Issue #1011, opened 2020-01-27
`home.sessionVariables` has no effect when using a graphical desktop environment


sigwinch28:
At the time of writing, the contents `home.sessionVariables` are placed in `~/.profile`.

As I use KDE Plasma, these variables never get loaded for me because:

- `/etc/profile` on NixOS 19.09 _does not_ source `~/.profile`
- sddm _does not_ source `~/.profile`
- Konsole and other terminal emulators start _non-login_ interactive shells by default (and therefore do not load `~/.bash_profile` or `~/.profile`, but `~/.bashrc` instead)

For example, if I set:
```
home.sessionVariables = { EDITOR = "emacs" };
```
I get:
```
[joe@mycomputer:~] $ echo $EDITOR
nano
```

My `~/.profile`, `~/.bash_profile`, and `~/.bashrc` files are all managed by home-manager and their contents are as follows:

```
# ~/.profile
# -*- mode: sh -*-

. "/home/joe/.nix-profile/etc/profile.d/hm-session-vars.sh"

```

```
# ~/.bash_profile
# -*- mode: sh -*-

# include .profile if it exists
[[ -f ~/.profile ]] && . ~/.profile

# include .bashrc if it exists
[[ -f ~/.bashrc ]] && . ~/.bashrc

```

```
# ~/.bashrc
# -*- mode: sh -*-

# Commands that should be applied only for interactive shells.
if [[ $- == *i* ]]; then
  HISTFILE="$HOME/.bash_history"
HISTFILESIZE=100000
HISTSIZE=10000

  shopt -s histappend
shopt -s checkwinsize
shopt -s extglob
shopt -s globstar
shopt -s checkjobs

  
fi

```

To make matters worse, it seems that NixOS doesn't make an authoritative decision about whether `~/.profile` (or actually, whether any other file) will be reliably sourced at login when using a graphical environment (https://github.com/NixOS/nixpkgs/issues/5200)

If I log in via the console or via SSH however, this _does_ count as a bash login shell and my environment variables are set correctly.

For now my current workaround is to put the following in my system's `/etc/nixos/configuration.nix` which applies a workaround at the system level:
```
environment.loginShellInit = ''
	if [ -e $HOME/.profile ]
	then
		. $HOME/.profile
	fi
'';
```

but I could just as easily put the following in my `home.nix`:
```
home.file.".xprofile".text = ''
	if [ -e $HOME/.profile ]
	then
		. $HOME/.profile
	fi
'';
```

[...]

PR #2681, opened 2022-02-02
faq: explain how to let HM manage shell profiles

### LINKED ISSUE #2680: Document how to manage shell profile
I saw https://github.com/nix-community/home-manager/blob/eee807560b4c2c7cdda00913056429d85a090eeb/docs/faq.adoc#L48 but I can't find how to let home-manager handle my .bash_profile and how to inject the existing code that was added over the past few decades.

wmertens:
fixes #2680


mainrs (approved):
LGTM

docs/faq.adoc
```diff
@@ -45,7 +45,7 @@ You could also opt to unistall _all_ of the packages from your profile with `nix
 === Why are the session variables not set?
 :foreign-env: https://github.com/oh-my-fish/plugin-foreign-env
 
-Home Manager is only able to set session variables automatically if it manages your Bash, Z shell, or fish shell configuration. If you don't want to let Home Manager manage your shell then you will have to manually source the `~/.nix-profile/etc/profile.d/hm-session-vars.sh` file in an appropriate way. In Bash and Z shell this can be done by adding
+Home Manager is only able to set session variables automatically if it manages your Bash, Z shell, or fish shell configuration (use `programs.bash`, `programs.zsh` or programs.fish` for that, see `man home-configuration.nix`). If you don't want to let Home Manager manage your shell then you will have to manually source the `~/.nix-profile/etc/profile.d/hm-session-vars.sh` file in an appropriate way. In Bash and Z shell this can be done by adding
```
berbiche, line 48:
```suggestion
Home Manager is only able to set session variables automatically if it manages your Bash, Z shell, or fish shell configuration (use <<opt-programs.bash>>, <<opt-programs.zsh>> or <<opt-programs.fish>> for that, see `man home-configuration.nix`). If you don't want to let Home Manager manage your shell then you will have to manually source the `~/.nix-profile/etc/profile.d/hm-session-vars.sh` file in an appropriate way. In Bash and Z shell this can be done by adding
```

ShamrockLee:
IIUC, the `shell` management for the session variables to work refers not only to the text-based shells (such as the Bash, Fish and Zsh login shell resides in tty1-tty6), but also to the graphical shell. The latter is especially critical for desktop users.

As the choice of graphical shells provided by HM is not as much as that by the NixOS configuration, users like me might choose not to let HM manage my graphical session, causing `home.sessionVariables` to be unavailable to graphical applications (including the terminal emulators). (#1011)

This might be workaround by injecting `config.home.sessionVariable` into the newly added `systemd.user.sessionVariables` (need someone to try that). Another terminal-emulator-only workaround is to inject those variables into `programs.{shell}.sessionVariables`.

It would be great if these can be documented, as the issue prevents a lot of modules that depends on `home.sessionVariables` from working properly.

ShamrockLee (commented):
Direct linking makes the document friendlier from the users' perspective.

rycee:
Thanks! Merged to master now.

PR #2683, opened 2022-02-02
pubs: add module

loicreynier:
### Description

Add [pubs](github.com/pubs/pubs) program module bringing basic configuration via `programs.pubs.config`. 

pubs configuration file uses ConfigObj syntax, which is similar to the INI files syntax but with extra functionnalities like nested sections. These functionnalities are used within pubs configuration file and prevent from using Nix's INI format parser. Here is an example of pubs configuration that cannot be parsed using Nix's INI format parser:

```ini
[plugins]

[[git]]
manual=False
```

For this reason, I opted for a stringly-typed configuration since the use of a structured `settings` option would require a custom parser.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -201,6 +201,9 @@
 
 /modules/programs/powerline-go.nix                    @DamienCassou
 
+/modules/programs/pubs.nix                            @loicreynier
+/tests/modules/programs/pubs                          @loicreynier
+
 /modules/programs/rbw.nix                             @ambroisie
 /tests/modules/programs/rbw                           @ambroisie
 
--- a/modules/lib/maintainers.nix
+++ b/modules/lib/maintainers.nix
@@ -83,6 +83,12 @@
     github = "kubukoz";
     githubId = 894884;
   };
+  loicreynier = {
+    name = "Loïc Reynier";
+    email = "loic@loireynier.fr";
+    github = "loicreynier";
+    githubId = 88983487;
+  };
   matrss = {
     name = "Matthias Riße";
     email = "matrss@users.noreply.github.com";
--- a/modules/programs/pubs.nix
+++ b/modules/programs/pubs.nix
@@ -1,0 +1,60 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.programs.pubs;
+
+in {
+  meta.maintainers = [ hm.maintainers.loicreynier ];
+
+  options.programs.pubs = {
+    enable = mkEnableOption "pubs";
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.pubs;
+      defaultText = literalExpression "pkgs.pubs";
+      description = "The package to use for the pubs script.";
+    };
+
+    extraConfig = mkOption {
+      type = types.lines;
+      default = "";
+      example = literalExpression ''
+        '''
+        [main]
+        pubsdir = ''${config.home.homeDirectory}/.pubs
+        docsdir = ''${config.home.homeDirectory}/.pubs/doc
+        doc_add = link
+        open_cmd = xdg-open
+
+        [plugins]
+        active = git,alias
+
+        [[alias]]
+
+        [[[la]]]
+        command = list -a
+        description = lists papers in lexicographic order
+
+        [[git]]
+        quiet = True
+        manual = False
+        force_color = False
+        ''''';
+      description = ''
+        Configuration using syntax written to
+        <filename>$HOME/.pubsrc</filename>.
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    home.file.".pubsrc" =
+      mkIf (cfg.extraConfig != "") { text = cfg.extraConfig; };
+  };
+}
--- a/tests/default.nix
+++ b/tests/default.nix
@@ -86,6 +86,7 @@ import nmt {
     ./modules/programs/pandoc
     ./modules/programs/pet
     ./modules/programs/powerline-go
+    ./modules/programs/pubs
     ./modules/programs/qutebrowser
     ./modules/programs/readline
     ./modules/programs/sagemath
--- a/tests/modules/programs/pubs/pubs-example-settings.nix
+++ b/tests/modules/programs/pubs/pubs-example-settings.nix
@@ -1,0 +1,36 @@
+{ config, lib, pkgs, ... }:
+
+{
+  programs.pubs = {
+    enable = true;
+
+    extraConfig = ''
+      [main]
+      pubsdir = ~/.pubs
+      docsdir = ~/.pubs/doc
+      doc_add = link
+      open_cmd = xdg-open
+
+      [plugins]
+      active = git,alias
+
+      [[alias]]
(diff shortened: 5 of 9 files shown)


modules/programs/pubs.nix
```diff
@@ -0,0 +1,58 @@
...
+    enable = mkEnableOption "pubs";
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.pubs;
+      defaultText = literalExpression "pkgs.pubs";
+      description = "The package to use for the pubs script.";
+    };
+
+    config = mkOption {
```
rycee, line 22: I would suggest `extraConfig` instead to better correspond with other similar options within Home Manager:
```suggestion
    extraConfig = mkOption {
```

modules/programs/pubs.nix
```diff
@@ -0,0 +1,58 @@
...
+        Configuration using syntax written to
+        <filename>$HOME/.pubsrc</filename>.
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    home.file.".pubsrc" = mkIf (cfg.config != "") { source = configFile; };
```
rycee, line 56: You can simplify this (and make it more general in the process) by assigning to the `text` field:
```suggestion
    home.file.".pubsrc" = mkIf (cfg.config != "") { text = cfg.config; };
```

modules/programs/pubs.nix
```diff
@@ -0,0 +1,58 @@
...
+      description = "The package to use for the pubs script.";
+    };
+
+    config = mkOption {
+      type = types.lines;
+      default = "";
+      example = ''
+        [main]
+        pubsdir = ${config.home.homeDirectory}/.pubs
+        docsdir = ${config.home.homeDirectory}/.pubs/doc
```
rycee, line 28: The example must not have a dependency on your configuration since it can cause unnecessary documentation builds. I think you want
```suggestion
        pubsdir = ''${config.home.homeDirectory}/.pubs
        docsdir = ''${config.home.homeDirectory}/.pubs/doc
```
instead.

rycee:
Thanks for the contribution! Added a few comments.

loicreynier:
Thanks for the comments, I implemented your revisions.

loicreynier:
I added a test to check configuration generation.

modules/programs/pubs.nix
```diff
@@ -0,0 +1,57 @@
...
+    enable = mkEnableOption "pubs";
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.pubs;
+      defaultText = literalExpression "pkgs.pubs";
+      description = "The package to use for the pubs script.";
+    };
+
+    extraConfig = mkOption {
```
sumnerevans, line 20: This is INI format, right? We should use `pkgs.formats.ini` if so (see the swaynag module for an example).

loicreynier: Not exactly, this is ConfigObj syntax which is an extension of the INI syntax with features such as `[[nested subsection]]` which are not supported by `pkgs.formats.ini`.

.github/CODEOWNERS
```diff
@@ -437,3 +437,5 @@ Makefile                                              @thiagokokada
 /modules/programs/ion.nix                             @jo1gi
 
 /modules/services/plex-mpv-shim.nix                   @starcraft66
+
+/modules/programs/pubs.nix                            @loicreynier
```
berbiche, line 441: Can you move this to the right place to preserve the ordering? Thanks.

modules/programs/pubs.nix
```diff
@@ -0,0 +1,57 @@
...
+
+with lib;
+
+let
+
+  cfg = config.programs.pubs;
+
+in {
+  options.programs.pubs = {
+    enable = mkEnableOption "pubs";
```
berbiche, line 11:
```suggestion
in {
  meta.maintainers = [ hm.maintainers.loicreynier ];

  options.programs.pubs = {
    enable = mkEnableOption "pubs";
```

modules/programs/pubs.nix
```diff
@@ -0,0 +1,57 @@
...
+
+        [[[la]]]
+        command = list -a
+        description = lists papers in lexicographic order
+
+        [[git]]
+        quiet = True
+        manual = False
+        force_color = False
+      '';
```
berbiche, line 43: IIRC multiline string examples still need to use `literalExpression`.

```suggestion
      example = literalExpression '''''
        [main]
        pubsdir = ''${config.home.homeDirectory}/.pubs
        docsdir = ''${config.home.homeDirectory}/.pubs/doc
        doc_add = link
        open_cmd = xdg-open

        [plugins]
        active = git,alias

        [[alias]]

        [[[la]]]
        command = list -a
        description = lists papers in lexicographic order

        [[git]]
        quiet = True
        manual = False
        force_color = False
      ''''';
```

tests/default.nix
```diff
@@ -166,5 +166,6 @@ import nmt {
     ./modules/services/xsettingsd
     ./modules/systemd
     ./modules/targets-linux
+    ./modules/programs/pubs
```
berbiche, line 169: Can you also preserve the ordering here?

tests/modules/programs/pubs/pubs-example-settings.nix
```diff
@@ -0,0 +1,36 @@
...
+        command = list -a
+        description = lists papers in lexicographic order
+
+        [[git]]
+        quiet = True
+        manual = False
+        force_color = False
+      '';
+    };
+
```
berbiche, line 30:
```suggestion
  test.stubs.pubs = { };
```

The package needs to be stubbed otherwise it will be fetched when running tests.

berbiche (changes requested):
Thanks!

I left a few comments for additional things to fix.

loicreynier (commented):
Thanks, I implemented your changes.

modules/lib/maintainers.nix
```diff
@@ -223,4 +223,9 @@
     github = "kmaasrud";
     githubId = 54394333;
   };
+    loicreynier = {
+    name = "Loïc Reynier";
+    email = "loic@loicreynier.fr";
+    github = "loicreynier";
+    githubId = 88983487;
```
berbiche, line 230:
```suggestion
  loicreynier = {
    name = "Loïc Reynier";
    email = "loic@loicreynier.fr";
    github = "loicreynier";
    githubId = 88983487;
  };
```

There seems to be a missing closing curly brace.

modules/programs/pubs.nix
```diff
@@ -0,0 +1,60 @@
...
+
+                [[[la]]]
+                command = list -a
+                description = lists papers in lexicographic order
+
+                [[git]]
+                quiet = True
+                manual = False
+                force_color = False
+              ''''';
```
berbiche, line 46: Can you reindent the example? Thanks

berbiche (changes requested):
Sorry for the delay!

loicreynier:
Whoops sorry for the missing bracket. The new indentation is now accepted by `./format`.

rycee:
Merged to master now. Thanks!

PR #2684, opened 2022-02-02
eww: add module

### LINKED ISSUE #2682: module request: eww
### Description

[eww](https://github.com/elkowar/eww) allows users to create custom widgets that work under any window manager. A configuration contains the following:

- `eww.yuck` file that defines the UI
- `eww.scss` file that contains the styling
- user-written scripts that might be used to fetch data.

### Notes

Apparently it isn't _that_ easy as the main configuration file, `eww.yuck`, can actually import other `.yuck` files as well. So a simple option of type `types.str` wouldn't be enough. The same goes for `scripts`. A user can have an arbitrary number of scripts that they call inside of `eww.yuck`. There would need to be a mechanism in-place that would allow one to add any number of files or maybe even a whole file hierarchy.

mainrs:
Closes #2682.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -76,6 +76,8 @@
 
 /modules/programs/emacs.nix                           @rycee
 
+/modules/programs/eww.nix                             @mainrs
+
 /modules/programs/exa.nix                             @kalhauge
 
 /modules/programs/firefox.nix                         @rycee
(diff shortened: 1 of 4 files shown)


rycee:
Thanks! Merged to master now.

PR #2686, opened 2022-02-02
zellij: add module

### LINKED ISSUE #2666: module request: zellij
### Description

[zellij](https://github.com/zellij-org/zellij) is terminal workspace written in Rust. It supports configuration via a rather large configuration file. It would be great if home-manager would contain a module for zellij

mainrs:
Closes #2666

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -245,6 +245,8 @@
 
 /modules/programs/zathura.nix                         @rprospero
 
+/modules/programs/zellij.nix                          @mainrs
+
 /modules/programs/zoxide.nix                          @marsam
 
 /modules/programs/zsh/prezto.nix                      @NickHu
--- a/modules/programs/zellij.nix
+++ b/modules/programs/zellij.nix
@@ -1,0 +1,51 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.programs.zellij;
+  yamlFormat = pkgs.formats.yaml { };
+
+in {
+  meta.maintainers = [ hm.maintainers.mainrs ];
+
+  options.programs.zellij = {
+    enable = mkEnableOption "zellij";
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.zellij;
+      defaultText = literalExpression "pkgs.zellij";
+      description = ''
+        The zellij package to install.
+      '';
+    };
+
+    settings = mkOption {
+      type = yamlFormat.type;
+      default = { };
+      example = literalExpression ''
+        {
+          theme = "custom";
+          themes.custom.fg = 5;
+        }
+      '';
+      description = ''
+        Configuration written to
+        <filename>$XDG_CONFIG_HOME/zellij/config.yaml</filename>.
+        </para><para>
+        See <link xlink:href="https://zellij.dev/documentation" /> for the full
+        list of options.
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    xdg.configFile."zellij/config.yaml" = mkIf (cfg.settings != { }) {
+      source = yamlFormat.generate "zellij.yaml" cfg.settings;
+    };
+  };
+}
--- a/modules/lib/maintainers.nix
+++ b/modules/lib/maintainers.nix
@@ -205,4 +205,10 @@
     github = "pinage404";
     githubId = 6325757;
   };
+  mainrs = {
+    name = "mainrs";
+    email = "5113257+mainrs@users.noreply.github.com";
+    github = "mainrs";
+    githubId = 5113257;
+  };
 }
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -2403,6 +2403,13 @@ in
           A new module is available: 'services.twmn'.
         '';
       }
+
+      {
+        time = "2022-02-16T23:50:35+00:00";
+        message = ''
+          A new module is available: 'programs.zellij'.
+        '';
+      }
     ];
   };
 }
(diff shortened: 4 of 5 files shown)


modules/programs/zellij.nix
```diff
@@ -0,0 +1,52 @@
...
+        type = yamlFormat.type;
+        default = { };
+        defaultText = "{}";
+        example = literalExpression ''
+          {
+            theme = "custom";
+            themes.custom.fg = 5;
+          }
+        '';
+        description = "The zellij configuration.";
```
berbiche, line 33:
```suggestion
        description = ''
          Configuration written to
          <filename>$XDG_CONFIG_HOME/zellij/config.toml</filename>.
          </para><para>
          See <link xlink:href="https://www.zellij.sh/book/configuration.html" /> for the full list
          of options.
        '';
```

modules/programs/zellij.nix
```diff
@@ -0,0 +1,52 @@
...
+    };
+  };
+
+  meta = {
+    description = "A terminal workspace with batteries included.";
+    homepage = "https://zellij.dev/";
+    maintainers = [ maintainers.mainrs ];
+    platforms = platforms.unix;
+    license = licenses.mit;
+  };
```
berbiche, line 51:
```suggestion
  meta.maintainers = [ maintainers.mainrs ];
```
AFAIK the module system does not use/have other meta attributes.

berbiche (changes requested):
Thanks for the PR!

modules/programs/zellij.nix
mainrs: I'll change this in one of my other PRs for `eww` as well! Thanks for letting me know!

happysalada:
@berbiche friendly ping (no worries if you're busy and no rush.), I'm interested by this.

modules/programs/zellij.nix
berbiche: Yup, the description is contained within a `<para>graph` and is automatically closed by one.
Adding new paragraphs is equivalent to adding a linebreak `<br>`.

modules/programs/zellij.nix
```diff
@@ -0,0 +1,52 @@
...
+
+      settings = mkOption {
+        type = yamlFormat.type;
+        default = { };
+        defaultText = "{}";
+        example = literalExpression ''
+          {
+            theme = "custom";
+            themes.custom.fg = 5;
+          }
```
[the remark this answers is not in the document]
mainrs: I took a look at the code and it uses the normal HashMap implementation which doesn’t make ordering guarantees. So I’d say no, it doesn’t matter.

.github/CODEOWNERS
```diff
@@ -428,3 +428,5 @@ Makefile                                              @thiagokokada
 /modules/programs/ion.nix                             @jo1gi
 
 /modules/services/plex-mpv-shim.nix                   @starcraft66
+
+/modules/programs/zellij.nix                          @mainrs
```
sumnerevans, line 432: This should go up with the other `/modules/programs`. Not sure why ion is down here, but it shouldn't be either.

modules/programs/zellij.nix
```diff
@@ -0,0 +1,52 @@
...
+        example = literalExpression "pkgs.zellij";
+        description = ''
+          The zellij package to install.
+        '';
+      };
+
+      settings = mkOption {
+        type = yamlFormat.type;
+        default = { };
+        defaultText = "{}";
```
sumnerevans, line 26:
```suggestion
        defaultText = literalExpression "{ }";
```

sumnerevans (changes requested):

.github/CODEOWNERS
mainrs: I don't think I read somewhere that it had to be alphabetically sorted. Might be worth doing that once in a separate PR and updating `CONTRIBUTING`. I might have missed it though.

berbiche (approved):
LGTM!

modules/programs/zellij.nix
```diff
@@ -0,0 +1,52 @@
...
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+    xdg.configFile."zellij/config.yaml" = mkIf (cfg.settings != { }) {
+      source = yamlFormat.generate "zellij.yaml" cfg.settings;
+    };
+  };
+
+  meta.maintainers = [ maintainers.mainrs ];
```
berbiche, line 51:
```suggestion
  meta.maintainers = [ hm.maintainers.mainrs ];
```

I missed this.

rycee:
Merged to master now. Thanks!

happysalada:
in the zellij docs, there are ways to unbind certain keybindings

the conf should be 
```
keybinds:
    unbind: [ Ctrl: 'p']
```

for the love of everything that is holy, I couldn't find the equivalent nix expression that would yield the corresponding yaml.

this is what I tried
```
keybinds = {
   unbind = [ "Ctrl: 'p'"];
};
```
as soon as I try to uncomment inside the square brackets, it doesn't preserve them.

[reference](https://zellij.dev/documentation/keybindings.html)

berbiche:
@happysalada 
> in the zellij docs, there are ways to unbind certain keybindings
> 
> the conf should be 
> ```
> keybinds:
>     unbind: [ Ctrl: 'p']
> ```

YAML allows using JSON notation for inline arrays (instead of using `-`).
The content of `unbind` is actually an array of one element, an object with key `Ctrl` and value `"p"`.

I think this should be what you are looking for:
``` nix
keybinds.unbind = [ { "Ctrl" = "p"; } ];
```

happysalada:
The formatting is not the same, which got me confused at first, but it actually works!
Thanks a lot for this!

PR #2688, opened 2022-02-02
man: add package option

### LINKED ISSUE #2634: `programs.man` should have a `package` option
### Description

The `programs.man` module should have a `package` option that lets me change the package added to `home.packages`. This should not affect `programs.man.generateCaches` as that uses `pkgs.man-db` directly anyway rather than the `pkgs.man` that is put in `home.packages`.

The reason here is I need to wrap `bin/man` to set a custom `MANPATH` (macOS `man` inserts Apple developer tools libraries into the default path but man-db doesn't know how to do this), but using an overlay to replace the package forces rebuilds of dependents (in my case that causes 67 rebuilt derivations). I can't just set this as a session variable as it needs to be calculated every time `man` is invoked (other env vars and system configuration influence the resulting developer tools paths).

mainrs:
Closes #2634

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/man.nix
+++ b/modules/programs/man.nix
@@ -2,7 +2,8 @@
 
 with lib;
 
-{
+let cfg = config.programs.man;
+in {
   options = {
     programs.man = {
       enable = mkOption {
@@ -13,6 +14,13 @@ with lib;
           command. This also includes "man" outputs of all
           <literal>home.packages</literal>.
         '';
+      };
+
+      package = mkOption {
+        type = types.package;
+        default = pkgs.man;
+        defaultText = literalExpression "pkgs.man";
+        description = "The man package to use.";
       };
 
       generateCaches = mkOption {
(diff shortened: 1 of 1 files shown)


modules/programs/man.nix
```diff
@@ -39,7 +46,7 @@ with lib;
   };
 
   config = mkIf config.programs.man.enable {
-    home.packages = [ pkgs.man ];
+    home.packages = [ cfg.package ];
```
berbiche, line 49:
```suggestion
    home.packages = [ config.programs.man.package ];
```

Or you can add
``` nix
let
  cfg = config.programs.man;
in
```
at the top.

berbiche (changes requested):
Thanks for the PR!

modules/programs/man.nix
mainrs: Yep, I missed that!

berbiche (approved):

PR #2690, opened 2022-02-02
chore: add question issue template

### LINKED ISSUE #2525: Create a template to ask support questions
### Description

We need a template for users to ask questions instead of submitting bug requests.

What should the template look like?

It should contain at least the following things:
- whether the user is using flakes (using a checkbox?)
- system information (NixOS / Darwin / generic Linux)
- standalone configuration or HM module

mainrs:
Closes #2525. An alternative would be to open up GitHub Discussions.

### DIFF
--- a/.github/ISSUE_TEMPLATE/question.yaml
+++ b/.github/ISSUE_TEMPLATE/question.yaml
@@ -1,0 +1,47 @@
+name: Question
+description: Ask a question
+title: 'question: '
+labels: [question]
+
+body:
+- type: textarea
+  id: question
+  attributes:
+    label: Question
+  validations:
+    required: true
+- type: textarea
+  id: maintainers
+  attributes:
+    label: Maintainer CC
+    description: |
+      Please @ people who are in the `meta.maintainers` list of the module in question.
+      If in doubt, check `git blame` for whoever last touched something.
+  validations:
+    required: false
+
+- type: markdown
+  attributes:
+    value: |
+      Information about how you use nix and home-manager and your system that runs them.
+- type: textarea
+  id: system
+  attributes:
+    label: System
+    description: Please run `nix-shell -p nix-info --run "nix-info -m"` and paste the output below.
+    render: markdown
+  validations:
+    required: true
+- type: checkboxes
+  id: usage
+  attributes:
+    label: Usage
+    options:
+    - label: Do you use nix flakes?
+      required: false
+    - label: Do you use home-manager as a standalone program?
+      required: false
+    - label: Do you use home-manager as a NixOS module?
+      required: false
+  validations:
+    required: false


.github/ISSUE_TEMPLATE/question.yaml
```diff
@@ -0,0 +1,47 @@
+name: Question
+description: Ask a question
+title: 'question: '
+labels: [question]
+
+body:
```
berbiche, line 6:
```suggestion
name: Question
description: Ask a question
title: 'question: '
labels: [question]

assignees: [rycee, berbiche]

body:
```

.github/ISSUE_TEMPLATE/question.yaml
```diff
@@ -0,0 +1,47 @@
...
+  id: usage
+  attributes:
+    label: Usage
+    options:
+    - label: Do you use nix flakes?
+      required: false
+    - label: Do you use home-manager as a standalone program?
+      required: false
+    - label: Do you use home-manager as a NixOS module?
+      required: false
```
berbiche, line 45:
```suggestion
    - label: Do you use Home Manager as a standalone program?
      required: false
    - label: Do you use Home Manager as a NixOS module?
      required: false
    - label: Do you use Home Manager as a nix-darwin module?
      required: false
```

.github/ISSUE_TEMPLATE/question.yaml
```diff
@@ -0,0 +1,47 @@
...
+    description: |
+      Please @ people who are in the `meta.maintainers` list of the module in question.
+      If in doubt, check `git blame` for whoever last touched something.
+  validations:
+    required: false
+
+- type: markdown
+  attributes:
+    value: |
+      Information about how you use nix and home-manager and your system that runs them.
```
berbiche, line 26:
```suggestion
      Information about how you use Nix and Home Manager and your system that runs them.
```

.github/ISSUE_TEMPLATE/question.yaml
```diff
@@ -0,0 +1,47 @@
...
+    required: true
+- type: textarea
+  id: maintainers
+  attributes:
+    label: Maintainer CC
+    description: |
+      Please @ people who are in the `meta.maintainers` list of the module in question.
+      If in doubt, check `git blame` for whoever last touched something.
+  validations:
+    required: false
```
berbiche, line 21:
```suggestion
    required: true
```

berbiche (changes requested):
Thanks for the PR!

mainrs:
It might be worth exploring the idea of having a workflow that automatically mentions the right persons for a given module. Maybe by having an input field with the module name instead.

.github/ISSUE_TEMPLATE/question.yaml
mainrs: I did apply the change. But it’s worth taking into account that some questions might relate to modules or internal library functions that do not have a maintainer. 

Having users enter a space might still be the better option though.

.github/ISSUE_TEMPLATE/question.yaml
```diff
@@ -0,0 +1,51 @@
...
+    description: |
+      Please @ people who are in the `meta.maintainers` list of the module in question.
+      If in doubt, check `git blame` for whoever last touched something.
+  validations:
+    required: true
+
+- type: markdown
+  attributes:
+    value: |
+      Information about how you use Nix and Home Manager and your system that runs them.
```
berbiche, line 28: Maybe the markdown should be a title?

image

berbiche: Correction: can this sentence be moved before "Please run ..."?

mainrs: @berbiche Which sentence? Should I move the `Information about...` part right before the `Please run...` part?

berbiche: The one above "System": "Information about how you use ..."

sumnerevans:
I think I'd prefer to just enable GitHub Discussions, but it's really up to @rycee

rycee:
I've seen people asking HM related questions under the "Help" category in the [NixOS Discourse](https://discourse.nixos.org/) instance and that seems like a fine place to me. If the volume gets too large then perhaps it would be possible to create a HM specific Help category?

@zimbatm Would it be OK to add a reference to the Discourse's "Help" topic in the Home Manager documentation?

khaneliman:
We've enabled discussions.

PR #2706, opened 2022-02-07
recoll: add module

foo-dogsquared:
### Description

Added [Recoll](https://www.lesbonscomptes.com/recoll/), a file index service for the desktop. It is oriented to be used by a normal user so it is a good addition to home-manager.

I use this module myself for two weeks but I don't feel confident for the custom settings format I created for this module so feel free to roast it. :)

### Checklist

Recoll

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/recoll.nix
+++ b/modules/services/recoll.nix
@@ -1,0 +1,192 @@
+{ config, options, lib, pkgs, ... }:
+
+with lib;
+
+# TODO: Fix the formatting of the resulting config.
+let
+  cfg = config.services.recoll;
+
+  # The key-value generator for Recoll config format. For future references,
+  # see the example configuration from the package (i.e.,
+  # `$out/share/recoll/examples/recoll.conf`).
+  mkRecollConfKeyValue = generators.mkKeyValueDefault {
+    mkValueString = v:
+      if v == true then
+        "1"
+      else if v == false then
+        "0"
+      else if isList v then
+        concatStringsSep " " v
+      else
+        generators.mkValueStringDefault { } v;
+  } " = ";
+
+  # A modified version of 'lib.generators.toINI' that also accepts top-level
+  # attributes as non-attrsets.
+  toRecollConf = { listsAsDuplicateKeys ? false }:
+    attr:
+    let
+      toKeyValue = generators.toKeyValue {
+        inherit listsAsDuplicateKeys;
+        mkKeyValue = mkRecollConfKeyValue;
+      };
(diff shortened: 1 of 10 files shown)


modules/services/recoll.nix
```diff
@@ -0,0 +1,170 @@
...
+        Documentation = [
+          "man:recoll"
+          "man:recollindex"
+          "https://www.lesbonscomptes.com/recoll/usermanual/"
+        ];
+      };
+
+      Service = {
+        ExecStart = "${cfg.package}/bin/recollindex";
+        Environment = [ "RECOLL_CONFDIR=${cfg.configDir}" ];
```
rycee, line 152: To avoid misunderstandings if the path contains a space or something:

```suggestion
        Environment = [ "RECOLL_CONFDIR=${escapeShellArg cfg.configDir}" ];
```

modules/services/recoll.nix
```diff
@@ -0,0 +1,170 @@
...
+        mkKeyValue = mkRecollConfKeyValue;
+      };
+      mkSectionName = name: strings.escape [ "[" "]" ] name;
+      config = mapAttrsToList (k: v:
+        if isAttrs v then
+          ''
+            [${mkSectionName k}]
+          '' + toKeyValue v
+        else
+          toKeyValue { "${k}" = v; }) attr;
```
rycee, line 37: Perhaps this produces unexpected output when a top-level attribute name is greater than an non-top-level attribute name? I.e.

``` nix
{
  a = { foo = "bar"; };
  b = 10;
}
```
would, I guess, produce something like
``` ini
[a]
foo = "bar"

b = 10
```
In this case the `b` key in the INI would be interpreted as being part of `[a]`?

To avoid this I imagine one would have to map the attribute set to a list and then sort this so that attributes with non-attribute set values come first.

rycee:
Thanks for the contribution! Looks quite good in general, I've just added a few comments.

modules/services/recoll.nix
foo-dogsquared: Nice catch! In Recoll, once a section is defined, it seems to consider all value definitions under that section. At least this is how the behavior of the application goes in my tests. I've added this as `recoll-config-format-order` test.

The updated implementation is pretty naive but it pass the tests. My nix-fu is not as good as I thought. ;p

rycee:
Sorry about the delay. Looks good now so I've merged to master 🙂

PR #2708, opened 2022-02-08
zsh: move sessionVariables from .zshrc to .zshenv

### LINKED ISSUE #2445: bug: Session vars are loaded in different startup scripts in Bash and ZSH

### Issue description

I don't know if this behavior classifies as a bug, but there is certainly a discrepancy between how session is loaded between Bash and ZSH.

**History**

Looking through some history, I found that in #183 it was concluded that loading the session should be moved from `.zshenv` to somewhere else. `.zprofile` came up, but in #190 loading the `hm-session-vars.sh` file was moved to `.zshrc` instead.

I couldn't find the final argument for that decision, but [this comment](https://github.com/nix-community/home-manager/issues/183#issuecomment-359381272) states (incorrectly I believe) that `.zprofile` is sourced after `.zshrc` whereas it is sourced _before_ `.zshrc` and only in login shells.

As of today, `hm-session-vars.sh` is still sourced in [.zshrc](https://github.com/nix-community/home-manager/blob/70c5b268e10025c70823767f4fb49e240b40151d/modules/programs/zsh.nix#L495)

**What does Bash do?**
Bash has a similar concept of startup scripts:

- `.profile` (or some alternatives) are loaded in login shells
- `.bashrc` is loaded in interactive shells

As of today `hm-session-vars.sh` is sourced in [.profile](https://github.com/nix-community/home-manager/blob/70c5b268e10025c70823767f4fb49e240b40151d/modules/programs/bash.nix#L190)

As a side note, session variables passed to `programs.bash` are also added to `.profile`.

**Why is it important?**
As I hinted above, different startup scripts are loaded at different times:

- `.profile` and `.zprofile` are loaded in _login_ shells (when you login from terminal, open an ssh session, etc)
- `.bashrc` and `.zshrc` are loaded in _interactive_ shells

Putting graphical interfaces aside for a moment, there are four cases to be examined:

- interactive, login shell: they work the same way
- interactive, non-login shell: Bash will inherit env vars, ZSH will load them (essentially the same, given Bash is the child of a Bash login shell)
- non-interactive, non-login shell: Both will inherit env vars from parent
- non-interactive, login shell: Bash will load env vars, but ZSH will not

Whether loading env variables in login shells or non-login interactive shells for the first time is the correct solution is a question. Personally, I tend to lean towards loading env vars in login shells, because you should be able to set them once and then forget about them (again, putting graphical user interfaces aside for a moment)

**A word about graphical environments**
Some are capable of loading `.profile` (or alternative scripts in the HOME dir), so if someone uses Bash and a desktop environment that can load `.profile`, they are already in good place. Alternatively, they can always just source `.profile` in whatever script they use on GUI launch (`.xinitrc`, `.xprofile`, etc).

ZSH users will have to create either `.profile` or source `.zprofile`.

But I think this is also pointing towards using startup scripts that are loaded in login shells.

**Conclusion**

I can't say objectively that the current solution is wrong, only that I think using `.zprofile` would be a more correct place for session loading.

Also, currently Bash and ZSH differ in behavior: Bash loads session in `.profile` (login shells), ZSH in `.zshrc` (interactive shells).

What do you think?

### Maintainer CC

Pinging @uvNikita as he made the change referenced above

Pinging @rycee and @Infinisil as they were the other participants in the conversation referenced above

### System information

```markdown
❯ nix-shell -p nix-info --run "nix-info -m"
 - system: `"x86_64-darwin"`
 - host os: `Darwin 20.6.0, macOS 10.16`
 - multi-user?: `no`
 - sandbox: `no`
 - version: `nix-env (Nix) 2.5pre20211007_844dd90`
 - channels(marksk): `"home-manager, nixpkgs-21.11pre327016.550dab224a2"`
 - nixpkgs: `/Users/marksk/.nix-defexpr/channels/nixpkgs`
```

jian-lin:
### Description

This patch moves both home.sessionVariables and
programs.zsh.sessionVariables from .zshrc to .zshenv. Additionally,
these two kinds of session variables will not be sourced more than
once to allow user-customized ones to take effect.

Before, session variables are in .zshrc, which causes non-interactive
shells to not be able to get those variables. For example, running a
command through SSH is in a non-interactive and non-login shell, which
suffers from this. With this patch, all kinds of shells can get
session variables.

The reason why these session variables are not moved to .zprofile is
that programs started by systemd user instances are not able to get
variables defined in that file. For example, GNOME
Terminal (gnome-terminal-server.service) is one of these programs and
doesn't get variables defined in .zprofile. As a result, the shells it
starts, which are interactive and non-login, do not get those
variables.

Fixes #2445

Related https://github.com/NixOS/nixpkgs/issues/33219
Related https://github.com/NixOS/nixpkgs/pull/45784

This file is not formatted before and is excluded by ./format, so I don't format it.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/zsh.nix
+++ b/modules/programs/zsh.nix
@@ -443,6 +443,19 @@ in
     })
 
     {
+      home.file."${relToDotDir ".zshenv"}".text = ''
+        # Environment variables
+        . "${config.home.profileDirectory}/etc/profile.d/hm-session-vars.sh"
+
+        # Only source this once
+        if [[ -z "$__HM_ZSH_SESS_VARS_SOURCED" ]]; then
+          export __HM_ZSH_SESS_VARS_SOURCED=1
+          ${envVarsStr}
+        fi
+      '';
+    }
+
+    {
       home.packages = with pkgs; [ zsh ]
         ++ optional cfg.enableCompletion nix-zsh-completions
         ++ optional cfg.oh-my-zsh.enable oh-my-zsh;
@@ -490,10 +503,6 @@ in
         ${optionalString cfg.enableSyntaxHighlighting
           "source ${pkgs.zsh-syntax-highlighting}/share/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh"
         }
-
-        # Environment variables
-        . "${config.home.profileDirectory}/etc/profile.d/hm-session-vars.sh"
-        ${envVarsStr}
 
         ${optionalString cfg.oh-my-zsh.enable ''
             # oh-my-zsh extra settings for plugins
--- a/tests/modules/programs/zsh/session-variables.nix
+++ b/tests/modules/programs/zsh/session-variables.nix
@@ -16,9 +16,9 @@ with lib;
     test.stubs.zsh = { };
 
     nmt.script = ''
-      assertFileExists home-files/.zshrc
-      assertFileRegex home-files/.zshrc 'export V1="v1"'
-      assertFileRegex home-files/.zshrc 'export V2="v2-v1"'
+      assertFileExists home-files/.zshenv
+      assertFileRegex home-files/.zshenv 'export V1="v1"'
+      assertFileRegex home-files/.zshenv 'export V2="v2-v1"'
     '';
   };
 }


jian-lin:
fixed test

jian-lin:
cc @berbiche as you recently merged several PRs related to zsh

teto (approved):
Makes sense to me. I will test this in my config.

clhodapp:
I believe that this is busted because it re-uses the same environment variable name as `hm-session-vars.sh`.

See from the code that generates `hm-session-vars.sh`:
https://github.com/nix-community/home-manager/blob/69536af27e86a9fc875d71cb9566ccccf47b5b60/modules/home-environment.nix#L547-L548

and the code that this puts into `.zshenv`:
https://github.com/nix-community/home-manager/blob/69536af27e86a9fc875d71cb9566ccccf47b5b60/modules/programs/zsh.nix#L447-L454

When `hm-session-vars.sh` gets sourced, it sets `__HM_ZSH_SESS_VARS_SOURCED` which causes the user's zsh-specific variables to get skipped!

clhodapp:
Never mind! I need to check my eyes! :eyeglasses: 

However! It does seem to be broken for some reason, as my prompt is no longer getting set... I'll post again if I find the real root cause

clhodapp:
Alrighty. It looks like the actual culprit was the prompt init code in `/etc/zshrc` that comes courtesy of the defaults from nixos:

https://github.com/NixOS/nixpkgs/blob/d5f237872975e6fb6f76eef1368b5634ffcd266f/nixos/modules/programs/zsh/zsh.nix#L90-L96

That, coupled with the fact that our variables were previously getting sourced _after_ that by virtue of being `~/.zshrc` but are now getting source _before_ it by virtue of being in `~/.zshenv`.

From the zsh man page:
```
STARTUP/SHUTDOWN FILES
       Commands  are  first read from /etc/zshenv; this cannot be overridden.  Subsequent behaviour is modified by the RCS and GLOBAL_RCS options; the former affects all startup files, while the second only affects global startup files (those shown here with an path starting with a /).
       If one of the options is unset at any point, any subsequent startup file(s) of the corresponding type will not be read.  It is also possible for a file in $ZDOTDIR to re-enable GLOBAL_RCS. Both RCS and GLOBAL_RCS are set by default.

       Commands are then read from $ZDOTDIR/.zshenv.  If the shell is a login shell, commands are read from /etc/zprofile and then $ZDOTDIR/.zprofile.  Then, if the shell is interactive, commands are read from /etc/zshrc and then $ZDOTDIR/.zshrc.  Finally,  if  the  shell  is  a  login
       shell, /etc/zlogin and $ZDOTDIR/.zlogin are read.

       When  a  login shell exits, the files $ZDOTDIR/.zlogout and then /etc/zlogout are read.  This happens with either an explicit exit via the exit or logout commands, or an implicit exit by reading end-of-file from the terminal.  However, if the shell terminates due to exec'ing an‐
       other process, the logout files are not read.  These are also affected by the RCS and GLOBAL_RCS options.  Note also that the RCS option affects the saving of history files, i.e. if RCS is unset when the shell exits, no history file will be saved.

       If ZDOTDIR is unset, HOME is used instead.  Files listed above as being in /etc may be in another directory, depending on the installation.

       As /etc/zshenv is run for all instances of zsh, it is important that it be kept as small as possible.  In particular, it is a good idea to put code that does not need to be run for every single shell behind a test of the form `if [[ -o rcs ]]; then ...' so that it  will  not  be
       executed when zsh is invoked with the `-f' option.

       Any of these files may be pre-compiled with the zcompile builtin command (see zshbuiltins(1)).  If a compiled file exists (named for the original file plus the .zwc extension) and it is newer than the original file, the compiled file will be used instead.

```

In the very short term, I've updated my system config to remove `prompt suse` from `programs.zsh.promptInit`.

jian-lin:
@clhodapp `PROMPT` should be set in `.zshrc` since it only needed by an interactive shell.

https://github.com/nix-community/home-manager/commit/2116fe6b50a5118d56f1f443cccf024abee9de40#r67162926

berbiche:
@NorthIsUp you could move your .zshenv to another file and source it: `programs.zsh.envExtra = ". $ZDOTDIR/my-zshenv";`

sagikazarmark:
I believe this PR broke `home.sessionPath`. 

Temporary solution:

```nix
{
    zsh.profileExtra = lib.optionalString (config.home.sessionPath != [ ]) ''
      export PATH="$PATH''${PATH:+:}${lib.concatStringsSep ":" config.home.sessionPath}"
    '';
}
```

Opened issue #2991

PR #2709, opened 2022-02-08
emacs: add an option for the shell launching Emacs to fix missing environment variable problem

### LINKED ISSUE #2139: Emacs service doesn't load environment variable unless seting programs.bash.enable.
### Issue description
In [services/emacs.nix](https://github.com/nix-community/home-manager/blob/0ada50fc9c620f7ad9f7c6ff70bf40514f4400a9/modules/services/emacs.nix#L97-L102), `pkgs.runtimeShell` is used to import user environment according comments. However `pkgs.runtimeShell` is hard coded to bash which means if user uses zsh and these environment such as `hm-session-var` will not be imported. 

A simple way to solve this is with `programs.bash.enable = true`. However, I don't think it is a good solution. Is there a more compatible method to do this?

#### Maintainer CC
@tadfisher

jian-lin:
### Description

This patch adds an option services.emacs.shell to choose the shell
launching an Emacs daemon.

This option defaults to bash (runtimeShell), which has backword
compatibility, and should be set to the user's shell to get the right
environment for Emacs.

Before, this shell is hardcoded to runtimeShell, i.e. bash. If a
user's shell is not bash and programs.bash.enable is false, the Emacs
daemon will miss some environment variables,
e.g. home.sessionVariables and programs.zsh.sessionVariables (if zsh
is the user's shell).

Fixes #2139 (for zsh, #2708 is also needed)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/emacs.nix
+++ b/modules/services/emacs.nix
@@ -44,6 +44,15 @@ in {
   options.services.emacs = {
     enable = mkEnableOption "the Emacs daemon";
 
+    shell = mkOption {
+      type = types.shellPackage;
+      default = pkgs.bash;
+      defaultText = literalExpression "pkgs.bash";
+      example = literalExpression "pkgs.zsh";
+      description =
+        "The shell that launches an Emacs daemon and should be the same as your shell.";
+    };
+
     package = mkOption {
       type = types.package;
       default = if emacsCfg.enable then emacsCfg.finalPackage else pkgs.emacs;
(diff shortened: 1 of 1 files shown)


jian-lin:
@tadfisher cc as you are the code owner

modules/services/emacs.nix
```diff
@@ -44,6 +44,15 @@ in {
   options.services.emacs = {
     enable = mkEnableOption "the Emacs daemon";
 
+    shell = mkOption {
+      type = types.shellPackage;
+      default = pkgs.bash;
```
otavio, line 49: Couldn't this use user's shell variable?

jian-lin: It's possible.

I just don't want to deal the weird quoting syntax with both shell and systemd. Writing a start script like what the nixos option `systemd.services.<name>.script` does is probably easier and clearer.

Feel free to open another PR. I would love to see either of them merged.

jian-lin:
After I switch to fish from zsh, this doesn't work.

PR #2712, opened 2022-02-09
mpv: allow `package = null`

loicreynier:
### Description

Allow to use `package = null` to generate MPV configuration files with Home Manager while using the system package / NixOS module.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/mpv.nix
+++ b/modules/programs/mpv.nix
@@ -49,10 +49,13 @@ let
   renderDefaultProfiles = profiles:
     renderOptions { profile = concatStringsSep "," profiles; };
 
-  mpvPackage = if cfg.scripts == [ ] then
-    cfg.package
+  mpvPackage = if cfg.package != null then
+    if cfg.scripts == [ ] then
+      cfg.package
+    else
+      pkgs.wrapMpv pkgs.mpv-unwrapped { scripts = cfg.scripts; }
   else
-    pkgs.wrapMpv pkgs.mpv-unwrapped { scripts = cfg.scripts; };
+    null;
 
 in {
   options = {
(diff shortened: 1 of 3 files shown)


SuperSandro2000:
Instead of overwriting the package to null this should be an extra option.

loicreynier:
Is there a naming convention for such an option?

SuperSandro2000:
I don't know. Maybe ``addToPath``? Overwriting anything to null is an anti pattern which always indicates that there could be an error.

rycee:
Why not just assign a derivation that produces an empty output to the `package` option?

loicreynier:
Great idea! However I do not how to implement such a derivation, I tried

```nix
package = pkgs.stdenv.mkDerivation {
  name = "mpv-dummy";
  dontUnpack = true;
  dontBuild = true;
  dontInstall = true;
};
``` 

which refuse to build:

```shell
error: builder for '/nix/store/...-mpv-dummy.drv' failed to produce output path for output 'out'
```

SuperSandro2000:
You need to `touch $out` in installPhase.

loicreynier:
`touch $out` in `installPhase` results in a `store path /nix/store/...-mpv-dummy is a file can't be merged into an environment using pkgs.buildEnv!` error. I fixed it using instead `mkdir $out`:

```nix
package = pkgs.stdenv.mkDerivation {
  name = "mpv-dummy";
  dontUnpack = true;
  dontBuild = true;
  installPhase = ''
    mkdir $out
  '';
};
```

The derivation is a better alternative since it can be implemented in every module without adding an option.

Issue #2324, opened 2021-09-13
Module request: Port the `nix` modules from NixOS


ShamrockLee:
### Description

The Nix package manager offers unprivileged users the ability to customize its behavior through configurations. Since the configuration can be done declaratively on NixOS, porting them to Home Manager would be beneficial to users.

Example use cases:

*   Use nixUnstable with flakes locally
`home.nix`
```Nix
{config, lib, pkgs, ...}:
{
  nix.enable = true;
  nix.package = pkgs.nixUnstable;
  nix.extraOptions = ''
    experimental-features = nix-command flakes
  '';
}
```

*   Custom registries

`flake.nix`
```Nix
{
  inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-20.09";
  inputs.nixpkgs-nixos-unstable.url = "github:NixOS/nixpkgs/nixos-unstable";
  inputs.home-manager = "github:nix-community/home-manager/release-20.09";
  inputs.home-manager.inputs.nixpkgs.follows = "nixpkgs";

  outputs = inputs@{ self, home-manager, ... }:
    let
      system = "x86_64-linux";
      username = "shamrock";
    in {
      homeManagerConfiguration.${username} = home-manager.lib.homeManagerConfiguration {
        configuration = import ./home.nix;
        inherit system username;
        homeDirectory = "/home/${username}";
        extraSpecialArgs = {
          # Inject the flakes here
          inherit (inputs) nixpkgs nixpkgs-nixos-unstable system-config-flake;
        };
      };
    };
}
```
`home.nix`
```Nix
args@{config, lib, pkgs, ... }: {
  nix = {
    enable = true;
    # Use the nix command inside the environment
    package = null;
    registry.local = {
      from = { type = "indirect"; id = "nixpkgs-local"; };
      flake = args.nixpkgs;
    };
    registry.unstable = {
      from = { type = "indirect"; id = "nixpkgs-local-unstable"; };
      flake = args.nixpkgs-unstable;
    };
  };
  # ...
}
```

    and then we can `nix search` the local nixpkgs input (without network connection) and `nix run` a program.
```console
$ nix search nixpkgs-local hello
* legacyPackages.x86_64-linux.gnome.iagno (3.38.1)
  Computer version of the game Reversi, more popularly called Othello

* legacyPackages.x86_64-linux.hello (2.10)
  A program that produces a familiar, friendly greeting

* legacyPackages.x86_64-linux.hello-unfree (1.0)
[...]

PR #2718, opened 2022-02-12
nix: add structural settings

polykernel:
Nix permits user level configurations through ~/.config/nix/nix.conf that allow
customization of system-wide settings and behavior. This is beneficial in chroot
environments and for per-user configurations. System level Nix configurations in the
form of /etc/nix/nix.conf can be specified declaratively via the NixOS nix module but as
of currently no counter part exists in home-manager.

This PR is a port of the RFC42 implementation for the NixOS nix module[1]
to home-manager. Non-applicable options have been excluded and the config generation
backends have been tweaked to the backends offered by home-manager. A notable change
from the NixOS module is a mandatory option to specify the Nix binary corresponding
to the version "nix.conf" should be generated against. This is necessary because
the validation phase is dependent on the `nix show-config` subcommand on the host platform.
While it is possible to avoid validation entirely, the lack of type checking was deemed too significant.
In NixOs, the version information can be retrieved from the `package` option itself which
declares the Nix binary system-wide. However in home-manager, there is no pure way to
detect the system Nix version and what state version the "nix.conf" should be generated
against. Thus an option is used to overcome this limitation by forcing the user to
specify the Nix package. Note this interaction can still be automated by forwarding
the system-wide Nix package to the home-manager module if needed.

Three unit tests were added to test the module behavior for the empty settings, the example
settings and the example registry configurations respectively.

[1] - https://github.com/NixOS/nixpkgs/pull/139075

### Description
Efforts towards https://github.com/nix-community/home-manager/issues/2324

Todo

- [x] Write unit tests

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/nix.nix
+++ b/modules/misc/nix.nix
@@ -6,8 +6,98 @@ let
 
   cfg = config.nix;
 
+  nixPackage = cfg.package;
+
+  isNixAtLeast = versionAtLeast (getVersion nixPackage);
+
+  nixConf = assert isNixAtLeast "2.2";
+    let
+
+      mkValueString = v:
+        if v == null then
+          ""
+        else if isInt v then
+          toString v
+        else if isBool v then
+          boolToString v
+        else if isFloat v then
+          floatToString v
+        else if isList v then
+          toString v
+        else if isDerivation v then
+          toString v
+        else if builtins.isPath v then
+          toString v
+        else if isString v then
+          v
+        else if isCoercibleToString v then
+          toString v
+        else
+          abort "The nix conf value: ${toPretty { } v} can not be encoded";
+
+      mkKeyValue = k: v: "${escape [ "=" ] k} = ${mkValueString v}";
+
+      mkKeyValuePairs = attrs:
+        concatStringsSep "\n" (mapAttrsToList mkKeyValue attrs);
+
+    in pkgs.writeTextFile {
+      name = "nix.conf";
+      text = ''
+        # WARNING: this file is generated from the nix.settings option in
+        # your Home Manager configuration at $XDG_CONFIG_HOME/nix/nix.conf.
+        # Do not edit it!
+        ${mkKeyValuePairs cfg.settings}
+        ${cfg.extraOptions}
+      '';
+      checkPhase =
+        if pkgs.stdenv.hostPlatform != pkgs.stdenv.buildPlatform then ''
+          echo "Ignoring validation for cross-compilation"
+        '' else ''
+          echo "Validating generated nix.conf"
+          ln -s $out ./nix.conf
+          set -e
+          set +o pipefail
+          NIX_CONF_DIR=$PWD \
+            ${cfg.package}/bin/nix show-config ${
+              optionalString (isNixAtLeast "2.3pre")
+              "--no-net --option experimental-features nix-command"
+            } \
+            |& sed -e 's/^warning:/error:/' \
+            | (! grep '${
+              if cfg.checkConfig then "^error:" else "^error: unknown setting"
+            }')
+          set -o pipefail
+        '';
+    };
+
+  semanticConfType = with types;
+    let
+      confAtom = nullOr (oneOf [ bool int float str path package ]) // {
+        description =
+          "Nix config atom (null, bool, int, float, str, path or package)";
+      };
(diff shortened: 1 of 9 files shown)


tests/modules/misc/nix/example-settings-expected.conf
```diff
@@ -0,0 +1,7 @@
+# WARNING: this file is generated from the nix.settings option in
+# your home-manager configuration at $XDG_CONFIG_HOME/nix/nix.conf.
```
berbiche, line 2:
```suggestion
# your Home Manager configuration at $XDG_CONFIG_HOME/nix/nix.conf.
```

modules/misc/nix.nix
```diff
@@ -68,19 +158,81 @@ in {
...
   };
 
-  config = mkIf (cfg.registry != { }) {
-    xdg.configFile."nix/registry.json".text = builtins.toJSON {
-      version = cfg.registryVersion;
-      flakes =
-        mapAttrsToList (n: v: { inherit (v) from to exact; }) cfg.registry;
+  config = mkIf cfg.enable {
+    assertions = [{
+      assertion = cfg.settings == { } || cfg.package != null;
```
berbiche, line 217:
```suggestion
      assertion = cfg.settings != { } -> cfg.package != null;
```
Up to you :smile:

modules/misc/nix.nix
```diff
@@ -68,19 +158,81 @@ in {
...
+    xdg.configFile = {
+      "nix/registry.json" = mkIf (cfg.registry != { }) {
+        source = jsonFormat.generate "registry.json" {
+          version = cfg.registryVersion;
+          flakes =
+            mapAttrsToList (n: v: { inherit (v) from to exact; }) cfg.registry;
+        };
+      };
+
+      "nix/nix.conf" = mkIf (cfg.settings != { }) { source = nixConf; };
```
berbiche, line 233:
```suggestion
      "nix/nix.conf".source = mkIf (cfg.settings != { }) nixConf;
```

berbiche (changes requested):
Thanks!

The only required change is changing `home-manager` for `Home Manager`, the other comments may be ignored.

tests/modules/misc/nix/example-settings-expected.conf
polykernel: I wasn't sure on the correct naming convention. Thanks!

polykernel:
Sorry for the late response, some life events got in the way. I have applied the required changes as requested.

berbiche (approved):
Thanks!

modules/misc/nix.nix
```diff
@@ -68,19 +158,81 @@ in {
...
-      flakes =
-        mapAttrsToList (n: v: { inherit (v) from to exact; }) cfg.registry;
+  config = mkIf cfg.enable {
+    assertions = [{
+      assertion = cfg.settings == { } || cfg.package != null;
+      message = ''
+        A corresponding Nix package must be specified via `nix.package` for generating
+        nix.conf.
+      '';
+    }];
```
ncfavier, line 222: This does not trigger when only `extraOptions` is set, causing https://github.com/nix-community/home-manager/issues/3644.

Also, is there a reason not to set `pkgs.nix` as the default package?

polykernel: The reason `pkgs.nix` is not set as the default is because the config generation checkPhase depends on the system nix version. The concern was that if the version of default nix package did not match the user's installed nix version, then the `show-config` nix command could potentially validate against an incompatible config schema. This is an issue when new releases introduce option changes that break backwards compatibility(e.g options being renamed/removed).

We can't assumed that home-manager has access to system state(i.e as a NixOS module) to determine the system nix version so the compromise was made to force the user to determine the version/schema to be used.

ncfavier: Fair enough, but we could at least try [`osConfig.nix.package`](https://github.com/nix-community/home-manager/blob/c43d4a3d6d9ef8ddbe2438362f5c775b4186000b/nixos/common.nix#L18)

ncfavier: https://github.com/nix-community/home-manager/pull/3647

PR #2720, opened 2022-02-13
modules/default.nix: Add useNixpkgsModule parameter

### LINKED ISSUE #2701: Unprivileged flake setup always imports nixpkgs

### Issue description

https://github.com/nix-community/home-manager/issues/616#issuecomment-1030646908

> @ShamrockLee While that works, you get an uncustomized nixpkgs. If you've already imported it, configured it, and added overlays elsewhere in your flake it's wasteful for home-manager to import it again and not very helpful.
> 
> I currently use this:
> 
> https://github.com/LunNova/nixos-configs/blob/1eda3c70d38a492b98a5fb1989f1c41761edffef/flake.nix#L204-L205

> You can override the pkgs argument actually
> 
> Anyway, @LunNova if you could move the discussion to a new issue it would be appreciated given the two problems are different.
> 
> The original issue is about using the global nixpkgs configuration when using Home Manager as module.
> 
> The original issue was resolved when the home-manager.useGlobalPkgs option was introduced.

### Maintainer CC

_No response_

### System information

```markdown
 - system: `"x86_64-linux"`
 - host os: `Linux 5.15.19-xanmod1-tt, NixOS, 22.05 (Quokka)`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.7.0pre20220127_558c4ee`
 - channels(lun): `""`
 - channels(root): `"nixos, nixpkgs"`
 - nixpkgs: `/etc/nix/path/nixpkgs`
```

LunNova:
### Description

Fixes #2701 by making it possible to disable the nixpkgs module with the unprivileged flake setup.

Sets `disableNixpkgsModule` to `true` by default when using `lib.homeManagerConfiguration`, to avoid reimporting `nixpkgs` and then copying the same exact config and overlays.

<details>
<summary> Click for some obsolete changes before #3037 went in </summary>
I also made this the default for `lib.homeManagerConfiguration` as I think it makes more sense for performance to use the `pkgs` input directly rather than reimporting it and copying config/overlays. I could change that back to being optional if desired, not sure.

The current requirement that `homeManagerConfiguration` take a `stateVersion`, `homeDirectory` and `username` arg feels weird to me as it prevents setting these in that user's configuration nix files (unless you mkForce). This is more often done this way and makes it easier to share the same configuration between use in nixos modules and use for the unprivileged flake setup.

I think it would be best if `homeManagerConfiguration` worked like this:

```
      homeManagerConfiguration = { configuration, system
          , extraModules ? [ ], extraSpecialArgs ? { }
          , pkgs ? builtins.getAttr system nixpkgs.outputs.legacyPackages
          , check ? true }@args:
          import ./modules {
            inherit pkgs check extraSpecialArgs useNixpkgsModule;
            configuration = { ... }: {
              imports = [ configuration ] ++ extraModules;
              // removed the home.* settings here and removed the args for them
              // they should be set in extraModules or configuration like all other hm config is
            };
          };
      };
```

But that wouldn't be backwards compatible.

</details>

Update for 2022-08: Most of that was done in https://github.com/nix-community/home-manager/pull/3037, this PR now only has the `useNixpkgsModule` changes.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -23,7 +23,7 @@
       lib = {
         hm = (import ./modules/lib/stdlib-extended.nix nixpkgs.lib).hm;
         homeManagerConfiguration = { modules ? [ ], pkgs, lib ? pkgs.lib
-          , extraSpecialArgs ? { }, check ? true
+          , extraSpecialArgs ? { }, check ? true, disableNixpkgsModule ? true
             # Deprecated:
           , configuration ? null, extraModules ? null, stateVersion ? null
           , username ? null, homeDirectory ? null, system ? null }@args:
@@ -50,11 +50,15 @@
               throwForRemovedArg extraModules # \
               throwForRemovedArg system;
           in throwForRemovedArgs (import ./modules {
-            inherit pkgs lib check extraSpecialArgs;
-            configuration = { ... }: {
-              imports = modules;
-              nixpkgs = { inherit (pkgs) config overlays; };
-            };
+            inherit pkgs lib check extraSpecialArgs disableNixpkgsModule;
+            configuration = { ... }:
+              ({
+                imports = modules;
+              } // (if disableNixpkgsModule then
+                { }
+              else {
+                nixpkgs = { inherit (pkgs) config overlays; };
+              }));
           });
       };
     } // utils.lib.eachDefaultSystem (system:
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -6,6 +6,8 @@
 , check ? true
   # Extra arguments passed to specialArgs.
 , extraSpecialArgs ? { }
+  # If disabled, the pkgs attribute passed to this function is used instead.
+, useNixpkgsModule ? true
 }:
 
 with lib;
@@ -25,7 +27,7 @@ let
 
   hmModules =
     import ./modules.nix {
-      inherit check pkgs;
+      inherit check pkgs useNixpkgsModule;
       lib = extendedLib;
     };
 
--- a/docs/nix-flakes.adoc
+++ b/docs/nix-flakes.adoc
@@ -69,6 +69,11 @@ writing a Home Manager configuration.
 
         # Optionally use extraSpecialArgs
         # to pass through arguments to home.nix
+
+        # Optionally set
+        # disableNixpkgsModule = false;
+        # to allow using nixpkgs.config/overlays inside the home configuration
+        # instead of directly using the passed pkgs
       };
     };
 }
--- a/docs/release-notes/rl-2211.adoc
+++ b/docs/release-notes/rl-2211.adoc
@@ -34,8 +34,8 @@ have been removed. Instead use the new `modules` argument, which
 accepts a list of NixOS modules.
 +
 Further, the `pkgs` argument is now mandatory and should be set to
-`nixpkgs.legacyPackages.${system}` where `nixpkgs` is the Nixpkgs
-input of your choice.
+an imported `nixpkgs`, for example `nixpkgs.legacyPackages.${system}`
+where `nixpkgs` is the Nixpkgs input of your choice.
 +
 For example, if your Flake currently contains
 +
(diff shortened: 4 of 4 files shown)


flake.nix
```diff
@@ -44,15 +44,15 @@
         homeManagerConfiguration = { configuration, system, homeDirectory
           , username, extraModules ? [ ], extraSpecialArgs ? { }
           , pkgs ? builtins.getAttr system nixpkgs.outputs.legacyPackages
-          , check ? true, stateVersion ? "20.09" }@args:
+          , check ? true, stateVersion ? "20.09", useNixpkgsModule ? false
```
berbiche, line 47:
```suggestion
          , check ? true, stateVersion ? "20.09", useNixpkgsModule ? true
```

Let's toggle this to false by default for the sake of backward compatibility.

berbiche (changes requested):
Would it be possible to update the flake document in the README and the manuals to mention this new option?

LunNova:
Reverted the flake.nix change, let's just add the extra flag to `modules/default.nix` since it's easy enough to write your own thing which imports it and I wasn't sure how to handle documentation/backcompat.

This should still be an improvement even if it's not as easy to use.

LunNova:
@rycee @berbiche is there anything I need to do to get this merged?

Most of what I wanted to do was done in #3037 (by just not being back compat which is cool), the only change left in this PR is adding a `useNixpkgsModule` option in `modules/default.nix` and it should be ready.

berbiche:
Hi, sorry, this PR is ready to be merged but it would be great if the documentation mentioned this new option.

LunNova:
Doesn't look like there is anywhere to document it with the current structure. If you use the CLI you don't import this yourself so it's not exposed, similarly if you use the flake.nix lib.homeManagerConfiguration it's not exposed yet.

I could add it as an option to that and document it, although I was trying to keep this minimal so I could get it merged :thinking:

LunNova:
@berbiche added the option to `lib.homeManagerConfiguration` so it's exposed and documented it. I defaulted it to `true`, happy to swap that to `false` if that's preferred.

LunNova:
1 failing test in CI, it's also failing on master.

modules/default.nix
```diff
@@ -6,6 +6,8 @@
 , check ? true
   # Extra arguments passed to specialArgs.
 , extraSpecialArgs ? { }
+  # If disabled, the pkgs attribute passed to this function is used instead.
+, useNixpkgsModule ? true
```
berbiche, line 10: I agree that setting this to `true` is the better choice, but I am sure there will be bug tickets related to this change.

berbiche (approved):
I'd like your input on this @rycee when you can :slightly_smiling_face:

berbiche:
> 1 failing test in CI, it's also failing on master.

This has been resolved on master

flake.nix
```diff
@@ -23,7 +23,7 @@
       lib = {
         hm = (import ./modules/lib/stdlib-extended.nix nixpkgs.lib).hm;
         homeManagerConfiguration = { modules ? [ ], pkgs, lib ? pkgs.lib
-          , extraSpecialArgs ? { }, check ? true
+          , extraSpecialArgs ? { }, check ? true, disableNixpkgsModule ? true
```
ncfavier, line 26: Why the negation? `useNixpkgsModule`. Also it should default to whatever its value is now, to preserve backwards compatibility.

LunNova:
I don't have the energy to update this right now. I did more than a year ago when I opened it but it's taken so long going through reviews that it doesn't feel worth it.

Hopefully someone else can make a PR for this.

nhooey:
I created this StackOverflow question to help everyone figure this out for the various different scenarios of using Nix:

* [StackOverflow: How to allow "unfree" packages in Nix, for each situation?: {NixOS, Nix, Nix with Flakes, Nix with Flakes and Home Manager}](https://stackoverflow.com/questions/77585228/how-to-allow-unfree-packages-in-nix-for-each-situation-nixos-nix-nix-wit)

Since GitHub issue comments are structured more like a forum of endless posts sorted by time instead of a proper question & answer site.

PR #2723, opened 2022-02-15
neovim: Support Lua for plugin configuration

cdunster:
### Description

This is to support the use of Lua as a language for plugin configuration in neovim.
Lua plugins are becoming the norm in neovim so it is useful to be able to configure them with Lua.

This should be expanded on later to add better support for Lua but this is a good first step.

P.S. I am very new to Nix and home-manager so any feedback is welcome, but please be kind :grin: 

### Checklist

    (_bash module tests not passing, I think this is an error on master though_)

    ```
    {component}: {description}

    {long description}
    ```

  (_not new but no current maintainer_)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -157,6 +157,9 @@
 /modules/programs/ne.nix                              @cwyc
 /tests/modules/programs/ne                            @cwyc
 
+/modules/programs/neovim.nix                          @cdunster
+/tests/modules/programs/neovim                        @cdunster
+
 /modules/programs/newsboat.nix                        @sumnerevans
 /tests/modules/programs/newsboat                      @sumnerevans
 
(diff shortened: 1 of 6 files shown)


cdunster:
> I prefer the other PR approach that just requires the generated lua config rather than the ugly heredoc sections:

No worries, I was actually planning on doing something similar as a next step but where each plugin configured with Lua had a separate file in the `.config/nvim/lua` subdirectory and then sourced in init.vim or init.lua as this is my preferred approach for my current dotfiles.

> I intend to merge https://github.com/nix-community/home-manager/pull/2716 as soon as my comment is adressed then we could look at generating an init.lua by default and converting these

Nice, just happy to have Lua supported so I can use my existing config.

> NB: even though lua plugins are nice, I suspect many less vocal) neovim users still use an init.vim (I do, well I have an hybrid setup), some operations are still easier in viml IMO. 

You're right but I think more and more people are moving to Lua and it's great for new people so adding support now makes sense.

PR #2729, opened 2022-02-18
pylint: add module

florpe:
### Description

Adds a module for the Python linter pylint.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/pylint.nix
+++ b/modules/programs/pylint.nix
@@ -1,0 +1,52 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.programs.pylint;
+
+  # generators.toINI fails when it encounters a list, so we have to set up
+  # some infrastructure.
+  mkVal = generators.mkValueStringDefault { };
+  mkINIVal = v:
+    if isList v then
+      builtins.concatStringsSep ", " (builtins.map mkVal v)
+    else
+      mkVal v;
+  mkINIKeyVal = generators.mkKeyValueDefault { mkValueString = mkINIVal; } "=";
+  mkINI = attrOfAttrs:
+    generators.toINI { mkKeyValue = mkINIKeyVal; } attrOfAttrs;
+  writeINI = name: attrsOfAttrs: pkgs.writeText name (mkINI attrsOfAttrs);
+
+in {
+  meta.maintainers = [ hm.maintainers.florpe ];
+  options.programs.pylint = {
+    enable = mkEnableOption "the pylint Python linter";
+    package = mkOption {
+      type = types.package;
+      default = pkgs.python3Packages.pylint;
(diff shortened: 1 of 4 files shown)


berbiche:
Hi,

This module allows you to configure a global pylintrc configuration file, but can you override this file on a per-project basis without unsetting `PYLINTRC`?

According to this StackOverflow [answer](https://stackoverflow.com/a/32672068), it's possible to set the file `~/.pylintrc` with the user configuration.
IMO this would be a better solution than setting an environment variable as the environment variable will only work correctly if the user has set `programs.$SHELL.enable` for each shell.

florpe:
According to the [pylint docs](https://pylint.pycqa.org/en/latest/user_guide/run.html#command-line-options), setting `PYLINTRC` or linking from `~/.pylintrc` will have the same effect - in either case `$PWD/.pylintrc` will take precedence. I see your point about shell integration though, and looking at other modules writing a dotfile seems like the more popular approach anyway, so I'll adjust the PR.

Is there a style guide for decisions like this?

berbiche:
Could you rebase your changes on the latest "master" to fix the failing test case?

> and looking at other modules writing a dotfile seems like the more popular approach anyway, so I'll adjust the PR.
> 
> Is there a style guide for decisions like this?


AFAIK we don't currently have any documentation regarding when to use dotfiles or environment variables.
Dotfiles are generally a better idea because they work without needing to source Home Manager's environment variables.
Sometimes we use environment variables by wrapping tools with a script that sets the right environment before starting.

berbiche (changes requested):

modules/programs/pylint.nix
```diff
@@ -0,0 +1,31 @@
...
+
+let
+
+  cfg = config.programs.pylint;
+  listToValue = concatMapStringsSep ",\n    " (generators.mkValueStringDefault { });
+  iniFormat = pkgs.formats.ini { inherit listToValue; };
+in {
+  meta.maintainers = [ hm.maintainers.florpe ];
+  options.programs.pylint = {
+    enable = mkEnableOption "The pylint Python linter.";
```
berbiche, line 13:
```suggestion
    enable = mkEnableOption "the pylint Python linter";
```

`mkEnableOption` automatically appends a final period.

berbiche (changes requested):
Small nit, lgtm besides that.

berbiche (changes requested):
> ./modules/programs/pylint.nix: not formatted

Unfortunately the formatting script has to be run again.

florpe:
It appears nixfmt has some difficulties with the newline in the list separator - now it's just `", "`, which is going to be ugly for longer lists, but at least the formatter is happy.

berbiche (approved):
LGTM!

PR #2731, opened 2022-02-18
lorri: add nixPackage option; add notifier

nyarly:
### Description

Adds `nixPackage` option, which allows lorri to use `nixUnstable`,
and therefore flake based shell.s

Adds a little notifier service, so that starts and ends of builds can be reported.

### Checklist

- [x] Code tested through `nix-shell --pure tests -A run.all`. _bash.nix test fails_

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/lorri.nix
+++ b/modules/services/lorri.nix
@@ -12,11 +12,21 @@ in {
   options.services.lorri = {
     enable = mkEnableOption "lorri build daemon";
 
+    enableNotifications = mkEnableOption "lorri build notifications";
+
     package = mkOption {
       type = types.package;
       default = pkgs.lorri;
       defaultText = literalExpression "pkgs.lorri";
       description = "Which lorri package to install.";
+    };
+
+    nixPackage = mkOption {
+      type = types.package;
+      default = pkgs.nix;
+      defaultText = literalExpression "pkgs.nix";
+      example = literalExpression "pkgs.nixVersions.unstable";
+      description = "Which nix package to use.";
     };
   };
 
@@ -45,7 +55,7 @@ in {
           Restart = "on-failure";
           Environment = let
             path = with pkgs;
-              makeSearchPath "bin" [ nix gitMinimal gnutar gzip ];
+              makeSearchPath "bin" [ cfg.nixPackage gitMinimal gnutar gzip ];
           in [ "PATH=${path}" ];
         };
       };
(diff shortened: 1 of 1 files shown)


modules/services/lorri.nix
```diff
@@ -60,6 +90,23 @@ in {
 
         Install = { WantedBy = [ "sockets.target" ]; };
       };
+
+      services.lorri-notify = mkIf (cfg.notify) {
```
berbiche, line 94:
```suggestion
      systemd.user.services.lorri-notify = mkIf (pkgs.stdenv.hostPlatform.isLinux && cfg.notify) {
```

Until we have a common backend to declare services for Darwin and Linux, the service will have to be gated.

modules/services/lorri.nix
```diff
@@ -6,18 +6,48 @@ let
...
       description = "Which lorri package to install.";
     };
+
+    nixPackage = mkOption {
+      type = types.package;
+      default = pkgs.nix;
+      defaultText = literalExpression "pkgs.nix";
+      description = "Which nix package to use.";
+      example = literalExpression "pkgs.nixUnstable";
+    };
```
berbiche, line 50: TODO: change the default value to `config.nix.package` once we add this option.

berbiche (changes requested):
Thanks!

I didn't test the notify script but I found your comment on the lorri issue tracker (https://github.com/nix-community/lorri/issues/77#issuecomment-1035564228) so I assume it works correctly!

modules/services/lorri.nix
nyarly: Not sure how to interpret this - just a note to the project, or does it need to be addressed?

modules/services/lorri.nix
```diff
@@ -60,6 +90,30 @@ in {
 
         Install = { WantedBy = [ "sockets.target" ]; };
       };
+
+      services.lorri-notify =
+        mkIf (pkgs.stdenv.hostPlatform.isLinux && cfg.notify) {
```
rycee, line 95: No need to check for Linux platform
```suggestion
        mkIf cfg.notify {
```

modules/services/lorri.nix
```diff
@@ -6,18 +6,48 @@ let
...
+    name = "lorri_notify";
+    executable = true;
+    text = ''
+      #! /usr/bin/env bash
+
+      lorri internal stream-events --kind live |\
+      jq --unbuffered '${jqFile}' |\
+      xargs -n 1 notify-send "Lorri Build"
+    '';
+  };
```
rycee, line 27: Can simplify using `writeShellScript`:
```suggestion
  notifyScript = pkgs.writeShellScript "lorri-notify" ''
    lorri internal stream-events --kind live \
      | jq --unbuffered '${jqFile}' \
      | xargs -n 1 notify-send "Lorri Build"
  '';
```

rycee:
Thanks! Looks good in general, I just added a few comments.

Additionally, please squash/split the commits so that there are two independent commits with the two added features. Something like

- lorri: add nixPackage option
- lorri: add support for notifying on build events

modules/services/lorri.nix
nyarly: I guess the thing here is that I'm not sure how useful the script is outside the specific context of a per-user systemd service. There's some discussion in the lorri issues about whether it makes sense to standardize this as a lorri subcommand - I'm not sure there's a universal version that makes sense.

rycee:
Thanks! I squashed and merged to master now. Note, I decided to rename the `notify` option to `enableNotifications` to make it more clear.

modules/services/lorri.nix
rycee: The `nix.package` option is available now but I left changing the default as an exercise for some intrepid soul 😉

PR #2734, opened 2022-02-19
just: add module for the just command runner

maximsmol:
### Description

Adds a module for [just](https://github.com/casey/just) and its shell completions.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -119,6 +119,8 @@
 
 /modules/programs/java.nix                            @ShamrockLee
 
+/modules/programs/just.nix                            @maximsmol
+
 /modules/programs/keychain.nix                        @marsam
 
 /modules/programs/kodi.nix                            @dwagenk
--- a/modules/programs/just.nix
+++ b/modules/programs/just.nix
@@ -1,0 +1,52 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.programs.just;
+
+in {
+  meta.maintainers = [ hm.maintainers.maximsmol ];
+
+  options.programs.just = {
+    enable = mkEnableOption
+      "just, a handy way to save and run project-specific commands";
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.just;
+      defaultText = literalExpression "pkgs.just";
+      description = "Package providing the <command>just</command> tool.";
+    };
+
+    enableBashIntegration = mkEnableOption "Bash integration" // {
+      default = true;
+    };
+
+    enableZshIntegration = mkEnableOption "Zsh integration" // {
+      default = true;
+    };
+
+    enableFishIntegration = mkEnableOption "Fish integration" // {
+      default = true;
+    };
+  };
+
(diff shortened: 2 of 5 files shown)


.github/CODEOWNERS
```diff
@@ -437,3 +437,5 @@ Makefile                                              @thiagokokada
 /modules/programs/ion.nix                             @jo1gi
 
 /modules/services/plex-mpv-shim.nix                   @starcraft66
+
+/modules/programs/just.nix                            @maximsmol
```
berbiche, line 441: Could this be moved at the right place to preserve the ordering? Thanks.

modules/programs/just.nix
```diff
@@ -0,0 +1,64 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.programs.just;
+
+in {
+  meta.maintainers = [ maintainers.maximsmol ];
```
berbiche, line 10:
```suggestion
  meta.maintainers = [ hm.maintainers.maximsmol ];
```

Home Manager's maintainers are available `lib.hm.maintainers`.

modules/programs/just.nix
```diff
@@ -0,0 +1,64 @@
...
+
+  options.programs.just = {
+    enable = mkEnableOption
+      "just is a handy way to save and run project-specific commands";
```
berbiche, line 14: This expands to `Whether to enable just is a handy way...`.
The phrasing seems odd.

berbiche (changes requested):
Thanks for the contribution!

modules/programs/just.nix
```diff
@@ -0,0 +1,65 @@
...
+
+  cfg = config.programs.just;
+
+in
+{
+  meta.maintainers = [ hm.maintainers.maximsmol ];
+
+  options.programs.just = {
+    enable = mkEnableOption
+      "just - a handy way to save and run project-specific commands";
```
sumnerevans, line 15:
```suggestion
      "just, a handy way to save and run project-specific commands";
```

modules/programs/just.nix
```diff
@@ -0,0 +1,65 @@
...
+      description = "Package providing the <command>just</command> tool.";
+    };
+
+    enableBashIntegration = mkOption {
+      default = true;
+      type = types.bool;
+      description = ''
+        Whether to enable Bash integration.
+      '';
+    };
```
sumnerevans, line 30: All of these enable options can use `mkEnableOption` like so:

```suggestion
    enableBashIntegration = mkEnableOption "Bash integration" // {
      default = true;
    };
```

maximsmol:
@sumnerevans just fyi, copied both the description style ("{name} - {description}") and the `mkOption` from `programs.fzf`—I could reformat those in another PR according to your suggestions as well if you think it'd be good

berbiche (approved):
LGTM!

berbiche:
Your PR needs to be rebased on "master" to include a fix that will allow the checks to succeed.

rycee:
Merged to master now. Thanks!

Issue #2722, opened 2022-02-15
bug: attribute 'shellDryRun' missing in 'modules/programs/bash.nix'


FabianGD:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

With the current update, the bash module in home-manager started breaking for me. I did not change anything in the program.bash set, but started getting the following error:

```
error: attribute 'shellDryRun' missing, at /home/fabian/.nix-defexpr/channels/home-manager/modules/programs/bash.nix:12:9
```

The issue persists after updating my user nixpkgs channel to the version stated below. 
This seems to be related to #2693 and #2668.

### Maintainer CC

@rycee

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 5.10.90, NixOS, 21.11 (Porcupine)`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.3.16`
 - channels(root): `"nixos-21.11.335130.386234e2a61"`
 - channels(fabian): `"home-manager, nixos-21.11.335946.c28fb0a4671, nixpkgs-unstable-22.05pre352484.1882c6b7368"`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`
```

maximsmol:
Running `sudo nix-channel --update` in addition to `nix-channel --update` fixed this issue for me on my Mac

Hugo-Heagren:
Can report that this is also happening to me, and `sudo nix-channel --update && nix-channel --update` hasn't helped.

MatrixManAtYrService:
I'm a new user, so I don't really know which way is up yet, but I am also running into this.  I just added

`<home-manager/nixos>` to my `configuration.nix` and then ran `nixos-rebuild switch`

Also, the update commands shown above have not helped.

MatrixManAtYrService:
Perhaps this is a no-brainer to those more familiar with home-manager, but removing this line from `/etc/nixos/configuration.nix` made the error go away:
```
  home-manager.users.matt = { pkgs, ...}: {
      programs.bash.enable = true;  # <--- remove this line
  };
```

I'm still reasoning through the consequences of that action, but I thought I'd mention it here in case it leads to useful workarounds for somebody.

montchr:
[...]

Issue #2540, opened 2021-12-09
bug: tmux bindings don't have notes, so they don't show up in `C-b ?`


lilyball:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

tmux 3.1 added the ability to attach notes to key bindings, with notes for all the default bindings. It also changed `C-b ?` to only list bindings with notes (this is `tmux list-keys -N`). It's still possible to list keys without notes using `tmux list-keys -Na`, but there's no binding for this.

The problem is home-manager includes its own key bindings, but these don't have notes, meaning they won't show up in `C-b ?`. Personally I think it was a mistake for tmux to suppress keys without notes by default, but perhaps this was done to try and force everyone to use notes?

In any case, home-manager's tmux module needs to attach notes to its keys. Otherwise they're not sufficiently discoverable for users. It should copy the current notes for any keys whose action has a default key binding, and use something sensible for any other keys.

This change should be safe to do on release branches from `release-20.09` onwards, as `nixos-20.09` contains tmux 3.1c (`nixos-20.03` has tmux 3.0a so that's too old for this).

### Maintainer CC

@marsam @toonn @expipiplus1

### System information

```markdown
- system: `"x86_64-darwin"`
 - host os: `Darwin 21.1.0, macOS 10.16`
 - multi-user?: `yes`
 - sandbox: `relaxed`
 - version: `nix-env (Nix) 2.4`
 - channels(lily): `"darwin, home-manager"`
 - channels(root): `"nixpkgs-21.11pre329221.c11d08f0239"`
 - nixpkgs: `/etc/nix/inputs/nixpkgs`
```

lilyball:
Argh, stale bots are the bane of my existence. Everything it says is a lie. People treat closed issues as _closed_, not as potential issues that were just shelved.

> Closed issues stay in the system for people to search, read, cross-reference, or even reopen

This especially is wrong. I'm the author of this issue and even I can't reopen it!

Stale bots only serve to make users feel like their issues don't matter and discourage future issues. This is actually a pretty significant issue with the `tmux` module. My `list-keys` binding now hides most of my bound keys.

lilyball:
Hell, this issue is still even labeled as `triage`. Why the hell is the stale bot closing issues that were _never even triaged_?!?

berbiche:
Sorry about the situation.

toonn:
@lilyball, #2742 does address this, maybe that at least reduces the frustration.

[...]

PR #2742, opened 2022-02-20
tmux: add notes to existing keybindings (#2540)

LogicalOverflow:
### Description

This PR adds notes to the keybindings created in the tmux config, making them show up when checking `C-b ?` or `tmux list-keys -N`, fixing #2540. I also adjusted the reference configs in the test cases to expect the notes.

Keys without notes are not displayed when checking `C-b ?` (since tmux 3.1), making them hard to discover for users.

The explanations are mostly taken from/based on the tmux man-page (which is licensed under Zero-Clause BSD).

I ran all tests which don't complain about the `shellDryRun` attribute missing (see #2722) and all tests that pass without the commit also pass with it (only `herbstluftwm-simple-config` fails, both with and without my commit).

### Checklist

- [ ] Code tested through `nix-shell --pure tests -A run.all`. (Mostly of done, see above)

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/tmux.nix
+++ b/modules/programs/tmux.nix
@@ -49,8 +49,8 @@ let
     ${optionalString cfg.newSession "new-session"}
 
     ${optionalString cfg.reverseSplit ''
-      bind v split-window -h
-      bind s split-window -v
+      bind -N "Split the pane into two, left and right" v split-window -h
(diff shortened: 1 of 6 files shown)


toonn (approved):
LGTM

berbiche (approved):
Thanks!

PR #1145, opened 2020-04-13
files: add an option to copy files instead of linking

rtimush:
### Description

This implements #155 by adding an option to copy files to the home instead of symlinking them:
```
home.file.foo = { 
  text = "content";
  copy = true;
};
```

Although the code works for me, it needs more testing and I have some questions regarding the approach taken, so I'm creating this pull request as a draft.

The idea is that we collect a list of files that we want to copy to the home directory in the `.home-manager-copy` file inside `home-files`. The activation script looks into this file to decide whether it needs to run `ln` or `cp`.  It also now compares file content with the previous generation and proceeds with an update or deletion only if the file hasn't been modified.

Open questions:

1. Is `copy` a good option name? Another one I can think of is `link = false`.
2. Is the `.home-manager-copy` file approach ok or it's better to keep linked and copied files separately?
3. Is there a way to write tests for the activation script?

### Checklist

- [x] Code formatted with `./format`.

- [x] Code tested through `nix-shell --pure tests -A run.all`.

- [ ] Test cases updated/added. See [example](https://github.com/rycee/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).

- [x] Commit messages are formatted like

    ```
    {component}: {description}

    {long description}
    ```

    See [CONTRIBUTING](https://github.com/rycee/home-manager/blob/master/CONTRIBUTING.md#commits) for more information and [recent commit messages](https://github.com/rycee/home-manager/commits/master) for examples.

- If this PR adds a new module

  - [ ] Added myself as module maintainer. See [example](https://github.com/rycee/home-manager/blob/068ff76a10e95820f886ac46957edcff4e44621d/modules/programs/lesspipe.nix#L6).

  - [ ] Added myself and the module files to `.github/CODEOWNERS`.

PR #2743, opened 2022-02-21
vscode: add mutable user settings option.

gcoakes:
### Description

User settings could not be modified once home-manager took over managing
settings. This setting allows user settings to diverge from what is
configured in the user's home-manager config. New settings from the home-manager config are still
applied but anything removed from the home-manager config will not be removed from the mutable settings. Similarly, any changes the user makes to the mutable settings will take precedence over the home-manager config settings.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/vscode.nix
+++ b/modules/programs/vscode.nix
@@ -135,6 +135,16 @@ in {
           or by Visual Studio Code.
         '';
       };
+
+      mutableUserSettings = mkOption {
+        type = types.bool;
+        default = false;
+        example = false;
+        description = ''
+          Whether user settings can be changed manually or by Visual
+          Studio Code.
+        '';
+      };
     };
   };
 
@@ -142,7 +152,7 @@ in {
     home.packages = [ cfg.package ];
 
     home.file = mkMerge [
-      (mkIf (cfg.userSettings != { }) {
+      (mkIf (!cfg.mutableUserSettings && cfg.userSettings != { }) {
         "${configFilePath}".source =
           jsonFormat.generate "vscode-user-settings" cfg.userSettings;
       })
@@ -172,5 +182,23 @@ in {
         "${extensionPath}".source = extensionsFolder;
       }))
     ];
+
+    home.activation =
+      mkIf (cfg.mutableUserSettings && cfg.userSettings != { }) {
+        injectVscodeSettings = lib.hm.dag.entryAfter [ "writeBoundary" ] ''
+          tmp="$(mktemp)"
+          if [[ -v DRY_RUN ]]; then
+            echo tmp="\$(mktemp)"
+            echo ${pkgs.jq}/bin/jq -s "'reduce .[] as \$x ({}; . * \$x)'" "${
+              jsonFormat.generate "vscode-user-settings" cfg.userSettings
+            }" "${configFilePath}" ">" "$tmp"
+          else
+            ${pkgs.jq}/bin/jq -s 'reduce .[] as $x ({}; . * $x)' "${
+              jsonFormat.generate "vscode-user-settings" cfg.userSettings
+            }" "${configFilePath}" > "$tmp"
+          fi
+          $DRY_RUN_CMD mv "$tmp" "${configFilePath}"
+        '';
+      };
   };
 }


gcoakes:
The test cases were failing for what appears to be an unrelated change. It looks like it is being seeing elsewhere  #2722.

sumnerevans:
The broken test has been fixed. Please reabse.

gcoakes:
@sumnerevans, I just rebased. For some reason the tests when run locally are hanging while copying glibc from the cache, but it appears that CI has me covered.

berbiche:
Hi,

How does this handle merging values that were added but then removed from Home Manager?

If I build a configuration with settings A, remove the setting and rebuild, will setting A still be there?

berbiche:
> New settings from the home-manager config are still
applied but anything removed from the home-manager config will not be removed from the mutable settings

I think in the activation script you have access to the previous generation so you could preserve the current Home Manager settings in a file (`home.file."something"...`) and when merging settings you could remove any value that is missing from this file and that was not changed by the user.

gcoakes:
@berbiche, can we just merge as is then if someone wants the diffing feature that can be a follow up patch? I just scoped out the changes a bit, and I'm not particularly interested in doing them. `jq` doesn't trivially provide a way to take the difference of arbitrarily structured objects. The current implementation is sufficient for my use.

If you don't want to merge, then I'll probably just extract the module and put it under a different config namespace in my local modules.

modules/programs/vscode.nix
```diff
@@ -135,14 +135,24 @@ in {
           or by Visual Studio Code.
         '';
       };
+
+      mutableUserSettings = mkOption {
+        type = types.bool;
+        default = false;
+        example = false;
```
turion, line 142:
```suggestion
        example = true;
```

modules/programs/vscode.nix
```diff
@@ -135,14 +135,24 @@ in {
...
+        description = ''
+          Whether user settings can be changed manually or by Visual
```
turion, line 144:
```suggestion
          Whether the user settings file `settings.json` can be changed by Visual
```

turion:
Sounds like a great idea to me. Possibly related upstream issue: https://github.com/microsoft/vscode/issues/147911

kristoff3r:
This would be nice to get merged

modules/programs/vscode.nix
```diff
@@ -172,5 +182,23 @@ in {
...
+              jsonFormat.generate "vscode-user-settings" cfg.userSettings
+            }" "${configFilePath}" ">" "$tmp"
+          else
+            ${pkgs.jq}/bin/jq -s 'reduce .[] as $x ({}; . * $x)' "${
+              jsonFormat.generate "vscode-user-settings" cfg.userSettings
+            }" "${configFilePath}" > "$tmp"
+          fi
+          $DRY_RUN_CMD mv "$tmp" "${configFilePath}"
+        '';
+      };
```
sumnerevans, line 202: This seems overly complicated. I think that it's fine if the dry-run creates temporary files. Just make sure that it doesn't actually override the `configFilePath`.

Regardless, this code needs a comment or two explaining what it's doing.

---

I guess this situation is also a vote in favor of having https://github.com/nix-community/home-manager/pull/1145 merged...

modules/programs/vscode.nix
```diff
@@ -135,14 +135,24 @@ in {
...
+
+      mutableUserSettings = mkOption {
+        type = types.bool;
+        default = false;
+        example = false;
+        description = ''
+          Whether user settings can be changed manually or by Visual
+          Studio Code.
+        '';
+      };
```
sumnerevans, line 147: I think the best option is to just use `mkEnableOption`

```suggestion
      mutableUserSettings = mkEnableOption "changing user settings in Visual Studio Code (rather than through Home Manager)";
      };
```

sumnerevans (changes requested):

kristoff3r:
I don't know if @gcoakes is still interested in finishing this up, otherwise I can take over and clean it up

sumnerevans:
@kristoff3r if you want to do that, please just open another PR and mention that it would close and supersede this PR.

poelzi:
I think I will take this one over soon

SaladinAyyub:
Just wanting to bump this as VS Code tries to write to settings.json whenever i open it.. it opens the file itself and shows this error


Screenshot From 2025-01-13 21-31-48

VS code currently imo has no option to disable this (disable vscode to write to settings.json). So just to get of the error.. this PR will be helpful to make vscode happy.

Issue #2724, opened 2022-02-15
mu should include aliases when initializing the database


exzombie:
Looking at how the [list of addresses is constructed](https://github.com/nix-community/home-manager/blob/master/modules/programs/mu.nix#L20), it appears that aliases are not taken into account. Without this, `mu` is unable to recognize all personal addresses.

ShamrockLee:
@KarlJoad

KarlJoad:
Correct. Aliases were not taken into account when this module was created, only the primary email address of each email account.

sumnerevans:
closed by https://github.com/nix-community/home-manager/pull/2748

PR #2748, opened 2022-02-22
mu: allow aliases to be used by mu configuration file

KarlJoad:
This has no effect if the user does not have any aliases defined for
any accounts.
This will also only add `--my-address=` to only accounts that are
enabled to be tracked by mu.

### Description

home-manager allows for users to define a list of aliases for a main email address.
Previously, this `mu` module did not understand these aliases.
Now `mu` will understand aliases and allow them to be treated "my-address", which makes them special in the eyes of mu.
This allows for filtering based on threads containing your email addresses and/or their aliases.

#2724

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/mu.nix
+++ b/modules/programs/mu.nix
@@ -12,12 +12,14 @@ let
   # Takes the list of accounts with mu.enable = true, and generates a
   # command-line flag for initializing the mu database.
   myAddresses = let
-    # List of account sets where mu.enable = true.
+    # Set of email account sets where mu.enable = true.
     muAccounts =
       filter (a: a.mu.enable) (attrValues config.accounts.email.accounts);
     addrs = map (a: a.address) muAccounts;
-    # Prefix --my-address= to each account's address with mu.enable.
-    addMyAddress = map (addr: "--my-address=" + addr) addrs;
+    # Construct list of lists containing email aliases, and flatten
+    aliases = flatten (map (a: a.aliases) muAccounts);
+    # Prefix --my-address= to each account's address AND all defined aliases
+    addMyAddress = map (addr: "--my-address=" + addr) (addrs ++ aliases);
   in concatStringsSep " " addMyAddress;
 
 in {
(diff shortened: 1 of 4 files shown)


modules/programs/mu.nix
```diff
@@ -12,12 +12,14 @@ let
...
+    # Set of email account sets where mu.enable = true.
     muAccounts =
       filter (a: a.mu.enable) (attrValues config.accounts.email.accounts);
     addrs = map (a: a.address) muAccounts;
-    # Prefix --my-address= to each account's address with mu.enable.
-    addMyAddress = map (addr: "--my-address=" + addr) addrs;
+    # Construct list of lists containing email aliases, and flatten
+    aliases = flatten (map (a: a.aliases) muAccounts);
+    # Prefix --my-address= to each account's address AND all defined aliases
+    addMyAddress = map (addr: "--my-address=" + addr) (addrs ++ aliases);
```
sumnerevans, line 22: This is getting complicated enough that a test feels useful. Not sure quite how to instrument one effectively, though. Maybe @rycee or @berbiche have ideas about how to do that (or maybe they disagree and we don't need a test).

KarlJoad: I agree that it is getting slightly more complicated. However, mu does not use a configuration file, only a set of command-line arguments passed during the initial index. I am not sure if home-manager could support that particular use. The command _is_ written out to the activation script, so that could be checked...

berbiche: A test would be a nice to have IMO.

KarlJoad: Agreed it would be nice. I am just not sure how to test it, because `mu` does not write a config file.

sumnerevans: I assume that `mu init` sets edits some files somewhere. Can we check the contents of those files?

KarlJoad: The only file that `mu init` creates is an empty database file with records containing the information you provide through the command line. Once you run `mu init`, you run `mu index` to populate that database with the relevant information.

The only file that is edited by this module is the activation script home-manager uses, by inserting an action to create the database.

berbiche (approved):
LGTM!

rycee:
Thanks! I added a basic test and merged to master.

PR #2753, opened 2022-02-23
Add lib argument to homeManagerConfiguration

dwarfmaster:
### Description

Add a `lib` argument to `homeManagerConfiguration` to specify the lib to use instead of the one from `pkgs`.

### Checklist

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -45,11 +45,11 @@
         homeManagerConfiguration = { configuration, system, homeDirectory
           , username, extraModules ? [ ], extraSpecialArgs ? { }
           , pkgs ? builtins.getAttr system nixpkgs.outputs.legacyPackages
-          , check ? true, stateVersion ? "20.09" }@args:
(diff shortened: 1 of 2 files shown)


ncfavier (approved):
LGTM. Out of curiosity, what is your use case for this?

dwarfmaster:
When I was refactoring my [configuration](https://github.com/dwarfmaster/home-nix), I was extending lib with some locally defined functions, but when I extended the lib in `pkgs` I had some infinite recursion, so I had a local `pkgs` with overlays and a local `lib` extended with some functions.

I have since noticed that I could use a different `pkgs` for home-manager, one extended with the complete lib, some I don't need this merged anymore, but since `nixosSystem` also has an optional lib argument it would make sense for `homeManagaerConfiguration` to also have one.

ncfavier:
OK, you might be interested in the discussion at https://github.com/NixOS/nixpkgs/pull/157056

berbiche (approved):
Thanks!
Merging after the tests.

PR #2757, opened 2022-02-24
xdg-user-dirs: allow paths and define sessionVariables

### LINKED ISSUE #2694: More consistent XDG support
### Description

The XDG module should set the environment variables for user directories if the option `xdg.userDirs.enable` is set to `true`, this would make it more consistent because the XDG module already sets some user specific variables like `XDG_CONFIG_HOME`:

https://github.com/nix-community/home-manager/blob/9d369c75ce2fdeb296ad42bcdc8c1a523c494550/modules/misc/xdg.nix#L89-L93

Currently the user directories aren't really usable in the configuration itself as they do not expand to the full path as theyr type is `types.str` :

https://github.com/nix-community/home-manager/blob/a52aed72c84a2a10102a92397339fa01fc0fe9cf/modules/misc/xdg-user-dirs.nix#L41-L45

This can be easily fixed by changing theyr type to `types.path` (this shouldn't cause any issues as they are defined per user anyway), which would again align with other XDG variables like `XDG_CONFIG_HOME`:

https://github.com/nix-community/home-manager/blob/a52aed72c84a2a10102a92397339fa01fc0fe9cf/modules/misc/xdg.nix#L45-L52

There could be a option to not set the environment variables for user directories, but as I see it would be pointless because once you want to manage them in your config you probably want them to be accessible system wide

ilkecan:
Changed the type of the user directory options from `str` to ~`either str path`~(`coercedTo path toString str`) to allow using path values.

The related session variable is defined for the default and the extra user directories now.

Closes #2694
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/xdg-user-dirs.nix
+++ b/modules/misc/xdg-user-dirs.nix
@@ -33,57 +33,62 @@ in {
     # https://gitlab.freedesktop.org/xdg/xdg-user-dirs/blob/master/man/user-dirs.dirs.xml
 
     desktop = mkOption {
-      type = types.str;
+      type = with types; coercedTo path toString str;
       default = "$HOME/Desktop";
       description = "The Desktop directory.";
     };
 
     documents = mkOption {
-      type = types.str;
+      type = with types; coercedTo path toString str;
       default = "$HOME/Documents";
       description = "The Documents directory.";
     };
 
     download = mkOption {
-      type = types.str;
+      type = with types; coercedTo path toString str;
       default = "$HOME/Downloads";
       description = "The Downloads directory.";
     };
 
     music = mkOption {
-      type = types.str;
+      type = with types; coercedTo path toString str;
       default = "$HOME/Music";
       description = "The Music directory.";
     };
 
     pictures = mkOption {
-      type = types.str;
+      type = with types; coercedTo path toString str;
       default = "$HOME/Pictures";
       description = "The Pictures directory.";
     };
 
     publicShare = mkOption {
-      type = types.str;
+      type = with types; coercedTo path toString str;
       default = "$HOME/Public";
       description = "The Public share directory.";
     };
 
     templates = mkOption {
-      type = types.str;
+      type = with types; coercedTo path toString str;
       default = "$HOME/Templates";
       description = "The Templates directory.";
     };
 
     videos = mkOption {
-      type = types.str;
+      type = with types; coercedTo path toString str;
       default = "$HOME/Videos";
       description = "The Videos directory.";
     };
 
     extraConfig = mkOption {
-      type = with types; attrsOf str;
+      type = with types; attrsOf (coercedTo path toString str);
       default = { };
-      example = { XDG_MISC_DIR = "$HOME/Misc"; };
+      defaultText = literalExpression "{ }";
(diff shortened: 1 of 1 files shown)


berbiche (changes requested):
AFAIK, the reason paths are not allowed is because it would lead to the entire path being imported in the nix store or prevent building on Flake configurations (Flake's can only refer to paths available in the git index).

ilkecan:
The paths being copied to the Nix store was not a problem because currently there isn't any antiquotation happening with those values. But ideally, this should be forward compatible.

The second point was indeed a problem.

I added `apply = toString` to the option definitions to avoid these problems.

modules/misc/xdg-user-dirs.nix
```diff
@@ -33,55 +33,64 @@ in {
...
       default = "$HOME/Videos";
       description = "The Videos directory.";
     };
 
     extraConfig = mkOption {
-      type = with types; attrsOf str;
+      type = with types; attrsOf (either str path);
+      apply = builtins.mapAttrs (_: toString);
       default = { };
       example = { XDG_MISC_DIR = "$HOME/Misc"; };
```
sumnerevans, line 95: let's fix this while you are at it already

```suggestion
      example = literalExpression ''
        {
          XDG_MISC_DIR = "$HOME/Misc";
        }
      '';
```

modules/misc/xdg-user-dirs.nix
```diff
@@ -33,55 +33,64 @@ in {
...
+      apply = toString;
       default = "$HOME/Videos";
       description = "The Videos directory.";
     };
 
     extraConfig = mkOption {
-      type = with types; attrsOf str;
+      type = with types; attrsOf (either str path);
+      apply = builtins.mapAttrs (_: toString);
       default = { };
```
sumnerevans, line 94:
```suggestion
      default = { };
      defaultText = literalExpression "{ }";
```

sumnerevans (changes requested):

modules/misc/xdg-user-dirs.nix
ilkecan: I am not sure I understand the reason for this change. Is it because of to prevent unnecessary whitespace at the end of the line?

modules/misc/xdg-user-dirs.nix
```diff
@@ -33,57 +33,71 @@ in {
...
+      type = with types; either str path;
+      apply = toString;
       default = "$HOME/Videos";
       description = "The Videos directory.";
     };
 
     extraConfig = mkOption {
-      type = with types; attrsOf str;
+      type = with types; attrsOf (either str path);
+      apply = builtins.mapAttrs (_: toString);
```
polykernel, line 93:
```suggestion
      type = with types; attrsOf (coercedTo path toString str);
```
minor nitpick, I think `coercedTo` is better suited here as only type coercion is happening, rather than transformation on the option value.

ilkecan: Thank you. I wasn't aware of the `coercedTo` type.

berbiche (approved):
Thanks again!

PR #2764, opened 2022-02-27
treewide: apply nixfmt to a few more files

rycee:
### Description

Removes a few of the remaining exceptions.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/mpd.nix
+++ b/modules/services/mpd.nix
@@ -18,9 +18,9 @@ let
     sticker_file        "${cfg.dataDir}/sticker.sql"
 
     ${optionalString (cfg.network.listenAddress != "any")
-      ''bind_to_address "${cfg.network.listenAddress}"''}
+    ''bind_to_address "${cfg.network.listenAddress}"''}
     ${optionalString (cfg.network.port != 6600)
-      ''port "${toString cfg.network.port}"''}
+    ''port "${toString cfg.network.port}"''}
 
     ${cfg.extraConfig}
   '';
@@ -54,7 +54,7 @@ in {
         type = with types; either path str;
         default = "${config.home.homeDirectory}/music";
         defaultText = "$HOME/music";
-        apply = toString;       # Prevent copies to Nix store.
+        apply = toString; # Prevent copies to Nix store.
         description = ''
           The directory where mpd reads music from.
         '';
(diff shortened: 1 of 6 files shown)


tests/modules/programs/gpg/override-defaults.nix
```diff
@@ -11,18 +11,17 @@ with lib;
...
 
       homedir = "${config.home.homeDirectory}/bar/foopg";
     };
 
     nmt.script = ''
       assertFileExists home-files/bar/foopg/gpg.conf
-      assertFileContent home-files/bar/foopg/gpg.conf ${./override-defaults-expected.conf}
+      assertFileContent home-files/bar/foopg/gpg.conf ${
+        ./override-defaults-expected.conf
+      }
```
SuperSandro2000, line 24: Prime example why nixpkgs-fmt is superior: It does not enforce line lengths.

modules/services/mpd.nix
```diff
@@ -162,25 +160,24 @@ in {
         Environment = "PATH=${config.home.profileDirectory}/bin";
         ExecStart = "${cfg.package}/bin/mpd --no-daemon ${mpdConf}";
         Type = "notify";
-        ExecStartPre = ''${pkgs.bash}/bin/bash -c "${pkgs.coreutils}/bin/mkdir -p '${cfg.dataDir}' '${cfg.playlistDirectory}'"'';
+        ExecStartPre = ''
+          ${pkgs.bash}/bin/bash -c "${pkgs.coreutils}/bin/mkdir -p '${cfg.dataDir}' '${cfg.playlistDirectory}'"'';
```
SuperSandro2000, line 164: and such change in nixpkgs can cause a mass rebuild and break things dramatically.

rycee: Hmm, I'm not certain why this would cause a mass rebuild? As far as I can tell they are semantically identical. The only change seems to be the initial newline and indentation, which should be stripped by Nix.

tests/modules/programs/gpg/override-defaults.nix
rycee: I think this is very much a matter of taste. Personally, I don't mind this wrapping at all.

modules/services/mpd.nix
rycee: That said, I think this code could be made more clear by using `pkgs.writeScript`.

Issue #2767, opened 2022-02-28
bug: xdg module breaks when used with current nixpkgs-unstable


euank:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

As was [announced upstream](https://discourse.nixos.org/t/breaking-changes-announcement-for-unstable/17574/7), `pkgs.makeDesktopItem` has a few breaking changes.

home-manager uses this library function in the xdg module here:

https://github.com/nix-community/home-manager/blob/ea85f4b1fdf3f25cf97dc49f4a9ec4eafda2ea25/modules/misc/xdg-desktop-entries.nix#L145-L156

The following changes to the method API impact home-manager's use of it, by my reading:

1. `mimeType` argument renamed to `mimeTypes`
2. `extraDesktopEntries` renamed to `extraConfig`
3. `categories` is now a list, not a semicolon separated string
4. `extraEntries` is removed, and all attributes are fully modeled in nix

Each of those will require changes to the xdg-module, with the 4th one requiring the most work I think.

For a specific example of something that breaks:

```nix
{
  xdg = {
    enable = true;
    desktopEntries = {
      firefox = {
        name = "Firefox";
        genericName = "Web Browser";
        exec = "firefox %U";
        terminal = false;
        categories = [ "Application" "Network" "WebBrowser" ];
        mimeType = [ "text/html" "text/xml" ];
      };
    };
  };
}
```

The above is taken from the example for `options.xdg.desktopEntries`, and with current nixos-unstable, results in:

```
error: anonymous function at /nix/store/6lk1r8frx7rw1spfbfj2hi88fb853w17-source/pkgs/build-support/make-desktopitem/default.nix:7:1 called with unexpected argument 'extraEntries'

       at /nix/store/1df131fdshngr3ydw2l98isy81rw2p4s-source/modules/misc/xdg-desktop-entries.nix:145:5:
```

### Maintainer CC

@cw

### System information

```markdown
$ nix flake metadata nixpkgs
Resolved URL:  github:NixOS/nixpkgs/nixpkgs-unstable
Locked URL:    github:NixOS/nixpkgs/22dc22f8cedc58fcb11afe1acb08e9999e78be9c
$ nix flake metadata home-manager
Resolved URL:  github:nix-community/home-manager
Locked URL:    github:nix-community/home-manager/ea85f4b1fdf3f25cf97dc49f4a9ec4eafda2ea25
```

PR #2770, opened 2022-03-02
Issues with DesktopEntries after #2496

SebTM:
### Description

My system wouldn't build due to different errors after #2496 (nixpkgs unstable, hm master - both local checkouts contains what I've seen the relevant commits made for this), I adjusted the module from what I found as "wrong". (with [help](https://github.com/NixOS/nixpkgs/commit/066a581a80c0939423ff8b2d4b451a07ea8e5865) of and the current code)

I'm unsure if there is a replacement for "extraEntries = config.extraConfig;" as changelog says it got removed (but I left it in for feedback). Currently not even the example from the HM-Docs was working for me, which brought me on point - maybe we should add a test against it.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/xdg-desktop-entries.nix
+++ b/modules/misc/xdg-desktop-entries.nix
@@ -4,8 +4,12 @@ with lib;
 
 let
   desktopEntry = {
+    imports = [
+      (mkRemovedOptionModule [ "extraConfig" ]
+        "The `extraConfig` option of `xdg.desktopEntries` has been removed following a change in Nixpkgs.")
+    ];
     options = {
-      # Since this module uses the nixpkgs/pkgs/build-support/make-desktopitem function, 
+      # Since this module uses the nixpkgs/pkgs/build-support/make-desktopitem function,
       # our options and defaults follow its parameters, with the following exceptions:
 
       # `desktopName` on makeDesktopItem is controlled by `name`.
@@ -13,14 +17,10 @@ let
       # `name` on makeDesktopItem is controlled by this module's key in the attrset.
       # This is the file's filename excluding ".desktop".
 
-      # `extraEntries` on makeDesktopItem is controlled by `extraConfig`,
-      # and `extraDesktopEntries` by `settings`,
+      # `extraConfig` on makeDesktopItem is controlled by `settings`,
       # to match what's commonly used by other home manager modules.
 
-      # `mimeType` and `categories` on makeDesktopItem ask for a string in the format "one;two;three;",
-      # for usability's sake we ask for a list of strings.
-
-      # Descriptions are taken from the desktop entry spec: 
+      # Descriptions are taken from the desktop entry spec:
       # https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#recognized-keys
 
       type = mkOption {
@@ -31,7 +31,7 @@ let
 
       exec = mkOption {
         description = "Program to execute, possibly with arguments.";
-        type = types.str;
+        type = types.nullOr types.str;
       };
 
       icon = mkOption {
@@ -78,9 +78,9 @@ let
 
       startupNotify = mkOption {
         description = ''
-          If true, it is KNOWN that the application will send a "remove" 
-          message when started with the <literal>DESKTOP_STARTUP_ID</literal> 
-          environment variable set. If false, it is KNOWN that the application 
+          If true, it is KNOWN that the application will send a "remove"
+          message when started with the <literal>DESKTOP_STARTUP_ID</literal>
+          environment variable set. If false, it is KNOWN that the application
           does not work with startup notification at all.'';
         type = types.nullOr types.bool;
         default = null;
@@ -103,19 +103,10 @@ let
         default = null;
       };
 
-      extraConfig = mkOption {
-        description = ''
-          Extra configuration. Will be appended to the end of the file and 
-          may thus contain extra sections.
-        '';
-        type = types.lines;
-        default = "";
-      };
-
       settings = mkOption {
         type = types.attrsOf types.string;
         description = ''
-          Extra key-value pairs to add to the <literal>[Desktop Entry]</literal> section. 
+          Extra key-value pairs to add to the <literal>[Desktop Entry]</literal> section.
           This may override other values.
         '';
         default = { };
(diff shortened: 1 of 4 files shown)


DanielSiepmann:
That fix should solve #2767?

SebTM:
> That fix should solve #2767?

Oh yes, I hate the GitHub search it doesn't showed me the issue and this time I searched everywhere for it :/

SebTM:
Okay, got it working mostly but there is one issue I'm not quite sure how to handle:

As I use "extraConfig = config.settings;" as Changelog says extraDesktopEntries got extraConfig - the test will fail because the value our extraConfig-Option which also seem to have another type (string vs object)? Do we want to drop the option in favor of settings?

berbiche (changes requested):
The only way to resolve the incompatibility quickly is to mark the option `extraConfig` as dropped with `mkRemovedOptionModule`.

Here's what I would add:
``` patch
diff --git a/modules/misc/xdg-desktop-entries.nix b/modules/misc/xdg-desktop-entries.nix
index 2c2f8079..f6db9feb 100644
--- a/modules/misc/xdg-desktop-entries.nix
+++ b/modules/misc/xdg-desktop-entries.nix
@@ -4,8 +4,12 @@ with lib;
 
 let
   desktopEntry = {
+    imports = [
+      (mkRemovedOptionModule [ "extraConfig" ] ''
+        The `extraEntries` option of `xdg.desktopEntries` has been removed following a change in Nixpkgs.'')
+    ];
     options = {
-      # Since this module uses the nixpkgs/pkgs/build-support/make-desktopitem function, 
+      # Since this module uses the nixpkgs/pkgs/build-support/make-desktopitem function,
       # our options and defaults follow its parameters, with the following exceptions:
 
       # `desktopName` on makeDesktopItem is controlled by `name`.
@@ -13,14 +17,10 @@ let
       # `name` on makeDesktopItem is controlled by this module's key in the attrset.
       # This is the file's filename excluding ".desktop".
 
-      # `extraEntries` on makeDesktopItem is controlled by `extraConfig`,
-      # and `extraDesktopEntries` by `settings`,
+      # `extraConfig` on makeDesktopItem is controlled by `settings`,
       # to match what's commonly used by other home manager modules.
 
-      # `mimeType` and `categories` on makeDesktopItem ask for a string in the format "one;two;three;",
-      # for usability's sake we ask for a list of strings.
-
-      # Descriptions are taken from the desktop entry spec: 
+      # Descriptions are taken from the desktop entry spec:
       # https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#recognized-keys
 
       type = mkOption {
@@ -31,7 +31,7 @@ let
 
       exec = mkOption {
         description = "Program to execute, possibly with arguments.";
-        type = types.str;
       };
 
       icon = mkOption {
@@ -78,9 +78,9 @@ let
 
       startupNotify = mkOption {
         description = ''
-          If true, it is KNOWN that the application will send a "remove" 
-          message when started with the <literal>DESKTOP_STARTUP_ID</literal> 
-          environment variable set. If false, it is KNOWN that the application 
+          If true, it is KNOWN that the application will send a "remove"
+          message when started with the <literal>DESKTOP_STARTUP_ID</literal>
+          environment variable set. If false, it is KNOWN that the application
           does not work with startup notification at all.'';
         type = types.nullOr types.bool;
         default = null;
@@ -103,19 +103,10 @@ let
         default = null;
       };
 
-      extraConfig = mkOption {
-        description = ''
-          Extra configuration. Will be appended to the end of the file and 
-          may thus contain extra sections.
-        '';
-        type = types.lines;
-        default = "";
-      };
-
       settings = mkOption {
         type = types.attrsOf types.string;
         description = ''
-          Extra key-value pairs to add to the <literal>[Desktop Entry]</literal> section. 
+          Extra key-value pairs to add to the <literal>[Desktop Entry]</literal> section.
           This may override other values.
         '';
         default = { };
@@ -132,6 +123,15 @@ let
         description = "Whether to validate the generated desktop file.";
         default = true;
       };
+
+      # Required for the assertions
+      # TODO: Remove me once `mkRemovedOptionModule` works correctly with submodules
+      assertions = mkOption {
+        type = types.listOf types.unspecified;
+        default = [ ];
+        visible = false;
+        internal = true;
+      };
     };
   };
 
@@ -147,8 +147,8 @@ let
         type exec icon comment terminal genericName startupNotify noDisplay
         prefersNonDefaultGPU;
       desktopName = config.name;
-      mimeTypes = if config.mimeType == null then [ ] else config.mimeType;
-      categories = if config.categories == null then [ ] else config.categories;
+      mimeTypes = optionals (config.mimeType != null) config.mimeType;
+      categories = optionals (config.categories != null) config.categories;
       extraConfig = config.settings;
     };
 in {
@@ -179,7 +179,7 @@ in {
   config = mkIf (config.xdg.desktopEntries != { }) {
     assertions = [
       (hm.assertions.assertPlatform "xdg.desktopEntries" pkgs platforms.linux)
-    ];
+    ] ++ flatten (catAttrs "assertions" (attrValues config.xdg.desktopEntries));
 
     home.packages = (map hiPrio # we need hiPrio to override existing entries
       (attrsets.mapAttrsToList makeFile config.xdg.desktopEntries));
diff --git a/tests/modules/misc/xdg/desktop-entries.nix b/tests/modules/misc/xdg/desktop-entries.nix
index adb6e31e..96fbaf5f 100644
--- a/tests/modules/misc/xdg/desktop-entries.nix
+++ b/tests/modules/misc/xdg/desktop-entries.nix
@@ -48,6 +48,15 @@ with lib;
       })
     ];
 
+    test.asserts.assertions.expected = let
+      currentFile = toString ./desktop-entries.nix;
+    in [
+      ''
+        The option definition `extraConfig' in `${currentFile}' no longer has any effect; please remove it.
+        The `extraEntries` option of `xdg.desktopEntries` has been removed following a change in Nixpkgs.
+      ''
+    ];
+
     nmt.script = ''
       assertFileExists home-path/share/applications/full.desktop
       assertFileExists home-path/share/applications/min.desktop
diff --git a/tests/modules/misc/xdg/desktop-full-expected.desktop b/tests/modules/misc/xdg/desktop-full-expected.desktop
index c9220fb1..784b6d83 100644
--- a/tests/modules/misc/xdg/desktop-full-expected.desktop
+++ b/tests/modules/misc/xdg/desktop-full-expected.desktop
@@ -14,6 +14,3 @@ StartupNotify=false
 Terminal=true
 Type=Application
 Version=1.4
-[X-ExtraSection]
-Exec=foo -o
-
diff --git a/tests/modules/misc/xdg/desktop-min-expected.desktop b/tests/modules/misc/xdg/desktop-min-expected.desktop
index 9475024a..1b36b57c 100644
--- a/tests/modules/misc/xdg/desktop-min-expected.desktop
+++ b/tests/modules/misc/xdg/desktop-min-expected.desktop
@@ -3,3 +3,4 @@ Exec=test --option
 Name=Test
 Terminal=false
 Type=Application
+Version=1.4
```

SebTM:
@berbiche Thanks for your help on this - have applied your patch and formatted the changes to fix the pipeline :)
If your are good with it, it seems ready to merge 🎉

berbiche (approved):
Thanks!

tasmo:
Please merge soon!

rycee:
I'll fix the commit message and merge in a little while.

rycee:
Merged to master now. Thanks!

Issue #1464, opened 2020-08-29
No programs.bash.enableCompletion?


DamienCassou:
### Issue description

I'm experimenting with home-manager outside of NixOS using Debian. I'm trying to have home-manager handle has many things as possible. I can't find the configuration to have home-manager enable bash completions.

What about adding the following to `programs.bash.bashrcExtra` or similar?

```
      . ${pkgs.bash-completion}/share/bash-completion/bash_completion
```

I can send a PR if you want.

NixOS has bashCompletion enabled by default.

#### Maintainer CC

/cc @rycee 

#### Technical details

 - system: `"x86_64-linux"`
 - host os: `Linux 5.4.59, NixOS, 20.09.git.f4195a89195 (Nightingale)`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.3.7`
 - channels(root): `""`
 - nixpkgs: `/home/cassou/Documents/projects/nix-system/nixpkgs`

acowley:
This was very helpful for me, thank you!

berbiche:
> What about adding the following to `programs.bash.bashrcExtra` or similar?
> 
> ```
>       . ${pkgs.bash-completion}/share/bash-completion/bash_completion
> ```
> 
> I can send a PR if you want.

Was a PR ever created?

DamienCassou:
I never got any answer so I never sent a PR.

-- 
Damien Cassou

"Success is the ability to go from one failure to another without
losing enthusiasm." --Winston Churchill

Issue #191, opened 2018-01-25
environment.pathsToLink alternative for home configuration


uvNikita:
System configuration has [environment.pathsToLink](https://github.com/NixOS/nixpkgs/blob/9c96edf8d7d64afeea0d9d1541e97cabcdb03ba8/nixos/modules/config/system-path.nix#L68) option that can be used to symlink additional directories to system profile. In particular, it's used in [bash](https://github.com/NixOS/nixpkgs/blob/9c96edf8d7d64afeea0d9d1541e97cabcdb03ba8/nixos/modules/programs/bash/bash.nix#L205) and [zsh](https://github.com/NixOS/nixpkgs/blob/9c96edf8d7d64afeea0d9d1541e97cabcdb03ba8/nixos/modules/programs/zsh/zsh.nix#L191) modules to add shell completion for packages that provide it.

I wonder if we could have a similar option, but for home configuration, meaning that declaring something like `environment.pathsToLink = "/share/zsh"` would symlink this folder from all packages to `~/.nix-profile/share/zsh`.

rycee:
Hmm, I'm a bit confused. What do you mean by *all* packages here. For packages in `home.packages` I understand their `/share/zsh` path will already be linked from `~/.nix-profile/share/zsh`.

uvNikita:
Nevermind, I was a bit confused and thought to have system packages' `/share/zsh` folders in `~/.nix-profile`. But then it makes sense to set `environment.pathsToLink` parameter in system configuration and reference it from `.zshrc` :)

Fixed system packages completions here https://github.com/rycee/home-manager/pull/196.

shlevy:
What about having headers available in the user profile? Currently I don't see `/include` linked in.

rycee:
@shlevy Aren't the headers typically placed in a `dev` output? You can try adding

    home.extraOutputsToInstall = [ "dev" ];

to your configuration.

shlevy:
I have that as well, though in this case I do have `/include` in a non-dev output of one of my drvs

PR #2783, opened 2022-03-09
bash: add bash-completion

maximsmol:
### Description

Add `bash-completion` support to the bash module, similar to the [NixOS module](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/programs/bash/bash-completion.nix). Solves #1464 + related to #191

Note sure what to do about tests with this one since it does not seem we currently test `.bashrc` (`initExtra`) in any way but will add one if needed

Note: sorry for the awful diff, the file was not formatted correctly on master so `nixfmt` messed it up

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -59,7 +59,7 @@
 
 /modules/programs/autorandr.nix                       @uvNikita
 
-/modules/programs/bash.nix                            @rycee
+/modules/programs/bash.nix                            @rycee @maximsmol
 
 /modules/programs/bat.nix                             @marsam
 
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -42,10 +42,14 @@ let
       };
     };
 
-    config = { id = mkDefault (builtins.hashString "sha256" config.message); };
+    config = {
+      id = mkDefault (builtins.hashString "sha256" config.message);
+    };
   });
 
-in {
+in
+
+{
   meta.maintainers = [ maintainers.rycee ];
 
   options = {
@@ -99,7 +103,7 @@ in {
       entries = mkOption {
         internal = true;
         type = types.listOf entryModule;
-        default = [ ];
+        default = [];
         description = "News entries.";
       };
     };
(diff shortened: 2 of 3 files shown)


maximsmol:
By the way, this makes a lot of the packages' bash completion work out of the box without any extra integration steps. So a lot of the `enableBashIntegration` options become kind of useless?

This also seems to already be the case for Zsh since it [also adds the nix profile paths to `fpath`?](https://github.com/nix-community/home-manager/blob/master/modules/programs/zsh.nix#L472-L474)

rycee:
Thanks for the contribution! I think most people enable bash completion option through the NixOS option. Will this change work OK with that? For example, will it enable bash completion twice?

Also, could you restore the formatting of the file? The Bash module is intentionally not formatted through nixfmt. You can use the `format` script in the project root to do code formatting and it will omit the excluded files.

wmertens:
this seems useful but indeed it would cause the completions to load twice.

However, the bash completion script itself sets `BASH_COMPLETION_VERSINFO=(2 11)`, so if that is set, the bash completion could be skipped?

@rycee out of interest, why is this file not autoformatted? Would `nixpkgsfmt` perform better?

rycee:
The `news.nix` file is not auto formatted since it would cause merge conflicts in various PRs. At some point the formatting exception should be removed.

I think the `enableXyzIntegration` options are best to be considered for things like starship and mcfly, i.e., more invasive shell integrations. Plain shell completions for installed tools should, in my opinion, always be installed by the Nixpkgs package and used automatically with no option to disable them.

Checking for `BASH_COMPLETION_VERSINFO` and making the sourcing conditional on that seems like a decent compromise.

jameysharp:
Disclaimer: I just started using home-manager this week and haven't looked closely at how programmable completion works before.

That said: it looks to me like you don't need to source anything from `/share/bash-completion/completions/`. The `bash_completion` script itself does so as-needed for every path in `$XDG_DATA_DIRS`, and it looks like something already adds the `share` subdir to that from each path in `$NIX_PROFILES`.

I'd love to see this merged; I was quite surprised, on trying out home-manager for the first time, to not get completions for man pages. I have a new machine where I have to use Ubuntu after years with NixOS, and I'm hoping to get some of the NixOS feel back using home-manager.

rycee:
Created #3131 with an alternative implementation, which deprecates this one.

PR #2785, opened 2022-03-10
bspwm: add alwaysResetDesktops

ncfavier:
### Description

If set to true, desktops configured in `monitors` will be reset every time the config is run.

If set to false, desktops will only be configured the first time the config is run.
This is useful if you want to dynamically add desktops and you don't want them to be destroyed if you re-run `bspwmrc`.

Tested.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


berbiche (approved):
Thanks!

Issue #2751, opened 2022-02-22
bug: environment variables are unavailable in ZSH since #2708 was merged


berbiche:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Context: https://github.com/nix-community/home-manager/commit/2116fe6b50a5118d56f1f443cccf024abee9de40#r67122011

### System information

```markdown
So far, this occurs on Darwin only.
```

clhodapp:
This is an issue for at least `PROMPT` on NixOS: https://github.com/nix-community/home-manager/pull/2708#issuecomment-1047520943

Maybe we should have an explicit setting for prompt init in home-manager to make it less fragile against stuff like this...

jwiegley:
It does seem that all of my settings do get set, including the `PROMPT`, it just doesn't have an effect:
```
johnw@Vulcan> env | grep PROMPT                                                                   ~
PROMPT=%m %~ $
PROMPT_DIRTRIM=2
RPROMPT=
johnw@Vulcan> echo $PROMPT                                                                        ~
%B%(?..[%?] )%b%n@%U%m%u>
johnw@Vulcan> export PROMPT='%m %~ $ '                                                            ~
Vulcan ~ $                                                                                        ~
```

jian-lin:
@jwiegley 

As I posted in https://github.com/nix-community/home-manager/commit/2116fe6b50a5118d56f1f443cccf024abee9de40#r67162926 , I think the reason for your output is that:

1. you export `PROMPT` in your `.zshenv`,
1. `PS1` or `PROMPT` is set in `/etc/zshrc` but not exported,
1. `env` only prints exported variables.

I can reproduce your output:
```
$ setopt interactivecomments
$ export PROMPT="exported_prompt> " # set by your .zshenv
exported_prompt> PS1="not_exported_prompt> " # set in /etc/zshrc. PS1 and PROMPT are the same, they use PS1 IIRC
not_exported_prompt> env | grep PROMPT
PROMPT=exported_prompt> 
not_exported_prompt> echo $PROMPT 
not_exported_prompt> 
not_exported_prompt>
```

What is the output of `echo $PROMPT` when your prompt is not what you set? Can you post content related to `PROMPT` or `PS1` in your  `/etc/zshrc` and `.zshrc` to confirm my guess?

If my guess is true, then it's not a bug, it a wrong user config.

`PROMPT` should not be exported since different shells have different syntax for it and will set it during initialization.
[...]

Issue #2739, opened 2022-02-19
bug: EDITOR and VISUAL are hardcoded when using prezto+zsh


Retrodynen:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Hi,

When using the below:

```nix
programs.zsh.prezto.enable = true;
home.sessionVariables.ALTERNATE_EDITOR = "emacs";
home.sessionVariables.EDITOR = "emacsclient -c";
home.sessionVariables.VISUAL = "emacsclient -c";
programs.zsh.envExtra = ''
export EDITOR='emacsclient -c'
export VISUAL='emacsclient -c'
'';
```

The EDITOR and VISUAL environment variables stay as 'nano' due to https://github.com/sorin-ionescu/prezto/blob/6c5657be8ef5f36054524ba171f254d5d9073e24/runcoms/zprofile#L20-L21

https://github.com/nix-community/home-manager/blob/master/modules/programs/zsh/prezto.nix#L354-L355

If I try envExtra or extraConfig.. the lines to change EDITOR and VISUAL are added to the top of the file which makes them useless as the hardcoded `EDITOR=nano` overrides it anyway :(

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 5.15.22, NixOS, 22.05 (Quokka)`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.6.0`
 - channels(root): `"home-manager, nixos"`
 - channels(syl): `""`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`
```

jian-lin:
Sorry my #2708 expose this issue
see https://github.com/nix-community/home-manager/issues/2751#issuecomment-1065964222

pinpox:
https://github.com/sorin-ionescu/prezto/pull/1992 has been merged and fixes this problem, just needs an update.

pinpox:
https://github.com/NixOS/nixpkgs/pull/168255 should fix this

holymonson:
Would someone confirm if https://github.com/NixOS/nixpkgs/pull/167843 has fixed this?

oati:
It was fixed for me

pinpox:
Just updated to latest unstable branch, can confirm `EDITOR` and `VISUAL` are now correct.

berbiche:
Thanks for resolving this issue at the source :smiley:

[...]

PR #2787, opened 2022-03-11
zsh: add `envExtraFirst` option + consolidate `.zshenv` generated file additions

montchr:
- Add a new option – `programs.zsh.envExtraFirst`. This option is the equivalent to `programs.zsh.initExtraFirst`, allowing for user-specified lines to be added at the very beginning of the generated `.zshenv` file.
- Combine all additions to `.zshenv` into one multiline block to avoid implicit/unclear ordering of the additions to the file.
- This may address #2751 by ensuring that `envExtra` is predictably appended to the end of `.zshenv`, which, based on my understanding of the issues in that ticket, would retain something similar to the initial order of `.zshenv` prior to #2708.

### Description

Currently, the `.zshenv` file generated by this module relies on an
implicit order of sections via the module's invocation of `lib.mkMerge`.
The current state leaves the resulting `.zshenv` file output unclear at
a glance.

With the changes in this commit, the text of the `.zshenv` file is
written explicitly in a single multiline string, with the conditional
sections handled with `lib.optionalString`. This approach makes the
text contents of the generated `.zshenv` file clear from reading the
source code.

By writing the generated file explicitly, at a glance we can easily
determine the order in which the various config values are loaded:

1. `envExtraFirst` – new option
2. `$ZDOTDIR`
3. oh-my-zsh vars
4. sourced home-manager session vars
5. `envVarsStr`
6. `envExtra`

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/zsh.nix
+++ b/modules/programs/zsh.nix
@@ -405,10 +405,6 @@ in
   };
 
   config = mkIf cfg.enable (mkMerge [
-    (mkIf (cfg.envExtra != "") {
-      home.file."${relToDotDir ".zshenv"}".text = cfg.envExtra;
-    })
-
     (mkIf (cfg.profileExtra != "") {
       home.file."${relToDotDir ".zprofile"}".text = cfg.profileExtra;
     })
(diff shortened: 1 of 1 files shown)


modules/programs/zsh.nix
```diff
@@ -444,6 +436,17 @@ in
...
+        ${cfg.envExtraFirst}
+
+        ${optionalString (cfg.dotDir != null) ''
+          ZDOTDIR=${zdotdir}
+        ''}
+
+        ${optionalString cfg.oh-my-zsh.enable ''
+          ZSH="${pkgs.oh-my-zsh}/share/oh-my-zsh";
+          ZSH_CACHE_DIR="${config.xdg.cacheHome}/oh-my-zsh";
+        ''}
```
montchr, line 448: I'm not an OMZ user, so I haven't verified this order personally. But considering the variables in question, it seemed to me like the kind of early-init configuration that belongs towards the top of the file.

jian-lin (approved):

jian-lin:
Actually, [prezto](https://github.com/nix-community/home-manager/blob/afe96e7433c513bf82375d41473c57d1f66b4e68/modules/programs/zsh/prezto.nix#L362-L363) also append something to `.zshenv`, which causes issue with `EDITOR` and `VISUAL`, see #2739 and [this](https://github.com/nix-community/home-manager/issues/2751#issuecomment-1065946619).

montchr:
@jian-lin Good call! I'll take a deeper look to see if there's anything else adding to these files that might need updating…

teto:
if this gets updater, it should follow the same approach as https://github.com/nix-community/home-manager/pull/6479 and rely on mkOrder IMO

teto:
it's made obsolete by https://github.com/nix-community/home-manager/pull/6479. Some changes I dont understand like reexporting ZDOTDIR. Closing as it's old but feel free to submit a more focused PR.

PR #2788, opened 2022-03-12
irssi: fix syntax error when no channels are specified

ReplayCoding:
### Description

TL;DR: Don't nest concatStringsSep

`concatStringsSep` will return an empty string when called with empty array. The `concatStringsSep` called with `v.channels` would generate an empty string for each channel:

```
programs.irssi = {
  networks = {
    network1 = {
      ...
    };
    network2 = {
      ...
    };
  };
};
```
generates `[ "" "" ]`, which the `concatStringsSep` called with `v.networks` will use to generate
```
,

```

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/irssi.nix
+++ b/modules/programs/irssi.nix
@@ -44,14 +44,15 @@ let
       }
     ''));
 
-  channelString = concatStringsSep cnl (flip mapAttrsToList cfg.networks (k: v:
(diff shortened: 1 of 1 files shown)


rycee:
Thanks! Merged to master now.

PR #2795, opened 2022-03-15
autorandr: add filter option

somasis:
### Description

xrandr's `--filter` option is useful for enabling integer scaling on displays that don't have support for it at a graphics card level; autorandr doesn't add this option itself to generated profiles, but happily accepts it if the user adds it to the config.

As an alternative to this change, perhaps an `extraConfig` attribute would be suitable; I didn't take that approach because I only needed this one option added.

### Checklist

FIXME(?): seems that `modules/misc/xdg-desktop-entries.nix` made them fail on my machine...?

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/autorandr.nix
+++ b/modules/programs/autorandr.nix
@@ -176,6 +176,13 @@ let
           }
         '';
       };
+
+      filter = mkOption {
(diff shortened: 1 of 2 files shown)


uvNikita (approved):
LGTM

Issue #2081, opened 2021-06-06
Xcursor module not follow XDG specification


SeraphyBR:
### Issue description

I believe this module should be creating the file in ".icons/default/index.theme", according to https://wiki.archlinux.org/title/Cursor_themes#XDG_specification, just like lxappearance does. 
Without this file my AwesomeWM will not use the cursor theme.

#### Maintainer CC

@rycee 

#### Technical details

```
 - system: `"x86_64-linux"`
 - host os: `Linux 5.12.7, NixOS, 21.11pre293730.1ca6b0a0cc3 (Porcupine)`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.3.12`
 - channels(root): `"nixos-21.11pre293730.1ca6b0a0cc3, home-manager, nixpkgs-21.03pre249813.26d3fbf2156"`
 - channels(seraphybr): `""`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`
```

sumnerevans:
CC: @league

league:
Thanks for the report. I'm not currently using this module, since I switched to KDE Plasma a few months ago. However I see the notes on the ArchWiki page you linked. **Can you add something like the following to your config,** and see if it does the right thing? If so, it would be easy to add to this module…

```nix
    home.file.".icons/default/index.theme".text = ''
      [icon theme]
      Inherits=${config.xsession.pointerCursor.name}
    '';
```

I'm not sure whether `index.theme` is intended to have other settings that could be clobbered by this. There doesn't seem to be a facility in home-manager (yet) for updating that file from various settings, so we'd be replacing it wholesale – if that makes sense. Thanks.

SeraphyBR:
Yes it works, I ended up doing it in my config as a temporary solution
![image](https://user-images.githubusercontent.com/33378179/121081384-0ca2c780-c7b3-11eb-8d7e-72cfa60e9c17.png)

Narice:
this doesn't work for me and I don't know why :/
I am using the[ `Qogir Cursors`](https://github.com/NixOS/nixpkgs/blob/nixos-21.05/pkgs/data/icons/qogir-icon-theme/default.nix#L27) theme:
```
        pointerCursor = {
            package = pkgs.qogir-icon-theme; # FIXME: default cursor on i3 instead of qogir
            name = "Qogir Cursors";
        };
```
I am on i3 + plasma.
[...]

PR #2805, opened 2022-03-19
xsession.pointerCursor: escape special characters in the cursor path

polykernel:
- Escape special character in the cursor name and default cursor file name
  in the cursor path for the xsetroot command.

### Description

small change to address https://github.com/nix-community/home-manager/issues/2081#issuecomment-859580927

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/xcursor.nix
+++ b/modules/xcursor.nix
@@ -36,6 +36,10 @@ let
     };
   };
 
+  cursorPath = "${cfg.package}/share/icons/${escapeShellArg cfg.name}/cursors/${
+      escapeShellArg cfg.defaultCursor
+    }";
+
 in {
   meta.maintainers = [ maintainers.league ];
 
@@ -68,9 +72,7 @@ in {
     home.packages = [ cfg.package ];
 
     xsession.initExtra = ''
-      ${pkgs.xorg.xsetroot}/bin/xsetroot -xcf ${cfg.package}/share/icons/${cfg.name}/cursors/${cfg.defaultCursor} ${
-        toString cfg.size
-      }
+      ${pkgs.xorg.xsetroot}/bin/xsetroot -xcf ${cursorPath} ${toString cfg.size}
     '';
 
     xresources.properties = {


modules/xcursor.nix
```diff
@@ -68,7 +68,7 @@ in {
     home.packages = [ cfg.package ];
 
     xsession.initExtra = ''
-      ${pkgs.xorg.xsetroot}/bin/xsetroot -xcf ${cfg.package}/share/icons/${cfg.name}/cursors/${cfg.defaultCursor} ${
+      ${pkgs.xorg.xsetroot}/bin/xsetroot -xcf "${cfg.package}/share/icons/${cfg.name}/cursors/${cfg.defaultCursor}" ${
```
ncfavier, line 71:
```suggestion
      ${pkgs.xorg.xsetroot}/bin/xsetroot -xcf ${cfg.package}/share/icons/${escapeShellArg cfg.name}/cursors/${escapeShellArg cfg.defaultCursor} ${
```

(Packages are top-level store paths and don't contain problematic characters)

ncfavier (changes requested):

modules/xcursor.nix
polykernel: I didn't know one can escape only problematic portions of the path. Thanks!

modules/xcursor.nix
```diff
@@ -68,9 +68,9 @@ in {
     home.packages = [ cfg.package ];
 
     xsession.initExtra = ''
-      ${pkgs.xorg.xsetroot}/bin/xsetroot -xcf ${cfg.package}/share/icons/${cfg.name}/cursors/${cfg.defaultCursor} ${
-        toString cfg.size
-      }
+      ${pkgs.xorg.xsetroot}/bin/xsetroot -xcf ${cfg.package}/share/icons/${
+        escapeShellArg cfg.name
+      }/cursors/${escapeShellArg cfg.defaultCursor} ${toString cfg.size}
```
sumnerevans, line 73: It may be worth extracting the cursor path to a variable.

sumnerevans (changes requested):
Can you please update the commit description to reflect what you actually did here? (It's not wrapping in quotes anymore)

sumnerevans (approved):

PR #2807, opened 2022-03-20
docs: fix README links

kidonng:
### Description

Fix some links in README which are either broken or redirected.

If anyone has better place to link to, suggestions welcome.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,18 +110,18 @@ License
 
 This project is licensed under the terms of the [MIT license](LICENSE).
 
-[Nix]: https://nixos.org/nix/
+[Nix]: https://nixos.org/explore.html
 [NixOS]: https://nixos.org/
-[Nixpkgs]: https://nixos.org/nixpkgs/
-[manual]: https://nix-community.github.io/home-manager/
+[Nixpkgs]: https://github.com/NixOS/nixpkgs
+[manual]: https://nix-community.github.io/home-manager/index.html
 [contributing]: https://nix-community.github.io/home-manager/#ch-contributing
 [manual usage]: https://nix-community.github.io/home-manager/#ch-usage
 [configuration options]: https://nix-community.github.io/home-manager/options.html
 [#home-manager]: https://webchat.oftc.net/?channels=home-manager
 [OFTC]: https://oftc.net/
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -109,19 +109,20 @@ License
 
 This project is licensed under the terms of the [MIT license](LICENSE).
 
-[Nix]: https://nixos.org/nix/
+[Nix]: https://nixos.org/guides/how-nix-works.html
 [NixOS]: https://nixos.org/
-[Nixpkgs]: https://nixos.org/nixpkgs/
-[manual]: https://nix-community.github.io/home-manager/
+[Nixpkgs]: https://github.com/NixOS/nixpkgs
+[manual]: https://nix-community.github.io/home-manager/index.html
```
sumnerevans, line 115: this link isn't broken. https://nix-community.github.io/home-manager/ serves the `index.html` page

sumnerevans (changes requested):
Not sure about the changes to the `Nix` and `Nixpkgs` links either. I don't remember what those links used to have.

README.md
kidonng: This avoids a unnecessary refresh when you click the links in table of contents which explicitly uses `/index.html`. This is also the case for the manual links below.

kidonng:
> Not sure about the changes to the `Nix` and `Nixpkgs` links either. I don't remember what those links used to have.

The last present version of [Nix](https://web.archive.org/web/20200426034512/https://nixos.org/nix/) and [Nixpkgs](https://web.archive.org/web/20200424062842/https://nixos.org/nixpkgs/) pages are found on Wayback Machine. The source files of these pages are removed in https://github.com/NixOS/nixos-homepage/pull/453, as a result of https://github.com/NixOS/nixos-homepage/pull/425.

From my understanding:

- The little info on Nix page is nuked. The Nixpkgs repository README indeed does not contain all that info, but I wonder if it is that useful for people who aren't familiar with Nix/NixOS.
- The info on Nix page is consolidated into the index page and "Features" page (which is now the "Explore" page). Since NixOS is already pointing to the index page, the "Explore" page seems to be the only destination what Nix can refer to, just my two cents.

PR #2818, opened 2022-03-21
lib.booleans: add yesNo function

### LINKED ISSUE #2772: Add Lib function for "yes" and "no"
### Description
As suggested from @sumnerevans in #2314 we could think about:

Maybe it's worth extracting this out to a function while you are at it. See the yesno function in the neomutt module.
We could probably extract it out to lib as well, but I'm fine with doing that in a separate PR.

This is the proper follow-up ticket for this :)

sumnerevans:
Closes #2772

### Description

Unify all of the "yesno" functions into a single function in `lib.hm`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


berbiche (approved):
LGTM!

PR #2819, opened 2022-03-22
i3-sway: Remove extra argument in call to `optionals`

stephaneyfx:
This would give the error "attempt to call something which is not a
function but a list" given that `optionals a b` returns a list.

### Description

This fixes the following error when running `home-manager switch` and `sway` is used in the configuration:
```text
error: attempt to call something which is not a function but a list

       at /home/.../src/home-manager/modules/services/window-managers/i3-sway/lib/functions.nix:96:16:

           95|               (optionalString (trayOutput != null) "tray_output ${trayOutput}")
           96|               (optionals colorsNotNull (indent
             |                ^
           97|                 (lists.subtractLists [ "" null ] [
```

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


stephaneyfx:
Thank you for your review @sumnerevans! It turns out the empty set argument was needed; it was just passed to the wrong function (`optionals` instead of `indent`). I amended the commit and its description.

stephaneyfx:
I forgot to run the formatting tool this time around so I had to push again. Sorry for the noise and thank you for restarting the tests.

PR #2822, opened 2022-03-23
types: fix `dagOf` behaviour with `mkIf`

ncfavier:
This makes definitions like `home.activation.foo = mkIf false "bar"` work, where previously they would complain about `home.activation.foo.data` being used but not defined.

The crucial part is that we don't call `convertAllToDags` in `dagOf.merge`, because we need to process `mkIf`/`mkMerge` properties first. So we let `attrEquivalent.merge` do its job normally, but give it a type `dagEntryOf` that does the conversion.

Ideally this shouldn't require so much boilerplate; I'd like to implement something like `types.changeInto dagContentType elemType dagEntryAnywhere` in nixpkgs.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/lib/types-dag.nix
+++ b/modules/lib/types-dag.nix
@@ -8,17 +8,27 @@ let
 
   isDagEntry = e: isAttrs e && (e ? data) && (e ? after) && (e ? before);
 
-  dagContentType = elemType:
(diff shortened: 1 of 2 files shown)


ncfavier:
cc @Infinisil @roberth

PR #2823, opened 2022-03-23
gpg: create homedir with 700 permissions

ncfavier:
It can happen in some cases that home-manager first runs before gpg creates its homedir, and it creates it with 755 permissions which the user then needs to change by hand.

Do this in the module instead: before linking files, make sure the homedir exists, and if it doesn't, create it with the right permissions.

We might drop the `mkMerge` if https://github.com/nix-community/home-manager/pull/2822 gets merged first. (EDIT: done)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/gpg.nix
+++ b/modules/programs/gpg.nix
@@ -282,7 +282,12 @@ in {
         source = "${keyringFiles}/pubring.kbx";
       };
 
-    home.activation = mkIf (cfg.publicKeys != [ ]) {
+    home.activation = {
+      createGpgHomedir =
+        hm.dag.entryBetween [ "linkGeneration" ] [ "writeBoundary" ] ''
+          $DRY_RUN_CMD mkdir -m700 -p $VERBOSE_ARG ${escapeShellArg cfg.homedir}
+        '';
+
       importGpgKeys = let
         gpg = "${cfg.package}/bin/gpg";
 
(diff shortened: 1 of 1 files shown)


berbiche (approved):
LGTM! Thanks

dhess:
I'm fairly certain this commit broke `programs.gpg` on macOS. I now get this error during activation:

```
mkdir: cannot create directory ‘/var/empty/.gnupg’: Operation not permitted
```

I have to set `programs.gpg.homedir` explicitly to my own homedir to make it stop complaining.

berbiche:
@dhess 
I'm not sure where the problem lies: the default value for gpg's home directory uses `home.homeDirectory`, an option that is required to be set.
Do you have a value set for this option?

dhess:
No, and I never have.

berbiche:
If you set a value for this option, does it solve the issue?

IRRC the `users.users` option in nix-darwin does not set the right home directory for your user. I cannot find the issue but it has happened before to other Darwin users.

dhess:
Thanks, I'll try that. It's odd that I've never had that set explicitly (nor `home.homeDirectory`) and never run into a problem before this.

ncfavier:
Was your `gpg.conf` created? If your `gpg.homedir` was `/var/empty/.gnupg` it should have been...

dhess:
> Was your `gpg.conf` created? If your `gpg.homedir` was `/var/empty/.gnupg` it shouldn't have been...

I don't understand your question. Since long before this change, I have configured `programs.gpg.settings` and `programs.gpg.scdaemonSettings` and everything worked fine.

dhess:
> edit: I have this in my own config:
> 
> ```nix
>   # Fix xdg.{dataHome,cacheHome} being empty in home-manager
>   users.users.nicolas = {
>     home = "/Users/nicolas";
>     isHidden = false;
>     shell = pkgs.zsh;
>   };
> ```

Thanks, that fixed the issue.

dhoppe:
Is there a particular reason why this has not been backported to release-21.11?

ncfavier:
https://github.com/nix-community/home-manager/pull/2919

PR #2824, opened 2022-03-23
misc: fix nix.conf generation

jqqqqqqqqqq:
nix/nix.conf should generate if cfg.extraOptions is not empty.

### Description

I've encountered `nix/nix.conf` not generated with the following config:

```
nix.extraOptions = ''
  experimental-features = nix-command flakes
'';
```

By checking the config files, this file is only generated when `cfg.settings != {}`. It missed one case that `extraOptions` is outside `settings`.

This PR fixes the logic and generates `nix/nix.conf` correctly.

It doesn't involve any package change so I guess most of the checklists are not applicable?

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/nix.nix
+++ b/modules/misc/nix.nix
@@ -230,7 +230,9 @@ in {
         };
       };
 
-      "nix/nix.conf" = mkIf (cfg.settings != { }) { source = nixConf; };
+      "nix/nix.conf" = mkIf (cfg.settings != { } || cfg.extraOptions != "") {
+        source = nixConf;
+      };
     };
   };
 


berbiche (approved):
LGTM!

Can you run the formatter? The tests are failing because of that.

jqqqqqqqqqq:
> LGTM!
> 
> Can you run the formatter? The tests are failing because of that.

Updated. Somehow it failed to run on my darwin-aarch64 with these errors:
```
error: cannot bootstrap GHC on this platform ('aarch64-darwin' with lib 'defaultLibc')
```

I ran it on another intel Mac and copied the result.

modules/misc/nix.nix
```diff
@@ -230,7 +230,9 @@ in {
         };
       };
 
-      "nix/nix.conf" = mkIf (cfg.settings != { }) { source = nixConf; };
+      "nix/nix.conf" = mkIf (cfg.settings != { } || cfg.extraOptions != "") {
```
otavio, line 233: The change looks good, but the commit log might be more detailed about the use case it is adding support for.

PR #2831, opened 2022-03-23
nix: fix attribute path of nix stable

rycee:


tests/modules/misc/nix/example-settings.nix
```diff
@@ -6,7 +6,7 @@ with lib;
   config = {
     nix = {
       package = config.lib.test.mkStubPackage {
-        version = lib.getVersion pkgs.nixStable;
+        version = lib.getVersion pkgs.nixVersions.stable;
```
SuperSandro2000, line 9: maybe we should add a fallback to nixStable?

rycee: Since this is only in HM master, which follows Nixpkgs unstable I think it should be safe to use the new name. We should take care not to backport this to the 21.11 branch, though.

PR #2840, opened 2022-03-26
nix: add support for `nix profile`

rycee:
Finishing up PR #2833


blaggacao:
@rycee is there a slight chance to back-port this to 21.11? That would be fantastic.

rycee:
@blaggacao I'm would be willing to do that if it turns out not to be too tricky. But I'd prefer to wait another week or so in case any issues arise.

PR #2843, opened 2022-03-26
programs/gnome-shell: init module

terlar:
### Description

Adds programs.gnome-shell for customizing Gnome Shell extensions and theme.

This is take two on #1795 by @tadfisher. Generally just rebased and updated with all new conventions.

- Changed maintainer to myself as I am not sure if @tadfisher is interested in this anymore
- Switched from uuid to extensionUuid without fallback (as it was quite a while that landed in nixpkgs and no-one is using this new module yet)
- Updated changelog
- This setting is now some kind of gVariant and needs to access the value as such `config.dconf.settings."org/gnome/shell".enabled-extensions.value`

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/gnome-shell.nix
+++ b/modules/programs/gnome-shell.nix
@@ -1,0 +1,115 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+let
+  cfg = config.programs.gnome-shell;
+
+  extensionOpts = { config, ... }: {
+    options = {
+      id = mkOption {
+        type = types.str;
+        example = "user-theme@gnome-shell-extensions.gcampax.github.com";
+        description = ''
+          ID of the GNOME Shell extension. If not provided, it
+          will be obtained from `package.extensionUuid`.
+        '';
+      };
+
+      package = mkOption {
+        type = types.package;
+        example = "pkgs.gnome.gnome-shell-extensions";
+        description = ''
+          Package providing a GNOME Shell extension in
+          `$out/share/gnome-shell/extensions/''${id}`.
+        '';
+      };
+    };
+
+    config = mkIf (hasAttr "extensionUuid" config.package) {
+      id = mkDefault config.package.extensionUuid;
+    };
+  };
+
+  themeOpts = {
+    options = {
+      name = mkOption {
+        type = types.str;
+        example = "Plata-Noir";
+        description = ''
+          Name of the GNOME Shell theme.
+        '';
+      };
+
+      package = mkOption {
+        type = types.nullOr types.package;
+        default = null;
+        example = literalExpression "pkgs.plata-theme";
+        description = ''
+          Package providing a GNOME Shell theme in
+          `$out/share/themes/''${name}/gnome-shell`.
+        '';
+      };
+    };
+  };
+
+in {
+  meta.maintainers = [ maintainers.terlar ];
+
+  options.programs.gnome-shell = {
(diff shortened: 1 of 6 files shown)


modules/programs/gnome-shell.nix
```diff
@@ -0,0 +1,114 @@
...
+        type = types.nullOr types.package;
+        default = null;
+        example = literalExpression "pkgs.plata-theme";
+        description = ''
+          Package providing a gnome-shell theme in
+          <filename>$out/share/themes/''${name}/gnome-shell</filename>.
+        '';
+      };
+    };
+  };
```
berbiche, line 52: It would be cool if the default value for the theme option was `gtk.theme.{package,name}`.

Having to configure the theme in yet another place in Home Manager is annoying IMO.

tests/modules/programs/gnome-shell/gnome-shell.nix
```diff
@@ -0,0 +1,94 @@
...
+
+  actualEnabledExtensions = catAttrs "value"
+    config.dconf.settings."org/gnome/shell".enabled-extensions.value;
+
+in {
+  nixpkgs.overlays = [
+    (self: super: {
+      gnome3 = super.gnome3.overrideScope' (gself: gsuper: {
+        gnome-shell-extensions = dummy-gnome-shell-extensions;
+      });
```
berbiche, line 42: Is `pkgs.gnome3` an alias to `pkgs.gnome`?
If so, could you use the latter.

berbiche (changes requested):
Thanks for the contribution.

The test cases are interesting.

tests/modules/programs/gnome-shell/gnome-shell.nix
terlar: Indeed, that seems to be the case:
https://github.com/NixOS/nixpkgs/blob/master/pkgs/top-level/aliases.nix#L469

Will update.

modules/programs/gnome-shell.nix
terlar: Good point, I guess you won't need to configure this option though. Personally I have `gtk.theme` configured, but I am not using this `program.gnome-shell.theme`.

I'm guessing that not all Gnome themes need to be GTK themes?

How would one go about to set a default here? I guess it would have to be mkDefault so a user could override it, but it still needs a conditional for the default as it should only be added if `gtk.theme` was configured.

Also `gtk.theme.package` supports being null (if the theme is part of the base offering), so perhaps that would need to be supported for `programs.gnome-shell.theme.package` as well.

terlar:
Rebased to resolve conflict in news entry.

flokli:
This needs another rebase.

berbiche (approved):
LGTM after a rebase!

terlar:
I'm back from some vacation and have rebased now, so should be ready to go

piegamesde (changes requested):

rycee:
Thanks! Finally merged this now. @piegamesde I think your suggestions seem reasonable so perhaps those could be implemented in future PRs?

terlar:
I agree those comments are good ideas and something we can create another PR for. Especially allowing extension packages should be pretty straightforward. The other one I guess we have to think about the interface regarding installing and not enabling as well as enabling locally installed ones

Issue #2652, opened 2022-01-18
Add option to sway.extraConfig to appear at the top of the config file


JonathanReeve:
### Description

When [using Pywal with Sway](https://github.com/dylanaraps/pywal/wiki/Customization#sway), Pywal has you load a sway include file at the beginning of the config file. This doesn't seem to be possible with the current home-manager setup, since `sway.extraConfig` adds stuff to the bottom of the file, but the Pywal include needs to be at the top of the file.

If there weree an option on sway.extraConfig, or a new option, sway.extraConfigEarly, that would allow for this Pywal include.

jakobkukla:
I would love to have this exact feature in the bspwm module.

berbiche:
> I would love to have this exact feature in the bspwm module.

Would you mind creating another feature request for bspwm?
It would be easier to track each feature requests separately.

JonathanReeve:
This issue is not stale—we're just waiting on #2847.

JonathanReeve:
Again, this issue is not stale, we're still waiting on #2847.

teto:
now that we have earlyExtraCofig AND a pywal module in HM, do you want to have another look @JonathanReeve ?

JonathanReeve:
Yep I think this is done!

PR #2847, opened 2022-03-28
sway, bspwm: add extraConfigEarly

### LINKED ISSUE #2665: Add option to bspwm.extraConfig to appear at the top of the config file
Sometimes it is needed to include files at the top of the configuration file (See [Using Pywal with bspwm](https://github.com/dylanaraps/pywal/wiki/Customization#bspwm) for example). It would be great if there was an option to add extra configuration at the top of the file instead of the default behavior.

mtoohey31:
### Description

Adds extraConfigEarly options for sway and bspwm configuration for use cases where custom instructions must be executed before other configuration options, such as sourcing pywal colour files.

Closes #2665, #2652.

Are there any other modules where this type of option would be appropriate? Sway and BSPWM were the only two I came across requests for, and they're the only two window managers that I'm familiar with so I haven't included anything else yet.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


tests/modules/services/window-managers/bspwm/bspwmrc
```diff
@@ -1,3 +1,5 @@
+
+
```
teto, line 2: we shouldnt have these empty lines.

mtoohey31: Fixed in 1d2a8100cf460c2a3990a37ede135292c7b19e80, the if statements are there to make sure there's a newline in between `extraConfigEarly` and the rest.

mtoohey31:
I just noticed that #2787 does a similar thing for ZSH environment configuration, but it uses a slightly different -First suffix. Would you like me to change the names of these options to `extraConfigFirst` for consistency?

modules/services/window-managers/i3-sway/sway.nix
```diff
@@ -263,53 +263,54 @@ let
...
-        ++ optional (gaps != null) gapsStr # gaps
-        ++ map floatingCriteriaStr floating.criteria # floating
-        ++ map windowCommandsStr window.commands # window commands
-        ++ map startupEntryStr startup # startup
-        ++ map workspaceOutputStr workspaceOutputAssign # custom mapping
-      )
-    else
-      [ ]) ++ (optional cfg.systemdIntegration ''
-        exec "systemctl --user import-environment; systemctl --user start sway-session.target"'')
+    ((if cfg.extraConfigEarly != "" then [ cfg.extraConfigEarly ] else [ ])
```
berbiche, line 266:
```suggestion
    ((optional (cfg.extraConfigEarly != "") cfg.extraConfigEarly)
```

tests/modules/services/window-managers/sway/default.nix
```diff
@@ -2,6 +2,7 @@
   sway-bar-focused-colors = ./sway-bar-focused-colors.nix;
   sway-bindkeys-to-code = ./sway-bindkeys-to-code.nix;
   sway-default = ./sway-default.nix;
+  sway-extra-config = ./sway-extra-config.nix;
```
berbiche, line 5: Could you merge this with another test?
Adding another separate test slows down testing unfortunately.

modules/services/window-managers/bspwm/default.nix
```diff
@@ -61,21 +61,23 @@ in {
 
     home.packages = [ cfg.package ];
 
-    xdg.configFile."bspwm/bspwmrc".source = pkgs.writeShellScript "bspwmrc" ''
-      ${concatStringsSep "\n" (mapAttrsToList formatMonitor cfg.monitors)}
+    xdg.configFile."bspwm/bspwmrc".source = pkgs.writeShellScript "bspwmrc"
+      ((if cfg.extraConfigEarly != "" then cfg.extraConfigEarly + "\n" else "")
```
berbiche, line 65:
```suggestion
      ((optionalString (cfg.extraConfigEarly != "") (cfg.extraConfigEarly + "\n"))
```

berbiche (changes requested):
Thanks, I think using `extraConfigEarly` is better over `extraConfigFirst`.
I believe we use the former syntax much more often.

tests/modules/services/window-managers/sway/default.nix
mtoohey31: Ah, makes sense, done in 84bfccd9205267c31f23da099a84519aa3cd1372.

modules/services/window-managers/bspwm/options.nix
```diff
@@ -177,6 +177,14 @@ in {
       '';
     };
 
+    extraConfigEarly = mkOption {
+      type = types.lines;
+      default = "";
+      description = ''
+        Like extraConfig, except commands are run at the start of the config
+                file.'';
```
rycee, line 185: Looks odd.

modules/services/window-managers/bspwm/default.nix
```diff
@@ -61,21 +61,23 @@ in {
...
     home.packages = [ cfg.package ];
 
-    xdg.configFile."bspwm/bspwmrc".source = pkgs.writeShellScript "bspwmrc" ''
-      ${concatStringsSep "\n" (mapAttrsToList formatMonitor cfg.monitors)}
+    xdg.configFile."bspwm/bspwmrc".source = pkgs.writeShellScript "bspwmrc"
+      ((optionalString (cfg.extraConfigEarly != "") (cfg.extraConfigEarly + "\n"))
+        + ''
+          ${concatStringsSep "\n" (mapAttrsToList formatMonitor cfg.monitors)}
 
-      ${concatStringsSep "\n" (mapAttrsToList formatSetting cfg.settings)}
```
rycee, line 67: Perhaps the lines that are currently before `${cfg.extraConfig}` could instead be added to `extraConfig` with something like `mkOrder 250` and the lines after added with `mkOrder 1750`. Then there would be no need for the `extraConfigEarly` since you could simply use something like `mkOrder 100 …` in your own configuration?

mtoohey31: I'm not familiar with many of the builtin functions, so correct me if I'm wrong, but would that approach only allow for a single extraConfig block in one of the possible positions? A user might want to do something custom at the start like source a colour scheme file and also do something after everything else like launch an application. If this is the case, we could always make the option an array of values whose order could be specified, but that seems excessive.

modules/services/window-managers/bspwm/options.nix
mtoohey31: Fixed in b7a6ebcf16cecff1bf9a2bd4fbeb8d6f422e04fb.

modules/services/window-managers/bspwm/default.nix
mtoohey31: Any thoughts on this ^? No worries if you're busy, I'm off this week though, so I have plenty of time to make changes if you'd like me to.

teto:
have you tried `xdg.configFile."bspwwnmrc".text = lib.mkBefore "my early test"` ?

mtoohey31:
> have you tried `xdg.configFile."bspwwnmrc".text = lib.mkBefore "my early test"` ?

Good idea, I hadn't tried that, but after some testing it doesn't look like it works because the `bspwm/bspwmrc` and `sway/config` files are declared using `.source` with a derivation, so that overrides any value provided via the `text` attribute, if I'm understanding the evaluation of `modules/lib/file-type.nix` correctly.

JonathanReeve:
Any chance we could merge this soon? I keep getting pinged by the stale bot in #2652 because this PR has been open for months.

berbiche (approved):
@mtoohey31 if you are willing to rebase yet again, I believe we could merge this as-is and review the mkBefore/mkAfter support at a later point in time.

mtoohey31:
:+1: rebased, there are a couple tests that are failing locally for me though (firefox-profile-settings and neovim-plugin-config), but I think they're not related?

PR #2850, opened 2022-03-30
git: Add option to use difftastic as diff tool

notarock:
### Description

Add the option to use [difftastic](https://github.com/Wilfred/difftastic), a syntax-aware diff tool, with git

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/git.nix
+++ b/modules/programs/git.nix
@@ -239,6 +239,34 @@ in {
         };
       };
 
+      difftastic = {
+        enable = mkEnableOption "" // {
+          description = ''
+            Enable the <command>difft</command> syntax highlighter.
+            See <link xlink:href="https://github.com/Wilfred/difftastic" />.
+          '';
+        };
+
+        background = mkOption {
+          type = types.enum [ "light" "dark" ];
+          default = "light";
+          example = "dark";
+          description = ''
+            Determines whether difftastic should use the lighter or darker colors
+            for syntax highlithing.
+          '';
+        };
+
+        color = mkOption {
+          type = types.enum [ "always" "auto" "never" ];
+          default = "auto";
+          example = "always";
+          description = ''
+            Determines when difftastic should color its output.
+          '';
+        };
+      };
+
       delta = {
         enable = mkEnableOption "" // {
           description = ''
@@ -274,6 +302,11 @@ in {
   config = mkIf cfg.enable (mkMerge [
     {
       home.packages = [ cfg.package ];
+      assertions = [{
+        assertion = !(cfg.delta.enable && cfg.difftastic.enable);
+        message =
+          "Only one of 'programs.git.delta.enable' or 'programs.git.difftastic.enable' can be set to true at the same time.";
+      }];
 
       programs.git.iniContent.user = {
         name = mkIf (cfg.userName != null) cfg.userName;
@@ -375,6 +408,18 @@ in {
           smudge = concatStringsSep " "
             ([ "git-lfs" "smudge" ] ++ skipArg ++ [ "--" "%f" ]);
         };
+    })
+
+    (mkIf cfg.difftastic.enable {
+      home.packages = [ pkgs.difftastic ];
+
+      programs.git.iniContent = let
+        difftCommand =
+          "${pkgs.difftastic}/bin/difft --color ${cfg.difftastic.color} --background ${cfg.difftastic.background}";
+      in {
+        diff.external = difftCommand;
+        core.pager = "${pkgs.less}/bin/less -XF";
+      };
     })
 
     (mkIf cfg.delta.enable {


modules/programs/git.nix
```diff
@@ -377,6 +403,19 @@ in {
...
 
+    (mkIf cfg.difftastic.enable {
+      home.packages = [ pkgs.difftastic ];
+
+      programs.git.iniContent =
+        let difftCommand = "${pkgs.difftastic}/bin/difft";
+        in {
+          core.pager = difftCommand;
+          interactive.diffFilter =
+            "${difftCommand} --color ${cfg.difftastic.color} --background ${cfg.difftastic.background}";
```
notarock, line 414: This was heavily inspired by how the `Delta` option were defined and used, so I am not sure if this would work exactly as it should

notarock:
Pretty sure it would be good to only allow one of either `difftastic` or `git` but I am now really good with nix, how can I achieve this?

modules/programs/git.nix
```diff
@@ -377,6 +403,19 @@ in {
         };
     })
 
+    (mkIf cfg.difftastic.enable {
```
sumnerevans, line 406: We should add an assertion that `cfg.difftastic.enable` and `cfg.delta.enable` are not enabled at the same time.

modules/programs/git.nix
```diff
@@ -377,6 +403,19 @@ in {
...
+      home.packages = [ pkgs.difftastic ];
+
+      programs.git.iniContent =
+        let difftCommand = "${pkgs.difftastic}/bin/difft";
+        in {
+          core.pager = difftCommand;
+          interactive.diffFilter =
+            "${difftCommand} --color ${cfg.difftastic.color} --background ${cfg.difftastic.background}";
+          difftastic = cfg.difftastic.options;
```
sumnerevans, line 415: `difftastic.options` isn't defined above (looks like you can just configure color and background) so this line is unnecessary

```suggestion
```

sumnerevans (changes requested):

modules/programs/git.nix
notarock: Done, I figured it out

modules/programs/git.nix
```diff
@@ -274,6 +300,11 @@ in {
   config = mkIf cfg.enable (mkMerge [
     {
       home.packages = [ cfg.package ];
+      assertions = [{
+        assertion = !(cfg.delta.enable && cfg.difftastic.enable);
+        message =
+          "Only one of delta or difftastic can be enabled at the same time.";
```
sumnerevans, line 306: I think we should specify the full option paths here.
```suggestion
          "Only one of 'git.delta.enable` or `git.delta.difftastic.enable` can be set to true at the same time.";
```

sumnerevans (changes requested):

modules/programs/git.nix
```diff
@@ -274,6 +300,11 @@ in {
   config = mkIf cfg.enable (mkMerge [
     {
       home.packages = [ cfg.package ];
+      assertions = [{
+        assertion = !(cfg.delta.enable && cfg.difftastic.enable);
+        message =
+          "Only one of 'git.delta.enable` or `git.delta.difftastic.enable` can be set to true at the same time.";
```
sumnerevans, line 306: Oh, I really screwed that suggestion up. This isn't markdown.

```suggestion
          "Only one of 'git.delta.enable' or 'git.delta.difftastic.enable' can be set to true at the same time.";
```

sumnerevans (changes requested):
sorry, one more thing

sumnerevans:
Also, please squash your commits to a single one that follows the contributing guide for commit messages.

modules/programs/git.nix
arjan-s: Shouldn't the second one be 'git.difftastic.enable' instead of 'git.delta.difftastic.enable'?

modules/programs/git.nix
```diff
@@ -239,6 +239,32 @@ in {
...
       };
 
+      difftastic = {
+        enable = mkEnableOption "" // {
+          description = ''
+            Enable the <command>difft</command> syntax highlighter.
+            See <link xlink:href="https://github.com/Wilfred/difftastic" />.
+          '';
+        };
+        background = mkOption {
```
rycee, line 249: Please separate the option definitions by an empty line. It helps make them more distinct. That is,
```suggestion

        background = mkOption {
```

and the same with `color` below.

modules/programs/git.nix
```diff
@@ -239,6 +239,32 @@ in {
...
+            Options to configure if difftastic should use the lighter or darker
+            colors for syntax highlithing.
+          '';
+        };
+        color = mkOption {
+          type = types.enum [ "always" "auto" "never" ];
+          default = "auto";
+          example = "always";
+          description = ''
+            Options to configure if difftastic should colors in it's output.
```
rycee, line 263: Perhaps something like
```suggestion
            Determines when difftastic should color its output.
```

modules/programs/git.nix
rycee: Also misses "programs.". Perhaps something like

> Only one of 'programs.git.delta.enable' or 'programs.git.difftastic.enable' can be enabled at the same time.

notarock: I changed it for 'programs.git.delta.enable` or `programs.git.difftastic.enable`

notarock:
Thanks for the review, I fixed it. Sorry for the constant back-and-forth, hopefully this is good now :sweat_smile:

modules/programs/git.nix
```diff
@@ -274,6 +302,11 @@ in {
   config = mkIf cfg.enable (mkMerge [
     {
       home.packages = [ cfg.package ];
+      assertions = [{
+        assertion = !(cfg.delta.enable && cfg.difftastic.enable);
+        message =
+          "Only one of 'programs.git.delta.enable` or `programs.git.difftastic.enable` can be set to true at the same time.";
```
sumnerevans, line 308:
```suggestion
          "Only one of 'programs.git.delta.enable' or 'programs.git.difftastic.enable' can be set to true at the same time.";
```

modules/programs/git.nix
```diff
@@ -239,6 +239,34 @@ in {
...
+          '';
+        };
+
+        background = mkOption {
+          type = types.enum [ "light" "dark" ];
+          default = "light";
+          example = "dark";
+          description = ''
+            Options to configure if difftastic should use the lighter or darker
+            colors for syntax highlithing.
```
sumnerevans, line 256:
```suggestion
            Determines whether difftastic should use the lighter or darker colors
            for syntax highlithing.
```

sumnerevans (changes requested):

lobre:
Hello @notarock,

Is there for convenience that you set the pager in that MR? I don’t see it mentioned in the documentation.

https://github.com/nix-community/home-manager/blob/5872aad1d0f841770db498cba9310d647c4aa376/modules/programs/git.nix#L483

https://difftastic.wilfred.me.uk/git.html

I like to set my own `core.pager` to `less -+FX` to enable mouse scroll. But when I am enabling `difftastic`, it conflicts and I cannot define my own config anymore.

Is it possible to keep that option free for the end user?

notarock:
> would you agree to having it removed, or maybe using mkDefault to avoid forcing it on users?

I don't have an opinion on this. Making it configurable with a default value seems like the way to go, but I don't have time to submit a pull request at this time. Feel free to open one if needed

lobre:
I decided to stick with the default git diff algorithm at the end instead of using difftastic. Making it configurable is the solution, but I will also let that duty of creating a PR to anyone requiring it in the future.

Thanks anyway for helping with that.

Issue #2852, opened 2022-03-30
services.screen-locker: Add option to configure x screensaver cycle when using xss-lock


rszamszur:
### Description

Add an option that will be used with `xset` to configure the X server's screensaver cycle along with timeout when using `xss-lock`.
```bash
$ xset -q
...
Screen Saver:
  prefer blanking:  yes    allow exposures:  yes
  timeout:  300    cycle:  5
$ xset --help
...
   For screen-saver control:
	 s [timeout [cycle]]
```

I've noticed that when using `services.screen-locker` with [`xsecurelock`](https://github.com/google/xsecurelock#automatic-locking) as locker command with `xss-lock` (in order to have support for suspend events) it doesn't lock properly because of high X screensaver cycle default value (600).

Setup with `xss-lock` from xsecurelock docs:
```
xset s 300 5
xss-lock -n /usr/lib/xsecurelock/dimmer -l -- xsecurelock
```

Now  `services.screen-locker.xss-lock` doesn't allow to set both timeout and cycle:
```
xset s 300
# vs
xset s 300 5
```

Now why this is even important in this case? From my tests, it seems that `xsecurelock/dimmer` locks the screen after the first cycle, even though the dimming lasts always the same amount (~3-4sec.). Meaning with the default 600 seconds cycle value `xsecurelock` needs 10minutes + timeout to actually lock the screen.

I know this setup might be quite unorthodox, but it's a simple feature that shouldn't break backward compatibility.

PR #2853, opened 2022-03-30
screen-locker: Add option to configure x screensaver cycle

rszamszur:
### Description

Some screen lockers which will be used with `xss-lock` might require
to set x screensaver timeout with cycle.

Resolves: #2852

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/screen-locker.nix
+++ b/modules/services/screen-locker.nix
@@ -94,6 +94,16 @@ in {
           Extra command-line arguments to pass to <command>xss-lock</command>.
         '';
       };
+
+      screensaverCycle = mkOption {
+        type = types.int;
+        default = 600;
+        description = ''
+          X server's screensaver cycle value expresed as seconds.
+          This will be used with <command>xset</command> to configure
+          the cycle along with timeout.
+        '';
+      };
     };
   };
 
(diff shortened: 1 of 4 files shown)


tests/modules/services/screen-locker/no-xautolock.nix
```diff
@@ -6,7 +6,10 @@
       enable = true;
       inactiveInterval = 5;
       lockCmd = "${pkgs.i3lock}/bin/i3lock -n -c AA0000";
-      xss-lock = { extraOptions = [ "-test" ]; };
+      xss-lock = {
```
rycee, line 9: Trailing whitespace
```suggestion
      xss-lock = {
```

modules/services/screen-locker.nix
```diff
@@ -94,6 +94,15 @@ in {
           Extra command-line arguments to pass to <command>xss-lock</command>.
         '';
       };
+
+      screensaverCycle = mkOption {
+        type = types.int;
+        default = 600;
+        description = ''
+          This will be used with <command>xset</command> to configure
+          the X server's screensaver cycle along with timeout.
```
rycee, line 103: Would probably be good to indicate that the unit is seconds.

rycee:
Thanks for the contribution! I've added a few minor comments.

rycee:
To fix the formatting problem you can run `./format` in the HM project root.

berbiche (approved):
LGTM!

PR #2855, opened 2022-03-31
xdg-mime-apps: add function `mimeAssociations`

rycee:
### Description

This convenience function allows automatic assignment of a package's associations to `xdg.mimeApps.defaultApplications`.

For example,

    xdg.mimeApps.defaultApplications =
      config.lib.xdg.mimeAssociations [ pkgs.gnome.evince ];

Based on code suggestion from @ryantrinkle.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/xdg-mime-apps.nix
+++ b/modules/misc/xdg-mime-apps.nix
@@ -72,20 +72,50 @@ in {
     };
   };
 
-  config = mkIf cfg.enable {
-    assertions =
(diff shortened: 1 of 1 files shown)


ambroisie:
By curiosity, why is it in `config.lib` ? Because it does IFD?

rycee:
@ambroisie Indeed, the function needs to have access to `pkgs` and we can't have that in the plain `lib.hm`.

PR #2859, opened 2022-04-02
Reuse DAG code and refactor it with help of built-in functions

rcerc:
### Description

This modifies `modules/lib/default.nix` to directly re-export `modules/lib/dag.nix` instead of each of the attribute values (e.g. `dagEntryAnywhere` as `entryAnywhere`) and renames the attributes to match the exported names. Some functions, like the built-in `attrValues` and `elem`, as well as the DAG `entryBetween`, are used within `modules/lib/dag.nix` for minor refactoring of the DAG functions. Also, the DAG functions (such as `isDag`) are referenced through `lib.hm.dag` instead of passing the `modules/lib/dag.nix` import to functions in separate files (e.g. `modules/lib/types-dag.nix`) that call them.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/lib/dag.nix
```diff
@@ -51,66 +50,54 @@ in rec {
...
   # Applies a function to each element of the given DAG.
-  dagMap = f: dag: mapAttrs (n: v: v // { data = f n v.data; }) dag;
-
-  # Create a DAG entry with no particular dependency information.
-  dagEntryAnywhere = data: {
-    inherit data;
-    before = [ ];
-    after = [ ];
-  };
+  map = f: dag: mapAttrs (n: v: v // { data = f n v.data; }) dag;
```
polykernel, line 94:
```suggestion
  map = f: mapAttrs (n: v: v // { data = f n v.data; });
```
This has minimal effect but I think eta reduction is still worth doing while we are at refactoring.

modules/lib/dag.nix
```diff
@@ -51,66 +50,54 @@ in rec {
...
-    inherit data after;
-    before = [ ];
-  };
-
-  dagEntryBefore = before: data: {
-    inherit data before;
-    after = [ ];
-  };
+  # Create a DAG entry with no particular dependency information.
+  entryAnywhere = data: hm.dag.entryBetween [ ] [ ] data;
```
polykernel, line 99:
```suggestion
  entryAnywhere = hm.dag.entryBetween [ ] [ ];
```

modules/lib/dag.nix
```diff
@@ -51,66 +50,54 @@ in rec {
...
 
+  entryAfter = after: data: hm.dag.entryBetween [ ] after data;
+  entryBefore = before: data: hm.dag.entryBetween before [ ] data;
```
polykernel, line 102:
```suggestion
  entryAnywhere = hm.dag.entryBetween [ ] [ ];

  entryAfter = hm.dag.entryBetween [ ];
  entryBefore = before: hm.dag.entryBetween before [ ];
```

modules/lib/dag.nix
rcerc: Thank you for catching that!

rycee:
Sorry about the delay, I will make sure to review this next week (week 28).

rycee:
Thanks! Looks good. Merged to master now 🙂

PR #2863, opened 2022-04-05
mpdris2: add password option

eboskma:
### Description

This adds a `password` option to `services.mpdris2`. We have an mpd server running on a separate machine at the office, which is password protected. 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


eboskma:
As a side question, since our mpd server is running on a separate machine, I don't need mpd on my own system. But for some reason, the mpdris2 module has an assertion that demands mpd is also enabled. Is this absolutely necessary?

sumnerevans:
Yes, probably can get rid of that assertion

sumnerevans:
maybe need to rebase on the latest `master` as well to fix the failing tests

eboskma:
Okay, I rebased and removed the assertion.

PR #2875, opened 2022-04-08
wezterm: add module

blmhemu:
### Description

Added a terminal emulator module [wezterm](https://wezfurlong.org/wezterm/)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


.github/CODEOWNERS
```diff
@@ -258,6 +258,9 @@
 /modules/programs/waybar.nix                          @berbiche
 /tests/modules/programs/waybar                        @berbiche
 
+/modules/programs/wezterm.nix                         @blmhemu
+/tests/modules/programs/wezterm                       @blmhemu
```
SuperSandro2000, line 262: As GitHub displays in the warning: This entry does nothing.

modules/programs/wezterm.nix
```diff
@@ -0,0 +1,58 @@
...
+            hide_tab_bar_if_only_one_tab = true,
+            default_prog = { "zsh", "--login", "-c", "tmux attach -t dev || tmux new -s dev" },
+            keys = {
+              {key="n", mods="SHIFT|CTRL", action="ToggleFullScreen"},
+            }
+          }
+        '';
+        description = ''
+          Configuration written to
+          <filename>$XDG_CONFIG_HOME/wezterm/wezterm.lua</filename>. See
```
SuperSandro2000, line 36: This breaks config reloading. Please use the workaround mentioned in https://github.com/wez/wezterm/issues/1697

.github/CODEOWNERS
blmhemu: Hey !  Could you pls point me how to solve this. It seems some other entries are being flagged as well.

sumnerevans: What you have is fine. It won't automatically notify you or request you to review, though since you aren't a member of the repository.

Eventually we should probably figure out a better system, but for now it's fine.

modules/programs/wezterm.nix
blmhemu: Fixed and tested the reloading works.

luxus:
is there a reason not to merge it?

rycee:
Thanks! Rebased to master in d9e03b7f8c67890b920c92d8154dd9b60bb1242f.

PR #2877, opened 2022-04-08
neomutt/signature: unset if disabled

### LINKED ISSUE #2876: bug: neomutt doesn't reset the signature when changing accounts

### Issue description

I've multiple mail accounts configured, and one of them has no signature. If I change to this account, the signature of the previous account is set. Please add `unset signature` in the account hook, if no signature is specified for the account.

### Maintainer CC

_No response_

### System information

```markdown
error (ignored): error: end of string reached
error: getting status of '/var/src/config': No such file or directory
 - system: `0`
 - host os: `Linux 5.10.109, NixOS, 21.11 (Porcupine)`
 - multi-user?: `no`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.5.0pre20211206_d1aaa7e`
 - channels(toon): `""`
 - channels(root): `""`
 - nixpkgs: `/nix/store/9dmig1pv9njj5kswvs8yvw3qp6b81zkd-source`
```

sumnerevans:
### Description

Closes #2876

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/neomutt.nix
+++ b/modules/programs/neomutt.nix
@@ -223,9 +223,11 @@ let
 
       # Extra configuration
       ${account.neomutt.extraConfig}
-    '' + optionalString (account.signature.showSignature != "none") ''
+    '' + (if (account.signature.showSignature == "none") then ''
+      unset signature
+    '' else ''
       set signature = ${pkgs.writeText "signature.txt" account.signature.text}
-    '' + optionalString account.notmuch.enable (notmuchSection account);
+    '') + optionalString account.notmuch.enable (notmuchSection account);
 
 in {
   options = {
--- a/tests/modules/programs/neomutt/hm-example.com-expected
+++ b/tests/modules/programs/neomutt/hm-example.com-expected
@@ -31,6 +31,7 @@ set trash='+Trash'
 # Extra configuration
 color status cyan default
 
+unset signature
 # notmuch section
 set nm_default_uri = "notmuch:///home/hm-user/Mail"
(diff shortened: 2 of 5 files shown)


modules/programs/neomutt.nix
```diff
@@ -223,9 +223,11 @@ let
 
       # Extra configuration
       ${account.neomutt.extraConfig}
-    '' + optionalString (account.signature.showSignature != "none") ''
```
rycee, line 226: An alternative may be to do something like
``` nix
setOption "signature" (
  if account.signature.showSignature == "none" then
    null
  else
    "${pkgs.writeText "signature.txt" account.signature.text}"
)
```
although I don't know if it is particularly better.

modules/programs/neomutt.nix
```diff
@@ -223,9 +223,11 @@ let
...
+    '' + (if (account.signature.showSignature == "none") then ''
```
rycee, line 226: If you prefer using this code then can remove the unnecessary parentheses:
```suggestion
    '' + (if account.signature.showSignature == "none" then ''
```

rycee:
Thanks! Looks good to me, just had a minor comment.

berbiche (approved):

modules/programs/neomutt.nix
sumnerevans: I think I'm going to keep it as-is. I don't think that is any better than what I have.

teto:
I dont use neomutt but the problem may arise for other settings too, is there a way to reset settings between accounts, i.e., to start from a clean slate ?

PR #2880, opened 2022-04-09
keychain: set SHELL during initialization

### LINKED ISSUE #2256: bug: keychain can print incorrect initialisation code for current shell

### Issue description

If you:

- have keychain enabled (`programs.keychain.enable`),
- are using fish as your login shell, and
- try to run bash from within fish,

bash will exit with:

```
bash: set: -U: invalid option
set: usage: set [-abefhkmnptuvxBCHP] [-o option-name] [--] [arg ...]
```

I suspect this also happens if you use zsh as your login shell, though I haven't tested this.

This is because bash doesn't change $SHELL if it's already set (and it is, by fish); keychain apparently has no option to always print e.g. the bash initialisation code, so it looks at $SHELL, and guesses wrongly.

### Maintainer CC

@marsam

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 5.10.52, NixOS, 21.05.1970.11c662074e2 (Okapi)`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.3.15`
 - channels(ash): `"nixos-unstable-21.11pre304626.8ecc61c91a5, home-manager-21.05"`
 - channels(root): `"nixos-21.05.1970.11c662074e2"`
```

Infinidoge:
### Description

Explicitly sets the `SHELL` environment variable, which should force Keychain to use the correct initialization code for the respective shells.

Tested with the integration enabled, and entering the `fish` shell with an incorrect environment variable (`SHELL=/run/current-system/sw/bin/zsh`). Previously gave a warning, which disappeared with the fix.

Fixes #2256

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/keychain.nix
+++ b/modules/programs/keychain.nix
@@ -100,13 +100,13 @@ in {
   config = mkIf cfg.enable {
     home.packages = [ cfg.package ];
     programs.bash.initExtra = mkIf cfg.enableBashIntegration ''
-      eval "$(${shellCommand})"
+      SHELL=bash eval "$(${shellCommand})"
     '';
     programs.fish.interactiveShellInit = mkIf cfg.enableFishIntegration ''
-      eval (${shellCommand})
+      SHELL=fish eval (${shellCommand})
     '';
     programs.zsh.initExtra = mkIf cfg.enableZshIntegration ''
-      eval "$(${shellCommand})"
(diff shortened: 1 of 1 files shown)


berbiche (approved):
Thanks!

PR #2882, opened 2022-04-09
git: Add option to use diff-so-fancy as diff tool

arjan-s:
### Description
Add the option to use [diff-so-fancy](https://github.com/so-fancy/diff-so-fancy) as diff tool in git. This makes the shown diffs human readable (and pretty) instead of machine readable.
This PR was inspired by https://github.com/nix-community/home-manager/pull/2850

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/git.nix
+++ b/modules/programs/git.nix
@@ -296,6 +296,65 @@ in {
           '';
         };
       };
+
+      diff-so-fancy = {
+        enable = mkEnableOption "" // {
+          description = ''
+            Enable the <command>diff-so-fancy</command> diff colorizer.
+            See <link xlink:href="https://github.com/so-fancy/diff-so-fancy" />.
+          '';
+        };
+
+        markEmptyLines = mkOption {
+          type = types.bool;
+          default = true;
+          example = false;
+          description = ''
+            Whether the first block of an empty line should be colored.
+          '';
+        };
+
+        changeHunkIndicators = mkOption {
+          type = types.bool;
+          default = true;
+          example = false;
+          description = ''
+            Simplify git header chunks to a more human readable format.
+          '';
+        };
+
+        stripLeadingSymbols = mkOption {
+          type = types.bool;
+          default = true;
+          example = false;
+          description = ''
+            Whether the <literal>+</literal> or <literal>-</literal> at
+            line-start should be removed.
+          '';
+        };
+
+        useUnicodeRuler = mkOption {
+          type = types.bool;
+          default = true;
+          example = false;
+          description = ''
+            By default, the separator for the file header uses Unicode
+            line-drawing characters. If this is causing output errors on
+            your terminal, set this to false to use ASCII characters instead.
+          '';
+        };
+
+        rulerWidth = mkOption {
+          type = types.nullOr types.int;
+          default = null;
+          example = false;
+          description = ''
+            By default, the separator for the file header spans the full
+            width of the terminal. Use this setting to set the width of
+            the file header manually.
+          '';
+        };
+      };
     };
   };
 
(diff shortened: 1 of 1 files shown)


modules/programs/git.nix
```diff
@@ -296,16 +296,71 @@ in {
...
   };
 
   config = mkIf cfg.enable (mkMerge [
     {
       home.packages = [ cfg.package ];
       assertions = [{
-        assertion = !(cfg.delta.enable && cfg.difftastic.enable);
+        assertion = !(cfg.delta.enable && cfg.difftastic.enable)
+          && !(cfg.delta.enable && cfg.diff-so-fancy.enable)
+          && !(cfg.difftastic.enable && cfg.diff-so-fancy.enable);
```
ambroisie, line 361: You might just want to put them in a list, and count that at most one is enabled.

arjan-s: @ambroisie Thanks for your review! I actually wanted to implement it that way, but couldn't figure out how to do it in Nix... Can you help with that?

olmokramer: @arjan-s You could do this for example

```
assertion = let
  enabled = [
    cfg.delta.enable
    cfg.diff-so-fancy.enable
    cfg.difftastic.enable
  ];
in lib.length (lib.filter lib.id enabled) == 1;
```

ambroisie (approved):

modules/programs/git.nix
```diff
@@ -296,16 +296,75 @@ in {
...
+          description = ''
+            Simplify git header chunks to a more human readable format.
+          '';
+        };
+        stripLeadingSymbols = mkOption {
+          type = types.bool;
+          default = true;
+          example = false;
+          description = ''
+            Should the pesky + or - at line-start be removed.
```
sumnerevans, line 328:
```suggestion
            Should the pesky <literal>+</literal> or <literal>-</literal> at line-start be removed?
```

modules/programs/git.nix
```diff
@@ -296,16 +296,75 @@ in {
...
+            Enable the <command>diff-so-fancy</command> diff colorizer.
+            See <link xlink:href="https://github.com/so-fancy/diff-so-fancy" />.
+          '';
+        };
+        markEmptyLines = mkOption {
+          type = types.bool;
+          default = true;
+          example = false;
+          description = ''
+            Should the first block of an empty line be colored.
```
sumnerevans, line 312:
```suggestion
            Should the first block of an empty line be colored?
```

sumnerevans (changes requested):

modules/programs/git.nix
rycee: I think it feels a bit odd for the documentation to ask the user questions. I would suggest to instead use the "Whether …" kind of phrasing. E.g.

> Whether the first block of an empty line should be colored.

modules/programs/git.nix
```diff
@@ -296,16 +296,75 @@ in {
...
       };
+
+      diff-so-fancy = {
+        enable = mkEnableOption "" // {
+          description = ''
+            Enable the <command>diff-so-fancy</command> diff colorizer.
+            See <link xlink:href="https://github.com/so-fancy/diff-so-fancy" />.
+          '';
+        };
+        markEmptyLines = mkOption {
```
rycee, line 307: Please make sure to separate option definitions by an empty line to make the module more readable. That is,
```suggestion

        markEmptyLines = mkOption {
```

modules/programs/git.nix
```diff
@@ -296,16 +296,75 @@ in {
...
       home.packages = [ cfg.package ];
       assertions = [{
-        assertion = !(cfg.delta.enable && cfg.difftastic.enable);
+        assertion = let
+          enabled = [
+            cfg.delta.enable
+            cfg.diff-so-fancy.enable
+            cfg.difftastic.enable
+          ];
+        in lib.length (lib.filter lib.id enabled) <= 1;
```
rycee, line 365:
```suggestion
        in count id enabled <= 1;
```

modules/programs/git.nix
arjan-s: > I think it feels a bit odd for the documentation to ask the user questions. I would suggest to instead use the "Whether …" kind of phrasing. E.g.
> 
> > Whether the first block of an empty line should be colored.

Yeah, I agree. I first opted to use the option descriptions from the original readme, but this seems nicer.

PR #2887, opened 2022-04-11
picom: remove deprecated option, add tests

thiagokokada:
### Description

Remove `services.picom.refreshRate`. This option was deprecated since release [v9](https://github.com/yshui/picom/releases/tag/v9), and it is generating the following warn in the log:

```
[ DD/MM/YYYY HH:MM:SS.mmm parse_config_libconfig WARN ] The refresh-rate option has been deprecated. Please remove it from your configuration file. If you encounter any problems without this feature, please feel free to open a bug report
```

To ensure that I didn't break anything, I decided to add a test to the module (since it didn't had before). To make testing easier, I did a small refactoring in the module, that includes adding the file to the `~/.config/picom/picom.conf` path that is the expected one for the application. This change should also make it easier to debug errors in the config (no more getting the current generated file from the service file).

Also removed an old workaround for MESA 18. Not tested, but I don't think it is needed anymore.

### Checklist

  + Mostly. If you use `services.picom.refreshRate` your config will break, but this should show the proper error thanks to the usage of `mkRemovedOptionModule`.

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/picom.nix
+++ b/modules/services/picom.nix
@@ -1,13 +1,14 @@
 { config, lib, pkgs, ... }:
 
-with lib;
-with builtins;
-
 let
+  inherit (builtins) toJSON toString;
+  inherit (lib)
+    concatStringsSep elemAt literalExpression mkEnableOption mkIf mkOption
+    mkRemovedOptionModule optional optionalAttrs optionalString types;
 
   cfg = config.services.picom;
 
-  configFile = pkgs.writeText "picom.conf" (optionalString cfg.fade ''
+  configFile = optionalString cfg.fade ''
     # fading
     fading = true;
     fade-delta    = ${toString cfg.fadeDelta};
@@ -46,8 +47,7 @@ let
     # other options
     backend = ${toJSON cfg.backend};
     vsync = ${toJSON cfg.vSync};
-    refresh-rate = ${toString cfg.refreshRate};
-  '' + cfg.extraOptions);
+  '' + cfg.extraOptions;
 
 in {
 
@@ -250,15 +250,6 @@ in {
       '';
     };
 
-    refreshRate = mkOption {
-      type = types.int;
-      default = 0;
-      example = 60;
-      description = ''
-        Screen refresh rate (0 = automatically detect).
-      '';
-    };
-
     package = mkOption {
       type = types.package;
       default = pkgs.picom;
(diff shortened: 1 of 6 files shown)


modules/services/picom.nix
```diff
@@ -301,15 +299,13 @@ in {
 
       Service = let
         experimentalBackendsFlag =
-          if cfg.experimentalBackends then " --experimental-backends" else "";
+          lib.optionalString cfg.experimentalBackends "--experimental-backends";
       in {
-        ExecStart = "${cfg.package}/bin/picom --config ${configFile}"
-          + experimentalBackendsFlag;
+        ExecStart = "${cfg.package}/bin/picom --config ${
+            config.xdg.configHome + "/picom/picom.conf"
```
thiagokokada, line 305: I tried to use `xdg.configFile."picom/picom.conf".target` instead, however this generated `.config/picom/picom.conf` instead of `/home/hm-user/.config/picom/picom.conf`.

If there is a better solution I am all ears.

thiagokokada:
Tested with my config, works fine: https://github.com/thiagokokada/nix-configs/commit/f0d30337d7c5c3f42573156ea04b1275fd0bfc90

thiagokokada:

modules/services/picom.nix
berbiche: There's `xdg.configFile.*.source` that points to the resulting file in the Nix store.

AFAIK there will only be a service restart if the settings are changed.

thiagokokada: > There's xdg.configFile.*.source that points to the resulting file in the Nix store.

If possible I would much prefer the way it currently is: pointing to the file inside the HOME. This way it is much easier to debug (e.g.: `systemctl cat --user picom.service` will show that the file is on `~/.config/picom/picom.conf`, so the user can open the file and see the contents).

> AFAIK there will only be a service restart if the settings are changed.

Not sure if I understood this comment here.

rycee: I think in general it is best to point directly to the configuration file in the Nix store. I.e., `${config.xdg.configFile."picom/picom.conf".source}` in this case as @berbiche mentioned.

The `systemctl cat --user picom.service` should then contain the path to the configuration file in the Nix store so it should still be reasonably easy to find the file.

As mentioned we then also get the benefit that the systemd service will be automatically restarted (assuming `systemd.user.startServices` is enabled) when the configuration is changed. The tool that does the restarts essentially compares the old and new service files and does the restart if they differ.

rycee: Btw, if you do switch to point directly to the configuration in the Nix store then you can still have stable tests using the `normalizeStorePaths` function. See

https://github.com/nix-community/home-manager/blob/7da4e6680f814562f070bad01922d6736e54be94/tests/modules/services/mpd/basic-configuration.nix#L11-L12

for an example

thiagokokada: Done.

thiagokokada:
Can you review again? CC @berbiche @rycee

rycee:
Nice! Merged to master now.

PR #2894, opened 2022-04-15
aerc: add module

wiIIiambonner:
### Description

Added a module for the `aerc` email client.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


podocarp:
Just a small thing I noticed, you added yourself as maintainer but did not specify `meta.maintainers` in the module

ratsclub:
I tried to use the default configuration files when the user doesn't set it's own configurations here but I could not proceed due to this error:

```
error: A definition for option `xdg.configFile.aerc/aerc.conf.source' is not of type `path'. Definition values:
       - In `/home/victor/Projects/home-manager/modules/programs/aerc.nix': "@aerc@/share/aerc/aerc.conf"
```

This is the relevant piece of code used before:

```nix
xdg.configFile."aerc/aerc.conf".source = if cfg.conf != { } then
    formatIni.generate "aerc.conf" cfg.conf
  else
    "${pkgs.aerc}/share/aerc/aerc.conf";
```

modules/programs/aerc.nix
```diff
@@ -0,0 +1,131 @@
...
+    mkKeyValue = k: v:
+      let value = (if isBool v then boolToString else toString) v;
+      in "${k} = ${value}";
+  };
+
+  configDir =
+    if pkgs.stdenv.isDarwin then
+      "Library/Application Support/aerc"
+    else
+      "${config.xdg.configHome}/aerc";
```
polykernel, line 18:
```suggestion
  configDir =
    if pkgs.stdenv.isDarwin then
      "Library/Application Support"
    else
      config.xdg.configHome;
```

modules/programs/aerc.nix
```diff
@@ -0,0 +1,131 @@
...
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+    meta.maintainers = [ maintainers.ratsclub ];
+
+    home.packages = [ cfg.package ];
+
+    xdg.configFile = {
```
polykernel, line 114:
```suggestion
    home.file.${configDir} = {
```

tests/modules/programs/aerc/with-config-file.nix
```diff
@@ -0,0 +1,68 @@
...
+        ''
+      }
+
+      assertFileContent home-files/.config/aerc/templates/foo ${
+        builtins.toFile "foo" ''
+          bar
+        ''
+      }
+    '';
+  };
```
polykernel, line 67:
```suggestion
    nmt.script = let
      configDir = if pkgs.stdenv.isDarwin then
          "home-files/Library/Application Support"
        else
          "home-files/.config";
      in ''
        assertFileExists ${configDir}/aerc/aerc.conf
        assertFileExists ${configDir}/aerc/binds.conf
        assertFileExists ${configDir}/aerc/stylesets/default
        assertFileExists ${configDir}/aerc/templates/foo
  
        assertFileContent ${configDir}/aerc/aerc.conf ${
          builtins.toFile "aerc.conf" ''
            [ui]
            mouse-enabled=true
            sort=-r date
          ''
        }
  
        assertFileContent ${configDir}/aerc/binds.conf ${
          builtins.toFile "binds.conf" ''
            [terminal]
            $ex=<C-x>
            $noinherit=true
          ''
        }
  
        assertFileContent ${configDir}/aerc/stylesets/default ${
          builtins.toFile "default" ''
            *.default = true
            *.selected.reverse = toggle
          ''
        }
  
        assertFileContent ${configDir}/aerc/templates/foo ${
          builtins.toFile "foo" ''
            bar
          ''
        }
      '';
  };
```

polykernel:
> I tried to use the default configuration files when the user doesn't set it's own configurations here but I could not proceed due to this error:
> 
> ```
> error: A definition for option `xdg.configFile.aerc/aerc.conf.source' is not of type `path'. Definition values:
>        - In `/home/victor/Projects/home-manager/modules/programs/aerc.nix': "@aerc@/share/aerc/aerc.conf"
> (use '--show-trace' to show detailed location information)
> ```
> 
> This is the relevant piece of code used before:
> 
> ```nix
> xdg.configFile."aerc/aerc.conf".source = if cfg.conf != { } then
>     formatIni.generate "aerc.conf" cfg.conf
>   else
>     "${pkgs.aerc}/share/aerc/aerc.conf";
> ```

I believe this is caused by the following line:
```nix
test.stubs.aerc = { };
```

Defining `tests.stubs.<package>` will override the package globally because it internally uses an overlay. You can avoid this by only supplying the stub package to the package option.
```nix
package = config.lib.test.mkStubPackage { };
```

<br>

I think the default should be sourced from `cfg.package` to not make the configuration depend on the nixpkgs version but I don't think it matters too much here given it is only a fallback.

ratsclub:
Thanks for the review!

> I think the default should be sourced from `cfg.package` to not make the configuration depend on the nixpkgs version but I don't think it matters too much here given it is only a fallback.

This makes me think that I shouldn't provide this default file as a simple modification of the package would break it...

modules/programs/aerc.nix
ratsclub: Done.

modules/programs/aerc.nix
ratsclub: Done.

tests/modules/programs/aerc/with-config-file.nix
ratsclub: Done.

teto:
do you plan to support hte generation of mailboxes via the mail module ?

ratsclub:
I can definitely try to do it so!

ratsclub:
@teto would it be able to merge the module as is and then I provide the account setup in another PR? I guess that it will be a lot more involved as it might support imap, notmuch and maildir and make the PR bigger to review.

ratsclub:
Unfortunately, the [config/config.go](https://git.sr.ht/~rjarry/aerc/tree/fb0e9e3e41cd263de139de0bf620cc69ee5d61ae/item/config/config.go#L853) file makes it impossible to store `accounts.conf` on the nix store...

teto:
> @teto would we be able to merge the module as is and then I provide the account setup in another PR? I guess that it will be a lot more involved as it might support imap, notmuch and maildir and make the PR bigger to review.

I understand. Yeah no pb. Other people have already reviewed the PR so I'll let them merge.

> Unfortunately, the [config/config.go](https://git.sr.ht/~rjarry/aerc/tree/fb0e9e3e41cd263de139de0bf620cc69ee5d61ae/item/config/config.go#L853) file makes it impossible to store accounts.conf on the nix store...

aerc should support a way that doesn't embed passwords in its config. I dont know if it's availaible or not but an aerc user may be able to negotiate a solution upstream.

ratsclub:
> aerc should support a way that doesn't embed passwords in its config. I dont know if it's availaible or not but an aerc user may be able to negotiate a solution upstream.

The only way I know of is reading the emails from `maildir` or `notmuch` and sending them through `msmtp`. All the other ways require an `accounts.conf` with passwords on it.

polykernel:
> This makes me think that I shouldn't provide this default file as a simple modification of the package would break it...

I think in general it is a good idea to keep the configuration and the package separate as much as possible to allow better composition but sometimes it is inevitable. I am not an aerc user myself but I trust you to make the right decision.

ratsclub:
Thanks for the trust!

I feel that we are ready to merge as of now. I don't intend to implement the account setting right now as this will involve an upstream effort. May we get this approved?

ratsclub:
I sent a patch upstream about the accounts too: https://lists.sr.ht/~rjarry/aerc-devel/patches/31289

ambroisie:
@wiIIiambonner there is a way to add a command for passwords, rather than writing the password itself in the configuration file (see [here](https://git.sr.ht/~rjarry/aerc/commit/b8208509f47301bcb03bc222f5e5765833478250)).

Is there a reason you closed this MR?

teto:
the new MR I suppose https://github.com/nix-community/home-manager/pull/3070

PR #2899, opened 2022-04-18
neomutt/signature: add support for command

sumnerevans:
### Description

Neomutt will run the given command and take the output from stdout and
use it as the signature for your email

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/accounts/email.nix
```diff
@@ -44,6 +44,13 @@ let
         '';
       };
 
+      command = mkOption {
+        type = with types; nullOr str;
```
berbiche, line 48:
```suggestion
        type = with types; nullOr (either str path);
```
This would allow using something like `pkgs.writeShellScript`

modules/programs/neomutt.nix
```diff
@@ -196,7 +196,16 @@ let
 
   accountStr = account:
     with account;
-    ''
+    let
+      signature = (if (account.signature.showSignature == "none") then
+        "unset signature"
+      else if account.signature.command != null then
+        ''set signature = "${account.signature.command}|"''
```
berbiche, line 203:
```suggestion
        ''set signature = "${toString account.signature.command}|"''
```
If my prior suggestion is applied, it becomes necessary.

tests/modules/programs/neomutt/neomutt-with-signature.nix
```diff
@@ -0,0 +1,47 @@
...
+
+    nmt.script = ''
+      assertFileExists home-files/.config/neomutt/neomuttrc
+      assertFileExists home-files/.config/neomutt/hm@example.com
+      assertFileContent home-files/.config/neomutt/neomuttrc ${
+        ./neomutt-expected.conf
+      }
+      assertFileContent home-files/.config/neomutt/hm@example.com ${
+        ./hm-example.com-signature-expected
+      }
```
berbiche, line 44:
```suggestion
      expectedSignature=$(normalizeStorePaths "home-files/.config/neomutt/hm@example.com")
      assertFileContent "$expectedSignature" ${
        ./hm-example.com-signature-expected
      }
```

The store path has to be changed to zeros with `normalizeStorePaths`.

sumnerevans: Thanks, I hadn't gotten around to figuring out how to do that.

modules/programs/neomutt.nix
```diff
@@ -196,7 +196,16 @@ let
 
   accountStr = account:
     with account;
-    ''
+    let
+      signature = if account.signature.showSignature == "none" then
+        "unset signature"
+      else if account.signature.command != null then
+        ''set signature = "${toString account.signature.command}|"''
```
rycee, line 203: Are you sure that you want `toString` here? If you set `command = ./signature.sh` (i.e., a path literal) shouldn't this command point to the file in the Nix store?

modules/accounts/email.nix
```diff
@@ -44,6 +44,13 @@ let
         '';
       };
 
+      command = mkOption {
+        type = with types; nullOr (either str path);
+        default = null;
+        example = "./signature.sh";
```
rycee, line 50: When giving a relative path here, what is it relative? The `~/.config/neomutt` directory? Would it be possible to have the type be just `path` to encourage people to put their signature script in the Nix store?

rycee:
Sorry, I force pushed the branch with a few minor cleanups before noticing the things I commented on above.

PR #2902, opened 2022-04-18
targets.genericLinux: include additional directories in XCURSOR_PATH

polykernel:
This commit appends system-wide icon and pixmap directory and the icon
directory in the home-manager profile to the XCURSOR_PATH session variable
for the generic linux target. This is necessary because the default prefix
for libXcursor resolves to the Nix store which excludes the aforementioned
directories from being searched for cursor themes. [1]

[1] - https://github.com/nix-community/home-manager/pull/2891#issuecomment-1101064521.

### Description

cc: @EHfive

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/targets/generic-linux.nix
+++ b/modules/targets/generic-linux.nix
@@ -43,6 +43,17 @@ in {
       "/var/lib/snapd/desktop"
     ];
 
+    # We need to append system-wide FHS directories due to the default prefix
+    # resolving to the Nix store.
+    # https://github.com/nix-community/home-manager/pull/2891#issuecomment-1101064521
+    home.sessionVariables = {
+      XCURSOR_PATH = "$XCURSOR_PATH\${XCURSOR_PATH:+:}" + concatStringsSep ":" [
+        "${config.home.profileDirectory}/share/icons"
+        "/usr/share/icons"
+        "/usr/share/pixmaps"
+      ];
+    };
+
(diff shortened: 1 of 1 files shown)


modules/targets/generic-linux.nix
```diff
@@ -43,6 +43,18 @@ in {
       "/var/lib/snapd/desktop"
     ];
 
+    # We need to append system-wide FHS directories due to the default prefix
+    # resolving to the Nix store.
+    # https://github.com/nix-community/home-manager/pull/2891#issuecomment-1101064521
+    home.sessionVariables = {
+      XCURSOR_PATH = concatStringsSep ":" [
+        "\${XCURSOR_PATH}"
```
berbiche, line 51: I think we need to use `sessionVariablesExtra` if we want to extend an environment variable.

berbiche, line 51:
```suggestion
      XCURSOR_PATH =
        "$XCURSOR_PATH\${XCURSOR_PATH:+:}"
        + concatStringsSep ":" [
```
This will need to be formatted.

polykernel: I forgot to consider the case where `XCURSOR_PATH` is null or unset. Thanks!

berbiche (approved):

PR #2905, opened 2022-04-21
mbsync: add darwin support via launchd

ryantking:
### Description

When the ~mbsync~ service is enabled, it will now create a ~launchd~ unit on Darwin systems. The ~systemd~ functionality is unchanged.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/services/mbsync.nix
```diff
@@ -75,35 +85,47 @@ in {
...
-        ExecStart =
-          "${cfg.package}/bin/mbsync ${concatStringsSep " " mbsyncOptions}";
-      } // (optionalAttrs (cfg.postExec != null) {
-        ExecStartPost = cfg.postExec;
-      }) // (optionalAttrs (cfg.preExec != null) {
-        ExecStartPre = cfg.preExec;
-      });
-    };
+  config = mkIf cfg.enable (mkMerge [
+    (optionalAttrs (hasAttr "sytemd" options) {
```
ambroisie, line 89: Probably better to use `pkgs.stdenv.hostPlatform.system` like `lib.hm.assertions.assertPlatform` does it.

modules/services/mbsync.nix
```diff
@@ -54,6 +55,15 @@ in {
       '';
     };
 
+    logDirectory = mkOption {
```
rycee, line 58: Isn't there some standard directory for logs? In XDG it seems [XDG_STATE_HOME](https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html) is intended for this.

So, unless some other directory is preferred under macos, I would suggest omitting the `logDirectory` option and always use `${config.xdg.stateHome}/<app name>` as the log directory.

ambroisie: I *think* using a log directory is an anti-pattern when using `systemd`.

rycee: @ambroisie Yeah, specifying a log directory for a systemd service would not be sensible. And I'm not very fond of the idea of adding a log directory option, I would prefer simply deciding on a base path where they all go, the user could always override the path by changing the value of XDG_STATE_HOME.

ryantking: Hmmm on second thought, I'm not sure macOS actually uses `XDG_STATE_HOME`. I believe launchd logs internally by default and you use the CLI to access them. There's not a lot of documentation that I can find. I think the two options here are:

1. Remove the output path option entirely, let users do `launchd.agents.mbsync.config.StandardOutPath = myPath` if they so choose.
2. If the user explicitly sets `XDG_STATE_HOME`, then use it. Probably want to use `mkDefault` so they can override it.

I think 2 in practice would work the best, but I'm hesitant to use an XDG variable that technically macOS doesn't actually make use of.

ryantking: CC: @rycee @ambroisie

ambroisie: I think that even though it does not use the XDG standard, it does have a standard location which has an analogous usage.

I do not have have any strong feelings regarding its use or not, I was simply stating that (IIRC) using a log directory goes somewhat against what is idiomatic whenever using systemd.

ryantking: Okay so after some further reading, it looks like both systemd and launchd log to some internal mechanism that can be redirected to files using `StandardOutput` and `ErrorOutput` keys in the unit files. So I will just completely remove those keys and leave it up to the user to specify, then we should have basically identical functionality on both the launchd and systemd units.

modules/services/mbsync.nix
ryantking: So is something like `(optionalAttrs (elem pkgs.stdenv.hostPlatform lib.platforms.linux) {` what you mean? Because that doesn't compile due to infinite recursion. I think it's because you can't check the platform on the top level, which is why I used the `hasAttr` pattern originally.

modules/services/mbsync.nix
```diff
@@ -75,35 +85,47 @@ in {
...
+    })
+    (optionalAttrs (hasAttr "launchd" options) {
+      launchd.agents.mbsync = {
+        enable = true;
+        config = ({
+          ProgramArguments = [ "${cfg.package}/bin/mbsync" ] ++ mbsyncOptions;
+          ProcessType = "Adaptive";
+          RunAtLoad = true;
+          StartInterval = 60
+            * 15; # TODO(ryantking): Use frequency here... somehow
```
ryantking, line 123: @ambroisie @rycee Any thoughts on how to handle the frequency? Launchd does support onCalendar, but you have to provide it in a dictionary, `<hour>00</hour><minute>00</minute>`, not just as a string like systems. I was playing around with writing a parser to parse from the string format to the dict for launchd, but that is questionable with the lackluster regex support that nix has, e.g., can't even do non-capturing groups.

What would you think about accepting a `or (str attrs)` as the type for frequency and instructing users which to use based on their system?

rycee: Yeah, that is a bit of a conundrum. Once upon a time I made [a little library](https://github.com/NixOS/nixpkgs/pull/43852) to handle durations in Nix. So in principle it should be possible to handle, at least simple cases, in a generic way. It would be necessary to keep an escape hatch for people who'd like to use something like "Thu,Fri 2012-*-1,5 11:12:13". I will think about that and perhaps resurrect the library.

For now I imagine it would be simplest to update the frequency documentation and make the user enter `frequency = "<hour>00</hour><minute>00</minute>"`. Not very pretty but would probably work until we have something better.

Do you have any link to documentation about the interval expressions supported by launchd?

modules/services/mbsync.nix
rycee: The trick is to almost never use `optionalAttrs` when dealing with NixOS modules. It is a sure-fire way to end up in infinite recursion. You'll want to use

    mkIf pkgs.stdenv.hostPlatform.isDarwin { … }

instead. For example

https://github.com/nix-community/home-manager/blob/684e85d01d333be91c4875baebb05b93c7d2ffaa/modules/targets/darwin/fonts.nix#L15

PR #2906, opened 2022-04-21
most: create enbable and pager options

j0hax:
### Description

Allow for setting Most as the default pager.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/most.nix
```diff
@@ -0,0 +1,22 @@
...
+
+let
+
+  cfg = config.programs.most;
+
+in {
+  options.programs.most = {
+    enable = mkEnableOption "Most, a powerful paging program";
+    sessionPager =
+      mkEnableOption "use Most as the session pager, e.g. for manpages";
```
rycee, line 13: I think this option could be removed. Presumably, if you enable the module you would always want Most as pager?

modules/programs/most.nix
```diff
@@ -0,0 +1,22 @@
...
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ pkgs.most ];
+    home.sessionVariables = mkIf cfg.sessionPager { PAGER = "most"; };
```
rycee, line 18: Could use an absolute path

```suggestion
    home.sessionVariables = mkIf cfg.sessionPager { PAGER = "${pkgs.most}/bin/most"; };
```

(or

```suggestion
    home.sessionVariables.PAGER = "most";
```

if the above suggestion is accepted).

PR #2908, opened 2022-04-24
firefox: add option to enable arkenfox userjs

MoritzBoehme:
Now Arkenfox user.js template can be added to the top of a profiles
user.js file and overrides through profile.settings or profile.extraConfig.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


MoritzBoehme:
I was having some issues as I tried to run the test:
```
error: attribute 'shellDryRun' missing

       at /home/moritz/home-manager/modules/programs/bash.nix:12:9:

           11|     checkPhase = ''
           12|       ${pkgs.stdenv.shellDryRun} "$target"
             |         ^
           13|     '';
```
I am a bit confused why this is, as this was added [quite a while ago](https://github.com/NixOS/nixpkgs/commit/f2065d81ad9d25bbe581aef5c4c80845d7bb3041)...

dwarfmaster:
It's a bit of a self-insert, but I also wrote an arkenfox module for home-manager [here](https://arkenfox.dwarfmaster.net/). It uses a script to extract all the settings from the `user.js` file and generate a nix file with the description. From that, it generates a documentation ([example](https://arkenfox.dwarfmaster.net/)) and a home-manager module (using import from derivation), that permits enabling/disabling settings per section, subsection and setting. If there is interest in such features, I could write a better documentation and prepare a pull request.

MoritzBoehme:
I just had a look at your module, and it looks way better structured and thought out than my implementation. I think it is the better option if this kind of feature should be implemented.

PR #2911, opened 2022-04-24
darwin: add support for 'defaults -currentHost' options

### LINKED ISSUE #2615: Darwin: Ability to set -currentHost defaults
### Description

I don't know when this became common but, in newer macOS versions, some preferences go to `~/Library/Preferences/ByHost/$filename.$UUID.plist` (i.e. `~/Library/Preferences/ByHost/com.apple.controlcenter.00000000-0000-0000-0000-00000000.plist`) now.

These can be accessed using `defaults -currentHost`, i.e. `defaults read -currentHost com.apple.controlcenter`. This is the place `BatteryShowPercentage` lives in macOS 11 for instance.

As a HM user, I want to be able to set these defaults just like the regular ones.

@midchildan

midchildan:
### Description
This adds a new option, `targets.darwin.currentHostDefaults`, which exposes additional macOS preferences. It is similar to `targets.darwin.defaults`, but it contains options that are only applicable for the current host instead of all hosts the logged in user has accounts on.

Closes #2615.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -605,6 +605,17 @@ in
           A new module is available: 'programs.librewolf'.
         '';
       }
+
+      {
+        time = "2022-07-24T13:17:01+00:00";
+        condition = hostPlatform.isDarwin;
+        message = ''
+          A new option is available: 'targets.darwin.currentHostDefaults'.
+
+          This exposes macOS preferences that are available through the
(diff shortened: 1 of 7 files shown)


rycee:
Thanks! Merged to master now.

PR #2916, opened 2022-04-25
imapnotify: Do not shell-escape password cmd

kirelagin:
### Description

Fix a regression introduced in #1675. (cc @duairc @teto)

tldr: `escapeShellArg` breaks the pipe in `pass some-account | head -n1`.

### Checklist

- [ ] Change is backwards compatible. (**ACTUALLY NOT**)

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/imapnotify.nix
+++ b/modules/services/imapnotify.nix
@@ -49,8 +49,7 @@ let
       inherit port;
       tls = account.imap.tls.enable;
       username = account.userName;
-      passwordCmd =
-        lib.concatMapStringsSep " " lib.escapeShellArg account.passwordCommand;
+      passwordCmd = lib.concatStringsSep " " account.passwordCommand;
       inherit (account.imapnotify) boxes;
     } // optionalAttrs (account.imapnotify.onNotify != "") {
       onNewMail = account.imapnotify.onNotify;


rycee:
In general I think it is better if the user is not expected to escape the command since different programs may execute the password command in different ways. Indeed, I would recommend avoiding using shell functionality in the command since the program may use `exec()` rather than `system()`.

If the password command relies on shell functionality then I would suggest being explicit about it and use something like
``` nix
passwordCommand = toString (pkgs.writeShellScript "my-passwd-cmd" ''
  pass some-account | head -n1
'');
```

kirelagin:
Well, I mean, I don’t really see how shell-escaping the password command does any good to anyone – what could the user potentially put in there that would require escaping? It just makes things more confusing and inconsistent: for example, `pass ... | head -n1` works perfectly fine with `mbsync`, but not with `imapnotify`.

I don’t have much evidence to cite, but it is my understanding that, within the NixOS/nixpkgs ecosystem, `somethingCommand` has always been taken to mean “a shell command”. And, indeed, “different programs may execute the password command in different ways”, but it is the job of the corresponding modules to make sure it works as expected (e.g. by wrapping it into `writeShellScript` if the program uses `exec`). As a matter of fact, giving arguments to an executable by separating them by spaces is “shell functionality” as well, but it is used right in the examples. Lastly, if it is not a shell command, then, I would say, the option should not be called “command” at all, but rather “exec” or something.

I don’t want to argue and put any pressure on anyone, so I’ll close the PR, but feel free to reopen and merge it in case you change your mind.

sand4rt:
It took a while to understand why it wasn’t working, but I eventually ended up with:

```nix
passwordCommand = toString (
  pkgs.writeShellScript "email-work" ''
    ${pkgs.pass}/bin/pass email/work | ${pkgs.coreutils}/bin/head -n1
  ''
);
```

perhaps there should be a comment in the docs about this?

PR #2925, opened 2022-04-30
home-environment: Stop exporting __HM_SESS_VARS_SOURCED

wahjava:
### Description

If the __HM_SESS_VARS_SOURCED is set in environment of parent process,
then the spawned child shell will inherit it, and won't set the
session variables one expects to be set in their home environment

And, I don't see any need for exporting environment variable, just
setting it is enough for it to be a guard.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


berbiche:
If I understand correctly, if the environment (`home.sessionVariables`, etc.) is changed after it has been loaded (for instance before starting your window manager) then new shells won't see the new environment variables?

wahjava:
> If I understand correctly, if the environment (`home.sessionVariables`, etc.) is changed after it has been loaded (for instance before starting your window manager) then new shells won't see the new environment variables?

Exactly

MonAaraj:
Hey, any progress on this PR? I've encountered this issue myself and this would be the solution. What's stopping this from being merged?

modules/home-environment.nix
```diff
@@ -551,7 +551,7 @@ in
           text = ''
             # Only source this once.
```
bew, line 552: in all these files, you could also change this comment to: `Only source this once per shell`

ncfavier:
Won't this introduce duplicated entries in `PATH`-like variables? I think this might be a problem?

wahjava:
> Won't this introduce duplicated entries in `PATH`-like variables? I think this might be a problem?

I've been using it for an year on multiple systems, and haven't experienced what you're pointing. Could you please elaborate why/how that may happen ?

Thanks

ncfavier:
Things like `export PATH=foo:$PATH` will result in duplicated `foo` entries in nested shells. I don't know if this is a problem.

wahjava:
> Things like `export PATH=foo:$PATH` will result in duplicated `foo` entries in nested shells. I don't know if this is a problem.

Well, I've not noticed it in practice, it did make switching home-manager profiles more deterministic for me when I was experimenting with session variables. And yes, I'd noticed NixOS uses similar concept (environment variable).

PR #2927, opened 2022-04-30
gpg-agent: shell integration is configurable

crazymanjinn:
### Description

In esoteric setups, automatically setting GPG_TTY to current tty is not
desired on every shell startup. This change adds configuration options
to allow user to disable that if desired.

### Checklist

Appears an unrelated test case is failing even before my change:
```
❯ nix-shell --pure tests -A run.all
error: attribute 'shellDryRun' missing

       at /home/user/projects/home-manager/modules/programs/bash.nix:12:9:

           11|     checkPhase = ''
           12|       ${pkgs.stdenv.shellDryRun} "$target"
             |         ^
           13|     '';
```

However, all the gpg-agent related tests do pass:
```
❯ for t in gpg-agent-default-homedir gpg-agent-override-homedir; do nix-shell --pure tests -A run.$t; done
❯ for t in gpg-agent-default-homedir gpg-agent-override-homedir; do nix-shell --pure tests -A run.$t; done
gpg-agent-default-homedir: OK
gpg-agent-override-homedir: OK
```

Change seems trivial enough it shouldn't need new testcases? Happy to be wrong, so please let me know if you would like me to add some.

    ```
    {component}: {description}

    {long description}
    ```

This PR does not.

### DIFF
--- a/modules/services/gpg-agent.nix
+++ b/modules/services/gpg-agent.nix
@@ -198,28 +198,16 @@ in {
         '';
       };
 
-      enableBashIntegration = mkOption {
+      enableBashIntegration = mkEnableOption "Bash integration" // {
         default = true;
-        type = types.bool;
-        description = ''
-          Whether to enable Bash integration.
-        '';
       };
 
-      enableZshIntegration = mkOption {
+      enableZshIntegration = mkEnableOption "Zsh integration" // {
         default = true;
-        type = types.bool;
-        description = ''
-          Whether to enable Zsh integration.
-        '';
       };
 
-      enableFishIntegration = mkOption {
+      enableFishIntegration = mkEnableOption "Fish integration" // {
         default = true;
-        type = types.bool;
-        description = ''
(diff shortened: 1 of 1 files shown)


modules/services/gpg-agent.nix
```diff
@@ -197,6 +197,30 @@ in {
...
         '';
       };
+
+      enableBashIntegration = mkOption {
+        default = true;
+        type = types.bool;
+        description = ''
+          Whether to enable Bash integration.
+        '';
+      };
```
sumnerevans, line 207: These all can be simplified as follows:

```suggestion
      enableBashIntegration = mkEnableOption "Bash integration" // {
        default = true;
      };
```

crazymanjinn: Ah I was just copy/pasting from some other module I saw. Good to know, thanks!

sumnerevans (approved):

PR #2928, opened 2022-05-01
tealdeer: add module

marsam:
### Description

Add https://github.com/dbrgn/tealdeer
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/tealdeer.nix
+++ b/modules/programs/tealdeer.nix
@@ -1,0 +1,53 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+let
+  cfg = config.programs.tealdeer;
+
+  tomlFormat = pkgs.formats.toml { };
+
+  configDir = if pkgs.stdenv.isDarwin then
+    "Library/Application Support"
+  else
+    config.xdg.configHome;
+
+in {
+  meta.maintainers = [ maintainers.marsam ];
+
+  options.programs.tealdeer = {
+    enable = mkEnableOption "Tealdeer";
+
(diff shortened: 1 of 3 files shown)


modules/programs/tealdeer.nix
```diff
@@ -0,0 +1,52 @@
...
+
+in {
+  meta.maintainers = [ maintainers.marsam ];
+
+  options.programs.tealdeer = {
+    enable = mkEnableOption "Tealdeer";
+
+    settings = mkOption {
+      type = tomlFormat.type;
+      default = { };
```
sumnerevans, line 22:
```suggestion
      default = { };
      defaultText = literalExpression "{ }";
```

sumnerevans (changes requested):
LGTM otherwise

sumnerevans (approved):

sumnerevans:
Oh, I think this needed an entry in the news file. @rycee are we still doing those? I haven't seen any lately (possibly because I keep forgetting to request that people add them :eyes:)

modules/programs/tealdeer.nix
rycee: Is there any particular need for the `defaultText` entry? I think the manual output should be decent even with just `default = {}`, although I haven't checked recently 🙂

rycee:
@sumnerevans I typically still add the news entry when merging the PR. I added one for tealdeer a little while ago.

Dunno if those news entries are particularly useful, maybe not?

PR #2937, opened 2022-05-04
sway: Make onChange script use cfg.package if set

mgurlitz:
### Description

On some machines, `pkgs.sway` is not usable and the `wayland.windowManager.sway.package` config can be used to substitute a working version. However `pkgs.sway` is still used for the `onChange` script, so the user may end up with two copies of sway installed or a broken build if `pkgs.sway` doesn't work.

Now if the `package` option is overriden, the `onChange` script will call `swaymsg` without pulling in `pkgs.sway`.

### Checklist

*I attempted this and was given an error "attribute 'shellDryRun' missing" in modules/programs/bash.nix*

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/window-managers/i3-sway/sway.nix
+++ b/modules/services/window-managers/i3-sway/sway.nix
@@ -431,12 +431,14 @@ in {
       home.packages = optional (cfg.package != null) cfg.package
         ++ optional cfg.xwayland pkgs.xwayland;
 
-      xdg.configFile."sway/config" = {
+      xdg.configFile."sway/config" = let
+        swayPackage = if cfg.package == null then pkgs.sway else cfg.package;
+      in {
         source = configFile;
         onChange = ''
           swaySocket="''${XDG_RUNTIME_DIR:-/run/user/$UID}/sway-ipc.$UID.$(${pkgs.procps}/bin/pgrep --uid $UID -x sway || true).sock"
           if [ -S "$swaySocket" ]; then
-            ${pkgs.sway}/bin/swaymsg -s $swaySocket reload
+            ${swayPackage}/bin/swaymsg -s $swaySocket reload
           fi
         '';
       };


modules/services/window-managers/i3-sway/sway.nix
```diff
@@ -436,7 +437,7 @@ in {
         onChange = ''
           swaySocket="''${XDG_RUNTIME_DIR:-/run/user/$UID}/sway-ipc.$UID.$(${pkgs.procps}/bin/pgrep --uid $UID -x sway || true).sock"
           if [ -S "$swaySocket" ]; then
-            ${pkgs.sway}/bin/swaymsg -s $swaySocket reload
+            ${swayPackage}/bin/swaymsg -s $swaySocket reload
```
berbiche, line 440: Can the `if else` be inlined here or the variable scoped to the `onChange` script?
Currently the variable is used once and declared 400 lines before its use.

berbiche (changes requested):
Thanks for the contribution!

modules/services/window-managers/i3-sway/sway.nix
mgurlitz: Thanks for the review/suggestion! I tried inlining it, but ./format wanted the `${}` to be spaced out over a few lines, so I applied your second idea of scoping the var instead.

berbiche (approved):
:+1:

rycee:
@mgurlitz When checking "Commit messages are formatted like …" please make sure that the commit message actually follows the pattern. The checklist is intended to help the reviewer, not mislead them.

mgurlitz:
@rycee My sincere apologies! I had it right in the original PR and then stomped it when fixing the comment from berbiche.

berbiche:
This is also my mistake, I didn't double check the commit message from the web interface 😕

rycee:
No worries. Mainly something to keep in mind 🙂

PR #2939, opened 2022-05-05
picom: sync module with the NixOS one

thiagokokada:
### Description

Sync `picom` module with the one in NixOS: https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/x11/picom.nix. It is not a 1<->1 copy though, since there were some cleanups.

This brings a few advantages:
- Use of float instead of strings to represent float values
- Use of structure settings
- Better type checking for some settings

Also, removing `services.compton` since this module was deprecated on 20.03 release and adapting it for the recent changes is more work than it is worth it.

### Checklist

  + Removal of `services.picom.extraOptions` (can be replicated on `services.picom.settings`)
  + Removal of `services.picom.blue*` (can be replicated on `services.picom.settings`)
  + Some options that used to be strings are now proper floats
  + Removal of `services.compton`

    ```
    {component}: {description}

    {long description}
    ```


thiagokokada:
CC @rycee @berbiche @sumnerevans.

thiagokokada:
Ping @rycee.

modules/services/compton.nix
```diff
@@ -1,43 +0,0 @@
...
-    (mkRenamedOptionModule (old "shadowOffsets") (new "shadowOffsets"))
-    (mkRenamedOptionModule (old "shadowOpacity") (new "shadowOpacity"))
-    (mkChangedOptionModule (old "vSync") (new "vSync") (v: v != "none"))
-  ];
-
-  options.services.compton.enable = mkEnableOption "Compton X11 compositor" // {
-    visible = false;
-  };
-
-  config = mkIf config.services.compton.enable {
```
berbiche, line 36: IMO we should gatekeep the module behind a stateVersion check. If stateVersion >= 22.05 then error else warn.

berbiche (commented):
Left two comments until I have time to review this.

modules/services/compton.nix
thiagokokada: I am not sure how this would work without making this module compatible with the `config.services.picom` itself (because this is a breaking change). So it would break either way.

Also, this module is deprecated since 2020-03-24 according to the news entry. Everybody that used this module should already have migrated.

thiagokokada: If you want I can keep this module but make it throw if activate. This should give a better error than just deleting the file.

But I am not sure if it is worth it, since as I said, this module is deprecated for a long time already.

thiagokokada:
Rebased with current `master`.

modules/services/picom.nix
```diff
@@ -256,33 +224,75 @@ in {
...
+
+      in mkOption {
+        type = topLevel;
+        default = { };
+        example = literalExpression ''
+          blur =
+            { method = "gaussian";
+              size = 10;
+              deviation = 5.0;
+            };
```
rycee, line 258: I think this example should be

```suggestion
          {
            blur = {
              method = "gaussian";
              size = 10;
              deviation = 5.0;
            };
          }
```

Otherwise the example will be rendered using invalid Nix syntax:

picomexample

But, I noticed that the same error is in the NixOS module so I'll leave this unchanged for now.

PR #2941, opened 2022-05-06
services/emacs: Add option to make service start with the session

DamienCassou:
### Description

Add services.emacs.startWithUserSession boolean to indicate that Emacs
must be started with the systemd user session. This is true by default
unless socket activation is also true.

In the past, the user had to choose between socket activation (to get
the Emacs service started when the user uses emacsclient) and
immediate start with the user session. When choosing immediate start
over socket activation and if the Emacs service is stopped at some
point, using emacsclient would start a new Emacs daemon but the
service would still be turned off. This situation would prevent
`home-manager switch` from completing successfully because it wouldn't
be able to start the Emacs service as Emacs is already running.

This new setting makes it possible to have both socket activation and
immediate start at the same time. In this scenario, Emacs is started
with the user session and, after the Emacs service is stopped, using
emacsclient starts the service again.

This new settings also makes it possible to have neither socket
activation nor immediate start.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/emacs.nix
+++ b/modules/services/emacs.nix
@@ -81,6 +81,17 @@ in {
       enable = mkEnableOption "systemd socket activation for the Emacs service";
     };
 
+    startWithUserSession = lib.mkOption {
+      type = lib.types.bool;
+      default = !cfg.socketActivation.enable;
+      defaultText =
+        literalExpression "!config.services.emacs.socketActivation.enable";
+      example = true;
+      description = ''
+        Whether to launch Emacs service with the systemd user session.
+      '';
+    };
+
     defaultEditor = mkOption rec {
       type = types.bool;
       default = false;
@@ -145,7 +156,7 @@ in {
           ExecStopPost =
             "${pkgs.coreutils}/bin/chmod --changes +w ${socketDir}";
         };
-      } // optionalAttrs (!cfg.socketActivation.enable) {
+      } // optionalAttrs (cfg.startWithUserSession) {
         Install = { WantedBy = [ "default.target" ]; };
       };
 
--- a/tests/modules/services/emacs/emacs-socket-and-startWithUserSession.nix
+++ b/tests/modules/services/emacs/emacs-socket-and-startWithUserSession.nix
@@ -1,0 +1,27 @@
+{ lib, pkgs, ... }:
+
+with lib;
+
+{
+  services.emacs = {
+    enable = true;
+    socketActivation.enable = true;
+    startWithUserSession = true;
+  };
+
+  nixpkgs.overlays = [
+    (self: super: rec {
+      emacs = pkgs.writeShellScriptBin "dummy-emacs-28.0.5" "" // {
+        outPath = "@emacs@";
+      };
+      emacsPackagesFor = _:
+        makeScope super.newScope (_: { emacsWithPackages = _: emacs; });
+    })
+  ];
+
+  nmt.script = ''
+    assertFileContains \
+      home-files/.config/systemd/user/emacs.service \
+      "WantedBy=default.target"
+  '';
+}
--- a/tests/modules/services/emacs/default.nix
+++ b/tests/modules/services/emacs/default.nix
@@ -4,4 +4,6 @@
   emacs-socket-27 = ./emacs-socket-27.nix;
   emacs-socket-28 = ./emacs-socket-28.nix;
   emacs-default-editor = ./emacs-default-editor.nix;
+  emacs-socket-and-startWithUserSession =
+    ./emacs-socket-and-startWithUserSession.nix;
 }


DamienCassou:
I've been using that for a week now and everything seems to work fine.

DamienCassou:
If nobody complains, I will just merge that PR.

DamienCassou:
Apparently I don't have commit access anymore. Could someone please review this PR or merge it?

jian-lin:
> When choosing immediate start
over socket activation and if the Emacs service is stopped at some
point, using emacsclient would start a new Emacs daemon

I cannot reproduce it. After I stop the emacs service, `emacsclient` errors:

``` console
> systemctl --user stop emacs
> emacsclient 
/nix/store/l7sfpiy16pkkm67n6m2dhcz5dc5yvwdg-emacs-pgtk-20220604.0/bin/emacsclient: file name or argument required
Try '/nix/store/l7sfpiy16pkkm67n6m2dhcz5dc5yvwdg-emacs-pgtk-20220604.0/bin/emacsclient --help' for more information
```

DamienCassou:
@jian-lin thank you very much for having a look at this PR. What you report seems unrelated to this PR. I think the command you use is just missing some arguments. From the command line, you can start emacsclient this way for example:

```session
$ emacsclient --alternate-editor= --create-frame
```

You can pass a filename as last argument if you want to but that is not necessary (see [the documentation for invoking emacsclient](https://www.gnu.org/software/emacs/manual/html_node/emacs/Invoking-emacsclient.html) for more information).

You can also start emacsclient by using your application launcher as there is a desktop file for emacsclient. Just search your applications for "Emacs (Client)".

jian-lin:
`--alternate-editor=` is the point. If you do not use it, there is no issue with current emacs module. But if you do, issue occurs.

Just curious, since you already use systemd to manage your emacs daemon, why don't you use `systemctl --user restart emacs` to restart it, which is a more reasonable choice to me?


Diff looks good to me though.

DamienCassou:
> Just curious, since you already use systemd to manage your emacs daemon, why don't you use systemctl --user restart emacs to restart it, which is a more reasonable choice to me?

That would be another solution indeed. What I do is just typing <kbd>M-x save-buffers-kill-emacs RET</kbd> and then using my application launcher to start emacsclient. I don't need any terminal.

jian-lin (approved):
A small nit, is it needed to add an assertion to not disable `startWithUserSession` and `socketActivation.enable` at the same time, which makes enabling this service useless?

DamienCassou:
@jian-lin wrote:
> A small nit, is it needed to add an assertion to not disable startWithUserSession and socketActivation.enable at the same time, which makes enabling this service useless?

I'm not sure. I don't see any problem doing the setup you present: it's just a workflow we personally don't need. I even mentioned that at the end of my commit message:

> This new settings also makes it possible to have neither socket activation nor immediate start.

modules/services/emacs.nix
```diff
@@ -81,6 +81,14 @@ in {
       enable = mkEnableOption "systemd socket activation for the Emacs service";
     };
 
+    startWithUserSession = lib.mkOption {
+      default = !cfg.socketActivation.enable;
```
rycee, line 85: Need to add something like

    defaultText = literalExample "config.services.emacs.socketActivation.enable";

otherwise the manual will depend on the configuration, wihch causes unnecessary rebuilds.

modules/services/emacs.nix
```diff
@@ -81,6 +81,14 @@ in {
...
+      example = cfg.socketActivation.enable;
```
rycee, line 86: Cannot make manual depend on the configuration:
```suggestion
      example = true;
```

tests/modules/services/emacs/emacs-socket-and-startWithUserSession.nix
```diff
@@ -0,0 +1,29 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+in {
```
rycee, line 7:
```suggestion
{
```

tests/modules/services/emacs/emacs-socket-and-startWithUserSession.nix
```diff
@@ -0,0 +1,29 @@
...
+  config = {
```
rycee, line 8: Can remove the `config = {` since no option is introduced:
```suggestion
```

tests/modules/services/emacs/emacs-socket-and-startWithUserSession.nix
```diff
@@ -0,0 +1,29 @@
+{ config, lib, pkgs, ... }:
```
rycee, line 1: Since `config` is unused:
```suggestion
{ lib, pkgs, ... }:
```

rycee (commented):
Thanks! I don't use the Emacs service myself but this seems to make sense. I just added a few comments.

I'm wonder, though, is there any situation where you would _not_ want socket activation? Perhaps we should remove that option and always do socket activation?

modules/misc/news.nix
```diff
@@ -2479,6 +2479,15 @@ in
           A new module is available: 'services.mopidy'.
         '';
       }
+
+      {
```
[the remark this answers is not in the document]
DamienCassou: removed

modules/services/emacs.nix
```diff
@@ -81,6 +81,14 @@ in {
...
+      description =
+        "Whether to launch Emacs service with the systemd user session. The default value is <literal>false</literal> if <literal>cfg.socketActivation.enable</literal> is <literal>true</literal>, <literal>true</literal> otherwise.";
```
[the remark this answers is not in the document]
DamienCassou: fixed

tests/modules/services/emacs/emacs-socket-and-startWithUserSession.nix
```diff
@@ -0,0 +1,29 @@
...
+    ];
+
+    services.emacs.enable = true;
+    services.emacs.socketActivation.enable = true;
+    services.emacs.startWithUserSession = true;
+
+    nmt.script = ''
+      assertFileContains \
+                         home-files/.config/systemd/user/emacs.service \
+                        "WantedBy=default.target"
```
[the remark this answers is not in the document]
DamienCassou: fixed

modules/services/emacs.nix
DamienCassou: fixed. I don't understand what I'm doing but I replaced `literalExample` with `literalExpression`.

modules/services/emacs.nix
DamienCassou: fixed

tests/modules/services/emacs/emacs-socket-and-startWithUserSession.nix
DamienCassou: removed

tests/modules/services/emacs/emacs-socket-and-startWithUserSession.nix
DamienCassou: fixed. I also removed `pkgs`

DamienCassou:
I think I have fixed all your feedback.

> is there any situation where you would not want socket activation? Perhaps we should remove that option and always do socket activation?

I don't know why someone would not want socket activation but there might be very good reasons.

modules/services/emacs.nix
rycee: Sorry, force of habit. The `literalExpression` is correct.

rycee:
Thanks! Merged to master now.

PR #2943, opened 2022-05-06
newsboat: allow imperative "urls" management

KFearsoff:
### Description

This PR adds 2 QoL changes: an option to manage the "urls" file imperatively and an option to enable a systemd service which would refresh the feeds periodically.

The former option is valuable if the user doesn't want to expose feeds in a git repo AND doesn't want to use `gitcrypt`, while still enjoying the benefits of managing Newsboat. It makes it possible to use `Syncthing`, `agenix` or `sops-nix` to manage "urls" in a safe way.

The latter option adds a simple systemd service that updates the Newsboat feeds (1 hour by default).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/newsboat.nix
+++ b/modules/programs/newsboat.nix
@@ -68,7 +68,10 @@ in {
           url = "http://example.com";
           tags = [ "foo" "bar" ];
         }];
-        description = "List of news feeds.";
+        description = ''
+          List of news feeds. Leave it empty if you want to manage feeds
+          imperatively, for example, using Syncthing.
+        '';
       };
 
       maxItems = mkOption {
@@ -121,16 +124,24 @@ in {
   };
 
   config = mkIf cfg.enable {
+    assertions = [{
+      assertion = cfg.queries != { } -> cfg.urls != [ ];
+      message = ''
+        Cannot specify queries if urls is empty. Unset queries if you
+        want to manage urls imperatively.
+      '';
+    }];
+
     home.packages = [ pkgs.newsboat ];
 
     # Use ~/.newsboat on stateVersion < 21.05 and use ~/.config/newsboat for
     # stateVersion >= 21.05.
     home.file = mkIf (versionOlder config.home.stateVersion "21.05") {
-      ".newsboat/urls".text = urlsFileContents;
+      ".newsboat/urls" = mkIf (cfg.urls != [ ]) { text = urlsFileContents; };
       ".newsboat/config".text = configFileContents;
     };
     xdg.configFile = mkIf (versionAtLeast config.home.stateVersion "21.05") {
-      "newsboat/urls".text = urlsFileContents;
+      "newsboat/urls" = mkIf (cfg.urls != [ ]) { text = urlsFileContents; };
       "newsboat/config".text = configFileContents;
     };
   };


teto:
I wonder if you could not achieve the same with      `home.file.".newsboat/urls".text = null`. 
As for the service we tend to have them in separate modules, for instance look at mbsync that has modules/programs/mbsync.nix and modules/services/mbsync.nix . Then I am not convinced we should have services that are this simple. It would be best to have an function `mkSystemdTimer` (maybe it exists) and present it better in the wiki/doc

KFearsoff:
I believe `home.file.".newsboat/urls".text = null` won't work. I have tested it a bit, and it seems like that option isn't effectively nullable. It complains about `home.file.".newsboat/urls".source` being used but never defined, and `home.file.".newsboat/urls".source` isn't nullable. I'm not too comfortable hacking on the core functionality to develop a workaround, and I think it's not that useful of a feature anyway.

As for the service, it's a good point. I guess I'll revert the systemd service commit and create a `mkSystemdTimer` function in a separate PR. It sounds useful enough to implement.

KFearsoff:
I decided that force-pushing (like nixpkgs repo tells you to) without the systemd service commit makes more sense than reverting the commit. Also rebased to the modern master.
Anyway, is this PR even required much? Please let me know if it is. Because at this point, I'm second-guessing myself, and I could always just close the PR and make local changes if it's not something that should be merged.

modules/programs/newsboat.nix
```diff
@@ -126,11 +133,11 @@ in {
     # Use ~/.newsboat on stateVersion < 21.05 and use ~/.config/newsboat for
     # stateVersion >= 21.05.
     home.file = mkIf (versionOlder config.home.stateVersion "21.05") {
-      ".newsboat/urls".text = urlsFileContents;
+      ".newsboat/urls" = mkIf cfg.declarativeUrls { text = urlsFileContents; };
```
rycee, line 136: Perhaps simply do
```suggestion
      ".newsboat/urls" = mkIf (cfg.urls != []) { text = urlsFileContents; };
```
instead? Then you don't need to add a special option for this, just update the `urls` option description.

KFearsoff: This looks great, let's go with that.

modules/programs/newsboat.nix
```diff
@@ -68,7 +68,7 @@ in {
           url = "http://example.com";
           tags = [ "foo" "bar" ];
         }];
-        description = "List of news feeds.";
+        description = "List of news feeds. Leave it empty if you want to manage feeds imperatively, e.g. using Syncthing.";
```
rycee, line 71: Best to avoid excessively long lines. For example,
```suggestion
        description = ''
          List of news feeds. Leave it empty if you want to manage feeds
          imperatively, for example, using Syncthing.
        '';
```

modules/programs/newsboat.nix
```diff
@@ -107,7 +107,7 @@ in {
         type = types.attrsOf types.str;
         default = { };
         example = { "foo" = ''rssurl =~ "example.com"''; };
-        description = "A list of queries to use.";
+        description = "A list of queries to use. This option won't work if <varname>urls</varname> is empty.";
```
rycee, line 110: I guess queries don't make much sense unless you have urls to query? If so perhaps best to use an assertion? Something like

```
assertions = [{
  assertion = cfg.queries != [] -> cfg.urls != [];
  message = "Cannot specify queries unless you have urls";
}]
```

perhaps.

KFearsoff:
Not stale. Still useful to me.

sumnerevans (approved):

rycee:
Thanks! Merged to master now.

PR #2945, opened 2022-05-07
Completely overhaul programs/vscode/haskell

schuelermine:
### Description

Updated programs.vscode.haskell to use the new official Haskell extension and haskell-language-server

### Checklist

       Note: `./format` failed with `bash: symbol lookup error: /usr/lib/libc.so.6: undefined symbol: __nptl_change_stack_perm, version GLIBC_PRIVATE`, formatted affected file using `nixfmt` instead.

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/vscode/haskell.nix
```diff
@@ -1,62 +1,33 @@
...
-      (self: super: { hie-nix = import ~/src/hie-nix {}; })
-    ]
-  '';
-
-in {
+in with lib; {
   options.programs.vscode.haskell = {
     enable = mkEnableOption "Haskell integration for Visual Studio Code";
 
-    hie.enable = mkOption {
```
teto, line 28: would be nice to add some warning instead so that it doesnt straightaway breaks configs

modules/programs/vscode/haskell.nix
```diff
@@ -1,62 +1,33 @@
...
-      '';
-      example = literalExpression ''
-        (import ~/src/haskell-ide-engine {}).hies + "/bin/hie-wrapper";
-      '';
+    hlsPackage = mkPackageOption pkgs "Haskell language server" {
+      default = [ "haskell-language-server" ];
     };
+
+    ghcupPackage =
+      mkPackageOption pkgs "GHCup" { default = [ "haskellPackages" "ghcup" ]; };
```
teto, line 16: I dont get why one would use ghcup when it can be installed via nix

modules/programs/vscode/haskell.nix
```diff
@@ -1,62 +1,33 @@
...
-        </para><para>
-        Because hie-nix is not packaged in Nixpkgs, you need to add it as an
-        overlay or set this option. Example overlay configuration:
-        <programlisting language="nix">${exampleOverlay}</programlisting>
-      '';
-      example = literalExpression ''
-        (import ~/src/haskell-ide-engine {}).hies + "/bin/hie-wrapper";
-      '';
+    hlsPackage = mkPackageOption pkgs "Haskell language server" {
+      default = [ "haskell-language-server" ];
```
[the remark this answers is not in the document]
schuelermine: What does the wrapper do?

modules/programs/vscode/haskell.nix
schuelermine: GHCup might have versions nixpkgs doesn’t yet have and it installs to the home path so it might be more compatible with some things. All weird issues to have when using home-manager but isn’t it nice to have options?

modules/programs/vscode/haskell.nix
schuelermine: Good point, will do

modules/programs/vscode/haskell.nix
```diff
@@ -1,62 +1,33 @@
...
+        "haskell.ghcupExecutablePath" = cfg.ghcupPackage;
+        "haskell.manageHLS" = "GHCup";
+      })
+
+      (mkIf (!cfg.useGhcup) {
+        "haskell.serverExecutablePath" = cfg.hlsPackage;
+      })
+    ];
+
+    programs.vscode.extensions = [ pkgs.vscode-extensions.haskell.haskell ];
```
ncfavier, line 31: This should probably still include `justusadam.language-haskell`, which is a dependency of `haskell.haskell`. Otherwise users will have to install it impurely.

schuelermine: Thanks, didn’t know

schuelermine: Has been done

modules/programs/vscode/haskell.nix
schuelermine: Done

schuelermine:
I’m planning on revisiting this once #3459 has been rejected or accepted.

PR #2949, opened 2022-05-09
programs.pywal: init

teto:
pywal allows to change your theme for much software:
https://github.com/dylanaraps/pywal/wiki/Customization#table-of-contents

home-manager is the perfect tool to leverage since a single setting
could enable wal-compatible theming for all this software.

In this PR, I've just enabled zsh, kitty and i3 but I hope follow up PRs will
enable more software.
At some point we may even need a blacklist.

Once enabled, run:
wal --theme random

to change your colorschemes.

edit: for neovim support we would need  https://github.com/dylanaraps/wal.vim/pull/26

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/pywal.nix
+++ b/modules/programs/pywal.nix
@@ -1,0 +1,78 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let cfg = config.programs.pywal;
+
+in {
+  options = { programs.pywal = { enable = mkEnableOption "pywal"; }; };
+
+  config = mkIf cfg.enable {
+
+    home.packages = [ pkgs.pywal ];
+
+    programs.zsh.initExtra = ''
+      # Import colorscheme from 'wal' asynchronously
+      # &   # Run the process in the background.
+      # ( ) # Hide shell job control messages.
+      (cat ~/.cache/wal/sequences &)
+
+      # Alternative (blocks terminal for 0-3ms)
+      cat ~/.cache/wal/sequences
+    '';
+
+    programs.kitty.extraConfig = ''
+      include ${config.xdg.cacheHome}/wal/colors-kitty.conf
+    '';
+
+    programs.rofi.theme."@import" =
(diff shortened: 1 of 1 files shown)


modules/programs/pywal.nix
```diff
@@ -0,0 +1,78 @@
...
+
+  config = mkIf cfg.enable {
+
+    home.packages = [ pkgs.pywal ];
+
+    programs.zsh.initExtra = ''
+      # Import colorscheme from 'wal' asynchronously
+      # &   # Run the process in the background.
+      # ( ) # Hide shell job control messages.
+      (cat ~/.cache/wal/sequences &)
```
rycee, line 18: I think you want `${config.xdg.cacheHome}` here (and just below) as well.

berbiche:
> At some point we may even need a blacklist.

What if we had an option that enables/disables it by default for every tool and then manually enable it?

berbiche:
There was a similar PR last year for a tool called "Flavours", see https://github.com/nix-community/home-manager/pull/2165

teto:
> What if we had an option that enables/disables it by default for every tool and then manually enable it?

that's my intention. I wanted to merge initial support so that others can join in the effort plus it makes other additions easier.

> here was a similar PR last year for a tool called "Flavours", see https://github.com/nix-community/home-manager/pull/2165

I had completely missed it thanks interesting. Looking forward to that PR being reopened :)

modules/programs/pywal.nix
teto: yes, fixed

teto:
damn I realized I had a copy of the module in my dotfiles and realized that after removing it could not load the pywal module from HM. I''ve also found an issue with rofi and have added neovim support locally. Please dont revert, I will fix soon

Misterio77:
You might be interested in [nix-colors](https://github.com/misterio77/nix-colors). It's something I came up with to replace flavours in my home-manager setup :)

Issue #2944, opened 2022-05-07
How to use `dag.entryBefore` with multiple modules?


lovesegfault:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

I have two config files:

`ssh.nix`:
```nix
{ lib, ... }: {
  programs.ssh = {
    enable = true;
    includes = [ "~/.ssh/config.private" ];
    matchBlocks = {
      "github.com" = lib.hm.dag.entryBefore [ "*" ] {
        extraOptions = {
          ControlMaster = "no";
          User = "git";
        };
      };
      "*".extraOptions = {
        AddKeysToAgent = "yes";
        ControlMaster = "auto";
        ControlPath = "~/.ssh/master-%r@%n:%p";
        ControlPersist = "30m";
        ForwardAgent = "no";
      };
    };
  };
}
```

`trusted-ssh.nix`:
```nix
{ config, lib, ... }: {
  programs.ssh.matchBlocks = {
    "github.com" = {
      extraOptions.IdentityFile = "~/.ssh/github";
    };
    "*" = {
      identityFile = [ "~/.ssh/bemeurer" "~/.ssh/yubikey_rsa.pub" ];
      extraOptions = {
        IdentitiesOnly = "yes";
        IdentityAgent = "/run/user/${toString config.home.uid}/gnupg/S.gpg-agent.ssh";
      };
    };
  };
}
```

If I try to build with both of them, I get the following:
```
error: The option `home-manager.users.bemeurer.programs.ssh.matchBlocks.*.after' is defined multiple times.

       Definition values:
       - In `ssh.nix': [ ]
       - In `trusted-ssh.nix': [ ]
(use '--show-trace' to show detailed location information)
```

Trying to add `entryBefore` to `trusted-ssh.nix` yields similar results:
```
error: The option `home-manager.users.bemeurer.programs.ssh.matchBlocks.aarch64.nixos.community.before' is defined multiple times.

       Definition values:
       - In `ssh.nix':
           [
             "*"
           ]
       - In `trusted-ssh.nix':
           [
             "*"
           ]
(use '--show-trace' to show detailed location information)
```

What is the right way of splitting a config across modules when using `dag`?

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 5.17.5, NixOS, 22.05 (Quokka), 22.05.20220505.836c52e`
 - multi-user?: `yes`
 - sandbox: `yes`
[...]

PR #2950, opened 2022-05-09
lib/types: make DAG entries mergeable

ncfavier:
### Description

Fixes https://github.com/nix-community/home-manager/issues/2944

Removes the `uniq` constraint on `after` and `before` so that we can merge multiple definitions for the same DAG entry:

```nix
{
  dag = mkMerge [
    {
      foo = lib.hm.dag.entryBefore [ "bar" ] {
        # definition 1
      };
    }
    {
      foo = lib.hm.dag.entryBefore [ "qux" ] {
        # definition 2
      };
    }
    {
      foo = {
        # definition 3
      };
    }
  ];
}
```

In this example `foo` will come before `bar` and `qux`.

It might also be worth considering enabling things like

```nix
{
  foo.after = "qux";
}
```

which currently would be treated as a definition of the underlying data (so the change wouldn't be backwards compatible), though that can be done with the current PR as

```nix
{
  foo = lib.hm.dag.entryAfter "qux" (mkMerge []);
}
```

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/lib/types-dag.nix
+++ b/modules/lib/types-dag.nix
@@ -3,8 +3,8 @@
 let
   inherit (lib)
     concatStringsSep defaultFunctor fixedWidthNumber imap1 isAttrs isList length
-    listToAttrs mapAttrs mkIf mkOption mkOptionType nameValuePair stringLength
-    types warn;
+    listToAttrs mapAttrs mkIf mkOrder mkOption mkOptionType nameValuePair
+    stringLength types warn;
 
   isDagEntry = e: isAttrs e && (e ? data) && (e ? after) && (e ? before);
 
@@ -13,21 +13,30 @@ let
       submoduleType = types.submodule ({ name, ... }: {
         options = {
           data = mkOption { type = elemType; };
-          after = mkOption { type = with types; uniq (listOf str); };
-          before = mkOption { type = with types; uniq (listOf str); };
+          after = mkOption { type = with types; listOf str; };
+          before = mkOption { type = with types; listOf str; };
         };
         config = mkIf (elemType.name == "submodule") {
           data._module.args.dagName = name;
         };
       });
-      maybeConvert = v: if isDagEntry v then v else dag.entryAnywhere v;
+      maybeConvert = def:
+        if isDagEntry def.value then
+          def.value
+        else
+          dag.entryAnywhere (if def ? priority then
+            mkOrder def.priority def.value
+          else
+            def.value);
     in mkOptionType {
       name = "dagEntryOf";
       description = "DAG entry of ${elemType.description}";
       # leave the checking to the submodule type
       merge = loc: defs:
-        submoduleType.merge loc
-        (map (def: def // { value = maybeConvert def.value; }) defs);
+        submoduleType.merge loc (map (def: {
+          inherit (def) file;
+          value = maybeConvert def;
+        }) defs);
     };
 
 in rec {
@@ -42,7 +51,7 @@ in rec {
     let attrEquivalent = types.attrsOf (dagEntryOf elemType);
     in mkOptionType rec {
       name = "dagOf";
-      description = "DAG of ${elemType.description}s";
+      description = "DAG of ${elemType.description}";
       inherit (attrEquivalent) check merge emptyValue;
       getSubOptions = prefix: elemType.getSubOptions (prefix ++ [ "<name>" ]);
       getSubModules = elemType.getSubModules;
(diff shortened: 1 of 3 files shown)


lovesegfault:
This works great!

lovesegfault:
Is landing this blocked on something?

lovesegfault:
@ncfavier Ping :)

ncfavier:
Not as far as I'm concerned :)

lovesegfault:
Pinging @rycee @sumnerevans for merging

lovesegfault:
Or, maybe @teto can merge?

berbiche (approved):
LGTM!

tests/modules/services/window-managers/sway/sway-no-xwayland.nix
```diff
@@ -17,7 +17,7 @@
         ${
           pkgs.writeText "expected" ''
             xwayland disable
-                    ''
```
sumnerevans, line 20: This is a good change, but irrelevant to this PR, so please revert and make a new PR for this

tests/lib/types/dag-merge.nix
```diff
@@ -22,6 +22,9 @@ in {
         between =
           mkIf true (dag.entryBetween [ "after" ] [ "before" ] "between");
       }
+      { last = "three"; }
+      { last = dag.entryAfter [ "after" ] "two"; }
+      { last = dag.entryAfter [ "before" ] "one"; }
```
sumnerevans, line 27: This test doesn't convince me that this actually works as intended. The test should ensure that the values are actually merged. In this case, it could be that they weren't actually merged correctly since it's at the end of the list anyway.

Maybe if you bounded it on both sides it would be more convincing? That is, having a merged `dag.entryBefore` on `last` as well that bounds where this can be in the output.

Also, how does this work with `entryBetween`?

sumnerevans (changes requested):
I think this change is probably fine, but it really needs some more test coverage to convince me it's correct.

tests/modules/services/window-managers/sway/sway-no-xwayland.nix
ncfavier: So the story behind this change is I used to run `bash format` instead of `./format` because I have nixfmt installed globally and it was faster, but as a result I was using nixfmt 0.5.0 instead of the pinned 0.4.0, so I assumed that a previous PR did not format that file properly when the formatting difference was only due to the update.

nixfmt was updated in https://github.com/nix-community/home-manager/pull/3029 so I'll just drop the commit.

tests/lib/types/dag-merge.nix
ncfavier: I've added both an `entryBetween` and an `entryBefore` to test this better. (Note that `entryBetween` has its arguments flipped; this really means "entry between before and after")

tests/lib/types/dag-merge.nix
```diff
@@ -22,6 +22,9 @@ in {
         between =
           mkIf true (dag.entryBetween [ "after" ] [ "before" ] "between");
       }
+      { merged = "three"; }
+      { merged = dag.entryBetween [ "after" ] [ "before" ] "two"; }
+      { merged = dag.entryBefore [ "between" ] "one"; }
```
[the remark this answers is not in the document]
ncfavier: I added more tests, including orderings on the `merged` values so that we don't rely on an unspecified ordering.

Note that `merged = mkAfter "foo";` would not work, first because it would hit https://github.com/NixOS/nixpkgs/pull/180701, but beyond that because it would order the *submodule* definition instead of the `data` definition. You'd need to do `merged = dag.entryAnywhere (mkAfter "foo");`.

modules/lib/types-dag.nix
```diff
@@ -42,7 +42,7 @@ in rec {
     let attrEquivalent = types.attrsOf (dagEntryOf elemType);
     in mkOptionType rec {
       name = "dagOf";
-      description = "DAG of ${elemType.description}s";
+      description = "DAG of ${elemType.description}";
```
ncfavier, line 45: This is to align with https://github.com/NixOS/nixpkgs/pull/170561

tests/lib/types/dag-merge.nix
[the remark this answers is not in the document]
ncfavier: Okay, we can do `merged = mkAfter "foo";` now. Basically taking the useless `priority` from the submodule definition (`after` and `before` are not order-sensitive) and shoving it into `data`. This is probably deep into nobody-will-never-use-this territory, but it seems more correct and doesn't cost anything.

rycee:
Thanks! Merged to master now.

PR #2957, opened 2022-05-13
nushell: update/fix configuration management, fixes #2952 and fixes #2837

### LINKED ISSUE #2837: nushell: module will need upgrade for 0.60.0 package
### Description

see: https://github.com/NixOS/nixpkgs/pull/165473

see: https://www.nushell.sh/blog/2022-03-22-nushell_0_60.html#a-new-config-file

cc: @Philipp-M

### LINKED ISSUE #2952: bug: Nushell config no longer works

### Issue description

Nushell configuration with home-manager no longer works, with the version of nushell shipped in nixpkgs (unstable). To reproduce:

1. Enable and configure nushell with home-manager
2. Run nushell with `nu`
3. Notice that it complains "No environment config file found at /home/jon/.config/nushell/env.nu ... Would you like to create one with defaults (Y/n):" even though a configuration exists at ~/.config/nu/config.toml. 

This is related to #2837 but is now a bug, and not a feature request. 

### Maintainer CC

@Philipp-M @marsam 

### System information

```markdown
/nix/store/r318fagbqjnz3fka0pj1a2iz08bb08c3-nix-info
copying path '/nix/store/r318fagbqjnz3fka0pj1a2iz08bb08c3-nix-info' from 'https://cache.nixos.org'...
 - system: `"x86_64-linux"`
 - host os: `Linux 5.17.5-76051705-generic, Pop!_OS, 22.04 LTS, nobuild`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.8.0`
 - channels(jon): `"home-manager, nixpkgs"`
 - channels(root): `"nixpkgs"`
 - nixpkgs: `/home/jon/.nix-defexpr/channels/nixpkgs`
```

Philipp-M:
### Description

nushell introduced a new way to configure nushell, described here: https://www.nushell.sh/book/configuration.html#nushell-configuration-with-env-nu-and-config-nu

### Checklist

- [x] Code tested through `nix-shell --pure tests -A run.all`. (This currently fails because of bash, but `nix-shell --pure tests -A run.nushell-example-settings` succeds)

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/nushell.nix
```diff
@@ -21,40 +21,55 @@ in {
...
-        let
-          prim = oneOf [ bool int str ];
-          primOrPrimAttrs = either prim (attrsOf prim);
-          entry = either prim (listOf primOrPrimAttrs);
-          entryOrAttrsOf = t: either entry (attrsOf t);
-          entries = entryOrAttrsOf (entryOrAttrsOf entry);
-        in attrsOf entries // { description = "Nushell configuration"; };
-      default = { };
+    config = mkOption {
+      type = types.nullOr types.path;
```
rycee, line 25: I would suggest having

```suggestion
      type = with types; nullOr (either str path);
```

and allow either a path to a file or a Nix string.

modules/programs/nushell.nix
```diff
@@ -21,40 +21,55 @@ in {
...
-      type = with types;
-        let
-          prim = oneOf [ bool int str ];
-          primOrPrimAttrs = either prim (attrsOf prim);
-          entry = either prim (listOf primOrPrimAttrs);
-          entryOrAttrsOf = t: either entry (attrsOf t);
-          entries = entryOrAttrsOf (entryOrAttrsOf entry);
-        in attrsOf entries // { description = "Nushell configuration"; };
-      default = { };
+    config = mkOption {
```
rycee, line 24: Perhaps `configFile` and `envFile` to avoid the heavily overloaded `config` name?

modules/programs/nushell.nix
Philipp-M: With Nix string you mean the contents of the file or a string of a path?

rycee: The string would be the content of the file. So that it becomes possible to write something like

```
programs.nushell.config = ''
  Whatever
'';
```

and if you want to use a file

```
programs.nushell.config = ./whatever.nu;
```

schuelermine:
Is this PR going to be merged?

rycee:
I had a think, and tilted over to having explicit text and source options (similar to how `home.file` does it). I made that change and pushed to this PR.

I also added a `mkRemovedOptionModule` for the old settings options to produce a nice error message for people with old configurations.

Please have a look, if it seems OK to you I'll merge.

Philipp-M:
Yes, looks good to me, nice, thanks!

rycee:
Cool. Merged now. I guess this should be backported to the 22.05 branch?

TheOPtimal:
Sorry, but why does it have either text or path? There's a good reason that other shells don't do this - now I can't make integrations for Nu,

Philipp-M:
I don't understand exactly can you explain your problem in more detail (and maybe open a separate issue for that)?

TheOPtimal:
> I don't understand exactly can you explain your problem in more detail (and maybe open a separate issue for that)?

Basically, e.g. Starship:

Screenshot from 2022-10-22 16-54-25

Has an option to automatically add it's shell hook to whatever shell you're using. Implementing this functionality for Nu becomes impossible because if the user has it configured to use a file, then we can't add extra hooks for other programs. `configFile` and `envFile` should just be text, so that other programs can add their shell hooks in no problem.

Philipp-M:
I see...

I've created a PR: #3347

This ensures backwards compatibility, you may want to test this.

TheOPtimal:
Thank you so much, this is great.

PR #2960, opened 2022-05-14
mujmap: add module

elizagamedev:
mujmap is a tool which synchronizes mail between a mail server and
notmuch via JMAP. It's very similar to lieer, so I heavily based the
implementation of the notmuch module on lieer's. I did not include an
equivalent to lieer's periodic synchronization service, however, because
I plan to soon introduce a daemon mode to mujmap.

https://github.com/elizagamedev/mujmap

This isn't quite ready for merge yet, as I'm still waiting on
NixOS/nixpkgs#172648. However, this is my first module contribution to
home-manager, and I thought it was worth it to receive early feedback.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/mujmap.nix
+++ b/modules/programs/mujmap.nix
@@ -1,0 +1,315 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.programs.mujmap;
+
+  mujmapAccounts =
+    filter (a: a.mujmap.enable) (attrValues config.accounts.email.accounts);
+
+  missingNotmuchAccounts = map (a: a.name)
+    (filter (a: !a.notmuch.enable && a.mujmap.notmuchSetupWarning)
+      mujmapAccounts);
+
+  notmuchConfigHelp =
+    map (name: "accounts.email.accounts.${name}.notmuch.enable = true;")
+    missingNotmuchAccounts;
+
+  settingsFormat = pkgs.formats.toml { };
+
+  filterNull = attrs: attrsets.filterAttrs (n: v: v != null) attrs;
+
+  configFile = account:
+    let
+      settings'' = if (account.jmap == null) then
+        { }
+      else
+        filterNull {
+          fqdn = account.jmap.host;
+          session_url = account.jmap.sessionUrl;
+        };
+
+      settings' = settings'' // {
+        username = account.userName;
+        password_command = escapeShellArgs account.passwordCommand;
+      } // filterNull account.mujmap.settings;
+
+      settings = if (hasAttr "fqdn" settings') then
+        (removeAttrs settings' [ "session_url" ])
+      else
+        settings';
+    in {
+      name = "${account.maildir.absPath}/mujmap.toml";
+      value.source = settingsFormat.generate
+        "mujmap-${lib.replaceStrings [ "@" ] [ "_at_" ] account.address}.toml"
+        settings;
+    };
+
+  tagsOpts = {
+    lowercase = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        If true, translate all mailboxes to lowercase names when mapping to notmuch
+        tags.
+      '';
+    };
+
+    directory_separator = mkOption {
+      type = types.str;
+      default = "/";
+      example = ".";
+      description = ''
+        Directory separator for mapping notmuch tags to maildirs.
+      '';
+    };
+
+    inbox = mkOption {
+      type = types.str;
+      default = "inbox";
+      description = ''
+        Tag for notmuch to use for messages stored in the mailbox labeled with the
+        <code>Inbox</code> name attribute.
+        </para><para>
+        If set to an empty string, this mailbox <emphasis>and its child
+        mailboxes</emphasis> are not synchronized with a tag.
+      '';
+    };
+
+    deleted = mkOption {
+      type = types.str;
+      default = "deleted";
+      description = ''
+        Tag for notmuch to use for messages stored in the mailbox labeled with the
+        <code>Trash</code> name attribute.
+        </para><para>
+        If set to an empty string, this mailbox <emphasis>and its child
+        mailboxes</emphasis> are not synchronized with a tag.
+      '';
+    };
+
+    sent = mkOption {
+      type = types.str;
+      default = "sent";
+      description = ''
+        Tag for notmuch to use for messages stored in the mailbox labeled with the
+        <code>Sent</code> name attribute.
+        </para><para>
+        If set to an empty string, this mailbox <emphasis>and its child
+        mailboxes</emphasis> are not synchronized with a tag.
+      '';
+    };
+
+    spam = mkOption {
+      type = types.str;
+      default = "spam";
+      description = ''
+        Tag for notmuch to use for messages stored in the mailbox labeled with the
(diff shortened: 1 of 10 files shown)


teto:
Great job ! 

My generated mujmap.toml contained `password_command = ["/nix/store/acfizib24wk6hq20qg5x9f3n8398q6lh-pass-show/bin/pass-show", "perso/fastmail_mc"]` which is apparently not supported by mujmap (it expects a string).

Also to setup the module I had to set mujmap.fqdn: in modules/accounts/email.nix there is a setting called "flavor" which lists "fastmail.com" and could be used to autoset fqdn

elizagamedev:
Pushed a commit to fix both of these issues.

teto:
@elizagamedev CI fails for mundane reasons (can't find you as a maintainer), could you fix it ? other than that it looks good we have tests, integration with the mail backend, a pretty good first module. I will do a better review after testing/CI passes

elizagamedev:
@teto I'm not actually sure how to fix CI, since I added myself to the maintainers list but it's still failing. Is there another place I missed?

teto:
ha yeah I've tried it locally and the maintainer infra is complex. @rycee any idea ?

elizagamedev:
I used snake case so that it could directly be serialized to the config file like the lieer module does, but if it is preferred to have them as camel case it should not be too hard to translate the names before serialization.

teto:
dont do it yet as I am unsure, waiting confirmation from the HM gods :)

teto:
while doing a nix
```
error: store path 'm34v88fkaklilvsvv75vld6q384zqyra-mujmap-MYUSERNAME@fastmail.com.toml' contains illegal character '@'

       … while evaluating 'check'
```
Once could do
`lib.replaceStrings ["@"] ["a"] "t@to" `
or call sanitizeDerivationName  from lib/strings.nix

teto:
so using the module I get 
```
$ mujmap sync -v
[2022-06-16T22:19:30Z WARN  mujmap::sync] Could not read mujmap state file `./mujmap.state.json': No such file or directory (os error 2)
Retrieving metadata... (11157 possibly changed)
Applying changes to notmuch database... (11157 new, 11157 changed, 0 destroyed)
error: Could not sync mail: Could not add new local email: Something went wrong trying to read or write a file
```
the error is not verbose enough for a proper diagnostic. I still have my maildir on another drive (ext4), can it be it ?
```
ls -l ~/maildir
lrwxrwxrwx 1 teto users 16 2021-10-02  /home/teto/maildir -> /mnt/ext/maildir
```

elizagamedev:
> Once could do `lib.replaceStrings ["@"] ["a"] "t@to" ` or call sanitizeDerivationName from lib/strings.nix

Done.

> the error is not verbose enough for a proper diagnostic. I still have my maildir on another drive (ext4), can it be it ?

This is possible. I added some new config options for v2.0 which I forgot to add to this module which solve that problem (the [troubleshooting section of the readme](https://github.com/elizagamedev/mujmap/blob/main/README.md#troubleshooting) now addresses this). They have been added to the module. Does messing with those settings help?

I also tweaked `fqdnXorSessionUrlAssertions`, since v2.0 can now auto-detect the fqdn from the domain part of the username. Now it just checks to make sure you don't specify both.

elizagamedev:
> I get the same error `error: store path 'm34v88fkaklilvsvv75vld6q384zqyra-mujmap-MYUSERNAME@fastmail.com.toml' contains illegal character '@'` with your change. Actually I think sanitizeDerivationName is broken, I've discovered and had the same problem recently. For now `lib.replaceStrings ["@"] ["a"] "t@to" ` is probably enough.

I changed it to replace "`@`" with "`_at_`" since that seems pretty unique. I don't know why it was working for me and not you, maybe a difference in nix version?

> I've found the piece that recommands camelCase https://nix-community.github.io/home-manager/index.html#sec-code-style

It actually looks like there's an exception to this in the [manual](https://nix-community.github.io/home-manager/). (Section 5.2.3, Add only valuable options)

> If the software uses a structured configuration format like a JSON, YAML, INI, TOML, or even a plain list of key/value pairs then consider using a settings option as described in [Nix RFC 42](https://github.com/Infinisil/rfcs/blob/config-option/rfcs/0042-config-option.md).

The RFC linked is what lieer does, and because this is modeled after lieer's module, I inherited this way of doing things without realizing it's the "new" way. So changing the module to use camelCase would actually be going backwards. Perhaps the code style sction you linked ought to be amended to point this exception out to prevent future confusion.

That said, the "essential-ness" of some of the config options I explicitly included is a bit questionable. I removed the following explicit options:

- `convert_dos_to_unix`
- `timeout`
- `retries`
- `concurrent_downloads`

modules/lib/maintainers.nix
```diff
@@ -281,4 +281,10 @@
       fingerprint = "D5D6 FD1F 0D9A 3284 FB9B  C26D 3F98 EC7E C2B8 7ED1";
     }];
   };
+  elizagamedev = {
+    email = "eliza@eliza.sh";
+    github = "elizagamedev";
+    githubId = 4576666;
+    name = "Eliza Velasquez";
+  };
```
rycee, line 289: You can remove this since you are already listed as a maintainer in Nixpkgs.
```suggestion
```

modules/programs/mujmap.nix
```diff
@@ -0,0 +1,312 @@
...
+    missingNotmuchAccounts;
+
+  fqdnForFlavor = flavor:
+    if flavor == "fastmail.com" then "fastmail.com" else null;
+
+  notBothFqdnAndSessionUrlAssertions = map (a: {
+    assertion = a.mujmap.settings.fqdn == null || a.mujmap.settings.session_url
+      == null;
+    message =
+      "For `accounts.email.accounts.${a.name}', `notmuch.settings' may not specify both `fqdn' and `session_url'";
```
rycee, line 26: Perhaps you mean
```suggestion
      "For `accounts.email.accounts.${a.name}', `mujsmap.settings' may not specify both `fqdn' and `session_url'";
```

modules/programs/mujmap.nix
```diff
@@ -0,0 +1,312 @@
...
+  missingNotmuchAccounts = map (a: a.name)
+    (filter (a: !a.notmuch.enable && a.mujmap.notmuchSetupWarning)
+      mujmapAccounts);
+
+  notmuchConfigHelp =
+    map (name: "accounts.email.accounts.${name}.notmuch.enable = true;")
+    missingNotmuchAccounts;
+
+  fqdnForFlavor = flavor:
+    if flavor == "fastmail.com" then "fastmail.com" else null;
```
rycee, line 20: It may be better to add support for jmap in `accounts/email.nix` and then add an entry for jmap in the

```nix
mkIf (config.flavor == "fastmail.com") { … }
```

block. Then that configuration could be used in other modules that support JMAP.

teto: that's a good idea, the same way we have an imapModule/smtpModule, we could have a jmapModule.

elizagamedev:
Thank you for your feedback, @rycee. I added a JMAP config option and modified the mujmap module to use it. Let me know if you want additional changes or if this is good to merge.

rycee:
Nice! I've squashed and merged to master now. Thanks for the contribution!

PR #2915, opened 2022-04-24
Remove platform check from the systemd module

midchildan:
### Description
This is related to https://github.com/nix-community/home-manager/pull/2497#discussion_r754814040.

The systemd module currently contains checks to make sure user services aren't defined on Darwin hosts. But with launchd support now in HM, this can result in feature gating boilerplate when trying to create modules that supports both systemd and launchd.

After this change, modules would be able to define both systemd services and launchd agents without the boilerplate. The systemd module would just ignore them on macOS and vice versa for launchd. I'm hoping this wouldn't be a big issue because most consumers of the systemd module already checks if it supports the target OS. I've also added checks for those that didn't.

For a more concrete example, here's how a module supporting both systemd and launchd might look before this change:

```nix
{ lib, pkgs, ... }:

let
  cfg = config.services.example;
  inherit (lib) mkIf mkMerge;
  inherit (pkgs.stdenv.hostPlatforms) isLinux isDarwin;
in {
  options.services.example.enable = mkEnableOption "example service";

  config = mkIf cfg.enable (mkMerge [
    (lib.mkIf isLinux {
      systemd.user.services.example = {
        Unit.Description = "example service";
        Service.ExecStart = "${pkgs.example}/bin/example";
        Install.WantedBy = [ "default.target" ];
      };
    })

    (lib.mkIf isDarwin {
      launchd.agents.example = {
        enable = true;
        config = {
          ProgramArguments = [ "${pkgs.example}/bin/example" ];
          ProcessType = "Background";
        };
      };
    })
  ]);
}
```

And here's how it'd look after this change:

```nix
{ lib, pkgs, ... }:

let
  cfg = config.services.example;
  inherit (lib) mkIf mkMerge;
in {
  options.services.example.enable = mkEnableOption "example service";

  config = mkIf cfg.enable {
    systemd.user.services.example = {
      Unit.Description = "example service";
      Service.ExecStart = "${pkgs.example}/bin/example";
      Install.WantedBy = [ "default.target" ];
    };

    launchd.agents.example = {
      enable = true;
      config = {
        ProgramArguments = [ "${pkgs.example}/bin/example" ];
        ProcessType = "Background";
      };
    };
  };
}
```

### Checklist

- [X] Change is backwards compatible.

- [X] Code formatted with `./format`.

- [X] Code tested through `nix-shell --pure tests -A run.all`.

[...]

PR #2964, opened 2022-05-17
gpg-agent: darwin support via launchd service

montchr:
### Description

- Add support for running the `gpg-agent` service on Darwin/macOS via `launchd`.
- Add support for `pinentry-mac` as `pinentry_program` and make it the default on macOS.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/gpg-agent.nix
+++ b/modules/services/gpg-agent.nix
@@ -89,7 +89,7 @@ let
   in hexString: (foldl' go initState (splitChars hexString)).ret;
 
 in {
-  meta.maintainers = [ maintainers.rycee ];
+  meta.maintainers = with maintainers; [ rycee montchr ];
 
   options = {
     services.gpg-agent = {
--- a/modules/services/gpg-agent.nix
+++ b/modules/services/gpg-agent.nix
@@ -8,11 +8,10 @@ let
 
   homedir = config.programs.gpg.homedir;
 
-  pinentryBinPath = optionalString (cfg.pinentryFlavor != null)
-    (if cfg.pinentryFlavor == "mac" then
-      "${pkgs.pinentry_mac}/Applications/pinentry-mac.app/Contents/MacOS/pinentry-mac"
-    else
-      "${pkgs.pinentry.${cfg.pinentryFlavor}}/bin/pinentry");
+  pinentryBinPath = (if cfg.pinentryFlavor == "mac" then
+    "${pkgs.pinentry_mac}/Applications/pinentry-mac.app/Contents/MacOS/pinentry-mac"
+  else
+    "${pkgs.pinentry.${cfg.pinentryFlavor}}/bin/pinentry");
 
   gpgSshSupportStr = ''
     ${gpgPkg}/bin/gpg-connect-agent updatestartuptty /bye > /dev/null
@@ -27,13 +26,13 @@ let
   # runs as a daemon, so this is already set.
   sshAuthSockStr =
     let sockPathCmd = "$(${gpgPkg}/bin/gpgconf --list-dirs agent-ssh-socket)";
-    in (if pkgs.stdenv.hostPlatform.isLinux then ''
+    in if pkgs.stdenv.hostPlatform.isLinux then ''
       if [[ -z "$SSH_AUTH_SOCK" ]]; then
         export SSH_AUTH_SOCK="${sockPathCmd}"
       fi
     '' else ''
       export SSH_AUTH_SOCK="${sockPathCmd}"
-    '');
+    '';
 
   gpgFishInitStr = ''
     set -gx GPG_TTY (tty)
@@ -206,6 +205,8 @@ in {
         type = types.nullOr (types.enum (pkgs.pinentry.flavors ++ [ "mac" ]));
         example = "gnome3";
         default = if pkgs.stdenv.hostPlatform.isMacOS then "mac" else "gtk2";
+        defaultText = literalExpression
+          ''if pkgs.stdenv.hostPlatform.isMacOS then "mac" else "gtk2"'';
         description = ''
           Which pinentry interface to use. If not
           <literal>null</literal>, it sets
@@ -278,19 +279,6 @@ in {
       '') cfg.sshKeys;
     })
 
-    {
-      launchd.agents.gpg-agent = {
-        enable = true;
-        config = {
-          ProgramArguments = [ "${gpgPkg}/bin/gpgconf" "--launch" "gpg-agent" ]
-            ++ optionals cfg.verbose [ "--verbose" ];
-          RunAtLoad = true;
-          KeepAlive.SuccessfulExit = false;
-          EnvironmentVariables.GNUPGHOME = homedir;
-        };
-      };
-    }
-
     # The systemd units below are direct translations of the
     # descriptions in the
     #
@@ -330,6 +318,17 @@ in {
         };
 
         Install = { WantedBy = [ "sockets.target" ]; };
+      };
+
+      launchd.agents.gpg-agent = {
+        enable = true;
+        config = {
+          ProgramArguments = [ "${gpgPkg}/bin/gpgconf" "--launch" "gpg-agent" ]
+            ++ optionals cfg.verbose [ "--verbose" ];
+          RunAtLoad = true;
+          KeepAlive.SuccessfulExit = false;
+          EnvironmentVariables.GNUPGHOME = homedir;
+        };
       };
     }
 
--- a/modules/services/gpg-agent.nix
+++ b/modules/services/gpg-agent.nix
@@ -32,14 +32,16 @@ let
   # we cannot use `gpgconf` directly because it heavily depends on system
   # state, but we need the values at build time. original:
   # https://github.com/gpg/gnupg/blob/c6702d77d936b3e9d91b34d8fdee9599ab94ee1b/common/homedir.c#L672-L681
-  gpgconf = dir:
+  sockRelPath = dir:
     let
       hash =
         substring 0 24 (hexStringToBase32 (builtins.hashString "sha1" homedir));
     in if homedir == options.programs.gpg.homedir.default then
-      "%t/gnupg/${dir}"
+      "gnupg/${dir}"
     else
-      "%t/gnupg/d.${hash}/${dir}";
+      "gnupg/d.${hash}/${dir}";
+  gpgconf = dir: "%t/${sockRelPath dir}";
+  gpgconf' = dir: "/tmp/${sockRelPath dir}";
 
   # Act like `xxd -r -p | base32` but with z-base-32 alphabet and no trailing padding.
   # Written in Nix for purity.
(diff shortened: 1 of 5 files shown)


tests/modules/services/gpg-agent/default-homedir.nix
```diff
@@ -2,7 +2,15 @@
```
rycee, line 2: To make the tests actually run for macOS, I think you would need to move the `./modules/services/gpg-agent` line in `tests/default.nix` to the common section.

rycee:
Thanks for the contribution! I've added a few comment 🙂

modules/services/gpg-agent.nix
```diff
@@ -256,79 +252,112 @@ in {
...
+        enable = true;
+        config = {
+          ProgramArguments = [
+            "/bin/sh"
+            "-c"
+            "/bin/ln -sf $GNUPGHOME/S.gpg-agent.ssh $SSH_AUTH_SOCK"
+          ];
+          RunAtLoad = true;
+          EnvironmentVariables = { GNUPGHOME = homedir; };
+          # KeepAlive.SuccessfulExit = false;
```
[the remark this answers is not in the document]
montchr: I uncommented it since it's pretty standard in background `launchd` configurations I've seen. I copied it over from the nix-darwin module initially and meant to double check its purpose before enabling.

modules/services/gpg-agent.nix
```diff
@@ -1,12 +1,8 @@
 { config, options, lib, pkgs, ... }:
-
```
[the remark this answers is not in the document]
montchr: Reverted the linespacing changes and reformatted.

montchr:
Update: I don't have access to a system running macOS at the moment (NixOS only right now), but I expect I will be able to follow up on this in the next couple weeks.

modules/services/gpg-agent.nix
```diff
@@ -256,79 +252,112 @@ in {
...
-          RefuseManualStart = true;
-        };
-
-        Service = {
-          ExecStart = "${gpgPkg}/bin/gpg-agent --supervised"
-            + optionalString cfg.verbose " --verbose";
-          ExecReload = "${gpgPkg}/bin/gpgconf --reload gpg-agent";
-          Environment = "GNUPGHOME=${homedir}";
+    (mkIf
+      (pkgs.stdenv.hostPlatform.isLinux && !pkgs.stdenv.buildPlatform.isDarwin)
```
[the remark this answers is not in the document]
montchr: I had been running into issues on Darwin with `gpgconf` but I don't think it applies any more.

montchr:
I think this is ready for another review, though I should note that I would prefer that it not be merged until someone can verify it on macOS. I expected that I'd be able to do so by this point, but it'll be a bit longer before I have a Mac to test on. In the meantime, I'll keep tracking this version of the module on NixOS (but it's been working fine there for a while now).

Also, the new tests pass, however some unrelated test failed:

```console
building '/nix/store/8xjprr5yi0r9j8j8b6p2nh3rn7jykdrm-neovim-0.7.0.drv'...
Generating remote plugin manifest
Error detected while processing function coc#rpc#start_server[13]..coc#util#get_data_home:
line   15:
E117: Unknown function: coc#float#create_notification
remote/host: generated rplugin manifest: /nix/store/8384rxk99idi5pz20hf233fdp69i5axi-neovim-0.7.0/rplugin.vim
Generating rplugin.vim failed!
error: builder for '/nix/store/8xjprr5yi0r9j8j8b6p2nh3rn7jykdrm-neovim-0.7.0.drv' failed with exit code 1;
       last 6 log lines:
       > Generating remote plugin manifest
       > Error detected while processing function coc#rpc#start_server[13]..coc#util#get_data_home:
       > line   15:
       > E117: Unknown function: coc#float#create_notification
       > remote/host: generated rplugin manifest: /nix/store/8384rxk99idi5pz20hf233fdp69i5axi-neovim-0.7.0/rplugin.vim
       > Generating rplugin.vim failed!
       For full logs, run 'nix log /nix/store/8xjprr5yi0r9j8j8b6p2nh3rn7jykdrm-neovim-0.7.0.drv'.
error: 1 dependencies of derivation '/nix/store/pmslzabl2mh96hyrmranmsnvg87s2ahv-home-manager-path.drv' failed to build
error: 1 dependencies of derivation '/nix/store/4dr1jnnlbpxsl3xq7bx963zk74vxr1rv-home-manager-generation.drv' failed to build
error: 1 dependencies of derivation '/nix/store/0ih32vl40zpck5vjjsw5vpfifg76m252-nmt-report-neovim-coc-config.drv' failed to build
```

treed:
I attempted to test this on my Mac by applying the patch over master. I did have to edit the patch to change context in the tests file (wezterm was added), but other than that it applied cleanly.

It didn't immediately work and I was unable to figure out where launchd keeps its logs. One page I found mentioned `/var/log/system.log` but nothing about this was there. I had to add the following to get any logging output:

```nix
    launchd.agents.gpg-agent.config = {
      StandardErrorPath = "/tmp/gpg-agent.err.log";
      StandardOutPath = "/tmp/gpg-agent.out.log";
    };
```
(This seems relevant to discussion I've seen elsewhere about standard logging paths.)

Stderr consists entirely of lines like the following:

```
gpg-agent[19259]: no gpg-agent running in this session
```

Which strikes me as sort of circular, but maybe someone with more familiarity with gpg-agent will know what the problem might be.

treed:
After a bit of digging around in `gpg-agent` source code and some searching, I figured out that the problem is that launchd does not automatically include the binary as arg 0: https://apple.stackexchange.com/questions/110644/getting-launchd-to-read-program-arguments-correctly

I then corrected the config to:

```nix
            ProgramArguments =
              [ "${gpgPkg}/bin/gpg-agent" "--supervised" (mkIf cfg.verbose "--verbose") ];
```

Which gave me a different error message:

```
gpg-agent[21550]: WARNING: "--supervised" is a deprecated option
gpg-agent (GnuPG) 2.3.6 starting in supervised mode.
no LISTEN_PID environment variable found in --supervised mode (ignoring)
no LISTEN_FDS or LISTEN_FDNAMES environment variables found in --supervised mode (assuming 1 active descriptor)
Fatal: file descriptor 3 must be valid in --supervised mode if LISTEN_FDNAMES is not set
```

I'm unsure how to get the socket through as those environment variables. I'm also curious what we're supposed to use other than `--supervised` if that's deprecated. It looks like they just deprecated it without replacement: https://dev.gnupg.org/rGca5d5142c6d6eaba4572a086f8473e4aebdd3f9e

treed:
I will probably leave this alone for now, but at the very least I think there's something wrong with the Socket options. I don't see anything with `/tmp/gnupg` existing while this service is being restarted, even if I watch with a while true loop to see if it's being deleted immediately as the service fails.

Also, I think there's an error with the mode, which should be converted from octal to decimal per the docstring (384 in the case of 600 octal), but changing this doesn't seem to solve anything.

montchr:
I've rebased the branch onto the latest from `master`.

I also noticed #3183 which addressed my opening note about `pinentry_mac`. That branch merges into this one cleanly. I've pushed the combined changes to a new branch on my fork which I am now tracking in my personal configs. 

https://github.com/montchr/home-manager/compare/gpg-agent-darwin...montchr:home-manager:gpg-agent-darwin-with-pinentry-mac?expand=1

https://github.com/montchr/dotfield/commit/d9ca7ac828b2b4b39ab336aed00ab093f89bb083

Commit signing has been working for me with the default pinentry flavor (though a sub-optimal UX). With #3183 in the combined branch, I'm now using `pinentry_mac`.

I haven't been using the SSH agent functionality lately, so I'll re-enable that and take a look at some of your suggestions @treed

montchr:
I've updated this to include the changes from #3183 and as discussed in that PR I've also updated the default pinentry program to `pinentry-mac` on macOS. I figured that since this change is specific to the _graphical macOS_ as opposed to the underlying Darwin system, Darwin still will use the `gtk2` default.

montchr:
I'm wondering if this logic might not work on Darwin:

https://github.com/nix-community/home-manager/blob/6acfdecd72cf81f27556a8522f8dd752e7f070a7/modules/services/gpg-agent.nix#L242-L246

FWIW nix-darwin does not do such a check, and I've confirmed that it works as expected:

https://github.com/LnL7/nix-darwin/blob/6349b99bc2b96ded34d068a88c7c5ced406b7f7f/modules/programs/gnupg.nix#L43-L49

And running this in an active shell session allows me to use `gpg-agent` for SSH in that session:

```console
home-manager on  gpg-agent-darwin
❯ export SSH_AUTH_SOCK="$(gpgconf --list-dirs agent-ssh-socket)"

home-manager on  gpg-agent-darwin
❯ ssh-add -L
ssh-rsa <stuff...> cardno:xx xxx xxx
```

But in a new shell session I get this:

```console
❯ echo $SSH_AUTH_SOCK
/private/tmp/com.apple.launchd.5gcAGrtCqp/Listeners
```

modules/services/gpg-agent.nix
```diff
@@ -9,24 +9,42 @@ let
 
   homedir = config.programs.gpg.homedir;
 
+  pinentryBinPath = optionalString cfg.pinentryFlavor != null
+    (if cfg.pinentryFlavor == "mac" then
```
montchr, line 13: I went for this specific series of conditions in an attempt to prevent pulling in any unnecessary pinentry packages. Especially in the case of `pinentry_mac` which is only supported on Darwin.

modules/services/gpg-agent.nix
```diff
@@ -256,79 +275,93 @@ in {
...
-          DirectoryMode = "0700";
+      })
+    ]))
+
+    (mkIf pkgs.stdenv.hostPlatform.isDarwin {
+      launchd.agents.gpg-agent = {
+        enable = true;
+        config = {
+          ProgramArguments = [ "${gpgPkg}/bin/gpg-connect-agent" "/bye" ];
+          RunAtLoad = cfg.enableSshSupport;
```
ratorx, line 361: Doesn't this make automatically starting gpg-agent dependent on having SSH support enabled? What if someone wanted to start automatically start GPG agent, but without SSH support (i.e. just enabled services.gpg-agent.?

Sorry if I'm missing something, but I don't see the launchd equivalent of `${gpgPkg}/bin/gpg-agent --supervised` (from the gpg-agent systemd user service) anywhere for Darwin. Is it not necessary due to the socket stuff?

sikmir:
What about [pinentry-touchid](https://github.com/jorgelbg/pinentry-touchid)? It would be really cool!

cmacrae:
Hey all 👋 Hope people don't mind me ressurecting this PR. Thanks for putting in the groundwork @montchr, nice to see you again :) I cooked up my own PR which was unknowingly, in large part, a repetition of these changes. Though you implemented some things nicer than I had!

I have a version of this that I'm using on my system (macOS Ventura 13.3, aarch64-darwin) with gpg-agent & ssh-agent functionality working flawlessly. It also considers some updates that were recently introduced to home-manager in which we no longer need to do so much control logic based on `stdenv.hostPlatform` (see #2915 ).

It appears #3840 was opened just the other day, but I don't think it addresses all the edgecases considered in changes here.

I'd love to see this accepted upstream, so what's the best path forward? A PR into your branch?

Based on their past involvement/being listed as reviewers, _courtesy ping_ for @ratorx & @rycee

modules/services/gpg-agent.nix
cmacrae: I think you're right @ratorx 👍 In my implementation, I simply have `RunAtLoad` set to `true`.

Regarding launching the agent, it appears (at least on my system, with `gpg-agent` 2.4.0), there no longer exists a `--supervised` option:
```
Options used for startup:
     --daemon                           run in daemon mode (background)
     --server                           run in server mode (foreground)
     --no-detach                        do not detach from the console
 -s, --sh                               sh-style command output
 -c, --csh                              csh-style command output
     --options FILE                     read options from FILE
```


I tried launching with what I thought was likely the closest thing: `gpg-agent --server --no-detach` (also tried without `--no-detach`), but seemingly any shell I launhced (which, with the intended & implemented configuration, will launch the agent if one isn't present) wasn't able to pick up the running gpg-agent process, so started its own one.

This is what I'm using to launch the agent, which works (and seems to be what's suggested in most of the literature I can find):
```nix
          ProgramArguments = [ "${gpgPkg}/bin/gpgconf" "--launch" "gpg-agent" ]
            ++ optionals cfg.verbose [ "--verbose" ];
```

The only caveat with this is that the process is only launched by launchd, not supervised in any manner. So, unloading the service will not terminate the disowned process, for instance.  
It doesn't seem there is an option for telling `gpgconf` to launch the `gpg-agent` component in a foreground manner.

But, it works nicely so far, in my experience :)

cmacrae:
> What about [pinentry-touchid](https://github.com/jorgelbg/pinentry-touchid) support? It would be really cool!

@sikmir sure :) If that fits your security posture, you could easily configure this module to use it with something like:
```nix
# Explicitly set to `null` as we infer "mac" if `stdenv.hostPlatform.isDarwin` is true
programs.gpg-agent.pinentryFlavor = null;
programs.gpg-agent.exraConfig = ''
  # or wherver Homebrew put pinentry-touchid
  pinentry-program /usr/local/opt/pinentry-touchid/bin/pinentry-touchid
'';
```

montchr:
Hey @cmacrae! Thanks for resurfacing this. It certainly stagnated but I've been using it in my own configs for a while since I've been super busy the past few months and dealing with the oddities of macOS can be… frustrating. So if you (or anyone else) has made some progress, a PR into this branch would be great. Though now that there's three separate PRs I'm not sure which would be the best path forward?

modules/services/gpg-agent.nix
montchr: > Doesn't this make automatically starting gpg-agent dependent on having SSH support enabled?

Yeah, I looked into this a while back and that does seem to be the case. I wasn't able to make much progress then, but it looks like @cmacrae has made a lot of progress testing out a fix. @cmacrae do you have a branch based off mine that I could test out? I thought you had opened a PR already but maybe not.

cmacrae:
@montchr Thanks for getting back :) Glad it's been working nicely for you. Understandable, we're all living our own busy lives! I've opened a PR to your branch with all the changes I've mentioned here. Once that's rebased in, we should be in a good position to get this merged. I updated the test cases and have confirmed all tests pass.

montchr:
I've just rebased onto `master` with @cmacrae's additions. I'm now tracking the latest commit in my personal configuration and git commit signing on `aarch64-darwin` is working, just as it was before.

However, some of the new tests are failing:

```console
gpg-agent-default-homedir: FAILED
For further reference please introspect /nix/store/y5lp84fzwq499bj5sl4abny9idyl6vys-nmt-report-gpg-agent-default-homedir
gpg-agent-override-homedir: FAILED
For further reference please introspect /nix/store/358ql8ibzmn6jm7qkdgg9xsggkiprval-nmt-report-gpg-agent-override-homedir
gpg-immutable-keyfiles: OK
gpg-mutable-keyfiles: OK
gpg-override-defaults: OK
```

Not quite sure what to make of the failure even after poking around in the store paths mentioned in the error output. @cmacrae any insights?

cmacrae:
> Not quite sure what to make of the failure even after poking around in the store paths mentioned in the error output. @cmacrae any insights?

Oh... weird 🤔 I ran the tests locally and they worked fine (as you can see in [dd69b63](https://github.com/nix-community/home-manager/pull/2964/commits/dd69b6314c296f63a2b5bb26fe6779d062405209#diff-af756363654efca06037e9e968de00b5004e553b35a0357b78c1d4f4e61f3394), I made changes to make sure they worked). I can take a look when I get a moment

tests/modules/services/gpg-agent/default-homedir.nix
```diff
@@ -2,7 +2,13 @@
 
 with lib;
 
-{
+let
+  inherit (pkgs.stdenv.hostPlatform) isLinux;
+
+  socketPath =
+    mkIf isLinux config.systemd.user.sockets.gpg-agent.Socket.ListenStream;
```
ncfavier, line 9:
```suggestion
    config.systemd.user.sockets.gpg-agent.Socket.ListenStream;
```

`mkIf` only makes sense within the module system (it simply wraps its value in an attribute set with the condition, see the [definition](https://github.com/NixOS/nixpkgs/blob/b55a4988418b73079e7c50ee574545a9c956ee0b/lib/modules.nix#L912-L915)). Since there is already an `optionalString` around the usage of `socketPath` this is entirely unnecessary.

modules/services/gpg-agent.nix
```diff
@@ -3,12 +3,17 @@
 with lib;
 
 let
-
   cfg = config.services.gpg-agent;
   gpgPkg = config.programs.gpg.package;
 
   homedir = config.programs.gpg.homedir;
 
+  pinentryBinPath = optionalString (cfg.pinentryFlavor != null)
```
ncfavier, line 11:
```suggestion
  pinentryBinPath =
```

This is redundant since there's already an `optionalString` guarding the use of `pinentryBinPath`. (And remove the parentheses)

modules/services/gpg-agent.nix
```diff
@@ -256,17 +278,26 @@ in {
...
+        enable = true;
+        config = {
+          ProgramArguments = [ "${gpgPkg}/bin/gpgconf" "--launch" "gpg-agent" ]
+            ++ optionals cfg.verbose [ "--verbose" ];
+          RunAtLoad = true;
+          KeepAlive.SuccessfulExit = false;
+          EnvironmentVariables.GNUPGHOME = homedir;
+        };
+      };
+    }
```
ncfavier, line 292: Merge this module with the next one?

tests/modules/services/gpg-agent/default-homedir.nix
```diff
@@ -11,8 +15,9 @@ with lib;
     test.stubs.gnupg = { };
     test.stubs.systemd = { }; # depends on gnupg.override
 
-    nmt.script = ''
-      in="${config.systemd.user.sockets.gpg-agent.Socket.ListenStream}"
+    nmt.script = optionalString isLinux ''
+      in="${socketPath}"
```
ncfavier, line 19: `in` is not set on darwin, so obviously the test fails on darwin?

jaudiger:
I was looking for gpg-agent on darwin, and found this PR. I just wanted to express my interest for this ongoing work 🌟

folliehiyuki:
#5786 was merged into master, should this PR be closed?

PR #2965, opened 2022-05-17
vscode: add options to disable update checks

wamserma:
### Description

Update notification popups are annoying when vscode/vscodium is
managed by home-manager. However, as these settings also require
the config to be managed via userSettings, they are disabled by
default.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/vscode.nix
+++ b/modules/programs/vscode.nix
@@ -34,6 +34,11 @@ let
   # TODO: On Darwin where are the extensions?
   extensionPath = ".${extensionDir}/extensions";
 
+  mergedUserSettings = cfg.userSettings
+    // optionalAttrs (!cfg.enableUpdateCheck) { "update.mode" = "none"; }
+    // optionalAttrs (!cfg.enableExtensionUpdateCheck) {
+      "extensions.autoCheckUpdates" = false;
+    };
 in {
   imports = [
     (mkChangedOptionModule [ "programs" "vscode" "immutableExtensionsDir" ] [
@@ -56,12 +61,28 @@ in {
         '';
       };
 
+      enableUpdateCheck = mkOption {
+        type = types.bool;
+        default = true;
+        description = ''
+          Whether to enable update checks/notifications.
+        '';
+      };
+
+      enableExtensionUpdateCheck = mkOption {
+        type = types.bool;
+        default = true;
+        description = ''
+          Whether to enable update notifications for extensions.
+        '';
+      };
+
       userSettings = mkOption {
         type = jsonFormat.type;
         default = { };
         example = literalExpression ''
           {
-            "update.channel" = "none";
+            "files.autoSave" = "off"
             "[nix]"."editor.tabSize" = 2;
           }
         '';
@@ -164,9 +185,9 @@ in {
     home.packages = [ cfg.package ];
 
     home.file = mkMerge [
-      (mkIf (cfg.userSettings != { }) {
+      (mkIf (mergedUserSettings != { }) {
         "${configFilePath}".source =
-          jsonFormat.generate "vscode-user-settings" cfg.userSettings;
+          jsonFormat.generate "vscode-user-settings" mergedUserSettings;
       })
       (mkIf (cfg.userTasks != { }) {
         "${tasksFilePath}".source =
--- a/tests/modules/programs/vscode/default.nix
+++ b/tests/modules/programs/vscode/default.nix
@@ -1,1 +1,4 @@
-{ vscode-keybindings = ./keybindings.nix; }
+{
+  vscode-keybindings = ./keybindings.nix;
+  vscode-update-checks = ./update-checks.nix;
+}
--- a/tests/modules/programs/vscode/keybindings.nix
+++ b/tests/modules/programs/vscode/keybindings.nix
@@ -27,12 +27,17 @@ let
     }
   ];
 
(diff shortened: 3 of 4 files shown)


teto (approved):

modules/programs/vscode.nix
```diff
@@ -56,12 +56,29 @@ in {
...
+      };
+
       userSettings = mkOption {
         type = jsonFormat.type;
         default = { };
         example = literalExpression ''
           {
-            "update.channel" = "none";
+            "update.mode" = "none"
+            "extensions.autoCheckUpdates" = false
```
sumnerevans, line 81:
```suggestion
```

This doesn't make a whole lot of sense to include here since we are encouraging people to use the other options.

modules/programs/vscode.nix
```diff
@@ -164,10 +181,16 @@ in {
...
+      (mkIf (cfg.userSettings != { } || cfg.disableUpdateCheck
+        || cfg.disableExtensionUpdateCheck) {
+          "${configFilePath}".source =
+            jsonFormat.generate "vscode-user-settings" (cfg.userSettings
+              // optionalAttrs (cfg.disableUpdateCheck) {
+                "update.mode" = "none";
+              } // optionalAttrs (cfg.disableExtensionUpdateCheck) {
+                "extensions.autoCheckUpdates" = false;
+              });
+        })
```
sumnerevans, line 193: It may be cleaner to merge `cfg.userSettings` and the two optional attr sets and then check if that is `{ }`. Then, just generate the whole config.

Something like:

```nix
let computedUserSettings = cfg.userSettings
  // optionalAttrs (cfg.disableUpdateCheck) {
    # ...
  }
  // optionalAttrs (cfg.disableExtensionUpdateCheck) {
    # ...
  };
in mkIf (computedUserSettings != { }) {
  # ...
}
```

modules/programs/vscode.nix
```diff
@@ -56,12 +56,29 @@ in {
         '';
       };
 
+      disableUpdateCheck = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          Enable to silence update popups. Updates userSettings.
+        '';
+      };
```
sumnerevans, line 65: I think it's better to have options be in the positive direction.

```suggestion
      enableUpdateCheck = mkEnableOption "update popups";
```

modules/programs/vscode.nix
```diff
@@ -56,12 +56,29 @@ in {
...
+
+      disableExtensionUpdateCheck = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          Enable to silence update notifications for extensions. Updates userSettings.
+        '';
+      };
```
sumnerevans, line 73:
```suggestion
      enableExtensionUpdateCheck = mkEnableOption "update notifications for extensions";
```

sumnerevans (changes requested):

modules/programs/vscode.nix
wamserma: I agree.

modules/programs/vscode.nix
wamserma: Yeah, replaced with a different example for a string arg.

wamserma:
ping @rycee

SuperSandro2000:
The update check should be disabled by default for installations managed by nix.

wamserma:
> The update check should be disabled by default for installations managed by nix.

At least for codium, I see update notifications.

modules/programs/vscode.nix
```diff
@@ -56,12 +61,28 @@ in {
         '';
       };
 
+      disableUpdateCheck = mkOption {
+        type = types.bool;
+        default = false;
```
SuperSandro2000, line 66:
```suggestion
        default = true;
```
I mean this should default to true

wamserma: As commented [here](https://github.com/nix-community/home-manager/pull/2965#discussion_r886076681), this would generate a config and subsequently a conflict with an existing `User/settings.json` for people that don't manage their vscode config via HM.

modules/programs/vscode.nix
```diff
@@ -34,6 +34,11 @@ let
   # TODO: On Darwin where are the extensions?
   extensionPath = ".${extensionDir}/extensions";
 
+  mergedUserSettings = cfg.userSettings
+    // optionalAttrs (cfg.disableUpdateCheck) { "update.mode" = "none"; }
+    // optionalAttrs (cfg.disableExtensionUpdateCheck) {
+      "extensions.autoCheckUpdates" = false;
```
berbiche, line 40: IMO it would be better to set these options in the config section:

```nix
{
  config.programs.vscode.userSettings = {
    "update.mode" = mkIf cfg.disableUpdateCheck "none";
    "extensions.autoCheckUpdates" = mkIf cfg.disableExtensionUpdateCheck false;
  };
}
```

berbiche (changes requested):
Small nit 🙂

modules/programs/vscode.nix
wamserma: I think this might lead to undesired results when some userSettings are passed in with an existing config and this module is expected to update only the relevant settings.

rycee:
Thanks! Merged to master now. Note, I changed "disable" to "enable" to avoid the negative logic. So the options are now `enableUpdateCheck` and `enableExtensionUpdateCheck`.

PR #2969, opened 2022-05-19
pass-secret-service: Add Tests, Remove Dependency on password-store Module

### LINKED ISSUE #2962: bug: pass-secret-service unnecessarily depends on password-store

### Issue description

The nix expression requires that `programs.password-store.enable` be `true`, basing the loaded password store directory on the environment variable configured within the `password-store` module. This is unnecessary, as [pass_secret_service uses pypass](https://github.com/mdellweg/pass_secret_service/blob/develop/setup.py), meaning the dependency is unnecessary, as well as potentially problematic if one wishes to use a different compliant pass tool, such as pypass or gopass, or use the service without installing pass itself. This directory should be configurable separately.

It may indeed be best to have it use a default of the environment variable `PASSWORD_STORE_DIR` or the one given in the `password-store` module if enabled, but it should still be configurable.

### Maintainer CC

@cab404

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 5.17.7, NixOS, 21.11 (Porcupine)`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.4`
 - channels(root): `"nixos"`
 - channels(david): `""`
```

houstdav000:
### Description

Allow setting the application package and storePath used by the config.
Since the programs.password-store home-manager module sets config values
via global environment variables, the default behavior of the module
should continue to behave as before for the user.

Additionally, performs some cleanup and adds a few tests that helped me
catch some bugs.

Closes #2962 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/pass-secret-service.nix
+++ b/modules/services/pass-secret-service.nix
@@ -2,31 +2,46 @@
 
 with lib;
 
-let serviceCfg = config.services.pass-secret-service;
+let
+  cfg = config.services.pass-secret-service;
+
+  serviceArgs =
+    optionalString (cfg.storePath != null) "--path ${cfg.storePath}";
 in {
-  meta.maintainers = [ maintainers.cab404 ];
+  meta.maintainers = with maintainers; [ cab404 houstdav000 ];
+
   options.services.pass-secret-service = {
     enable = mkEnableOption "Pass libsecret service";
+
+    package = mkPackageOption pkgs "pass-secret-service" { };
+
+    storePath = mkOption {
+      type = with types; nullOr str;
+      default = null;
+      defaultText = "~/.password-store";
+      example = "/home/user/.local/share/password-store";
+      description = "Absolute path to password store.";
+    };
   };
-  config = mkIf serviceCfg.enable {
+
+  config = mkIf cfg.enable {
     assertions = [
       (hm.assertions.assertPlatform "services.pass-secret-service" pkgs
         platforms.linux)
-
-      {
-        assertion = config.programs.password-store.enable;
-        message = "The 'services.pass-secret-service' module requires"
-          + " 'programs.password-store.enable = true'.";
-      }
     ];
 
     systemd.user.services.pass-secret-service = {
-      Unit = { Description = "Pass libsecret service"; };
-      Service = {
-        # pass-secret-service doesn't use environment variables for some reason.
-        ExecStart =
-          "${pkgs.pass-secret-service}/bin/pass_secret_service --path ${config.programs.password-store.settings.PASSWORD_STORE_DIR}";
+      Unit = {
+        AssertFileIsExecutable = "${cfg.package}/bin/pass_secret_service";
+        Description = "Pass libsecret service";
+        Documentation = "https://github.com/mdellweg/pass_secret_service";
+        PartOf = [ "default.target" ];
       };
+
+      Service = {
+        ExecStart = "${cfg.package}/bin/pass_secret_service ${serviceArgs}";
+      };
+
       Install = { WantedBy = [ "default.target" ]; };
     };
   };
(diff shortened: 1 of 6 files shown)


houstdav000:
Tests for code changed works, but test `xdg-desktop-entries` fails on my machine.

houstdav000:
Added a missing "m" to "comand" to make the manual build.

modules/services/pass-secret-service.nix
```diff
@@ -2,31 +2,47 @@
...
+      Unit = {
+        Description = "Pass libsecret service";
+        PartOf = [ "default.target" ];
+      };
+
       Service = {
         # pass-secret-service doesn't use environment variables for some reason.
         ExecStart =
-          "${pkgs.pass-secret-service}/bin/pass_secret_service --path ${config.programs.password-store.settings.PASSWORD_STORE_DIR}";
+          "${cfg.package}/bin/pass_secret_service --path ${cfg.storePath}";
```
rycee, line 43: Can take the opportunity to do

```suggestion
          "${cfg.package}/bin/pass_secret_service --path ${escapeShellArg cfg.storePath}";
```

so that, e.g., spaces in the path won't cause problems.

modules/services/pass-secret-service.nix
```diff
@@ -2,31 +2,47 @@
...
+      type = types.package;
+      default = pkgs.pass-secret-service;
+      defaultText = literalExpression "pkgs.pass-secret-service";
+      description = "Package providing <command>pass_secret_service</command>.";
+    };
+
+    storePath = mkOption {
+      type = types.str;
+      default = "$PASSWORD_STORE_DIR";
+      defaultText = "$PASSWORD_STORE_DIR";
```
rycee, line 22: No need to put a `defaultText` attribute since the value is a plain string.
```suggestion
```

modules/services/pass-secret-service.nix
```diff
@@ -2,31 +2,47 @@
...
+    package = mkOption {
+      type = types.package;
+      default = pkgs.pass-secret-service;
+      defaultText = literalExpression "pkgs.pass-secret-service";
+      description = "Package providing <command>pass_secret_service</command>.";
+    };
+
+    storePath = mkOption {
+      type = types.str;
+      default = "$PASSWORD_STORE_DIR";
```
rycee, line 21: Have you verified that this works? I suspect that it wouldn't work since the `PASSWORD_STORE_DIR` environment variable is not imported into the systemd session unless the X session is managed by HM. Even then it won't be imported until the graphical sessiont target is starting, i.e., long after the default target.

I think, instead, that it would make sense to omit the default value for the `storePath` variable. In the `password-store.nix` file add

    services.pass-secret-service.storePath = mkDefault cfg.settings.PASSWORD_STORE_DIR;

Then a user that has password-store enabled should get the correct path and other users will get an error unless they set a path themselves.

modules/services/pass-secret-service.nix
houstdav000: Fixed in c211c509

modules/services/pass-secret-service.nix
houstdav000: Added in 80e43c10

modules/services/pass-secret-service.nix
houstdav000: Propagating the value in `password-store.nix` makes sense, though I wonder if it's good for `password-store.nix` to be setting its own default, since pass itself defaults to `~/.password-store`, and that's why we are forced to set the directory for pass_secret_service with `--path` regardless of the environment.

Either way, I'll definitely implement something else than what I have, like

```Nix
services.pass-secret-service.storePath = mkDefault cfg.settings.PASSWORD_STORE_DIR;
```

and have the `--path` arg only be added if `storePath` is not null or something.

houstdav000:
@rycee Any further requests? This PR is still awaiting review

rycee:
Thanks! Merged to master now.

PR #2970, opened 2022-05-19
mopidy: add support for extraConfigFiles

lilyinstarlight:
### Description

Follow-up to #2656:

I use `extraConfigFiles` to avoid storing secrets in the Nix store with sops-nix, and this config option (or a functionally similar one that allows importing files at runtime) is required for that to work. See code at https://github.com/lilyinstarlight/foosteros/blob/08fb35ccff35af18d6436149002665ca703858ba/hosts/bina/configuration.nix#L551-L553 in which I've been using `extraConfigFiles` for an example

cc: @foo-dogsquared

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/mopidy.nix
+++ b/modules/services/mopidy.nix
@@ -45,6 +45,9 @@ let
 
   settingsFormat = mopidyConfFormat { };
 
+  configFilePaths = concatStringsSep ":"
+    ([ "${config.xdg.configHome}/mopidy/mopidy.conf" ] ++ cfg.extraConfigFiles);
+
 in {
   meta.maintainers = [ hm.maintainers.foo-dogsquared ];
 
@@ -97,6 +100,16 @@ in {
         more details.
       '';
     };
+
+    extraConfigFiles = mkOption {
+      default = [ ];
+      type = types.listOf types.path;
+      description = ''
+        Extra configuration files read by Mopidy when the service starts.
+        Later files in the list override earlier configuration files and
+        structured settings.
+      '';
+    };
   };
 
   config = mkIf cfg.enable {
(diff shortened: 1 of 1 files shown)


modules/services/mopidy.nix
```diff
@@ -97,6 +97,16 @@ in {
         more details.
       '';
     };
+
+    extraConfigFiles = mkOption {
+      default = [ ];
+      type = types.listOf types.str;
+      description = ''
+        Extra config file read by Mopidy when the service starts.
```
rycee, line 105:
```suggestion
        Extra configuration files read by Mopidy when the service starts.
```

modules/services/mopidy.nix
```diff
@@ -97,6 +97,16 @@ in {
         more details.
       '';
     };
+
+    extraConfigFiles = mkOption {
+      default = [ ];
+      type = types.listOf types.str;
```
rycee, line 103: I think you always want an absolute path, so can do a bit better than `str`:
```suggestion
      type = types.listOf types.path;
```

modules/services/mopidy.nix
```diff
@@ -113,7 +123,13 @@ in {
         After = [ "network.target" "sound.target" ];
       };
 
-      Service = { ExecStart = "${mopidyEnv}/bin/mopidy"; };
+      Service = {
+        ExecStart = "${mopidyEnv}/bin/mopidy --config ${
+            concatStringsSep ":"
+            ([ "${config.xdg.configHome}/mopidy/mopidy.conf" ]
+              ++ cfg.extraConfigFiles)
```
rycee, line 130: Can extract to a variable (inside the `let` bindings at the top), then use the variable both here and below.

lilyinstarlight: Should I extract the entire `concatStringsSep` call or only extract `[ "${config.xdg.configHome}/mopidy/mopidy.conf" ] ++ cfg.extraConfigFiles`? Thank you for the review!

rycee: I would say the entire thing, so that the command in both places contains simply something like `--config ${configFilePaths}`.

lilyinstarlight:
@rycee, all of the review comments should be addressed now and I made an additional grammatical fix to the description (overrides -> override). Let me know if anything else needs to be fixed!

rycee:
Thanks! Merged now.

Issue #2297, opened 2021-08-26
Enable searching Home Manager options from search.nixos.org


berbiche:
### Description

With the merge of this PR: <https://github.com/NixOS/nixos-search/pull/324>, it seems like it would be possible to search Home Manager options from <https://search.nixos.org/flakes>.

What would the Home Manager project need to do @ysndr @ncfavier :smiley:?

edit: link: https://github.com/NixOS/nixos-search/issues/323

ncfavier:
Would love to see this. I think the main problem is that there isn't an official flake output for "modules that are not nixos- or darwin-specific", like home-manager modules. Maybe we could just start using `modules` unofficially though.

ysndr:
Exactly, the main problem right now is that the modules hm provides are not publicated by the flake
There’s the integration with the nixos configuration but that’s all as far as I can see.
I don’t know if there’s big hurdles to make hm’s config nixos compatible (it shouldn’t as it already integrates) and make them publicly visible as a module in the flake

terlar:
Still relevant. Docs is already part of the flake now. Not sure if it is in the correct format to be picked up though.

ncfavier:
For the record I just looked into this a bit, and realised that exposing the modules is not enough: we would also need to expose the `lib` that HM modules expect (`stdlib-extended.nix`) in order to be able to evaluate them.

Another option might be to expose *options* (the argument to `lib.optionAttrSetToDocList`, probably) directly in the flake rather than modules. That way nixos-search doesn't need to know about the specifics of how HM modules expect to be evaluated (and we might imagine adding support for other "module scopes" in the future).

On the nixos-search side, it should be rather straightforward to add a `scope` field to options (like `nixos`, `home-manager`, etc.) and display different scopes in separate tabs.

eveeifyeve:
@rycee this issue can be closed.

teto:
home-manager is now available as a source eg: https://search.nixos.org/options?channel=unstable&query=neovim&source=home_manager

PR #1934, opened 2021-04-16
Add flake.lock & workflow to update it

i077:
### Description

This PR implements suggestions from #1913, adding a flake.lock file so we can `nix run github:nix-community/home-manager` without any hiccups. The nixpkgs input is currently set to track `nixos-unstable`. If you want to test this out, you can run `nix run github:nix-community/home-manager/pull/1934/head`, which should give you the help text.

To keep this file up to date, the PR also adds a GitHub workflow that
- Runs every Sunday and Wednesday at 03:51 (I just picked a random time)
- Runs `nix flake update --commit-lock-file`, committing as a nonexistent user called `flakebot`
- Pushes this commit ~~directly to master~~ to a new branch called `flake-update` with a custom token `FLAKEBOT_TOKEN` (this needs to be added to the repo)
- Opens a PR (or updates an existing PR) to merge this commit to master. This is done with the same custom token so that other workflows can be run against the PR.

Also updates the README to change the section on a flake-based config a bit.

I also wanted to modify the testing workflow to actually run home-manager's tests, but was having a bit of trouble exposing the tests under the `checks` output of the flake, so I will leave that out for now. ~~This means that the update workflow will just regularly push updates to master, whether they pass tests or not.~~

cc @berbiche since we were talking about this in the other PR.

### Checklist

- [x] Change is backwards compatible.

- [x] Code formatted with `./format`.

- [ ] Code tested through `nix-shell --pure tests -A run.all`.

- [ ] Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).

- [ ] Commit messages are formatted like

    ```
    {component}: {description}

    {long description}
    ```

    See [CONTRIBUTING](https://github.com/nix-community/home-manager/blob/master/doc/contributing.adoc#sec-commit-style) for more information and [recent commit messages](https://github.com/nix-community/home-manager/commits/master) for examples.

- If this PR adds a new module

  - [ ] Added myself as module maintainer. See [example](https://github.com/nix-community/home-manager/blob/068ff76a10e95820f886ac46957edcff4e44621d/modules/programs/lesspipe.nix#L6).

  - [ ] Added myself and the module files to `.github/CODEOWNERS`.

PR #2971, opened 2022-05-20
Prepare inclusion in nixos-search

ncfavier:
### Description

Lays the groundwork for https://github.com/NixOS/nixos-search/issues/323 by making
```
nix run nixos-search#flake-info -- flake github:nix-community/home-manager
```
succeed. Note that this PR does **NOT** solve the [problem](https://github.com/nix-community/home-manager/issues/2297#issuecomment-1079698898) of how to expose Home Manager modules/options in the flake, it only ensures that we can index packages and NixOS options.

Best reviewed one commit at a time.

This PR partly supersedes https://github.com/nix-community/home-manager/pull/1934, although it does not add an auto-update workflow.

Tested the NixOS module and the manual. nix-darwin testers welcome.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,6 +1,16 @@
 { pkgs ? import <nixpkgs> { } }:
 
-rec {
+let
+
+  flake = (import
+    (let lock = builtins.fromJSON (builtins.readFile ./flake.lock);
+    in fetchTarball {
+      url =
+        "https://github.com/edolstra/flake-compat/archive/${lock.nodes.flake-compat.locked.rev}.tar.gz";
+      sha256 = lock.nodes.flake-compat.locked.narHash;
+    }) { src = ./.; }).defaultNix;
+
+in rec {
   docs = with import ./docs { inherit pkgs; }; {
     html = manual.html;
     manPages = manPages;
@@ -16,4 +26,6 @@ rec {
   nixos = import ./nixos;
 
   path = ./.;
+
+  inherit (flake) inputs;
 }
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,43 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1654230545,
+        "narHash": "sha256-8Vlwf0x8ow6pPOK2a04bT+pxIeRnM1+O0Xv9/CuDzRs=",
+        "owner": "nixos",
+        "repo": "nixpkgs",
+        "rev": "236cc2971ac72acd90f0ae3a797f9f83098b17ec",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nixos",
+        "ref": "nixos-unstable",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "nixpkgs": "nixpkgs",
+        "utils": "utils"
+      }
+    },
+    "utils": {
+      "locked": {
+        "lastModified": 1653893745,
+        "narHash": "sha256-0jntwV3Z8//YwuOjzhV2sgJJPt+HY6KhU7VZUL0fKZQ=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "1ed9fb1935d260de5fe1c2f7ee0ebaae17ed2fa1",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    }
+  },
+  "root": "root",
+  "version": 7
+}
--- a/flake.lock
+++ b/flake.lock
@@ -1,5 +1,21 @@
 {
   "nodes": {
+    "flake-compat": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1650374568,
+        "narHash": "sha256-Z+s0J8/r907g149rllvwhb4pKi8Wam5ij0st8PwAh+E=",
+        "owner": "edolstra",
+        "repo": "flake-compat",
+        "rev": "b4a34015c698c7793d592d66adbab377907a2be8",
+        "type": "github"
+      },
+      "original": {
+        "owner": "edolstra",
+        "repo": "flake-compat",
+        "type": "github"
+      }
+    },
     "nixpkgs": {
       "locked": {
         "lastModified": 1654230545,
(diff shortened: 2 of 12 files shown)


teto:
I am happy to see this initiative but why mixing it with format changes ? can these go in another MR ?

Is the following expected ?
```
$ nix run nixos-search#flake-info -- flake github:ncfavier/home-manager/nixos-searchable
warning: Using saved setting for 'extra-substituters = https://nixos-search.cachix.org' from ~/.local/share/nix/trusted-settings.json.
warning: Using saved setting for 'extra-trusted-public-keys = nixos-search.cachix.org-1:1HV3YF8az4fywnH+pAd+CXFEdpTXtv9WpoivPi+H70o=' from ~/.local/share/nix/trusted-settings.json.
[2022-06-05T10:03:11Z ERROR command_run] command 'nix eval --json --no-write-lock-file -f /tmp/.tmpRoip4S/extract.nix --override-flake input-flake github:ncfavier/home-manager/nixos-searchable --argstr flake github:ncfavier/home-manager/nixos-searchable all' failed: exit status: 1
    stdout:
    
    stderr:
    warning: Nix search path entry '/nix/var/nix/profiles/per-user/root/channels/nixos' does not exist, ignoring
    error: file 'nixpkgs' was not found in the Nix search path (add it using $NIX_PATH or -I)
    
           at /tmp/.tmpRoip4S/extract.nix:5:21:
    
                4|
                5|   nixpkgs = (import <nixpkgs> {});
                 |                     ^
                6|   lib = nixpkgs.lib;
    
Error: Getting flake info caused an error: Failed to gather information about github:ncfavier/home-manager/nixos-searchable

Caused by:
    command 'nix eval --json --no-write-lock-file -f /tmp/.tmpRoip4S/extract.nix --override-flake input-flake github:ncfavier/home-manager/nixos-searchable --argstr flake github:ncfavier/home-manager/nixos-searchable all' failed: exit status: 1
```

ncfavier:
> I am happy to see this initiative but why mixing it with format changes ? can these go in another MR ?

I suppose they can, but does it matter? This is just a big clean-up PR.

> Is the following expected ?

Nope, that's a bug in nixos-search. In the meantime you can do `export NIX_PATH=nixpkgs=channel:nixpkgs-unstable`.

teto:
arf nmd issue: 
```
    error: a 'aarch64-darwin' with features {} is required to build '/nix/store/rsiqrib1rv2shc6azv12xcbkvxs9hj32-nmd.drv', but I am a 'x86_64-linux' with features {benchmark, big-parallel, kvm, nixos-test}
```

ncfavier:
Ah, I guess I didn't run into that because that path was substituted by the nix-community cache.

So we should probably add nmd and nmt to the flake inputs (which I did initially but then I thought it wasn't needed).

ncfavier:
Should be fixed now

flake.nix
```diff
@@ -1,43 +1,31 @@
 {
   description = "Home Manager for Nix";
 
-  outputs = { self, nixpkgs }:
-    let
-      # List of systems supported by home-manager binary
-      supportedSystems = with nixpkgs.lib.platforms; linux ++ darwin;
+  inputs.nixpkgs.url = "nixpkgs/nixos-unstable";
```
teto, line 4: this seems to work but I would rather see "github:nixpkgs/nixos-unstable" that is less magical; same for flake-utils

ncfavier:
Thanks for the quick merge!

@rycee quick heads-up since I don't know how familiar you are with flakes: inputs like nmd and nmt can now be updated with `nix flake lock --update-input nmd` (you'll need to enable the `flakes` and `nix-command` experimental features). But maybe we want to automate this as part of https://github.com/nix-community/home-manager/pull/1934?

default.nix
```diff
@@ -1,6 +1,16 @@
...
-rec {
+let
+
+  flake = (import
+    (let lock = builtins.fromJSON (builtins.readFile ./flake.lock);
+    in fetchTarball {
+      url =
+        "https://github.com/edolstra/flake-compat/archive/${lock.nodes.flake-compat.locked.rev}.tar.gz";
+      sha256 = lock.nodes.flake-compat.locked.narHash;
+    }) { src = ./.; }).defaultNix;
```
rycee, line 11: This feels pretty ugly to me. Why not simply change the `pkgs.fetchFromGitLab` to `fetchTarball` when fetching nmd and nmt?

flake.lock
```diff
@@ -0,0 +1,94 @@
...
+        "rev": "236cc2971ac72acd90f0ae3a797f9f83098b17ec",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nixos",
+        "ref": "nixos-unstable",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
```
rycee, line 34: I want it to be explicit to the user which Nixpkgs they are using in their configuration. I.e., HM should never be responsible for pinning Nixpkgs on behalf of users. So if we are pinning Nixpkgs here wouldn't we also need to change

    pkgs ? builtins.getAttr system nixpkgs.outputs.legacyPackages

to simply

    pkgs

in the `homeManagerConfiguration` function?

Ideally, perhaps disallowing

without an accompanying

    home-manager.inputs.nixpkgs.follows = "nixpkgs";

would be most reassuring but I don't know if that is possible.

integration-common.nix
```diff
@@ -0,0 +1,124 @@
+# This module is the common base for the NixOS and nix-darwin modules.
```
rycee, line 1: This should be moved out of the project root directory. For example, you can put it in the nixos directory instead.

rycee (commented):
I don't know much about Flakes so I'll defer to @teto's judgement about merging. I just added a few minor comments that I didn't see discussed, probably because they are too blindingly obvious for people who aren't ignorant about Flakes 🙂

default.nix
ncfavier: This is only a compatibility layer that should progressively become useless as we transition to flakes and the new nix CLI.

> Why not simply change the `pkgs.fetchFromGitLab` to `fetchTarball` when fetching nmd and nmt?

We could, but if we're going to use flakes, we might as well use them fully, i.e. for all inputs. I think they're a nicer way to manage dependencies, and they make updating and overriding easier.

flake.lock
ncfavier: > So if we are pinning Nixpkgs here wouldn't we also need to change

Agreed, I'll do that in a follow-up PR.

> Ideally, perhaps disallowing

That's not possible, and I think it would be overkill: we only need to ensure that the `nixpkgs` input is not passed down to the HM modules, but we may (and should) use the locked `nixpkgs` for flake packages.

integration-common.nix
ncfavier: Will do in follow-up.

ncfavier:
https://github.com/nix-community/home-manager/pull/3028

default.nix
rycee: I agree that eventually it will be worth switching over to flakes but I'm skeptical about delving too deep while the feature is marked as experimental. At least unless there is a really compelling reason. But in this case there is a straight forward way to avoid the pkgs reference without touching flakes and needing compatibility shims.

I also would like to move the nmt and nmd projects over to sourcehut at some point and it doesn't seem like the flake-compat project supports the sourcehut URLs. I'll probably maintain a mirror at the gitlab location so this is not a blocker, though.

ncfavier: It's also not a blocker because we can just use URLs like `git+https://git.sr.ht/~rycee/nmt`.

wmertens:
so what is now the way to pass nixpkgs? It is not clear to me at all. I tried passing `pkgs = inputs.nixpkgs.outputs.legacyPackages` but then `lib` is missing.

The docs for flakes haven't been updated yet.

default.nix
SuperSandro2000: > I'm skeptical about delving too deep while the feature is marked as experimental.

It is only marked as experimental to not have to create backwards compatibility while it still might change.

ncfavier:
`nixpkgs.outputs.legacyPackages.${system}`

PR #2974, opened 2022-05-22
bspwm: add missing rule setting `rectangle`

ncfavier:
Also add a `freeformType` so that we don't have to do this in the future.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


berbiche (approved):
Thanks for the PR!

PR #2979, opened 2022-05-24
nushell: update to reflect new settings style

schuelermine:
### Description

Updated the nushell module to reflect the current nushell configuration style.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


tests/modules/services/window-managers/sway/sway-no-xwayland.nix
```diff
@@ -17,7 +17,7 @@
         ${
           pkgs.writeText "expected" ''
             xwayland disable
-                    ''
```
sumnerevans, line 20: please remove this change. It seems entirely unrelated

sumnerevans (changes requested):

modules/programs/nushell.nix
```diff
@@ -21,7 +29,27 @@ in {
       description = "The package to use for nushell.";
     };
 
-    settings = mkOption {
+    config = mkOption {
+      type = types.lines;
+      default = "";
+      defaultText = literalExpression ''""'';
```
sumnerevans, line 35: I don't think `defaultText` isn't necessary for default-empty strings (at least I can't find another instance of that in the repo.

rycee:
Seems like a duplicate of #2957.

rycee:
I believe this is fixed since the merge of the PR mentioned previously.

PR #2982, opened 2022-05-27
getmail: Use newer getmail6 over getmail package

romanofski:
The `getmail` package will soon be removed from nixpkgs. The
`nixos-unstable` channel already has it removed and using the service
will result in:

    error: getmail has been removed from nixpkgs, migrate to getmail6

Upgrade to the getmail6 package which is already available and backwards
compatible.

### DIFF
--- a/modules/services/getmail.nix
+++ b/modules/services/getmail.nix
@@ -44,7 +44,7 @@ in {
 
     systemd.user.services.getmail = {
       Unit = { Description = "getmail email fetcher"; };
-      Service = { ExecStart = "${pkgs.getmail}/bin/getmail ${configFiles}"; };
+      Service = { ExecStart = "${pkgs.getmail6}/bin/getmail ${configFiles}"; };
     };
 
     systemd.user.timers.getmail = {


teto:
could you amend and repush to retrigger the ci ? getmail was removed in https://github.com/NixOS/nixpkgs/pull/154697/files#diff-44c521cada1a7bc163a0f625ac1bf8326cb83e5b302536cdc0f6874163186670

modules/services/getmail.nix
```diff
@@ -44,7 +44,7 @@ in {
 
     systemd.user.services.getmail = {
       Unit = { Description = "getmail email fetcher"; };
-      Service = { ExecStart = "${pkgs.getmail}/bin/getmail ${configFiles}"; };
+      Service = { ExecStart = "${pkgs.getmail6}/bin/getmail ${configFiles}"; };
```
berbiche, line 47:
```suggestion
      Service = { ExecStart = "${pkgs.getmail6 or pkgs.getmail}/bin/getmail ${configFiles}"; };
```

If the package isn't available in stable nixpkgs then I recommend the patch above.

teto: this is a bit confusing and makes troubleshooting harder I think. I would rather merge as is. HM master is supposed to target unstable anyway

romanofski: Yeah I was wondering whether we should not better make it explicit (`assert`?) that `getmail` is deprecated and the user has to upgrade. Fine with me either way.

teto: I think the nix error is clear enough. Let's not dwell on this.

rycee:
@romanofski When opening the PR did you get a template message with a checklist of things to do? If you did and intentionally removed it then please don't do that. It is there to help us review PRs and to avoid inconsistencies in the project. It is fine to disagree with the project contribution guidelines but please raise your concerns publically instead of surreptitiously removing the PR checklist.

romanofski:
> @romanofski When opening the PR did you get a template message with a checklist of things to do? If you did and intentionally removed it then please don't do that. It is there to help us review PRs and to avoid inconsistencies in the project. It is fine to disagree with the project contribution guidelines but please raise your concerns publically instead of surreptitiously removing the PR checklist.

@rycee I'm very sorry. I did not mean to disagree with any of the guidelines. I presume I removed the checklist since I thought it was a one line change. Again I'm very sorry if it promoted bad behaviour or disagreement. I will not remove the template when filing a new PR.

PR #2983, opened 2022-05-27
doc: make documentation links more visible

kubukoz:
### Description

Making the documentation links stand out more - previously it was easy to miss them, because they were only single words. Also adding a Usage section to highlight them even more - I can't begin to tell you how many times I've googled "home-manager configuration" or ended up on the URL https://rycee.gitlab.io/home-manager/options.html instead of https://nix-community.github.io/home-manager/options.html .

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,11 +6,16 @@ using the [Nix][] package manager together with the Nix libraries
 found in [Nixpkgs][]. It allows declarative configuration of user
 specific (non global) packages and dotfiles.
 
+Usage
+-----
+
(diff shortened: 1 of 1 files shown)


kubukoz:
@rycee please take a look :)

rycee:
Thanks, I think this change makes good sense. Merged to master now.

Btw, the pages at https://rycee.gitlab.io/home-manager/ and https://nix-community.github.io/home-manager/ should be identical so no need to worry about ending up on one or the other 🙂

PR #2984, opened 2022-05-29
firefox: support setting search engine & adding custom engines

### LINKED ISSUE #2318: Support search.json.mozlz4 customization
This would allow us to configure our search engine lists.

See https://github.com/soufianesakhi/firefox-search-engines-helper

kira-bruneau:
### Description

With this change, it's now possible to configure the default search engine in Firefox with `programs.firefox.profiles.<name>.search.default` and add custom engines with `programs.firefox.profiles.<name>.search.engines`.

It's also recommended to enable `programs.firefox.profiles.<name>.search.force = true` since Firefox
will replace the symlink for the search configuration on every launch, but note that you'll loose any existing configuration by enabling this. 

Here's an example of how I'm using this in my home-manager config: https://gitlab.com/kira-bruneau/home-config/-/blob/028c1a50c82f8cad2be3d6693e6c53c67bf7e7c2/package/firefox/default.nix#L9-93.

**NOTE**: The hash verification is only required for custom search engines, it's ignored when setting your default engine to a builtin engine, but I set this up to still generate the hash anyway to be consistent with Firefox.

Closes #2318 @pshirshov

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

~- If this PR adds a new module~

### DIFF
--- a/modules/programs/firefox.nix
+++ b/modules/programs/firefox.nix
@@ -8,6 +8,8 @@ let
 
   cfg = config.programs.firefox;
 
+  jsonFormat = pkgs.formats.json { };
+
   mozillaConfigPath =
     if isDarwin then "Library/Application Support/Mozilla" else ".mozilla";
 
@@ -106,7 +108,7 @@ let
     '';
 
 in {
-  meta.maintainers = [ maintainers.rycee ];
+  meta.maintainers = [ maintainers.rycee maintainers.kira-bruneau ];
 
   imports = [
     (mkRemovedOptionModule [ "programs" "firefox" "enableAdobeFlash" ]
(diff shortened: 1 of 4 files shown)


pshirshov:
Excited to see this, thank you! Any chance it can be merged soon?

kira-bruneau:
I decided to [rename the `enable` option to `force`](https://github.com/nix-community/home-manager/compare/16dc28a6a89a766b5ac60807f92bffac9edeeb39..19c0bd4787d1f82d5515cac35cd1009a2d92d9ac) to make it clearer what it's actually doing.

pshirshov:
Is there any blocker for this to be merged? :(

pshirshov:
@[peterhoeg](https://github.com/peterhoeg)

pshirshov:
@rycee , sorry about bothering you, but could you have a look at this please?

modules/programs/firefox.nix
```diff
@@ -444,6 +532,124 @@ in {
...
+                        updateInterval = "_updateInterval";
+                        updateURL = "_updateURL";
+                        iconUpdateURL = "_iconUpdateURL";
+                        iconURL = "_iconURL";
+                        iconMapObj = "_iconMapObj";
+                        metaData = "_metaData";
+                        orderHint = "_orderHint";
+                        definedAliases = "_definedAliases";
+                        urls = "_urls";
+                      }.${name} or name;
```
Atemu, line 589: Perhaps this could be done prettier using `genAttrs`.

modules/programs/firefox.nix
```diff
@@ -444,6 +532,124 @@ in {
...
+                      "By modifying this file, I agree that I am doing so "
+                      + "only within $appName itself, using official, user-driven search "
+                      + "engine selection processes, and in a way which does not circumvent "
+                      + "user consent. I acknowledge that any attempt to change this file "
+                      + "from outside of $appName is a malicious act, and will be responded "
+                      + "to accordingly.";
+
+                    salt = profile.path + profile.search.default
+                      + (builtins.replaceStrings [ "$appName" ] [ "Firefox" ]
+                        disclaimer);
```
Atemu, line 635: Probably better to use a function and antiquotation rather than translating the JS code 1:1.

modules/programs/firefox.nix
```diff
@@ -444,6 +532,124 @@ in {
...
+
+                    hash = removeSuffix "\n" (builtins.readFile
+                      (pkgs.runCommand "hash" { inherit salt; } ''
+                        echo -n "$salt" | ${pkgs.openssl}/bin/openssl dgst -sha256 -binary | base64 > "$out"
+                      ''));
+                  in hash;
```
Atemu, line 641: This is IFD. We can hash strings within Nix though and perhaps it's not too hard to do base64 encoding?

Why do we need the value of the hash in Nix though? Isn't it always the same?

Couldn't we just substitute it in the JSON? This way we could avoid the IFD.

Atemu (changes requested):

modules/programs/firefox.nix
kira-bruneau: Oh right that's a great point This definitely shouldn't be an IFD. I think I just did it this way because the hashing defaults to base16.

I'll try looking into a Nix-only alternative.

Atemu: Looking into this further, I don't think this is easily doable in pure Nix. You'd need to implement a base16 to base64 converter.

See https://github.com/NixOS/nix/issues/3151

I think the substitution approach is the best we can do.

modules/programs/firefox.nix
kira-bruneau: Oh yep! That could make things simpler, but searchForm requires two underscores so it would still be an exception.

modules/lib/maintainers.nix
```diff
@@ -101,6 +101,12 @@
       fingerprint = "2BE3 BAFD 793E A349 ED1F  F00F 04D0 CEAF 916A 9A40";
     }];
   };
+  kira-bruneau = {
+    name = "Kira Bruneau";
+    email = "kira.bruneau@pm.me";
+    github = "kira-bruneau";
+    githubId = 382041;
+  };
```
ncfavier, line 109: This is not needed since you're already in the nixpkgs maintainer list.

pshirshov:
@Atemu , @ncfavier : any chance you may accept it?

pshirshov:
@kira-bruneau seems like the P/R doesn't work on macOS: 

```
error: Package ‘mozlz4a-2018-08-23’ in /nix/store/qr489cwcx2l5b9m234g5ipqxan0mr7fd-source/pkgs/tools/compression/mozlz4a/default.nix:27 is not supported on ‘x86_64-darwin’, refusing to evaluate.
```

Though I don't see anything linux-specific in the python script ([formula](https://github.com/NixOS/nixpkgs/blob/nixos-22.05/pkgs/tools/compression/mozlz4a/default.nix#L27), [script](https://gist.githubusercontent.com/kaefer3000/73febe1eec898cd50ce4de1af79a332a/raw/a266410033455d6b4af515d7a9d34f5afd35beec/mozlz4a.py)), do you think you may just incorporate that script into your PR?

pshirshov:
https://github.com/NixOS/nixpkgs/pull/196021 - I guess that may help

Atemu:
@pshirshov neither of us are maintainers here; we only provided feedback on the code.

pshirshov:
@rycee : please...

rycee:
Thanks! Fixed a few minor issues and merged to master now.

I think some of the code could benefit by having more let bindings to make it a bit more understandable, see e.g. the definition of `orderedEngineList`, but it's OK for now.

pshirshov:
To be honest, even as it is it's totally understandable and it's a real game changer. Many thanks!

rycee:
Indeed, this is a very nice feature and @kira-bruneau did a great job implementing it! So, I hope my comment above does not seem too negative, it was meant purely as constructive criticism.

kira-bruneau:
@rycee Yeah... I was worried it was too messy / hard to follow :sweat_smile:, but there's also just a lot going on to get this to work. I'll try to clean it up some more. Would it make sense to try to split this into a separate file?

rycee:
@kira-bruneau No need for a separate file, I think. It seems to me the thing that would help readability the most in this case is "naming things".

For example, instead of having, loosely modeled on `orderedEngineList`,

``` nix
imap (i: x: long expression) xs ++ map (y: long expression) (expression)
```

one could try doing something like

``` nix
let
  descriptiveName1 = i: x: long expression;
  descriptiveName2 = y: long expression;
  descriptiveName3 = expression;
in
  imap descriptiveName1 xs ++ map descriptiveName2 descriptiveName3
```

the `long expression`s themselves may benefit from let bindings, in which case the variable could be placed either within the function definition or on the same level as `descriptiveNameX`. Typically I would move the definition out of the function unless it uses a function argument.

PR #2986, opened 2022-05-29
fish: enable manpage completion

jian-lin:
### Description

This patch follows a similar patch[1] in nixpkgs. With this patch,
fish can complete manpages for programs installed through
home-manager, e.g. using home.packages.

[1]: https://github.com/NixOS/nixpkgs/pull/91794

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -285,6 +285,9 @@ in {
     {
       home.packages = [ cfg.package ];
 
+      # Support completion for `man` by building a cache for `apropos`.
+      programs.man.generateCaches = mkDefault true;
+
       xdg.dataFile."fish/home-manager_generated_completions".source = let
         # paths later in the list will overwrite those already linked
         destructiveSymlinkJoin = args_@{ name, paths, preferLocalBuild ? true


rcerc (approved):

modules/programs/fish.nix
```diff
@@ -285,6 +285,8 @@ in {
     {
       home.packages = [ cfg.package ];
 
+      programs.man.generateCaches = true;
```
rcerc, line 288: Sorry for the duplicate review. I think we should only set a default value for `programs.man.generateCaches` so the user can override it without using `lib.mkOverride`. (This is what Nixpkgs currently does.) We might as well include a comment to clarify what generating the caches does too.
```suggestion
      # Required for man completions
      programs.man.generateCaches = mkDefault true;
```

rcerc (changes requested):

modules/programs/fish.nix
jian-lin: Thanks. `mkDefault` is added. I prefer to explain the reason of a commit in its commit message though.

jian-lin:
Tests related to fish are passed.

rcerc (approved):

rycee:
Isn't the purpose of the existing `fish/home-manager_generated_completions` file to create these completions? @hyperfekt Do you know how that file relates to the change in this PR?

hyperfekt:
Yes, fish should not need access to the man database for any programs installed through home-manager, as it creates completions for those without user input and using every installed output's files directly.

hyperfekt:
Sorry, I replied after only a cursory look. The module creates completions for commands from their manpages, not completions for the `man` command to fill in the names of commands. So this is does something else entirely but is easily confused for each other.

rycee:
Thanks @jian-lin for the contribution and @hyperfekt for the clarification. I added a comment to the added line and merged to master.

PR #2987, opened 2022-05-29
programs.zsh: set ZPLUG_HOME before loading zplug

pitkling:
### Description

This is a minor bugfix. The option `zplugHome` of the `zplug` module allows to set the directory to which `zplug` downloads plugins. The option sets the shell variable `ZPLUG_HOME` accordingly. However, currently this is set **after** initializing `zplug` via `source ${pkgs.zplug}/init.zsh`. To be taken into account, the variable must be set before the initialization. To fix this, I simply swapped the two lines. This order is also used in `zplug`'s own tests, see [misc/zshrc in zplug](https://github.com/zplug/zplug/blob/c4dea766566b168a32dbfa8d10335e525ce39fcc/misc/zshrc#L6-L11).

### Checklist

- [ ] ~~Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).~~

    ```
    {component}: {description}

    {long description}
    ```

  - [ ] ~~Added myself and the module files to `.github/CODEOWNERS`.~~

### DIFF
--- a/modules/programs/zplug.nix
+++ b/modules/programs/zplug.nix
@@ -45,9 +45,9 @@ in {
     home.packages = [ pkgs.zplug ];
 
     programs.zsh.initExtraBeforeCompInit = ''
-      source ${pkgs.zplug}/init.zsh
-
       export ZPLUG_HOME=${cfg.zplugHome}
+
+      source ${pkgs.zplug}/init.zsh
 
       ${optionalString (cfg.plugins != [ ]) ''
         ${concatStrings (map (plugin: ''


klefevre (approved):

pitkling:
Let me know if I can/should make anything concerning this PR to get it going forward (I'm not very experienced with Github, so I might well have missed something I should have done…).

h7x4 (approved):

PR #567, opened 2019-02-07
ssh: add possibility to use match conditions inside a host match block

tuxinaut:
This change enables to create host match blocks where it's possible to check
against flowing arguments.

- canonical
- exec
- host
- originalhost
- user
- localuser

PR #2992, opened 2022-05-31
ssh: add generic Match support for matchBlocks

### LINKED ISSUE #2769: SSH: Create actual `Match` blocks
### Description

Currently the option `programs.ssh.matchBlocks` only allows creating `Host` blocks but not `Match` blocks.

This is slightly confusing but the main issue is that there is no way (that I could find) to create `Match` blocks
except creating them manually.

`Match` is basically a superset of `Host` as it allows more options to be set besides the host name pattern (including
custom shell commands for instance). There is no difference in the content of the respective blocks so the implementing
logic is already there.

It would be great to be able to create `Match` blocks as well as `Host` blocks.

I've created a dumb prototype which tries to reuse the `Host` declaration - currently it's very simple ~and untested~ _and works well for the author_ : https://github.com/jficz/home-manager/tree/add-match-to-ssh

Any comments, tips or feedback welcome. No need to be PC, I can handle anything ;)

jficz:
### Description

Allow creating actual `match` block in ssh config using `Match` instead of `Host`.

This is a preliminary `Match` support which doesn't try to exhaustively define the `Match` rules but
rather takes a more conservative approach and lets users define their own verbatim `Match` rules
as a string.

_It's better than nothing_ but it's not an ideal solution (yet). It works well for the author though.

Fixes #2769 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


esclear:
I tried / am trying to do the same in #2370 (though I started with a narrower case there).
#567 was along those lines as well.
I'll look at this later and may have some input.

cptMikky:
I took inspiration in the two but took a different approach so I could "ship a fully functional prototype" so to speak.

esclear:
As the match condition overwrites the host condition, I think it should cause an error or at least a warning, if both, host and match, are defined.

As I don't have a strong preference of specifying the match condition manually over specifying the parts in a set and this would solve my problems / applications, I'd like to see this land.
However, I don't know how others feel about this and have not received feedback about this in #2370 for a long time :/

SuperSandro2000:
> As the match condition overwrites the host condition, I think it should cause an error or at least a warning, if both, host and match, are defined.

No, there shouldn't be any warnings when that is done. ssh_config does not consider this and it is totally fine to mix and match multiple Host and Match. The ssh_config parser is very simple and goes from top to bottom. If a setting is set it can no longer be changed later but no errors are emitted. In a complex ssh_config you could have some defaults at the end of the file which are overwritten by host specific settings earlier.

This means we don't need complex logic around this and worst case it breaks configs which are working like intended.

jficz:
I think he meant that both Match and Host are set for a single entry in nix home configuration.

That would make sense but I couldn't find a way to make that work without breaking other logic - namely default
Host by entry name (https://github.com/nix-community/home-manager/pull/2992/files#diff-a58732e56068a41ed3b90c4069d4c4934a01dd8f1372633f8aae82995ee14495R289).

That's why I went with CSS-like priorities - the more specific the settings the higher priority it gets.

esclear:
@jficz is correct about my concern being what happens when botsh `host` and `match` options are set for one match block.

 I don't have much experience with defining options myself, but a falling back to the `dagName` to use as the Host, if neither `host` nor `match` options are set, seems like it should be doable. I may even try to build that, though it almost certainly won't be very idiomatic.

esclear:
I'd like to propose this:
```diff
diff --git a/modules/programs/ssh.nix b/modules/programs/ssh.nix
index 00616d5a..0d089927 100644
--- a/modules/programs/ssh.nix
+++ b/modules/programs/ssh.nix
@@ -285,15 +285,14 @@ let
         description = "Extra configuration options for the host.";
       };
     };
-
-    config.host = mkDefault dagName;
   });

-  matchBlockStr = cf: concatStringsSep "\n" (
+  matchBlockStr = key: cf: concatStringsSep "\n" (
     let
-      matchHead = if cf.match != null then
-        "Match ${cf.match}"
-      else "Host ${cf.host}";
+      hostOrDagName = if cf.host != null then cf.host else key;
+      matchHead = if cf.match != null
+        then  "Match ${cf.match}"
+        else "Host ${hostOrDagName}";
     in [ "${matchHead}" ]
     ++ optional (cf.port != null)            "  Port ${toString cf.port}"
     ++ optional (cf.forwardAgent != null)    "  ForwardAgent ${lib.hm.booleans.yesNo cf.forwardAgent}"
@@ -506,7 +505,7 @@ in
         ++ (optional (cfg.includes != [ ]) ''
           Include ${concatStringsSep " " cfg.includes}
         '')
-        ++ (map (block: matchBlockStr block.data) matchBlocks)
+        ++ (map (block: matchBlockStr block.name block.data) matchBlocks)
       )}

       Host *
@@ -522,5 +521,9 @@ in

         ${replaceStrings ["\n"] ["\n  "] cfg.extraConfig}
     '';
+
+    warnings = mapAttrsToList
+        (n: v: "In the SSH config match block `programs.ssh.matchBlocks.${n}` sets both of the host and match options.\nThe match option takes precedence.")
+        (filterAttrs (n: v: v.data.host != null && v.data.match != null) cfg.matchBlocks);
   };
 }
```
This would maintain the fallback to the `dagName`, if there is neither a `match` nor `host` option set and also warns for every match block where both, `match` and `option` are set.

But yeah, it might not be the most idiomatic solution.

modules/programs/ssh.nix
```diff
@@ -60,10 +60,20 @@ let
...
+          `Host` pattern used by this conditional block. See `ssh_config(5)` for `Host` block details. `match` option takes precedence if defined.
+        '';
+      };
+
+      match = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        example = "host <hostname> canonical\nhost <hostname> exec \"ping -c1 -q 192.168.17.1\"";
+        description = ''
+          `Match` block conditions used by this block. See `ssh_config(5)` for `Match` block details. `match` option takes precedence over `host` if defined.
```
sumnerevans, line 76:
```suggestion
          <literal>Match</literal> block conditions used by this block. See
           <citerefentry>
             <refentrytitle>ssh_config</refentrytitle>
             <manvolnum>5</manvolnum>
           </citerefentry>
           for <literal>Match</literal> block details.
           This option takes precedence over
           <option>ssh.matchBlocks.*.match<option>
           if defined.
```

Not quite sure what the actual option name is, please correct if necessary.

tests/modules/programs/ssh/match-blocks-match-and-hosts.nix
```diff
@@ -0,0 +1,32 @@
...
+          match = "host xyz canonical";
+          port = 2223;
+        };
+
+        "* !github.com" = { port = 516; };
+      };
+    };
+
+    home.file.assertions.text = builtins.toJSON
+      (map (a: a.message) (filter (a: !a.assertion) config.assertions));
```
sumnerevans, line 22: Are you sure this is necessary?

tests/modules/programs/ssh/match-blocks-match-and-hosts.nix
```diff
@@ -0,0 +1,32 @@
...
+
+    nmt.script = ''
+      assertFileExists home-files/.ssh/config
+      assertFileContent \
+        home-files/.ssh/config \
+        ${./match-blocks-match-and-hosts-expected.conf}
+      assertFileContent home-files/assertions ${./no-assertions.json}
```
sumnerevans, line 29: Same with this? I don't think you should have to check assertions yourself.

sumnerevans (changes requested):
This seems like the most general way to approach this problem. It covers all of the edge cases, but is a little bit more manual, which I think is OK as it avoids some of the pitfalls that the other attempts at adding this have fallen into.

jficz:
thanks all for the patches, I'll try to review and apply them as soon as I have some free cycles, hopefully some time this week

tests/modules/programs/ssh/match-blocks-match-and-hosts.nix
jficz: honestly, I don't know. I've never written any tests like this before. This is a _inspiraton_ (read: _copy-paste_) from another test which I adapted to the best of my understanding

tests/modules/programs/ssh/match-blocks-match-and-hosts.nix
jficz: same as above, I'm open to any suggestions, tips, criticism and fixes. That being said, I didn't create this from scratch but rather copied from already existing tests in the ssh module so unless I missed something (which is very possible), this is "how it's done"

jficz:
suggested news entry:
`programs.ssh.matchBlocks.*` now supports literal `Match` blocks via `programs.ssh.matchBlocks.*.match` option as an alternative to plain `Host` blocks.

@rycee @berbiche I believe the PR is now ready to be merged

jficz:
@sumnerevans could you re-review the change pls so it doesn't block the MR? also @rycee , @berbiche anything else I can do to help with pushing the PR?

esclear (approved):
Looks good to me, maybe add the proposed news entry to the PR?

teto:
@jficz  if you update the news entry, I can try to merge this tomorrow

jficz:
@teto added

PR #3003, opened 2022-06-05
swaylock: Add module

rcerc:
### Description

This adds a module to configure [swaylock](https://github.com/swaywm/swaylock) (a screen locker for Wayland). All swaylock options must be specified through `programs.swaylock.settings` and the user must give them appropriate types (because there are no Home Manager options for specific swaylock options). This is done for forward compatibility (but could be changed if necessary). Unless `programs.swaylock.settings` is an empty set, the swaylock configuration file is automatically added to the user's home directory (this should be fine because the user can forcefully disable it by setting `programs.swaylock.settings` to `lib.mkForce { }`).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/swaylock.nix
+++ b/modules/programs/swaylock.nix
@@ -1,0 +1,32 @@
+{ config, lib, ... }:
+
+let cfg = config.programs.swaylock;
+in {
+  meta.maintainers = [ lib.hm.maintainers.rcerc ];
+
+  options.programs.swaylock.settings = lib.mkOption {
+    type = with lib.types; attrsOf (oneOf [ bool float int str ]);
+    default = { };
+    description = ''
+      Default arguments to <command>swaylock</command>. An empty set
+      disables configuration generation.
+    '';
+    example = {
+      color = "808080";
+      font-size = 24;
+      indicator-idle-visible = false;
+      indicator-radius = 100;
+      line-color = "ffffff";
+      show-failed-attempts = true;
+    };
+  };
+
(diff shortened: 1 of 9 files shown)


modules/programs/swaylock.nix
```diff
@@ -0,0 +1,32 @@
...
+      font-size = 24;
+      indicator-idle-visible = false;
+      indicator-radius = 100;
+      line-color = "ffffff";
+      show-failed-attempts = true;
+    };
+  };
+
+  config.xdg.configFile."swaylock/config" = lib.mkIf (cfg.settings != { }) {
+    text = lib.concatStrings (lib.mapAttrsToList (n: v:
```
teto, line 25: maybe you could use `generators.toINIWithGlobalSection ` from nixpkgs since it looks like an ini file.

teto (approved):

modules/programs/swaylock.nix
rcerc: I was hoping to use a generator like that, but swaylock only seems to accept boolean options specified as `name` instead of `name=true` (at least on my system).

teto: we should have a way to parametrize this in nixpkgs t o avoid having to rewrite this every time.

rcerc: Maybe, although it might be non-standard INI syntax. A custom `mkKeyValue` function could be passed to handle these cases though. Do you suggest that in this case?

PR #3006, opened 2022-06-07
docs: disable `_module.check` for nixos/nix-darwin modules

ncfavier:
Currently we're maintaining a "mock" module made of sink options, which requires updating whenever the definitions in the nixos/nix-darwin modules change.

Instead, set `_module.check` to false so that definitions in those modules are simply ignored.

This is [what we do](https://github.com/NixOS/nixos-search/blob/75e8457105ee6a9d3f60c71add59661bb809f236/flake-info/src/commands/flake_info.nix#L53) in nixos-search.

This could probably also be done in nmd by default.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -18,6 +18,8 @@ let
     }];
   };
 
+  dontCheckDefinitions = { _module.check = false; };
+
   buildModulesDocs = args:
     nmd.buildModulesDocs ({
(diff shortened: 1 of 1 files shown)


rycee:
Thanks! Looks much cleaner, merged to master now.

PR #3010, opened 2022-06-08
emacs: extend startWithUserSession to start after graphical session

jian-lin:
### Description

It's useful if you want emacs daemon to get some environment variables, e.g. ibus related variables when using Gnome.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/emacs.nix
+++ b/modules/services/emacs.nix
@@ -81,14 +81,14 @@ in {
       enable = mkEnableOption "systemd socket activation for the Emacs service";
     };
 
-    startWithUserSession = lib.mkOption {
-      type = lib.types.bool;
+    startWithUserSession = mkOption {
+      type = with types; either bool (enum [ "graphical" ]);
       default = !cfg.socketActivation.enable;
       defaultText =
         literalExpression "!config.services.emacs.socketActivation.enable";
-      example = true;
+      example = "graphical";
       description = ''
-        Whether to launch Emacs service with the systemd user session.
+        Whether to launch Emacs service with the systemd user session.  If it is <literal>true</literal>, Emacs service is started by default.target.  If it is <literal>"graphical"</literal>, Emacs service is started by graphical-session.target.
       '';
     };
 
@@ -115,6 +115,11 @@ in {
           Description = "Emacs text editor";
           Documentation =
             "info:emacs man:emacs(1) https://gnu.org/software/emacs/";
+
+          After = optional (cfg.startWithUserSession == "graphical")
+            "graphical-session.target";
+          PartOf = optional (cfg.startWithUserSession == "graphical")
+            "graphical-session.target";
 
           # Avoid killing the Emacs session, which may be full of
           # unsaved buffers.
@@ -156,8 +161,15 @@ in {
           ExecStopPost =
             "${pkgs.coreutils}/bin/chmod --changes +w ${socketDir}";
         };
-      } // optionalAttrs (cfg.startWithUserSession) {
-        Install = { WantedBy = [ "default.target" ]; };
+      } // optionalAttrs (cfg.startWithUserSession != false) {
+        Install = {
+          WantedBy = [
+            (if cfg.startWithUserSession == true then
+              "default.target"
+            else
+              "graphical-session.target")
+          ];
+        };
       };
 
       home = {
--- a/tests/modules/services/emacs/default.nix
+++ b/tests/modules/services/emacs/default.nix
@@ -1,6 +1,8 @@
 {
   emacs-service-27 = ./emacs-service-27.nix;
   emacs-service-28 = ./emacs-service-28.nix;
+  emacs-service-28-after-graphical-session-target =
+    ./emacs-service-28-after-graphical-session-target.nix;
   emacs-socket-27 = ./emacs-socket-27.nix;
   emacs-socket-28 = ./emacs-socket-28.nix;
   emacs-default-editor = ./emacs-default-editor.nix;
--- a/tests/modules/services/emacs/emacs-service-28-after-graphical-session-target.nix
+++ b/tests/modules/services/emacs/emacs-service-28-after-graphical-session-target.nix
@@ -1,0 +1,40 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+{
+  config = {
+    nixpkgs.overlays = [
+      (self: super: rec {
+        emacs = pkgs.writeShellScriptBin "dummy-emacs-28.2" "" // {
+          outPath = "@emacs@";
+        };
+        emacsPackagesFor = _:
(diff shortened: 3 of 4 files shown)


Dietr1ch:
I kinda wanted to propose the same thing, but I'm unsure on what's the best way to fix emacs's graphical sessions when using the daemon.

Waiting for `default.target` makes it start too early, but waiting for `graphical-session.target` may have unintended consequences,
  - What happens if your graphical session gets broken?
    - It seems that it's better to just wait for `default.target`, and somehow fix things once the graphical session comes up (no clue on whether this is feasible).
  - What happens if you just ssh into your machine.
    - Will the daemon start?
    - If not, then doing `systemctl --user start emacs` might also not work.


Anyway, I do think that I just want to wait for the `graphical-session.target` by now, things don't break often, and if emacs is not running when I ssh it's fine, I can either use emacs remotely with TRAMP or just start emacs in the foreground if I really need emacs to be running locally.

jian-lin:
Thanks for your review.

> What happens if your graphical session gets broken? 

Well, I think you should fix your graphical session then. A broken graphical session is not a common case in my opinion. Even in this case, you can use a non-daemon emacs to fix it.

> What happens if you just ssh into your machine. 

> Will the daemon start?

It will not start if you enable this new option.

> If not, then doing systemctl --user start emacs might also not work.

In my test, `systemctl --user start emacs` can start an emacs daemon.

Here is my test steps:

1. boot my laptop to GDM, but not log in
2. ssh into my laptop
3. `systemctl --user status emacs` shows it isn't started
4. `systemctl --user start emacs` will start the daemon

I guess emacs daemon can be manually started because graphical-session.target is not in this transaction, so the `After` doesn't apply. I am not sure though. 

> It seems that it's better to just wait for default.target, and somehow fix things once the graphical session comes up (no clue on whether this is feasible).

I prefer to let systemd handle the starting order. Updating environment variables for a running emacs daemon is not easy.

rycee:
I'm not sure this is needed in the module since it should be relatively simple to override with whichever target you want. Something like

```
systemd.user.services.emacs = {
  Unit.After = [ "graphical-session.target" ];
  Install.WantedBy = lib.mkForce [ "graphical-session.target" ];
};
```

perhaps.

jian-lin:
> I'm not sure this is needed in the module since it should be relatively simple to override with whichever target you want

A lot of module options can be relatively simple to override thanks to flexibility of the module system, e.g. https://github.com/nix-community/home-manager/pull/2941.

IMHO, starting emacs daemon after graphical session target like any other graphical programs is a common use case, e.g. @Dietr1ch also needs this, and is suitable to be contained in this module.

Dietr1ch:
I'd appreciate having the option around. It'd be easier to discover, and it could be set up in a way that auto-detects when you are running a graphical server, which would give a great experience to users IMO.

Without this, users would need to start wondering why things do not work, and hopefully run into this or a similar issue to understand that they'll need to tweak the systemd unit to wait for their graphical environment. That's not a great experience, even though users where free to easily override their configs in the right way from the start.

jian-lin:
(My reply through email seems to be lost by github.)

> it could be set up in a way that auto-detects when you are running a
> graphical server, which would give a great experience to users IMO.

That would be great but I have no idea how we implement this.

If it's on NixOS and home-manager is used as a NixOS module, we can get
the system config from `nixosConfig`. But what NixOS option should we
check? Also, there are cases where home-manager is used as a standalone
tool and even on a foreign system.

jian-lin:
> it could be set up in a way that auto-detects when you are running a
> graphical server, which would give a great experience to users IMO.

That would be great but I have no idea how we implement this.

If it's on NixOS and home-manager is used as a nixos module, we can get
the system config from `nixosConfig`. But what nixos option should we
check? Also, there are cases where home-manager is used as a standalone
tool and even on a foreign system.

futile:
This would be extremely useful to have, as currently the emacs daemon is broken for everyone that uses GUI emacs. Which should be quite some people.

Don't really care whether it's this PR, or a simple option that performs the override mentioned by @rycee in https://github.com/nix-community/home-manager/pull/3010#issuecomment-1166241254, but it should be more easily discoverable than having to find this PR/issue #3443.

DamienCassou:
> This would be extremely useful to have, as currently the emacs daemon is broken for everyone that uses GUI emacs.

I have been using this module with GUI Emacs for years without any trouble. I'm not quite sure what kind of trouble people are having. The PR description mentions Gnome which I'm not using though.

I'm fine merging the PR or making graphical-session the default target if that helps people.

SohamG:
Just finally fixed my ever-broken emacs-daemon/client setup. This was the sole reason I couldnt just use the emacs client setup from home-manager out of the box. Would love to see this merged.

DamienCassou:
It would be nice to merge this PR.

modules/services/emacs.nix
```diff
@@ -101,13 +101,27 @@ in {
         editor using the <envar>EDITOR</envar> environment variable.
       '';
     };
+
+    afterGraphicalSessionTarget = mkOption {
```
ncfavier, line 105: Instead of a new option, we should extend the type of `startWithUserSession` from `bool` to something like `enum [ false true "graphical" ]`.

ncfavier (changes requested):

modules/services/emacs.nix
```diff
@@ -82,13 +82,13 @@ in {
     };
 
     startWithUserSession = lib.mkOption {
-      type = lib.types.bool;
+      type = lib.types.enum [ false true "graphical" ];
```
ncfavier, line 85: Actually I feel this is a bit better:

```suggestion
      type = with lib.types; either bool (enum [ "graphical" ]);
```

modules/services/emacs.nix
```diff
@@ -156,9 +161,17 @@ in {
           ExecStopPost =
             "${pkgs.coreutils}/bin/chmod --changes +w ${socketDir}";
         };
-      } // optionalAttrs (cfg.startWithUserSession) {
-        Install = { WantedBy = [ "default.target" ]; };
-      };
+      } // optionalAttrs (cfg.startWithUserSession == true
+        || cfg.startWithUserSession == "graphical") {
```
ncfavier, line 165:
```suggestion
      } // optionalAttrs (cfg.startWithUserSession != false) {
```

jian-lin: I slightly prefer the origin code for better clarity but does not have a strong opinion.

ncfavier: The original code seems easier to forget to extend the condition when new values are added to the enum (maybe irrelevant here, but as a general principle).

ncfavier (approved):

SohamG:
Thanks for the quick action everyone!

PR #3013, opened 2022-06-10
integration-common: set hmModule's description directly

ncfavier:
Depends on https://github.com/NixOS/nixpkgs/pull/173621 which just landed, let's wait a couple weeks before merging.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


rycee:
Thanks, merged to master now! Looks nice 👍

PR #3017, opened 2022-06-12
xdg-user-dirs: `createXdgUserDirectories` after `linkGeneration`

ncfavier:
In the scenario where some XDG user directory is a symlink defined by `home.file`, we want the symlink to be created before we try to `mkdir -p` that directory, as it will then silently succeed.

On the other hand, if we create the directory first, creating the symlink will fail (in a surprising way, see https://github.com/nix-community/home-manager/pull/3018).

We lose nothing by doing this as `linkGeneration` creates the directories it needs.

We can also fix the `mkIf` weirdness since https://github.com/nix-community/home-manager/pull/2822.

cc @thiagokokada 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/xdg-user-dirs.nix
+++ b/modules/misc/xdg-user-dirs.nix
@@ -120,12 +120,10 @@ in {
 
     home.sessionVariables = directories;
 
-    home.activation = mkIf cfg.createDirectories {
-      createXdgUserDirectories = let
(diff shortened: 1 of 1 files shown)


thiagokokada (approved):

rycee:

PR #3018, opened 2022-06-12
files: avoid surprises when linking files

ncfavier:
In the unlikely (but possible) event that a directory is created in place of a target file *after* `checkLinkTargets` but *before* `linkGeneration`, currently the link is created *inside* the directory, which is surprising (see https://github.com/nix-community/home-manager/pull/3017 for context).

To avoid this, use `ln -T` and bail if it fails.

~Also bail if backing up the target fails.~

Ideally, `checkLinkTargets` and `linkGeneration` should be more tightly coupled.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/files.nix
+++ b/modules/files.nix
@@ -185,8 +185,9 @@ in
               $VERBOSE_ECHO "Skipping '$targetPath' as it is identical to '$sourcePath'"
             else
               # Place that symlink, --force
+              # This can still fail if the target is a directory, in which case we bail out.
               $DRY_RUN_CMD mkdir -p $VERBOSE_ARG "$(dirname "$targetPath")"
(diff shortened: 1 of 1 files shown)


berbiche (changes requested):
Thanks for the PR!

I'm pretty sure I hit this situation at some point but I don't remember.

The `-T` flag does not seem to be POSIX and does not seem to be supported by BSD `ln` (so not supported by Darwin/Mac).

Do you know whether the GNU coreutil package is in the activation script's $PATH? If so, my previous comment can be ignored but a note should be left.

ncfavier:
cc @rycee, I still think this change is good

ncfavier:
I've removed the other change, added a comment and rebased on master.

PR #1767, opened 2021-01-30
Improve flake homeManagerConfiguration

terlar:
### Description

This includes two changes:

#### Fix pkgs passed to configuration
When I passed a custom pkgs which adds a few new packages via overlays, they did not end up in my configuration modules, but when I updated `extraSpecialArgs` to include the `pkgs` then all worked fine.

#### Separate config from custom modules
I have a bunch of custom home-manager modules, some private ones not ready for prime-time and some coming from third party packages. I think it would be nice to separate those from the regular configuration (just like home-manager itself does). I named it `extraModules`.

Instead of:
```
configuration = {
  imports = [
    emacs-config.homeManagerModules.emacsConfig
    ./home.nix
   ];
};
```
    
You can do:
```
configuration = ./home.nix;
extraModules = [ emacs-config.homeManagerModules.emacsConfig ];
```

### Checklist

- [x] Change is backwards compatible.

- [x] Code formatted with `./format`.

- [x] Code tested through `nix-shell --pure tests -A run.all`.

- [ ] Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).

- [x] Commit messages are formatted like

    ```
    {component}: {description}

    {long description}
    ```

PR #3023, opened 2022-06-14
flake.nix: Allow homeManagerConfiguration to accept a list of modules

### LINKED ISSUE #2717: Pass argumnets from the flake to `home.nix` using `_module.args` instead of `extraSpecialArgs`?
According to NixOS/Nixpkgs [`lib/modules.nix`](https://github.com/NixOS/nixpkgs/blob/master/lib/modules.nix#L92-L95)

> This (`specialArgs`) should only be used for special arguments that need to be evaluated when resolving module structure (like in `imports`). For everything else, there's` _module.args`. If `specialArgs.modulesPath` is defined it will be used as the base path for disabledModules.

Should change the way how variables are passed, either by instructing users to pass extra arguments through `_modules.args`, or by creating another parameter to do the job?

ShamrockLee:
### Description

This allows users to pass through multiple modules to function `homeManagerConfiguration` in their `flake.nix`. The use cases include NixOS-configuration-like

```Nix
{
  outputs = inputs: {
     configuration =  [
      (import ./home.nix)
      ({ config, ... }: { _module.args =  {
        # Pass extra arguments here
      };})
    ];
  };
}
```

Fixes #2717

### Checklist

- [ ] Code tested through `nix-shell --pure tests -A run.all`. (not applicable)

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -38,7 +38,10 @@
           import ./modules {
             inherit pkgs lib check extraSpecialArgs;
             configuration = { ... }: {
-              imports = [ configuration ] ++ extraModules;
+              imports = (if builtins.isList configuration then
+                configuration
+              else
+                [ configuration ]) ++ extraModules;
               home = { inherit homeDirectory stateVersion username; };
               nixpkgs = { inherit (pkgs) config overlays; };
             };


ncfavier:
`homeManagerConfiguration` already accepts a list of modules in `extraModules`. We could probably unify both arguments into `modules` (see https://github.com/nix-community/home-manager/pull/1767#discussion_r605609607), but I don't think there's much point to this now.

https://github.com/nix-community/home-manager/issues/2717 is a non-issue because you can already do what you want in two ways:

```nix
homeManagerConfiguration {
    configuration = {
        imports = [ ./home.nix ];
        _module.args = { ... };
    };
}
```

or

```nix
homeManagerConfiguration {
    configuration = {
        _module.args = { ... };
    };
    extraModules = [ ./home.nix ];
}
```

ShamrockLee:
> 

Thanks a lot for explaining that. I'll close both of the issue and the PR.

Issue #2985, opened 2022-05-29
bug: hm installation instructions for NixOS (section 1.2) are broken/incomplete


motiejus:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

I have followed the [documentation](https://nix-community.github.io/home-manager/index.html#sec-install-nixos-module) to install home-manager on a fresh NixOS machine. It failed this way:

```
$ sudo nixos-rebuild switch
error: file 'home-manager/nixos' was not found in the Nix search path (add it using $NIX_PATH or -I), at /etc/nixos/configuration.nix:11:7
(use '--show-trace' to show detailed location information)
```

The channel is added for the system:

```
$ sudo nix-channel --list
home-manager https://github.com/nix-community/home-manager/archive/release-21.11.tar.gz
nixos https://nixos.org/channels/nixos-21.11
```

I have found #1267, but I am still not clear on what the correct resolution should be. Regardless, my goals are a bit different:

1. Find a resolution that is good enough to be documented. This is the first step that prospective home-manager users run, so should be robust. I will need your help here.
2. Update the official manual for the exact/correct instructions how to install home-manager on NixOS. I will do that.

### System information

```yaml
- system: `"x86_64-linux"`
- host os: `Linux 5.10.109, NixOS, 21.11 (Porcupine)`
- multi-user?: `yes`
- sandbox: `yes`
- version: `nix-env (Nix) 2.3.16`
- channels(root): `"nixos-21.11.337714.9bc0e974545"`
- channels(motiejus): `""`
- nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos
```

motiejus:
> The home-manager channel should be added for your user, not root. I guess the `#` prompts in the manual are misleading.

Thanks, I guess that was the culprit.

However, I no longer plan to use home-manager, so closing.

ncfavier:
I think I spoke too fast and assumed you were running `home-manager switch`. `sudo nixos-rebuild switch` should use the root channels, so my guess is that your NIX_PATH was broken somehow.

PR #3025, opened 2022-06-15
docs: use `$` prompts for nix-channel commands

ncfavier:
Channels should be added for the normal user, not for root, because the home-manager configuration is not supposed to be evaluated as root.

Fixes https://github.com/nix-community/home-manager/issues/2985

### DIFF
--- a/docs/faq.adoc
+++ b/docs/faq.adoc
@@ -143,8 +143,8 @@ You can add the `nixpkgs-unstable` channel by running
 
 [source,console]
 ----
-# nix-channel --add https://nixos.org/channels/nixpkgs-unstable nixpkgs-unstable
-# nix-channel --update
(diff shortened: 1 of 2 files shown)


teto (approved):

rycee:
I guess this doesn't hurt. But when I used to use `nix-channel` I always managed the channels using root. It should work fine to do so since the root channels are available to all as far as I understand:

``` console
$ tr ':' '\n' <<<$NIX_PATH | grep root
/nix/var/nix/profiles/per-user/root/channels
```

rycee:
Merged to master now 🙂

ncfavier:
Actually I think I confused myself... the commands in the NixOS module section should indeed run as root. The other ones probably not.

PR #3029, opened 2022-06-16
format: update and remove exceptions

ncfavier:
Switches to nixfmt 0.5.0 and removes exceptions for files without open PRs. (used [this script](https://github.com/ncfavier/config/blob/d565fc7cb6daf4a9e847f3eed681c886df8eba2e/modules/git.nix#L105-L123) to check)

### DIFF
--- a/home-manager/home-manager.nix
+++ b/home-manager/home-manager.nix
@@ -1,9 +1,5 @@
-{ pkgs ? import <nixpkgs> {}
-, confPath
-, confAttr ? null
-, check ? true
-, newsReadIdsFile ? null
-}:
+{ pkgs ? import <nixpkgs> { }, confPath, confAttr ? null, check ? true
+, newsReadIdsFile ? null }:
 
 let
   inherit (pkgs.lib)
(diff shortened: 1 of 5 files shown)


home-manager/home-manager.nix
```diff
@@ -1,78 +1,64 @@
-{ pkgs ? import <nixpkgs> {}
-, confPath
-, confAttr ? null
-, check ? true
-, newsReadIdsFile ? null
-}:
+{ pkgs ? import <nixpkgs> { }, confPath, confAttr ? null, check ? true
+, newsReadIdsFile ? null }:
```
teto, line 2: I preferred the oldversion, is that tweakable in nixfmt ?

ncfavier: No :/ https://github.com/serokell/nixfmt/issues/46
We could keep the exception for this file.

teto: There are some improvements in the formatting but the degradations annoy me more than the improvements so personnally I would advocate to keep the same nixfmt version, just remove the exceptions. Let's see what others think

Issue #2028, opened 2021-05-19
Mako module stopped work


timsofteng:
Hello. I find `mako` notification module stopped work. Daemon doesn't start. I can't see it in my processes and notifications doen't work.
Is it knowing issue?

sumnerevans:
I have the following in my config to fix the issues with dbus: https://git.sr.ht/~sumner/home-manager-config/tree/master/item/modules/window-manager/wayland.nix#L64

berbiche:
This might be of interest: https://github.com/alebastr/sway-systemd

asymmetric:
Can we reopen this issue? As it is, the `mako` module is broken.

@sumnerevans could you point a permalink to the relevant portion of your config?

berbiche:
@asymmetric hi, do you have multiple notification daemons on your system? They are likely fighting to register for the notification daemon.

If you run the following command, you should get whether a notification daemon (mako or something else) is receiving notifications: `dbus-send --print-reply --dest=org.freedesktop.DBus /org/freekdesktop/DBus org.freedesktop.DBus.ListNames | grep Notifications`

asymmetric:
@berbiche I don't have any notification daemon running, actually. Running that snippet returns empty.

To fix the issue, I now run mako explicitly in my sway config with `exec mako`, but this wasn't necessary before.

Can you instead confirm that just having `programs.mako.enable = true` in your config gives a different result?

sumnerevans:
Not sure if this actually fixes people's problems, but I see this in my git commit log, which means that it maybe is necessary? https://github.com/sumnerevans/home-manager-config/commit/f82aacb32cb25145b614bc1d4fd907617c416b6e#diff-60d8bd9845e95cde122a9132e44f5da4e55231a3505d10548de91e751aaa2afa

samhh:
Is the mako module as currently defined supposed to start it as a daemon? I've had to define my own systemd unit instead:

```nix
{
  systemd.user.services.mako = {
    Install.WantedBy = [ "sway-session.target" ];
    Service.ExecStart = "${pkgs.mako}/bin/mako";
  };
}
```

benjumanji:
Every single solution here seems to miss the forest for the trees.

```
  xdg.dataFile."dbus-1/services/fr.emersion.mako.service" = {
    enable = config.services.mako.enable;
    source = "${pkgs.mako}/share/dbus-1/services/fr.emersion.mako.service";
  };
```

[...]

PR #3031, opened 2022-06-17
sway: import dbus env vars and explicitly specify them 

### LINKED ISSUE #2806: bug: sway systemd integration is imperfect (breaking things at times)

### Issue description

I've got some changes staged up for sway and friends but wanted to file an issue for visibility and to make sure I don't forget these fixes on some fork.

Particularly, `systemctl --user import-environment` should be replaced with `dbus-update-activation-environment --systemd`. I've also seen some issues with how the target binding works that causes some services to not be restarted if you restart the sway target, instead requiring that the user restart both the sway and graphical-session targets.

I do have fixes for most of this stuff.

**Particularly** the environment fix will make GCR/pinentry-gnome3 actually reliable. As-is today, if you start sway, then stop sway, and start it again, pinentry will likely not work or will fall back to curses.

Finally, I think we should:
- encapsulate `sway` itself in a systemd --user unit, bound to the graphical session target
- have `sway-start` that invokes `systemctl --user start graphical-session.target`, maybe it warns if it's already running and tries to suggest which tty to switch to
- have `sway-stop` that invokes `systemctl --user stop graphical-session.target`

I want this because:
- I keep catching stale env vars or sockets lying around. systemd is great at handling these things, we should leverage it

I suspect that maybe the dbus environment issue may very likely have to do with why many users have opted to not use it (yet?).

### Maintainer CC

_No response_

### System information

```markdown
n/a
```

oxalica:
### Description

Supersede and closes #2385 
Fix #2806 and #2028

- Importing all environment variables is considered deprecated for
  `systemdctl import-environment`. The list of variables are picked
  from:
  https://github.com/swaywm/sway/wiki/Systemd-integration#managing-user-applications-with-systemd

  The `XDG_CURRENT_DESKTOP` is said to be required for portals, see:
  https://github.com/nix-community/home-manager/pull/2385#issuecomment-1026454162

- DBus activation environment should also be updated. Otherwise, DBus
  activated programs, without a coresponding systemd service, cannot get a
  correct environment and would fail, eg. `mako`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/window-managers/i3-sway/sway.nix
+++ b/modules/services/window-managers/i3-sway/sway.nix
@@ -309,7 +309,7 @@ let
       )
     else
       [ ]) ++ (optional cfg.systemdIntegration ''
-        exec "systemctl --user import-environment; systemctl --user start sway-session.target"'')
+        exec "${pkgs.dbus}/bin/dbus-update-activation-environment --systemd DISPLAY WAYLAND_DISPLAY SWAYSOCK XDG_CURRENT_DESKTOP; systemctl --user start sway-session.target"'')
       ++ (optional (!cfg.xwayland) "xwayland disable") ++ [ cfg.extraConfig ]));
 
   defaultSwayPackage = pkgs.sway.override {
@@ -320,7 +320,7 @@ let
   };
 
 in {
-  meta.maintainers = with maintainers; [ alexarice sumnerevans sebtm ];
+  meta.maintainers = with maintainers; [ alexarice sumnerevans sebtm oxalica ];
 
   options.wayland.windowManager.sway = {
     enable = mkEnableOption "sway wayland compositor";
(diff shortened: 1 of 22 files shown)


modules/lib/maintainers.nix
```diff
@@ -223,6 +223,16 @@
...
+  oxalica = {
+    email = "oxalicc@pm.me";
+    github = "oxalica";
+    githubId = 14816024;
+    name = "oxalica";
+    keys = [{
+      longkeyid = "ed25519/0x7571654CF88E31C2";
+      fingerprint = "F90F FD6D 585C 2BA1 F13D  E8A9 7571 654C F88E 31C2";
+    }];
+  };
```
ncfavier, line 235: Shouldn't be needed as you're already in nixpkgs.

oxalica: Fixed.

sumnerevans (approved):
LGTM

berbiche (approved):

PR #3035, opened 2022-06-18
docs: revert to `#` prompt on NixOS

ncfavier:
See https://github.com/nix-community/home-manager/pull/3025#issuecomment-1159565524, sorry about that.

I don't understand how nix-darwin works so I'm not sure how the channels should be added there.


rycee:
Merged to master, thanks!

PR #3028, opened 2022-06-16
Clean up after #2971

ncfavier:
Applies review comments from https://github.com/nix-community/home-manager/pull/2971#pullrequestreview-1008311355

PR #3037, opened 2022-06-19
homeManagerConfiguration: remove default stateVersion, update docs

ncfavier:
Updates the docs after https://github.com/nix-community/home-manager/pull/3028, removes the now useless `system` argument, and removes the default `stateVersion` value because it's now quite old and the user should set it explicitly anyway.

Fixes https://github.com/nix-community/home-manager/pull/2647

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -29,19 +29,40 @@
 
       lib = {
         hm = import ./modules/lib { lib = nixpkgs.lib; };
-        homeManagerConfiguration = { configuration, system, homeDirectory
-          , username, extraModules ? [ ], extraSpecialArgs ? { }, pkgs
-          , lib ? pkgs.lib, check ? true, stateVersion ? "20.09" }@args:
-          assert nixpkgs.lib.versionAtLeast stateVersion "20.09";
+        homeManagerConfiguration = { modules ? [ ], pkgs, lib ? pkgs.lib
+          , extraSpecialArgs ? { }, check ? true
+            # Deprecated:
+          , configuration ? null, extraModules ? null, stateVersion ? null
+          , username ? null, homeDirectory ? null, system ? null }@args:
+          let
+            throwForRemovedArg = v:
+              lib.throwIf (v != null) ''
+                The 'homeManagerConfiguration' arguments
 
-          import ./modules {
+                  - 'configuration',
+                  - 'username',
+                  - 'homeDirectory'
+                  - 'stateVersion',
+                  - 'extraModules', and
+                  - 'system'
+
+                have been removed. Instead use the arguments 'pkgs' and
+                'modules'. See the 22.11 release notes for more.
+              '';
+
+            throwForRemovedArgs = throwForRemovedArg configuration # \
+              throwForRemovedArg username # \
+              throwForRemovedArg homeDirectory # \
+              throwForRemovedArg stateVersion # \
+              throwForRemovedArg extraModules # \
+              throwForRemovedArg system;
+          in throwForRemovedArgs (import ./modules {
             inherit pkgs lib check extraSpecialArgs;
             configuration = { ... }: {
-              imports = [ configuration ] ++ extraModules;
-              home = { inherit homeDirectory stateVersion username; };
+              imports = modules;
               nixpkgs = { inherit (pkgs) config overlays; };
             };
-          };
+          });
       };
     } // utils.lib.eachDefaultSystem (system:
       let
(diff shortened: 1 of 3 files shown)


terlar (commented):
Perhaps all required attributes should be first or last? Might be easier to spot what is required to provide

terlar (approved):

rycee:
Since there are two independent changes here they should go into two separate commits. That should also help limit the subject line to [50 characters](https://cbea.ms/git-commit/#limit-50).

flake.nix
```diff
@@ -29,9 +29,9 @@
 
       lib = {
         hm = import ./modules/lib { lib = nixpkgs.lib; };
-        homeManagerConfiguration = { configuration, system, homeDirectory
-          , username, extraModules ? [ ], extraSpecialArgs ? { }, pkgs
-          , lib ? pkgs.lib, check ? true, stateVersion ? "20.09" }@args:
+        homeManagerConfiguration = { configuration, pkgs, lib ? pkgs.lib
+          , homeDirectory, username, stateVersion, extraModules ? [ ]
```
rycee, line 33: Do you know why `stateVersion` is here at all? Same with `username` and `homeDirectory`. Seems they could easily be defined inside `configuration`?

ncfavier: I was wondering that as well. The problem if we remove them is that `stateVersion` becomes optional as it has a default value of 18.09, and this in turn seems to entail default values for `username` and `homeDirectory`, which we don't want.

I think `stateVersion` should either not have a default value, or default to `pkgs.lib.trivial.release` [like in nixpkgs](https://github.com/NixOS/nixpkgs/blob/e0a42267f73ea52adc061a64650fddc59906fc99/nixos/modules/misc/version.nix#L83-L85). Then maybe we can remove those arguments.

ncfavier: Additionally if we remove the `stateVersion` argument we lose the ability to assert that it's ≥ 20.09, which is needed to avoid the uninformative error about looking up `<nixpkgs>` in pure mode. Maybe that's fine though, it's 2022.

ncfavier: cc @berbiche @terlar as you might have thoughts on this

terlar: As the person who added `stateVersion` I would be fine with removing it, but perhaps we should pass some option to the configuration letting Home Manager know the config came from a flake (or that it needs to be "pure") so we could assert the stateVersion is ≥ 20.09.

The error is quite confusing if you have an older `stateVersion` and use flakes.

There has been some discussion here as well #3041 2647

ncfavier: I did try adding a module assertion directly in the flake, but it was useless as the problematic `<nixpkgs>` is evaluated first. Not sure if we can fix that.

rycee: I've been meaning to remove the default value of `stateVersion` from the option definition and will do that in a separate commit. Will do that change before merging this.

ncfavier: I've done that already: [`96ebe57` (#3037)](https://github.com/nix-community/home-manager/pull/3037/commits/96ebe57b47900d49e9989bc99973f991b83ca0e1)

flake.nix
```diff
@@ -29,12 +29,26 @@
 
       lib = {
         hm = import ./modules/lib { lib = nixpkgs.lib; };
-        homeManagerConfiguration = { configuration, pkgs, lib ? pkgs.lib
-          , extraModules ? [ ], extraSpecialArgs ? { }, check ? true }@args:
+        homeManagerConfiguration = { modules ? [ ], pkgs, lib ? pkgs.lib
+          , extraSpecialArgs ? { }, check ? true
+          , configuration ? "configuration is removed, use modules instead"
```
ncfavier, line 34: This seems ugly, I'd rather suggest something like `{ configuration ? null }: lib.throwIf (configuration != null) "configuration is removed, use modules instead" ...`

rycee: Nice, i looked a little for something like `throwIf` but obviously not enough. Updated now. Btw, apologies for putting commits on your branch, you may have been planning to add something like this already. I had a bit of time to spend today and took the opportunity to try out flakes a bit.

ncfavier: No worries but you did remove https://github.com/nix-community/home-manager/commit/96ebe57b47900d49e9989bc99973f991b83ca0e1. Are you planning to do that in a separate PR, or just a rebasing failure?

rycee: It no longer seems necessary since I removed the default state version in https://github.com/nix-community/home-manager/commit/3c710201d59188754517eca0e6bd89ca34863e9c, which is already in master.

rycee:
I think it is fine for now and even I managed to set up a working flake for a standalone HM configuration 🙂

Thanks! Merged to master now.

PR #807, opened 2019-08-19
Run sudo with -s in the darwin module

toonn:
Currently activation is run with `sudo -i` this defaults to the user's login
shell. This can lead to problems if the user's shell isn't set properly.

By passing `-s` rather than `-i`, `sudo` runs `activate` in `SHELL` instead.
I'm assuming at this point in the activation `SHELL` contains the path to a
bash in the nix store. This should always be a valid shell to run the
`activate` script with.

From the `sudo` manual it seems like this cannot be fixed if `SHELL`
isn't set at this point or by passing a command to `-s` because that
command is then passed to the user's shell.

I had some trouble turning off `useUserPackages` in the darwin module because I'd set my shell manually using `chsh -s $(which fish)`. This set the shell to `/etc/profiles/per-user/${USER}/bin/fish` and that directory disappeared when turning off `useUserPackages` so the HM `activate` script couldn't run because `sudo -i` couldn't find that shell anymore.

I believe my fix makes this work no matter what happens to the user's shell. Note that shell setup, using `chsh` is still required, home-manager doesn't fix this. However, I don't know whether it's important `activate` is run with *a* or specifically the user's login shell.

PR #3040, opened 2022-06-19
Provide SHELL for sudo -s invocation of activation script

### LINKED ISSUE #2900: bug: Activation script fails after `sudo` attempts to use an invalid shell

### Issue description

After updating my `home-manager` input to the current master branch, my activation script appears to be broken on both an intel and m1 machine:

```
setting up groups...
setting up users...
setting up ~/Applications...
applying patches...
setting up /etc...
system defaults...
setting up launchd services...
configuring networking...
configuring keyboard...
configuring fonts...
Activating home-manager configuration for kclejeune
sudo: /usr/bin/bash: command not found
```

I assume this is due to the change introduced by #807, since the failing line is:

```
sudo -u kclejeune -s --set-home /nix/store/2wb1b2kyzlfb5k6mr66di7qi0aldmq0d-activation-kclejeune
```

I've been using home-manager on both m1 and intel machines prior to this change without any activation script issues (including setup from scratch) so I don't think the issue motivating #807 is universal.

### Maintainer CC

@kalbasit @toonn 

### System information

```markdown
 - system: `"x86_64-darwin"`
 - host os: `Darwin 21.4.0, macOS 10.16`
 - multi-user?: `yes`
 - sandbox: `no`
 - channels(kclejeune): `""`
 - channels(root): `"darwin, home-manager, nixpkgs-20.09pre228453.dcb64ea42e6"`
 - nixpkgs: `/etc/nixpkgs`
```

toonn:
### Description

In #807 I changed the flag passed to `sudo` from `-i` to `-s` so `sudo`
wouldn't use a non-existent shell defined in the `passwd` file. kalbasit
also reported in that PR that `-i` didn't work for them anymore on an M1
Mac, presumably because Apple changed something in newer versions of
macOS.

Some users reported that this broke the behavior for them because
`SHELL` was set to a path that didn't even exist on their system. It's
unclear how this came to be but it shows that my assumption that `SHELL`
would be set to a reasonable shell by Home-Manager at this point in the
activation is false.

As a way around this problem we can explicitly set `SHELL` when running
the activation script to a value that we know will be good, like
`${pkgs.bash}`.

One change in behavior this causes is that the activation script will
always be run by bash, not the user's shell. If the script is generated
by Home-Manager this is fine since it can be generated taking into
account the supported set of functions and behaviors. If the intent is
for the activation script to possibly be run by non-bash and even
non-POSIX shells, like tcsh, ksh or Xonsh, then this fix will not
suffice.

Fixes #2900

**Note**: I'm not sure how to test this properly, advice appreciated.

### Checklist

- [] Code formatted with `./format`.
   Couldn't do this because of a hash mismatch:
   ```
   error: ca hash mismatch importing path '/nix/store/rqc038b0jxgifl017wbmqanlp8ybk0k0-source';
         specified: sha256:0rbh9d0fskapgyzy812yxrzd87wj1jlzn46wi7ymyls2bpnbkw1g
         got:       sha256:19kaagb3ljgzhcyxcinqiskjbcasikfdbrrhvy8rj2wrxygcphmz
    ```

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/nix-darwin/default.nix
+++ b/nix-darwin/default.nix
@@ -15,7 +15,7 @@ in {
       system.activationScripts.postActivation.text = concatStringsSep "\n"
         (mapAttrsToList (username: usercfg: ''
           echo Activating home-manager configuration for ${username}
-          SHELL=${pkgs.bash} sudo -u ${username} -s --set-home ${
+          sudo -u ${username} --set-home ${
             pkgs.writeShellScript "activation-${username}" ''
               ${lib.optionalString (cfg.backupFileExtension != null)
               "export HOME_MANAGER_BACKUP_EXT=${
--- a/nix-darwin/default.nix
+++ b/nix-darwin/default.nix
@@ -15,7 +15,7 @@ in {
       system.activationScripts.postActivation.text = concatStringsSep "\n"
         (mapAttrsToList (username: usercfg: ''
           echo Activating home-manager configuration for ${username}
-          sudo -u ${username} -s --set-home ${
+          SHELL=${pkgs.bash} sudo -u ${username} -s --set-home ${
             pkgs.writeShellScript "activation-${username}" ''
               ${lib.optionalString (cfg.backupFileExtension != null)
               "export HOME_MANAGER_BACKUP_EXT=${


nix-darwin/default.nix
```diff
@@ -28,4 +28,3 @@ in {
         '') cfg.users);
     })
   ];
-}
```
rycee, line 31: I assume this is a mistake?

rycee:
LGTM but I think some nix-darwin users should give it a try before merging.

I only have a minor comments about the commit message, could you make the first line a bit shorter? I think something like

> nix-darwin: improve invocation of activation script

should suffice since the rest of the excellent commit message explains things quite clearly.

Also "Home-Manager" should be "Home Manager".

About using bash for the activation script, this is an absolute requirement. We use quite a few bash specific features in the activation script. The `writeShellScript` function should generate a script that will get executed by bash from Nixpkgs. If `SHELL` is set to some odd value at this point, wouldn't it be a problem with the nix-darwin activation script?

nix-darwin/default.nix
toonn: Yes. Failed to check the diff after rebasing : /

toonn:
@rycee, thanks for the rubber ducking, you made me realize that both `-i` and `-s` were misguided and the only crucial part of `-i` was `--set-home`.

toonn:
As far as I'm concerned this is ready for merge. I have no way to test this because I can't reproduce the issue from #2900 but I'm fairly confident this is the proper fix.

rycee:
Yeah, I think enough time has passed. I've merged to master now and we'll see if that produces any additional feedback. Thanks a lot for the fixing!

PR #3042, opened 2022-06-20
firefox: add support for nested bookmarks

### LINKED ISSUE #2577: Allow firefox bookmarks to have folders
### Description
Currently setting Firefox's bookmarks only allows creating bookmarks without folders.

```nix
programs.firefox.profiles."testProfile" = {
  wikipedia = {
    keyword = "wiki";
    url = "https://en.wikipedia.org/wiki/Special:Search?search=%s&go=Go";
  };
  "kernel.org" = {
    url = "https://www.kernel.org";
  };
}
```

I think what would be better is if we could do

```nix
programs.firefox.profiles."testProfile" = {
  "Bookmarks Toolbar" = {
    Wikis = {
      Nix = {
        "Nix package manager" = {
          url = "https://en.wikipedia.org/wiki/Nix_package_manager";
          tags = "Nix,wiki";
        };
        NixOs = {
          url = "https://en.wikipedia.org/wiki/NixOS";
          tags = "NixOS,os,wiki";
        };
      };
    };
    "kernel.org" = {
      url = "https://www.kernel.org";
    };
  };
}
```

to create bookmarks inside the bookmark toolbar with this format

```
.
├── 📂 Wikis
│     └── 📂 Nix
│           ├── 🔖 Nix package manager
│           └── 🔖 NixOs
└── 🔖 kernel.org
```

I am achieve something similar in my setup [here](https://github.com/Th3Whit3Wolf/nixos/blob/main/profiles/users/browsers/firefox/default.nix) that calls [mkBookmarks.nix](https://github.com/Th3Whit3Wolf/nixos/blob/main/profiles/users/browsers/firefox/mkBookmarks.nix)(This is a slightly edited version of [netscape-bookmarks](https://github.com/bahamas10/node-netscape-bookmarks) and passes in the bookmarks as a nix attrset (which get converted to json and becomes a file). However it requires that folders have a key named contents which values are either another folder or the bookmark. Here is the code my current setup would require to achieve the same folder structure as above.

```nix
programs.firefox.profiles."testProfile" = {
  "Bookmarks Toolbar" = {
    contents = {
      Wikis = {
        contents = {
          Nix = {
            contents = {
              "Nix package manager" = {
                url = "https://en.wikipedia.org/wiki/Nix_package_manager";
                tags = "Nix,wiki";
              };
              NixOs = {
                url = "https://en.wikipedia.org/wiki/NixOS";
                tags = "NixOS,os,wiki";
              };
            };
          };
        };
      };
      "kernel.org" = {
        url = "https://www.kernel.org";
      };
    };
  };
}
```

I am going to be very busy with work and moving over the next couple months I will try to write something then to achieve the more ideal API format I mentioned earlier. However if someone wants to beat me to it (and doesn't mind javascript) I would suggest looking at what I have made so far [here again](https://github.com/Th3Whit3Wolf/nixos/blob/main/profiles/users/browsers/firefox/mkBookmarks.nix) which, again, is based on [netscape-bookmarks](https://github.com/bahamas10/node-netscape-bookmarks) and modified slightly to work with nix.

h7x4ABk3g:
### Description

Change type of `firefox.profile.<name>.bookmarks` to
allow for nested bookmarks with user defined order.

The old format is still available.

Closes #2577 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/firefox.nix
+++ b/modules/programs/firefox.nix
@@ -42,7 +42,7 @@ let
 
   mkUserJs = prefs: extraPrefs: bookmarks:
     let
-      prefs' = lib.optionalAttrs ({ } != bookmarks) {
+      prefs' = lib.optionalAttrs ([ ] != bookmarks) {
         "browser.bookmarks.file" = toString (firefoxBookmarksFile bookmarks);
         "browser.places.importBookmarksHTML" = true;
       } // prefs;
@@ -65,13 +65,37 @@ let
         "&gt;"
         "&amp;"
       ];
-      mapper = _: entry: ''
-        <DT><A HREF="${escapeXML entry.url}" ADD_DATE="0" LAST_MODIFIED="0"${
-          lib.optionalString (entry.keyword != null)
-          " SHORTCUTURL=\"${escapeXML entry.keyword}\""
-        }>${escapeXML entry.name}</A>
-      '';
-      bookmarksEntries = lib.attrsets.mapAttrsToList mapper bookmarks;
+
+      indent = level:
+        lib.concatStringsSep "" (map (lib.const "  ") (lib.range 1 level));
+
+      bookmarkToHTML = indentLevel: bookmark:
+        ''
+          ${indent indentLevel}<DT><A HREF="${
+            escapeXML bookmark.url
+          }" ADD_DATE="0" LAST_MODIFIED="0"${
+            lib.optionalString (bookmark.keyword != null)
+            " SHORTCUTURL=\"${escapeXML bookmark.keyword}\""
+          }>${escapeXML bookmark.name}</A>'';
+
+      directoryToHTML = indentLevel: directory: ''
+        ${indent indentLevel}<DT><H3>${escapeXML directory.name}</H3>
+        ${indent indentLevel}<DL><p>
+        ${allItemsToHTML (indentLevel + 1) directory.bookmarks}
+        ${indent indentLevel}</p></DL>'';
+
+      itemToHTMLOrRecurse = indentLevel: item:
+        if item ? "url" then
+          bookmarkToHTML indentLevel item
+        else
+          directoryToHTML indentLevel item;
+
+      allItemsToHTML = indentLevel: bookmarks:
+        lib.concatStringsSep "\n"
+        (map (itemToHTMLOrRecurse indentLevel) bookmarks);
+
+      bookmarkEntries = allItemsToHTML 1
+        (if isAttrs bookmarks then lib.attrValues bookmarks else bookmarks);
     in pkgs.writeText "firefox-bookmarks.html" ''
       <!DOCTYPE NETSCAPE-Bookmark-file-1>
       <!-- This is an automatically generated file.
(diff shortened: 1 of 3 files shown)


modules/programs/firefox.nix
```diff
@@ -42,7 +42,7 @@ let
 
   mkUserJs = prefs: extraPrefs: bookmarks:
     let
-      prefs' = lib.optionalAttrs ({ } != bookmarks) {
+      prefs' = lib.optionalAttrs ([ ] != bookmarks) {
```
rcerc, line 45: Since the `bookmarks` option could also be an attribute set (when using the old structure), this should check that `bookmarks` is not an empty list nor an empty set. The same should probably be done when calling `mkUserJs`.
```suggestion
      prefs' = lib.optionalAttrs (bookmarks != [ ] && bookmarks != { }) {
```

modules/programs/firefox.nix
```diff
@@ -233,37 +257,78 @@ in {
...
+                      description = "Bookmarks within directory.";
+                    };
                   };
+                }) // {
+                  description = "directory submodule";
                 };
-              }));
-              default = { };
+              in with types;
+              either (attrsOf bookmarkType)
```
rcerc, line 304: Would it be feasible to continue using an attribute set and distinguish between bookmarks and folders by inspecting their attributes (instead of using a list) when the user wishes to use both? That seems to involve less migration, but I might be missing something.

tests/modules/programs/firefox/profile-settings.nix
```diff
@@ -15,13 +15,30 @@ lib.mkIf config.test.enableBig {
     profiles.bookmarks = {
       id = 2;
       settings = { "general.smoothScroll" = false; };
-      bookmarks = {
-        wikipedia = {
+      bookmarks = [
```
rcerc, line 18: I think that both test cases should be included to ensure that the old bookmarks-only attribute set structure for the `bookmarks` option continues to work.

rcerc (changes requested):

modules/programs/firefox.nix
h7x4ABk3g: I tried doing that in the start, but attribute sets does not keep any order. We could still model it as an attribute set and let users use the home-manager DAG functionality, but it seems painful if they want to keep a custom order for all of the bookmarks.

modules/programs/firefox.nix
h7x4ABk3g: I'll change this. Thank you

tests/modules/programs/firefox/profile-settings.nix
h7x4ABk3g: That makes sense. I'll add it back in.

modules/programs/firefox.nix
rcerc, line 304: To simplify handling both attribute sets and lists for the `bookmarks` option, `lib.coercedTo` could be used here to convert attribute sets to lists:
```suggestion

                nodeType = types.either bookmarkType directoryType;
              in with types;
              coercedTo (attrsOf nodeType) attrValues (listOf nodeType);
```
Doing this means that `bookmarks` does not have to be converted when calling `allItemsToHTML` nor checked against both an empty list *and* set (i.e. `bookmarks != [ ]` could be used instead of `bookmarks != [ ] && bookmarks != { }`).

rycee:
Thanks! Rebased to master in 64c745fe1c019b724e241a4f3cfbe3c661f54712.

PR #3048, opened 2022-06-22
neovim/coc: fix `withNodeJs` value when enabling coc

addict3d:
### Description

Coc needs nodejs. This changes the enabling logic to use 'boolean or' instead of an 'attr or default' expression.

Currently, nodejs is enabled by a 'select attrpath or default' expression:

> withNodeJs = cfg.withNodeJs or cfg.coc.enable;

However, the option `neovim.withNodeJs` defaults to false and will exist, which means if you only set `neovim.coc.enable = true;` then node will be missing.

I don't know how to test for this from the test suite, but am willing to try if someone can suggest how to proceed.

CC @sumnerevans Looks like a very minor oversight from nix-community/home-manager#2801
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/neovim.nix
+++ b/modules/programs/neovim.nix
@@ -368,7 +368,7 @@ in {
 
     neovimConfig = pkgs.neovimUtils.makeNeovimConfig {
       inherit (cfg) extraPython3Packages withPython3 withRuby viAlias vimAlias;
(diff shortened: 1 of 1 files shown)


sumnerevans (approved):
Ack, I write too much Python haha. Good catch.

PR #3056, opened 2022-06-26
Unflake nmt and nmd

rycee:
### Description

This simplifies the code a bit and avoids using experimental Flake functionality. If Flakes become stable before NixOS 22.11 then we can consider having nmd and nmt as Flake inputs. Maybe could then also avoid the need for flake-compat.

Note, instead of `pkgs.fetchFromGitLab` this commit uses the builtin function `fetchTarball` to avoid the `pkgs` dependency.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,16 +1,6 @@
 { pkgs ? import <nixpkgs> { } }:
 
-let
-
-  flake = (import
-    (let lock = builtins.fromJSON (builtins.readFile ./flake.lock);
-    in fetchTarball {
-      url =
-        "https://github.com/edolstra/flake-compat/archive/${lock.nodes.flake-compat.locked.rev}.tar.gz";
-      sha256 = lock.nodes.flake-compat.locked.narHash;
-    }) { src = ./.; }).defaultNix;
-
-in rec {
+rec {
   docs = with import ./docs { inherit pkgs; }; {
     html = manual.html;
     manPages = manPages;
@@ -26,6 +16,4 @@ in rec {
   nixos = import ./nixos;
 
   path = ./.;
-
-  inherit (flake) inputs;
 }
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -1,9 +1,15 @@
-{ pkgs, nmdSrc ? (import ./.. { }).inputs.nmd
+{ pkgs
 
-  # Note, this should be "the standard library" + HM extensions.
+# Note, this should be "the standard library" + HM extensions.
 , lib ? import ../modules/lib/stdlib-extended.nix pkgs.lib }:
 
 let
+
+  nmdSrc = fetchTarball {
+    url =
+      "https://gitlab.com/api/v4/projects/rycee%2Fnmd/repository/archive.tar.gz?sha=91dee681dd1c478d6040a00835d73c0f4a4c5c29";
+    sha256 = "07szg39wmna287hv5w9hl45wvm04zbh0k54br59nv3yzvg9ymlj4";
+  };
 
   nmd = import nmdSrc { inherit lib pkgs; };
 
--- a/flake.lock
+++ b/flake.lock
@@ -1,21 +1,5 @@
 {
   "nodes": {
-    "flake-compat": {
-      "flake": false,
-      "locked": {
-        "lastModified": 1650374568,
-        "narHash": "sha256-Z+s0J8/r907g149rllvwhb4pKi8Wam5ij0st8PwAh+E=",
-        "owner": "edolstra",
-        "repo": "flake-compat",
-        "rev": "b4a34015c698c7793d592d66adbab377907a2be8",
-        "type": "github"
-      },
-      "original": {
-        "owner": "edolstra",
-        "repo": "flake-compat",
-        "type": "github"
-      }
-    },
     "nixpkgs": {
       "locked": {
         "lastModified": 1654953433,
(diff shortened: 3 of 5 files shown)


default.nix
```diff
@@ -26,6 +16,4 @@ in rec {
   nixos = import ./nixos;
 
   path = ./.;
-
-  inherit (flake) inputs;
```
rycee, line 30: @ncfavier Was the `inputs` here intended as a field for public consumption. I'm a bit weary about removing it now that it has been published but I don't see what purpose it has as a public field so perhaps it is not used much?

ncfavier: No it was internal.

rycee: Sweet. Thanks!

SuperSandro2000:
So essentially we are doing a step backward to a half hearted flake usage and hardcode nmd and nmt somewhere in home-manager?

Not using flakes makes harder to update the two tools and try out custom patches. Right now I can just overwrite the input and rebuild home-manager against another branch of nmd or nmt. After this change I need to do the following steps:
- clone home-manager
- change the path to home-manager to the source code
- change the fetcher of nmd and nmt
- generate the hashes
- rebuild home-manager
- rebuild home-manager again to use the new tools

With flakes I only need to run the following command twice:


``bash
$ home-manager switch --override-input nmd ~/nmd
```


> If Flakes become stable before NixOS 22.11

Flakes development is not tied to any nixos release number.

> Maybe could then also avoid the need for flake-compat.

That will always be required to use flake lock files in none flake code. Essentially it is just some boilerplate code to get the hash out of the flake lock and re-use the downloaded tarball. Pretty clever to be honest.

---

I am very disappointing that you merged this PR as is.

**Comment by ncfavier** [comment]
To be entirely fair, you still couldn't really use `--override-input` because flake-compat does not support that (`--override-input` doesn't modify the lock file). The only thing that would be affected are the `docs-*` flake packages. Using the overridden `nmd` e.g. in the modules would have required some more plumbing. I agree that doing that is probably desirable though.

**Comment by oxalica** [comment]
How could I override nmd inside home-manager with current `fetchurl`? I need [a patch](https://gitlab.com/rycee/nmd/-/merge_requests/9) on nmd to make h-m evaluate since #3570.

**Comment by ncfavier** [comment]
Replace

https://github.com/nix-community/home-manager/blob/684bdb386cec7d4f16e0da9f694c8ab50ad2cf2a/docs/default.nix#L8-L12

with

```nix
  nmdSrc = pkgs.fetchFromGitLab {
    owner = "ncfavier";
    repo = "nmd";
    rev = "no-ifd";
    hash = ""; # TOFU
  };
```

SuperSandro2000:
Do not even overlays work?

ncfavier:
> ```nix
>   nmdSrc = pkgs.fetchFromGitLab {
> ```

Maybe that would cause IFD though, which is precisely the thing you're trying to work around...

In that case try `https://gitlab.com/ncfavier/nmd/-/archive/no-ifd/nmd-no-ifd.tar.gz` in the `fetchTarball`

oxalica:
> Replace
> 
> https://github.com/nix-community/home-manager/blob/684bdb386cec7d4f16e0da9f694c8ab50ad2cf2a/docs/default.nix#L8-L12
> 
> with
> ```
>   nmdSrc = pkgs.fetchFromGitLab {
>     owner = "ncfavier";
>     repo = "nmd";
>     rev = "no-ifd";
>     hash = ""; # TOFU
>   };
> ```

How can I do this without forking `home-manager` and manually rebase it every time? When `nmd` is a flake input, I could use `inputs.nmd.follows` to override it. But now it's very hard to (auto) patch a remote nix file inside `home-manager`.
Related: https://github.com/NixOS/nix/issues/3920

ncfavier:
There are several things we could do, not necessarily mutually exclusive:
- revert this PR and do the thing I was suggesting to propagate flake inputs to the modules
- add a `manual.nmd` (or `nmdSrc`) option to allow overriding nmd modularly

@rycee thoughts?

rycee:
Ultimately I think the idea is to switch over to Flakes … once they move out of experimental status. I'm not sure it's worth spending much effort on making nmd overrideable until then.

What I do while working on nmd is to simply comment out the `nmdSrc` assignment in my local HM clone and set `nmdSrc = ../../nmd;` instead 🙂

SuperSandro2000:
> once they move out of experimental status

IMO do the move now. I've been using flakes for well over a year and I can't remember a major breakage in that time. The lock file format stayed the same and the cli stayed the same. I am using flakes every day and all the nixos infrastructure I am doing is using flakes.

What I can remember is that the nixos manual changed in breaking ways multiple times in the last months and getting in the fix for nmd was always unnecessarily hard because home-manager is not using flakes correct.

And even if flakes have a breaking change eventually, I am sure the manual broke at least once again until that.

> I'm not sure it's worth spending much effort on making nmd overrideable until then.

It is. We already spend more time discussing it than it would have took to implement it. There was even a flake implementation already there that was reverted. Also if NixOS wouldn't be overwritable by default in many places than it would be harder to use and IMO a worse distro. One of NixOS greatest features is it's extensibility.

> What I do while working on nmd is to simply comment out the `nmdSrc` assignment in my local HM clone and set `nmdSrc = ../../nmd;` instead slightly_smiling_face

Compared to the flake steps that would be necessary those are more steps to do and foremost they are imperative. 

For flakes I would just need to add another input and add a follow to that. Maybe the follow could even contain a flake rev, not sure about that right now.

terlar:
I just ran into this issue, I am trying to build Home Manager in a controlled environment where Internet access is on a per allow-list basis. Either I have to add `git.sr.ht` to the allow-list which is not directly within my control or if this was a flake input I could change the source to a fork of nmd easily via input overrides and call it a day.

PR #3062, opened 2022-06-29
keychain: set SHELL correctly in bash

sersorrel:
### Description

bash apparently handles command substitution slightly differently than
fish and zsh. in bash:

```console
$ export FOO=x
$ FOO=y echo $(sh -c 'echo $FOO')
x
```

whereas in zsh:

```console
$ export FOO=x
$ FOO=y echo $(sh -c 'echo $FOO')
y
```

so we have to assign $SHELL within the substitution in bash.

see #2256 and #2880.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/keychain.nix
+++ b/modules/programs/keychain.nix
@@ -100,13 +100,13 @@ in {
   config = mkIf cfg.enable {
     home.packages = [ cfg.package ];
     programs.bash.initExtra = mkIf cfg.enableBashIntegration ''
-      SHELL=bash eval "$(${shellCommand})"
+      eval "$(SHELL=bash ${shellCommand})"
     '';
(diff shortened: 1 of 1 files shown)


berbiche (approved):
Thanks for the contribution!

berbiche:
On my machine, I get the first behavior with zsh 5.9.

sumnerevans:
I also get the first behaviour with zsh 5.9 and bash 5.1.

sersorrel:
huh, testing again, I see that behaviour in zsh too, I wonder what I did wrong first time. I'll make the change for zsh as well.

rycee:
Thanks! Merged to master now.

PR #3068, opened 2022-07-02
zsh: source zsh-syntax-highlighting at the end of .zshrc

maxbrunet:
### Description

`zsh-syntax-highlighting.zsh` must be sourced at the end of the `.zshrc` file, otherwise widgets created after are not properly highlighted (e.g. from oh-my-zsh...): https://github.com/zsh-users/zsh-syntax-highlighting#faq

~I moved `zsh-autosuggestions` with it for style only, I think it is good to keep these 2 plugins grouped, and it should not cause any issue.~

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/zsh.nix
+++ b/modules/programs/zsh.nix
@@ -500,10 +500,6 @@ in
           "source ${pkgs.zsh-autosuggestions}/share/zsh-autosuggestions/zsh-autosuggestions.zsh"
         }
 
-        ${optionalString cfg.enableSyntaxHighlighting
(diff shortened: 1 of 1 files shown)


sumnerevans (approved):
Please rebase on the latest `master` as the build has been fixed there.

I also wonder if we should move `cfg.initExtra` to be the absolute last thing... It's in an awkward place before the aliases right now.

I am not quite convinced that this is entirely backwards compatible (especially if people had custom stuff in `cfg.initExtra`). I'm not sure how much we care about that though tbh.

sumnerevans (changes requested):
Code change looks fine. Please squash your commits and rebase on the latest `master`.

sumnerevans:
Thanks for the contribution!

PR #3069, opened 2022-07-03
swayidle: fix systemd service

### LINKED ISSUE #2811: bug: systemd (user?) units do not include /bin in path by default

### Issue description

swayidle uses `sh` from PATH, so it seems like it doesn't work until I add it in manually.

```diff
+         Environment = "PATH=/bin";
```

Can we "fix" this globally by somehow telling systemd to always include /bin in the path for a service?

### Maintainer CC

_No response_

### System information

```markdown
n/a, latest master of home-manager and nixos-unstable.
```

jneem:
### Description

swayidle executes commands using "sh -c" and so its PATH must contain
a shell. This adds such PATH entry to the environment of the systemd
service.

Fixes #2811, by applying the suggestion in [this comment](https://github.com/nix-community/home-manager/pull/2849#discussion_r846411403)

### Checklist

- [ ] Code tested through `nix-shell --pure tests -A run.all`. *(I tested using `nix-shell --pure tests -A run.swayidle-basic-configuration`; I couldn't run all the tests because the neovim test is failing for me)*

    ```
    {component}: {description}

    {long description}
    ```


sumnerevans:
Please rebase on latest `master`. The build failure has been fixed there already.

sumnerevans (approved):
I think this makes sense. I'll let @berbiche make the final call as he was the one who suggested the change in the first place.

modules/services/swayidle.nix
```diff
@@ -108,6 +108,8 @@ in {
 
       Service = {
         Type = "simple";
+        # swayidle executes commands using "sh -c", so the PATH needs to contain a shell.
+        Environment = [ "PATH=${dirOf pkgs.runtimeShell}" ];
```
berbiche, line 112: Thinking back at my solution, it might be better to simply use `lib.makeBinPath [ pkgs.bash ]` instead of the `dirOf` "hack".
What do you think @jneem @sumnerevans?

sumnerevans: I guess it only needs to have access to to the `sh` command, eh? I think just putting `bash` into the `PATH` would be sufficient then.

jneem: Sure, that seems to work. I've also rebased on latest master.

jneem:
... and I forgot to go through the checklist the second time. Sorry! it should work now...

jneem:
I'm not sure what to do about this latest CI failure. It seems like master suffers from the same problem...

berbiche (approved):

tests/modules/services/swayidle/basic-configuration.nix
```diff
@@ -47,6 +47,9 @@
 
       assertFileExists $serviceFile
       assertFileRegex $serviceFile 'ExecStart=.*/bin/swayidle ${expectedArgs}'
+      assertFileRegex $serviceFile 'Environment=.*PATH=${
+        dirOf pkgs.runtimeShell
+      }'
```
berbiche, line 52: You have to fix the regex here.

berbiche (changes requested):
Oops, I pressed the wrong button before submitting my comment. 
Sorry for the noise.

rycee:
Thanks! Merged to master now.

archer-65:
After these merged commits swayidle service no longer works although active.
Adding swayidle command to sway startup list works perfectly.

PR #983, opened 2020-01-11
mpv: allow string values in scripts list

rycee:

PR #3071, opened 2022-07-04
mpv: prohibit string values in scripts

sersorrel:
### Description

Strings as scripts don't currently work, and probably haven't worked since b567d273.

See #983.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/mpv.nix
+++ b/modules/programs/mpv.nix
@@ -77,7 +77,7 @@ in {
       };
 
       scripts = mkOption {
-        type = with types; listOf (either package str);
+        type = with types; listOf package;
         default = [ ];
         example = literalExpression "[ pkgs.mpvScripts.mpris ]";
         description = ''


sumnerevans:
Maybe worth mentioning that this is basically a pure revert of https://github.com/nix-community/home-manager/pull/983

berbiche:
Another fix would be to map `cfg.scripts` to be a list of packages as defined in https://github.com/nix-community/home-manager/pull/983#issuecomment-1173294624.

Something like the following:
``` nix
let 
  scripts = lib.imap1 (idx: script:
    let name = "home-manager-script-${toString idx}"; in
    ((pkgs.writeTextFile {
      inherit name;
      executable = true;
      destination = "/share/mpv/scripts/${name}.lua";
      text = script;
    }).overrideAttrs (old: {
      passthru.scriptName = "${name}.lua";
    }))
  ) cfg.scripts;
in
pkgs.wrapMpv pkgs.mpv-unwrapped { inherit scripts; }
```

rycee:
I guess we'll merge this now since the strings don't actually work, if I understand it correctly (I don't use scripts myself).

A better solution would probably be to have a good way to write Lua scripts in Nixpkgs. At some point I would like to resurrect https://github.com/NixOS/nixpkgs/pull/124080 and the Lua package seems like a good candidate.

rycee:
Merged now. Apologies to anybody still using the string type. Feel free to make a PR that implements something along @berbiche's suggestion 🙂

PR #3073, opened 2022-07-07
codeowners: cleanup alphabetization

sumnerevans:
### Description

We've tried to keep the CODEOWNERS file alphabetized, but it's drifted some with people accidentally adding to the bottom. This fixes all of those issues.

One question that I have is whether the few at the top were intentionally clustered (they are mainly @rycee). If they need to be clustered together, maybe it's worth adding "sections" to the file and noting the exceptions. Alternatively, we just go with what I have right now which aggressively alphabetizes everything.

Signed-off-by: Sumner Evans <me@sumnerevans.com>

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,6 +2,13 @@
 
 /flake.nix                                            @bqv @kisik21
 
+Makefile                                              @thiagokokada
+
+/modules/config/home-cursor.nix                       @polykernel @league
+
+/modules/config/i18n.nix                              @midchildan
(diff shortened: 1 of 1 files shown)


rycee:
Thanks for the cleanup! I've merged it to master now.

I think a "best effort" alphabetic ordering should be sufficient. I have some vague memory that later entries override earlier entries, hence the wildcart (`*`) being at the top. Beside that I don't think order matter much.

PR #3076, opened 2022-07-08
git: Add option to use difftastic in side-by-side or inline view

Congee:
### Description
Currently there's no way to see a unified view of `git diff` output with difftastic. This PR enables it.

### Checklist

   pandoc-defaults: FAILED
   helix-example-settings: FAILED

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/git.nix
+++ b/modules/programs/git.nix
@@ -265,6 +265,16 @@ in {
             Determines when difftastic should color its output.
           '';
         };
+
+        display = mkOption {
+          type =
+            types.enum [ "side-by-side" "side-by-side-show-both" "inline" ];
+          default = "side-by-side";
+          example = "inline";
+          description = ''
+            Determines how the output displays - in one column or two columns.
+          '';
+        };
       };
 
       delta = {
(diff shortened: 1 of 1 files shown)


rycee:
Thanks! Rebased to master in 0639aa34f1c2e584598c19a38990e81ad2b86ae2 with fixes to the commit message and the `difftCommand` definition.

modules/programs/git.nix
```diff
@@ -476,8 +486,11 @@ in {
       home.packages = [ pkgs.difftastic ];
 
       programs.git.iniContent = let
-        difftCommand =
-          "${pkgs.difftastic}/bin/difft --color ${cfg.difftastic.color} --background ${cfg.difftastic.background}";
+        difftCommand = "${pkgs.difftastic}/bin/difft" + concatStringsSep " " [
```
rycee, line 489: This will produce `…/bin/difft--color …` which will cause an error when git tries to run it.

Congee: thank you!, see https://github.com/nix-community/home-manager/pull/3089

PR #3077, opened 2022-07-08
spectrwm: add module

loicreynier:
### Description

Add [spectrwm](https://github.com/conformal/spectrwm) window manager module.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/services/window-managers/spectrwm.nix
```diff
@@ -0,0 +1,142 @@
...
+      bindings = mkOption {
+        type = types.attrsOf types.string;
+        default = { };
+        example = literalExpression ''
+          {
+            term = "Mod+Return";
+            restart = "Mod+Shift+r";
+            quit = "Mod+Shift+q";
+          }
+        '';
```
berbiche, line 72: Can you use a command as a key? Like "alacritty --some-flag"?

If so, the `programs` option is redundant when you can manage everything with nix expressions (IMO).

modules/misc/news.nix
```diff
@@ -583,6 +583,14 @@ in
           A new module is available: 'services.sctd'.
         '';
       }
+
+      {
+        time = "2022-07-08T17:47:06+00:00";
+        condition = hostPlatform.isLinux;
+        message = ''
+          A new module is available: 'services.xsession.windowManager.spectrwm'.
```
berbiche, line 591:
```suggestion
          A new module is available: 'xsession.windowManager.spectrwm'.
```

berbiche:
Thanks for the contribution!

Clean PR with only a small nit 👍

berbiche (changes requested):
Thanks for the contribution!

Clean PR with only a small nit 👍

modules/services/window-managers/spectrwm.nix
loicreynier: The `action` key such as the `program` identifer must be a single word.

modules/misc/news.nix
loicreynier: Thanks, I missed this.

loicreynier:
Thanks for the review, last forced push should fix it.

modules/services/window-managers/spectrwm.nix
```diff
@@ -0,0 +1,142 @@
...
+            workspace_limit = 5;
+            focus_mode = "manual";
+            focus_close = "next";
+          }
+        '';
+        description = "Spectrwm settings.";
+      };
+
+      bindings = mkOption {
+        type = types.attrsOf types.string;
```
rycee, line 64: Instead of `types.string` use `types.str` which has more predictable behavior:

```suggestion
        type = types.attrsOf types.str;
```

modules/services/window-managers/spectrwm.nix
```diff
@@ -0,0 +1,142 @@
...
+            term = "Mod+Return";
+            restart = "Mod+Shift+r";
+            quit = "Mod+Shift+q";
+          }
+        '';
+        description = "Spectrwm keybindings.";
+      };
+
+      unbindings = mkOption {
+        type = types.listOf types.string;
```
rycee, line 77:
```suggestion
        type = types.listOf types.str;
```

modules/services/window-managers/spectrwm.nix
```diff
@@ -0,0 +1,142 @@
...
+            "MOD+t"
+          ]
+        '';
+        description = ''
+          List of keybindings to disable from default Spectrwm configuration.
+        '';
+      };
+
+      programs = mkOption {
+        type = types.attrsOf types.string;
```
rycee, line 95:
```suggestion
        type = types.attrsOf types.str;
```

modules/services/window-managers/spectrwm.nix
```diff
@@ -0,0 +1,142 @@
...
+          {
+            term = "alacritty";
+            search = "dmenu -ip -p 'Window name/id:';
+          }
+        '';
+        description = "Spectrwm programs variables.";
+      };
+
+      quirks = mkOption {
+        type = types.attrsOf types.string;
```
rycee, line 107:
```suggestion
        type = types.attrsOf types.str;
```

tests/modules/services/window-managers/spectrwm/spectrwm-simple-config.nix
```diff
@@ -0,0 +1,23 @@
...
+  xsession.windowManager.spectrwm = {
+    enable = true;
+    settings = {
+      bar_enabled = true;
+      bar_format = "+S [+R:+I] %a %b %d [%R]";
+      bar_justify = "center";
+      modkey = "Mod4";
+    };
+    programs = { term = "alacritty"; };
+    bindings = { term = "MOD+Shift+Return"; };
```
rycee, line 13: Maybe add some unbinding as well?

rycee:
Thanks! I added a few minor comments.

loicreynier:
Thanks for the review @rycee, I fix string types and added unbindings to the test.

Although the `waybar-systemd-with-graphical-session-target` fails, I don't understand why.

rycee:
Thanks! Rebased to master in a3b778e672e2a24f9307190da320782aa654b712.

The test error is unrelated.

PR #3078, opened 2022-07-09
tests: fix impurity of _module.args.pkgsPath

ShamrockLee:
### Description

This patch fixes the impurity caused by the last element of the above `modules` (i.e. `pkgsModule`) setting `pkgsPath` with `mkDefault` to `<nixpkgs>` if `config.home.stateVersion` is before 20.09.

Considering that this piece of configuration is used for testing, it should be fine to specify pkgsPath directly.

If applied, it will allow the flake-based test  running (provided by by #3024) work without the `--impure` flag.

### Checklist

- [X] Change is backwards compatible. (I should think so)

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/tests/default.nix
+++ b/tests/default.nix
@@ -15,6 +15,10 @@ let
     check = false;
   } ++ [
     {
+      # Bypass <nixpkgs> reference inside modules/modules.nix to make the test
+      # suite more pure.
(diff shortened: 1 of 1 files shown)


rycee:
Merged now. Thanks!

The test error seems to be unrelated to this commit.

PR #3081, opened 2022-07-10
xdg-desktop-entries: allow icon path type

hall:
### Description

The [XDG desktop spec](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s06.html#key-icon) allows an icon value to be an absolute path. Since the `xdg.desktopEntries` config passes the `icon` value, unmodified, to [`makeDesktopIcon`](https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/make-desktopitem/default.nix) a `path` type is valid.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/xdg-desktop-entries.nix
+++ b/modules/misc/xdg-desktop-entries.nix
@@ -38,7 +38,7 @@ let
 
       icon = mkOption {
         description = "Icon to display in file manager, menus, etc.";
-        type = types.nullOr types.str;
+        type = types.nullOr (types.either types.str types.path);
         default = null;
       };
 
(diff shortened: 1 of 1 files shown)


hall:
CI failures appear unrelated to this PR (and are affecting others) so I won't take any action to fix them unless someone says otherwise.

hall:
Ah, I didn't realize they were related in that way; good to know. However, I get the following error when using a path without the changes in the PR:

```
❯ cat home.nix
{
  xdg.desktopEntries.example = {
    exec = "true";
    icon = ./icon.png;
  };
}

❯ home-manager build -f home.nix
error: A definition for option `xdg.desktopEntries.example.icon' is not of type `null or string'. Definition values:
       - In `/home/bryton/test/home.nix': /home/bryton/test/icon.png

❯ home-manager --version
22.05
```

rycee:
Ah yes, you are entirely right! Looking closer the `path` type will allow coercion of paths while `str` does not. So I've rebased this to master in b8bb5f291a4131e94b1a49b15d2735fcff581fa3, thanks!.

PR #3083, opened 2022-07-10
gpg-agent: set Environment to a list

NickCao:
### Description
In my specific setup, due to the merge of https://github.com/nix-community/home-manager/pull/3031, the environment variable `GTK2_RC_FILES` is no longer visible to `gpg-agent.service`, thus the `pinentry` it launches no longer follows the system gtk theme. In order to restore the previous behavior, I set
```nix
systemd.user.services.gpg-agent.Service.Environment = [ "GTK2_RC_FILES=${config.home.sessionVariables.GTK2_RC_FILES}" ];
```
However the `Environment` attr is a string instead of a list in the `gpg-agent` module, causing an evaluation failure. This change restores the possibility to set custom environment variables in this service.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/gpg-agent.nix
+++ b/modules/services/gpg-agent.nix
@@ -271,7 +271,7 @@ in {
           ExecStart = "${gpgPkg}/bin/gpg-agent --supervised"
             + optionalString cfg.verbose " --verbose";
           ExecReload = "${gpgPkg}/bin/gpgconf --reload gpg-agent";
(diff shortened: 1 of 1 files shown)


rycee:
Thanks! Rebased to master in 1e66e035e18ca02d72ebbbc83e4e75fa0acdf1af.

PR #3085, opened 2022-07-10
yt-dlp: add module

marsam:
### Description

[yt-dlp](https://github.com/yt-dlp/yt-dlp) is a command-line tool to download videos from YouTube.com and other sites (youtube-dl fork)
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/yt-dlp.nix
+++ b/modules/programs/yt-dlp.nix
@@ -1,0 +1,47 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.programs.yt-dlp;
+
+in {
+  meta.maintainers = [ maintainers.marsam ];
+
+  options.programs.yt-dlp = {
+    enable = mkEnableOption "yt-dlp";
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.yt-dlp;
+      defaultText = literalExpression "pkgs.yt-dlp";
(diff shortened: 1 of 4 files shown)


modules/programs/yt-dlp.nix
```diff
@@ -0,0 +1,47 @@
...
+    enable = mkEnableOption "yt-dlp";
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.yt-dlp;
+      defaultText = literalExpression "pkgs.yt-dlp";
+      description = "Package providing the <command>yt-dlp</command> tool.";
+    };
+
+    settings = mkOption {
```
rycee, line 22: I would suggest renaming this to `extraConfig` since `settings` typically is reserved for structured configuration à la RFC 42.

marsam: Updated. Thanks!

rycee:
Thanks! I've merged this to master now. I'm wondering, though, if it wouldn't be possible to have a structured `settings` option instead of `extraConfig`. Perhaps the example `extraConfig` could be

``` nix
programs.yt-dlp.settings = {
  embed-thumbnail = true;
  embed-subs = true;
  sub-langs = "all";
  downloader = "aria2c";
  downloader-args = "aria2c:'-c -x8 -s8 -k1M'";
}
```

loicreynier:
Thanks for adding this! I added a structured `settings` option in #3213.

PR #3099, opened 2022-07-19
Fix tests maybe?

ncfavier:
~This reverts commit 8d5b07fc83d579cd196125e698454b4eb4850646.~

~It seems like this commit is [breaking](https://github.com/nix-community/home-manager/runs/7390484468?check_suite_focus=true) the tests in GitHub Actions, where coreutils becomes coreutils-stage4 for reasons unknown.~

### DIFF
--- a/modules/services/window-managers/bspwm/options.nix
+++ b/modules/services/window-managers/bspwm/options.nix
@@ -4,7 +4,11 @@ with lib;
 
 let
 
+  primitive = with types; oneOf [ bool int float str ];
+
   rule = types.submodule {
+    freeformType = with types; attrsOf primitive;
+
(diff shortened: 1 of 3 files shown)


ncfavier:
Nope, nevermind.

rycee:
Yeah, I expect the problem to have arisen due to some change in Nixpkgs, not in HM 😕

ncfavier:
Looks like this could be it https://github.com/NixOS/nixpkgs/pull/169378

ncfavier:
...and it looks like a fix is in the works https://github.com/NixOS/nixpkgs/pull/179961 so this will fix itself eventually.

PR #3112, opened 2022-07-25
firefox: make it possible to have nested folders (for boomarks)

rhoriguchi:
### Description

This makes it possible to have a dir in a dir. This is needed to have a dir in the bookmark toolbar since that is already a dir.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/firefox.nix
+++ b/modules/programs/firefox.nix
@@ -288,7 +288,7 @@ in {
                     };
 
                     bookmarks = mkOption {
-                      type = types.listOf bookmarkType;
+                      type = types.listOf nodeType;
                       default = [ ];
                       description = "Bookmarks within directory.";
                     };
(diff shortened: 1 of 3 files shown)


fufexan (approved):

teto:
I agree in principle, I have a question first: is it possible to mix this declarative approach with imperative bookmarks ?

rhoriguchi:
> I agree in principle, I have a question first: is it possible to mix this declarative approach with imperative bookmarks ?

Yes, it's somewhat possible. Before the bookmark option existed, I had an option like this `home.file.".mozilla/firefox/PROFILE_NAME/bookmarks.html".source = ./bookmarks.html`. If you have a `bookmarks.html` in your Firefox profile on the first startup it will import all of them and you can edit them as you please. You can also use the `browser.bookmarks.file` option which does the same, `bookmarks.html` is just the default.

This module sets [browser.places.importBookmarksHTML](http://kb.mozillazine.org/Browser.places.importBookmarksHTML) to true, that's why on every restart it will import your bookmarks again. Normally it would be changed from `true` to `false` once it imports the bookmarks but since it's pinned to `true` that will never happen.

A way to address this would be to add an optional option that controls if `browser.places.importBookmarksHTML` is used.

berbiche (approved):
Nice!

PR #3119, opened 2022-07-27
gh: add `extensions` option

amesgen:
### Description

Allows to configure [gh extensions](https://cli.github.com/manual/gh_extension). Naturally, the imperative commands such as `gh extension install` and `gh extension remove` then stop working as expected.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/gh.nix
+++ b/modules/programs/gh.nix
@@ -95,6 +95,15 @@ in {
       mkEnableOption "the gh git credential helper for github.com" // {
         default = true;
       };
+
+    extensions = mkOption {
+      type = types.listOf types.package;
+      default = [ ];
+      description = ''
+        gh extensions, see <link xlink:href="https://cli.github.com/manual/gh_extension"/>.
+      '';
+      example = literalExpression "[ pkgs.gh-eco ]";
+    };
   };
 
   config = mkIf cfg.enable {
@@ -106,5 +115,12 @@ in {
     programs.git.extraConfig.credential."https://github.com".helper =
       mkIf cfg.enableGitCredentialHelper
       "${cfg.package}/bin/gh auth git-credential";
+
+    xdg.dataFile."gh/extensions" = mkIf (cfg.extensions != [ ]) {
+      source = pkgs.linkFarm "gh-extensions" (builtins.map (p: {
+        name = p.pname;
+        path = "${p}/bin";
+      }) cfg.extensions);
+    };
   };
 }
--- a/tests/modules/programs/gh/extensions.nix
+++ b/tests/modules/programs/gh/extensions.nix
@@ -1,0 +1,27 @@
+{ config, lib, pkgs, ... }:
+
+{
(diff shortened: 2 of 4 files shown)


modules/programs/gh.nix
```diff
@@ -106,5 +115,12 @@ in {
     programs.git.extraConfig.credential."https://github.com".helper =
       mkIf cfg.enableGitCredentialHelper
       "${cfg.package}/bin/gh auth git-credential";
+
+    xdg.dataFile."gh/extensions" = mkIf (cfg.extensions != null) {
```
rycee, line 119: Perhaps it would suffice to have

```suggestion
    xdg.dataFile."gh/extensions" = mkIf (cfg.extensions != []) {
```

Then the `nullOr` could be removed.

modules/programs/gh.nix
```diff
@@ -106,5 +115,12 @@ in {
...
+      source = toString (pkgs.linkFarm "gh-extensions" (builtins.map (p: {
```
rycee, line 120: I don't think you need `toString` here
```suggestion
      source = pkgs.linkFarm "gh-extensions" (builtins.map (p: {
```

tests/modules/programs/gh/extensions.nix
```diff
@@ -0,0 +1,18 @@
+{ config, lib, pkgs, ... }:
+
+{
+  config = {
```
rycee, line 4: Can do without the `config` attribute.

```suggestion
```

tests/modules/programs/gh/extensions.nix
```diff
@@ -0,0 +1,18 @@
...
+    programs.gh = {
+      enable = true;
+      extensions = [ pkgs.gh-eco ];
```
rycee, line 7: Need to add a stub for `gh-eco` as well to avoid the unnecessary download. Adding

```
test.stubs.gh-eco = {
  buildScript = "mkdir -p $out/bin; touch $out/bin/gh-eco";
};
```

may suffice.

modules/programs/gh.nix
amesgen: I am not necessarily opposed, but did you see my remark from the PR description:

> I made the option nullable and null by default to avoid errors when people (probably on non-NixOS) used `gh extension install` before and hence already have a non-empty folder `.local/share/gh/extensions`.

WDYT?

amesgen: Good point, will change :+1:

tests/modules/programs/gh/extensions.nix
amesgen: Thanks, that makes sense. I had to slightly expand this, in particular, I changed `tests.stubs` to set `pname` in a9ca816d5df484f29d69196de7f85d8688a8a9fd.

rycee:
Thanks! Merged to master now.

PR #3122, opened 2022-07-31
bugfix and slighty improved behavior on the "files" module

niacdoial:
### Description

- fixed a(n assumed) bug where hidden files directly in `$STORE/$HASH-home-manager-files` are not symlinked into the home directory

- allow recursive `home.file` directive to be merged in an existing directory in the home directory (managed files are still "weaker" than already-present files unless `force` is used)
- similarly, add an option (`allowRecursiveMerge`) to have two recursive `home.file` directives to have the same target directory
  (although I'm not sure what is to happend if there is an actually conficting file in those directories)

### Checklist

- [ ] Fully determine behavior
  - what happens when recursive+allowRecursiveMerge is used, and one of the merged sources is a single file?
  - what takes precedence when there is a single file/file (or file/folder) conflict when merging when constructing home-manager-files?

  - Does "backwards compatible" include the fact that users might have relied on what is here assumed to be a bug?
    (todo: fully determine what footguns there would be in doing so)
  - users relying on recursive+force overriding the full target directory might experience surprises: test whether it is the case

  - whoops, probably undone with subsequent commits :')

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/files.nix
+++ b/modules/files.nix
@@ -43,10 +43,18 @@ in
       let
         dups =
           attrNames
-            (filterAttrs (n: v: v > 1)
-            (foldAttrs (acc: v: acc + v) 0
-            (mapAttrsToList (n: v: { ${v.target} = 1; }) cfg)));
-        dupsStr = concatStringsSep ", " dups;
+            (filterAttrs
+             (n: v: (v.count > 1 && v.nonrec > 0))
+             (foldAttrs (acc: v: {
+                 count=acc.count + v.count; nonrec= acc.nonrec+v.nonrec;
+              })
+              {count=0;nonrec=0;}
+              (mapAttrsToList (n: v: {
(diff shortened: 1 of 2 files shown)


niacdoial:
note: rewrote history to make commits messages compliant with present instructions

niacdoial:
…a lot of very embarrassing bugs squashed, but it seems the tests were borked before my own work:
```
> Generating remote plugin manifest
> Error detected while processing function coc#rpc#start_server[13]..coc#util#get_data_home:
> line   15:
> E117: Unknown function: coc#float#create_notification
> remote/host: generated rplugin manifest: /nix/store/vgx09lr1cs9ggv2s587dl83zf33x8qm9-neovim-0.7.0/rplugin.vim
> Generating rplugin.vim failed!
```

niacdoial:
…okay this is just too WIP and awkward to even be a pull request. bye for now.

PR #3123, opened 2022-08-01
hyfetch: add module

lilyinstarlight:
### Description

Adds a simple module for [hyfetch](https://github.com/hykilpikonna/hyfetch), which is in nixpkgs now (NixOS/nixpkgs#182944)

I've not actually contributed a module before and this one is pretty minimal, but I think I covered everything I saw in the contribution docs (except a news item which I'm happy to add if desired). Let me know if I'm missing something though!

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -123,6 +123,9 @@ Makefile                                              @thiagokokada
 /modules/programs/htop.nix                            @bjpbakker
 /tests/modules/htop                                   @bjpbakker
 
+/modules/programs/hyfetch.nix                         @lilyinstarlight
+/tests/modules/programs/hyfetch                       @lilyinstarlight
+
 /modules/programs/i3status.nix                        @JustinLovinger
 
 /modules/programs/i3status-rust.nix                   @workflow
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -632,6 +632,13 @@ in
           A new module is available: 'services.recoll'.
         '';
       }
+
+      {
+        time = "2022-08-01T16:35:28+00:00";
+        message = ''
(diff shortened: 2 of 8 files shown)


rycee:
Thanks! Merged to master now. It is typically best if the person that does the merge adds the news entry, then we avoid unnecessary merge conflicts.

luxus:
it doesn't work with the default config. after the setup assistant it runs into a logical permission error.
the defaults should have settings populated

lilyinstarlight:
> it doesn't work with the default config. after the setup assistant it runs into a logical permission error. the defaults should have settings populated

I'm not at my computer at the moment (so can't test), but to clarify are you just doing `{ programs.hyfetch.enable = true; }` and getting the error?

luxus:
yes, because the default is a `{}` empty file it tries to setup with the setup assistant and cannot write the file at the end
with this it works:
```
  programs.hyfetch = {
    enable = true;
    settings = {
      preset = "rainbow";
      mode = "rgb";
      color_align = {
        mode = "horizontal";
      };
    };
  };
```

lilyinstarlight:
I'll make a PR when I get home to have it not write the config file if settings is empty (I might default it to `null` for that case and add a conditional)

Thank you for letting me know!

rycee:
Something like

```
xdg.configFile."hyfetch.json" = mkIf (cfg.settings != {}) {
  source = jsonFormat.generate "hyfetch.json" cfg.settings;
};
```

may work.

lilyinstarlight:
Yep that's roughly what I was thinking @rycee. I'll make the change later and add a test for it too when I'm at my computer again in a few hours

PR #3129, opened 2022-08-06
bashmount: add module

AndersonTorres:
### Description

As my first contribution, a straightforward module.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -646,6 +646,13 @@ in
           A new module is available: 'programs.wezterm'.
         '';
       }
+
+      {
+        time = "2022-08-08T16:11:22+00:00";
+        message = ''
+          A new module is available: 'programs.bashmount'.
+        '';
+      }
     ];
   };
 }
--- a/modules/programs/bashmount.nix
+++ b/modules/programs/bashmount.nix
@@ -1,0 +1,34 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.programs.bashmount;
+
+in {
+  meta.maintainers = [ maintainers.AndersonTorres ];
+
+  options.programs.bashmount = {
+    enable = mkEnableOption "bashmount";
+
+    extraConfig = mkOption {
+      type = types.lines;
+      default = "";
+      description = ''
+        Configuration written to
+        <filename>$XDG_CONFIG_HOME/bashmount/config</filename>. Look at
+        <link xlink:href="https://github.com/jamielinux/bashmount/blob/master/bashmount.conf" />
+        for explanation about possible values.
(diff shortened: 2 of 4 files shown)


modules/programs/bashmount.nix
```diff
@@ -0,0 +1,34 @@
...
+
+  cfg = config.programs.bashmount;
+
+in {
+  meta.maintainers = [ maintainers.AndersonTorres ];
+
+  options.programs.bashmount = {
+    enable = mkEnableOption "bashmount";
+
+    settings = mkOption {
```
rycee, line 15: The `settings` name is typically reserved for structured configuration à la [RFC 42](https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md#part-1-structural-settings-instead-of-stringly-extraconfig). For stringly configuration we keep using `extraConfig`:

```suggestion
    extraConfig = mkOption {
```

modules/misc/news.nix
```diff
@@ -639,6 +639,13 @@ in
           A new module is available: 'programs.hyfetch'.
         '';
       }
+
+      {
```
rycee, line 643: No need to add a news entry, the merger does that. Then you won't risk merge conflicts.

AndersonTorres: I need to re-read that part of manual, it looked ambiguous to me.

AndersonTorres: Done!

modules/programs/bashmount.nix
```diff
@@ -0,0 +1,34 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.programs.bashmount;
+
+in {
+  meta.maintainers = [ maintainers.AndersonTorres ];
```
AndersonTorres, line 10: Just to know if I understood:

I am already a maintainer of Nixpkgs. Will HM recognize me here?

modules/programs/bashmount.nix
AndersonTorres: I have used `rtorrent.nix` as example. Is there any difference?

modules/programs/bashmount.nix
rycee: Yes, it will. There is also a step in the CI pipeline that checks that the maintainer is a valid reference, so since the pipeline passed you are recognized.

modules/programs/bashmount.nix
rycee: No different, I think the option in the rtorrent module should have been `extraConfig` as well. Just some mistake.

rycee:
Looks good! Merged to master now.

PR #3141, opened 2022-08-13
files: allow target conflicts between recursively-linked directories (exclusively)

niacdoial:
### Description

Loosened the "no two home.file directives may have the same target" constraint:
Multiple directives may share a target if and only if they are recursively-linked directories

### Checklist

  note: one bout of weirdness is the use of `pathExists "${sourceStorePath v}/.."` to detect if the source path is a directory. This might warrant a double check for cross-platform compatibility

  note: due to the large changes caused by this, formatting was done in a separate, reproducible commit.

   (note: the line in the tests about neovim was disabled, as that test failed both before and after my changes. this is however not committed)

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/files.nix
+++ b/modules/files.nix
@@ -43,17 +43,19 @@ in
       let
         dups =
           attrNames
-            (filterAttrs (n: v: (v.count > 1 && v.nonrec))
+            (filterAttrs (n: v: ((v.count > 1) && (v.nonrec || v.hasfile) ))
             (foldAttrs (v: acc: {
                 count = add acc.count v.count;
                 nonrec = acc.nonrec || v.nonrec;
+		hasfile = acc.hasfile || v.hasfile;
           }) {
             count = 0;
-            nonrec = false;
+            nonrec = false; hasfile = false;
           } (mapAttrsToList (n: v: {
             ${v.target} = {
               count = 1;
               nonrec = (!v.recursive);
+	      hasfile = !(pathExists "${sourceStorePath v}/..");
             };
           }) cfg)));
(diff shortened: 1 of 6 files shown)


niacdoial:
If you saw the previous (closed WIP) pull request, this is a simpler version, which doesn't re-fix what was already fixed in the lest few months[?].

Also, I'm no longer trying to understand how `force` interacts with `recursive`, as there is no apparent breaking change in that area here.
`force` seems to straight-up not work when `recursive` is active, both on per-file conflict and in terms of overriding the full directory, whatever the "ideal" behavior of that interaction is supposed to be

rycee:
Thanks for the contribution! I was having a look at this and in general it seems fine. But I do feel a bit reluctant to make `source` a mergeable option since it seems quite unintuitive to me. If I needed to merge source directories and link them recursively then the first thing that comes to mind is to use `symlinkJoin` like so

```nix
home.file."test-recursive-joined" = {
  source = pkgs.symlinkJoin {
    name = "test-recursive-joined";
    paths = [ ./test1 ./test2 ];
  };
  recursive = true;
};
```

Do you really need the mergeability or is the `symlinkJoin` solution sufficient?

niacdoial:
There would probably be some usecases where being able to have separate directives (if they are in different files, which have completely different purposes, maybe?)
…But I have to admit I outright didn't know about `symlinkJoin`, and it would fit my usecase nicely.
Should I do a pull request that simply mentions that function in the docs for `home.file` instead?

niacdoial:
…@rycee should I?

PR #3149, opened 2022-08-15
clipmenu: add launcher option

loicreynier:
### Description

This adds an option to set the launcher command for Clipmenu (which is set with the `CM_LAUNCHER` session variable).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/clipmenu.nix
+++ b/modules/services/clipmenu.nix
@@ -18,6 +18,16 @@ in {
       defaultText = "pkgs.clipmenu";
       description = "clipmenu derivation to use.";
     };
+
+    launcher = mkOption {
+      type = types.nullOr types.str;
+      default = null;
+      example = "rofi";
+      description = ''
+        Launcher command, if not set, <command>dmenu</command>
+        will be used by default.
+      '';
+    };
   };
 
   config = mkIf cfg.enable {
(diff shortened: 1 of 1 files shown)


modules/services/clipmenu.nix
```diff
@@ -18,6 +18,15 @@ in {
       defaultText = "pkgs.clipmenu";
       description = "clipmenu derivation to use.";
     };
+
+    launcher = mkOption {
+      type = types.nullOr types.str;
+      default = null;
```
sumnerevans, line 24:
```suggestion
      default = null;
      example = "rofi";
```

sumnerevans (changes requested):
Looks good otherwise!

loicreynier:
Thanks! Just added the example.

modules/services/clipmenu.nix
```diff
@@ -28,6 +38,9 @@ in {
 
     home.packages = [ cfg.package ];
 
+    home.sessionVariables =
+      mkIf (cfg.launcher != null) { CM_LAUNCHER = cfg.launcher; };
```
berbiche, line 42: If this environment variable is needed by the daemon, then it should be set in the Service Environment; otherwise, it won't be available unless imported (`systemctl --user import-environment CM_LAUNCHER`).

The benefit of setting it there is that it should cause a restart of the service if the value changes.

loicreynier: The environment variable is not needed by the daemon but only by the CLI launcher. It can be changed without restarting the service.

rycee:
Thanks! Merged to master now.

PR #3150, opened 2022-08-16
aerc: improve module

### LINKED ISSUE #3621: bug: programs.aerc.enable seems to always create binds.conf

### Issue description

The `aerc` module seems to unconditionally crate an almost empty file at `~/.config/aerc/binds.conf` even when no binds are set. This overwrites all the default bindings.

### Maintainer CC

@lukasngl

### System information

```markdown
 - system: `"x86_64-linux"`
 - host os: `Linux 6.1.7-200.fc37.x86_64, Fedora Linux, 37.20230123.0 (Kinoite), nobuild`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.12.0`
 - channels(root): `"nixpkgs"`
 - nixpkgs: `/var/home/user/.nixpkgs`
```
```

### PR DESCRIPTION
**Pull request by lukasngl**
Follow-up to  #3070.
  - Added empty lines after definitons
  - Improved warning realted to `extraConfig.general.unsafe-accounts-conf = true;`
  - Only generate files if they would have content, i.e. the respective config has non-empty values
  - Let values from `<account>.aerc.extraAccounts` take precedence over all derived values

Closes #3621

### CODE REVIEW

**Comment by luxus** [comment]
is it ready for merge? then you have to change from draft

File: modules/programs/aerc.nix [hunk 1]
```diff
@@ -102,64 +109,87 @@ in {
...
+
+    accountsExtraAccounts = mapAttrs accounts.mkAccount aerc-accounts;
+
+    accountsExtraConfig = mapAttrs accounts.mkAccountConfig aerc-accounts;
+
+    accountsExtraBinds = mapAttrs accounts.mkAccountBinds aerc-accounts;
+
+    joinContextual = contextual: joinCfg (map mkINI (attrValues contextual));
+
+    recursivelyEmpty = x:
```
berbiche, line 135:
```suggestion
    isRecursivelyEmpty = x:
```
IMO this makes more sense, but it's up to you.

berbiche (approved):
Thanks for the clean up.

modules/programs/aerc.nix
lukasngl: Thanks, its definitely more sensible.

vbrandl:
I have the problem, that home-manager creates an empty `binds.conf`, effectively disabling all keybindings. Does this PR fix that issue?

lukasngl:
> I have the problem, that home-manager creates an empty binds.conf, effectively disabling all keybindings. Does this PR fix that issue?
Yes, e36acb9 should adress this issue.

> Edit: A similar problem exists for aerc.conf when setting programs.aerc.extraConfig.general.unsafe-accounts-conf = true. aerc.conf is created with only the line unsafe-accounts-conf = true, disabling default filters for e.g. text/plain.
If your config was empty, the module (with this PR merged) would not try to create the aerc.conf.

vbrandl:
I think even with an non-empty config, the default values should also be written to `aerc.conf` and `binds.conf`.

Also the accounts in `accounts.conf` are currently sorted alphabetically. The first account in `accounts.conf` is opened by default when starting aerc. Placing the primary account on top would make it the default mailbox when opening aerc.

lukasngl:
> I think even with an non-empty config, the default values should also be written to `aerc.conf` and `binds.conf`.

Well, that would be nice. How would you do it?

> Also the accounts in `accounts.conf` are currently sorted alphabetically. The first account in `accounts.conf` is opened by default when starting aerc. Placing the primary account on top would make it the default mailbox when opening aerc.

Good point, fixed in a096894e.

> I also just sent a patch (https://lists.sr.ht/~rjarry/aerc-devel/patches/36790) to make the supplied filters work nice with NixOS.

The patch probably won't be accepted, see for example nix-shell(1) man page:
> The lines starting with #! nix-shell specify nix-shell options (see above). Note that you cannot write  #!  /usr/bin/env nix-shell -i ...  because many operating systems only allow one argument in #! lines.

You could run the filters as `${pkgs.awk}/bin/awk -f ${pkgs.aerc}/share/aerc/filters/colorize`.

vbrandl:
> > I think even with an non-empty config, the default values should also be written to `aerc.conf` and `binds.conf`.
> 
> Well, that would be nice. How would you do it?

While it isn't pretty, defining a set for the default `binds.conf` and
`aerc.conf` and merging it with the user supplied settings might work...

vbrandl:
I agree that anything related to the default configuration is out of scope for this PR.

Just my thoughts:

Not generating `binds.conf` if the extraBinds configuration is empty solves the problem for the happy path but if one were to configure some `extraBinds`, a file only containing the defined binds will be created, removing all default bindings. I think at least the key should be renamed to just `binds` since those binds wouldn't be 'extra' as in 'additional'.

Also since `unsafe-accounts-conf` must be set, the `extraConfig` set is never empty, so I end up with `aerc.conf` with only `unsafe-accounts-conf` set to true. This means there is no `[filters]` section and messages cannot be viewed. Having to duplicate the complete configuration for `aerc.conf` in home-manager only to get the default behaviour seems wierd to me...

I'm not sure how this could be fixed nicely. Making aerc itself fallback to default values for undefined keys in the configuration seems to be the cleanest solution to me.

sumnerevans:
This may need a rebase on the latest `master` to get the tests working

rodrgz (approved):

jleightcap:
Hello!

Is there any status PR? Unfortunately `aerc` isn't particularly usable, IMO, without some of this PR's fixes. if it's not just awaiting review I'm happy to continue work on this.

Thanks :slightly_smiling_face:

PR #3151, opened 2022-08-17
nheko: add module

gvolpe:
### Description

Add module for the [nheko](https://github.com/Nheko-Reborn/nheko) program.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


tests/default.nix
```diff
@@ -95,6 +95,7 @@ import nmt {
     ./modules/programs/newsboat
     ./modules/programs/nix-index
     ./modules/programs/nnn
+    ./modules/programs/nheko
```
sumnerevans, line 98: "h" comes before "i" so this should be between newsboat and nix-index

modules/programs/nheko.nix
```diff
@@ -0,0 +1,83 @@
...
+let
+  cfg = config.programs.nheko;
+
+  iniType = with types;
+    let
+      primitiveType = either str (either bool (either int float));
+      multipleType = either primitiveType (listOf primitiveType);
+      sectionType = attrsOf multipleType;
+      supersectionType = attrsOf (either multipleType sectionType);
+    in attrsOf supersectionType;
```
sumnerevans, line 14: This looks like a pretty standard INI file... we should be using `pkgs.formats.ini`

(See `modules/services/pantalaimon.nix` as an example)

modules/programs/nheko.nix
```diff
@@ -0,0 +1,83 @@
...
+            animateImagesOnHover = false;
+            "sidebar\\roomListWidth" = 308;
+          };
+        }
+      '';
+      description = ''
+        Attribute set of Nheko preferences (converted to INI file).
+
+        For now, it is recommended to run nheko and sign-in before filling in the configuration
+        settings in this module, as nheko writes the access token to $HOME/.config/nheko/nheko.conf
```
sumnerevans, line 60:
```suggestion
        settings in this module, as nheko writes the access token to <filename>$HOME/.config/nheko/nheko.conf</filename>
```

sumnerevans (changes requested):

tests/default.nix
gvolpe: damn, a hiccup 🥲

modules/programs/nheko.nix
gvolpe: Oh nice, wasn't aware of that, I'll check it out!

gvolpe:
@sumnerevans thanks for the review! I believe I addressed all your points and even fixed the tests, as the `nmt.script` line was in the wrong place in one of the tests 😅

gvolpe:
`nheko` seems to be marked as broken for MacOS, anything I should do here?

modules/programs/nheko.nix
```diff
@@ -0,0 +1,82 @@
...
+
+        For now, it is recommended to run nheko and sign-in before filling in the configuration
+        settings in this module, as nheko writes the access token to <filename>$XDG_CONFIG_HOME/nheko/nheko.conf</filename>
+        the first time we sign in, and we need that data into these settings for the correct
+        functionality of the application.
+
+        This a temporary inconvenience, however, as nheko has plans to move the authentication
+        stuff into the local database they currently use. Once this happens, this will no
+        longer be an issue.
+      '';
```
berbiche, line 66:
```suggestion
      description = ''
        Attribute set of Nheko preferences (converted to INI file).
        </para><para>
        For now, it is recommended to run nheko and sign-in before filling in the configuration
        settings in this module, as nheko writes the access token to <filename>$XDG_CONFIG_HOME/nheko/nheko.conf</filename>
        the first time we sign in, and we need that data into these settings for the correct
        functionality of the application.
        </para><para>
        This a temporary inconvenience, however, as nheko has plans to move the authentication
        stuff into the local database they currently use. Once this happens, this will no
        longer be an issue.
      '';
```

modules/programs/nheko.nix
```diff
@@ -0,0 +1,82 @@
...
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    home.file."${configDir}/nheko/nheko.conf" = mkIf (cfg.config != { }) {
+      text = ''
+        ; Generated by Home Manager.
+
+        ${generators.toINI { } (attrsKeyToSnakeCase cfg.config)}
```
berbiche, line 77:
```nix
generators.toINI {
  mkKeyValue = k: v: mkKeyValueDefault {} "=" (camelCaseToSnakeCase k) v;
} cfg.config
```

tests/modules/programs/nheko/nheko-example-settings.nix
```diff
@@ -0,0 +1,80 @@
+{ pkgs, ... }:
+
+{
+  config = {
```
berbiche, line 4:
```suggestion
{ ... }:

{
```

tests/default.nix
```diff
@@ -93,6 +93,7 @@ import nmt {
     ./modules/programs/ne
     ./modules/programs/neomutt
     ./modules/programs/newsboat
+    ./modules/programs/nheko
```
berbiche, line 96: > `nheko` seems to be marked as broken for MacOS, anything I should do here?

The package needs to be stubbed in the tests.

tests/modules/programs/nheko/nheko-empty-settings.nix
```diff
@@ -0,0 +1,11 @@
+{ pkgs, ... }:
+
+{
+  config = {
+    programs.nheko.enable = true;
```
berbiche, line 5:
```suggestion
    programs.nheko.enable = true;
    
    test.stubs.nheko = { };
```

tests/modules/programs/nheko/nheko-example-settings.nix
```diff
@@ -0,0 +1,80 @@
...
+
+        window = {
+          height = 482;
+          width = 950;
+        };
+      };
+
+    };
+
+    nmt.script = ''
```
berbiche, line 74:
```suggestion
    };
    
    test.stubs.nheko = { };

    nmt.script = ''
```

berbiche (changes requested):
Thanks for the contribution!

modules/programs/nheko.nix
gvolpe: Aren't those tags in reverse order? I see a closing tag first, then an opening tag, not sure if that's correct.

modules/programs/nheko.nix
gvolpe: Could you explain why this longer version would be preferred?

tests/default.nix
gvolpe: Ah that's what those stubs were for! I had them before but removed them because the tests were passing anyway, but of course, I'm on NixOS not on Mac 😄

modules/programs/nheko.nix
gvolpe: Oh I guess we'd get rid of the `attrsKeyToSnakeCase` intermediate function, making it shorter in the end, is that correct?

modules/programs/nheko.nix
gvolpe: Added them anyway, as suggested.

modules/programs/nheko.nix
gvolpe: I think I got it. It's working now as suggested.

gvolpe:
Thanks for your review, @berbiche ! Left one or two questions, but I addressed all your points.

modules/programs/nheko.nix
```diff
@@ -0,0 +1,87 @@
...
+
+  iniFmt = pkgs.formats.ini { };
+
+  configDir = if pkgs.stdenv.hostPlatform.isDarwin then
+    "Library/Application Support"
+  else
+    config.xdg.configHome;
+
+  camelCaseToSnakeCase =
+    replaceStrings upperChars (map (s: "_${s}") lowerChars);
```
rycee, line 16: We typically don't bother doing this for RFC-42 type `settings` options. That is, having 

    general.disable_certificate_validation = false

is OK.

gvolpe: Can we have it this way then? Or RFC-42 prohibits this? I really dislike snake case on nix settings by default 😅

modules/lib/maintainers.nix
```diff
@@ -305,4 +305,10 @@
     github = "lukasngl";
     githubId = 69244516;
   };
+  gvolpe = {
+    name = "Gabriel Volpe";
+    email = "volpegabriel@gmail.com";
+    github = "gvolpe";
+    githubId = 443978;
+  };
```
rycee, line 313: Since you are already in Nixpkgs' `maintainers/maintainer-list.nix` file you don't need to be here:

```suggestion
```

modules/programs/nheko.nix
```diff
@@ -0,0 +1,87 @@
...
+
+  inherit (generators) mkKeyValueDefault toINI;
+in {
+  meta.maintainers = [ hm.maintainers.gvolpe ];
```
rycee, line 20:
```suggestion
  meta.maintainers = [ maintainers.gvolpe ];
```

modules/programs/nheko.nix
rycee: Typically, the description of `settings` options contain an upstream link to where the user can find information about available program options. That upsteam documentation would show the snake-case names so it may confuse a user when they see non-snake-case names here.

But, I have no strong feeling about it so you can keep the `camelCaseToSnakeCase` if you want. As long as it is unambiguous, i.e., nheko doesn't have any option that contains a uppercase character.

gvolpe: I'm afraid all these settings are very much under-documented, I had to dig them up, so no documentation link...

gvolpe:
@rycee thanks for your review, addressed all your points 👍🏽

gvolpe:
Hi folks, appreciate if anyone could hit the "Approve and run" button for the CI jobs, and perhaps give it a final review. I believe I addressed all your points, thank you @rycee @berbiche @sumnerevans  :)

rycee:
Nice, merged to master now 🙂

PR #3156, opened 2022-08-19
zsh: add option for `zsh-history-substring-search`

CardboardTurkey:
### Description

Add an option for installing [zsh-history-substring-search](https://github.com/zsh-users/zsh-history-substring-search) outside of `prezto`.

I added it as a submodule of `zsh` but perhaps it would make more sense within the `zsh.history` module?

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/zsh.nix
+++ b/modules/programs/zsh.nix
@@ -188,6 +188,28 @@ let
     };
   };
 
+  historySubstringSearchModule = types.submodule {
+    options = {
+      enable = mkEnableOption "history substring search";
+      searchUpKey = mkOption {
+        type = types.str;
+        default = "^[[A";
+        description = ''
+          The key code to be used when searching up.
+          The default of <literal>^[[A</literal> corresponds to the UP key.
+        '';
+      };
+      searchDownKey = mkOption {
+        type = types.str;
+        default = "^[[B";
+        description = ''
+          The key code to be used when searching down.
+          The default of <literal>^[[B</literal> corresponds to the DOWN key.
+        '';
+      };
+    };
+  };
+
 in
 
 {
@@ -293,6 +315,12 @@ in
       enableSyntaxHighlighting = mkOption {
         default = false;
         description = "Enable zsh syntax highlighting";
+      };
+
+      historySubstringSearch = mkOption {
+        type = historySubstringSearchModule;
+        default = {};
+        description = "Options related to zsh-history-substring-search.";
       };
 
       history = mkOption {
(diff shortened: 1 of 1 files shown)


modules/programs/zsh.nix
```diff
@@ -555,10 +575,18 @@ in
         ${dirHashesStr}
 
         ${optionalString cfg.enableSyntaxHighlighting
-          # Load zsh-syntax-highlighting last, after all custom widgets have been created
+          # Load zsh-syntax-highlighting after all custom widgets have been created
           # https://github.com/zsh-users/zsh-syntax-highlighting#faq
           "source ${pkgs.zsh-syntax-highlighting}/share/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh"
         }
+        ${optionalString cfg.historySubstringSearch.enable
```
berbiche, line 582: The default value for `historySubstringSearch` is an empty attribute set so the `enable` does not always exist.
```suggestion
        ${optionalString (cfg.historySubstringSearch.enable or false)
```

modules/programs/zsh.nix
```diff
@@ -188,6 +188,20 @@ let
     };
   };
 
+  historySubstringSearchModule = types.submodule {
+    options = {
+      enable = mkEnableOption "enable";
+      upAndDownKeys = mkOption {
+        default = [ "^[[A" "^[[B" ];
+        type = types.listOf types.str;
```
berbiche, line 196: Can you split this option?
Using a list is opaque for this sort of option and does not work well if you want to override only one keybind.
It also leads to weird logic like the `elemAt` in the `config` section.

berbiche (changes requested):
Thanks for the contribution.

I don't know zsh and this module well enough to have an informed opinion but the addition does not feel substantive to me.

CardboardTurkey:
Thanks a lot for the review! Changes implemented.

> the addition does not feel substantive to me

I don't think I understand what you mean by substantive here. Note the substring package has to be activated _after_ the syntax-highlighting one. Without this PR I don't see how a user can use both packages and also achieve this.

berbiche:
> Note the substring package has to be activated _after_ the syntax-highlighting one. Without this PR I don't see how a user can use both packages and also achieve this

Thanks for the explanation 👍

modules/programs/zsh.nix
```diff
@@ -188,6 +188,28 @@ let
...
 
+  historySubstringSearchModule = types.submodule {
+    options = {
+      enable = mkEnableOption "enable";
+      searchUpKey = mkOption {
+        type = types.str;
+        default = "^[[A";
+        description = ''
+          The key code to be used when searching up.
+          The default of ^[[A corresponds to the UP key.
```
berbiche, line 199:
```suggestion
          The default of <literal>^[[A</literal> corresponds to the UP key.
```

modules/programs/zsh.nix
```diff
@@ -188,6 +188,28 @@ let
...
+        '';
+      };
+      searchDownKey = mkOption {
+        type = types.str;
+        default = "^[[B";
+        description = ''
+          The key code to be used when searching down.
+          The default of ^[[B corresponds to the DOWN key.
```
berbiche, line 207:
```suggestion
          The default of <literal>^[[B</literal> corresponds to the DOWN key.
```

berbiche (changes requested):
Small nits

modules/programs/zsh.nix
```diff
@@ -188,6 +188,28 @@ let
     };
   };
 
+  historySubstringSearchModule = types.submodule {
+    options = {
+      enable = mkEnableOption "enable";
```
berbiche, line 193:
```suggestion
      enable = mkEnableOption "history substring search";
```

berbiche (approved):
Sorry, I also missed this, otherwise lgtm.

CardboardTurkey:
Fix applied!

sumnerevans (approved):

CardboardTurkey:
Is there anything blocking this being merged?

berbiche:
Sorry, thanks again!

PR #3158, opened 2022-08-19
tmux: mouse support

cor:
### Description

Allows you to enable mouse support in tmux

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/tmux.nix
+++ b/modules/programs/tmux.nix
@@ -94,6 +94,7 @@ let
       bind-key -N "Kill the current pane" x kill-pane
     ''}
 
+    set  -g mouse             ${boolToStr cfg.mouse}
     setw -g aggressive-resize ${boolToStr cfg.aggressiveResize}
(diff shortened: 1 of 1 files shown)


modules/programs/tmux.nix
```diff
@@ -204,6 +205,14 @@ in {
         description = "VI or Emacs style shortcuts.";
       };
 
+      mouse = mkOption {
+        default = false;
+        example = true;
+        description = ''
+          Mouse support.
+        '';
+      };
```
berbiche, line 214: This can be simplified:
```suggestion
      mouse = mkEnableOption "mouse support";
```

berbiche (changes requested):
Thanks for the contribution!

modules/programs/tmux.nix
cor: Ah this is much nicer, thanks!

berbiche:
Hi,
Can you update the tmux tests to reflect your change?
Thanks

PR #3168, opened 2022-08-22
neovim runtime

teto:
- programs.neovim: update comments/cleanup code
- programs.neovim: add PLUGIN.runtime support

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


andOrlando:
Hi, it seems like this commit may have killed my neovim installation. As of rebuilding yesterday, neovim.nix in the home-manager directory matches up with the one in this merge and `plugins` seems to just not work. With this configuration in home-manager the standard keybinds for commenting with vim-commentary don't work, leading me to believe the plugin didn't get loaded. Furthermore, when changing the plugins in plugins I see no message about neoivm having to rebuild itself when rebuilding my config. Any idea what could be going wrong?

```nix
programs.neovim = {
  enable = true;
  plugins = with pkgs.vimPlugins; [ vim-commentary ];
};
```

teto:
hum I dont see that here, could you check the generated vimrc ? visible in tail $(readlink -f $(which nvim)). also your init.vim. what if you bind your keys at runtime ?

andOrlando:
Result of `tail $(readlink -f $(which nvim))`
```
LUA_PATH=${LUA_PATH#';'}
LUA_PATH=${LUA_PATH%';'}
export LUA_PATH
LUA_CPATH=${LUA_CPATH:+';'$LUA_CPATH';'}
LUA_CPATH=${LUA_CPATH/';''/nix/store/k27ppkfykx4j112hci1nkyz3dd12g28k-luajit-2.1.0-2022-04-05-env/lib/lua/5.1/?.so'';'/';'}
LUA_CPATH='/nix/store/k27ppkfykx4j112hci1nkyz3dd12g28k-luajit-2.1.0-2022-04-05-env/lib/lua/5.1/?.so'$LUA_CPATH
LUA_CPATH=${LUA_CPATH#';'}
LUA_CPATH=${LUA_CPATH%';'}
export LUA_CPATH
exec -a "$0" "/nix/store/wkaaff9wj1bg58xjara8ypxl89n37p22-neovim-unwrapped-0.7.2/bin/nvim"  '--cmd' 'let g:loaded_node_provider=0' '--cmd' 'let g:loaded_python_provider=0' '--cmd' 'let g:python3_host_prog='\''/nix/store/ji1gl53i1lyawqby1nblvmkg2r80gjhr-neovim-0.7.2/bin/nvim-python3'\''' '--cmd' 'let g:ruby_host_prog='\''/nix/store/ji1gl53i1lyawqby1nblvmkg2r80gjhr-neovim-0.7.2/bin/nvim-ruby'\''' "$@" 
```
init.vim just sources a file that has all my plugins in it. I don't understand what you mean "what if you bind your keys at runtime."

teto:
You should see a  '--cmd' 'set packpath^=/nix/store/gkwkj4nn5sd5d6an33990s3pg2m9hr85-vim-pack-dir' in the wrapper. Do you by any chance mix a stable nixos with an unstable home-manager ? that would explain why it fails for you

andOrlando:
That is most certainly what happened. Thanks for your help!

ELD:
@teto Curious: how does mixing unstable home-manager and stable NixOS cause the issue? I looked at both the home-manager file and the NixOS one and couldn't find where an adverse interaction would happen.

FWIW, I was using the `nixos-unstable` with home-manager `master` and ran into this.

teto:
`set packdir` used to be set in the init.vim, it's now set as a neovim argument in the wrapper (to avoid creating an init.vim just for that, to ease the transition towards init.lua for those who wanna use a lua config only). The nixpkgs PR is https://github.com/NixOS/nixpkgs/pull/184364

ELD:
I must be doing something wrong because even using `nixos-unstable` to install packages (including Neovim), I can't get it to work with home-manager `master`. Any tips? I'm using a flake with each input and adding `unstable` via a nixpkgs overlay.

Sorry for all the questions!

teto:
is your config public ? please check that your nixpkgs version contains the nixpkgs commit I've mentioned in my previous post. You can see your nixpkgs version via `nix flake metadata`

Quentin-M:
Letting you know that this PR breaks neovim for darwin / nix-darwin completely.
Here is the `init.lua` you write: `vim.cmd [[source /var/empty/.config/nvim/init-home-manager.vim]]`
Path should be `/Users/<username>/.config/nvim/init.lua`, `/var/empty/.config` is not a thing.

teto:
the nix code does `"vim.cmd [[source ${config.xdg.configHome}/nvim/init-home-manager.vim]]"`. I am not sure how XDG transposes to darwin: is the /var/empty the correct translation of XDG or is that something in your setup ?
A possible alternative would be to rely on neovim to find the correct folder, i.e.: `"vim.cmd [[source stdpath('config')./init-home-manager.vim]]"`

> https://github.com/NixOS/nixpkgs/issues/193903 also broke darwin fwiw.

It's a different issue, comment in the linked PR (with more information). Maybe you just hit https://github.com/NixOS/nixpkgs/issues/193903#issuecomment-1264628610 ? (which is not darwin specific)

QuentinM-Hilbtec:
There's nothing about `xdg.configHome`, or `XDG_CONFIG_HOME` in my setup. `XDG_CONFIG_HOME` is simply not supported on darwin.

If you look around, for nix modules wanting to rely on it, they have conditions verifying the platform, or whether the env var is defined, and using `~/.config/` as a fallback. This applies to home-manager embedded modules too.

This behavior is in line with `nix.conf`'s [documentation](https://manpages.ubuntu.com/manpages/impish/man5/nix.conf.5.html):

```
Nix reads settings from two configuration files:
• The system-wide configuration file sysconfdir/nix/nix.conf (i.e.  /etc/nix/nix.conf on
  most systems), or $NIX_CONF_DIR/nix.conf if NIX_CONF_DIR is set.
• The user configuration file $XDG_CONFIG_HOME/nix/nix.conf, or [~/.config/nix/nix.conf](file:///~/.config/nix/nix.conf)
   if XDG_CONFIG_HOME is not set.
```

teto:
Even before this MR, the neovim module was writing init.vim via xdg.configFile, this didn't change ? where was it written and why would that be different now ? do you think you could provide a PR to fix the issue ?

QuentinM-Hilbtec:
Fair enough, apologies - looks like I was probably misled and someone else might've went wrong (?) while dissecting the commits.. I've just tried to converge based on your latest code (https://github.com/NixOS/nixpkgs/pull/194483) which is aimed to address the pname issue (separate from the empty path stuff), and the config file looks OK... Not even seeing the init-home-manager.lua call, not super sure how it gets injected and why it would get injected wrong, I haven't really spent time reading the code at that point.. Cheers!

PR #3170, opened 2022-08-22
programs.zsh: generate a more compact config

teto:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/zsh.nix
+++ b/modules/programs/zsh.nix
@@ -18,10 +18,6 @@ let
     mapAttrsToList (k: v: "alias ${k}=${lib.escapeShellArg v}") cfg.shellAliases
   );
 
-  globalAliasesStr = concatStringsSep "\n" (
-    mapAttrsToList (k: v: "alias -g ${k}=${lib.escapeShellArg v}") cfg.shellGlobalAliases
-  );
-
   dirHashesStr = concatStringsSep "\n" (
     mapAttrsToList (k: v: ''hash -d ${k}="${v}"'') cfg.dirHashes
   );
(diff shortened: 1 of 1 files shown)


teto:
I was surprised the zsh tests worked locally but apparently they just grep for stuff so having fewer newlines doesn't impact them @uvNikita

modules/programs/zsh.nix
```diff
@@ -547,10 +548,12 @@ in
 
         # Aliases
         ${aliasesStr}
-
-        # Global Aliases
-        ${globalAliasesStr}
```
sumnerevans, line 552: Can we remove the definition of `globalAliasesStr` from above in the file?

modules/programs/zsh.nix
```diff
@@ -460,44 +460,45 @@ in
...
         # as all $fpath entries will be traversed again.
         ${optionalString (cfg.enableCompletion && !cfg.oh-my-zsh.enable && !cfg.prezto.enable)
           cfg.completionInit
         }
-
-        ${optionalString cfg.enableAutosuggestions
-          "source ${pkgs.zsh-autosuggestions}/share/zsh-autosuggestions/zsh-autosuggestions.zsh"
+        ${optionalString cfg.enableAutosuggestions "
+          source ${pkgs.zsh-autosuggestions}/share/zsh-autosuggestions/zsh-autosuggestions.zsh
+		"
```
sumnerevans, line 501: This formatting looks a little suspect...

modules/programs/zsh.nix
teto: good catch

modules/programs/zsh.nix
teto: I've restored the version from master

teto: @sumnerevans is that betteR ?

otavio (approved):

sumnerevans (approved):

PR #3173, opened 2022-08-22
udiskie: avoid dependency on tray.target when tray is disabled

lilyinstarlight:
### Description

This PR removes the dependecy on `tray.target` when configuring udiskie to not use the system tray (motivated by me having `services.udiskie.tray = "never"` in my hm config but also no `tray.target` in my system, leading to errors even though `tray.target` is unnecessary)

I additionally added tests for the old and new behaviors for udiskie configuration (and made a drive-by typo fix). Quick question though: Do I need to add the `tests/modules/services/udiskie` directory to `.github/CODEOWNERS` and if so, should I add myself rather than the module maintainer or both of us or something else?

Edit: The test failure on Ubuntu in GHA seems unrelated to this PR and caused by neovim changes in nixpkgs. It looks like #3168 will fix that test, though

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/udiskie.nix
+++ b/modules/services/udiskie.nix
@@ -109,8 +109,9 @@ in {
     systemd.user.services.udiskie = {
       Unit = {
         Description = "udiskie mount daemon";
-        Requires = [ "tray.target" ];
-        After = [ "graphical-session-pre.target" "tray.target" ];
+        Requires = lib.optional (cfg.tray != "never") "tray.target";
+        After = [ "graphical-session-pre.target" ]
+          ++ lib.optional (cfg.tray != "never") "tray.target";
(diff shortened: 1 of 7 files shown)


rycee:
Thanks! Merged to master now.

lilyinstarlight:
> Thanks! Merged to master now.

Awesome, thank you! Should we add a line in `.github/CODEOWNERS` for `tests/modules/services/udiskie`, since this PR added tests but it's not currently in the CODEOWNERS file?

rycee:
Ah, sorry. I forgot about that. Yeah, I've updated the codeowners file now. Thanks for the reminder.

PR #3183, opened 2022-08-26
services.gpg-agent: allow usage of pinentry-mac

treed:
pinentry-mac is a separate package, and thus not included in the list
of flavors, and also it requires a little bit of extra handling path-wise.

### Description

I made this change so that I could use pinentry-mac rather than the default I was getting, which didn't work well when accessed from outside a terminal.

I didn't see any relevant tests to update.

I considered adding a note to the docstring that "mac" only works on Darwin, but that seemed perhaps too obvious. Let me know if I should go ahead and add it anyway, or feel free to edit and add that.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/gpg-agent.nix
+++ b/modules/services/gpg-agent.nix
@@ -179,7 +179,7 @@ in {
       };
 
       pinentryFlavor = mkOption {
-        type = types.nullOr (types.enum pkgs.pinentry.flavors);
+        type = types.nullOr (types.enum (pkgs.pinentry.flavors ++ [ "mac" ]));
         example = "gnome3";
         default = "gtk2";
         description = ''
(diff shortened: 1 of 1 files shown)


modules/services/gpg-agent.nix
```diff
@@ -179,7 +179,7 @@ in {
       };
 
       pinentryFlavor = mkOption {
-        type = types.nullOr (types.enum pkgs.pinentry.flavors);
+        type = types.nullOr (types.enum (pkgs.pinentry.flavors ++ [ "mac" ]));
         example = "gnome3";
         default = "gtk2";
```
montchr, line 184: What do you think about setting the default "flavor" to `mac` when `isDarwin` here? I've tried all the `pkgs.pinentry.flavors` on macOS and the experience is pretty terrible, so `pinentry_mac` seems like a sensible default on macOS IMO.

montchr (approved):
Nice! I just set this up in my personal configs and it worked perfectly.

https://github.com/montchr/dotfield/commit/d9ca7ac828b2b4b39ab336aed00ab093f89bb083

I merged your branch into a new branch on my fork incorporating the changes from my PR #2964 and your changes in this PR. No conflicts.

modules/services/gpg-agent.nix
treed: Seems reasonable to me!

treed:
Yeah, I'm happy to have the changes combined into one PR. I'll go ahead and close this one.

PR #3187, opened 2022-08-29
fluxbox: add module

AndersonTorres:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/window-managers/fluxbox.nix
+++ b/modules/services/window-managers/fluxbox.nix
@@ -1,0 +1,127 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.xsession.windowManager.fluxbox;
+
+in {
+  meta.maintainers = [ maintainers.AndersonTorres ];
+
+  options = {
+    xsession.windowManager.fluxbox = {
+      enable = mkEnableOption "Fluxbox window manager";
+
+      package = mkOption {
+        type = types.package;
+        default = pkgs.fluxbox;
+        defaultText = literalExpression "pkgs.fluxbox";
+        description = "Package to use for running Fluxbox WM.";
+      };
+
+      init = mkOption {
+        type = types.lines;
+        default = "";
+        description = ''
+          Init configuration for Fluxbox, written to
+          <filename>~/.fluxbox/init</filename>. Look at the
+          <citerefentry>
+          <refentrytitle>fluxbox</refentrytitle>
+          <manvolnum>1</manvolnum></citerefentry> manpage for details.
+        '';
+      };
+
+      apps = mkOption {
+        type = types.lines;
+        default = "";
+        description = ''
+          Apps configuration for Fluxbox, written to
+          <filename>~/.fluxbox/apps</filename>. Look at the
+          <citerefentry><refentrytitle>fluxbox</refentrytitle>
+          <manvolnum>1</manvolnum></citerefentry> manpage for details.
+        '';
+      };
+
+      keys = mkOption {
+        type = types.lines;
+        default = "";
+        description = ''
+          Keyboard shortcuts configuration for Fluxbox, written to
+          <filename>~/.fluxbox/keys</filename>. Look at the
+          <citerefentry><refentrytitle>fluxbox-keys</refentrytitle>
+          <manvolnum>1</manvolnum></citerefentry> manpage for details.
+        '';
+      };
+
+      menu = mkOption {
+        type = types.lines;
+        default = "";
+        description = ''
+          Menu configuration for Fluxbox, written to
+          <filename>~/.fluxbox/menu</filename>. Look at the
+          <citerefentry><refentrytitle>fluxbox-menu</refentrytitle>
+          <manvolnum>1</manvolnum></citerefentry> manpage for details.
+        '';
+      };
+
+      slitlist = mkOption {
+        type = types.lines;
+        default = "";
+        description = ''
+          Slitlist configuration for Fluxbox, written to
+          <filename>~/.fluxbox/slitlist</filename>. Look at the
+          <citerefentry><refentrytitle>fluxbox</refentrytitle>
+          <manvolnum>1</manvolnum></citerefentry> manpage for details.
+        '';
+      };
+
+      windowmenu = mkOption {
+        type = types.lines;
+        default = "";
+        description = ''
+          Window menu configuration for Fluxbox, written to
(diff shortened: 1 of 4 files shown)


AndersonTorres:
@rycee what the errors mean?

modules/services/window-managers/fluxbox.nix
```diff
@@ -0,0 +1,135 @@
...
+        defaultText = literalExpression "pkgs.fluxbox";
+        description = "Package to use for running Fluxbox WM.";
+      };
+
+      init = mkOption {
+        type = types.lines;
+        default = "";
+        description = ''
+          Init configuration for Fluxbox, written to
+          <filename>~/.fluxbox/init<filename>. Look at the fluxbox(1) manpage
```
rycee, line 25: Missing a `/` 🙂

```suggestion
          <filename>~/.fluxbox/init</filename>. Look at the fluxbox(1) manpage
```

The same mistake also appears in the options below.

rycee: Also the man page reference can be made with `citerefentry`. That is, in total something like

``` xml
<filename>~/.fluxbox/init</filename>. See
<citerefentry>
  <refentrytitle>fluxbox</refentrytitle>
  <manvolnum>1</manvolnum>
</citerefentry>
for details.
```

modules/services/window-managers/fluxbox.nix
```diff
@@ -0,0 +1,135 @@
...
+
+      extraStartupCommands = mkOption {
+        type = with types; listOf str;
+        default = "";
+        example = "";
+        description = ''
+          Extra startup commands executed before <command>fluxbox</command>
+          starts.
+
+          Caution! Commands that should keep running after the window manager
```
rycee, line 98: Unfortunately, to create a paragraph break you need `</para><para>` like so:

```suggestion
          starts.
          </para><para>
          Caution! Commands that should keep running after the window manager
```

modules/services/window-managers/fluxbox.nix
```diff
@@ -0,0 +1,135 @@
...
+      ".fluxbox/windowmenu" =
+        mkIf (cfg.windowMenu != "") { text = cfg.windowmenu; };
+    };
+
+    xsession.windowManager.command = let
+      commandLine = concatStringsSep " " (remove ""
+        ([ "${fluxbox}/bin/fluxbox" "-rc ~/.fluxbox/init" ]
+          ++ (escapeShellArgs cfg.extraCommandLineArgs)));
+    in ''
+      ${cfg.extraStartupCommands}
```
rycee, line 129: This seems unnecessary since [`xsession.initExtra`](https://nix-community.github.io/home-manager/options.html#opt-xsession.initExtra) accomplishes the same thing:

https://github.com/nix-community/home-manager/blob/df79df8be10bc54d79118ac6167a92b779344228/modules/xsession.nix#L204-L206

modules/services/window-managers/fluxbox.nix
```diff
@@ -0,0 +1,135 @@
...
+
+      exec ${commandLine} & fbpid=$!
+      wait $fbpid
```
rycee, line 132: Hmm, isn't this the same as

```suggestion
      ${commandLine}
```

?

rycee:
Thanks! Looks fine in general, I've added a few comments 🙂

modules/services/window-managers/fluxbox.nix
AndersonTorres: Yes it is. I was following the advice from Fluxbox manual, that suggests this in order to accomodate post-start commands. 

But I will simplify this,

modules/services/window-managers/fluxbox.nix
AndersonTorres: How exactly I pass command line arguments to the fluxbox binary?

modules/services/window-managers/fluxbox.nix
```diff
@@ -0,0 +1,116 @@
...
+        mkIf (cfg.windowMenu != "") { text = cfg.windowmenu; };
+    };
+
+    xsession.windowManager.command = let
+      commandLine = concatStringsSep " " (remove ""
+        ([ "${cfg.package}/bin/fluxbox" "-rc ~/.fluxbox/init" ]
+          ++ (escapeShellArgs cfg.extraCommandLineArgs)));
+    in ''
+      exec ${commandLine}
+    '';
```
rycee, line 114: Can simplify this. The `exec` needs to be removed since it would prevent the [cleanup actions](https://github.com/nix-community/home-manager/blob/41790ba656bafc023f48ccdbbe7816d30fd52d76/modules/xsession.nix#L208-L214) that happen after the window manager command finishes.

```suggestion
    xsession.windowManager.command = concatStringsSep " " (remove ""
        ([ "${cfg.package}/bin/fluxbox" "-rc ~/.fluxbox/init" ]
          ++ (escapeShellArgs cfg.extraCommandLineArgs)));
```

modules/services/window-managers/fluxbox.nix
```diff
@@ -0,0 +1,116 @@
...
+        default = pkgs.fluxbox;
+        defaultText = literalExpression "pkgs.fluxbox";
+        description = "Package to use for running Fluxbox WM.";
+      };
+
+      init = mkOption {
+        type = types.lines;
+        default = "";
+        description = ''
+          Init configuration for Fluxbox, written to <filename>~/.fluxbox/init</filename>. Look at the <citerefentry><refentrytitle>fluxbox</refentrytitle><manvolnum>1</manvolnum></citerefentry> manpage for details.
```
rycee, line 24: Best to avoid very long lines, can do something like

```suggestion
          Init configuration for Fluxbox, written to
          <filename>~/.fluxbox/init</filename>. See
          <citerefentry>
            <refentrytitle>fluxbox</refentrytitle>
            <manvolnum>1</manvolnum>
          </citerefentry>
          for details.
```

modules/services/window-managers/fluxbox.nix
rycee: The way you do it now with `escapeShellArgs cfg.extraCommandLineArgs` seems just fine.

modules/services/window-managers/fluxbox.nix
rycee: No, the exec will cause the rest of the script to be ignored since it performs a process substitution. See my other comment for a proposal on how to fix.

AndersonTorres:
Now it should be OK.

rycee:
Thanks! Merged to master now.

PR #3192, opened 2022-08-30
gpg-agent: invert grab and no-grab behavior

kisonecat:
Home-manager should explicitly output `grab` when `cfg.grabKeyboardAndMouse` is true.  Previously home-manager emitted `no-grab` when `cfg.grabKeyboardAndMouse` was false.

### Description

The GNU Privacy Guard 2.3 man page for `gpg-agent` describes the `--grab` and `--no-grab` options as follows:

> Tell the pinentry to grab the  keyboard  and mouse.   This  option  should  be used on X-Servers to avoid X-sniffing attacks. Any use of  the  option --grab overrides an used option --no-grab.  The default is --no-grab.

### Checklist

This may be "more" backwards compatible if `cfg.grabKeyboardAndMouse` defaulted to false.

Spurious warning: 

> In file /home/jim/downloads/me/home-manager/tests/lib/types/list-or-dag-merge.nixa list is being assigned to the option 'tested.dag'.

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/gpg-agent.nix
+++ b/modules/services/gpg-agent.nix
@@ -216,7 +216,7 @@ in {
     {
       home.file."${homedir}/gpg-agent.conf".text = concatStringsSep "\n"
         (optional (cfg.enableSshSupport) "enable-ssh-support"
-          ++ optional (!cfg.grabKeyboardAndMouse) "no-grab"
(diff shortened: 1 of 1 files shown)


rycee:
Thanks! Rebased to master in 7b512c94ffd714d18067257d08f7b5da6def947a.

Issue #3175, opened 2022-08-22
bug: neovim: cannot override nodejs version when coc is enabled


IvarWithoutBones:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

The neovim module unconditionally adds nodejs to `PATH` in the generated wrapper when  `coc.enable` is set:
https://github.com/nix-community/home-manager/blob/688e5c85b7537f308b82167c8eb4ecfb70a49861/modules/programs/neovim.nix#L362

It does not allow for overriding the version, which is problematic. I use both coc and github copilot, but the latter does not work with the latest version of nodejs, it requires version 14-17. This is of course an issue upstream, but I currently cannot work around it by supplying an older version of using `extraPackages` and settomg `withNodeJS = false`. The latest version gets added in the wrapper first because of coc, then the version i specificed gets added further down and ends up being ignored.

I propose one of the following fixes:
* Make the version of NodeJS configurable with an option like `nodejs = { enable = true; package = pkgs.nodejs_16_x; }`
* Make `withNodeJs` `null` by default, and make the `coc.enable` option set it to `true` if its not `false` already. This would allow me to set `withNodeJs = false`, and handle adding the package to the environent myself without breaking existing configurations.

### Maintainer CC

Since you last touched this file and helped out with an neovim related issue before: @sumnerevans

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 5.19.2, NixOS, 22.11 (Raccoon), 22.11.20220822.54060e8`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.10.3`
 - channels(root): `"nixos-21.11"`
 - channels(ivv): `"nixpkgs"`
 - nixpkgs: `/home/ivv/.nix-defexpr/channels/nixpkgs`
```

teto:
the only moment nodejs is passed is here pkgs/applications/editors/neovim/wrapper.nix but because of the makeOverridable in   `wrapNeovim = neovim-unwrapped: lib.makeOverridable (neovimUtils.legacyWrapper neovim-unwrapped);` I am not sure one can actually override nodejs. I dont have time to look into it so I hope you can find a solution. And yes the config is quite complex and could be simplified, I've started the process but the next step may require breaking changes :o

hanckmann:
Did you find a fix for this?
[...]

PR #3194, opened 2022-09-01
neovim: allow disabling NodeJS when CoC is enabled

IvarWithoutBones:
### Description

Previously enabling CoC would unconditionally add the latest version of NodeJS to the environment, without allowing the user to override it. That is a problem in the case a different plugin does not support the latest version of NodeJS, and only works on an 
older release.

I encountered this problem with github-copilot, which only supports NodeJS 14-17. It is entirely broken if used in combination with CoC right now.

This PR fixes the problem by allowing you to provide any version of NodeJS. You can set `withNodeJs = null` and `extraPackages = [ nodejs-16_x ]`.

Fixes https://github.com/nix-community/home-manager/issues/3175

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/neovim.nix
+++ b/modules/programs/neovim.nix
@@ -138,11 +138,14 @@ in {
       };
 
       withNodeJs = mkOption {
-        type = types.bool;
+        type = types.nullOr types.bool;
         default = false;
         description = ''
           Enable node provider. Set to <literal>true</literal> to
           use Node plugins.
+
+          This will be automatically enabled if <varname>coc.enable</varname> is set,
+          which can be circumvented by setting this option to <literal>null</literal>.
         '';
       };
 
@@ -341,9 +344,13 @@ in {
 
     neovimConfig = pkgs.neovimUtils.makeNeovimConfig {
       inherit (cfg) extraPython3Packages withPython3 withRuby viAlias vimAlias;
-      withNodeJs = cfg.withNodeJs || cfg.coc.enable;
       plugins = map suppressNotVimlConfig pluginsNormalized;
       customRC = cfg.extraConfig;
+      withNodeJs = if (cfg.withNodeJs == true || cfg.coc.enable
+        && cfg.withNodeJs != null) then
+        true
+      else
+        false;
     };
 
   in mkIf cfg.enable {


IvarWithoutBones:
All neovim related tests pass, except `neovim-plugin-config`. That is not a regression, it also fails for me on master.

modules/programs/neovim.nix
```diff
@@ -138,11 +138,14 @@ in {
...
       withNodeJs = mkOption {
-        type = types.bool;
+        type = types.nullOr types.bool;
         default = false;
         description = ''
           Enable node provider. Set to <literal>true</literal> to
           use Node plugins.
+
+          This will be automatically enabled if <varname>coc.enable</varname> is set,
+          which can be circumvented by setting this option to <literal>null</literal>.
```
teto, line 148: what's the point of having this to null and coc.enable true if it doesn't bring nodejs ? shouldn'it be an error instead ?

modules/programs/neovim.nix
```diff
@@ -341,9 +344,13 @@ in {
 
     neovimConfig = pkgs.neovimUtils.makeNeovimConfig {
       inherit (cfg) extraPython3Packages withPython3 withRuby viAlias vimAlias;
-      withNodeJs = cfg.withNodeJs || cfg.coc.enable;
       plugins = map suppressNotVimlConfig pluginsNormalized;
       customRC = cfg.extraConfig;
+      withNodeJs = if (cfg.withNodeJs == true || cfg.coc.enable
```
teto, line 349: or you know just use the boolean `cfg.withNodeJs == true || cfg.coc.enable  && cfg.withNodeJs != null)`, . Actually I would add parentheses for clarity `cfg.withNodeJs == true || (cfg.coc.enable        && cfg.withNodeJs != null))`

modules/programs/neovim.nix
IvarWithoutBones: With that you can supply an older version of nodejs yourself using `extraPackages` which coc will pick up.

teto (changes requested):
so i think it solves the wrong thing. coc needs nodejs to work, doesn't it ? the problem is that one version overrides the other ? 
You could instead disable withNodeJs and pass your node to extraMakeWrapperArgs .

What if you override nodejs in an overlay ? wouldn't that be simpler ?

A proper fix could be for this function 
 https://github.com/NixOS/nixpkgs/blob/bc9914ada7bdc6d2fbb2126d7c4358ab87289ee8/pkgs/applications/editors/neovim/utils.nix#L28 to accept an environment instead of a boolean.

IvarWithoutBones:
> so i think it solves the wrong thing. coc needs nodejs to work, doesn't it ? the problem is that one version overrides the other ?
You could instead disable withNodeJs and pass your node to extraMakeWrapperArgs .

Thing is, `withNodeJs` gets enabled unconditionally when coc is enabled. I cannot disable it myself.

> What if you override nodejs in an overlay ? wouldn't that be simpler ?

That'd cause a lot more rebuilds though, I would like to only use the older version of node in my nvim environment. 

> A proper fix could be for this function
https://github.com/NixOS/nixpkgs/blob/bc9914ada7bdc6d2fbb2126d7c4358ab87289ee8/pkgs/applications/editors/neovim/utils.nix#L28 to accept an environment instead of a boolean

That would indeed be cleaner, I'll get on that. Are you opposed the solution from this PR as a temporary workaround? That would need some time to get reviewed/merged in nixpkgs, and here in home-manager afterwards.

It's fairly annoying that this cannot be fixed from an end-users module without having to resort to overlaying `nodejs` in its entirety.

teto:
> Are you opposed the solution from this PR as a temporary workaround? 

Yes, the logic is very confusing. As a workaround you can override the call to neovim to pass your own nodejs.

IvarWithoutBones:
Okay, then I'll close this for now.

sumnerevans:
Trying to override via
```
programs.neovim = {
  package = pkgs.neovim-unwrapped.override {
    nodejs = pkgs.nodejs_17_x;
  };
};
```
is not working for me.

I think it's because `makeNeovimConfig` doesn't respect the `nodejs` passed in to the package. Has anyone found a solution here?

PR #3196, opened 2022-09-02
flake: correct nix-darwin flake description

jameschensmith:
### Description

Corrects copy/paste mistake in the nix-darwin flake description.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/docs/nix-flakes.adoc
+++ b/docs/nix-flakes.adoc
@@ -200,7 +200,7 @@ is similar to that of NixOS. The `flake.nix` would be:
 [source,nix]
 ----
 {
-  description = "NixOS configuration";
+  description = "Darwin configuration";
 
   inputs = {
     nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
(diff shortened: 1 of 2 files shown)


berbiche (approved):
Thanks!

docs/nix-flakes.adoc
```diff
@@ -188,7 +188,7 @@ is similar to that of NixOS. The `flake.nix` would be:
 [source,nix]
```
teto, line 188: shouldn't we include or link templates/nix-darwin/flake.nix here ? else both flakes will go out of sync

jameschensmith: You mean, for example, like `include::../templates/nix-darwin/flake.nix[]`? That's actually a great idea.

teto: yes

rycee:
Thanks! Merged to master now. Unfortunately the include doesn't work since the documentation build happens inside a sandbox that doesn't have access to the wider project code.

PR #3198, opened 2022-09-03
Fix mako

Rosuavio:
### Description

Enabling `mako` does not actually cause it to start for me or set anything up so that mako will handle notifications in any way, also its named `programs.mako` but in the `modules/services` folder and behaves more like a service anyway.

This changes `programs.mako` to `services.mako` and enable a mako systemd service when mako is enabled.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/mako.nix
+++ b/modules/services/mako.nix
@@ -4,13 +4,16 @@ with lib;
 
 let
 
-  cfg = config.programs.mako;
+  cfg = config.services.mako;
 
 in {
   meta.maintainers = [ maintainers.onny ];
 
+  imports =
+    [ (mkRenamedOptionModule [ "programs" "mako" ] [ "services" "mako" ]) ];
+
   options = {
-    programs.mako = {
+    services.mako = {
       enable = mkEnableOption ''
         Mako, lightweight notification daemon for Wayland
       '';
--- a/modules/services/mako.nix
+++ b/modules/services/mako.nix
@@ -18,6 +18,14 @@ in {
         Mako, lightweight notification daemon for Wayland
       '';
 
+      autoStart = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          Automaticly start mako with graphical-session.target.
+        '';
+      };
+
       package = mkOption {
         type = types.package;
         default = pkgs.mako;
@@ -367,6 +375,10 @@ in {
         ExecStart = getExe cfg.package;
         ExecReload = "${cfg.package}/bin/makoctl reload";
       };
-    };
+
+    } // (if autoStart then {
+      Install.WantedBy = [ "graphical-session.target" ];
+    } else
+      { });
   };
 }
--- a/modules/services/mako.nix
+++ b/modules/services/mako.nix
@@ -305,6 +305,15 @@ in {
     assertions =
       [ (hm.assertions.assertPlatform "services.mako" pkgs platforms.linux) ];
 
+    xdg.dataFile."dbus-1/services/fr.emersion.mako.service".text =
+      generators.toINI { } {
+        "D-BUS Service" = {
+          Name = "org.freedesktop.Notifications";
+          Exec = getExe cfg.package;
+          SystemdService = "mako.service";
+        };
+      };
+
     home.packages = [ cfg.package ];
 
     xdg.configFile."mako/config" = {
@@ -341,6 +350,23 @@ in {
 
         ${cfg.extraConfig}
       '';
+    };
+
+    systemd.user.services.mako = {
+      Unit = {
+        Description = "Lightweight Wayland notifcation deamon";
+        Documentation = "man:mako(1)";
+        PartOf = [ "graphical-session.target" ];
+        After = [ "graphical-session-pre.target" ];
+      };
+
+      Service = {
+        Type = "dbus";
+        BusName = "org.freedesktop.Notifications";
+        ExecCondition = "/bin/sh -c '[ -n \"$WAYLAND_DISPLAY\" ]'";
+        ExecStart = getExe cfg.package;
+        ExecReload = "${cfg.package}/bin/makoctl reload";
+      };
     };
   };
 }


Rosuavio:
@onny I know this is supper rough, but I want to make the experience for using mako smoother, I had to write a service config locally to use it. As my time writing a systemd service just to get something to handle notifications, it was not a good experience.

Rosuavio:
@rycee, also it feels wired having this in `programs` and I know the current changes are not backwards compatible in any way and it has no migration story but I did not know what else to do. I would like to know how to properly move `mako` from `programs` to `services`, if that is what we want.

sumnerevans:
We should at least add a `mkRemovedOptionModule` with an explanation of how to use the new module.

rycee:
Yeah, like @sumnerevans mentioned. Try adding a `mkRemovedOptionModule` for the old `programs.mako` path in the imports.

modules/services/mako.nix
```diff
@@ -4,17 +4,28 @@ with lib;
...
 let
 
-  cfg = config.programs.mako;
+  cfg = config.services.mako;
 
 in {
   meta.maintainers = [ maintainers.onny ];
 
+  imports =
+    [ (mkRenamedOptionModule [ "programs" "mako" ] [ "services" "mako" ]) ];
```
Rosuavio, line 13: Is this better than `mkRemovedOptionModule`?

modules/services/mako.nix
```diff
@@ -302,6 +313,15 @@ in {
...
       [ (hm.assertions.assertPlatform "services.mako" pkgs platforms.linux) ];
 
+    xdg.dataFile."dbus-1/services/fr.emersion.mako.service".text =
+      generators.toINI { } {
+        "D-BUS Service" = {
+          Name = "org.freedesktop.Notifications";
+          Exec = getExe cfg.package;
+          SystemdService = "mako.service";
+        };
+      };
```
Rosuavio, line 323: I saw [in `dunst`](https://github.com/nix-community/home-manager/blob/5408e27961599b1350b651f88715daf6e67244a7/modules/services/dunst.nix#L135) we just reuse the dbus service file provided from the repo. 
I wanted to use [the dbus service file provided by `mako`](https://github.com/emersion/mako/blob/a69041b0fde6ea7808c0e883c4b82b39da40641e/fr.emersion.mako.service.in), but unlike the one provided by `dunst` it lacks the `SystemdService` field. The [arch guys seem to patch this file](https://github.com/archlinux/svntogit-community/blob/2fae8c15a0459ac6562daee5b0d39ac6f038ba13/trunk/systemd-dbus-activation.patch) to add the `SystemdService` field. I am going to see what I can do upstream so we can avoid writing this our self.

modules/services/mako.nix
```diff
@@ -4,17 +4,28 @@ with lib;
...
         Mako, lightweight notification daemon for Wayland
       '';
 
+      autoStart = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          Automaticly start mako with graphical-session.target.
+        '';
+      };
```
Rosuavio, line 27: Might be the first option I wrote, any feedback is welcome.

modules/services/mako.nix
```diff
@@ -339,5 +359,26 @@ in {
...
+        BusName = "org.freedesktop.Notifications";
+        ExecCondition = "/bin/sh -c '[ -n \"$WAYLAND_DISPLAY\" ]'";
+        ExecStart = getExe cfg.package;
+        ExecReload = "${cfg.package}/bin/makoctl reload";
+      };
+
+    } // (if autoStart then {
+      Install.WantedBy = [ "graphical-session.target" ];
+    } else
+      { });
```
Rosuavio, line 382: Is this the way to do something like this?

Rosuavio (commented):
Okay I tried to clean it up as much as I could!

Rosuavio:
Should I add myself as a maintainer?

modules/services/mako.nix
Rosuavio: It seems like the `SystmedService` line was purposefully removed, https://github.com/emersion/mako/commit/dbd9c2bc98ef0cec7801a049f4bf8d61f7be36d0, hoping to find out why.

Also I opened a patch to nixpkgs https://github.com/NixOS/nixpkgs/pull/191448 copying arch.

Rosuavio:
So after further investigation. I have some question and issues that need resolving before I can continue. If can any get any assistance that would help a lot @rycee @sumnerevans @onny.

## Service or Program?
Currently mako is in a state between service and program in how it is presented in hm. (For devs) Its in the services folder, its named `programs.mako` and there is code calling it `services.mako` https://github.com/nix-community/home-manager/blob/5427f3d1f0ea4357cd4af0bffee7248d640c6ffc/modules/services/mako.nix#L303

### Which should it be? 
I can understand both sides. 

If we consider it a program, then this module can primarily provide access to configuring mako, and letting the user decide how to start mako separately ( there are many ways to start it, and the [mako docs](https://github.com/emersion/mako/blob/master/README.md#running) recommends both d-bus activation and starting is threw sway). A user could add config to sway to start mako, or they can write there own d-bus service file (writing the service file would have more pitfalls). Either way if we consider this module to be program configuration at its core, then it leaves more options for the user. We can then add options like `dbus-activated`, `sway-activated`, `systemd-integration` and `systemd-activated` that makes some common things easier without  forcing the user to use mako as a service.

Now if we consider it a service, then the module primarily sets up mako as an available service and also provides access to simple configuration. This would mean there would have to be a on-by-default method to make mako available to handle and display notifications (the go-to seems like d-bus activation) with options to change how its started or disable it from starting automatically or automatically handling notifications. And if we think most users are going to want to use it as a service then it provides simpler options and a clearer API.

I am leaning towards it being a service. I would like to hear other's thoughts.

## Enabling its service-like behavior
Currently the PR actually setups up mako in a way that is not officially supported by the project, though it works (D-bus activation + systemd service).  I think it should be optional to use the systemd integration but for some reason when I try setting mako up as just a d-bus service then I get this error `failed to create display` from mako. This happens because [`wl_display_connect`](https://manpages.debian.org/experimental/libwayland-doc/wl_display_connect.3.en.html), retunrs null, so I think mako is not seeing `WAYLAND_DISPLAY`or ` XDG_RUNTIME_DIR` or they are returning the wrong values.

When I run `systemctl --user show-environment`
I get `WAYLAND_DISPLAY=wayland-1` and `XDG_RUNTIME_DIR=/run/user/1000`, which seems right. I think that since my user's dbus service launches mako when my system gets a notification, that service needs access to these environment variables to for mako to see them.

This leaves me with two problems, I don't know how to see what environment variables mako is seeing when it is launched as a dbus service, and (if needed) I don't know how to provide those environment variable to my user's dbus service (and subsequently mako).

modules/services/mako.nix
rycee: Yeah, that does indeed seem nicer.

modules/services/mako.nix
ambroisie: You probably want `lib.optional` instead.

```suggestion
      Install.WantedBy = lib.optional autoStart "graphical-session.target";
    };
```

ambroisie (changes requested):

Rosuavio:
Boy, do I have egg on my face. I opened this PR because mako was not working for me. Today I found out that this was fixed in https://github.com/nix-community/home-manager/pull/3031, but I was using release-22.05.

My takeaways from this my research..,.
1. Nothing needs to be done to make mako work (we do not need to write our own dbus service file), the one packaged works fine if the current implementation of `wayland.windowManager.sway.systemdIntegration`
2. Maybe `programs.mako` should still be renamed to `services.mako` just for consistency.
3. Mako can be integrated with systemd, it is not officially supported  but the mako repo provides a systemd service file for those who want it.
  1. Systemd integration can provide some usefulness `systemctl --user status mako.service`
  2. Systemd integration can also enable an `autoStart` (start on `graphical-session.target`) vs the current "OnDemand" nature of D-Bus activation. (Don't know why anyone would want to use it)
  3. Arch from some reason enables systemd integration even though its not officially supported by mako.

So I am going to simply close this and open a few other PRs.
1. Rename `programs.mako` -> `services.mako`
2. Add some kind of systemd integration option to the `nixpkgs.mako` derivation (Why not Arch does it, might be useful)
3. Added systemd related options to `services.mako` (one for enabling the systemd service and another for starting mako on `graphical-session-target`).

PR #3201, opened 2022-09-04
git: add config helper for hooks

cmoog:
### Description

This adds a configuration option to `programs.git` that aids in the configuration of custom Git [hooks](https://git-scm.com/docs/githooks). 

###### Example Usage
```nix
{
  programs.git = {
    enable = true;
    hooks = {
      pre-commit = ./path-to-hook.sh;
    };
  };
}
```

### Checklist

This seems to be failing with an unrelated error?

### DIFF
--- a/modules/programs/git.nix
+++ b/modules/programs/git.nix
@@ -201,6 +201,21 @@ in {
         '';
       };
 
+      hooks = mkOption {
+        type = types.attrsOf types.path;
+        default = { };
+        example = literalExpression ''
+          {
+            pre-commit = ./pre-commit-script;
+          }
+        '';
+        description = ''
+          Configuration helper for Git hooks.
+          See <link xlink:href="https://git-scm.com/docs/githooks" />
+          for reference.
+        '';
+      };
+
       iniContent = mkOption {
         type = gitIniType;
         internal = true;
@@ -446,6 +461,15 @@ in {
         commit.gpgSign = cfg.signing.signByDefault;
         tag.gpgSign = cfg.signing.signByDefault;
         gpg.program = cfg.signing.gpgPath;
+      };
+    })
+
+    (mkIf (cfg.hooks != { }) {
+      programs.git.iniContent = {
+        core.hooksPath = let
+          entries =
+            mapAttrsToList (name: path: { inherit name path; }) cfg.hooks;
+        in toString (pkgs.linkFarm "git-hooks" entries);
       };
     })
 
--- a/tests/modules/programs/git/git-with-hooks.nix
+++ b/tests/modules/programs/git/git-with-hooks.nix
@@ -1,0 +1,26 @@
+{ pkgs, ... }:
+
+{
+  programs.git = {
+    enable = true;
+    hooks = { pre-commit = ./git-pre-commit-hook.sh; };
+  };
+
+  nmt.script = ''
+    function getGitConfig() {
+      ${pkgs.gitMinimal}/bin/git config \
+        --file $TESTED/home-files/.config/git/config \
+        --get $1
+    }
+
+    assertFileExists home-files/.config/git/config
+    hookPath=$(getGitConfig core.hooksPath)
+    assertLinkExists $hookPath/pre-commit
+
+    actual="$(readlink "$hookPath/pre-commit")"
+    expected="${./git-pre-commit-hook.sh}"
+    if [[ $actual != $expected ]]; then
+      fail "Symlink $hookPath/pre-commit should point to $expected via the Nix store, but it actually points to $actual."
+    fi
+  '';
+}
--- a/tests/modules/programs/git/default.nix
+++ b/tests/modules/programs/git/default.nix
@@ -5,4 +5,5 @@
   git-with-str-extra-config = ./git-with-str-extra-config.nix;
   git-with-signing-key-id = ./git-with-signing-key-id.nix;
   git-without-signing-key-id = ./git-without-signing-key-id.nix;
+  git-with-hooks = ./git-with-hooks.nix;
 }
--- a/tests/modules/programs/git/git-pre-commit-hook.sh
+++ b/tests/modules/programs/git/git-pre-commit-hook.sh
@@ -1,0 +1,2 @@
+#!/bin/sh
+echo "pre-commit hook..."


modules/programs/git.nix
```diff
@@ -201,6 +201,22 @@ in {
         '';
       };
 
+      hooks = mkOption {
+        type = types.listOf (types.submodule {
+          options = {
+            name = mkOption { type = types.str; };
+            text = mkOption { type = types.str; };
+          };
+        });
```
rycee, line 210: Would

```suggestion
        type = types.attrsOf types.lines;
```

work? Looking at the `symlinkJoin` call, the name needs to be unique anyway?

Also, presumably one would like to be able to merge the hook content, hence `types.lines`.

modules/programs/git.nix
```diff
@@ -201,6 +201,22 @@ in {
         '';
       };
 
+      hooks = mkOption {
```
rycee, line 204: You need to add an option description. Also make sure to add a test case that uses the new option.

rycee:
Thanks! Added a few comments.

modules/programs/git.nix
cmoog: Good catch. This is definitely nicer.

modules/programs/git.nix
```diff
@@ -447,6 +465,19 @@ in {
         gpg.program = cfg.signing.gpgPath;
       };
     })
+    (mkIf (cfg.hooks != { }) {
+      programs.git.iniContent = {
+        core.hooksPath = "${pkgs.symlinkJoin {
+          name = "git-hooks";
+          paths = attrValues (mapAttrs (name: text:
+            (pkgs.writeTextFile {
```
cmoog, line 473: As is, this requires the user to declare a shell interpreter. We _could_ automatically prepend that here instead.

modules/programs/git.nix
```diff
@@ -201,6 +201,24 @@ in {
...
+        default = { };
+        example = {
+          hooks = {
+            pre-commit = ''
+              #!/bin/sh
+              echo "formatting..."
+              # run formatting here
+            '';
+          };
+        };
```
sumnerevans, line 219:
```suggestion
        example = literalExpression ''
          {
            hooks = {
              pre-commit = ''
                #!/bin/sh
                echo "formatting..."
                # run formatting here
              '';
            };
          }
        '';
```

modules/programs/git.nix
sumnerevans: Why not instead use `types.path` and then allow users to choose how they want to generate the scripts? For example, they could do:

```
pre-commit = writeShellScript "cool-script" '...';
```

sumnerevans (changes requested):

modules/programs/git.nix
cmoog: Hmmmm... that would certainly be more flexible. It doesn't feel as clean to me though. @rycee thoughts?

modules/programs/git.nix
```diff
@@ -447,6 +465,19 @@ in {
         gpg.program = cfg.signing.gpgPath;
       };
     })
+    (mkIf (cfg.hooks != { }) {
+      programs.git.iniContent = {
+        core.hooksPath = "${pkgs.symlinkJoin {
```
olmokramer, line 470: I have tried to add hooks to my configuration in the past, and ran into a problem with symlinks. The thing is that when the hooks directory contains symlinks to the nix store, when git initializes a new repository it will just copy the symlinks to `.git/hooks` instead of the actual files. Then when I changed a hook and subsequently ran a nix garbage collect, the symlinks in various `.git/hooks` directories suddenly didn't point to a script anymore because they were gc'ed. So for this to work properly, the path that `core.hooksPath` points to should contain only regular files, not symlinks to the nix store.

olmokramer: After writing the above I looked up my config and it instead uses the `init.templatedir` setting, not `core.hooksPath`. So I don't think the above actually applies, as it seems `init.templatedir` is copied to `.git/hooks` on `git init` while `core.hooksPath` isn't and so doesn't suffer from the dead symlink problem.

I'm not sure, though, if this should use `core.hooksPath` instead of `init.templatedir`, as the former prevents one to use hooks on a per-repository basis.

olmokramer: PS this is the config I'm using, where `./hooks` is a directory containing all my hooks:

```nix
{ pkgs, ... }: {
  config.git.extraConfig.init.templatedir =
    "${pkgs.runCommandNoCCLocal "git-hooks" {} ''
      mkdir -p $out/hooks
      cp -r "${./hooks}"/* $out/hooks
    ''}";
```

cmoog: I agree that for certain setups, `templatedir` is the appropriate place to put hooks. That said, for a top-level `hooks` configuration, home-manager placing them in templateDir that would be non-obvious behavior in my view.

cmoog: >  core.hooksPath isn't and so doesn't suffer from the dead symlink problem

I can confirm that usage of `core.hooksPath` works as intended. The symlink problem is not an issue.

modules/programs/git.nix
rycee: I think having it expect the path to a script seems quite reasonable. Then it would also be more straight-forward to pull scripts from external sources. For example if there is a company specified pre-commit hook then you could define a package for it and do `hooks.pre-commit = pkgs.acme-corp.pre-commit-hook`.

I wonder, though if one would like to handle the situation where you have multiple hook scripts? I.e., should it be possible to do something like

``` nix
hooks.pre-commit = mkMerge [
  (writeShellScript "check-trailing-whitespace" "…")
  (writeShellScript "check-no-deploy-credentials" "…")
]
```

?

tests/modules/programs/git/git-with-hooks.nix
```diff
@@ -0,0 +1,26 @@
...
+          --file $TESTED/home-files/.config/git/config \
+          --get $1
+      }
+
+      assertFileExists home-files/.config/git/config
+      hookPath=$(getGitConfig core.hooksPath)
+      assertLinkExists $hookPath/pre-commit
+
+      actual="$(readlink "$hookPath/pre-commit")"
+      expected="${toString ./git-pre-commit-hook.sh}"
```
cmoog, line 20: Did I implement this properly?

rycee:
Thanks! I made a few slight cleanups and merged to master.

PR #3202, opened 2022-09-04
swayidle: allow wayland targets other than sway-session.target

erikrodrgz:
### Description

Allow another Wayland targets, as river-session.target or hyprland-session.target, to use swayidle.service which is  
 hard-coded to sway-session.target.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/swayidle.nix
+++ b/modules/services/swayidle.nix
@@ -96,6 +96,15 @@ in {
       default = [ ];
       description = "Extra arguments to pass to swayidle.";
     };
+
+    systemdTarget = mkOption {
+      type = types.str;
+      default = "sway-session.target";
+      description = ''
+        Systemd target to bind to.
+      '';
+    };
+
   };
 
   config = mkIf cfg.enable {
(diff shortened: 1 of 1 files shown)


sumnerevans (approved):

modules/services/swayidle.nix
```diff
@@ -114,7 +123,7 @@ in {
           "${cfg.package}/bin/swayidle -w ${concatStringsSep " " args}";
       };
 
-      Install = { WantedBy = [ "sway-session.target" ]; };
+      Install = { WantedBy = [ cfg.systemdTarget ]; };
```
rycee, line 126: Hmm, if swayidle works with any compositor, wouldn't `graphical-session.target` be the most appropriate value? Seems unnecessary to have to set it to whichever compositor you happen to use.

erikrodrgz: > Hmm, if swayidle works with any compositor, wouldn't `graphical-session.target` be the most appropriate value? Seems unnecessary to have to set it to whichever compositor you happen to use.

The way it was merged allows swayidle to be linked to another target than "graphical-session.target" being 
 more flexible.
The Default option could be "graphical-session.target", following the implemented in kanshi, wlsunset and waybar.

PR #3203, opened 2022-09-05
msmtp: allow sending email from aliased addresses

lheckemann:
### Description

This allows sending emails from all the configured aliases.

### Checklist

- [?] Change is backwards compatible. This will fail if an email address is configured both as an account name and as an account alias. I consider this unlikely enough that I think it's negligible, but feedback is welcome.

    ```
    {component}: {description}

    {long description}
    ```


rycee:
Thanks! I think the account name / alias conflict is unlikely and probably indicates some confused configuration anyway. Merged now.

PR #3204, opened 2022-09-05
editorconfig: add module

loicreynier:
### Description

Add a module to generate `~/.editorconfig` configuration file.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


loicreynier:
The module works fine in my configuration but the configuration file is not generated by the test. I must have miss something.

modules/misc/editorconfig.nix
```diff
@@ -0,0 +1,54 @@
...
+  cfg = config.editorconfig;
+
+  iniFormat = pkgs.formats.ini { };
+
+in {
+  meta.maintainers = with maintainers; [ loicreynier ];
+
+  options.editorconfig = {
+    enable = mkEnableOption "editorconfig home configuration file";
+    settings = mkOption {
```
rycee, line 16:
```suggestion

    settings = mkOption {
```

modules/misc/editorconfig.nix
```diff
@@ -0,0 +1,54 @@
...
+      type = iniFormat.type;
+      default = { };
+      description = ''
+        Configuration written to <filename>$HOME/.editorconfig</filename>.
+        <code>root = true</code> is automaticaly added to the file,
```
rycee, line 21:
```suggestion
        <code>root = true</code> is automatically added to the file,
```

modules/misc/editorconfig.nix
```diff
@@ -0,0 +1,54 @@
...
+            indent_style = "space";
+            indent_size = 4;
+          };
+        };
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home.file.".editorconfig" = mkIf (cfg.settings != { }) {
```
rycee, line 42: Can simplify to something like
```
  config = mkIf (cfg.enable && cfg.settings != { }) {
    home.file.".editorconfig".text = …
```

modules/modules.nix
```diff
@@ -23,6 +23,7 @@ let
     ./launchd/default.nix
     ./manual.nix
     ./misc/dconf.nix
+    ./misc/editorconfig.nix
     ./misc/debug.nix
```
sumnerevans, line 27:
```suggestion
    ./misc/debug.nix
    ./misc/editorconfig.nix
```

modules/misc/editorconfig.nix
```diff
@@ -0,0 +1,53 @@
...
+          };
+        };
+      '';
+    };
+  };
+
+  config = mkIf (cfg.enable && cfg.settings != { }) {
+    home.file.".editorconfig".text = let
+      renderedSettings = generators.toINIWithGlobalSection { } {
+        globalSection = { root = true; };
```
sumnerevans, line 45: We should provide a way for the user to add things to the global section.

modules/misc/editorconfig.nix
```diff
@@ -0,0 +1,53 @@
...
+  options.editorconfig = {
+    enable = mkEnableOption "EditorConfig home configuration file";
+
+    settings = mkOption {
+      type = iniFormat.type;
+      default = { };
+      description = ''
+        Configuration written to <filename>$HOME/.editorconfig</filename>.
+        <code>root = true</code> is automatically added to the file,
+        it must no be added in here.
```
sumnerevans, line 23:
```suggestion
        it must not be added here.
```

sumnerevans (changes requested):

modules/misc/editorconfig.nix
sumnerevans: Oh, I didn't realize that there was only one thing that could go in the global section. I guess you use `[*]` for other settings.

It may be worth providing a way to make it non-root, but I am not quite sure about the utility there, so this is probably fine

sumnerevans (changes requested):
Please rebase your commits down to a single one with the changes.

modules/misc/editorconfig.nix
loicreynier: Exactly, all settings other than `root = true` are in `[*]`.

I added a `rootConfig` option to provide a way to make it non-root. Not sure about the name though. Is there a naming convention for this kind of option? Maybe `enableRootConfig`?

rycee: I'd say to just leave it hard coded as `root = true` (i.e., not introduce an option for it). I doubt anybody would need to set it to `false` for the `~/.editorconfig` file. If somebody really wants it they can open an issue and we can worry about it then.

loicreynier:
I removed the `rootConfig` option and merged commits.

sumnerevans (approved):

PR #3206, opened 2022-09-07
fzf: add colors option

loicreynier:
### Description

Add an option to specify fzf colors with an attribute set rather than writing the entire option string.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/fzf.nix
+++ b/modules/programs/fzf.nix
@@ -6,6 +6,10 @@ let
 
   cfg = config.programs.fzf;
 
+  renderedColors = colors:
+    concatStringsSep ","
+    (mapAttrsToList (name: value: "${name}:${value}") colors);
+
 in {
   imports = [
     (mkRemovedOptionModule [ "programs" "fzf" "historyWidgetCommand" ]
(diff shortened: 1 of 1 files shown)


modules/programs/fzf.nix
```diff
@@ -141,7 +163,8 @@ in {
         FZF_CTRL_T_COMMAND = cfg.fileWidgetCommand;
         FZF_CTRL_T_OPTS = cfg.fileWidgetOptions;
         FZF_DEFAULT_COMMAND = cfg.defaultCommand;
-        FZF_DEFAULT_OPTS = cfg.defaultOptions;
+        FZF_DEFAULT_OPTS = cfg.defaultOptions
+          ++ [ "--color ${renderedColors cfg.colors}" ];
```
sumnerevans, line 167: This doesn't work when there are no colors specified.

sumnerevans (changes requested):

modules/programs/fzf.nix
loicreynier: Oh yes indeed, fixed.

loicreynier:
@sumnerevans is there anything left to fix that prevents this from being merged?

h7x4 (approved):
LGTM :+1:

sumnerevans (approved):

PR #3207, opened 2022-09-07
swiftbar: add module

IvarWithoutBones:
### Description

This adds a module for [SwiftBar](https://github.com/swiftbar/SwiftBar#script-metadata), an app that allows you to define scriptable menu bar items on MacOS. I recently PR'd it to nixpkgs, figured a home-manager module would make it a lot nicer to use. 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -157,6 +157,7 @@ let
     ./programs/skim.nix
     ./programs/sm64ex.nix
     ./programs/sqls.nix
+    ./programs/swiftbar.nix
     ./programs/ssh.nix
     ./programs/starship.nix
     ./programs/swaylock.nix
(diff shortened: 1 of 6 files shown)


modules/modules.nix
```diff
@@ -157,6 +157,7 @@ let
     ./programs/skim.nix
     ./programs/sm64ex.nix
     ./programs/sqls.nix
+    ./programs/swiftbar.nix
```
IvarWithoutBones, line 160: Should this be guarded by an `isDarwin` optional? The module itself asserts the platform is Darwin already, but I'm not sure that would cause the manpage entries to not get generated on Linux.

hodgesd:
This is good work… will it be merged?

PR #3209, opened 2022-09-08
mention `sudo nixos-rebuild switch` uses root user channels

DrRuhe:
To me it was not obvious that when using home-manager as a module in the nixos config it will use the channels of the root user when running `sudo nixos-rebuild switch`.

### Description

 `sudo nixos-rebuild switch` uses the channels configured for the root user. The instructions should contain a Note pointing this out so new users are aware of this.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/docs/installation.adoc
+++ b/docs/installation.adoc
@@ -124,21 +124,21 @@ deployed through NixOps.
 
 To make the NixOS module available for use you must `import` it into
 your system configuration. This is most conveniently done by adding a
-Home Manager channel. For example, if you are following Nixpkgs master
-or an unstable channel, you can run
+Home Manager channel to the root user. For example, if you are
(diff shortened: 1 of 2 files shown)


rycee:
Hmm, yeah. The documentation is a bit vague here. It uses the customary way of having `#` indicate commands that should be run as the root user and `$` for the regular user. I imagine if one is not used to this notation it can be unclear.

A more clear way may be to change the commands above this change from `#` to `$ sudo`. A user logged in as root would know that the sudo part is unnecessary and can omit that.

If we do that change then perhaps this added note becomes unnecessary? Especially since `nixos-rebuild switch` always should be run as root.

rycee:
Thanks! Merged to master now.

PR #3212, opened 2022-09-09
password-store: allow shell integrations

eaglesemanation:
### Description

Options to source password-store shell integrations

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/password-store.nix
+++ b/modules/programs/password-store.nix
@@ -53,6 +53,18 @@ in {
         available keys.
       '';
     };
+
+    enableBashIntegration = mkEnableOption "pass bash integration" // {
+      default = true;
+    };
+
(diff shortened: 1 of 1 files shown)


loicreynier:
I don't think this is necessary. When enabling completion for shell interpreters, completion script directories are sourced in the shell configuration file. We should not have to deal with standard completion scripts individually.

rycee:
Indeed, I think this should be unnecessary. I use Bash and have the custom `pass` completion available without any additional work.

Note, when using Home Manager on a distro different from NixOS you may have to include `targets.genericLinux = true` in your configuration to make everything available.

eaglesemanation:
@rycee `targets.genericLinux.enable = true;` fixed it for me, thanks! I'll close the PR

PR #3213, opened 2022-09-09
yt-dlp: add `settings` option

loicreynier:
### Description

Adds a structured `settings` option for the recently added `yt-dlp` module. Also adds a test for the generated configuration file.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/yt-dlp.nix
+++ b/modules/programs/yt-dlp.nix
@@ -6,6 +6,12 @@ let
 
   cfg = config.programs.yt-dlp;
 
+  renderSettings = mapAttrsToList (name: value:
+    if isBool value then
+      if value then "--${name}" else "--no-${name}"
+    else
+      "--${name} ${toString value}");
+
 in {
   meta.maintainers = [ maintainers.marsam ];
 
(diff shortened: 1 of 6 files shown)


modules/programs/yt-dlp.nix
```diff
@@ -6,6 +6,11 @@ let
 
   cfg = config.programs.yt-dlp;
 
+  renderedSettings = settings:
+    concatStringsSep "\n" (mapAttrsToList
+      (name: value: if isBool value then "--${name}" else "--${name} ${value}")
```
rycee, line 11: If you set

    embed-thumbnail = false;

then I guess the generated file should not contain `--embed-thumbnail` at all?

tests/modules/programs/yt-dlp/yt-dlp-simple-config.nix
```diff
@@ -0,0 +1,24 @@
+{ ... }:
+
+{
+  programs.yt-dlp = {
+    enable = true;
+    settings = {
+      embed-thumbnail = true;
+      embed-subs = true;
```
rycee, line 8: Try having one boolean option with a `false` value:

```suggestion
      embed-subs = false;
```

tests/modules/programs/yt-dlp/yt-dlp-simple-config.nix
```diff
@@ -0,0 +1,24 @@
...
+      sub-langs = "all";
+      downloader = "aria2c";
+      downloader-args = "aria2c:'-c -x8 -s8 -k1M'";
+    };
```
rycee, line 12: Try

```suggestion
      trim-filenames = 30;
    };
```

tests/modules/programs/yt-dlp/yt-dlp-simple-config.nix
```diff
@@ -0,0 +1,24 @@
...
+    extraConfig = ''
+      /home/user/.yt-dlp.conf
+    '';
+  };
+
```
rycee, line 17: No need to actually download the yt-dlp package in the test, so can stub it out:

```suggestion

    test.stubs.yt-dlp = { };

```

loicreynier:
Thanks for the fixes, should be good now.

modules/programs/yt-dlp.nix
```diff
@@ -6,6 +6,13 @@ let
 
   cfg = config.programs.yt-dlp;
 
+  renderedSettings = settings:
+    concatStringsSep "\n" (mapAttrsToList (name: value:
+      if (isBool value) then
+        if value then "--${name}" else ""
```
rycee, line 12: Looking closer at the yt-dlp man page, there are some Boolean options that have a `--no-` prefix. For example, `--embed-subs` and `--no-embed-subs`. So perhaps this should be

```suggestion
        if value then "--${name}" else "--no-${name}"
```

Some Boolean options don't seem to have a `--no`-variant, however, such as `--extract-audio`. But I guess for now we can just assume that people shouldn't set those to `false` 🙂 The alternative is to allow `null` and make `null` valued attributes result in nothing in the generated configuration.

loicreynier: I implemented your suggestion but did not understand the point of allowing `null` if it is for making `null` valued attributes result in nothing in the configuration.

rycee:
Thanks! Merged to master now with @ambroisie's suggestion applied. Also fixed the test case; it needs `nmt.script`, not `nmt.scripts`.

PR #3216, opened 2022-09-09
redshift/gammastep logging verbosity

arcnmx:
### Description

redshift spams the systemd journal since the module enables `-v` verbose mode by default, every few seconds reporting things like `Period: Transition (99.62% day)`. This adds an option that configures it to be quieter, but leaves the default as-is for compatibility with existing configs.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

CC @thiagokokada

### DIFF
--- a/modules/services/redshift-gammastep/lib/options.nix
+++ b/modules/services/redshift-gammastep/lib/options.nix
@@ -110,6 +110,8 @@ in {
       '';
     };
 
+    enableVerboseLogging = mkEnableOption "verbose service logging";
+
     tray = mkOption {
       type = types.bool;
       default = false;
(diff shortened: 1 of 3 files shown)


modules/services/redshift-gammastep/lib/options.nix
```diff
@@ -110,6 +110,14 @@ in {
       '';
     };
 
+    verbose = mkOption {
+      type = types.bool;
+      default = true;
+      description = ''
+        Enable verbose service logging.
+      '';
+    };
```
sumnerevans, line 119:
```suggestion
    verbose = mkEnableOption "verbose service logging" // {
      default = true;
    };
```

modules/services/redshift-gammastep/lib/options.nix
```diff
@@ -110,6 +110,8 @@ in {
       '';
     };
 
+    verbose = mkEnableOption "verbose service logging" // { default = true; };
```
rycee, line 113: I'm not sure why verbose logging is enabled by default, it's mostly interesting for debug purposes and could be off most of the time. So perhaps simply have it default to false?

I would prefer to rename the option to something like `enableVerboseLogging` or something similarly … verbose.

So in summary:

```suggestion
    enableVerboseLogging = mkEnableOption "verbose service logging";
```

arcnmx: Yeah I only left it on for backward-compatibility with existing configs, but I personally agree that it makes more sense to be off by default. Done!

rycee:

PR #3218, opened 2022-09-09
systemd: make unit files with `runCommand`

### LINKED ISSUE #3217: bug: mv: inter-device move failed: '/tmp/nix-build-emacs.service.drv-0/.attr-1' to "/nix/store/xyz-emacs.service'/emacs.service'"; unable to remove target: Permission denied

### Issue description

When building service files I get the above error.

I believe this is because on this machine /tmp is mounted on a different drive than /nix/store (not tmpfs) and home-manager uses `writeTextFile` to write systemd unit files: https://github.com/nix-community/home-manager/blob/9f7fe353b613d0e45d7a5cdbd1f13c96c15803dd/modules/systemd.nix#L41

... which, in turn, uses `mv`:
https://github.com/NixOS/nixpkgs/blob/2eb8607fd30f53c25dce32de0b17ce3fb6ab2007/pkgs/build-support/trivial-builders.nix#L128

nixpkgs uses `runCommand` to write units: https://github.com/NixOS/nixpkgs/blob/ea3f05e3e3fb5eaf6309deb443a1200f6389ae8a/nixos/lib/systemd-lib.nix#L20

### Maintainer CC

_No response_

### System information

```markdown
(sensitive information elided, no channels were used for this error to occur)

 - system: `"x86_64-linux"`
 - host os: `Linux 5.4.122, NixOS, 20.09pre-git (Nightingale)`
 - multi-user?: `no`
 - sandbox: `no`
 - version: `nix-env (Nix) 2.3.11`
```
```

### PR DESCRIPTION
**Pull request by jsoo1**
### Description 

Fixes #3217

Avoids errors of the following sort when /tmp is mounted on a different drive.

```
mv: inter-device move failed: '/tmp/nix-build-emacs.service.drv-0/.attr-1' to "/nix/store/xyz-emacs.service'/emacs.service'"; unable to remove target: Permission denied
```

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/systemd.nix
+++ b/modules/systemd.nix
@@ -38,11 +38,16 @@ let
 
       # Needed because systemd derives unit names from the ultimate
       # link target.
-      source = pkgs.writeTextFile {
-        name = pathSafeName;
+      source' = pkgs.runCommand pathSafeName {
+        preferLocalBuild = true;
+        allowSubstitutes = false;
         text = toSystemdIni serviceCfg;
-        destination = "/${filename}";
-      } + "/${filename}";
+      } ''
+        mkdir -p $out
+        echo -n "$text" > $out/${filename}
+      '';
+
+      source = "${source'}/${filename}";
 
       install = variant: target: {
         name = "systemd/user/${target}.${variant}/${filename}";


jsoo1:
`nix-shell --pure tests -A run.all` failed for me??

jsoo1:
I _believe_ this change is backwards-compatible but I am not 100% certain.

jsoo1:
Hm. I'm not 100% sure this actually fixes the issue. Simply bumping home-manager did the trick.

That said, I think this might be good to do?

rycee:
Hmm, I think if this is an issue with `writeTextFile` then it needs to be fixed in Nixpkgs. Although, I'm not sure what could cause it. Seems like `writeTextFile` attempted to write the target file twice (the second time being after making Nix store read-only)? Quite curious.

PR #3219, opened 2022-09-11
lib: add two new gvariant types

scotabroad:
Add GVariant variant and dictionary entry types

### Description

Added support for GVariant variant and dictionary entry types.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/docs/writing-modules.adoc
+++ b/docs/writing-modules.adoc
@@ -167,9 +167,11 @@ Builds a GVariant array containing the given list of elements, where each elemen
 - `hm.gvariant.type.int64`
 - `hm.gvariant.type.uint64`
 - `hm.gvariant.type.double`
+- `hm.gvariant.type.variant`
 - `hm.gvariant.type.arrayOf type`
 - `hm.gvariant.type.maybeOf type`
 - `hm.gvariant.type.tupleOf types`
+- `hm.gvariant.type.dictionaryEntryOf types`
 --
 +
 where `type` and `types` are themselves a type and list of types, respectively.
@@ -185,3 +187,9 @@ Builds a GVariant maybe value containing the given GVariant element.
 +
 `hm.gvariant.mkTuple elements`:::
 Builds a GVariant tuple containing the given list of elements, where each element is a GVariant value.
++
+`hm.gvariant.mkVariant element`:::
+Builds a GVariant variant which contains the value of a GVariant element.
++
+`hm.gvariant.mkDictionaryEntry elements`:::
+Builds a GVariant dictionary entry containing the given list of elements, where each element is a GVariant value.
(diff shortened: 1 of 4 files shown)


z0al:
How can this be used to represent dictionaries? currently, I do this to workaround the support:

```nix
{...}: 
{
  dconf.settings."org/gnome/shell/extensions/rounded-window-corners" = {
    global-rounded-corner-settings = {
      _type = "gvariant";
      type = "";
      value = "{'padding': <{'left': <uint32 1>, 'right': <uint32 1>, 'top': <uint32 1>, 'bottom': <uint32 1>}>, 'keep_rounded_corners': <{'maximized': <false>, 'fullscreen': <false>}>, 'border_radius': <uint32 8>, 'smoothing': <uint32 1>, 'enabled': <true>}";
      __toString = self:
        "@a{sv} ${self.value}";
    };
  };
}
```

scotabroad:
Dictionaries, as described here https://docs.gtk.org/glib/gvariant-text.html#dictionaries-and-dictionary-entries are actually an array of dictionary entries. A dictionary of type `a{sv}` could either be of form `[{"one", <1>}, {"two, <2>}]` or `{"one": <1>, "two": <2>}`. Technically, the second is syntactic sugar for the first one, which I found to easily take advantage of the `lib.hm.gvariant.mkArray` function and nix lists. A dictionary could thus be made as follows using this pull request: 
```nix
{...}:
let
  mkDictionaryEntry = lib.hm.gvariant.mkDictionaryEntry;
  mkVariantEntry = lib.hm.gvariant.mkVariant;
{
  dconf.settings."test/simple-dictionary" = {
    sample = [
      (mkDictionaryEntry ["one" (mkVariant 1)])
      (mkDictionaryEntry ["two" (mkVariant 2)])
      (mkDictionaryEntry ["three" (mkVariant 3)])
      (mkDictionaryEntry ["four" (mkVariant 4)])
    ];
  };
 }
 ```
 This generates a dictionary formatted as `[{"one", <1>}, {"two", <2>}. {"three", <3>}, {"four", <4>}]`, the array part being automatic as all nix lists get interpreted as gvariant arrays. Dconf-Editor will display dictionaries of this format in the other format when reading values as they are the same.

rycee:
Thanks! This seems quite nice! The `gvariant` type needs to be updated as well, though:

https://github.com/nix-community/home-manager/blob/95559181518533741c516826ae377a51814569c3/modules/lib/types.nix#L68-L104

I'm not entirely certain how best to define it but I imagine `mergeOneOption` should suffice for the variant values and something similar to the array definition for the dictionary.

scotabroad:
I have updated modules/lib/types.nix accordingly. My implementation for variants was inspired by the implementation for maybe. Dictionary entries are tuples. I have decided against implementing something for dictionaries, given dictionaries are just arrays of dictionary entries, and arrays are already implemented.

rycee:
Thanks! Looks good. Merged to master now.

PR #3224, opened 2022-09-14
micro: add bindings, ensurePlugins, and trueColor options

pbsds:
### Description

Add ability to configure keybinds and a micro-specific environ variable.
I also added the ability to install micro plugins during activation.

In the future i/someone might end up packaging the plugins properly, hence the odd, deprecateable, option name `ensurePlugins`.

### Checklist

- [x] Code tested through `nix-shell --pure tests -A run.micro`.

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/micro.nix
```diff
@@ -9,12 +9,19 @@ let
...
 
 in {
-  meta.maintainers = [ hm.maintainers.mforster ];
+  meta.maintainers = [ hm.maintainers.mforster maintainers.pbsds ];
 
   options = {
     programs.micro = {
       enable = mkEnableOption "micro, a terminal-based text editor";
 
+      trueColor = mkOption {
```
rycee, line 18: Perhaps

```suggestion
      enableTrueColor = mkOption {
```

since we often use the `enable` prefix for Boolean options.

modules/programs/micro.nix
```diff
@@ -31,13 +38,72 @@ in {
...
+      };
+
+      ensurePlugins = mkOption {
+        type = types.listOf types.str;
+        default = [ ];
+        example = literalExpression ''
+          [
+            "aspell"
+          ]
+        '';
```
rycee, line 70: I think this example is simple enough that `literalExpression` is overkill:

```suggestion
        example = [ "aspell" ];
```

modules/programs/micro.nix
```diff
@@ -31,13 +38,72 @@ in {
...
+        if ! test -d ${config.xdg.configHome}/micro/plug/${
+          lib.escapeShellArg pluginName
+        }; then
+          (set -x
+            $DRY_RUN_CMD ${pkgs.micro}/bin/micro -plugin install ${
+              lib.escapeShellArg pluginName
+            }
+          )
+        fi
+      '';
```
rycee, line 103: Hmm, this doesn't seem very Nixy to me. Is it not possible to package the plugins as Nix packages and, for example, symlink the plugins into place?

This also seems like it would require network access? Typically, the activation script should be possible to run without network availability.

pbsds: I agree. This was a temporary approach. Should i package all the plugins in nixpkgs, or here?

rycee:
Oh, and you can improve the commit message by avoiding to put everything the summary. Something like

```
micro: add a few new options

This adds the options

- `bindings`,
- `ensurePlugins`, and
- `enableTrueColor`.
```

modules/programs/micro.nix
rycee: We're very much against putting packages in the HM repository 🙂

The ideal is to have them in Nixpkgs. The second best is to have them in some other publicly accessible package repository, which is handy if the package set is automatically generated. See, e.g., the [Firefox add-on package set](https://gitlab.com/rycee/nur-expressions/-/blob/master/pkgs/firefox-addons/generated-firefox-addons.nix).

pbsds:
Draft until #192680 is merged. I'm plan on giving it some love during the holydays

phip1611:
Any update here? @rycee

rycee:
@phip1611 Not much I can do. I think this is waiting for the plugins to be available in Nixpkgs?

pbsds:
I have a new version of https://github.com/NixOS/nixpkgs/pull/192680 i've yet to push, but i have to test it properly and currently i'm crunching on a paper. In april i will finally have more time. :)

PR #3230, opened 2022-09-17
Fix pueue for empty settings

maisiliym:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/pueue.nix
+++ b/modules/services/pueue.nix
@@ -39,7 +39,7 @@ in {
 
     home.packages = [ cfg.package ];
 
-    xdg.configFile."pueue/pueue.yml".source = configFile;
+    xdg.configFile = mkIf (cfg.settings != { }) { "pueue/pueue.yml".source = configFile; };
 
     systemd.user = {
       services.pueued = {
@@ -49,7 +49,7 @@ in {
 
         Service = {
           Restart = "on-failure";
-          ExecStart = "${cfg.package}/bin/pueued -v -c ${configFile}";
+          ExecStart = "${cfg.package}/bin/pueued -v  -c ${configFile}";
         };
 
         Install.WantedBy = [ "default.target" ];


modules/services/pueue.nix
```diff
@@ -39,7 +41,7 @@ in {
 
     home.packages = [ cfg.package ];
 
-    xdg.configFile."pueue/pueue.yml".source = configFile;
+    xdg.configFile = mkIf explicitSettings { "pueue/pueue.yml".source = configFile; };
```
rycee, line 44: I would prefer

```suggestion
    xdg.configFile = mkIf (cfg.settings != { }) { "pueue/pueue.yml".source = configFile; };
```

since we use that everywhere, the `explicitSettings` indirection makes it seem like something special is going on.

modules/services/pueue.nix
```diff
@@ -49,7 +51,7 @@ in {
 
         Service = {
           Restart = "on-failure";
-          ExecStart = "${cfg.package}/bin/pueued -v -c ${configFile}";
+          ExecStart = "${cfg.package}/bin/pueued -v";
```
rycee, line 54: The configuration file should still be defined here, otherwise we won't know when the service needs to be restarted on configuration changes.

rycee:
CC @AndersonTorres

rycee:
@maisiliym The CI error is related to formatting, you can run the `format` script in the project root to do that.

modules/services/pueue.nix
```diff
@@ -7,8 +7,10 @@ let
   cfg = config.services.pueue;
   yamlFormat = pkgs.formats.yaml { };
   configFile = yamlFormat.generate "pueue.yaml" cfg.settings;
+  explicitSettings = cfg.settings != { };
```
AndersonTorres, line 10: Needless alias, it just pollutes the patch.

```suggestion
```

AndersonTorres (changes requested):

LiGoldragon:
Is it more energy to make the changes rather than document them? I guess github is easier to use than git.

LiGoldragon:
Maybe it's my workflow, but it was easier for me to make the changes than use github to document then changes I would make. Just saying.

AndersonTorres:
What exactly are you saying? For me Github is merely the web interface for a git repo.

rycee:
Thanks! Rebased to master.

Issue #1907, opened 2021-04-07
Neovim option to avoid writing init.vim


DanCardin:
### Issue description

https://github.com/nix-community/home-manager/commit/8e0c1c55fbb7f16f9fd313275ddf63c97b34394c, specifically
```
    xdg.configFile."nvim/init.vim".text = neovimConfig.neovimRcContent;
```

appears to have introduced a "regression" in my usage, given that I define much of my dotfile/XDG_CONFIG standalone in a way that doesn't depend on nix, so I therefore intentionally have a `.config/nvim/init.vim`. When running `home-manager switch` i now see the following:

```
Existing file '<elided>/.config/nvim/init.vim' is in the way of '/<elided>/.config/nvim/init.vim'
Please move the above files and try again or use -b <ext> to move automatically.
```

Ideally, there would be an option available to disable this behavior, but if it's an antigoal to provide this kind of customization, is there a straightforward way for me to override or patch remove that particular line, so that it doesn't attempt to write the `init.vim` file?

#### Maintainer CC

@teto 

#### Technical details

Can do this, but it feels unimportant given the above context.

teto:
I am in a similar scenario where my config is a complete mess of nix/vimscript/lua. 
One question first, do you have a config generated in nix ? if you've got no config, home-manager could skip writing the init.vim (which is not the case I think because we always have some init.vim like `set compatible` because it's generated by the vimrcContent function).

I have a fork of home-manager where I had changed home-manager to generate `init.generated.vim` instead of `init,vim`. And I would source it from my manual init.vim. I now do the reverse (to more easily test the home-manager master branch,

With the upcoming neovim, one can write a lua config too. I have one that I source too via `luafile`. I guess we will have to support luaConfig for plugins etc and generate a lua file as well.

DanCardin:
My home-manager config for neovim looks like:
```
  programs.neovim = {
    enable = true;
    package = pkgs.neovim-unwrapped;
    vimAlias = true;
    withNodeJs = false;
    withPython3 = true;
    withRuby = false;

    extraPython3Packages = (ps: with ps; [
      jedi
      pynvim
      pkgs.python37Packages.python-language-server
      pkgs.python37Packages.pyls-mypy
      pkgs.python37Packages.pyls-isort
      pkgs.python37Packages.pyls-black
    ]);
  };
```

Which I only mention to show that i am only using it for making available dependencies for the lsp tooling.

[...]

PR #3233, opened 2022-09-17
programs.neovim: default to init.lua

teto:
We change the current logic: instead of writing an init.vim which loads lua/init-home-manager.lua, we write an init.lua that sources init.vim

This commit also avoids writing any of these files if the plugins have no config.

Should fix https://github.com/nix-community/home-manager/issues/1907

Needs some tests

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/neovim.nix
+++ b/modules/programs/neovim.nix
@@ -368,13 +368,11 @@ in {
           "nvim/init.vim" = mkIf (neovimConfig.neovimRcContent != "") {
             text = neovimConfig.neovimRcContent;
           };
-          "nvim/init.lua" =
-            mkIf (hasLuaConfig || neovimConfig.neovimRcContent != "") {
-              text = lib.optionalString hasLuaConfig
-                "vim.cmd.source ${config.xdg.configFile."nvim/init.vim"}"
-                + lib.optionalString hasLuaConfig
-                config.programs.neovim.generatedConfigs.lua;
-            };
+          "nvim/init.lua".text =
+            lib.optionalString (neovimConfig.neovimRcContent != "")
+            "vim.cmd.source ${config.xdg.configHome}/nvim/init.vim"
+            + lib.optionalString hasLuaConfig
+            config.programs.neovim.generatedConfigs.lua;
           "nvim/coc-settings.json" = mkIf cfg.coc.enable {
             source = jsonFormat.generate "coc-settings.json" cfg.coc.settings;
           };
--- a/modules/programs/neovim.nix
+++ b/modules/programs/neovim.nix
@@ -368,11 +368,14 @@ in {
           "nvim/init.vim" = mkIf (neovimConfig.neovimRcContent != "") {
             text = neovimConfig.neovimRcContent;
           };
-          "nvim/init.lua".text =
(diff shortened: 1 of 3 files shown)


modules/programs/neovim.nix
```diff
@@ -357,17 +357,17 @@ in {
...
-            (hasAttr "lua" config.programs.neovim.generatedConfigs)
-            "lua require('init-home-manager')";
+          text = neovimConfig.neovimRcContent;
         };
-        "nvim/lua/init-home-manager.lua" =
-          mkIf (hasAttr "lua" config.programs.neovim.generatedConfigs) {
-            text = config.programs.neovim.generatedConfigs.lua;
+        "nvim/init.lua" = mkIf (hasLuaConfig || neovimConfig.neovimRcContent != "") {
+          text = config.programs.neovim.generatedConfigs.lua 
+            + lib.optionalString hasLuaConfig "vim.cmd([[source ~/.config/nvim/init.vim]])";
```
teto, line 370: of course this needs to be adapted.

AndersonTorres (changes requested):
`./format` the code

AndersonTorres (approved):

ambroisie (approved):

johanneshorner:
Using these commits with only "programs.neovim.enable = true" does create an empty "init.lua" file for me. Thats not correct right?
The file should only be created if "extraConfig" or some other config options are set. Except maybe I'm missing some default option I'd have to set?

teto:
@johanneshorner thanks that's exactly the kind of feedback I was looking for. I've adjusted the test to check that. Should be good now

johanneshorner:
@teto Just retested with the new commit. Works exactly as described in your pull request now. I can just symlink my full neovim config to "$HOME/.config/nvim" and there are no more conflicts. Thanks for the all the hard work :-).

rycee:
Nicely done! :+1:

Smaug123:
This PR broke my setup on Darwin, https://github.com/Smaug123/nix-dotfiles/blob/server-flake/home.nix#L130 (commit 418bab12b271a0a909778e1b27f0bd0e9e16edb8), and I'm still trying to work out how to fix it. When I start `neovim` now:

```
Error detected while processing /Users/patrick/.config/nvim/init.lua:
E5113: Error while calling lua chunk: /Users/patrick/.config/nvim/init.lua:1: Vim(source):E484: Can't open file /var/empty/.config/nvim/init-home-manager.vim
stack traceback:
        [C]: in function 'cmd'
        /Users/patrick/.config/nvim/init.lua:1: in main chunk
```

If I add the configuration `xdg.configHome = "/Users/patrick"` to make the path correct, neovim does start without errors, but my Git config is no longer picked up. (I'd never heard of XDG before this bug, so I am having to learn a lot of stuff quite quickly now.)

ffung:
@Smaug123 Not sure if you have resolved this, but this worked for me https://github.com/LnL7/nix-darwin/issues/423#issuecomment-1238515147

Smaug123:
> @Smaug123 Not sure if you have resolved this, but this worked for me https://github.com/LnL7/nix-darwin/issues/423#issuecomment-1238515147

No apparent effect, I'm afraid.

PR #3235, opened 2022-09-18
Files: Allow `target="."`, and add symlinkJoin to the documentation

niacdoial:
### Description

- Allow the use the home directory as a target for *recursively-linked directories*
- Add `symlinkJoin` to the documentation for `home.file`

### Checklist

 -> wont fix just yet, changes are too big to be in a simple pull request (both files.nix and file-type.nix are on the ignore list of ./format)

 -> arguably not needed for this change
 -> would need more discussion anyway, as the previous behavior tripped a "this should have been caught earlier" check, which cannot be used in tests

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/files.nix
+++ b/modules/files.nix
@@ -343,8 +343,13 @@ in
           # We therefore simply log the conflict and otherwise ignore it, mainly
           # to make the `files-target-config` test work as expected.
           if [[ -e "$realOut/$relTarget" ]]; then
-            echo "File conflict for file '$relTarget'" >&2
-            return
+            if [[ -d "$realOut/$relTarget" && $(realpath "$realOut/$relTarget") == $realOut && $recursive ]]; then
+                # exception: allow a "collision" for a recureively linked home *directory*
+                :
+            else
+                echo "File conflict for file '$relTarget'" >&2
+                return
+            fi
           fi
 
           # Figure out the real absolute path to the target.
--- a/modules/lib/file-type.nix
+++ b/modules/lib/file-type.nix
@@ -25,6 +25,13 @@ in
           defaultText = literalExpression "<name>";
           description = ''
             Path to target file relative to ${basePathDesc}.
+            </para><para>
+            Please note that only a single 
+            <xref linkend="opt-home.file"/> directive
+            may write to a given target.
+            In order to link several directories to the same place, you may use
+            <literal>symlinkJoin</literal> as such:</para><para>
+            <literal>source = pkgs.symlinkJoin{name="myjoin"; dirs=[ home-package /some/path ]; postBuild="echo custom shell script here";};</literal>
           '';
         };
 


niacdoial:
There's just a hickup which might pause this pull request:

I have a weird unrelated issue which might require this to be paused.
Using the builtin `pkgs.symlinkJoin` within my `home.nix` file fails (the symlinkJoin-based derivation appears (and its "symlink-joined-dir.drv" file too), but never gets built.)
However, copying the code for `symlinkJoin` from nixpkgs's `trivial-builders.nix` into `home.nix` and using that copy of the function fixes this?

I'll paste the exact code (and version info, etc) if you want, in a later comment.

This pull request might be the "better solution" when compared to my earlier #3141.

ajgrf:
I was about to submit a PR with nearly the same fix, except that instead of making a special case out of `target=".";` I just tested that both the source and target were directories (so `lndir` will work):

```bash
          # If the target already exists then we have a collision. Note, this
          # should not happen due to the assertion found in the 'files' module.
          # We therefore simply log the conflict and otherwise ignore it, mainly
          # to make the `files-target-config` test work as expected.
          if [[ -e "$realOut/$relTarget" && ]]; then
            if [[ $recursive && -d "$realOut/$relTarget" && -d "$source" ]]; then
              : skip
            else
              echo "File conflict for file '$relTarget'" >&2
              return
            fi
          fi
```

I think this is probably better, because I would also expect recursively-linked directories to work elsewhere, too. For example, this works with my solution but not with yours:

```nix
xdg.dataFile.hello.text = "world\n";
home.file.cowsay = {
  target = ".local";
  source = "${pkgs.cowsay}";
  recursive = true;
};
```

modules/files.nix
```diff
@@ -343,8 +343,13 @@ in
           # We therefore simply log the conflict and otherwise ignore it, mainly
           # to make the `files-target-config` test work as expected.
           if [[ -e "$realOut/$relTarget" ]]; then
-            echo "File conflict for file '$relTarget'" >&2
-            return
+            if [[ -d "$realOut/$relTarget" && $(realpath "$realOut/$relTarget") == $realOut && $recursive ]]; then
+                # exception: allow a "collision" for a recureively linked home *directory*
```
ajgrf, line 347:
```suggestion
            if [[ $recursive && -d "$realOut/$relTarget" && -d "$source" ]]; then
              # exception: allow a "collision" for a recursively linked *directory*
```

I don't think there's any reason to make a special case out of `target = "."`. Recursively-linked directories should work everywhere. Just test whether source and target are both directories so we know `lndir` will work.

modules/files.nix
```diff
@@ -343,8 +343,13 @@ in
...
+                :
+            else
+                echo "File conflict for file '$relTarget'" >&2
+                return
```
ajgrf, line 351:
```suggestion
              :
            else
              echo "File conflict for file '$relTarget'" >&2
              return
```

Indent with 2 spaces like the rest of the file.

ajgrf (changes requested):

niacdoial:
Sorry, but there are… a few issues with the changes you are suggesting (save for the typo. well spotted.)
Though, I created a previous pull request that does what you are interested in, at #3141.

First, the script you modified (the bash script that checks for collisions) is filling the derivation (`/nix/store/HASH-home-manager-files/`, go look at where `$realOut` is defined there), so it doesn't even see collisions with directories that only exist in the target `/home/USER/`. what means is that your quote here:

> For example, this works with my solution but not with yours:

isn't correct.

Second, this wouldn't actually allow multiple recursively-linked directories to have the same destination, because there is a check in the nix code that runs before that bash script. That nix code also needed changes in #3141.

Lastly, but most importantly, the check you removed introduces a huge footgun:
consider the following (…with some overdramatic names for emphasis):

```nix
home.file.myfiles = {
  target = ".config";
  source = /data/user/my-really-important-and-really-delicate-config-files-that-are-worth-gold;
  recursive = false;
};
home.file.expconfig = {
  target = ".config/sway";
  source = "${pkgs.some-half-assed-experimental-config-generator-that-is-currently-incomplete}";
  recursive = true;
};
```

modules/files.nix
niacdoial: I think the code style for nixos uses 4-space indents for bash code and 2-space indents for nix code.
…though now that I think about it, what is in the current code of home-manager is not up to style?
hm…
well, at least the file would be uniform.

niacdoial:
…so.
Basically, looking at this again, I just realize how condescending I sounded to you, even after the edit.

Anyway, if someone looks at this and can make decisions about the project, I'm not sure whether this pull or #3141 is the most appropriate for home-manager. so I guess until there's a decision on that, this will stay a stale pull.

niacdoial:
if someone stumbles upon this issue and it's not dealt with, here's a workaround that works on nixos unstable (and only unstable as of 2023-02-18):
```nix
  home.file = builtins.listToAttrs (
    let link_me =
     (readDir pkgs.inhome-data.outDir);
    in
    builtins.map ( filename: {
      name = filename;
      value = {
        source = pkgs.inhome-data + "/" + filename;
        target = ( "./" +  filename);
        recursive = (link_me."${filename}" == "directory");
      };
    })
    (builtins.attrNames link_me)
  );
```

On NixOs stable, `(readDir pkgs.inhome-data.outDir);` will not work because `builtins.readDir` is not defined.
instead, you need to generate that manually, with the value looking like (for example) `{".config" = "directory"; "bin" = "directory"; ".profile" = "symlink"; ".bashrc" = "symlink";}`

ajgrf:
@niacdoial No problem, don't worry about it.

Is there anything blocking this from being merged? I rely pretty heavily on this feature in my own config, and it would be nice to be able to remove the giant workaround from my dotfiles that I need without this upstream.

niacdoial:
For both this pull request and the one I refer to, what blocks the merge is that it has not been reviewed yet?
I guess the main question is which of the two should be merged (they are kinda mutually exclusive, since they are different workarounds for the same problem). I don't think I can do anything to speed things up on my end.

PR #3238, opened 2022-09-19
blesh: add module

aiotter:
### Description

This PR implements @akinomyoga's [ble.sh](https://github.com/akinomyoga/ble.sh) module.
cf. https://github.com/akinomyoga/ble.sh/issues/216

~~In order to load ble.sh earlier than other programs like starship, I have to modify `home.file` module so that you can edit files from other modules.~~
~~The main implementation of this change resides in `modules/lib/file-overlay.nix`.~~
~~With `import modules/lib/file-overlay.nix { inherit pkgs config; }` you can get a simple function which takes original `home.file` config and returns modified one.~~

~~I'm not sure whether this is the best practice. Please let me know if you have better ideas.~~

I'll remove them and use `mkBefore` and `mkAfter` instead. While rewriting this PR is kept draft.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -81,6 +81,9 @@ Makefile                                              @thiagokokada
 
 /modules/programs/beets.nix                           @rycee
 
+/modules/programs/blesh.nix                           @aiotter
+/tests/modules/programs/blesh                         @aiotter
+
 /modules/programs/bottom.nix                          @polykernel
 /tests/modules/programs/bottom                        @polykernel
 
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -698,6 +698,13 @@ in
       }
 
       {
+        time = "2022-09-19T11:03:53+00:00";
+        message = ''
+          A new module is available: 'programs.blesh'.
+        '';
+      }
+
+      {
         time = "2022-09-21T22:42:42+00:00";
         condition = hostPlatform.isLinux;
         message = ''
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -57,6 +57,7 @@ let
     ./programs/bashmount.nix
     ./programs/bat.nix
     ./programs/beets.nix
+    ./programs/blesh.nix
     ./programs/bottom.nix
     ./programs/broot.nix
     ./programs/browserpass.nix
--- a/modules/programs/blesh.nix
+++ b/modules/programs/blesh.nix
@@ -1,0 +1,96 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let cfg = config.programs.blesh;
+in {
+  meta.maintainers = [ maintainers.aiotter ];
+
+  options.programs.blesh = {
+    enable = mkEnableOption "ble.sh";
+
+    package = mkPackageOption pkgs "ble.sh" { default = [ "blesh" ]; };
+
+    enableBashIntegration = mkOption {
+      type = types.bool;
+      default = true;
+      description = ''
+        Load ble.sh at the beginning of `~/.bashrc` and attach it at the end.
+        No effect if <code>programs.bash.enable</code> is false.
+      '';
+    };
+
+    options = mkOption {
+      type = types.attrsOf types.str;
+      default = { };
+      description = ''
+        Change setting variables with <code>bleopt</code> function.
+      '';
+      example = { complete_auto_delay = "300"; };
+    };
+
+    faces = mkOption {
+      type = types.attrsOf types.str;
+      default = { };
+      description = ''
+        Change graphic settings with <code>ble-face</code> function.
+      '';
+      example = { region = "bg=60,fg=white"; };
+    };
+
+    imports = mkOption {
+      type = types.listOf types.str;
+      default = [ ];
+      description = ''
+        Search listed script files and sources them with <code>ble-import</code> if they are not yet sourced. Scripts are searched from <code>import_path</code> setting variables.
+        Use <xref linkend="opt-programs.blesh.blercExtra"/> for delayed imports.
+      '';
+      example = [ "contrib/bash-preexec" ];
+    };
+
+    blercExtra = mkOption {
+      type = types.str;
+      default = "";
+      description = ''
+        Extra configlations to be added at the end of <filename>~/.blerc</filename>.
+      '';
+      example = ''
+        function my/complete-load-hook {
+          bleopt complete_auto_delay=300
+        }
+
+        blehook/eval-after-load complete my/complete-load-hook
+      '';
+    };
+  };
+
+  config = let
+    stringifyAttrs = cmd: options:
+      concatStringsSep "\n"
+      (mapAttrsToList (k: v: "${cmd} ${k}=${escapeShellArg v}") options);
+    stringifyLists = cmd: options:
+      concatStringsSep "\n" (map (v: "${cmd} ${escapeShellArg v}") options);
+    optionsStr = stringifyAttrs "bleopt" cfg.options;
+    facesStr = stringifyAttrs "ble-face" cfg.faces;
+    importsStr = stringifyLists "ble-import" cfg.imports;
+  in mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    home.file.".blerc".source = pkgs.writeTextFile {
+      name = "blerc";
+      text =
+        concatStringsSep "\n" [ optionsStr facesStr importsStr cfg.blercExtra ];
+      checkPhase = ''
+        ${pkgs.stdenv.shellDryRun} "$target"
+      '';
+    };
+
+    programs.bash.bashrcExtra = mkIf cfg.enableBashIntegration (mkBefore ''
+      [[ $- == *i* ]] && source '${config.programs.blesh.package}/share/blesh/ble.sh' --noattach
+    '');
+
+    programs.bash.initExtra = mkIf cfg.enableBashIntegration (mkAfter ''
+      [[ ''${BLE_VERSION-} ]] && ble-attach
+    '');
+  };
+}
--- a/modules/programs/blesh.nix
+++ b/modules/programs/blesh.nix
@@ -20,6 +20,49 @@ in {
       '';
     };
 
+    rcfile = let
+      stringifyAttrs = cmd: options:
+        concatStringsSep "\n"
+        (mapAttrsToList (k: v: "${cmd} ${k}=${escapeShellArg v}") options);
+      stringifyLists = cmd: options:
+        concatStringsSep "\n" (map (v: "${cmd} ${escapeShellArg v}") options);
+      optionsStr = stringifyAttrs "bleopt" cfg.options;
+      facesStr = stringifyAttrs "ble-face" cfg.faces;
+      importsStr = stringifyLists "ble-import" cfg.imports;
+      blerc = pkgs.writeTextFile {
+        name = "blerc";
+        text = concatStringsSep "\n" [
+          optionsStr
+          facesStr
+          importsStr
+          cfg.blercExtra
+        ];
+        checkPhase = ''
+          ${pkgs.stdenv.shellDryRun} "$target"
+        '';
+      };
+    in mkOption {
+      type = types.nullOr (types.either types.path types.str);
+      default = blerc;
+      defaultText = "Path to the generated file";
+      example = "~/.blerc";
+      description = ''
+        Path of the ble init file. This value will be passed to ble.sh as a shell option <code>--rcfile</code>.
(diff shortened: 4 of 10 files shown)


berbiche:
That is indeed how you use `mkBefore` and `mkAfter`. The module system takes care of ordering correctly based on the type defined by the option. Here `bashrcExtra` is a string (`types.lines` or `types.str`) and allows ordering of the declarations.

modules/lib/file-overlays.nix
```diff
@@ -0,0 +1,22 @@
...
+# Latter function will be applied latter and get more precedance.
+[
+  (self: super:
+    pkgs.lib.attrsets.optionalAttrs
+    (with config.programs.blesh; enable && manageBashrc && super ? ".bashrc") {
+      ".bashrc" = super.".bashrc" // {
+        source = pkgs.writeTextFile {
+          inherit (super.".bashrc".source) name;
+          text = ''
+            [[ $- == *i* ]] && source '${config.programs.blesh.package}/share/ble.sh' --noattach
```
akinomyoga, line 13: I was waiting for https://github.com/NixOS/nixpkgs/pull/185866 made by @tars0x9752 to be processed first, but it seems to take time to finally merge it, so let me write this now. As I have commented in the review comments below https://github.com/NixOS/nixpkgs/pull/181963#pullrequestreview-1056052774, the current install location is wrong. Currently, all the related files are put directly under `$prefix/share` (such as `$prefix/share/ble.sh`, `$prefix/share/keymap`, and `$prefix/share/lib`), but all these files should be placed under the directory `$prefix/share/blesh` such as `$prefix/share/blesh/ble.sh`).

Could you first process them including other comments below https://github.com/NixOS/nixpkgs/pull/181963#pullrequestreview-1056052774 (probably after concluding https://github.com/NixOS/nixpkgs/pull/185866) before proceeding with this PR? Also, it would be great if you could edit [the Wiki page](https://github.com/akinomyoga/ble.sh/wiki/Manual-A1-Installation#user-content-nixpkgs) after that.

aiotter: Whoops, I thought the wrong installation path was already fixed on https://github.com/NixOS/nixpkgs/pull/185866 because `installFlags` attribute was removed there, and I completely forgot to fix them in this PR. But as this PR passes the tests without fixing the path, the installation path was actually not fixed yet.

I'll see why the current path is still being wrong without `installFlags` attribute, and also fix the path in this PR. Before these investigation this PR shouldn't be merged, so I'm converting it to draft.

akinomyoga: Thank you for your reply!

> Whoops, I thought the wrong installation path was already fixed on [NixOS/nixpkgs#185866](https://github.com/NixOS/nixpkgs/pull/185866) because `installFlags` attribute was removed there,

I think the installation path isn't fixed by NixOS/nixpkgs#185866 as far as I see lines [[1]](https://github.com/NixOS/nixpkgs/pull/185866/files#diff-a321819ac826b9eccaf69268d655b5826335e81f73da8d822bcd33efaf801ab9R38) and [[2]](https://github.com/NixOS/nixpkgs/pull/185866/files#diff-a321819ac826b9eccaf69268d655b5826335e81f73da8d822bcd33efaf801ab9L41). I have been thinking that we could process this in a separate PR than NixOS/nixpkgs#185866, but now the merging process turned out to be very slow. If @tars0x9752 could agree, now I'm becoming to think we can bundle the fix of the installation path with NixOS/nixpkgs#185866 although it might not be the original purpose of that PR.

aiotter: Ah, I remembered that the PR removes build process and installs ble.sh by just copying the built scripts. So the [line [1]](https://github.com/NixOS/nixpkgs/pull/185866/files#diff-a321819ac826b9eccaf69268d655b5826335e81f73da8d822bcd33efaf801ab9R38) should be moved to `$out/share/blesh` instead of `$out/share`.

I agree that the path is better to be fixed on the PR.

DavSanchez:
Hi! https://github.com/NixOS/nixpkgs/pull/185866 has been merged for a while now. Would this be ready for review?

tannerlegvold:
Bump... @rycee

D3vil0p3r:
@aiotter @akinomyoga what about this PR?

D3vil0p3r:
@aiotter @akinomyoga @rycee what about this PR?

PS: @akinomyoga you should fix the link on [your blesh repository about [home-manager's documentation](https://nix-community.github.io/home-manager/options.html#opt-programs.blesh.enable) at the bottom of the page because the link is old and broken.

akinomyoga:
@aiotter It's been quite a while since the discussion took place, and I'm not a user of Home Manager and unfamiliar with it. So I forgot most of the details of the discussion. Could you take a look at this again?

> PS: @akinomyoga you should fix the link on [your blesh repository about [home-manager's documentation](https://nix-community.github.io/home-manager/options.html#opt-programs.blesh.enable) at the bottom of the page because the link is old and broken.

Ah, thanks for the information. I've fixed it. You could actually edit it by yourself. I'm not restricting the users who can edit the wiki pages.

aiotter:
My bad, I missed the mention.

I got an advice from @SuperSandro2000 about configuration options: https://github.com/NixOS/nixpkgs/pull/201345#issuecomment-1322576091.
As `rcfile` is just a bash script, maybe we should add no configuration options other than `programs.blesh.extraConfig`.

I'm planning to remove `programs.blesh.options`, `programs.blesh.faces`, and `programs.blesh.imports`.
Does anyone have some thought about it?

Current way of configuration:
```nix
  config = {
    programs.blesh = {
      enable = true;
      options = {
        prompt_ps1_transient = "trim:same-dir";
        prompt_ruler = "empty-line";
      };
      faces = { auto_complete = "fg=240"; };
      imports = [ "contrib/bash-preexec" ];
      blercExtra = ''
        function my/complete-load-hook {
          bleopt complete_auto_history=
          bleopt complete_ambiguous=
          bleopt complete_menu_maxlines=10
        };
        blehook/eval-after-load complete my/complete-load-hook
      '';
    };
  };
```

will be changed to:
```nix
  config = {
    programs.blesh = {
      enable = true;
      extraConfig = ''
        bleopt prompt_ps1_transient='trim:same-dir'
        bleopt prompt_ruler='empty-line'
        ble-face auto_complete='fg=240'
        ble-import 'contrib/bash-preexec'
        function my/complete-load-hook {
          bleopt complete_auto_history=
          bleopt complete_ambiguous=
          bleopt complete_menu_maxlines=10
        };
        blehook/eval-after-load complete my/complete-load-hook
      '';
    };
  };
```

akinomyoga:
Thank you for your reply. May I ask where the content of `extraConfig` is supposed to go?

If it's supposed to go into `~/.bashrc`, I'd have to say it should instead be put in `~/.blerc` or `$XDG_CONFIG_HOME/blesh/init.sh`. The settings in `~/.bashrc` are not fetched on `ble-reload` and `ble-update`, so those settings would vanish after `ble-update` in the session.

If it's supposed to be in `~/.blerc`, maybe "extra" sounds strange because there do not seem to be basic parts of the settings.

aiotter:
@akinomyoga If `programs.blesh.rcfile` is `null`(default), the content of the `extraConfig` will become a text file and then it will be loaded by adding the following line at the very head of `~/.bashrc`:
```bash
[[ $- == *i* ]] && source /path/to/ble.sh --attach=none --rcfile=/path/to/generated-text-file
```

And at the very end of the file as well:
```bash
[[ ${BLE_VERSION-} ]] && ble-attach
```

If `programs.blesh.rcfile` is specified, it must be a type of `path` or `str`, and that value will be supplied to `--rcfile` argument.

---

I agree with you that `extraConfig` sounds strange with no other configurable values. However it seems common in home-manager modules.
Please take a look at [programs.bashmount.extraConfig](https://nix-community.github.io/home-manager/options.xhtml#opt-programs.bashmount.extraConfig) for example. It has no other options other than `extraConfig`. It even generates no config file when the user set nothing there.

I'm not familiar with home-manager very much. There may be a better naming.

akinomyoga:
Thank you for the explanation!

> I agree with you that `extraConfig` sounds strange with no other configurable values. However it seems common in home-manager modules.
> Please take a look at [programs.bashmount.extraConfig](https://nix-community.github.io/home-manager/options.xhtml#opt-programs.bashmount.extraConfig) for example.

OK, it indeed seems a convention. Then, let's follow the convention here.

akinomyoga:
@aiotter Can we have another option to switch how the blesh-related codes are arranged in `~/.bashrc`?

I have been recommending people using the combination `source ble.sh --attach=none` & `ble-attach` as a *more* reliable way, but I recently stumbled upon a case where the simple `source ble.sh` works while `source ble.sh --attach=none` & `ble-attach` doesn't work. So I think it would be good to provide an option to switch that.

I'm naively thinking about a config for the attaching strategy such as `programs.blesh.attach = "manual"` (default) producing the configuration mentioned in https://github.com/nix-community/home-manager/pull/3238#issuecomment-1946005014, while `programs.blesh.attach = "prompt"` inserting a single line `[[ $- == *i* ]] && source /path/to/ble.sh --attach=prompt` at the very end of `~/.bashrc`.

aiotter:
@akinomyoga That's indeed possible. I'll implement in that way.

Btw I couldn't find out what is happened when you specify `--attach=prompt`. Is there a document for this option?

akinomyoga:
> Btw I couldn't find out what is happened when you specify `--attach=prompt`. Is there a document for this option?

Ah, I think the explanation is only found in the output of `source /path/to/ble.sh --help`. To begin with, I actually haven't explained the meaning of `--prompt=none` used in the recommended setup.

akinomyoga:
P.S. Actually, `--attach=prompt` is the default strategy, so it can be in principle omitted. However, there is another case where we want to specify it to work around a problem: Some Bash configuration tries to perform `source ~/.bashrc` inside a function. When we only specify the filename to the `source` builtin without arguments, positional parameters `$1`, `$2`, ... in the context of `source` are inherited by the sourced file. Then, if `source /path/to/ble.sh` without arguments is placed in `~/.bashrc`, ble.sh ends up unexpectedly receiving the arguments originally specified to the function that sourced `~/.bashrc`.

Maybe it is easier to see an example:

```bash
$ cat A.sh
# A.sh (.bashrc)
echo "A.sh: $*"
source B.sh
$ cat B.sh
# B.sh (ble.sh)
echo "B.sh: $*"
$ function my_source { echo "my_source: $*"; source "$1"; }
$ my_source A.sh
my_source: A.sh
A.sh: A.sh
B.sh: A.sh
```

For this reason, to inhibit the unexpected inheritance of the positional parameters, we want to specify at least one parameter as a workaround. Then, I think `--attach=prompt` would be a good one because it doesn't change behavior and it also makes the selected attach strategy explicit.

theonekingboo:
how's the state of this now?

RAVENz46:
Still important to me

sbancuz:
Don't know if there is still development for this pr but, in the case direnv support is enabled then ble.sh must be the last thing that gets initialized.

For now my workaround is to disable direnv.bashIntegration and putting this at the end with mkAfter
```
eval "$(/nix/store/823qi3acdfxnvn0ylygjg6jabx7b4q31-direnv-2.34.0/bin/direnv hook bash)"

[[ ${BLE_VERSION-} ]] && ble-attach
```

just-maiyak:
Hello :) Will this PR still be looked into ?

akinomyoga:
@aiotter, may I ask what the current status of this PR is? Is there anything I can help with? If you are busy or don't use ble.sh anymore, maybe someone else can take over the PR. I personally don't use NixOS and Home Manager, so I don't have an environment to test it, though.

aiotter:
@akinomyoga @just-maiyak Sorry for my late reply! I missed the notification.
I'm sorry to say that I do not currently use ble.sh. I had to switch to zsh.
If someone is interested in adding ble.sh into Home Manager, they are free to take over my code.

just-maiyak:
@aiotter no problem, I also ended up switching back to another shell, thanks for replying !

kaBeech:
@akinomyoga I'm considering taking over this PR. Your advice in [#3238 (comment)](https://github.com/nix-community/home-manager/pull/3238#issuecomment-1946053896) about the attaching strategy makes sense. Is there anything else you would want to change/augment in the code/plans discussed here?

Honestly it may be May before I have time to dig into this but I'm starting to think ahead.

> I had to switch to zsh.

FWIW, I discovered `ble.sh` yesterday because my new employer is giving me a MacBook and I was looking for an excuse *not* to switch to zsh. I was almost ready to make the jump (lured in by [ZVM](https://github.com/jeffreytse/zsh-vi-mode)) until I found `ble.sh` :smile_cat:

Thanks @aiotter for all the great work on this!

PR #3240, opened 2022-09-19
Allows lists of strings in extra configuration of `mbsync` (mail accounts).

### LINKED ISSUE #3228: mbsync: extraConfig enum missing variant
As described in [mbsync](https://isync.sourceforge.io/mbsync.html), the following account configuration line is valid (and probably desirable).

```
SSLVersions TLSv1.3 TLSv1.2
```

However, this cannot be represented using the `extraConfigType`.

https://github.com/nix-community/home-manager/blob/5408e27961599b1350b651f88715daf6e67244a7/modules/programs/mbsync-accounts.nix#L7

I _think_ this should allow for `attrsOf (listOf (separatedString " "))`, meaning I can supply the following value.

```nix
SSLVersions = [ "TLSv1.3" "TLSv1.2" ];
```

If I just use a string, then the output (below) results in an error (edit: `Unrecognized SSL version`).

```
SSLVersions "TLSv1.3 TLSv1.2"
```

I'd be happy to make the change, if it's ok.

dbaynard:
### Description

This supports multi-string values, like `SSLVersions TLSv1.3 TLSv1.2`.

Fixes #3228 

### Checklist

    It uses `lib.types.oneOf`, introduced in 2019.

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/mbsync-accounts.nix
+++ b/modules/programs/mbsync-accounts.nix
@@ -4,7 +4,8 @@ with lib;
 
 let
 
-  extraConfigType = with lib.types; attrsOf (either (either str int) bool);
+  extraConfigType = with lib.types;
+    attrsOf (oneOf [ str int bool (listOf str) ]);
 
   perAccountGroups = { name, config, ... }: {
     options = {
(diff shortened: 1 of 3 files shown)


rycee:
Thanks! The `oneOf` is no worry, the HM master branch targets Nixpkgs unstable and we already use things that are not even in Nixpkgs 22.05.

Issue #3020, opened 2022-06-13
NIX_PATH: unbound variable on Windows WSL


101v:
I tried to install home manager on Windows WSL (Ubuntu) using the following command:
`nix-shell '<home-manager>' -A install`

It failed with the following error message:

> Creating initial Home Manager generation...
> 
> /nix/store/inphj230qfkd25zpf995p5kggbbyqk44-home-manager/bin/home-manager: line 90: NIX_PATH: unbound variable
> Uh oh, the installation failed! Please create an issue at
> 
>     https://github.com/nix-community/home-manager/issues
> 
> if the error seems to be the fault of Home Manager.

Addtional info:
Kernel Release: 5.10.16.3-microsoft-standard-WSL2
Kernel Version: #1 SMP Fri Apr 2 22:23:49 UTC 2021
Machine: X86_64

101v:
Apologies found the solution in the documentation - on a Non-NixOS we have to add following:
`export NIX_PATH=$HOME/.nix-defexpr/channels:/nix/var/nix/profiles/per-user/root/channels${NIX_PATH:+:$NIX_PATH}`

SuperSandro2000:
Did you source the home-manager file mentioned in the docs?

Issue #2580, opened 2021-12-21
bug: fresh install fails


ErnestKz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

I added and updated the home-manager channel minutes ago on my fresh MacOS and nix install.

```
ek@Ernests-MacBook-Pro ~ % nix-shell '<home-manager>' -A install

Creating initial Home Manager generation...

/nix/store/67hwm7hjzqb5dlw00632mfcb2daady9d-home-manager/bin/home-manager: line 73: NIX_PATH: unbound variable
Uh oh, the installation failed! Please create an issue at

    https://github.com/nix-community/home-manager/issues

if the error seems to be the fault of Home Manager.
```

### System information

```markdown
ek@Ernests-MacBook-Pro ~ % nix-shell -p nix-info --run "nix-info -m"                                                         
 - system: `"x86_64-darwin"`
 - host os: `Darwin 21.2.0, macOS 10.16`
 - multi-user?: `yes`
 - sandbox: `no`
 - version: `nix-env (Nix) 2.5.1`
 - channels(ek): `"home-manager"`
 - channels(root): `"nixpkgs-22.05pre339739.1dd151f0c0c"`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixpkgs`
```

ErnestKz:
My bad, didn't see this: https://github.com/nix-community/home-manager/issues/2564

Issue #2564, opened 2021-12-15
bug: cannot install home-manager


diohabara:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

This is the command I execute.

```bash
nix-shell '<home-manager>' -A install
```

The result is here.

```bash
Creating initial Home Manager generation...

/nix/store/i9zl4x42wwkjy2gbj749aci3y25f4r94-home-manager/bin/home-manager: line 73: NIX_PATH: unbound variable
Uh oh, the installation failed! Please create an issue at

    https://github.com/nix-community/home-manager/issues

if the error seems to be the fault of Home Manager.
```

### System information

```markdown
nix-shell -p nix-info --run "nix-info -m"
 - system: `"x86_64-darwin"`
 - host os: `Darwin 21.2.0, macOS 10.16`
 - multi-user?: `yes`
 - sandbox: `no`
 - version: `nix-env (Nix) 2.4`
 - channels(tuampon): `"home-manager"`
 - channels(root): `"nixpkgs-22.05pre338556.fc58c3761e9"`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixpkgs`
```

shirakaba:
Happening for me as well on a newer macOS, on M1 rather than Intel:

```
 - system: `"aarch64-darwin"`
 - host os: `Darwin 21.1.0, macOS 12.0.1`
 - multi-user?: `yes`
 - sandbox: `no`
 - version: `nix-env (Nix) 2.4`
 - channels(jamie): `"home-manager"`
 - channels(root): `"nixpkgs-22.05pre337993.e6df26a654b"`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixpkgs`
```

This is happening for me on a brand new M1 MacBook Pro with nothing installed and no shell config.

```sh
# Install Nix
sh <(curl -L https://nixos.org/nix/install) --darwin-use-unencrypted-nix-store-volume --daemon

# Install Home Manager
nix-channel --add https://github.com/nix-community/home-manager/archive/master.tar.gz home-manager
nix-channel --update
nix-shell '<home-manager>' -A install
```

<details>
<summary>See console output</summary>

```
this derivation will be built:
  /nix/store/n72syqmx8l29d8nfdlyshxknmhjm0d19-home-manager.drv
these 40 paths will be fetched (26.52 MiB download, 137.42 MiB unpacked):
[...]

PR #3241, opened 2022-09-20
home-manager: Add home-manager search path to EXTRA_NIX_PATH

zhaofengli:
### Description

Since 2.4, the Nix installer [no longer](https://github.com/NixOS/nix/commit/ec9dd9a5aeb9b7fca170d86b8906ffac7e5ef057) adds the `NIX_PATH` environment variable to shell profiles, because Nix now has a built-in search path if it's unset. However, home-manager still relies on it being set and prepends search paths for `home-manager`. Even though the documentations have been updated to include instructions for non-NixOS users, this is a bad first-time experience (https://github.com/nix-community/home-manager/issues/2564, https://github.com/nix-community/home-manager/issues/3020, https://github.com/nix-community/home-manager/issues/2580).

Instead of relying on NIX_PATH being set and prepending to it, let's not touch it at all and add our paths as `-I` arguments. We trust that Nix will use a sensible default.

Tested by installing home-manager on Ubuntu 22.04 with a fresh multi-user Nix installation.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/docs/installation.adoc
+++ b/docs/installation.adoc
@@ -58,16 +58,6 @@ and if you follow a Nixpkgs version 22.11 channel you can run
 $ nix-channel --add https://github.com/nix-community/home-manager/archive/release-22.11.tar.gz home-manager
 $ nix-channel --update
 ----
-+
-On non-NixOS, you may have to add
-+
-[source,bash]
-export NIX_PATH=$HOME/.nix-defexpr/channels:/nix/var/nix/profiles/per-user/root/channels${NIX_PATH:+:$NIX_PATH}
-+
-to your shell (see https://github.com/NixOS/nix/issues/2033[nix#2033]
-and
-https://discourse.nixos.org/t/where-is-nix-path-supposed-to-be-set/16434/8[this
-reply on the Nix Discourse]).
 
 3. Run the Home Manager installation command and create the first Home
 Manager generation:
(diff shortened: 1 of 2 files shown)


zhaofengli:
Hi @rycee, any opinion on this?

ncfavier:
FWIW the "unbound variable" error is easy to fix: `NIX_PATH=…${NIX_PATH:+:"$NIX_PATH"}`.

I guess using `EXTRA_NIX_PATH` should work as well?

zhaofengli:
> FWIW the "unbound variable" error is easy to fix: `NIX_PATH=…${NIX_PATH:+:"$NIX_PATH"}`.

No, because if you set `NIX_PATH` at all, `nixpkgs` must be there as well otherwise legacy channel-based usage is broken.

ncfavier (approved):

ncfavier:
How can I reproduce the problem in a test? If I unset NIX_PATH then I get errors about nixpkgs not being found, instead of NIX_PATH being unbound: https://github.com/ncfavier/home-manager/actions/runs/3881937196/jobs/6621499934#step:5:5

zhaofengli:
Adding `nix-channel --update` [reproduces](https://github.com/zhaofengli/home-manager/actions/runs/3881986152/jobs/6621605972) it.

ncfavier:
Indeed, and throwing in your patch [doesn't](https://github.com/ncfavier/home-manager/actions/runs/3882011263/jobs/6621658314), so I'm satisfied.

PR #3243, opened 2022-09-20
batsignal: add module

loicreynier:
### Description

Adds [batsignal](https://github.com/electrickite/batsignal) module.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/batsignal.nix
+++ b/modules/services/batsignal.nix
@@ -1,0 +1,183 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.services.batsignal;
+
+  commandLine = concatStringsSep " " ([ "${cfg.package}/bin/batsignal" ]
+    ++ optional (cfg.warningLevelPercent != null)
+    "-w ${toString cfg.warningLevelPercent}"
+    ++ optional (cfg.criticalLevelPercent != null)
+    "-c ${toString cfg.criticalLevelPercent}"
+    ++ optional (cfg.criticalLevelPercent != null)
+    "-d ${toString cfg.dangerLevelPercent}"
+    ++ optional (cfg.fullLevelPercent != null)
+    "-f ${toString cfg.fullLevelPercent}"
+    ++ optional (cfg.warningLevelMessage != null)
+    "-W '${cfg.warningLevelMessage}'"
+    ++ optional (cfg.criticalLevelMessage != null)
+    "-C '${cfg.criticalLevelMessage}'"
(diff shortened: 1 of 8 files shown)


loicreynier:
I wanted to add the `dangerLevelCommand` option to the test, but since it generates an entry in the store for the generated shell script I don't know how to deal with it:

```service
[Service]
ExecStart=@batsignal@/bin/batsignal -w 15 -c 5 -d 2 -f 5 -D /nix/store/fv4mq8q355016vraa178bndbqa75w4rx-batsignal-dangercmd -n 'Internal' -m 120 -a 'Batsignal daemon' -I battery
```

loicreynier:
I fixed the test (typo in expected service config), simplified the package option with `mkPackageOption` and rebased into `master`.

loicreynier:
I added the `dangerLevelCommand` to the test using `normalizeStorePaths`. I also added all the other options to the test which is now complete.

modules/services/batsignal.nix
```diff
@@ -0,0 +1,181 @@
...
+  meta.maintainers = with maintainers; [ loicreynier ];
+
+  options.services.batsignal = {
+    enable = mkEnableOption "batsignal battery monitor daemon";
+    package = mkPackageOption pkgs "batsignal" { };
+
+    warningLevelPercent = mkOption {
+      type = types.nullOr (types.ints.between 0 100);
+      default = null;
+      example = 15;
```
sumnerevans, line 43: I think we should line up these defaults with what batsignal does: https://github.com/electrickite/batsignal/blob/master/batsignal.1.in#L34

```suggestion
      default = 15;
      example = 20;
```

modules/services/batsignal.nix
```diff
@@ -0,0 +1,181 @@
...
+      '';
+    };
+
+    warningLevelMessage = mkOption {
+      type = types.nullOr types.lines;
+      default = null;
+      example = literalExpression ''
+        Battery low
+      '';
+      description = "Warning level message.";
```
sumnerevans, line 86: I think this lines up a bit better with the man page: https://github.com/electrickite/batsignal/blob/master/batsignal.1.in#L46

(also do the same for all of the other `*LevelMessage` descriptions)

```suggestion
      description = "The message to show when the battery reaches the warning level.";
```

sumnerevans (changes requested):
Looks good, overall.

loicreynier:
Thanks for the review @sumnerevans, I changed what your requested.

PR #3258, opened 2022-09-23
awesome: fix luaModules using pkgs.lua instead of awesome.lua

necauqua:
Mirroring my change to awesome module in nixos:
https://github.com/NixOS/nixpkgs/commit/b79f9e9b8a822ac64710d67d26b558f9aba6d47b

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/window-managers/awesome.nix
+++ b/modules/services/window-managers/awesome.nix
@@ -6,7 +6,7 @@ let
 
   cfg = config.xsession.windowManager.awesome;
   awesome = cfg.package;
-  getLuaPath = lib: dir: "${lib}/${dir}/lua/${pkgs.luaPackages.lua.luaversion}";
+  getLuaPath = lib: dir: "${lib}/${dir}/lua/${awesome.lua.luaversion}";
   makeSearchPath = lib.concatMapStrings (path:
     " --search ${getLuaPath path "share"}"
     + " --search ${getLuaPath path "lib"}");


rycee:
I don't use Awesome so I can't judge this very well but I assume it is appropriate.

So I'd be happy to merge, just please shorten the commit message summary to max 50 chars or so. No need to put the entire description in the summary, you can write a more expansive description inside the commit message body. See [seven rules](https://cbea.ms/git-commit/#limit-50).

necauqua:
If you use awesome with a non-system lua (system `lua` is `lua5_2` and you set awesome lua to be `lua5_3` for a concrete example) then `getLuaPath` is completely broken - in short the lua version is mentioned twice, once as part of the `lib` var and equal to awesomes lua version and second time as `pkgs.luaPackages.lua.luaversion` (which is "5.2" in current nixpkgs) - so you get a module path which looks like `../5.3/../5.2` and never ever resolves to anything actulally existing, so luaModules never get loaded.

Didn't notice going over 50 chars, will amend

modules/services/window-managers/awesome.nix
```diff
@@ -6,7 +6,7 @@ let
 
   cfg = config.xsession.windowManager.awesome;
   awesome = cfg.package;
-  getLuaPath = lib: dir: "${lib}/${dir}/lua/${pkgs.luaPackages.lua.luaversion}";
+  getLuaPath = lib: dir: "${lib}/${dir}/lua/${awesome.lua.luaversion}";
```
teto, line 9: I would like to avoid out-of-tree reimplementation of lua utilities such as getLuaPath here. I would like to use lua flat installs in the future but that requires packages not to replicate their own lua wrapping code https://github.com/NixOS/nixpkgs/issues/56398#issuecomment-912096841 . 

I have some ongoing work to improve the lua ecosystem https://github.com/NixOS/nixpkgs/pull/177556/files#diff-558d30a16316ebf76080fb16d8b82a341313e48617e224e2acf30b5f64952990 but until then could you use https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/lua-modules/lib.nix#L39 instead ?

rycee: Maybe we can merge this as-is for now and do the improvement in a future PR?

necauqua: Yeah, while I appreciate the effort of reducing duplicate code/boilerplate when dealing with Lua that you seem to be working on, this PR is a oneline fix and just mirrors the same fix I made in nixpkgs a little while ago.

I am totally out of context for the Lua ecosystem improvements you're making and a jump from 'oneline fix' to 'use the entirely new better lua system' was a bit too much for me so I forgot about this PR :sweat_smile:

PR #3266, opened 2022-09-26
Mako systemd

Rosuavio:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -225,7 +225,7 @@ let
     ./services/keynav.nix
     ./services/lieer.nix
     ./services/lorri.nix
-    ./services/mako.nix
+    ./services/mako
     ./services/mbsync.nix
     ./services/mopidy.nix
     ./services/mpd.nix
--- a/modules/services/mako.nix
+++ b/modules/services/mako.nix
@@ -4,13 +4,16 @@ with lib;
 
 let
 
-  cfg = config.programs.mako;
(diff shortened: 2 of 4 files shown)


Rosuavio:
Maybe I should do something like...
https://github.com/nix-community/home-manager/blob/9e7394523eb4f298528d457e316fc752bdf07151/modules/programs/waybar.nix#L198-L212

v4rakh:
Thanks working on this! I recently ran into the situation resulting in `exec failed: No such file or directory`, because `mako` has been spawned by ` systemd --user` which doesn't have the proper `sh` on `PATH`, but [mako expects this being POSIX compliant](https://github.com/emersion/mako/blob/683d8826f29f678bc4752323329da0b3d7a502cd/notification.c#L420). I changed to starting `mako` with `sway` and this works fine, but I think a dedicated user service is the right choice to provide all flexibility needed.

By the way, this popped up with a very simple example of using _actions_ in `mako`:

```
[summary="Screenshot copied to clipboard"]
on-button-right=exec ${pkgs.wl-screenshot}/bin/wl-screenshot upload
```

Rosuavio:
Closed because it seems we don't need a systemd service for mako, and she sh thing seems solved in nixpkgs https://github.com/nix-community/home-manager/issues/3893#issuecomment-1969618171 now.

PR #3268, opened 2022-09-26
kdeconnect: stop using removed alias

sweenu:
### Description

Since https://github.com/NixOS/nixpkgs/commit/d06207386df9a53fe01f8a30130dfc9a839fc8fc, the kdeconnect package cannot be used by its alias.
Other packages might be affected.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


adisbladis (approved):

rycee:
Thanks! This should be fixed now with the merge of #3282.

PR #3269, opened 2022-09-26
borgmatic: add module

DamienCassou:
### Description

add [borgmatic](https://torsion.org/borgmatic/) module.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -343,6 +343,11 @@ Makefile                                              @thiagokokada
 
 /modules/services/betterlockscreen.nix                @SebTM
 
+/modules/programs/borgmatic.nix                       @DamienCassou
+/modules/services/borgmatic.nix                       @DamienCassou
+/tests/modules/programs/borgmatic                     @DamienCassou
+/tests/modules/services/borgmatic                     @DamienCassou
+
 /modules/services/caffeine.nix                        @uvNikita
 
 /modules/services/cbatticon.nix                       @pmiddend
--- a/modules/services/borgmatic.nix
+++ b/modules/services/borgmatic.nix
@@ -1,0 +1,87 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  serviceConfig = config.services.borgmatic;
+  programConfig = config.programs.borgmatic;
+in {
+  meta.maintainers = [ maintainers.DamienCassou ];
+
+  options = {
+    services.borgmatic = {
+      enable = mkEnableOption "Borgmatic service";
+
+      frequency = mkOption {
+        type = types.str;
+        default = "hourly";
+        description = ''
+          How often to run borgmatic when
+          <code language="nix">services.borgmatic.enable = true</code>.
+          This value is passed to the systemd timer configuration as
+          the onCalendar option. See
+          <citerefentry>
+            <refentrytitle>systemd.time</refentrytitle>
+            <manvolnum>7</manvolnum>
+          </citerefentry>
+          for more information about the format.
+        '';
+      };
+    };
+  };
+
+  config = mkIf serviceConfig.enable {
+    assertions = [
+      (lib.hm.assertions.assertPlatform "services.borgmatic" pkgs
+        lib.platforms.linux)
+    ];
+
+    systemd.user = {
+      services.borgmatic = {
+        Unit = {
+          Description = "borgmatic backup";
+          # Prevent borgmatic from running unless the machine is
+          # plugged into power:
+          ConditionACPower = true;
+        };
+        Service = {
+          Type = "oneshot";
+
+          # Lower CPU and I/O priority:
+          Nice = 19;
+          CPUSchedulingPolicy = "batch";
+          IOSchedulingClass = "best-effort";
+          IOSchedulingPriority = 7;
+          IOWeight = 100;
+
+          Restart = "no";
+          LogRateLimitIntervalSec = 0;
+
+          # Delay start to prevent backups running during boot:
+          ExecStartPre = "sleep 3m";
+
+          ExecStart = ''
+            ${pkgs.systemd}/bin/systemd-inhibit \
+              --who="borgmatic" \
+              --why="Prevent interrupting scheduled backup" \
+              ${programConfig.package}/bin/borgmatic \
+                --stats \
+                --verbosity -1 \
+                --list \
+                --syslog-verbosity 1
+          '';
+        };
+      };
+
+      timers.borgmatic = {
+        Unit.Description = "Run borgmatic backup";
+        Timer = {
+          OnCalendar = serviceConfig.frequency;
+          Persistent = true;
+          RandomizedDelaySec = "10m";
+        };
+        Install.WantedBy = [ "timers.target" ];
+      };
+    };
+  };
+}
--- a/tests/modules/services/borgmatic/basic-configuration.nix
+++ b/tests/modules/services/borgmatic/basic-configuration.nix
@@ -1,0 +1,22 @@
+{ config, pkgs, ... }:
+
+{
+  config = {
+    services.borgmatic = {
+      enable = true;
+      frequency = "weekly";
+    };
+
+    test.stubs.borgmatic = { };
+
+    nmt.script = ''
+      assertFileContent \
+        $(normalizeStorePaths home-files/.config/systemd/user/borgmatic.service) \
+        ${./basic-configuration.service}
+
+      assertFileContent \
+        home-files/.config/systemd/user/borgmatic.timer \
+        ${./basic-configuration.timer}
+    '';
+  };
+}
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -748,6 +748,20 @@ in
           A new module is available: 'programs.discocss'.
         '';
       }
+
+      {
+        time = "2022-10-16T19:49:46+00:00";
+        condition = hostPlatform.isLinux;
+        message = ''
+          Two new modules are available:
+
+            - 'programs.borgmatic' and
+            - 'services.borgmatic'.
+
+          use the first to configure the borgmatic tool and the second if you
+          want to automatically run scheduled backups.
+        '';
+      }
     ];
   };
 }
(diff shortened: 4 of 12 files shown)


DamienCassou:
Running all tests fail for me:
```
$ nix-shell --pure tests -A run.all
building '/nix/store/h1n8vmasnrzikwdzbaf2vk1a15cd0sw7-nmt-report-neovim-coc-config.drv'...
error: cannot coerce null to a string

       at /home/cassou/Documents/projects/nix-system/nixpkgs/pkgs/applications/editors/neovim/utils.nix:64:40:

           63|       pluginConfig = p:
           64|         if (p.config or "") != "" then ''
             |                                        ^
           65|           " ${p.plugin.pname or p.plugin.name} {{{
```

But the new tests pass:

```
$ nix-shell --pure tests -A run.borgmatic-basic-configuration
borgmatic-basic-configuration: OK
```

DamienCassou:
This PR introduces a home-manager service because there are 2 systemd units included (a service and a timer). But most of the module really deals with generating the configuration file of borgmatic which has nothing to do with systemd. Should I decompose my module into 2? 1 program and 1 service?

modules/services/borgmatic.nix
```diff
@@ -0,0 +1,298 @@
...
+    };
+  };
+
+  config = mkIf cfg.enable {
+    assertions = [
+      (lib.hm.assertions.assertPlatform "services.borgmatic" pkgs
+        lib.platforms.linux)
+    ];
+
+    # ."borgmatic.d/first.yaml".text = writeConfig cfg.config;
```
rycee, line 239: Can remove commented code.

```suggestion
```

tests/modules/services/borgmatic/basic-configuration.nix
```diff
@@ -0,0 +1,40 @@
...
+
+      actual_source_directory=$($yq ".location.source_directories[0]" $config_file)
+      if [[ $actual_source_directory != ${source_directory} ]]; then
+         fail "Expected source directory to be ${source_directory} but was $actual_source_directory"
+      fi
+
+      one_file_system=$($yq ".location.one_file_system" $config_file)
+      if [[ $one_file_system != "true" ]]; then
+         fail "Expected one_file_system to be true but it was $one_file_system"
+      fi
```
rycee, line 37: Perhaps it would be possible to simply use `assertFileContent` to verify the entire generated file?

modules/services/borgmatic.nix
```diff
@@ -0,0 +1,298 @@
...
+          timer configuration as the onCalendar option.  See
+          <citerefentry>
+            <refentrytitle>systemd.time</refentrytitle>
+            <manvolnum>7</manvolnum>
+          </citerefentry>
+          for more information about the format.
+        '';
+      };
+
+      configs = mkOption {
```
rycee, line 207: Perhaps `backups` instead of `configs`? Mainly to avoid the extremely overloaded "config" word :-)

rycee: If not, then perhaps `backupConfigs`?

modules/services/borgmatic.nix
```diff
@@ -0,0 +1,298 @@
...
+
+  extraConfigOption = mkOption {
+    type = with types; attrsOf (oneOf [ str bool path int ]);
+    default = { };
+    description = "Extra settings.";
+  };
+
+  locationModule = types.submodule {
+    options = {
+      source_directories = mkOption {
```
rycee, line 30: Since this is not using freeform modules, I think it would be more appropriate to use camel case for the options. I.e., `sourceDirectories` instead of `source_directories`.

modules/services/borgmatic.nix
```diff
@@ -0,0 +1,298 @@
...
+          literalExpression ''"''${pkgs.password-store}/bin/pass borg-repo"'';
+      };
+      extraConfig = extraConfigOption;
+    };
+  };
+
+  retentionModule = types.submodule {
+    options = {
+      keep_within = mkNullableOption {
+        type = types.strMatching "[[:digit:]][Hdwmy]";
```
rycee, line 60: How about "14d"? :-)

rycee: Also, the format is different from `frequency` i the consistency check frequency?

modules/services/borgmatic.nix
```diff
@@ -0,0 +1,298 @@
...
+        }) cfg.configs;
+
+    home.packages = [ cfg.package ];
+
+    systemd.user = {
+      services.borgmatic = {
+        Unit = {
+          Description = "borgmatic backup";
+          Wants = [ "network-online.target" ];
+          After = [ "network-online.target" ];
```
rycee, line 253: Unfortunately, `network-online.target` is not available for the user systemd session so I think this will start with no consideration made to network status.

modules/services/borgmatic.nix
```diff
@@ -0,0 +1,298 @@
...
+        default = "hourly";
+        description = ''
+          How often to run borgmatic.  This value is passed to the systemd
+          timer configuration as the onCalendar option.  See
+          <citerefentry>
+            <refentrytitle>systemd.time</refentrytitle>
+            <manvolnum>7</manvolnum>
+          </citerefentry>
+          for more information about the format.
+        '';
```
rycee, line 204: Since this is the only option (beside the package) one that is direcly used by the timer service I would rename the module to `programs.borgmatic` and have an accompanying service module with a single `enable` option. Then can change this description to something like:

```suggestion
        description = ''
          How often to run borgmatic when
          <code language="nix">services.borgmatic.enable = true</code>.
          This value is passed to the systemd timer configuration as
          the onCalendar option. See
          <citerefentry>
            <refentrytitle>systemd.time</refentrytitle>
            <manvolnum>7</manvolnum>
          </citerefentry>
          for more information about the format.
        '';
```

tests/modules/services/borgmatic/basic-configuration.nix
rycee: Could also assert the content of the generated service and timer files…

.github/CODEOWNERS
```diff
@@ -334,6 +334,8 @@ Makefile                                              @thiagokokada
 
 /modules/services/betterlockscreen.nix                @SebTM
 
+/modules/services/borgmatic.nix                       @DamienCassou
+
```
rycee, line 338: Can also add the test directory:

```suggestion
/modules/services/borgmatic.nix                       @DamienCassou
/tests/modules/services/borgmatic                     @DamienCassou

```

rycee:
Thanks for the contribution! Looks quite nice! I do think that the module would benefit from splitting into a program part and a service part. E.g., I use borg for backups and usually prefer to run the backup manually so I would only enable the program module.

modules/services/borgmatic.nix
DamienCassou: fixed

modules/services/borgmatic.nix
```diff
@@ -0,0 +1,298 @@
...
+      };
+      extraConfig = extraConfigOption;
+    };
+  };
+
+  configModule = types.submodule {
+    options = {
+      location = mkOption {
+        type = locationModule;
```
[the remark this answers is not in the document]
DamienCassou: changed

modules/services/borgmatic.nix
DamienCassou: it makes sense, fixed.

modules/services/borgmatic.nix
DamienCassou: I didn't know that, thank you. Removed.

modules/services/borgmatic.nix
DamienCassou: Done. This change makes the implementation of `writeConfig` more complex as each attribute as to be referenced manually now.

modules/services/borgmatic.nix
DamienCassou: > How about "14d"?

you are perfectly right. Fixed and used "14d" as value for the unit test

> the format is different from frequency i the consistency check frequency?

this is expected. One configuration option comes from borg and the other one from borgmatic. Here is the reference documentation:
- https://torsion.org/borgmatic/docs/reference/configuration/
- https://borgbackup.readthedocs.io/en/stable/usage/prune.html

tests/modules/services/borgmatic/basic-configuration.nix
DamienCassou: > Perhaps it would be possible to simply use assertFileContent to verify the entire generated file?

I will do it if you really want me to but I would prefer not because it would make the test more fragile, i.e., the test would fail for unrelated reasons. E.g.,
- the test would fail if `builtins.toYAML` starts sorting keys of the input attribute set;
- the test would fail if `builtins.toYAML` starts output a file that looks like typical YAML instead of [outputting JSON](https://github.com/NixOS/nixpkgs/blob/master/lib/generators.nix#L243) as it does today

By using a YAML parser and checking attributes one by one, I make sure borgmatic will be happy with the configuration file without checking a particular implementation detail of `builtins.toYAML`.

Instead, I added checks for much more options. I hope it is ok with you. If you still prefer me to use `assertFileContent`, I will.

> Could also assert the content of the generated service and timer files…

done. I used `assertFileContent` here.

DamienCassou (commented):
Thank you so much for the detailed feedback! I think I took care of all of it.

modules/services/borgmatic.nix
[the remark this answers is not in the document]
rycee: Seems it still is a submodule? I see

```
location = mkOption {
  type = types.submodule {
    options = {
      sourceDirectories = mkOption {
…
```

DamienCassou:
@rycee would you mind having a second look please? I took care of your feedback.

rycee:
@DamienCassou I think the comment https://github.com/nix-community/home-manager/pull/3269#discussion_r984914078 still holds, there is no need to use submodules in these places so I would prefer having the simpler solution with plain attribute sets.

modules/services/borgmatic.nix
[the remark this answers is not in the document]
DamienCassou: I changed some of them but I forgot others... I'm sorry. I have fixed them all hopefully now.

DamienCassou:
@rycee I hope I took care of all your feedback now. Would you mind having another look please.

rycee:
Thanks! Merged to master now.

PR #3272, opened 2022-09-26
urxvt: fix package name

superherointj:
At nixpkgs, `rxvt_unicode` alias was deprecated in favor of `rxvt-unicode`. Updating accordingly.

https://github.com/NixOS/nixpkgs/blob/master/pkgs/top-level/aliases.nix#L1005

PR #3273, opened 2022-09-27
broot: fix build

Chan-Siu-Man:
### Description

Fix `conf.hjson` path in `broot.nix`, unblocking builds.

For broot version >= 1.14.0, the default config file should be at
[`resources/default-conf/conf.hjson`](https://github.com/Canop/broot/blob/master/resources/default-conf/conf.hjson).

This PR fixes this error:

    FileNotFoundError: [Errno 2] No such file or directory: '/nix/store/dikbik0phcq4lf7al6s1g6imcx2grfqs-broot-1.15.0.tar.gz/resources/default-conf.hjson'

### Checklist

- [x] Code tested through `nix-shell --pure tests -A run.broot`.
  `btop` test fails on master as well.

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/broot.nix
+++ b/modules/programs/broot.nix
@@ -206,16 +206,33 @@ in {
   config = mkIf cfg.enable {
     home.packages = [ cfg.package ];
 
-    xdg.configFile."broot/conf.toml".source =
-      tomlFormat.generate "broot-config" cfg.settings;
+    xdg.configFile."broot" = {
+      recursive = true;
+      source = pkgs.symlinkJoin {
+        name = "xdg.configFile.broot";
+        paths = [
+          (pkgs.writeTextDir "conf.toml" (builtins.readFile
+            (tomlFormat.generate "broot-config" cfg.settings)))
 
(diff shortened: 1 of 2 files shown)


rycee:
Thanks! Just need to fix the test error, squash the commits. and fix the "Commit messages are formatted like" checklist item.

modules/programs/broot.nix
```diff
@@ -138,6 +138,9 @@ let
         "broot --print-shell-function ${shell} > $out"
       }
     '';
+
+  defaultConfIsDirectory = builtins.compareVersions cfg.package.version "1.14.0"
+    >= 0;
```
rycee, line 143: No need for this. We target Nixpkgs unstable and it doesn't have any older versions of broot packaged.

```suggestion
```

dermetfan (approved):
Thanks for finding and fixing my mistake so quickly! I squashed your commits manually to tick the checkboxes.

teto:
there seems to be a broot related issue, not sure if tha'ts related https://github.com/nix-community/home-manager/pull/3272

PR #3284, opened 2022-09-29
oh-my-posh: add module

arjan-s:
### Description
Add a new module for [oh-my-posh](https://ohmyposh.dev/)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/oh-my-posh.nix
+++ b/modules/programs/oh-my-posh.nix
@@ -1,0 +1,94 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+let
+
+  cfg = config.programs.oh-my-posh;
+
+  jsonFormat = pkgs.formats.json { };
+
+  configArgument = if cfg.settings != { } then
+    "--config ${config.xdg.configHome}/oh-my-posh/config.json"
+  else if cfg.useTheme != null then
+    "--config ${cfg.package}/share/oh-my-posh/themes/${cfg.useTheme}.omp.json"
+  else
+    "";
+
+in {
+  meta.maintainers = [ maintainers.arjan-s ];
+
+  options.programs.oh-my-posh = {
+    enable = mkEnableOption "oh-my-posh, a prompt theme engine for any shell";
+
+    package = mkPackageOption pkgs "oh-my-posh" { };
+
+    settings = mkOption {
+      type = jsonFormat.type;
+      default = { };
+      example = literalExpression ''
+        builtins.fromJSON (builtins.unsafeDiscardStringContext (builtins.readFile "''${pkgs.oh-my-posh}/share/oh-my-posh/themes/space.omp.json"))'';
+      description = ''
+        Configuration written to
+        <filename>$XDG_CONFIG_HOME/oh-my-posh/config.json</filename>. See
+        <link xlink:href="https://ohmyposh.dev/docs/configuration/overview"/>
+        for details. The <literal>useTheme</literal> option is ignored when this
+        option is used.
+      '';
+    };
+
+    useTheme = mkOption {
+      type = types.nullOr types.str;
+      default = null;
+      description = ''
+        Use one of the official themes. This should be a name from this list:
+        <link xlink:href="https://ohmyposh.dev/docs/themes"/>. Because a theme
+        is essentially a configuration file, this option is not used when a
+        <literal>configFile</literal> is set.
+      '';
+    };
+
+    enableBashIntegration = mkOption {
+      type = types.bool;
+      default = true;
+      description = ''
+        Whether to enable Bash integration.
+      '';
+    };
+
+    enableZshIntegration = mkOption {
+      type = types.bool;
+      default = true;
+      description = ''
+        Whether to enable Zsh integration.
+      '';
+    };
+
+    enableFishIntegration = mkOption {
+      type = types.bool;
+      default = true;
+      description = ''
+        Whether to enable Fish integration.
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    xdg.configFile."oh-my-posh/config.json" = mkIf (cfg.settings != { }) {
+      source = jsonFormat.generate "oh-my-posh-settings" cfg.settings;
+    };
+
+    programs.bash.initExtra = mkIf cfg.enableBashIntegration ''
+      eval "$(${cfg.package}/bin/oh-my-posh init bash ${configArgument})"
+    '';
+
+    programs.zsh.initExtra = mkIf cfg.enableZshIntegration ''
+      eval "$(${cfg.package}/bin/oh-my-posh init zsh ${configArgument})"
+    '';
+
+    programs.fish.shellInit = mkIf cfg.enableFishIntegration ''
+      ${cfg.package}/bin/oh-my-posh init fish ${configArgument} | source
+    '';
+  };
+}
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -237,6 +237,9 @@ Makefile                                              @thiagokokada
 
 /modules/programs/octant.nix                          @06kellyjac
 
+/modules/programs/oh-my-posh.nix                      @arjan-s
+/tests/modules/programs/oh-my-posh                    @arjan-s
+
 /modules/programs/opam.nix                            @marsam
 
 /modules/programs/openssh.nix                         @rycee
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -800,6 +800,13 @@ in
           A new module is available: 'programs.k9s'.
         '';
       }
+
+      {
+        time = "2022-11-01T23:57:50+00:00";
+        message = ''
+          A new module is available: 'programs.oh-my-posh'.
+        '';
+      }
     ];
   };
 }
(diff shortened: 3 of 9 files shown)


modules/programs/oh-my-posh.nix
```diff
@@ -0,0 +1,79 @@
...
+  themeArgument =
+    optionalString ((cfg.configFile == null) && (cfg.useTheme != null))
+    " --config ${pkgs.oh-my-posh}/share/oh-my-posh/themes/${cfg.useTheme}.omp.json";
+
+in {
+  meta.maintainers = [ maintainers.arjan-s ];
+
+  options.programs.oh-my-posh = {
+    enable = mkEnableOption "oh-my-posh";
+
```
AndersonTorres, line 20:
```suggestion

    package = mkPackageOption pkgs "oh-my-posh" { };
```

modules/programs/oh-my-posh.nix
```diff
@@ -0,0 +1,79 @@
...
+      default = true;
+      type = types.bool;
+      description = ''
+        Whether to enable Fish integration.
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable (mkMerge [{
+    home.packages = [ pkgs.oh-my-posh ];
```
AndersonTorres, line 65:
```suggestion
    home.packages = [  cfg.package ];
```

modules/programs/oh-my-posh.nix
```diff
@@ -0,0 +1,79 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+let
+
+  cfg = config.programs.oh-my-posh;
+
+  configArgument =
+    optionalString (cfg.configFile != null) " --config ${cfg.configFile}";
```
AndersonTorres, line 9:
```suggestion
    optionalString (cfg.configFile != null) "--config ${cfg.configFile}";
```

modules/programs/oh-my-posh.nix
```diff
@@ -0,0 +1,79 @@
...
+
+  themeArgument =
+    optionalString ((cfg.configFile == null) && (cfg.useTheme != null))
+    " --config ${pkgs.oh-my-posh}/share/oh-my-posh/themes/${cfg.useTheme}.omp.json";
```
AndersonTorres, line 13:
```suggestion
    "--config ${pkgs.oh-my-posh}/share/oh-my-posh/themes/${cfg.useTheme}.omp.json";
```

modules/programs/oh-my-posh.nix
```diff
@@ -0,0 +1,79 @@
...
+
+  config = mkIf cfg.enable (mkMerge [{
+    home.packages = [ pkgs.oh-my-posh ];
+
+    programs.bash.initExtra = mkIf cfg.enableBashIntegration ''
+      eval "$(${pkgs.oh-my-posh}/bin/oh-my-posh init bash${configArgument}${themeArgument})"
+    '';
+
+    programs.zsh.initExtra = mkIf cfg.enableZshIntegration ''
+      eval "$(${pkgs.oh-my-posh}/bin/oh-my-posh init zsh${configArgument}${themeArgument})"
```
AndersonTorres, line 72:
```suggestion
      eval "$(${pkgs.oh-my-posh}/bin/oh-my-posh init zsh ${configArgument} ${themeArgument})"
```

modules/programs/oh-my-posh.nix
```diff
@@ -0,0 +1,79 @@
...
+        Whether to enable Fish integration.
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable (mkMerge [{
+    home.packages = [ pkgs.oh-my-posh ];
+
+    programs.bash.initExtra = mkIf cfg.enableBashIntegration ''
+      eval "$(${pkgs.oh-my-posh}/bin/oh-my-posh init bash${configArgument}${themeArgument})"
```
AndersonTorres, line 68:
```suggestion
      eval "$(${pkgs.oh-my-posh}/bin/oh-my-posh init bash ${configArgument} ${themeArgument})"
```

modules/programs/oh-my-posh.nix
```diff
@@ -0,0 +1,79 @@
...
+    '';
+
+    programs.zsh.initExtra = mkIf cfg.enableZshIntegration ''
+      eval "$(${pkgs.oh-my-posh}/bin/oh-my-posh init zsh${configArgument}${themeArgument})"
+    '';
+
+    programs.fish.shellInit = mkIf cfg.enableFishIntegration ''
+      ${pkgs.oh-my-posh}/bin/oh-my-posh init fish${configArgument}${themeArgument} | source
```
AndersonTorres, line 76:
```suggestion
      ${pkgs.oh-my-posh}/bin/oh-my-posh init fish ${configArgument} ${themeArgument} | source
```

AndersonTorres (changes requested):

arjan-s:
Thank you, I have updated the code according to your suggestions.

AndersonTorres:
@rycee ping

modules/programs/oh-my-posh.nix
```diff
@@ -0,0 +1,81 @@
...
+
+in {
+  meta.maintainers = [ maintainers.arjan-s ];
+
+  options.programs.oh-my-posh = {
+    enable = mkEnableOption "oh-my-posh";
+
+    package = mkPackageOption pkgs "oh-my-posh" { };
+
+    configFile = mkOption {
```
rycee, line 23: Could this be a structured `settings` option using `pkgs.formats.json`?

modules/programs/oh-my-posh.nix
```diff
@@ -0,0 +1,81 @@
...
+let
+
+  cfg = config.programs.oh-my-posh;
+
+  configArgument =
+    optionalString (cfg.configFile != null) "--config ${cfg.configFile}";
+
+  themeArgument =
+    optionalString ((cfg.configFile == null) && (cfg.useTheme != null))
+    "--config ${cfg.package}/share/oh-my-posh/themes/${cfg.useTheme}.omp.json";
```
rycee, line 13: Perhaps something more straight forward would be simpler:

```suggestion
  configArgument =
    if cfg.configFile != null then
      "--config cfg.configFile"
    else if cfg.useTheme != null then
      "--config ${cfg.package}/share/oh-my-posh/themes/${cfg.useTheme}.omp.json"
    else
      "";
```

modules/programs/oh-my-posh.nix
arjan-s: Note that a "config" is really a "theme" for oh-my-posh. Defining it as a structured option is probably possible, but it would get very big very fast when defining a theme that way. See for example the default theme: https://github.com/JanDeDobbeleer/oh-my-posh/blob/main/themes/default.omp.json
I expect people to want to take one of the official themes and modify them slightly for their use. Having to first rewrite an entire theme into a `pkgs.formats.json` setting is a lot of work...

modules/programs/oh-my-posh.nix
arjan-s: That is definitely nicer. Thank you!

modules/programs/oh-my-posh.nix
rycee: I think if you want to use an existing JSON file then that shouldn't be a problem using something like

``` nix
programs.oh-my-posh.settings = builtins.fromJSON (
  builtins.readFile "${cfg.package}/share/oh-my-posh/themes/${cfg.useTheme}.omp.json")
```

In principle you could do that and then override individual keys within the theme using regular module system overrides.

arjan-s: Thank you for your input. I will look into how this needs to be done, as I have no idea at the moment. :)

arjan-s: @rycee I updated the module according to your comments. It still works fine when using `useTheme`, but when using a variation of the `fromJSON` line in my local config, I get the error "the string '{theme JSON here}' is not allowed to refer to a store path". Probably this is caused by my lack of deeper understandig of Nix, but perhaps you could help with it?

arjan-s: @rycee friendly ping :)

arjan-s: @rycee Sorry for bothering you again... I'm still hoping for an answer. :)
If I add the code below to my config, I'm getting the error `the string '{theme JSON here}' is not allowed to refer to a store path`.
```
  programs = {
    oh-my-posh = {
      enable = true;
      useTheme = "space";
      settings = builtins.fromJSON (builtins.readFile "${pkgs.oh-my-posh}/share/oh-my-posh/themes/unicorn.omp.json");
    };
  };
```

arjan-s: Okay, I finally figured it out. The correct way to do it is:
```nix
  programs = {
    oh-my-posh = {
      enable = true;
      settings = builtins.fromJSON (builtins.unsafeDiscardStringContext (builtins.readFile "${pkgs.oh-my-posh}/share/oh-my-posh/themes/space.omp.json"));
    };
  };
```

arjan-s:
@rycee I believe this PR is now ready for merging. I've implemented all suggestions and tested all options.
Could you please review again? Thank you!

arjan-s:
I also added the settings example now. Not sure if the syntax is completely correct though, because the docs and manpages don't build [like they should](https://rycee.gitlab.io/home-manager/index.html#sec-guidelines-module-maintainer), with an error from `modules/services/picom.nix`.

arjan-s:
Sorry, forgot to run `nixfmt` before my previous push...

rycee:
Thanks! Merged to master now.

PR #3285, opened 2022-09-29
pls: add module

arjan-s:
### Description
Add module with aliases for [pls](https://dhruvkb.github.io/pls/)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/pls.nix
+++ b/modules/programs/pls.nix
@@ -1,0 +1,36 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.programs.pls;
+
+  aliases = {
+    ls = "${cfg.package}/bin/pls";
+    ll =
+      "${cfg.package}/bin/pls -d perms -d user -d group -d size -d mtime -d git";
+  };
+
+in {
+  meta.maintainers = [ hm.maintainers.kalhauge ];
+
+  options.programs.pls = {
+    enable =
+      mkEnableOption "pls, a modern replacement for <command>ls</command>";
+
+    package = mkPackageOption pkgs "pls" { };
+
+    enableAliases = mkEnableOption "recommended pls aliases";
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    programs.bash.shellAliases = mkIf cfg.enableAliases aliases;
+
+    programs.fish.shellAliases = mkIf cfg.enableAliases aliases;
+
+    programs.zsh.shellAliases = mkIf cfg.enableAliases aliases;
+  };
+}
--- a/tests/modules/programs/pls/bash.nix
+++ b/tests/modules/programs/pls/bash.nix
@@ -1,0 +1,29 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+{
+  config = {
+    programs = {
+      bash.enable = true;
+
+      pls = {
+        enable = true;
+        enableAliases = true;
+        package = config.lib.test.mkStubPackage { outPath = "@pls@"; };
+      };
+    };
+
+    test.stubs.pls = { };
+
+    nmt.script = ''
+      assertFileExists home-files/.bashrc
+      assertFileContains \
+        home-files/.bashrc \
+        "alias ls='@pls@/bin/pls'"
+      assertFileContains \
+        home-files/.bashrc \
+        "alias ll='@pls@/bin/pls -d perms -d user -d group -d size -d mtime -d git'"
+    '';
+  };
+}
--- a/tests/modules/programs/pls/fish.nix
+++ b/tests/modules/programs/pls/fish.nix
@@ -1,0 +1,36 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+{
+  config = {
+    programs = {
+      fish.enable = true;
+
+      pls = {
+        enable = true;
+        enableAliases = true;
+        package = config.lib.test.mkStubPackage { outPath = "@pls@"; };
+      };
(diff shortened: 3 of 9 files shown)


modules/programs/pls.nix
```diff
@@ -0,0 +1,34 @@
...
+in {
+  meta.maintainers = [ hm.maintainers.kalhauge ];
+
+  options.programs.pls = {
+    enable =
+      mkEnableOption "pls, a modern replacement for <command>ls</command>";
+
+    enableAliases = mkEnableOption "recommended pls aliases";
+  };
+
```
berbiche, line 24: It would be useful to have a package option.
```suggestion
  options.programs.pls = {
    enable =
      mkEnableOption "pls, a modern replacement for <command>ls</command>";
      
    package = mkPackageOption pkgs "pls" { };

    enableAliases = mkEnableOption "recommended pls aliases";
  };

```

modules/programs/pls.nix
```diff
@@ -0,0 +1,34 @@
...
+with lib;
+
+let
+
+  cfg = config.programs.pls;
+
+  aliases = {
+    ls = "${pkgs.pls}/bin/pls";
+    ll =
+      "${pkgs.pls}/bin/pls -d perms -d user -d group -d size -d mtime -d git";
```
berbiche, line 12:
```suggestion
    ls = "${cfg.package}/bin/pls";
    ll =
      "${cfg.package}/bin/pls -d perms -d user -d group -d size -d mtime -d git";
```

Generally, my only issue with aliases and commands like these is that they might be garbage collected if the shell session (or terminal) lives long enough for the user (or a scheduled service) to run the `nix-collect-garbage` and invalidate the aliases.

I don't think HM has any good solution to this problem sadly.

modules/programs/pls.nix
```diff
@@ -0,0 +1,34 @@
...
+
+  options.programs.pls = {
+    enable =
+      mkEnableOption "pls, a modern replacement for <command>ls</command>";
+
+    enableAliases = mkEnableOption "recommended pls aliases";
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ pkgs.pls ];
```
berbiche, line 26:
```suggestion
    home.packages = [ cfg.package ];
```

tests/modules/programs/pls/bash.nix
```diff
@@ -0,0 +1,28 @@
...
+
+{
+  config = {
+    programs = {
+      bash.enable = true;
+
+      pls = {
+        enable = true;
+        enableAliases = true;
+      };
```
berbiche, line 13:
```suggestion
      pls = {
        enable = true;
        enableAliases = true;
        package = config.lib.test.mkStubPackage { outPath = "@pls@"; };
      };
```

tests/modules/programs/pls/bash.nix
```diff
@@ -0,0 +1,28 @@
...
+    test.stubs.pls = { };
+
+    nmt.script = ''
+      assertFileExists home-files/.bashrc
+      assertFileContains \
+        home-files/.bashrc \
+        "alias ls='${pkgs.pls}/bin/pls'"
+      assertFileContains \
+        home-files/.bashrc \
+        "alias ll='${pkgs.pls}/bin/pls -d perms -d user -d group -d size -d mtime -d git'"
```
berbiche, line 25: Using the stubbed package,
```suggestion
        "alias ls='@pls@/bin/pls'"
      assertFileContains \
        home-files/.bashrc \
        "alias ll='@pls@/bin/pls -d perms -d user -d group -d size -d mtime -d git'"
```

tests/modules/programs/pls/fish.nix
```diff
@@ -0,0 +1,35 @@
...
+    };
+
+    nmt.script = ''
+      assertFileExists home-files/.config/fish/config.fish
+      assertFileContains \
+        home-files/.config/fish/config.fish \
+        "alias ls '${pkgs.pls}/bin/pls'"
+      assertFileContains \
+        home-files/.config/fish/config.fish \
+        "alias ll '${pkgs.pls}/bin/pls -d perms -d user -d group -d size -d mtime -d git'"
```
berbiche, line 32: Same thing about the stubbed package.

tests/modules/programs/pls/zsh.nix
```diff
@@ -0,0 +1,31 @@
...
+    };
+
+    nmt.script = ''
+      assertFileExists home-files/.zshrc
+      assertFileContains \
+        home-files/.zshrc \
+        "alias ls='${pkgs.pls}/bin/pls'"
+      assertFileContains \
+        home-files/.zshrc \
+        "alias ll='${pkgs.pls}/bin/pls -d perms -d user -d group -d size -d mtime -d git'"
```
berbiche, line 28: Same thing about the stubbed package.

berbiche (changes requested):
Thanks for the contribution!

Overall, there are only a few things to fix.

modules/programs/pls.nix
arjan-s: Okay, I thought I heard that package options were only recommended when there actually were multiple packages for an application. No problem though, I added it now.

modules/programs/pls.nix
arjan-s: I've used these kind of aliases (from exa) for quite a while, and they've never been garbage collected from my .bashrc. Are you sure that's a thing?

arjan-s:
Thank you for your review. I've updated the code according to your suggestions.

berbiche (approved):
LGTM!

PR #3287, opened 2022-09-30
neovim: support for additional configuration languages

Misterio77:
Hello, folks!

Here's my attempt at making the neovim module support any configuration language.

This was mostly motivated by the fact that `extraConfig` currently only works with `viml` (forcing the user to (ab)use `plugins.*.config` to write lua config); as well as the listing of `fennel` and `teal` as supported languages when they're actually silently ignored.

Please let me know if there's anything to be improved (or even if this idea does not make any sense at all).

PS: I've run format in a second commit, hope it helps making reviewing a bit easier.

## Description

The main change is adding a `configLanguages` option. This option, that by default comes with `viml` and `lua` configured, allows the user to add support for any language they prefer, through whatever plugin and/or lua module. Anything that can be called by lua code works (although it's targeted at languages that compile to lua, such as fennel and teal). It includes an example on adding `fennel` support through `moonwalk`, there's also a test that uses both fennel, viml, lua, and teal simultaneously, pretty cool.

The other notable change is making `extraConfig` an attribute set (still compatible with strings by use of `coercedTo`, tests still pass as expected). It allows users to add configurations in any supported language (similarly to `plugins.*.config`).

### Smaller changes
There's a few smaller changes too (if required, I can get these into separate PRs):
- Added a `extraRuntime` option, for adding files similarly to `plugins.*.runtime`, for those that are not coupled to plugins (e.g. a custom colorscheme or ftplugin).
- `plugins` now uses the `coercedTo` type instead of being manually normalized. `coercedTo` gives us normalization for free!
- A few options had type `lines` when it worked exactly the same as `str`. AFAIK, `lines` only matters when regarding options that will merge. This does not apply to `generatedConfigs` (it's `readOnly`!) or `plugins.*.config` (it's impossible to specify configs to a single `plugins` element in two different modules). Changed them to `str`.
- `withRuby` was changed to `types.bool` to match its siblings (python and node).

### Tests

Changes:
- I've added a new test that (as mentioned) configures the editor using four different languages (viml, lua, teal, and fennel), and tests it.
- `no-init`'s viml file path was updated to match the new one. Similarly to the previous change (https://github.com/nix-community/home-manager/commit/bd83eab6220226085c82e637931a7ae3863d9893), this shouldn't affect existing configurations at all.
- `plugin-config` now also tests a lua plugin configuration. The expected files were inlined into the test, IMO it helps declutter the directory and makes understanding the test easier.
- `runtime` now also tests `extraRuntime`.

## Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/neovim.nix
+++ b/modules/programs/neovim.nix
@@ -54,8 +54,7 @@ let
 
       plugin = mkOption {
         type = types.package;
-        description =
-          "Package providing the plugin";
+        description = "Package providing the plugin";
       };
 
       runtime = mkOption {
@@ -89,7 +88,8 @@ let
         description = "Lua packages required to support this language.";
       };
       vimPlugins = mkOption {
-        type = with types; listOf (coercedTo package (v: { plugin = v; }) pluginWithConfigType);
+        type = with types;
+          listOf (coercedTo package (v: { plugin = v; }) pluginWithConfigType);
         default = [ ];
         description = "Vim plugins required to support this language.";
         example = literalExpression ''
@@ -99,7 +99,8 @@ let
       enableScript = mkOption {
         type = with types; nullOr str;
         default = null;
-        description = "Lua script required to enable support for this language.";
+        description =
+          "Lua script required to enable support for this language.";
         example = literalExpression ''
           require("moonwalk").add_loader("fnl", function(src, path)
               return require("fennel").compileString(src, { filename = path })
@@ -125,7 +126,8 @@ let
     };
   };
 
-  langVimPlugins = flatten (mapAttrsToList (_: lang: lang.vimPlugins) cfg.configLanguages);
+  langVimPlugins =
+    flatten (mapAttrsToList (_: lang: lang.vimPlugins) cfg.configLanguages);
   allPlugins = cfg.plugins ++ langVimPlugins ++ optional cfg.coc.enable {
     type = "viml";
     plugin = cfg.coc.package;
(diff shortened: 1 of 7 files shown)


teto:
thanks for trying to get into this. I didn't know about coercedTo and that's a function I definitely wished I would have known earlie, feel free to send a patch with just that change.

It's hard to draw a line where to stop the new features but to me this feels too complex for home-manager, too many new options to maintain.

I dont like the distinction between language plugins: one should be able to just add dependencies and home-manager should automatically deduce which language configuration to enable without config. In the nixpkgs meta field, you can distinguish between python/lua/viml package.
Same could be said with the  `importScript`  option. It doesn't add much convenience compared to `xdg.configFile."nvim/init.lua" = ""`. A "enableFennelNvim" would bring more value but if we start doing this, I am afraid we will have "enableSignifiy" etc which would turn into a maintenance nightmare (except if this lives as an out of tree module and we can change things in this module to help both to work better).

Pegasust:
Just to clarify from @teto, if you have existing config with `init.lua` instead of `init.vim`, you could hack together something like:

```nix
# home.nix
{
...
  xdg.configFile."nvim/init.lua".text = builtins.readFiles /path/to/localfs/nvim/init.lua
}
```

Not sure why this isn't documented in [https://mipmip.github.io/home-manager-option-search/?xdg.configFile](home-manager options search: xdg.configFile), probably it's undocumented in the source code.

Misterio77:
Hey, sorry for the delay; have been a little busy lately.

I put this together mostly to deal with the inconsistency of only being able to inline vimscript, while having the possibility to configure with viml or lua. I don't really use teal or fennel, just added the support as it was a small change.

> to me this feels too complex for home-manager, too many new options to maintain.

Perhaps we could support just lua and vim script (on both `extraConfig` and `plugins`), and if people use a name that is not vim or lua, output to a file with that extension? That would make it possible to use any language without adding new options.

> you could hack together something like

Thanks for the clarification! Yeah, I'm aware of this possibility; I usually prefer configuring with `extraConfig` and `plugins`, so I mostly just have a painpoint with the inconsistency.

ncfavier:
Note that [`programs.neovim.extraLuaConfig`](https://nix-community.github.io/home-manager/options.html#opt-programs.neovim.extraLuaConfig) exists now.

Misterio77:
> Note that [`programs.neovim.extraLuaConfig`](https://nix-community.github.io/home-manager/options.html#opt-programs.neovim.extraLuaConfig) exists now.

That's definitely a good enough solution for me. Thanks for letting me know :)

Closing this for now, then. If anyone wants to reuse something from this for another PR, feel free!

PR #3291, opened 2022-09-30
imapnotify: Add launchd agent

dbaynard:
### Description

This extends the imapnotify (goimapnotify) configuration with a launchd agent that matches the systemd unit (and shares as much code as possible).

~Annoyingly I'd forgotten that `goimapnotify` is malconfigured in `nixpkgs`, in that it works (and is supported) on macos, but the metadata `platforms` are linux only. So while this is blocked on that, I'm making the PR now, and I'll update it when the upstream change is merged.~

Edit: NixOS/nixpkgs#193797 (now merged, ~awaiting~ and deployed [deployment to nixos-unstable](https://nixpk.gs/pr-tracker.html?pr=193797))

~I added a semi-golden test: it uses the `goimapnotify` store path, so I've passed that as a parameter. I've seen the `pkgs.writeText` pattern used in other tests, but I didn't spot this case, so I hope it's ok.~

I've seen that #2915 is still open; if this is merged first, there'll be additonal redundant platform checking, from this.

I've a few comments that I'll put in a self-review.

### Checklist

- [x] Change is backwards compatible. (but the version of `goimapnotify` that lists darwin in platforms is only in `master`, at the moment)

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/imapnotify.nix
+++ b/modules/services/imapnotify.nix
@@ -37,6 +37,30 @@ let
       };
     };
 
+  genAccountAgent = account:
+    let name = safeName account.name;
+    in {
+      name = "imapnotify-${name}";
+      value = {
+        enable = true;
+        config = {
+          ProgramArguments = [
+            "${getExe cfg.package}"
+            "-conf"
+            "${config.xdg.configHome}/imapnotify/${configName account}"
+          ];
+          KeepAlive = true;
+          ThrottleInterval = 30;
+          ExitTimeOut = 0;
+          ProcessType = "Background";
+          RunAtLoad = true;
+        } // optionalAttrs account.notmuch.enable {
+          EnvironmentVariables.NOTMUCH_CONFIG =
+            "${config.xdg.configHome}/notmuch/default/config";
+        };
+      };
+    };
+
   genAccountConfig = account:
     pkgs.writeText (configName account) (let
       port = if account.imap.port != null then
@@ -92,8 +116,6 @@ in {
             + concatMapStringsSep ", " (a: a.name) badAccounts;
         };
     in [
-      (lib.hm.assertions.assertPlatform "services.imapnotify" pkgs
-        lib.platforms.linux)
       (checkAccounts (a: a.maildir == null) "maildir configuration")
       (checkAccounts (a: a.imap == null) "IMAP configuration")
       (checkAccounts (a: a.passwordCommand == null) "password command")
@@ -101,6 +123,8 @@ in {
     ];
 
     systemd.user.services = listToAttrs (map genAccountUnit imapnotifyAccounts);
+
+    launchd.agents = listToAttrs (map genAccountAgent imapnotifyAccounts);
 
     xdg.configFile = listToAttrs (map (account: {
       name = "imapnotify/${configName account}";
--- a/modules/services/imapnotify.nix
+++ b/modules/services/imapnotify.nix
@@ -8,6 +8,8 @@ let
 
   safeName = lib.replaceStrings [ "@" ":" "\\" "[" "]" ] [ "-" "-" "-" "" "" ];
 
+  configName = account: "imapnotify-${safeName account.name}-config.json";
+
   imapnotifyAccounts =
     filter (a: a.imapnotify.enable) (attrValues config.accounts.email.accounts);
 
@@ -19,7 +21,10 @@ let
         Unit = { Description = "imapnotify for ${name}"; };
 
         Service = {
-          ExecStart = "${getExe cfg.package} -conf ${genAccountConfig account}";
+          ExecStart =
+            "${getExe cfg.package} -conf '${config.xdg.configHome}/imapnotify/${
+              configName account
+            }'";
           Restart = "always";
           RestartSec = 30;
           Type = "simple";
@@ -33,7 +38,7 @@ let
     };
 
   genAccountConfig = account:
-    pkgs.writeText "imapnotify-${safeName account.name}-config.json" (let
+    pkgs.writeText (configName account) (let
       port = if account.imap.port != null then
         account.imap.port
       else if account.imap.tls.enable then
@@ -96,5 +101,10 @@ in {
     ];
 
     systemd.user.services = listToAttrs (map genAccountUnit imapnotifyAccounts);
+
+    xdg.configFile = listToAttrs (map (account: {
+      name = "imapnotify/${configName account}";
+      value.source = genAccountConfig account;
+    }) imapnotifyAccounts);
   };
 }
--- a/modules/services/imapnotify.nix
+++ b/modules/services/imapnotify.nix
@@ -19,9 +19,7 @@ let
         Unit = { Description = "imapnotify for ${name}"; };
 
         Service = {
-          ExecStart = "${pkgs.goimapnotify}/bin/goimapnotify -conf ${
-              genAccountConfig account
-            }";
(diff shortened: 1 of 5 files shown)


modules/services/imapnotify.nix
```diff
@@ -34,6 +34,28 @@ let
       };
     };
 
+  genAccountAgent = account:
+    let name = safeName account.name;
```
dbaynard, line 38: `safeName` converts `hw@example.com` to `hw-example.com` but for the `plist` label it should arguably replace the `.` with a `-` too (for `hw-example-com`).

tests/modules/programs/goimapnotify/default.nix
```diff
@@ -0,0 +1 @@
+{ goimapnotify-launchd = ./launchd.nix; }
```
dbaynard, line 1: I've only tested this on darwin — I don't know whether the `launchd` tests run on linux (I can test this, though).

dbaynard: Ah, tests are passing in CI.

dbaynard:
`goimapnotify` at `master` has the correct platforms for darwin (though I'm not sure this is being backported). All builds in `nixpkgs` suported by `home-manager` _should_ work, but require the following override to build (I'm sure there's a more elegant way of doing it with `lib.recursiveUpdate` but 🤷 ). I’m not sure how to handle these circumstances.

```nix
drv: drv.overrideAttrs (old: {
    meta = old.meta or { } // {
      platforms = old.meta.platforms or [ ] ++ lib.platforms.darwin;
    };
  });
```

dbaynard:
I'm just pushing a commit that should get macos CI showing green.

dbaynard:
NixOS 22.11 has the working version of `goimapnotify` so I'll make the necessary changes to this in the next week.

dbaynard:
I'm happy with this, the tests work, and the upstream dependency requires no changes. Ready to go!

modules/services/imapnotify.nix
```diff
@@ -34,8 +37,32 @@ let
...
+    in {
+      name = "imapnotify-${name}";
+      value = {
+        enable = true;
+        config = {
+          ProgramArguments = [
+            "${cfg.package}/bin/${cfg.executable}"
```
ncfavier, line 48:
```suggestion
            (getExe cfg.package)
```

modules/services/imapnotify.nix
```diff
@@ -79,14 +124,21 @@ in {
...
       (checkAccounts (a: a.passwordCommand == null) "password command")
       (checkAccounts (a: a.userName == null) "username")
     ];
 
-    systemd.user.services = listToAttrs (map genAccountUnit imapnotifyAccounts);
+    systemd.user.services = lib.mkIf (!pkgs.stdenv.isDarwin)
+      (listToAttrs (map genAccountUnit imapnotifyAccounts));
+
+    launchd.agents = lib.mkIf (pkgs.stdenv.isDarwin)
+      (listToAttrs (map genAccountAgent imapnotifyAccounts));
```
ncfavier, line 137: You can get rid of the `mkIf`s after rebasing

modules/services/imapnotify.nix
```diff
@@ -19,9 +21,10 @@ let
         Unit = { Description = "imapnotify for ${name}"; };
 
         Service = {
-          ExecStart = "${pkgs.goimapnotify}/bin/goimapnotify -conf ${
-              genAccountConfig account
-            }";
+          ExecStart =
+            "${cfg.package}/bin/${cfg.executable} -conf '${config.xdg.configHome}/imapnotify/${
```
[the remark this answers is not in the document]
dbaynard: Ah nice; TIL about `meta.mainProgram`.

dbaynard:
Thanks for the review — I'll rebase, make the changes and confirm everything is ok, then ping for re-review.

dbaynard:
This is still on my roadmap; I'll get to it in the next few months (hopefully soon).

dbaynard:
I've rebased but I haven't been able to figure out the test failure.

Here's my test.

```nix
    nmt.script = let
      serviceFileName =
        "org.nix-community.home.imapnotify-hm-example.com.plist";
    in ''
      serviceFile=LaunchAgents/${serviceFileName}
      assertFileExists $serviceFile
      assertFileContent $serviceFile ${./launchd.plist}
    '';
```

```console
Expected LaunchAgents/org.nix-community.home.imapnotify-hm-example.com.plist to be same as /nix/store/shpsifrkxyil9gd65295q5kx0ihyqvwx-launchd.plist but were different:
--- actual
+++ expected
@@ -17,7 +17,7 @@
        <string>Background</string>
        <key>ProgramArguments</key>
        <array>
-               <string>/nix/store/ib1yskg8kk1hxzn32yi3sjjpphq45sna-goimapnotify/bin/goimapnotify</string>
+               <string>@goimapnotify@/bin/goimapnotify</string>
                <string>-conf</string>
                <string>/home/hm-user/.config/imapnotify/imapnotify-hm-example.com-config.json</string>
        </array>

```

I don't see how the call to `assertFileContent` differs from the `helix` example.

```nix
''
      assertFileContent \
        home-files/.config/helix/languages.toml \
        ${./languages-expected.toml}
''
```

In both cases the relevant parameter comes from `lib.test.mkStubPackage { name = "pkgname" }`. I tried with the argument `outPath`, with `name` and with both, and in all cases it generated (a variant of) the error here, where the generated code includes a store path, not the path relative to `@goimapnotify@`.

It's plausible that this behaves differently on launchd units — I haven't looked through the code for the tests.

Any suggestions?

ncfavier:
The problem is that `getBin` does not play well with our stubs, because we just naively override `outPath` but this does not take care of `pkg.out`. I'll try a fix later.

dbaynard:
OK; no rush on my part. Let me know if you'd like me to take a look at it, too. Thanks!

ncfavier:
https://github.com/nix-community/home-manager/pull/4050

This was not the only problem with this PR: you'll also have to pass `outPath = "@goimapnotify@";` to `mkStubPackage`, as the other tests do.

ncfavier:
Rebasing to master should fix the tests now.

kciredor:
Very eager to use this! Thanks @dbaynard and @ncfavier for the upcoming review 😀

ncfavier (approved):

kciredor:
That was fast, thanks 🥳

modules/services/imapnotify.nix
```diff
@@ -19,9 +21,10 @@ let
         Unit = { Description = "imapnotify for ${name}"; };
 
         Service = {
-          ExecStart = "${pkgs.goimapnotify}/bin/goimapnotify -conf ${
-              genAccountConfig account
-            }";
+          ExecStart =
+            "${getExe cfg.package} -conf '${config.xdg.configHome}/imapnotify/${
```
rycee, line 25: Won't this prevent automatic restart of the service when the account configuration changes?

rycee: Had a closer look and yeah it doesn't look like this will work. To be able to restart the service on configuration change we must ensure that the unit file content change between the generations. I would propose to keep using the direct Nix store path here. You can still create the link in config home, if you want.

dbaynard: Ah, please forgive me.  I've made #4212 though I can't test it right now.

PR #3292, opened 2022-10-01
sioyek: enable multiple bindings for the same command

foo-dogsquared:
### Description

In Sioyek, [multiple keybindings can be set for the same command](https://github.com/ahrm/sioyek/blob/749a88fb927419ec85f795881c8f0c76600a06b3/pdf_viewer/keys.config#L26-L27). It would be nice for home-manager to be able to set that as well.

Since the configuration is getting a teeeeeny more complex, I also added a test.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/sioyek.nix
+++ b/modules/programs/sioyek.nix
@@ -1,11 +1,16 @@
 { config, lib, pkgs, ... }:
+
 with lib;
+
 let
+
   cfg = config.programs.sioyek;
 
-  renderAttrs = attrs:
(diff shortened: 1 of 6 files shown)


foo-dogsquared:
cc @podocarp

podocarp:
Did not know this at first! Thanks a lot.

rycee:
Thanks! Merged to master. Note, I switched the rendering over to use `generators.toKeyValue`.

PR #3294, opened 2022-10-02
ci: introduce labeler workflow

teto:
automatically tag PRs depending on touched files

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,0 +1,24 @@
+name: "Label PR"
+
+on:
+  - pull_request_target
+
+# WARNING:
+# When extending this action, be aware that $GITHUB_TOKEN allows some write
+# access to the GitHub API. This means that it should not evaluate user input in
+# a way that allows code injection.
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  labels:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/labeler.yml
```diff
@@ -0,0 +1,25 @@
+name: "Label PR"
+
+on:
+  pull_request_target:
+    types: [edited, opened, synchronize, reopened]
```
marsam, line 5: I think this line is not necessary. By default this event is run on `opened`, `synchronize`, and `reopened`. 
Nixpkgs adds `edited` to cover when a PR changes its base branch, which is very unlikely in home-manager
```suggestion
```

.github/workflows/labeler.yml
```diff
@@ -0,0 +1,24 @@
+name: "Label PR"
+
+on:
+  pull_request_target:
```
marsam, line 4: nitpick
```suggestion
  - pull_request_target
```

marsam (approved):

.github/workflows/labeler.yml
teto: it's not equivalent is it ? from a dict to a list. Wouldn't it be better to remove it altogether since it doesn't seem used

marsam (approved):

rycee:

PR #3295, opened 2022-10-03
sbt: cache `passwordCommand` output

plaflamme:
### Description

This will cache the output of `passwordCommand` per authentication realm.

Context: the `credentials` key in `sbt` is a `TaskKey[Seq[Credentials]]`. In `sbt`, tasks are evaluated on-demand and their output is not cached. This particular key is referenced by all submodules in a project. When the command is relatively expensive (e.g.: `pass show foo`), this results in several seconds of delay when doing basic things like `compile` or `test` which makes this unusable without some kind of caching.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

  - ~[ ] Added myself and the module files to `.github/CODEOWNERS`.~

### DIFF
--- a/modules/programs/sbt.nix
+++ b/modules/programs/sbt.nix
@@ -8,13 +8,16 @@ let
     addSbtPlugin("${plugin.org}" % "${plugin.artifact}" % "${plugin.version}")
   '';
 
-  renderCredential = cred: ''
-    credentials += Credentials("${cred.realm}", "${cred.host}", "${cred.user}", "${cred.passwordCommand}".!!.trim)
-  '';
+  renderCredential = idx: cred:
+    let symbol = "credential_${toString idx}";
+    in ''
+      lazy val ${symbol} = "${cred.passwordCommand}".!!.trim
+      credentials += Credentials("${cred.realm}", "${cred.host}", "${cred.user}", ${symbol})
+    '';
 
   renderCredentials = creds: ''
     import scala.sys.process._
-    ${concatStrings (map renderCredential creds)}'';
+    ${concatStrings (imap0 renderCredential creds)}'';
 
   renderRepository = value:
     if isString value then ''
--- a/tests/modules/programs/sbt/credentials.nix
+++ b/tests/modules/programs/sbt/credentials.nix
@@ -19,8 +19,10 @@ let
   ];
   expectedCredentialsSbt = pkgs.writeText "credentials.sbt" ''
     import scala.sys.process._
-    credentials += Credentials("Sonatype Nexus Repository Manager", "example.com", "user", "echo password".!!.trim)
-    credentials += Credentials("Sonatype Nexus Repository Manager X", "v2.example.com", "user1", "echo password1".!!.trim)
+    lazy val credential_0 = "echo password".!!.trim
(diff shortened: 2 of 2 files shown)


rycee:
Maintainer CC: @kubukoz

plaflamme:
FWIW: this assumes `Caffeine` is on the `sbt` classpath. Since we don't really use any of the fancy `Caffeine` features (which might be interesting), I can re-write this to use `ConcurrentHashMap` (or similar) so we only depend on jdk / scala types.

kubukoz:
_is_ caffeine included with sbt?

a CHM is probably good enough. Maybe we could even make a `lazy val` for each cred's password?

plaflamme:
@kubukoz yes, `Caffeine` is used in `sbt`, but it's not part of `sbt`'s API; so assuming its is a risk. I'll remove it, it's just not useful.

Also FWIW, I originally went with a `lazy val` approach, but switched to a map because it was less clunky in the `nix` code here. For declaring a `val`, you have to come up with a valid scala symbol, which was a little annoying. I'm happy to rewrite to that approach since it'd be based on less assumptions in general.

kubukoz:
yeah, you're right, it's probably not worth the difference. CHM is available since at least Java 8, which is a 2.12 (so, sbt 1.0) requirement, so it should be fine (Caffeine does indeed sound like it could disappear from the classpath any day).

plaflamme:
ok @kubukoz [here's](https://github.com/nix-community/home-manager/pull/3295/commits/22acef89de3ba4ddc292c0081a49217a3575d912) the `lazy val` approach. WDYT?

kubukoz:
I'm satisfied :) I'll give this a spin locally.

kubukoz:
I'm getting compilation errors:

```
/Users/kubukoz/.sbt/1.0/credentials.sbt:4: error: credential_41dbb854a44088597c878eb68ff6c354e1c6fbe6792fcebdfa9554ce293bf01a is already defined as lazy value credential_41dbb854a44088597c878eb68ff6c354e1c6fbe6792fcebdfa9554ce293bf01a
lazy val credential_41dbb854a44088597c878eb68ff6c354e1c6fbe6792fcebdfa9554ce293bf01a = "cat /Users/kubukoz/secrets/sonatype.txt".!!.trim
         ^
sbt.compiler.EvalException: Type error in expression
[error] sbt.compiler.EvalException: Type error in expression
```

looks like my realms aren't unique. It might be a common case... WDYT about using the host for the hash instead? Better yet, maybe a host+username combination (but I'm not sure if it's possible to have multiple credentials per host, maybe it's worth checking)

plaflamme:
Yeah, I was afraid of that, thankfully using the `lazy val` based approach caught this. It would have silently been broken with the `Map`-based approach. Thanks for testing, I'll add a case in the tests while fixing this up.

kubukoz:
oof, yeah, that's a good point. :)

plaflamme:
@kubukoz I fixed it by simply using the index in the credentials list instead. Not sure why I went with hashing in the first place since this is a lot simpler and more robust. PTAL [here](https://github.com/nix-community/home-manager/pull/3295/commits/3596ab87b231c20ad5da765398f0bca32d63f509).

kubukoz (approved):

rycee:
Thanks! Merged to master now.

PR #3296, opened 2022-10-03
sbt: allow managing the `~/.sbt/repositories` file

plaflamme:
### Description

`sbt` allows overriding the default repositories to use to resolve
dependencies. This is often used with proxies and/or private
repositories to host internal packages.

This change adds a `repositories` attribute to `sbt` to allow specifying
the values that will go in `~/.sbt/repositories` file.

NOTE: I had to use `${baseConfigPath}/../repositories` to place the file in the right place. This seems brittle, but I can't find a better way without breaking compatibility (which seems worse).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

  - ~[ ] Added myself and the module files to `.github/CODEOWNERS`.~

### DIFF
--- a/modules/programs/sbt.nix
+++ b/modules/programs/sbt.nix
@@ -16,6 +16,17 @@ let
     import scala.sys.process._
     ${concatStrings (map renderCredential creds)}'';
 
+  renderRepository = value:
+    if isString value then ''
+      ${value}
+    '' else ''
+      ${concatStrings (mapAttrsToList (name: value: "${name}: ${value}") value)}
+    '';
+
+  renderRepositories = repos: ''
+    [repositories]
+    ${concatStrings (map renderRepository cfg.repositories)}'';
+
   sbtTypes = {
     plugin = types.submodule {
       options = {
@@ -68,6 +79,11 @@ let
   cfg = config.programs.sbt;
 
 in {
+  imports = [
+    (mkRemovedOptionModule [ "programs" "sbt" "baseConfigPath" ]
+      "Use programs.sbt.baseUserConfigPath instead, but note that the semantics are slightly different.")
+  ];
+
   meta.maintainers = [ maintainers.kubukoz ];
 
   options.programs.sbt = {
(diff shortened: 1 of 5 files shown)


rycee:
Maintainer CC: @kubukoz

kubukoz:
About `baseConfigPath`, I'd say breaking compat would be fine as long as we pick a different name and have the old one issue a warning/error.

Not sure about the path traversal, I'm fine with it but it's surely not ideal.

plaflamme:
Ok, I think this deprecation path might not be too bad:

* add, say `baseSbtConfigPath`, and have it point at `~/.sbt` by default 
* make the existing `baseConfigPath` have an empty default value
* then at "runtime":
  * if `baseConfigPath` is undefined
    * proceed with using `baseSbtConfigPath` throughout
  * if `baseConfigPath` is defined:
    * issue a deprecation warning
    * use `baseConfigPath`for `credentials` and `plugins` as before
    * if `repositories` is used:
      * exit with an error requesting to switch to new `baseSbtConfigPath`

Does that work? If so, I'm not 100% confident my nix-foo is up to par, but I can take a stab.

kubukoz:
Yes, that sounds perfect! Let me know if you need help - I haven't worked on HM in a while but might help on the pure "nix side".

plaflamme:
@kubukoz while investigating this, I found out about `mkRemovedOptionModule`. I went ahead and simply deprecated the option in favour of a new one in [this commit](https://github.com/nix-community/home-manager/pull/3296/commits/b06cd4e54251c64f2ae08fdfdf7d4aaf86213b59). I believe this is what you suggested originally. PTAL. I can still try the other more complicate approach, but this seems like a fine tradeoff since I doubt users are actually specifying this option.

kubukoz:
I'm fine with either, `mkRemovedOptionModule` looks great!

kubukoz (approved):
Great work!

kubukoz:
@rycee can you help get this merged?

plaflamme:
Thanks! And thanks for reviewing!

rycee:
Thanks for the contribution and productive reviewing! Merged to master now.

Issue #2752, opened 2022-02-22
bug: swayidle command not generating correctly


harrisonthorne:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Based on the source code for [escapeShellArg](https://github.com/NixOS/nixpkgs/blob/7f9b6e2babf232412682c09e57ed666d8f84ac2d/lib/strings.nix#L318) and the swayidle module, I'm expecting my timeout commands to be wrapped in single quotes.

However, the module is producing this instead:

```console
/nix/store/6slnlma1imvpk2ih91xp54n7hpgff0qa-swayidle-1.7.1/bin/swayidle -w timeout 570 notify-send -u low -t 29500 -- 'Are you still there?' 'Your system will lock itself soon.' timeout 600 $HOME/.config/sway/scripts/lock.fish --bg-color 002b36 --fg-color 93a1a1 --primary-color 2aa198 --warning-color ffaa00 --error-color ff0000 timeout 630 swaymsg 'output * dpms off' resume swaymsg 'output * dpms on' before-sleep $HOME/.config/sway/scripts/lock.fish --bg-color 002b36 --fg-color 93a1a1 --primary-color 2aa198 --warning-color ffaa00 --error-color ff0000
```

which gives me this error in the systemd unit, `systemctl --user status swayidle.service`:

```
Feb 22 13:38:49 ponytower systemd[3002]: Started Idle manager for Wayland.
Feb 22 13:38:49 ponytower swayidle[64891]: 2022-02-22 13:38:49 - [Line 880] Unsupported command '630'
Feb 22 13:38:49 ponytower systemd[3002]: swayidle.service: Main process exited, code=exited, status=255/EXCEPTION
Feb 22 13:38:49 ponytower systemd[3002]: swayidle.service: Failed with result 'exit-code'.
```

[Link to swayidle commands in my config](https://codeberg.org/harrisonthorne/nixos-config/src/commit/533ad87742844b11b788cfc4b15fd6e81157cae7/homes/harrison/services.nix#L33)

### Maintainer CC

@c0deaddict 

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 5.10.101, NixOS, 22.05 (Quokka)`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.7.0pre20220127_558c4ee`
 - channels(root): `"nixos-22.05pre353535.48d63e924a2"`
[...]

PR #3298, opened 2022-10-04
services/swayidle: fix examples

wuyoli:
### Description
The example did not work before, since the systemd service does not have swayidle in its $PATH.

### Checklist

  did not go through, but error is unrelated?

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/swayidle.nix
+++ b/modules/services/swayidle.nix
@@ -73,7 +73,7 @@ in {
       default = [ ];
       example = literalExpression ''
         [
-          { timeout = 60; command = "swaylock -fF"; }
+          { timeout = 60; command = "${pkgs.swaylock}/bin/swaylock -fF"; }
         ]
       '';
       description = "List of commands to run after idle timeout.";
@@ -84,7 +84,7 @@ in {
       default = [ ];
       example = literalExpression ''
         [
-          { event = "before-sleep"; command = "swaylock"; }
+          { event = "before-sleep"; command = "${pkgs.swaylock}/bin/swaylock"; }
           { event = "lock"; command = "lock"; }
         ]
       '';


wuyoli:
This solves a problem, which was mentioned by @PedroHLC in #2752

modules/services/swayidle.nix
```diff
@@ -73,7 +73,7 @@ in {
       default = [ ];
       example = literalExpression ''
         [
-          { timeout = 60; command = "swaylock -fF"; }
+          { timeout = 60; command = "${pkgs.swaylock}/bin/swaylock -fF"; }
```
uvNikita, line 76:
```suggestion
          { timeout = 60; command = ''"''${pkgs.swaylock}/bin/swaylock -fF"''; }
```

modules/services/swayidle.nix
```diff
@@ -84,7 +84,7 @@ in {
       default = [ ];
       example = literalExpression ''
         [
-          { event = "before-sleep"; command = "swaylock"; }
+          { event = "before-sleep"; command = "${pkgs.swaylock}/bin/swaylock"; }
```
uvNikita, line 87:
```suggestion
          { event = "before-sleep"; command = ''"''${pkgs.swaylock}/bin/swaylock"''; }
```

uvNikita (changes requested):
We need to escape `${...}` expressions, see e.g: https://github.com/nix-community/home-manager/blob/e4e639dd4dc3e431aa5b5f95325f9a66ac7e0dd9/modules/misc/gtk.nix#L154

modules/services/swayidle.nix
PedroHLC: I think you want the parameters (`-fF`) on the outside of the quotes

wuyoli:
@uvNikita 
yes, now, that you say it it is obvius, that this has to be escaped. But why wouldn't i just use a backslash?

nix manual says
>The special characters `"` and `\` and the character sequence `${` must be escaped by prefixing them with a backslash (`\`).

rycee:
Actually, in this case there is no need to escape since [`scrubDerivations`](https://gitlab.com/rycee/nmd/-/blob/91dee681dd1c478d6040a00835d73c0f4a4c5c29/default.nix#L7-30) will automatically put the correctly escaped version in the documentation.

rycee:
Rebased to master in 6427ae95789c04534363d2b68289996f08077c0c 🙂

PR #3301, opened 2022-10-04
qutebrowser: add greasemonkey userscript option

IvarWithoutBones:
### Description
This adds the `programs.qutebrowser.greasemonkey` option, a list of packages that represent greasemonkey userscripts.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/qutebrowser.nix
+++ b/modules/programs/qutebrowser.nix
@@ -268,6 +268,27 @@ in {
       '';
     };
 
+    greasemonkey = mkOption {
+      type = types.listOf types.package;
+      default = [ ];
+      description = ''
+        Greasemonkey userscripts to add to qutebrowser's <filename>greasemonkey</filename> directory.
+      '';
+      example = literalExpression ''
+        [
+          (pkgs.fetchurl {
+            url = "https://raw.githubusercontent.com/afreakk/greasemonkeyscripts/1d1be041a65c251692ee082eda64d2637edf6444/youtube_sponsorblock.js";
+            sha256 = "sha256-e3QgDPa3AOpPyzwvVjPQyEsSUC9goisjBUDMxLwg8ZE=";
+          })
+          (pkgs.writeText "some-script.js" '''
+            // ==UserScript==
+            // @name  Some Greasemonkey script
+            // ==/UserScript==
+          ''')
+        ]
+      '';
+    };
+
     extraConfig = mkOption {
       type = types.lines;
       default = "";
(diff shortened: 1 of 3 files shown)


somasis:
This looks mostly okay, but there's a few things I wonder about:

- Currently there's no handling of qutebrowser's `@require` downloading; this could cause errors on startup (and scripts to not load) if scripts in `~/.config/qutebrowser/greasemonkey` have any `@require` statements not satisfied by the script derivative itself, since the generated `greasemonkey` directory is read-only (because it's in the nix store) and thus qutebrowser can't download anything itself. There ought to be a way to do the `require` downloading declaratively (or at least, make the `requires` directory writable).
- Maybe `programs.qutebrowser.greasemonkey` should also be able to take paths and strings in addition to derivatives? Otherwise for scripts one is referring to within the configuration path, you have to do something like `pkgs.writeTextFile "script.js" ./script.js`. I'm not really sure if there's precedent for doing this in this option but it was just something that occurred to me while configuring this on my system.

somasis:
> This looks mostly okay, but there's a few things I wonder about:
> 
> * Currently there's no handling of qutebrowser's `@require` downloading; this could cause errors on startup (and scripts to not load) if scripts in `~/.config/qutebrowser/greasemonkey` have any `@require` statements not satisfied by the script derivative itself, since the generated `greasemonkey` directory is read-only (because it's in the nix store) and thus qutebrowser can't download anything itself. There ought to be a way to do the `require` downloading declaratively (or at least, make the `requires` directory writable).

Actually, I wasn't entirely correct; qutebrowser downloads the `@require`d scripts to its data directory (`~/.local/share/qutebrowser/greasemonkey/requires`), so it's probably not as big a problem I made it out to be.

somasis:
Anything still preventing this from being merged?

modules/programs/qutebrowser.nix
```diff
@@ -325,9 +344,24 @@ in {
         '';
       };
 
+    home.file.".qutebrowser/quickmarks" =
+      mkIf (cfg.quickmarks != { } && pkgs.stdenv.hostPlatform.isDarwin) {
+        text = quickmarksFile;
+      };
+
```
sumnerevans, line 351: why did this get moved? If for no reason, then please undo this change.

tests/modules/programs/qutebrowser/default.nix
```diff
@@ -2,4 +2,5 @@
   qutebrowser-settings = ./settings.nix;
   qutebrowser-keybindings = ./keybindings.nix;
   qutebrowser-quickmarks = ./quickmarks.nix;
+  qutebrowser-greasemonkey = ./greasemonkey.nix;
```
sumnerevans, line 5: these lines really should be sorted, but I guess it was already unsorted, so this is fine

sumnerevans (changes requested):

sisyphushappy:
Opened #4239 to finalize this PR.

rycee:
Resolved by merging the PR mentioned above.

PR #3303, opened 2022-10-05
mpd: add package to home path

ncfavier:
It's always useful to have the man pages available.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/mpd.nix
+++ b/modules/services/mpd.nix
@@ -198,6 +198,8 @@ in {
 
       Install = { WantedBy = [ "sockets.target" ]; };
     };
+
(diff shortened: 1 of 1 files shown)


modules/services/mpd.nix
```diff
@@ -198,6 +198,8 @@ in {
 
       Install = { WantedBy = [ "sockets.target" ]; };
     };
+
+    home.packages = [ cfg.package ];
```
rycee, line 202: May be possible to do a more targeted install using `getMan`. I.e. something like

```
home.packages = [ (getMan cfg.package) ];
```

PR #3304, opened 2022-10-06
contour: add module

ivan770:
### Description

Add [Contour terminal](https://github.com/contour-terminal/contour) module.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/contour.nix
+++ b/modules/programs/contour.nix
@@ -1,0 +1,1193 @@
+{ config, lib, pkgs, ... }:
+with lib;
+let
+  cfg = config.programs.contour;
+
+  filteredKeybindings =
+    map (keybind: filterAttrs (_: v: v != null) keybind) cfg.keybindings;
+in {
+  meta.maintainers = [ maintainers.ivan770 ];
+
+  options.programs.contour = {
+    enable = mkEnableOption "Contour terminal";
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.contour;
+      defaultText = literalExpression "pkgs.contour";
+      description = ''
+        Contour package to install
+      '';
+    };
+
+    settings = mkOption {
+      type = types.submodule {
+        options = {
+          platform_plugin = mkOption {
+            type = types.enum [ "auto" "xcb" "cocoa" ];
+            default = "auto";
+            description = ''
+              Overrides the auto-detected platform plugin to be loaded.
+            '';
+          };
+
+          renderer = mkOption {
+            type = types.submodule {
+              options = {
+                backend = mkOption {
+                  type = types.enum [ "default" "software" "OpenGL" ];
+                  default = "OpenGL";
+                  description = ''
+                    Backend to use for rendering the terminal onto the screen.
+                  '';
+                };
+
+                tile_hashtable_slots = mkOption {
+                  type = types.int;
+                  default = 4096;
+                  description = ''
+                    Number of hashtable slots to map to the texture tiles.
+                    Larger values may increase performance, but too large may also decrease.
+                    This value is rounted up to a value equal to the power of two.
+                  '';
+                };
+
+                tile_cache_count = mkOption {
+                  type = types.int;
+                  default = 4000;
+                  description = ''
+                    Number of tiles that must fit at lest into the texture atlas.
+
+                    This does not include direct mapped tiles (US-ASCII glyphs,
+                    cursor shapes and decorations), if tile_direct_mapping is set to true).
+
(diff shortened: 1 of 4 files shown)


modules/programs/contour.nix
```diff
@@ -0,0 +1,1193 @@
+{ config, lib, pkgs, ... }:
+with lib;
+let
+  cfg = config.programs.contour;
+
+  filteredKeybindings =
+    map (keybind: filterAttrs (_: v: v != null) keybind) cfg.keybindings;
+in {
+  meta.maintainers = [ maintainers.ivan770 ];
```
rycee, line 9:
```suggestion
  meta.maintainers = [ hm.maintainers.ivan770 ];
```

modules/programs/contour.nix
```diff
@@ -0,0 +1,1193 @@
...
+    package = mkOption {
+      type = types.package;
+      default = pkgs.contour;
+      defaultText = literalExpression "pkgs.contour";
+      description = ''
+        Contour package to install
+      '';
+    };
+
+    settings = mkOption {
```
rycee, line 23: You can replace this very long definition of `settings` with one using `formats.yaml` like in the Pueue module: https://github.com/nix-community/home-manager/blob/1a8e35d2e53ed2ccd9818fad9c9478d56c655661/modules/services/pueue.nix#L19-L33

modules/programs/contour.nix
```diff
@@ -0,0 +1,1193 @@
...
+      '';
+    };
+
+    extraConfig = mkOption {
+      type = types.attrs;
+      default = { };
+      description = ''
+        Extra configuration that overrides primary settings.
+      '';
+    };
```
rycee, line 1177: No need for this when the user can use an RFC-42 style `settings` option:

```suggestion
```

modules/programs/contour.nix
```diff
@@ -0,0 +1,1193 @@
...
+            ];
+            description = ''
+              Action to be executed on mapped input activation.
+            '';
+          };
+
+          mode = mkOption {
+            type = types.nullOr types.str;
+            default = null;
+            description = lib.mdDoc ''
```
rycee, line 971: Need to use docbook. HM doesn't use the documentation system of NixOS.

```suggestion
            description = ''
```

modules/programs/contour.nix
ivan770: I'm not sure if that's a preferred way, since, well, we lose all the type info

modules/programs/contour.nix
```diff
@@ -0,0 +1,1193 @@
...
+              * Select    : The terminal has currently an active grid cell selection (such as selected text).
+              * Insert    : The Insert input mode is active, that is the default and one way to test that the input mode is not in normal mode or any of the visual select modes.
+
+              You can combine these modes by concatenating them via | and negate a single one
+              by prefixing with ~.
+            '';
+          };
+        };
+      });
+      default = [
```
rycee, line 986: Can't the default simply be `[ ]`?

ivan770: That would result in Contour having no keybindings at all, which is rather unexpected and requires user to re-generate config manually to obtain default values

rycee: Ah, that's unfortunately. I imagine you will then have to do something like in the i3 module: https://github.com/nix-community/home-manager/blob/1a8e35d2e53ed2ccd9818fad9c9478d56c655661/modules/services/window-managers/i3-sway/i3.nix#L22-L111

Note, the example and description shows the user how they can extend the default keybindings.

modules/programs/contour.nix
rycee: Unfortunately, it is the preferred way to keep the option list maintainable. Exceptions can be made when the configuration options are very stable (such as for SSH or i3) but it seems to me that this may not be the case with Contour.

If you want your can make the settings option into a freeform module with explicit options specified for a small number of keys that most people would want to edit.

pastadudes:
hello! i want to supersede this pr (since i would really like this module)
for `.github/CODEOWNERS` do i keep `/modules/programs/contour.nix                         @ivan770` or do i add my github handle or replace it with mine?

khaneliman:
> hello! i want to supersede this pr (since i would really like this module) for `.github/CODEOWNERS` do i keep `/modules/programs/contour.nix @ivan770` or do i add my github handle or replace it with mine?

We don't really use the CODEOWNERS too much atm, if you want to maintain a module just add `maintainers` with your entry in it.

PR #3308, opened 2022-10-06
blueman-applet: fix service name in example

adamcstephens:
### Description

Simple documentation update

### Checklist

  ```
  {component}: {description}

  {long description}
  ```

### DIFF
--- a/modules/services/blueman-applet.nix
+++ b/modules/services/blueman-applet.nix
@@ -13,7 +13,7 @@ with lib;
           be enabled system-wide. You can enable it in the system
           configuration using
           <programlisting language="nix">
-            services.blueman.enable = true;
+            services.blueman-applet.enable = true;
           </programlisting>
         '';
       };


modules/services/blueman-applet.nix
```diff
@@ -13,7 +13,7 @@ with lib;
           be enabled system-wide. You can enable it in the system
           configuration using
           <programlisting language="nix">
-            services.blueman.enable = true;
+            services.blueman-applet.enable = true;
```
rycee, line 16: Are you sure this is correct? My understanding of the description is that for

    services.blueman-applet.enable = true;

to actually work for your HM configuration you need to have

    services.blueman.enable = true;

in your NixOS configuration. Searching for "blueman" among the NixOS configuration options shows no blueman-applet module: https://search.nixos.org/options?query=blueman

adamcstephens: No, I'm wrong. Sorry for the spam.

rycee: No worries.

PR #3319, opened 2022-10-11
dircolors: add the option `configPath`

gxmas:
### Description

Add the option `dircolors.configPath` so that the path of the dircolors file can be configurable.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/dircolors.nix
+++ b/modules/programs/dircolors.nix
@@ -67,6 +67,19 @@ in {
         Extra lines added to <filename>.dir_colors</filename> file.
       '';
     };
+
+    configPath = mkOption {
+      type = types.path;
+      default = "${config.home.homeDirectory}/.dir_colors";
+      defaultText = "~/.dir_colors";
+      apply = toString;
+      description = ''
+        Path to the <filename>.dir_colors</filename> file.
+      '';
+      example = literalExpression ''
+        "''${config.home.homeDirectory}/.dir_colors"
+      '';
+    };
   };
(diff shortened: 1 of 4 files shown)


modules/programs/dircolors.nix
```diff
@@ -67,6 +67,19 @@ in {
         Extra lines added to <filename>.dir_colors</filename> file.
       '';
     };
+
+    configPath = mkOption {
```
rycee, line 71: Could just remove this and unconditionally use `xdg.configFile`?

rycee: The dir_colors man page does say

```
~/.dir_colors
       Per-user configuration file.
```

so in principle it is incorrect to place it anywhere else. But, I don't know of anything that uses this file except the line calling dircolors, which is under HM control. So I guess it would be OK to place it elsewhere.

In fact, it wouldn't be necessary to place it anywhere in the home directory, could simply reference it directly in the Nix store. I.e. something like

``` bash
eval $(${pkgs.coreutils}/bin/dircolors -b ${configFile})
```

where

```
configFile = pkgs.writeText "dir-colors" (concatStringsSep "\n" ([ ]
      ++ mapAttrsToList formatLine cfg.settings ++ [ "" ]
      ++ optional (cfg.extraConfig != "") cfg.extraConfig))
```

rycee:
Maintainer CC: @JustinLovinger

PR #3242, opened 2022-09-20
WIP qutebrowser: add userscript management options

somasis:
### Description

This commit adds the ability to declaratively manage qutebrowser's greasemonkey scripts from within Home Manager. Currently, greasemonkey scripts have to be managed imperatively, or with liberal use of `xdg.dataFile."qutebrowser/greasemonkey/<name>"`.

This pull request is marked as a WIP because of a few outstanding issues:

1. qutebrowser, when it loads greasemonkey scripts, will read all `@require` keywords, and then download the scripts. This is a problem, because (aside from that being impure) this commit makes `~/.local/share/qutebrowser/greasemonkey` read-only, and so it'll fail to write the required files anyway. qutebrowser shouldn't have to download the required scripts.

2. Required scripts should be managed differently than they are here. `xdg.dataFile."qutebrowser/greasemonkey"` and `xdg.dataFile."qutebrowser/greasemonkey/requires"` clash with each other, and as a result, none of the required scripts are placed in `~/.local/share/qutebrowser/greasemonkey/requires`.

3. I'm not pleased with the way scripts dependencies are declared. Ideally, they'd be managed somewhat similarly to Kakoune's plugins. However, this entails more infrastructure be present for creating derivatives from greasemonkey scripts, which, it doesn't right now.

4. How required files are managed should be simplified. qutebrowser's `@require` downloaded is a bit nasty with the filenames is produces: `~/.local/share/qutebrowser/greasemonkey/requires/https:__code.jquery.com_jquery-3.3.1.slim.min.js`. I'm not sure if it demands that these correspond with the `@require` keyword, but I presume it does...

The structure of qutebrowser's greasemonkey directory looks like this:

```
.config/qutebrowser/greasemonkey/
  user_edited_script.user.js
.local/share/qutebrowser/greasemonkey/
  requires/
    slashes_changed_to_underscores.js
  qutebrowser_managed_script.js
```

The method of configuring it that I have in my head looks is like this:

```nix
# all derivatives in userScripts get merged into one directory, erroring
# when files conflict of course

# produces:
#   ./
#     requires/
#       jquery.min.js
#       lib_mbimport.js
#       lib_mbimportstyle.js
#       lib_mbimportlinks.js
#       lib_logger.js
#     bandcamp_importer.user.js
#     Zoom_redirector.user.js

programs.qutebrowser.userScripts = [
  # standalone script, has no `@require`'d scripts

  # produces:
  #   ./
  #     Zoom_redirector.user.js

[...]

PR #3322, opened 2022-10-11
qutebrowser: allow for specifying multiple commands in bindings

somasis:
### Description

This command adds the ability to specify lists of qutebrowser commands as values for key bindings,
which avoids the need for concatenating commands with ` ;; `.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/qutebrowser.nix
+++ b/modules/programs/qutebrowser.nix
@@ -131,7 +131,7 @@ in {
     };
 
     keyBindings = mkOption {
-      type = types.attrsOf (types.attrsOf types.str);
+      type = with types; attrsOf (attrsOf (separatedString " ;; "));
       default = { };
       description = ''
         Key bindings mapping keys to commands in different modes. This setting
@@ -245,6 +245,11 @@ in {
             "<Ctrl-v>" = "spawn mpv {url}";
             ",p" = "spawn --userscript qute-pass";
             ",l" = '''config-cycle spellcheck.languages ["en-GB"] ["en-US"]''';
+            "<F1>" = mkMerge [
+              "config-cycle tabs.show never always"
+              "config-cycle statusbar.show in-mode always"
+              "config-cycle scrolling.bar never always"
+            ];
           };
           prompt = {
             "<Ctrl-y>" = "prompt-yes";
--- a/tests/modules/programs/qutebrowser/keybindings.nix
+++ b/tests/modules/programs/qutebrowser/keybindings.nix
@@ -13,6 +13,11 @@ with lib;
         normal = {
           "<Ctrl-v>" = "spawn mpv {url}";
           ",l" = ''config-cycle spellcheck.languages ["en-GB"] ["en-US"]'';
+          "<F1>" = mkMerge [
+            "config-cycle tabs.show never always"
+            "config-cycle statusbar.show in-mode always"
+            "config-cycle scrolling.bar never always"
+          ];
         };
         prompt = { "<Ctrl-y>" = "prompt-yes"; };
       };
@@ -34,6 +39,7 @@ with lib;
             c.bindings.default = {}
             config.bind(",l", "config-cycle spellcheck.languages [\"en-GB\"] [\"en-US\"]", mode="normal")
             config.bind("<Ctrl-v>", "spawn mpv {url}", mode="normal")
+            config.bind("<F1>", "config-cycle tabs.show never always ;; config-cycle statusbar.show in-mode always ;; config-cycle scrolling.bar never always", mode="normal")
             config.bind("<Ctrl-y>", "prompt-yes", mode="prompt")''
         }
     '';


modules/programs/qutebrowser.nix
```diff
@@ -131,7 +137,7 @@ in {
     };
 
     keyBindings = mkOption {
-      type = types.attrsOf (types.attrsOf types.str);
+      type = with types; attrsOf (attrsOf (either (listOf str) str));
```
h7x4, line 140:
```suggestion
      type = let coerce = concatStringsSep " ;; ";
      in with types; attrsOf (attrsOf (coercedTo (listOf str) coerce str));
```

modules/programs/qutebrowser.nix
```diff
@@ -29,7 +29,13 @@ let
...
     let
       formatKeyBinding = m: k: c:
-        ''config.bind("${k}", "${escape [ ''"'' ] c}", mode="${m}")'';
+        ''
+          config.bind("${k}", "${
+            if builtins.isList c then
+              escape [ ''"'' ] (concatStringsSep " ;; " c)
+            else
+              escape [ ''"'' ] c
+          }", mode="${m}")'';
```
h7x4, line 38:
```suggestion
        ''config.bind("${k}", "${escape [ ''"'' ] c}", mode="${m}")'';
```

h7x4 (changes requested):
Could you update `/tests/modules/programs/qutebrowser/keybindings.nix` with your `<F1>` example as well?

somasis:
Updated the test and applied your suggestions.

modules/programs/qutebrowser.nix
h7x4: I don't think you will need this block anymore. `c` is already coerced to a string, so `builtins.isList c` should never occur.

somasis: You're right :) I accidentally the commit that I pushed and it was only half changed.

h7x4:
Nice :+1: . I think the only thing remaining is removing the WIP commit (maybe you could move it to a new PR?), and then this is ready for merge as far as I can see.

somasis:
> I think the only thing remaining is removing the WIP commit

whoops :)

It's from an old PR I scrapped (#3242)

h7x4 (approved):
LGTM :+1:

modules/programs/qutebrowser.nix
```diff
@@ -131,7 +131,8 @@ in {
     };
 
     keyBindings = mkOption {
-      type = types.attrsOf (types.attrsOf types.str);
+      type = let coerce = concatStringsSep " ;; ";
+      in with types; attrsOf (attrsOf (coercedTo (listOf str) coerce str));
```
ncfavier, line 135:
```suggestion
      type = with types; attrsOf (attrsOf (separatedString " ;; "));
```

modules/programs/qutebrowser.nix
```diff
@@ -245,6 +246,11 @@ in {
             "<Ctrl-v>" = "spawn mpv {url}";
             ",p" = "spawn --userscript qute-pass";
             ",l" = '''config-cycle spellcheck.languages ["en-GB"] ["en-US"]''';
+            "<F1>" = [
```
ncfavier, line 249:
```suggestion
            "<F1>" = mkMerge [
```

tests/modules/programs/qutebrowser/keybindings.nix
```diff
@@ -13,6 +13,11 @@ with lib;
         normal = {
           "<Ctrl-v>" = "spawn mpv {url}";
           ",l" = ''config-cycle spellcheck.languages ["en-GB"] ["en-US"]'';
+          "<F1>" = [
```
ncfavier, line 16:
```suggestion
          "<F1>" = mkMerge [
```

modules/programs/qutebrowser.nix
somasis: Is there a reason to prefer one way over the other (readability or otherwise)? The first version of the type was suggested just before your review. https://github.com/nix-community/home-manager/pull/3322#discussion_r1028533628

ncfavier: It's a trade-off between simplicity of types and brevity of definitions. I am partial to the former, but it is only a suggestion.

somasis: I'll just go with your suggestion, it looks nicer to me as well.

modules/programs/qutebrowser.nix
```diff
@@ -131,7 +131,7 @@ in {
     };
 
     keyBindings = mkOption {
-      type = types.attrsOf (types.attrsOf types.str);
+      type = with types; attrsOf (attrsOf (separatedString " ;; "));
```
berbiche, line 134: If we make it even more complex with `types.coercedTo`, it becomes possible to use raw lists (i.e. without `mkMerge`) 😃 

```nix
attrsOf (attrsOf (coercedTo (listOf str) (concatStringsSep " ;; ") (separatedString " ;; ")))
```

`separatedString` remains needed to merge definitions like
```nix
# a.nix
keyBindings."a" = "foo";

# b.nix
keyBindings."a" = [ "bar" "baz" ];
```
The final order of commands depends on the evaluation order 🤷

berbiche (approved):
LGTM!

You can ignore my comment if you want. No changes required.

modules/programs/qutebrowser.nix
ncfavier: That's where we're coming from 😅 https://github.com/nix-community/home-manager/pull/3322#discussion_r1072772539

berbiche: Oops, as you can see I did not check your previous comments

somasis:
Anything still holding this back from being merged?

somasis:
thanks! :)

PR #3332, opened 2022-10-16
himalaya: fix broken config to support 0.6.x

toastal:
### Description

Himalaya had breaking config changes from 0.5.x to 0.6.x. @ambroisie mentioned wanting to step down from maintaining the module.

This isn't backwards compatible because the new config isn't. It now explicitly requires `backend` and `sender` are set to an enum which requires certain properties. I suppose `backend = "imap"` and `sender = "smtp"` could be `default` values to follow what was previously set as the only and required options, but I believe it would be better to be explicit moving forwards--just like `himalaya`.

This fix allowed me to verify that the Vim plugin still worked too https://github.com/NixOS/nixpkgs/pull/195927/files

### Checklist

- [ ] Code tested through `nix-shell --pure tests -A run.all`. (failed with `neovim`)
- [x] Code tested through `nix-shell --pure tests -A run.himalaya`.

    ```
    {component}: {description}

    {long description}
    ```

### Comments

Please offer a non-proprietary option for contribution instead of being limited to the Microsoft-owned GitHub forge. Identity is tied to a Microsoft GitHub account too which is as barrier to contribution and requires giving away ones data.

I don't like the use of `nixfmt` because it makes changes that break `git` diffs. I prefer `nixpkgs-fmt` as it has a philosophy of "expand, never contract" and so lines never get pulled up and cuddled if the string or set is now shorter making it harder to review.

### DIFF
--- a/modules/programs/himalaya.nix
+++ b/modules/programs/himalaya.nix
@@ -11,7 +11,7 @@ let
     toHimalayaConfig = account:
       {
         email = account.address;
-        name = account.realName;
+        display-name = account.realName;
         default = account.primary;
 
         mailboxes = {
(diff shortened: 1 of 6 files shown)


modules/programs/himalaya.nix
```diff
@@ -20,30 +20,49 @@ let
...
         smtp-passwd-cmd = lib.escapeShellArgs account.passwordCommand;
         smtp-host = account.smtp.host;
         smtp-port = account.smtp.port;
         smtp-starttls = account.smtp.tls.useStartTls;
-      } // (lib.optionalAttrs (account.signature.showSignature == "append") {
-        # FIXME: signature cannot be attached
-        signature = account.signature.text;
-      }) // account.himalaya.settings;
+      } else if account.himalaya.sender == "sendmail" then
+        if account.himalaya ? sendmail-cmd then {
```
ambroisie, line 53: You only check the settings existence but don't add it to the configuration?

toastal: @ambroisie  I'm not entirely sure how to do this in Nix syntax... it's like an ADT where some setting are dependent on both the `sender` and `backend` properties.

```
type Sender =
  | IMAP { imap-port :: Number, ...etc }
  | Sendmail { sendmail-cmd :: String }
```

toastal: In this case, the `sendmail-cmd` is not needed or used if the `sender` is `imap`.

ambroisie: Yes I understood that part, my confusion was just that `sendmail-cmd` seemed to not be set, but that's only because I forgot that `// account.himalaya.settings` is done at the end.

ambroisie: Now that I am looking at this again, I'm thinking that perhaps either the `throw` should be removed (and let `himalaya` itself complain about the missing command) or perhaps it should be an actual assertion, using the module system.

Will let the maintainers weight in about what's usually done in home-manager.

toastal: I'm okay with any of those options.

toastal: I did the latter... removed the `-cmd` check

ambroisie (approved):

rycee:
Thanks! Is this all good for merge?

About contributions outside GitHub you can try sending patches to my public inbox on sr.ht: https://lists.sr.ht/~rycee/public-inbox For now that is an unofficial channel, though.

toastal:
@rycee I've been using it for a while now. There seems to be a bug inside the Vim plugin for me, but that's not related to getting the config up and running so I could launch `himalaya` to see even see that error.

NomisIV:
I hate to be pushy, but I'm waiting for this to be merged as I'm using himalaya as my main email client. Are there any blockers for merging?

toastal:
I've been trying to do the occasional rebase so folks could use in the interim. It's worked fine for me for couple of months now.
-- 
toastal ไข่ดาว | https://toast.al
PGP: 7944 74b7 d236 dab9 c9ef  e7f9 5cce 6f14 66d4 7c9e

teto:
@rycee do we follow some convention to signal breaking changes ? in "conventionalcommits", you can detect breaking changes by looking for BREAKING CHANGE in the last line of the commit. The first line contains also a `!`. Maybe we could adopt some scheme like that ? that's keeping me from merging. We need some kind of log for breaking changes between releases like nixos does

rycee:
Thanks! Merged to master now.

Issue #2803, opened 2022-03-18
Add librewolf support to the firefox config module


TilCreator:
I want to use Librewolf instead of Firefox, mostly because Mozilla is doing more and more "evil" things.
I tried to just set `programs.firefox.package` to `pkgs.librewolf`(and create new profiles). But this fails with something in the firefox wrapper:
```
error: attribute 'gtk3' missing

       at /nix/store/21d58mybzbw2h1piiyr95yfh1k2s3ajh-nixos/nixos/pkgs/applications/networking/browsers/firefox/wrapper.nix:179:23:

          178|       nativeBuildInputs = [ makeWrapper lndir replace jq ];
          179|       buildInputs = [ browser.gtk3 ];
             |                       ^
          180|
```
I don't know exactly how the Firefox and Librewolf package differ, but I would guess that just a further check in https://github.com/nix-community/home-manager/blob/master/modules/programs/firefox.nix#L337-L354 for Librewolf would be enough to make this work?

SuperSandro2000:
librewolf is missing this part https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/networking/browsers/firefox/common.nix#L363-L374

TilCreator:
I don't quite get it, isn't Librewolf build with that Nix file?

TilCreator:
I think it would also be better to have an extra module for Librewolf, like with I3 and sway. So that one can have Librewolf profiles in ~/.librewolf and Firefox profiles in ~/.mozilla/firefox. What do you think?

SuperSandro2000:
> I don't quite get it, isn't Librewolf build with that Nix file?

No, it uses this one https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/networking/browsers/firefox/librewolf/default.nix

> What do you think?

I don't know, I am not using it.

TilCreator:
> No, it uses this one https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/networking/browsers/firefox/librewolf/default.nix

But that file is only used as source in https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/networking/browsers/firefox/packages.nix#L62 that is then used in https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/networking/browsers/firefox/common.nix
Please correct me if I'm wrong, I'm sure you have more experience and there has to be a reason why it doesn't work after all.

> I don't know, I am not using it.

[...]

PR #3339, opened 2022-10-17
firefox: use different paths if librewolf is used

TilCreator:
### Description
~~Makes the `mozillaConfigPath`, `firefoxConfigPath` and `profilesPath` available to the user
https://github.com/nix-community/home-manager/issues/2803#issuecomment-1276635894~~

Adds handling for the firefox module to use different profile paths if the package is changed to librewolf.  
This is kind of a solution for #2803 and #4179 

I'm currently not shure how good this solution is and will leave this as a draft and without tests until I used this for a bit and wrote tests.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/firefox.nix
+++ b/modules/programs/firefox.nix
@@ -10,13 +10,21 @@ let
 
   jsonFormat = pkgs.formats.json { };
 
-  mozillaConfigPath =
-    if isDarwin then "Library/Application Support/Mozilla" else ".mozilla";
+  isLibrewolf = cfg.package.meta.mainProgram == "librewolf";
 
-  firefoxConfigPath = if isDarwin then
-    "Library/Application Support/Firefox"
+  mozillaConfigPath =
+    if isLibrewolf then
+      (if isDarwin then "Library/Application Support/Librewolf" else ".librewolf")
+    else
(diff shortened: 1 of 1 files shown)


rycee:
Seems like the module could, e.g., look at the browser `binaryName` to automatically determine these paths for librewolf. I think it would be inappropriate to force such details on the user.

montchr:
It would be helpful to have these values available. Currently I'm copying the logic into my own config file. 

Seeing as the options shouldn't necessarily be configurable since Firefox expects exactly those paths, perhaps these would be more suitable as read-only options? Or alternatively, accessible by way of a function defined in a separate file so that users could import it directly.

modules/programs/firefox.nix
```diff
@@ -10,13 +10,15 @@ let
 
   jsonFormat = pkgs.formats.json { };
 
-  mozillaConfigPath =
```
rycee, line 13: The `mozillaConfigPath` variable is now used for the `native-messaging-hosts` directory.

Looking at https://librewolf.net/docs/faq/#how-do-i-get-native-messaging-to-work it seems `~/.librewolf/native-messaging-hosts` is the path for LibreWolf. So perhaps could have `mozillaConfigPath = "~/.librewolf"` when `isLibrewolf == true`?

PR #3345, opened 2022-10-22
hypr: add module

AndersonTorres:
Now that @AndersonTorres added Hypr X11 window manager to Nixpkgs, it is a good time to add it to Home Manager too!

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -527,6 +527,8 @@ Makefile                                              @thiagokokada
 /modules/services/window-managers/herbstluftwm        @olmokramer
 /tests/modules/services/window-managers/herbstluftwm  @olmokramer
 
+/modules/services/window-managers/hypr                @AndersonTorres
+
 /modules/services/window-managers/i3-sway/i3.nix      @sumnerevans @sebtm
 /tests/modules/services/window-managers/i3            @sumnerevans @sebtm
 
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -292,6 +292,7 @@ let
     ./services/window-managers/bspwm/default.nix
     ./services/window-managers/fluxbox.nix
     ./services/window-managers/herbstluftwm.nix
+    ./services/window-managers/hypr/default.nix
     ./services/window-managers/i3-sway/i3.nix
     ./services/window-managers/i3-sway/sway.nix
     ./services/window-managers/i3-sway/swaynag.nix
--- a/modules/services/window-managers/hypr/default.nix
+++ b/modules/services/window-managers/hypr/default.nix
@@ -1,0 +1,29 @@
+{ config, lib, pkgs, ... }:
+
+let
+
+  inherit (lib) hm types mkEnableOption mkIf mkOption mkPackageOption platforms;
+
+  cfg = config.xsession.windowManager.hypr;
+
+in {
+  meta.maintainers = with lib.maintainers; [ AndersonTorres ];
+
+  options = import ./options.nix { inherit config lib pkgs; };
+
+  config = mkIf cfg.enable {
+    assertions = [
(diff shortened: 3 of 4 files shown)


modules/services/window-managers/hypr.nix
```diff
@@ -0,0 +1,46 @@
+{ config, lib, pkgs, ... }:
+
+let
+
+  inherit (lib) hm types mkEnableOption mkIf mkOption mkPackageOption platforms;
```
AndersonTorres, line 5: Hello, @rycee !

Can I use this more explicit listing of auxiliary `lib` functions?

AndersonTorres:
@rycee ping!

AndersonTorres:
@rycee ping

AndersonTorres:
@rycee another ping

modules/services/window-managers/hypr.nix
```diff
@@ -0,0 +1,46 @@
...
+
+      extraConfig = mkOption {
+        type = types.lines;
+        default = "";
+        description = ''
+          Configuration written to
+          <filename>$XDG_CONFIG_HOME/hypr/hypr.conf</filename>. See
+          <link xlink:href="https://github.com/hyprwm/Hypr/wiki/Configuring-Hypr" />
+          for explanation about possible values.
+        '';
```
thiagokokada, line 26: The config file format here should be easy enough to implement using structured config with attrset instead of `extraConfig` style.

AndersonTorres: I believe it would need at least some form of parsing. 

The official example creates at least two levels of options:

```
# top level
gaps_in=5
border_size=1
gaps_out=20
rounding=0
max_fps=60

# sub level
Bar {
    height=20
    monitor=0
    enabled=1
    mod_pad_in=8
    no_tray_saving=1 # using this doesnt save the tray between reloads but fixes an issue with the bar disappearing.
}
```

And, well, parsing

thiagokokada: This still looks like a highly structured format that can be easily represented by an attrset. See `rofi` module for an example of a way more complex configuration format that is also using structured config instead of `extraConfig`.

> And, well, parsing is something prone to errors...

Well, if you want to argue that way, I would argue that it is way easier to commit config mistakes in `extraConfig`. At least with structured config, as long it compiles you can be pretty certain that there is no formatting mistakes.

thiagokokada: Remember that you can (and should) write tests for modules and the tests can easily show that your parser is correct. You can't easily write tests for `extraConfig` since anything goes though.

AndersonTorres: > it is way easier to commit config mistakes in extraConfig.

It merely delegates the error to the final program, namely Hypr.

> At least with structured config, as long it compiles 

As long as the parser (written in Nix) is not bugged.

> See rofi module for an example

Fine. I will do that.

thiagokokada: > It merely delegates the error to the final program, namely Hypr.

Hypr is a window manager right? Delegating the errors to the final program is really bad in this case, since this may mean a broken session. If the user is using a Display Manager or something, it is even worse since you generally can't see the program output.

> As long as the parser (written in Nix) is not bugged.

This is why tests are recommended to be written.

thiagokokada: There is also the fact that structured config is just superior inside the module system. You can easily overwrite a config in a specific host for example using structured config, while `extraConfig` is addictive so you need a few hacks to make it work.

AndersonTorres:
@rycee is everything OK?

AndersonTorres:
A month without response. I will close this.

PR #3347, opened 2022-10-23
Support starship nushell integration

### LINKED ISSUE #3058: programs.starship.enableNushellIntegration
enable starship in nushell.

Philipp-M:
### Description

This PR adds the two options `extraConfig` and `extraEnv` to nushell. These value of these options are appended/merged with `configFile` and `envFile`. This is to ensure backwards compatibility.

Motivation for this is the starship nushell integration, which requires to extend either env and config files.
Unfortunately nushell is limited in sourcing or conditionally setting global environment variables, which is why the terminal compatibility check isn't done compared to the other starship shell integrations.

Fixes #3058

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/nushell.nix
+++ b/modules/programs/nushell.nix
@@ -6,14 +6,17 @@ let
 
   cfg = config.programs.nushell;
 
-  tomlFormat = pkgs.formats.toml { };
-
   linesOrSource = name:
     types.submodule ({ config, ... }: {
       options = {
         text = mkOption {
           type = types.lines;
-          default = "";
+          default = if config.source != null then
+            builtins.readFile config.source
+          else
+            "";
+          defaultText = literalExpression
+            "if source is defined, the content of source, otherwise empty";
           description = ''
             Text of the nushell <filename>${name}</filename> file.
             If unset then the source option will be preferred.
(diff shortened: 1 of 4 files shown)


rycee:
I think it may be worth simply removing the option of having a source path and only allowing a `types.lines` option. Having the file source makes things a bit tricky when other modules want to integrate with the nushell configuration. A user that wants to use a separate file can use `readFile` like you do implicitly.

It would break backwards compatibility but in this case it may be worth it.

Philipp-M:
Yeah I've thought about this as well, and I agree, this would definitely simplify things.

Do you have a suggestion for the name of the option(s), should I stay with with `configFile` or `configText` or go with `initExtra` like other shells (with the later possibility to abstract nushell code with nix, if it makes sense anyhow) or just use `extraConfig` or something else?
(Same for `envFile` etc.)

TheOPtimal:
> Yeah I've thought about this as well, and I agree, this would definitely simplify things.
> 
> Do you have a suggestion for the name of the option(s), should I stay with with `configFile` or `configText` or go with `initExtra` like other shells (with the later possibility to abstract nushell code with nix, if it makes sense anyhow) or just use `extraConfig` or something else? (Same for `envFile` etc.)

Since nuShell has a seperate env.nu and config.nu, initExtra or extraConfig don't make much sense. What about *just* `env` and `config` as the option names?

teto (approved):

Philipp-M:
Sorry for the late update...

> What about _just_ `env` and `config` as the option names?

Well the reasons are somewhat elucidated here: https://github.com/nix-community/home-manager/pull/2957#discussion_r877605437

But I'm not sure, I also think `config` and `env` may be fine but I certainly agree with the argument there, I think my closest preference would be `configText` otherwise, I'll let @rycee decide.

rycee:
Thanks! Merged to master now.

The `extraConfig` and `extraEnv` is fine to be consistent with other modules.

Issue #2519, opened 2021-11-29
starship settings require ordered attrsets


arcnmx:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Starship uses an [IndexMap](https://github.com/starship/starship/pull/1782) for its module settings that use dictionaries/attrsets so that they can be processed in declaration order. This doesn't translate well to nix, which doesn't really have ordered attrsets - and so certain settings are impossible to set properly via `programs.starship.settings`.

### System information

```markdown
host os: `Linux 5.15.4, NixOS, 22.05 (Quokka)`
version: `nix-env (Nix) 2.3.16`
```

berbiche:
Hi,

Do you have an example setting that cannot be set currently with Home Manager?

arcnmx:
Specifically - because starship requires that values be set in the order they should be processed - you cannot set...

```toml
[directory.substitutions]
"/a/b" = "b"
"/a" = "a"
```

because in nix, sets are always sorted by keys, so...

```nix
{
  directory.substitutions = {
    "/a/b" = "b";
    "/a" = "b";
  };
}
```

... results in an incorrect configuration:

```toml
[directory.substitutions]
"/a" = "a"
"/a/b" = "b"
```

rycee:
Pretty unfortunate 🙁 May be worth adding an `extraConfig` of `type.lines` to cope with such situations.

rycee:
I'm thinking that perhaps something like

``` nix
{
  directory.substitutions = literalExpression ''
    "/a/b" = "b"
    "/a" = "a"
  '';
}
```

may be interesting?

arcnmx:
eh a DAG is what I'd call interesting, that's a more boring approach :stuck_out_tongue: 
but sure, it could solve this!

(implementation might be a little complicated, plus unlike JSON there isn't necessarily such thing as a ubiquitous TOML literal value since the syntax is somewhat context-sensitive, you might have to disambiguate by saying `literalTableKeyValues` or something, but it can be hacked together regardless)

rcerc:
[...]

PR #3353, opened 2022-10-25
yambar: add module

carpinchomug:
### Description

This adds a module for [Yamber](https://codeberg.org/dnkl/yambar), a modular status panel for X11 and Wayland.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/yambar.nix
+++ b/modules/programs/yambar.nix
@@ -1,0 +1,55 @@
+{ config, lib, pkgs, ... }:
+
+let
+
+  cfg = config.programs.yambar;
+  yamlFormat = pkgs.formats.yaml { };
+
+in {
+  meta.maintainers = [ lib.maintainers.carpinchomug ];
+
+  options.programs.yambar = {
+    enable = lib.mkEnableOption "Yambar";
+
+    package = lib.mkPackageOption pkgs "yambar" { };
+
+    settings = lib.mkOption {
+      type = yamlFormat.type;
+      default = { };
+      example = lib.literalExpression ''
+        bar = {
+          location = "top";
+          height = 26;
+          background = "00000066";
+
+          right = [
+            {
+              clock.content = [
+                {
+                  string.text = "{time}";
+                }
+              ];
+            }
+          ];
+        };
+      '';
+      description = ''
+        Configuration written to {file}`$XDG_CONFIG_HOME/yambar/config.yml`.
+        See {manpage}`yambar(5)` for options.
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    assertions = [
+      (lib.hm.assertions.assertPlatform "programs.yambar" pkgs
+        lib.platforms.linux)
+    ];
+
+    home.packages = [ cfg.package ];
+
+    xdg.configFile."yambar/config.yml" = lib.mkIf (cfg.settings != { }) {
+      source = yamlFormat.generate "config.yml" cfg.settings;
+    };
+  };
+}
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -1642,6 +1642,20 @@ in {
           https://github.com/hyprwm/hyprpaper for more.
         '';
       }
+
+      {
+        time = "2024-05-10T21:28:38+00:00";
+        condition = hostPlatform.isLinux;
+        message = ''
+          A new module is available: 'programs.yambar'.
+
+          Yambar is a lightweight and configurable status panel for X11 and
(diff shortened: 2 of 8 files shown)


modules/programs/yambar.nix
```diff
@@ -0,0 +1,67 @@
...
+  options = {
+    programs.yambar = {
+      enable = mkEnableOption "Yambar";
+
+      package = mkOption {
+        type = types.package;
+        default = pkgs.yambar;
+        defaultText = literalExpression "pkgs.yambar";
+        description = "The Yambar package to install";
+      };
```
h7x4, line 19:
```suggestion
      package = mkPackageOption pkgs "yambar" { };
```

h7x4 (changes requested):

m-bdf:
When using the [map particle](https://www.mankier.com/5/yambar-particles#Map), yambar evaluates the conditions in the order they appear in the config file.
However, the YAML generator always outputs Nix attribute sets alphabetically, so using a map in this modules's `settings` option won't work as expected (same as #2519, potential future solution in #3481).

For example, the Nix value
```nix
{
  river.content = {
    map.conditions = {
      "focused" = …;
      "occupied" = …;
      "id < 10" = …;
    };
  };
}
```
will be rendered to YAML as
```yaml
river:
  content:
    map:
      conditions:
        focused: …
        id < 10: …
        occupied: …
```
(notice the wrong order).

---

FWIW, here is a workaround for this example, using `~(…)` to negate each condition:
```nix
{
  river.content = {
    map.conditions = {
      "focused" = …;
      "~(focused)" = {
        map.conditions = {
          "occupied" = …;
          "~(occupied)" = {
            map.conditions = {
              "id < 10" = …;"
            };
          };
        };
      };
    };
  };
}
```
rendered as
```yaml
river:
  content:
    map:
      conditions:
        focused: …
        ~(focused):
          map:
            conditions:
              occupied: …
              ~(occupied):
                map:
                  conditions:
                    id < 10: …
```
(which is verbose and ugly, but will be evaluated as intended).

Also, here is a Nix helper function to generate the boilerplate from an ordered list of condition/particle pairs:
```nix
mkOrderedMapParticle =
  foldl' (res: { condition, particle }: {
    map.conditions = {
      "${condition}" = particle;
    } // optionalAttrs (res != null) {
      "~(${condition})" = res;
    };
  });
```
used like this:
```nix
{
  river.content = mkOrderedMapParticle null [
    { condition = "focused"; particle = …; }
    { condition = "occupied"; particle = …; }
    { condition = "id < 10"; particle = …; }
  ];
}
```
(the first argument being an optional default particle).

madsrumlenordstrom:
> ```nix
> {
>   river.content = mkOrderedMapParticle null [
>     { condition = "focused"; particle = …; }
>     { condition = "occupied"; particle = …; }
>     { condition = "id < 10"; particle = …; }
>   ];
> }
> ```
I would be satisfied using this solution.

ivan770:
Not sure if it's a proper scope for this PR, but would it be reasonable to add a systemd-based activation support for yambar?

madsrumlenordstrom:
> Not sure if it's a proper scope for this PR, but would it be reasonable to add a systemd-based activation support for yambar?

I think it would be a good idea to add systemd activation. I actually use this in my personal home-manager configuration. You can check it out [here](https://github.com/madsrumlenordstrom/nixos-config/blob/dd981dcb9e55c5e8b6e2bccff0485304573314b9/modules/home-manager/yambar.nix). I am no expert in systemd services so maybe there is a better way to do it.

modules/programs/yambar.nix
```diff
@@ -0,0 +1,62 @@
...
+        '';
+        description = ''
+          Configuration written to
+          <filename>$XDG_CONFIG_HOME/yambar/config.yml</filename>.
+          See
+          <citerefentry>
+           <refentrytitle>yambar</refentrytitle>
+           <manvolnum>5</manvolnum>
+          </citerefentry>
+          for options.
```
rycee, line 44: Can switch to Markdown

```suggestion
          Configuration written to
          {file}`$XDG_CONFIG_HOME/yambar/config.yml`.
          See {manpage}`yambar(5)` for options.
```

rycee:
Thanks and apologies for the delay. I've merged this to master now.

I agree that this module should define a systemd service. Well worth a future PR for any interested person.

About the particles, it does seem like an odd way to use YAML and I think the proposed workaround doesn't look too bad. So probably also worth a future PR.

PR #3360, opened 2022-10-29
vscode: fix examples and add a test

natsukium:
### Description

I fixed invalid examples of userSettings and userTasks and added userTasks' test.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/vscode.nix
+++ b/modules/programs/vscode.nix
@@ -82,7 +82,7 @@ in {
         default = { };
         example = literalExpression ''
           {
-            "files.autoSave" = "off"
+            "files.autoSave" = "off";
             "[nix]"."editor.tabSize" = 2;
           }
         '';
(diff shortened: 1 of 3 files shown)


modules/programs/vscode.nix
```diff
@@ -97,14 +97,14 @@ in {
...
               {
-                "type": "shell",
-                "label": "Hello task",
-                "command": "hello",
+                "type" = "shell";
+                "label" = "Hello task";
+                "command" = "hello";
               }
-            ]
+            ];
```
marsam, line 107: What do you think of using the following?
```nix
{
  version = "2.0.0";
  tasks = [
    {
      type = "shell";
      label = "Hello task";
      command = "hello";
    }
  ];
}
```

tests/modules/programs/vscode/tasks.nix
```diff
@@ -0,0 +1,43 @@
...
+    "${config.xdg.configHome}/Code/User/tasks.json";
+
+  tasks = {
+    "version" = "2.0.0";
+    "tasks" = [{
+      "type" = "shell";
+      "label" = "Hello task";
+      "command" = "hello";
+    }];
+  };
```
marsam, line 17: ditto

modules/programs/vscode.nix
natsukium: No problem, but I followed the style in settings.json above.
The tasks.json does not have a flattened key like "files.autosave", so there is less confusion without the quotes.

rycee:
Applied @marsam's suggestions and merged to master. Thanks for the contribution!

PR #3362, opened 2022-10-29
irssi: add option for SASL external authentication

deejayem:
### Description

Add option `programs.irssi.networks.<name>.saslExternal`. When enabled, this sets `sasl_username` for the network to the same value as `programs.irssi.networks.<name>.nick`, and sets `sasl_mechanism` to `EXTERNAL` (this means irssi will authenticate using SASL + certFP, the latter being configured with `programs.irssi.networks.<name>.server.ssl.certificateFile`).

An alternative would be to add three separate options for setting `sasl_username`, `sasl_password` and `sasl_mechanism` (which can be `PLAIN` to use `sasl_password`, or` EXTERNAL` to use certFP), but the approach I have taken avoids duplicating the nick/username, and prevents people from putting passwords in their config.

SASL + certFP is quite popular now, and LiberaChat restricts connecting without using SASL (i.e. depending on your IP, you can only connect this way), so I think this would be a very useful addition, instead of needing to add an entry for each network to extraConfig.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/irssi.nix
+++ b/modules/programs/irssi.nix
@@ -23,6 +23,11 @@ let
         type = "${v.type}";
         nick = "${quoteStr v.nick}";
         autosendcmd = "${concatMapStringsSep ";" quoteStr v.autoCommands}";
+        ${
+          lib.optionalString (v.saslExternal) ''
+            sasl_username = "${quoteStr v.nick}";
+              sasl_mechanism = "EXTERNAL";''
+        }
       };
     ''));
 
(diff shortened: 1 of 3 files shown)


modules/programs/irssi.nix
```diff
@@ -23,6 +23,12 @@ let
         type = "${v.type}";
         nick = "${quoteStr v.nick}";
         autosendcmd = "${concatMapStringsSep ";" quoteStr v.autoCommands}";
+        ${
+          lib.optionalString (v.saslExternal) ''
+            sasl_username = "${quoteStr v.nick}";
+              sasl_mechanism = "EXTERNAL";
```
deejayem, line 29: I had to add two extra spaces on line 29 to get consistent indentation in the config produced in the test. I'm not sure if that's the correct thing to do though.

PR #3366, opened 2022-10-30
i3blocks: add module

VAWVAW:
### Description

Adds a module for [i3blocks](https://github.com/vivien/i3blocks) a modular status command for bars.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


tests/modules/programs/i3blocks/with-custom.nix
```diff
@@ -0,0 +1,47 @@
...
+            a=foo
+            zz=bar
+
+            [time]
+            command=date '+%d.%m.%4Y %T'
+            interval=2
+            value1=2.500000
+
+            [asdf]
+            command=fdsa
```
h7x4, line 43:
```suggestion
            a = "foo"
            zz = "bar"

            [asdf]
            command = "fdsa"

            [time]
            command = "date '+%d.%m.%4Y %T'"
            interval = 2
            value1 = 2.5
```

I have tested that i3blocks accepts this file, even with quotation marks and spaces interlaced around `=`

h7x4 (changes requested):
Looks good! But I think it would be better to use the builtin TOML functionality to generate the config file.

VAWVAW:
I tried using TOML at first too. But the order of the blocks in the output of i3blocks depends on the order of the blocks in the config file. Using your change the block `asdf` would be displayed on the left of the block `time` in the test and the user has no option to change the order of the blocks except for prefixing the name of the block to prevent alphabetic ordering.

h7x4:
Ah, I see! That makes sense. But I think we should still make use of the builtin TOML generator for ease of maintenance, and in case `i3-blocks` suddenly chooses to support values other than `str` and `number`.

How about something like this?

```nix
    blocks = mkOption {
      type = types.listOf tomlFormat.type;
      default = [ ];
      description = ''
        Blocks to add to i3blocks <filename>config</filename> file. See
        <link xlink:href="https://github.com/vivien/i3blocks#i3blocks-properties" />
        for options.
      '';
        [
          {
            name = "time";
            command = "date '+%d.%m.%4Y %T'";
            interval = 5;
          };
        ]
      '';
    };

    ...

    xdg.configFile."i3blocks/config" =
      mkIf (cfg.blocks != { } || cfg.blocks != { }) {
        source = let
          blocks = pipe cfg.blocks [
            (map (b: { ${b.name} = (removeAttrs b [ "name" ]); }))
            (map (tomlFormat.generate "i3blocks-config-block"))
            (concatStringsSep "\n")
          ];
          globalVars =
            tomlFormat.generate "i3blocks-config-globalVars" cfg.globalVars;
        in pkgs.concatText "i3blocks-config" [ globalVars blocks ];
      };

```

VAWVAW:
Unfortunately, i3blocks doesn't use flat toml. It just hard cuts at the `=`. So in that case the value of 'command ' (with a space) would be ' "date '+%d.%m.%4Y %T'"'. So I think we may be able use the ini generator. I wasn't able to find documentation for it but from looking at the source and trying, it looks like ini only accepts a set of sets. That would sort the blocks alphabetically again. Any idea whether this could be overwritten?

VAWVAW:
I found `generators.toINI` which works but needs manual type definition.
``` nix
blocks = mkOption {
  type = types.listOf (with types; attrsOf (oneOf [ str number ]));
  default = [ ];
};

...

xdg.configFile."i3blocks/config" =
  mkIf (cfg.blocks != { } || cfg.blocks != { }) {
    text = concatStringsSep "\n" [
      (generators.toINI {} cfg.globalVars)
      (pipe cfg.blocks [
        (map (b: { ${b.name} = (removeAttrs b [ "name" ]); }))
        (map (generators.toINI {} ))
        (concatStringsSep "\n")
      ])
    ];
  };

```

h7x4:
To be honest, I haven't found any good documentation for the `pkgs.formats` either. I think these was earlier part of the nixpkgs `lib.generators` attrset, [which is pretty well documented in the manual](https://ryantm.github.io/nixpkgs/functions/generators/), but the new ones comes with types and stuff as well. I take it you've already read this file?

https://github.com/NixOS/nixpkgs/blob/master/pkgs/pkgs-lib/formats.nix

There is a way to override order in home-manager, by using `hm.dag`. But I think this will become somewhat painful if you need custom order for all the blocks(compared to just one specific value needing to be behind another within an attrset).

The code I posted above keeps the order of the blocks by individually converting each block to `toml`, and then concatenating them together as text in the order of the list. If you modify it to use `iniFormat = pkgs.formats.ini { }` and replacing all instances of `tomlFormat` with `iniFormat`, I think it should work the way you intend it to.

VAWVAW:
Yes, I took a look at the source code. The problem that persists with `formats.ini` is the type of the option. `iniFormat.type` only allows for a set of sets while we want a list of sets. So we could use the parser but we still need manual type definition.

VAWVAW:
`formats.ini` doesn't work because it adds the formatting with spaces and quotes like `formats.toml`. I also realized that `generators.toINI` only works for the blocks, because it expects a set of sets as well. Do you have any other idea than to copy part of the `generators.toINI` definition?

h7x4:
I took a deeper look into things today, and I learned some new things:

- Despite the INI format being quite loosely defined, everyone seems to agree one some ideas. One of those ideas is that there are no such thing as "global options" at the top of the file. Everything has to be put under block headers. This is probably the reason why `nixpkgs/pkgs/pkgs-lib/formats.nix` doesn't export its `iniAtom` type - it shouldn't need to as long as you're dealing with proper INI.

- The last commit to [vivien/i3blocks](https://github.com/vivien/i3blocks) was three years ago. I do not expect a PR asking to optionally put global options under `[Global]` would go through anytime soon.

These two options come to mind:

- We could generate `globalOptions` as one ini block, and just cut off the text header afterwards. It would look something like this (not tested):

```nix
globalOptionsText =
  let globalOptionsAsIniBlock = { global = globalOptions };
  in pipe globalOptionsAsIniBlock [
    iniFormat.generate
    builtins.readFile
    (splitString "\n")
    (drop 1)
    (concatStringsSep "\n")
  ];
```

Then we use the trick with mapping `iniFormat.generate` over all the blocks separately before concatenating all the text in the end.

- We could just consider this to be another format and use your original code. The config format doesn't show any signs of changing anytime soon ¯\\\_(ツ)\_/¯

I don't have any preferences at the moment, so I think I'll wait and see what the home-manager admins think.

PR #3378, opened 2022-11-02
xfconf: add module

chuangzhu:
### Description

Add a module for Xfconf, the configuration system used by XFCE, similar to GNOME/GTK's dconf.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/misc/xfconf.nix
```diff
@@ -0,0 +1,79 @@
...
+    };
+  };
+
+  config = mkIf (cfg.enable && cfg.settings != { }) {
+    home.activation.xfconfSettings = hm.dag.entryAfter [ "installPackages" ]
+      (let
+        commands = mapAttrsToList (channel: properties:
+          mapAttrsToList (property: value: ''
+            $DRY_RUN_CMD ${pkgs.xfce.xfconf}/bin/xfconf-query -n \
+              -c "${channel}" -p "/${property}" ${withType value}
```
rycee, line 75: I think you are risking problems with escaping. For example, consider a user wanting to set a string value that contains a quote character (`"`). Perhaps something like

```suggestion
            $DRY_RUN_CMD ${pkgs.xfce.xfconf}/bin/xfconf-query \
              ${escapeShellArgs ([ "-n" "-c" "${channel}" "-p" "/${property}" ]
                ++ withType value)}
```

modules/misc/xfconf.nix
```diff
@@ -0,0 +1,79 @@
...
+    else if builtins.isInt v then
+      "-t int -s ${toString v}"
+    else if builtins.isFloat v then
+      "-t float -s ${toString v}"
+    else if builtins.isString v then
+      ''-t string -s "${toString v}"''
+    else if builtins.isList v then
+      "-a ${concatMapStringsSep " " withType v}"
+    else
+      throw "unexpected xfconf type: ${builtins.typeOf v}";
```
rycee, line 21: Together with the `escapeShellArgs` change below:

```suggestion
    if builtins.isBool v then
      [ "-t" "bool" "-s" (if v then "true" else "false") ]
    else if builtins.isInt v then
      [ "-t" "int" "-s" (toString v) ]
    else if builtins.isFloat v then
      [ "-t" "float" "-s" (toString v) ]
    else if builtins.isString v then
      [ "-t" "string" "-s" v ]
    else if builtins.isList v then
      [ "-a" ] ++ withType v
    else
      throw "unexpected xfconf type: ${builtins.typeOf v}";
```

modules/misc/xfconf.nix
```diff
@@ -0,0 +1,79 @@
...
+          };
+        }
+      '';
+      description = ''
+        Settings to write to the Xfconf configuration system.
+      '';
+    };
+  };
+
+  config = mkIf (cfg.enable && cfg.settings != { }) {
```
rycee, line 69: This module is Linux specific so need to add an assertion:

```suggestion
  config = mkIf (cfg.enable && cfg.settings != { }) {
    assertions = [
      (hm.assertions.assertPlatform "xfconf" pkgs platforms.linux)
    ];

```

rycee:
Thanks! I've added a few comments.

pjones:
@chuangzhu I tried to use the new `xfconf` module but it doesn't look like it works:

```
hm-activate-pjones[4004]: Failed to init libxfconf: Cannot autolaunch D-Bus without X11 $DISPLAY.
```

This results in the home-manager activation failing.  As far as I can tell there's no way to run `xfconf-query` without XFCE running.

Any suggestions?

gou4shi1:
I meet the same issue as @pjones

PR #3381, opened 2022-11-03
home-environment: reset PATH in activation script

rycee:
### Description

Starting with state version 22.11 we completely reset the PATH variable in the activation script. This is to avoid impurities and unexpected results if the activation script accidentally uses a command found in the user's PATH.

### Checklist

- [x] Change is backwards compatible. _Should be backwards compatible due to the state version guard_

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/docs/release-notes/rl-2211.adoc
+++ b/docs/release-notes/rl-2211.adoc
@@ -97,3 +97,17 @@ These changes are only active if the `home.stateVersion` option is set to "22.11
 value of <<opt-xdg.userDirs.music>> if <<opt-xdg.userDirs.enable>> is
 enabled. Otherwise it is undefined and must be specified in the user
 configuration.
+
+* The activation script now resets `PATH` before running. Before, the
+user's `PATH` environment variable would be used in the script and
+this made it possible for commands in the activation script to run
+arbitrary commands accessible to the user. We now restrict the
+activation script to commands that are explicitly specified.
++
+There is no official way to restore the old behavior. We attempt to
+make the activation script as reproducible as possible and honoring
+the user's `PATH` reduces reproducibility.
++
+If you need to run a command in an activation script block then refer
+to the command by its absolute command path, such as
+`${pkgs.hello}/bin/hello`.
--- a/modules/home-environment.nix
+++ b/modules/home-environment.nix
@@ -346,12 +346,18 @@ in
 
     home.emptyActivationPath = mkOption {
       internal = true;
-      default = false;
       type = types.bool;
+      default = versionAtLeast stateVersion "22.11";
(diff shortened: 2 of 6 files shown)


rycee:
@midchildan Could you help me out here? I'm trying to make the HM activation script more self-contained by resetting the user's `PATH` at the beginning of the activation. The idea is to make the script more robust and reproducible since it won't be possible to run any random tool that happens to be on the user's system.

The worry I have is that the launchd activation script block won't work since it uses `launchctl`, which presumably comes from the user's `PATH` variable. Would it be possible to reference the full path of `launchctl` in the script block? For example by doing something similar to the `systemctlPath` option in the systemd module: https://github.com/nix-community/home-manager/blob/93335810751f0404fe424e61ad58bc8e94bf8e9d/modules/systemd.nix#L108-L117

midchildan:
On macOS Monterey, `launchctl` and `defaults` are located in `/bin/launchctl` and `/usr/bin/defaults`, respectively. I was able to run them both with an empty `PATH`, exactly like they're used in the activation script.

rycee:
@midchildan Hmm, I'm surprised it works with an empty PATH. On my system, for example, I get

``` console
$ systemctl --version
systemd 250 (250.4)
+PAM +AUDIT -SELINUX +APPARMOR +IMA +SMACK +SECCOMP +GCRYPT -GNUTLS +OPENSSL +ACL +BLKID +CURL +ELFUTILS +FIDO2 +IDN2 -IDN +IPTC +KMOD +LIBCRYPTSETUP +LIBFDISK +PCRE2 -PWQUALITY -P11KIT -QRENCODE +BZIP2 +LZ4 +XZ +ZLIB +ZSTD -BPF_FRAMEWORK -XKBCOMMON +UTMP -SYSVINIT default-hierarchy=unified

$ PATH="" systemctl --version
bash: systemctl: No such file or directory
```

So for systemctl we need the absolute path in the activation script. I would have expected the same to be true for launchctl and defaults.

midchildan:
I did specify the full path to invoke the commands though. What I meant was that it worked with the same command line arguments.

rycee:
Ah, sorry I misunderstood. It's great that they are at specific locations. So if I understand things correctly, it should suffice to, for example, change

    $DRY_RUN_CMD launchctl bootstrap "$domain" "$dstPath"

to

    $DRY_RUN_CMD /bin/launchctl bootstrap "$domain" "$dstPath"

? If so, would you mind making a PR with such a change? I'll merge it to master and rebase this branch on top.

midchildan:
Yes, replace the commands with their full path should be enough. I made a PR in #3384 to make the changes.

PR #3389, opened 2022-11-05
targets/generic-linux: use the correct nix package

blaggacao:


blaggacao:
didn't work ... :smirk:

PR #3390, opened 2022-11-05
targets/generic-linux: use the correct nix package

blaggacao:


rycee:
I'll close this as a wontfix. The 21.11 branch is unsupported for quite a while so we don't need to do any more commits there.

PR #3395, opened 2022-11-06
programs/bat: add extraPackages

sielicki:
bat has a few additional/extra packages in nixpkgs, allow for these to be specified via hm.

### DIFF
--- a/modules/programs/bat.nix
+++ b/modules/programs/bat.nix
@@ -30,6 +30,16 @@ in {
       '';
     };
 
+    extraPackages = mkOption {
+      type = types.listOf types.package;
+      default = [ ];
+      example = literalExpression
+        "with pkgs.bat-extras; [ batdiff batman batgrep batwatch ];";
+      description = ''
+        Additional bat packages to install.
+      '';
+    };
+
     themes = mkOption {
       type = types.attrsOf types.lines;
       default = { };
@@ -51,7 +61,7 @@ in {
   };
 
   config = mkIf cfg.enable {
-    home.packages = [ pkgs.bat ];
+    home.packages = [ pkgs.bat ] ++ cfg.extraPackages;
 
     xdg.configFile = mkMerge ([{
       "bat/config" =


modules/programs/bat.nix
```diff
@@ -30,6 +30,15 @@ in {
       '';
     };
 
+    extraPackages = mkOption {
+      type = types.listOf types.package;
+      default = [ ];
+      example = "with pkgs.bat-extras: [ batdiff batman batgrep batwatch ];";
```
d-dervishi, line 36:
```suggestion
      example = "with pkgs.bat-extras; [ batdiff batman batgrep batwatch ];";
```

d-dervishi (changes requested):
Maybe it would be better to use a function taking `pkgs.bat-extras` and returning a subset of the packages, like `program.emacs.extraPackages` does? So having the type of `extraPackages` be `functionTo (listOf package)` or `hm.types.selectorFunction`?

modules/programs/bat.nix
h7x4: When grepping through home-manager for the keyword `extraPackages`, it seems like emacs is the only one passing their pkgs subset through a function (of several modules like `neovim`, `nnn`, `xmonad` and even `texlive`). I think the package subset function is a more common pattern in nixpkgs, where I know at least python and haskell uses it to create environment derivations.

I think it's worth noting that the function doesn't really restrict you from using packages from outside the function argument. It just becomes a fancy `with subpkgs;` statement. I'm not sure if there's many advantages to be gained by embracing that pattern, apart from a strong indication of what packages should be used?

Do you have any thoughts on the matter @rycee?

sielicki: in general I disagree with making it resemble an overlay because, it's not really like emacs or vim where there are dozens of plugins that are growing all the time and/or where nix users are going to regularly write drvs on their own. It's also not particularly tied to the version of bat being used, which is the main motivation of `emacsPackagesFor`. There's only a handful and they're just bash scripts.

This is just a simple convenience method to install these via home manager, under the same module that you're using for bat itself. If users do want to include some hypothetical contrib/third party bat plugin, they can still write a drv and put it into this array.

rycee: I think it'll be sufficient to use the plain list of packages in this case.

modules/programs/bat.nix
```diff
@@ -30,6 +30,16 @@ in {
       '';
     };
 
+    extraPackages = mkOption {
+      type = types.listOf types.package;
+      default = [ ];
+      example = literalExpression
+        "with pkgs.bat-extras: [ batdiff batman batgrep batwatch ];";
```
sielicki, line 37: this needs to be a semicolon

sielicki: fixed

rycee:
Thanks! Rebased to master in 50c9bccb6abc52811a59db620606e016fcde32bd.

PR #3398, opened 2022-11-08
zsh: add ignoreAllDups, findNoDups, saveNoDups, bang, verify, beep history options

fmoda3:
### Description

Added new zsh history options to zsh module

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/zsh.nix
+++ b/modules/programs/zsh.nix
@@ -83,12 +83,43 @@ let
         '';
       };
 
+      ignoreAllDups = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          If a new command line being added to the history
+          list duplicates an older one, the older command
+          is removed from the list (even if it is not the
+          previous event).
+        '';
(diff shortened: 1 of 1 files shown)


h7x4 (approved):
All zsh-relevant tests run fine (although there seems to be no relevant tests for these environment variables).

Code seems clean, considering that previous code follows a pattern of not using `mkEnableOption` for boolean settings, even for options with `false` as default.

teto:
I think it's a mistake to override ZSH defaults like we do. We should have those as freeform settings to allow nix reasoning on those values but without having to add all of ZSH options

teto:
@fmoda3 I am personally not keen on merging this since it goes in the wrong direction but if you changed the implementation to a freeform "settings" (grep "freeform" in the codebase to see what I am alluding to) you could achieve the same objective while improving the module. You dont have to port legacy options, it can be done separately.

PR #3399, opened 2022-11-08
lib, zellij: move zellij configuration to the new KDL format

### LINKED ISSUE #3364: zellij is migrating to KDL configuration
### Description

On startup of the latest versions of zellij, it warns the user that it has migrated their config file from YAML to KDL.

The home-manager module should be updated to output KDL, depending on the package version.

h7x4:
Added a generator for the KDL document language.
This is in order for home-manager to natively generate the new config format for zellij, as described in #3364.

There is not a one to one mapping between KDL and nix types, but I've used KDLs JSON-IN-KDL syntax to translate attrsets.

Closes #3364 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/tests/lib/generators/tokdl-result.txt
+++ b/tests/lib/generators/tokdl-result.txt
@@ -1,0 +1,41 @@
+a 1
+b "string"
+bigFlatItems 23847590283751 1.239000 "multiline \" \" \"\nstring\n" null
+c "multiline string\nwith special characters:\n\t \n \\" \"\n"
+extraAttrs 2 true arg1=1 arg2=false {
+	nested {
+		a 1
+		b null
+	}
+}
+flatItems 1 2 "asdf" true null
+listInAttrsInList {
+	list1 {
+		- {
+			a 1
+		}
+		- {
+			b true
+		}
+		- {
+			c null
+			d {
+				- {
+					e "asdfadfasdfasdf"
+				}
+			}
+		}
+	}
+	list2 {
+		- {
+			a 8
+		}
+	}
+}
+nested {
+	- 1 2
+	- true false
+	- 
+	- null
+}
+unsafeString " \" \n 	 "
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -16,6 +16,9 @@ Makefile                                              @thiagokokada
 
 /modules/launchd                                      @midchildan
 
+/modules/lib/generators.nix                           @h7x4
+/test/lib/generators                                  @h7x4
+
 /modules/misc/dconf.nix                               @rycee
 
 /modules/misc/editorconfig.nix                        @loicreynier
--- a/modules/lib/default.nix
+++ b/modules/lib/default.nix
@@ -6,6 +6,7 @@ rec {
   assertions = import ./assertions.nix { inherit lib; };
 
   booleans = import ./booleans.nix { inherit lib; };
+  generators = import ./generators.nix { inherit lib; };
   gvariant = import ./gvariant.nix { inherit lib; };
   maintainers = import ./maintainers.nix;
   strings = import ./strings.nix { inherit lib; };
--- a/modules/lib/generators.nix
+++ b/modules/lib/generators.nix
@@ -1,0 +1,102 @@
+{ lib }:
+
+{
+  toKDL = { }:
+    let
+      inherit (lib) concatStringsSep splitString mapAttrsToList any;
+      inherit (builtins) typeOf replaceStrings elem;
+
+      # ListOf String -> String
+      indentStrings = let
+        # Although the input of this function is a list of strings,
+        # the strings themselves *will* contain newlines, so you need
+        # to normalize the list by joining and resplitting them.
+        unlines = lib.splitString "\n";
+        lines = lib.concatStringsSep "\n";
+        indentAll = lines: concatStringsSep "\n" (map (x: "	" + x) lines);
+      in stringsWithNewlines: indentAll (unlines (lines stringsWithNewlines));
+
+      # String -> String
+      sanitizeString = replaceStrings [ "\n" ''"'' ] [ "\\n" ''\"'' ];
+
+      # OneOf [Int Float String Bool Null] -> String
+      literalValueToString = element:
+        lib.throwIfNot
+        (elem (typeOf element) [ "int" "float" "string" "bool" "null" ])
+        "Cannot convert value of type ${typeOf element} to KDL literal."
+        (if typeOf element == "null" then
+          "null"
+        else if element == false then
+          "false"
+        else if element == true then
+          "true"
+        else if typeOf element == "string" then
+          ''"${sanitizeString element}"''
+        else
+          toString element);
+
+      # Attrset Conversion
+      # String -> AttrsOf Anything -> String
+      convertAttrsToKDL = name: attrs:
+        let
+          optArgsString = lib.optionalString (attrs ? "_args")
+            (lib.pipe attrs._args [
+              (map literalValueToString)
+              (lib.concatStringsSep " ")
+              (s: s + " ")
+            ]);
+
+          optPropsString = lib.optionalString (attrs ? "_props")
+            (lib.pipe attrs._props [
+              (lib.mapAttrsToList
+                (name: value: "${name}=${literalValueToString value}"))
+              (lib.concatStringsSep " ")
+              (s: s + " ")
+            ]);
+
+          children =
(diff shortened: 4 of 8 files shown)


rycee:
Neat! Don't forget to run `./format`, though.

h7x4:
I had some problems while running the last `nix-shell --pure tests -A run.all`.
`run.broot` seems to be broken.
However, this was broken before I made any commits, and all relevant tests (just `run.tokdl`) runs fine, so I think a separate PR should be made to fix broot.

quasigod-io:
I'm running into an error when trying to test your branch
```
error: undefined variable 'versionAtleast'

       at /nix/store/5i0glv8zf7fc8fa70mql59din6gy70x1-source/modules/programs/zellij.nix:60:32:

           59|     home.file."${configDir}/config.kdl" = mkIf
           60|       (cfg.settings != { } && (versionAtleast cfg.package.version "0.32.0")) {
             |                                ^
           61|         text = lib.hm.generators.toKDL { } cfg.settings;

       … while evaluating the attribute 'condition'
```

Looks like it's just a capitalization error, I think it should be `versionAtLeast`.

colemickens:
I am doing a few things at once here, but at quick test gave me an issue -- it seems the converter is translating bools to ints and then Zellij isn't happy about it.

```
  × Failed to parse Zellij configuration
   ╭─[/home/cole/.config/zellij/config.kdl:5:1]
 5 │ scrollback_editor "hx"
 6 │ simplified_ui 1
   · ───────┬───────
   ·        ╰── Property simplified_ui must be true or false, found 1
 7 │ theme "default"
   ╰────
```

but in my nix:

```nix
{
  config = {
    home-manager.users.cole = { pkgs, ... }: {
      programs.zellij = {
        enable = true;
        settings = {
          default_mode = "normal";
          ui.pane_frames.rounded_corners = true;
          default_layout = "compact";
          default_shell = "nu";
          simplified_ui = true;
        };
      };
    };
  };
```

colemickens:
Looking at the PR, it seems that the test in fact checks for this. I'm not familiar with KDL to know if Zellij is being picky or if the test is testing the wrong thing, but I thought I'd mention.

h7x4:
> Looking at the PR, it seems that the test in fact checks for this. I'm not familiar with KDL to know if Zellij is being picky or if the test is testing the wrong thing, but I thought I'd mention.

Thanks! It is an overlook on my part. I should have inspected the test code more closely.

If you're wondering, the problem is that `builtins.toString` translates `true` and `false` to `"1"` and `"0"` instead of `"true"` and `"false"`.

michaelBelsanti:
What syntax would you use to write a keybind in this? I'm trying to recreate the default Zellij config as a test, but I can't get output matching the syntax of binds, for example:
```
locked {
        bind "Ctrl g" { SwitchToMode "Normal"; }
}
```

h7x4:
@michaelBelsanti 

> What syntax would you use to write a keybind in this? I'm trying to recreate the default Zellij config as a test, but I can't get output matching the syntax of binds, for example:
> 
> ```
> locked {
>         bind "Ctrl g" { SwitchToMode "Normal"; }
> }
> ```

If you deconstruct the semantics of the statement, it would be the following:

The root set contains one node `locked`, which has a nested set as its only value.
That set has one node `bind` which connects to two values: `"Ctrl g"` and `{ SwitchToMode "Normal"; }`

You can represent it like this:

```nix
let
  myConfig = {
    locked = {
      bind = [
        "Ctrl g"
        { SwitchToMode = "Normal"; }
      ];
    };
  };
in g.toKDL { } myConfig
```

The function will produce this as its output:

```txt
locked {
  bind {
    - "Ctrl g"
    - {
      SwitchToMode "Normal"
    }
  }
}
```

However, (and please correct me if I'm wrong here), this should be semantically equivalent to your original KDL config

tests/lib/generators/tokdl-result.txt
```diff
@@ -0,0 +1,41 @@
...
+	}
+}
+flatItems 1 2 "asdf" true null
+listInAttrsInList {
+	list1 {
+		- {
+			a 1
+		}
+		- {
+			b 1
```
colemickens, line 18: I think this should be `b true`

colemickens:
I commented on a test, and still have an issue with `simplified_ui 1` in my config after trying [6dca1bc](https://github.com/nix-community/home-manager/pull/3399/commits/6dca1bcdee922e113008c569e1099ca5ab122e2b)

michaelBelsanti:
@h7x4 
I tried using that syntax like this: 
```
      keybinds = {
        "shared_except \"move\" \"locked\"" = {
          bind = [
            "Ctrl m"
            { SwitchToMode = "Move"; }
          ];
        };
      };

```
which output this config: 
```
keybinds {
	shared_except "move" "locked" {
		bind {
			- "Ctrl m"
			- {
				SwitchToMode "Move"
			}
		}
	}
}
```
Unfortunately, Zellij fails to parse this and outputs this error:
```
  × Failed to parse Zellij configuration
   ╭─[/home/quasi/.config/zellij/config.kdl:4:1]
 4 │ 		bind {
 5 │ 			- "Ctrl m"
   ·    ─────┬────
   ·         ╰── Unsupported action: -
 6 │ 			- {
   ╰────
   ```
   
For context, I am trying to recreate this from my own Zellij configuration:
```
keybinds {
	shared_except "move" "locked" {
	    bind "Ctrl m" { SwitchToMode "move";
	}
}
```

File: tests/lib/generators/tokdl-result.txt (hunk 1 above)
Comment (h7x4) [reply tests/lib/generators/tokdl-result.txt]: Fixed bools!

**Comment by h7x4** [comment]
Added support for [Arguments](https://github.com/kdl-org/kdl/blob/main/SPEC.md#argument) and [Properties](https://github.com/kdl-org/kdl/blob/main/SPEC.md#property), in a similar fashion to #3731. Although the generator now extends a little bit past the JSON-IN-KDL standard, it should make it easier to produce correct results for zellij to parse.

@michaelBelsanti example config

```
keybinds = {
  shared_except = {
    _props = [ "move" "locked" ];
    bind = {
      _props = [ "Ctrl m" ];
      SwitchToMode = "move";
    };
  };
};
```

**Comment by shanesveller** [comment]
Sorry for the necro-reply but how would one express this kind of syntax from the [default keybinds](https://github.com/zellij-org/zellij/blob/7dddaaccd1eb8aa162b9e5747a0504433852e54f/zellij-utils/assets/config/default.kdl#L49-L51)?

```kdl
tab {
    bind "Ctrl t" { SwitchToMode "Normal"; }
    bind "r" { SwitchToMode "RenameTab"; TabNameInput 0; }
    # ...
}
```

In particular I think relative ordering is situationally significant. For a contrived example, `bind "something" { Resize "Decrease Left"; NewPane; }` will have different outcomes than `bind "something" { NewPane; Resize "Decrease Left";  }`.

I experimented with this Nix syntax:

```nix
programs.zellij.keybinds.tab = [
  {
    bind = {
      _args = ["Ctrl t"];
      SwitchToMode = "Normal";
    };
  }
  {
    bind = {
      _args = ["r"];
      SwitchToMode = "RenameTab";
      TabNameInput = 0;
    };
  }
];
```

Which produces this KDL:
```kdl
tab {
    - {
        bind "Ctrl t" {
            SwitchToMode "Normal"
        }
    }
    - {
        bind "r" {
            SwitchToMode "RenameTab"
            TabNameInput 0
        }
    }
}
```
This may be syntactically valid as far as the format is concerned, but using Zellij 0.39.2, `zellij -c some-generated-kdl-path setup --check` reports:
```
  × Failed to parse Zellij configuration
    ╭─[/nix/store/m76z996zf38f7mfgly0xs38cyhk0kgmm-hm_zellijconfig.kdl:4:1]
  4 │         tab {
  5 │ ╭─▶         - {
  6 │ │               bind "Ctrl t" {
  7 │ │                   SwitchToMode "Normal"
  8 │ │               }
  9 │ ├─▶         }
    · ╰──── Unknown keybind instruction: '-'
 10 │             - {
    ╰────
  help: For more information, please see our configuration guide: https://zellij.dev/
        documentation/configuration.html
```

Unbinding existing defaults would have the same issue, and using `clear-defaults=true` args would also run into the above situation - how would you repeat `bind`, `unbind`, or other repeatable terms with differing contents in a Nix data structure?

h7x4:
> Sorry for the necro-reply but how would one express this kind of syntax [...]

No worries. The current KDL generator is quite broken, and can only represent a small subset of what KDL is capable of. It's a known issue, and it's currently being worked on (see #4614 for some extra details). I'd recommend just writing out the KDL config as plaintext in the meanwhile.

PR #3405, opened 2022-11-10
dbus: pass absolute path of dbus-daemon to dbus-run-session

NickCao:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/dconf.nix
+++ b/modules/misc/dconf.nix
@@ -102,7 +102,7 @@ in {
       if [[ -v DBUS_SESSION_BUS_ADDRESS ]]; then
         export DCONF_DBUS_RUN_SESSION=""
       else
-        export DCONF_DBUS_RUN_SESSION="${pkgs.dbus}/bin/dbus-run-session"
+        export DCONF_DBUS_RUN_SESSION="${pkgs.dbus}/bin/dbus-run-session --dbus-daemon=${pkgs.dbus}/bin/dbus-daemon"
       fi
 
       if [[ -v oldGenPath ]]; then


oluceps (approved):

oxalica:
This would expect that `dbus-daemon` is not started when `dbus-run-session` is called, and try to spawn `dbus-daemon` with the environment of `dbus-run-session`? Doesn't this spawned one conflict with user `dbus.service`, and have may-be-different envvars between every boot?

oxalica:
I assume the main issue here is that `dconf` depends on the user `dbus.services`. Could we just move this dconf setup into a user service after `dbus`, and just reload/restart it during activation?

tstachl:
@oxalica just jumping in here really quickly, this pull request solves an issue I ran into today. I'm currently pointing my flake input to @NickCao. So, while I appreciate your concern for future greatness, I think your suggestion should become an item on the todo list and this should be merged to mitigate an occurring issue.

brianmcgee:
I am having trouble with https://github.com/danth/stylix applying GTK styles, activation script as failing with:

`dbus-run-session: failed to execute message bus daemon 'dbus-daemon': No such file or directory`

Switching my flake input to @NickCao's and all is good.

NickCao:
Did a few experiments, and finds out that this issue only manifests when `programs.dconf.enable = false;` (nixos option).

Myaats:
> Did a few experiments, and finds out that this issue only manifests when `programs.dconf.enable = false;` (nixos option).

I tried to set it to true (in nixos options), but dconf activation still fails on my device on boot. It executes without issues when I am already in a GNOME session and restart the home-manager-$USER service manually.
```
Nov 21 16:11:23 shun hm-activate-mats[879]: Starting Home Manager activation
Nov 21 16:11:23 shun hm-activate-mats[879]: Activating checkFilesChanged
Nov 21 16:11:23 shun hm-activate-mats[879]: Activating checkLinkTargets
Nov 21 16:11:23 shun hm-activate-mats[879]: Activating writeBoundary
Nov 21 16:11:23 shun hm-activate-mats[879]: Activating installPackages
Nov 21 16:11:23 shun hm-activate-mats[879]: Activating dconfSettings
Nov 21 16:11:23 shun hm-activate-mats[1318]: dbus-run-session: failed to execute message bus daemon 'dbus-daemon': No such file or directory
Nov 21 16:11:23 shun hm-activate-mats[1316]: dbus-run-session: EOF reading address from bus daemon
Nov 21 16:11:24 shun systemd[1]: home-manager-mats.service: Main process exited, code=exited, status=127/n/a
```

Unsure if this is a separate issue or not

NickCao:
However the underlying issue is, as long as the current activation script runs into the else branch, it immediately fails. I think we should merge this and rethink about the logic around dconf in a separate issue.

thiagokokada:
Would also be nice to backport this to release-22.11 if this is merged.

rycee:
Thanks! Rebased to master in 02c05460339e5239ef6fc83eecea047e7e879704. Also cherry-picked to 22.11.

PR #3410, opened 2022-11-12
swaylock: add enable and package option

katexochen:
Signed-off-by: Paul Meyer <49727155+katexochen@users.noreply.github.com>

### Description

Adding `enable` and `package` options to the swaylock module. 

:warning: These changes are not fully backwards compatible, as the config file from `settings` won't be written if enable isn't set. We could write the settings file when non-empty even if the module isn't enabled to retain backwards compatibility, but I couldn't find another module that does this, so I didn't implement it for now.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/swaylock.nix
+++ b/modules/programs/swaylock.nix
@@ -40,6 +40,11 @@ in {
   };
 
   config = mkIf cfg.enable {
+    assertions = [
+      (lib.hm.assertions.assertPlatform "programs.swaylock" pkgs
+        lib.platforms.linux)
+    ];
+
     home.packages = [ cfg.package ];
 
     xdg.configFile."swaylock/config" = mkIf (cfg.settings != { }) {
--- a/modules/programs/swaylock.nix
+++ b/modules/programs/swaylock.nix
@@ -1,32 +1,54 @@
-{ config, lib, ... }:
+{ pkgs, config, lib, ... }:
+
+with lib;
 
 let cfg = config.programs.swaylock;
 in {
-  meta.maintainers = [ lib.hm.maintainers.rcerc ];
+  meta.maintainers = [ hm.maintainers.rcerc ];
 
-  options.programs.swaylock.settings = lib.mkOption {
-    type = with lib.types; attrsOf (oneOf [ bool float int str ]);
-    default = { };
-    description = ''
-      Default arguments to <command>swaylock</command>. An empty set
-      disables configuration generation.
-    '';
-    example = {
-      color = "808080";
-      font-size = 24;
-      indicator-idle-visible = false;
-      indicator-radius = 100;
-      line-color = "ffffff";
-      show-failed-attempts = true;
+  options.programs.swaylock = {
+    enable = mkOption {
+      type = lib.types.bool;
+      default = versionOlder config.home.stateVersion "23.05"
+        && (cfg.settings != { });
+      defaultText = literalExpression ''
+        true  if state version < 23.05 and settings ≠ { },
+        false otherwise
+      '';
+      example = true;
+      description = "Whether to enable swaylock.";
+    };
+
+    package = mkPackageOption pkgs "swaylock" { };
+
+    settings = mkOption {
+      type = with types; attrsOf (oneOf [ bool float int str ]);
+      default = { };
+      description = ''
+        Default arguments to <command>swaylock</command>. An empty set
+        disables configuration generation.
+      '';
+      example = {
+        color = "808080";
+        font-size = 24;
+        indicator-idle-visible = false;
+        indicator-radius = 100;
+        line-color = "ffffff";
+        show-failed-attempts = true;
+      };
     };
   };
 
-  config.xdg.configFile."swaylock/config" = lib.mkIf (cfg.settings != { }) {
-    text = lib.concatStrings (lib.mapAttrsToList (n: v:
-      if v == false then
-        ""
-      else
-        (if v == true then n else n + "=" + builtins.toString v) + "\n")
-      cfg.settings);
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    xdg.configFile."swaylock/config" = mkIf (cfg.settings != { }) {
+      text = concatStrings (mapAttrsToList (n: v:
+        if v == false then
+          ""
+        else
+          (if v == true then n else n + "=" + builtins.toString v) + "\n")
+        cfg.settings);
+    };
   };
 }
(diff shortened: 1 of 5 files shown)


modules/programs/swaylock.nix
```diff
@@ -1,32 +1,44 @@
...
+
+with lib;
 
 let cfg = config.programs.swaylock;
 in {
-  meta.maintainers = [ lib.hm.maintainers.rcerc ];
+  meta.maintainers = [ hm.maintainers.rcerc ];
+
+  options.programs.swaylock = {
+    enable = mkEnableOption "swaylock";
```
sumnerevans, line 10: We could add an assertion to ensure that if `programs.swaylock.settings != { }`, that `programs.swaylock.enable` must be `true`.

That would make the migration easier.

katexochen: Good idea, I will implement this. :)

ambroisie: I'd suggest making it a waning instead of an error.

katexochen: I'm not quite sure how to exactly combine the combine these things. If enable isn't set, should we still write the config file in addition to printing the warning? Or should only the warning be printed and config isn't written?
  
  ```nix
  config = {
    home.packages = mkIf cfg.enable [ cfg.package ];

    warnings = mkIf (!cfg.enable && cfg.settings != { }) [''
      Warning: swaylock.settings is set but swaylock is disabled
    ''];

    xdg.configFile."swaylock/config" = mkIf (cfg.settings != { }) {
    # OR
    xdg.configFile."swaylock/config" = mkIf (cfg.enable && cfg.settings != { }) {
```

Comment (katexochen) [reply modules/programs/swaylock.nix:10]: @rycee @rcerc any opinion on this topic? I also wonder if there is a policy for the this in general. Should every program have an `enable`/`package` option? I guess this is what I would expect as a user...

Comment (katexochen) [reply modules/programs/swaylock.nix:10]: >  I do not see why someone would configure swaylock in Home Manager but not want it to generate a configuration file. (If the user would really like to disable it, they could reset programs.swaylock.settings with lib.mkForce { }.)

I understand your point, but isn't that the case for most/all home-manager programs? For what reason does the enable option then exist?

Comment (rcerc) [reply modules/programs/swaylock.nix:10]: I cannot answer that (maybe someone else can), but I am not against adding an `enable` option to the `swaylock` module.

Comment (katexochen) [reply modules/programs/swaylock.nix:10]: Sure, just wanted to understand... :)

**Comment by katexochen** [comment]
Sorry, I hadn't much time lately. I've changed the enable option to only scope the package, so the settings are still written to a config file even if enable isn't  set. This way we keep the backwards compatibility.

File: modules/programs/swaylock.nix (hunk 1 above)
Comment (ambroisie) [reply modules/programs/swaylock.nix:10]: I'm personally of the opinion that an `enable` option can be good, is more in-line with the rest of the modules, and a minimal change. Same reasoning for `package`.

File: tests/modules/programs/swaylock/settings.nix [hunk 2]
```diff
@@ -1,11 +1,13 @@
...
-  programs.swaylock.settings = {
-    color = "808080";
-    font-size = 24;
-    indicator-idle-visible = false; # Test that this does nothing
-    indicator-radius = 100;
-    line-color = "ffffff";
-    show-failed-attempts = true;
+{ config, ... }: {
+  programs.swaylock = {
+    settings = {
```
rcerc, line 3: `settings` does not have to be nested here because the `enable` and `package` options are not being set alongside it any more. However, if it's all right to change this file in this pull request, we could remove `{ ... }:` to simplify it because it does not accept any parameters anyway.

rcerc (approved):
Thanks for spending time on this!

tests/modules/programs/swaylock/settings.nix
katexochen: Fixed it. :)

rcerc (approved):
Thanks!

modules/programs/swaylock.nix
rycee: Sorry for the delay. It is typically good to have an enable option, in some cases i've omitted it and I've usually have regretted it in hindsight. Personally I'm not fond of the package option but people seem very attached to them so I don't mind if they are added.

In this case I would suggest adding an enable option with the default set to something like

    versionOlder config.home.stateVersion "23.05"

so that old configurations that expect `settings != {}` to be sufficient will continue working. There should be no warning or assertion. The main point of having an enable option is to be able to unconditionally set the other module options and toggle the whole thing on or off with the enable option. See e.g. [my configuration of the gnome-terminal module](https://git.sr.ht/~rycee/configurations/tree/89ad940adc7ca74c60c76003dbee9ad12060c3fc/item/user/common.nix#L677-689).

In other words, if you add an enable option then setting it to false should result in no change to the user's environment. As if the user did not use the module at all.

katexochen: I've added the default as suggested, nice idea. I don't get why the tests are currently failing.

tests/modules/programs/swaylock/enabled.nix
rcerc, line 1: Should this not also test that `~/.config/swaylock/config` appears when `enable` is set and `settings` is specified?

katexochen: That case is checked in `test/modules/programs/swaylock/settings.nix`, isn't it?

I saw, that the stateVersion in `tests/default.nix` is outdated, but that doesn't seem to be the problem with the failing CI, it says: `error: The option 'home.stateVersion' is used but not defined.`

rcerc: > That case is checked in `test/modules/programs/swaylock/settings.nix`, isn't it?

Oh, I think I understand; `enabled.nix` ensures an empty configuration file is not unnecessarily generated if no settings are specified.

> I saw, that the stateVersion in `tests/default.nix` is outdated, but that doesn't seem to be the problem with the failing CI, it says: `error: The option 'home.stateVersion' is used but not defined.`

Yeah, I don't think that is why the CI is failing either. When I run the tests locally on NixOS, they succeed. Even odder is the macOS CI output, which (to my inexperienced eyes) looks like it's trying to test swaylock despite it belonging to the `lib.optionals isLinux` list in `tests/default.nix`.

rcerc (approved):

rycee:
Thanks! I made some additional tweaks and merged to master 🙂

oxalica:
The default behavior of swaylock config generation changes depending on `stateVersion` due to this PR, but is missed from release notes 2305.

PR #3411, opened 2022-11-12
xdg-user-dirs: Allow setting to null to skip setting

amarshall:
### Description

Previously, this module was all-or-nothing with its pre-defined user dirs. This allows e.g. `xdg.userDirs.desktop = null;` to opt-out of some configuration while still benefiting from the rest.

### Checklist

I did not run the full test suite, but ran `run.xdg-user-dirs-null` and `run.mpd-xdg-music-dir` as those are the only tests that appear to exercise `xdg.userDirs`.

    ```
    {component}: {description}

    {long description}
    ```


rycee:
Thanks! Rebased to master in bc90de24d898655542589237cc0a6ada7564cb6c.

PR #3413, opened 2022-11-13
polybar: allow service to read from systemd PATH variable

saltnpepper97:
### Description

Removed hard coded **PATH** variable in order to read from systemd directly. 
This is necessary as it will allow user created scripts to be able to run when using polybar.service.
Currently there is a hard coded variable for **PATH** that does not point to all the programs on the system, causing programs called in scripts to not be found.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/polybar.nix
+++ b/modules/services/polybar.nix
@@ -211,7 +211,6 @@ in {
 
       Service = {
         Type = "forking";
-        Environment = "";
         ExecStart =
           let scriptPkg = pkgs.writeShellScriptBin "polybar-start" cfg.script;
           in "${scriptPkg}/bin/polybar-start";
--- a/modules/services/polybar.nix
+++ b/modules/services/polybar.nix
@@ -211,6 +211,7 @@ in {
 
       Service = {
         Type = "forking";
+        Environment = "$PATH";
         ExecStart =
           let scriptPkg = pkgs.writeShellScriptBin "polybar-start" cfg.script;
           in "${scriptPkg}/bin/polybar-start";
--- a/modules/services/polybar.nix
+++ b/modules/services/polybar.nix
@@ -211,7 +211,7 @@ in {
 
       Service = {
         Type = "forking";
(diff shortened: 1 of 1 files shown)


rycee:
Please fix the commit message to fit [the guidelines](https://github.com/nix-community/home-manager/blob/master/docs/contributing.adoc#sec-commit-style).

Also, please add a description as to why this change is necessary.

modules/services/polybar.nix
```diff
@@ -211,7 +211,7 @@ in {
 
       Service = {
         Type = "forking";
-        Environment = "PATH=${cfg.package}/bin:/run/wrappers/bin";
+        Environment = "";
```
rycee, line 214: If the `PATH` is no longer necessary, then why keep `Environment = ""`?

saltnpepper97: Sure I can remove it altogether. Would I have to open a new pull request??

teto:
> Removed hard coded PATH variable in order to read from systemd directly.

Is there a default for systemd ? what is it ? (honest question, it seemed to me at least on nixos that it was quite empty)

saltnpepper97:
> > Removed hard coded PATH variable in order to read from systemd directly.
> 
> Is there a default for systemd ? what is it ? (honest question, it seemed to me at least on nixos that it was quite empty)
I understand that however I feel as though if the PATH variable is not hard coded people can import whatever paths are needed into systemd themselves
I understand your concern however the paths that are hard coded into the module now cannot allow for a lot of user created scripts to run with polybar when running from the systemd service file. This is due to a lot of the programs being under /nix/store/ or /run/current-system/sw/bin for me at least.

h7x4:
Would it be better to do this the opposite way?

Declare an `extraPackages` option and then either:

1. Add them to the systemd-units `PATH`

2. Create a wrapper with a `PATH`, as in the `nnn` module:

https://github.com/nix-community/home-manager/blob/4fa1ba72a39da509399cce43995e4c1e1960ce9c/modules/programs/nnn.nix#L111-L122

PR #3423, opened 2022-11-16
picom: remove experimentalBackends, add extraArgs

thiagokokada:
### Description

`--experimental-backends` flag was removed in the recent released picom v10. Using it now will result in the program exiting.

v10 also introduces its counter-part, `--legacy-backends`. However this will be removed soon. Instead of adding this as an separate option, add `extraArgs` option so for those that they want they can pass it manuall. It is also more future proof.

### Checklist

  + Removal of option, but it will trigger a proper warning.

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/picom.nix
+++ b/modules/services/picom.nix
@@ -54,6 +54,8 @@ in {
   imports = [
     (mkRemovedOptionModule [ "services" "picom" "refreshRate" ]
       "The option `refresh-rate` has been deprecated by upstream.")
+    (mkRemovedOptionModule [ "services" "picom" "experimentalBackends" ]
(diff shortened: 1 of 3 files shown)


thiagokokada:
The bump was merged recently: https://github.com/NixOS/nixpkgs/pull/201535

It will be in the 22.11 release. so it would be nice to merge it soon.

CC @rycee.

sumnerevans (approved):

sumnerevans:
corresponding Picom changelog: https://github.com/yshui/picom/releases/tag/v10-rc1

PR #3426, opened 2022-11-21
github actions: only run actions on the main repo

h7x4:
### Description

My notification inbox for my home-manager fork looks like this:

image

I could just turn off actions in the forks repository settings, but I suspect I'm not the only one with this issue?

This PR turns off all actions for every repo except `nix-community/home-manager`

Please do tell me if there is a clear intention behind keeping the actions running for forks though. In that case, it should probably stated explicitly in the `Contributing` section of the manual.

### DIFF
--- a/.github/workflows/github_pages.yml
+++ b/.github/workflows/github_pages.yml
@@ -5,6 +5,7 @@ on:
       - master
 jobs:
   publish:
+    if: github.repository_owner == 'nix-community'
     strategy:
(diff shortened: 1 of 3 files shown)


teto (commented):
it can be useful to run the checks in one's fork before opening a PR. You can disable the github actions from your settings instead.

h7x4:
Fair enough

PR #3439, opened 2022-11-24
Bump master release to 23.05

Lillecarl:
### Description

According to my interpretation of
https://github.com/nix-community/home-manager#releases home-manager/master tracks nixos-unstable, which is now on release 23.05.

### Checklist

Doesn't really apply?


rycee:
Thanks for the reminder! I've updated the master branch to use 23.05. Will update the stable version when the release actually happens.

rycee:
Should be resolved since a while back now.

PR #3444, opened 2022-11-27
neovim: Source neovimRcContent directly from store

amarshall:
### Description

The previous version linked the file into home, then sourced that. Since
nothing else expects that file to be there, this is unnecessary.
Additionally, doing so made it impossible to test a built config without
switching, e.g. using `XDG_CONFIG_HOME=… nvim` or `nvim -u`. This
remedies that, at least for this particular reference.

To test this, change from asserting contents of the config file to
actually starting nvim, outputting sentinel values, and then asserting
their values are present. This way it’s tested that nvim loaded the
config, rather than that some config is in a specific place.

This is all in one commit as the test, as written now, would not have
worked before since the previously hard-coded home path was not an
actual file in the test environment.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/tests/modules/programs/neovim/plugin-config.nix
+++ b/tests/modules/programs/neovim/plugin-config.nix
@@ -7,16 +7,14 @@ with lib;
     programs.neovim = {
       enable = true;
       extraConfig = ''
-        " This 'extraConfig' should be present in vimrc
+        let g:hmExtraConfig='HM_EXTRA_CONFIG'
       '';
       plugins = with pkgs.vimPlugins; [
         vim-nix
         {
           plugin = vim-commentary;
           config = ''
-            " plugin-specific config
-            autocmd FileType c setlocal commentstring=//\ %s
-            autocmd FileType c setlocal comments=://
+            let g:hmPlugins='HM_PLUGINS_CONFIG'
           '';
         }
       ];
@@ -24,11 +22,12 @@ with lib;
     };
 
     nmt.script = ''
-      vimrc="$TESTED/home-files/.config/nvim/init-home-manager.vim"
-      vimrcNormalized="$(normalizeStorePaths "$vimrc")"
-
-      assertFileExists "$vimrc"
-      assertFileContent "$vimrcNormalized" "${./plugin-config.vim}"
+      vimout=$(mktemp)
+      echo "redir >> /dev/stdout | echo g:hmExtraConfig | echo g:hmPlugins | redir END" \
+        | ${pkgs.neovim}/bin/nvim -es -u "$TESTED/home-files/.config/nvim/init.lua" \
+        > "$vimout"
+      assertFileContains "$vimout" "HM_EXTRA_CONFIG"
+      assertFileContains "$vimout" "HM_PLUGINS_CONFIG"
     '';
   };
 }
--- a/modules/programs/neovim.nix
+++ b/modules/programs/neovim.nix
@@ -365,15 +365,13 @@ in {
       in mkMerge (
         # writes runtime
         (map (x: x.runtime) pluginsNormalized) ++ [{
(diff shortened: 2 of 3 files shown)


tests/modules/programs/neovim/plugin-config.nix
```diff
@@ -7,28 +7,27 @@ with lib;
...
       '';
       plugins = with pkgs.vimPlugins; [
         vim-nix
         {
           plugin = vim-commentary;
           config = ''
-            " plugin-specific config
-            autocmd FileType c setlocal commentstring=//\ %s
-            autocmd FileType c setlocal comments=://
+            let g:hmPlugins='%%hmPlugins%%'
```
teto, line 17: any reason this takes the form '%%hmPlugins%%' with all the `%` ?

teto: `%%` is sometimes used as a placeholder for stuff, also it appears in neovim paths (in backups IIRC). I would prefer something more vanilla like `HM_PLUGINS`. Other than that good to go.

teto:
thanks !

dbaynard:
> Additionally, doing so made it impossible to test a built config without switching, e.g. using `XDG_CONFIG_HOME=… nvim` or `nvim -u`. This remedies that, at least for this particular reference.

This makes `nvim -u` much more complicated; though setting `XDG_CONFIG_HOME` still works. Is there a straightforward technique to test neovim configuration changes?

I feel a little like `LongtimeUser4`, from https://xkcd.com/1172/…

dbaynard:
Oh I wouldn't say frustration. I see the improvement, yet I don't see the benefit it brings to testing — so I thought I'd ask.

I was looking for how you test configuration changes, without a `switch`, _after_ the change from this PR. Specifically, I have been (quickly) loading the `init.vim` to check for any errors, before I switch to the new configuration. If it were straightforward to test the whole thing, I'd just do that. 

My testing (for home manager files more broadly) is to run a hacky script that builds `".#${PLATFORM?}Configurations.${HOSTNAME?}.config.home-manager.users.${USER?}.home-files"` and then pops up an fzf window to inspect whichever file I want. It also prints the file path, so I can open neovim with `nvim -u` pointed to that. My `init.lua` just sources the `init.vim`.

That approach no longer works as conveniently. The only way I can think to access the `pkgs.writeText` result path is to parse the `init.lua` file.

I'm open to a better way of doing this (where 'this' means everything in this message and more). Ideally I could call `nix run '.#neovim'` or something similarly simple but with a longer attr path.

(Oh, and I use linux and darwin, so a `nixos-container` call wouldn't work, I don't think?)

dbaynard:
That… yes, that is an excellent point. Thanks 🤦 

Is that what you're doing?

dbaynard:
I often find I want to inspect the `init.vim` file, though — do you have a suggestion for how I might do that?

dbaynard:
OK. I think this discussion will be helpful for others, so thanks for helping to explain this. This change _is_ a regression, then, but the workarounds are reasonable.

TIL about `gf` so thanks for that, too! I've been using `bat` and `tmux-thumbs` to do something similar.

I think this works for me, then? It should work with `nixosConfigurations`, too, on nixos. Perhaps I need to add a new flake output for home manager configurations and then I can build _just_ my home-manager modules 🤔 

```zsh
sed -rn 's#.+(/nix/store/.+?nvim-init-home-manager.vim).+#\1#p' "$(nix build --no-link --print-out-paths \
    "n#darwinConfigurations.${HOSTNAME:-$(hostname -s)}.config.home-manager.users.${USER}.home-files" \
  | tail -n1)/.config/nvim/init.lua" | xargs bat
```

PR #3459, opened 2022-11-30
programs: add various programming languages

schuelermine:
### Description

Added modules for Haskell, Python, and Rust that provide packages for them. Added modules for Cabal, GHC, GHCup, Stack, and Mypy, Pip, Pytest, and Cargo, Clippy, RLS, Rust-Analyzer, RustC, Rustfmt, RustUp.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


schuelermine:
RE @teto: I find that this is valuable to other people as it allows more flexibility in configuring their environment. Some people like having a dev environment without a nix shell and/or use the repls as scripting tools or recreationally.

schuelermine:
RE @teto: I find that this is valuable to other people as it allows more flexibility in configuring their environment. Some people like having a dev environment without a nix shell and/or use the repls as scripting tools or recreation

schuelermine:
There already exists pylint.nix, and it’s possible to want to use these tools outside of a project. Also, it’s not hard to maintain, all I’m doing is plucking these tools from nixpkgs into the global environment, and setting up a single config file.

schuelermine:
Also, these allow configuring a toolchain version that is followed by tools you add to your config.
Also, some people work on projects where they don’t control the configuration.

schuelermine:
At any rate, does anybody have any idea why the pipeline might be failing? I tested it on my machine & it seems to work. Is there something pulling the pylint option out of `options` in the install script?

loicreynier:
Although I am not in favor of using HM to configure global dev toolchains, I think that having options to configure global configuration files for tools such as you did with `~/.pylintrc` is a good idea. Those global configurations files are useful when associated tools are used outside project environments. HM already supports [EditorConfig](https://github.com/nix-community/home-manager/blob/e38ce0ae16b68515c913ab50177bc7624f65f569/modules/misc/editorconfig.nix).

Rosuavio:
> I am opposed to that type of support in HM:
> - Development tools are tied to projects so it makes no sense to have it global. IMO global tools is an anti-pattern in nix

I think someone very useful about nix is the layering, system config can be overridden be user config, overridden by shell config. 

While for a particular project it's good to have your tools setup as close to the project as you can (a development shell), sometimes it's useful to have tools at your disposal in a more adhoc manner. Like before you even make a repo for your project. Ghci, for example, is very useful outside of the context of any project as a tool to experiment with ides in Haskell.

teto:
`nix shell nixpkgs#ghc` and I get a shell with ghci and I can choose the version without rebuilding the HM profile.
If you want it permanent, `nix profile install nixpkgs#ghc` (which can be added to HM packages).
Compare that with maintaining code in HM that is already short-staffed. Surely some people find it useful but my question is whether we want to support this in HM (when you can use out of tree HM modules).

~/.pylintrc is a bit different since there is some state involved. I have trouble imaging what a dedicated module could improve over `home.file.".pylintrc".text = ''";` but ready to be convinced (maybe a different output depending on pylint version but then it would screw you in nix-shells ?).

loicreynier:
> ~/.pylintrc is a bit different since there is some state involved. I have trouble imaging what a dedicated module could improve over `home.file.".pylintrc".text = ''";` but ready to be convinced (maybe a different output depending on pylint version but then it would screw you in nix-shells ?).

Clearly, in the case of `~/.pylintrc` the module would just improve by providing the conversion from an attribute set to a proper INI format.

schuelermine:
> as you did with `~/.pylintrc`

I want to make it clear that I did not add the pylint config, it was already there. I just integrated it into the `programs.python` options set declared in this PR. The module was originally contributed by @florpe.

schuelermine:
Also, in my opinion, the mere consolidation of program installation and configuration into semantically named options is already added value of a module.

Another opposing opinion from @Kranzes:
> Adding these modules will lead new users to forget or not understand that they should still have a shell.nix or a devshell in their repo
> new users don't understand the idea of putting all of the tools needed for development in a devshell hermetically in the repo
> more new users not adding their tools to devshell too as they already have them globally will lead to more projects not having reproducible build envs 
> meaning we get the issue of other distros and package managers
> where is user clones a repo
> then uses their own version of the compiler
> or development tools

rycee:
Thanks for the contribution! Clearly quite a bit of work and thought has gone into this.

I must say, however, that I'm feeling a bit reluctant to merge this for the reasons that have already been raised. Specifically that it encourages a mode of development that more or less goes against the "best practice" (at least as perceived by me) of using Nix project specific development environments. For consistency, I think that the pylint module probably should not have been merged.

That said, I think it would be fair to somehow make this available to people that truly need/want this type of configuration. Would it be OK to keep this in a separate repository and then add a reference to it in the HM manual or README? I think some kind of "Third party modules" or "Third party extensions" chapter would be quite useful since I imagine more people may have modules and other things that enhance HM but are too opinionated or experimental to be upstreamed.

It would be a situation similar to how I maintain the [programs.emacs.init](https://gitlab.com/rycee/nur-expressions/-/blob/adde79a140c8527210d42dc3578fe1c4dc1042b3/hm-modules/emacs-init.nix) and [theme.base16](https://gitlab.com/rycee/nur-expressions/-/blob/adde79a140c8527210d42dc3578fe1c4dc1042b3/hm-modules/theme-base16/default.nix) modules outside HM.

schuelermine:
> Would it be OK to keep this in a separate repository and then add a reference to it in the HM manual or README?

That sounds like a great compromise! I’m going to create a repo that contains these modules

schuelermine:
@rycee I have created a repository: https://github.com/schuelermine/xhmm

rycee:
@schuelermine Nice! I've created #3485. Please comment there if you'd like to expand it.

PR #3465, opened 2022-12-03
just: deprecate since completions work out of the box now

### LINKED ISSUE #3449: bug: `programs.just.enable` causes zsh to error on start

### Issue description

When one uses `programs.just.enable` it creates an entry in the `.zshrc` that causes `zsh` to print an error message on start.

The error is similar to the one below:

```text
_arguments:comparguments:327: can only be called from completion function
```

This has been reported by another user without a GH account in the inofficial nix discord, and I am posting it here on their behalf.

We were able to identify the following snippet:

https://github.com/nix-community/home-manager/blob/master/modules/programs/just.nix#L43-L45

Perhaps using `programs.zsh.plugins` would be a better option? Or any other way to extend the `FPATH` appropriately?

### Maintainer CC

For `just`:

* @maximsmol

For zsh (no maintainers in module, therefore pinging random commiters from first page of history)

* @rycee 
* @zimbatm 
* @oxalica 

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.0.2, NixOS, 23.05 (Stoat), 23.05.20221125.5dc7114`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.12.0pre20221126_534332c`
 - channels(nmelzer): `""`
 - channels(root): `"nixos-21.05"`

(ran this on my machine, though the issue is obvious and not related to my setup)
```

maximsmol:
### Description

Closes #3449. Also deprecates the entire thing as completions work out of the box now! Also removes the old completions implementation to avoid double-initializing them

Note: there are no tests for this module and the tests fail with unrelated `No such file or directory: '/nix/store/61h76ya021588gp3g6my168pib7in8w9-broot-1.12.0.tar.gz/resources/default-conf/conf.hjson'` so I'm just ignoring that requirement

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/just.nix
+++ b/modules/programs/just.nix
@@ -9,44 +9,14 @@ let
 in {
   meta.maintainers = [ hm.maintainers.maximsmol ];
 
-  options.programs.just = {
-    enable = mkEnableOption
-      "just, a handy way to save and run project-specific commands";
-
-    package = mkOption {
-      type = types.package;
-      default = pkgs.just;
-      defaultText = literalExpression "pkgs.just";
-      description = "Package providing the <command>just</command> tool.";
-    };
-
-    enableBashIntegration = mkEnableOption "Bash integration" // {
-      default = true;
-    };
-
-    enableZshIntegration = mkEnableOption "Zsh integration" // {
-      default = true;
-    };
-
-    enableFishIntegration = mkEnableOption "Fish integration" // {
-      default = true;
-    };
-  };
-
-  config = mkIf cfg.enable {
-    home.packages = [ cfg.package ];
-
-    programs.bash.initExtra = mkIf cfg.enableBashIntegration ''
-      source ${cfg.package}/share/bash-completion/completions/just.bash
-    '';
-
-    programs.zsh.initExtra = mkIf cfg.enableZshIntegration ''
-      source ${cfg.package}/share/zsh/site-functions/_just
-    '';
-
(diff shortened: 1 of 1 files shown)


rycee:
Thanks! I switched it over to use `mkRemovedOptionModule` and merged to master.

Animeshz:
Fish completions actually doesn't work unless sourced.

srid:
Zsh completions doesn't work either (on macOS).

rycee:
Hmm, that seems odd. Perhaps macOS needs something similar to what `targets.genericLinux.enable = true` does? See https://github.com/nix-community/home-manager/blob/19b87b9ae6ecfd81104a2a36ef8364f1de1b54b1/modules/targets/generic-linux.nix#L73-L89

@Animeshz Are you using NixOS or some other distro (or macOS)?

rycee:
Apologies, I now notice that the generic-linux thing probably doesn't apply to this. I think maybe https://github.com/nix-community/home-manager/blob/19b87b9ae6ecfd81104a2a36ef8364f1de1b54b1/modules/programs/zsh.nix#L583-L585 is more relevant.

srid:
FTR, I use home-manager along with `nix-darwin`, and I am able to get completion working with https://github.com/srid/nixos-config/pull/50

PR #3474, opened 2022-12-05
emacs: add note about `inhibit-startup-message`

rycee:
### Description

Hopefully this can avoid a bit of confusion.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


Zimmi48:
What is the recommended mechanism to inhibit the startup message? Putting the configuration in `.emacs` with `file.".emacs".source`?

PR #3481, opened 2022-12-07
Allow generating a configuration file containing ordered name-value pairs from a DAG

rcerc:
This adds a `lib.hm.dag.toJson` function to convert DAGs to JSON objects of ordered name-value pairs. The package set used by Home Manager is also extended so modules can generate configuration files of other formats containing ordered name-value pairs.

The package set used by the Home Manager, NixOS and nix-darwin modules is extended with an overlay stored in `modules/pkgs/` because I could not think of a better way to provide the configuration file generators that depend on Nixpkgs (they require the [`remarshal`](https://github.com/remarshal-project/remarshal) package in particular).

This pull request is supposed to be a step towards a solution to #2519. Since I might not have tested these changes adequately (and maybe broke something I’m not aware of), this is a draft pull request for now.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


arcnmx:
it's worth noting another example of a module that could benefit from this: `programs.imv.settings`: [if `options.suppress_default_binds = true`, the `binds` section must appear after the `options` section](https://todo.sr.ht/~exec64/imv/5)

ncfavier:
Could this also resolve https://github.com/nix-community/home-manager/issues/4090 ?

xfzv:
> it's worth noting another example of a module that could benefit from this: `programs.imv.settings`: [if `options.suppress_default_binds = true`, the `binds` section must appear after the `options` section](https://todo.sr.ht/~exec64/imv/5)

Indeed, I'm having this issue. With:

```nix
{
  programs.imv = {
    enable = true;
    settings = {
      options = {
        suppress_default_binds = true;
      };
      binds = {
        "q" = "quit";
      };
}
```

The final config in `$XDG_CONFIG_HOME/imv/config` is:

```ini
# Generated by Home Manager.
[binds]
q=quit

[options]
suppress_default_binds=true
```

and since `suppress_default_binds=true` is written last, all bindings are cleared.

I guess I could use `xdg.configFile."imv/config".text` to force the correct order but is there a better workaround nowadays?

PR #3499, opened 2022-12-15
ssh: add addKeysToAgent option

lockejan:
### Description

add addKeysToAgent option to ssh module

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/ssh.nix
+++ b/modules/programs/ssh.nix
@@ -361,6 +361,17 @@ in
       '';
     };
 
+    addKeysToAgent = mkOption {
+      type = types.str;
+      default = "no";
+      description = ''
+        When enabled, a private key that is used during authentication will be
+        added to ssh-agent if it is running (with confirmation enabled if
+        set to 'confirm'). The argument must be 'no' (the default), 'yes', 'confirm'
+        (optionally followed by a time interval), 'ask' or a time interval (e.g. '1h').
+      '';
+    };
+
     compression = mkOption {
       default = false;
       type = types.bool;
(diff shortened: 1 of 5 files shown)


modules/programs/ssh.nix
```diff
@@ -355,6 +355,16 @@ in
       '';
     };
 
+    addKeysToAgent = mkOption {
+      type = types.enum [ "yes" "no" "ask" "confirm" ];
```
rycee, line 359: Unfortunately I don't think an `enum` would work in this case since it wouldn't allow a value with a timeout, such as "1h" or "confirm 1h". If we add this option then I suspect it would have to be a plain `types.str`.

lockejan: My concerns persist, but I've changed the PR as suggested. I've rebased and ran all tests again. I leave the final decision to @rycee . Feel free to close or merge.

lockejan:
My concerns persist, but I've changed the PR as suggested. I've rebased and ran all tests again. I leave the final decision to @rycee . Feel free to close or merge.

sumnerevans (approved):

sumnerevans:
Probably needs a rebase to get the tests passing.

lockejan:
> Probably needs a rebase to get the tests passing.

Maybe you are right. I rebased and ran the tests again locally, this time against the flake inputs `nix develop --ignore-environment .#all`.
No errors whatsoever.
Should be fine now.

rycee:
Sorry about the delay! Finally round to merging this 🙂

sersorrel:
>  Change is backwards compatible.

unfortunately this change inscrutably breaks behaviour for anyone who set `programs.ssh.extraConfig` to `"AddKeysToAgent yes"` already, since per ssh_config(5) "for each parameter, the first obtained value will be used" and the new `programs.ssh.addKeysToAgent` inserts its `AddKeysToAgent` line before `extraConfig`.

PR #3505, opened 2022-12-17
`exa`: add more options

athre0z:
### Description

This PR adds more configuration options for the `exa` module.

### Checklist

- [x] Test cases updated/added (doesn't appear like any exist)

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/exa.nix
+++ b/modules/programs/exa.nix
@@ -46,6 +46,8 @@ with lib;
       ++ optional cfg.git "--git" ++ cfg.extraOptions);
 
     aliases = {
+      # Use `command` instead of hardcoding the path to exa so that aliases don't
+      # go stale after a system update.
       exa = "command ${cmd}";
       ls = "exa";
       ll = "exa -l";
@@ -61,5 +63,9 @@ with lib;
     programs.zsh.shellAliases = mkIf cfg.enableAliases aliases;
 
     programs.fish.shellAliases = mkIf cfg.enableAliases aliases;
+
+    # ion doesn't support the standard `command` built-in. Or recursive alias expansion.
+    programs.ion.shellAliases =
+      mkIf cfg.enableAliases (aliases // { exa = cmd; });
   };
 }
--- a/modules/programs/exa.nix
+++ b/modules/programs/exa.nix
@@ -42,9 +42,8 @@ with lib;
   config = let
     cfg = config.programs.exa;
 
-    cmd = escapeShellArgs ([ "${cfg.package}/bin/exa" ]
-      ++ optional cfg.icons "--icons" ++ optional cfg.git "--git"
-      ++ cfg.extraOptions);
+    cmd = escapeShellArgs ([ "exa" ] ++ optional cfg.icons "--icons"
+      ++ optional cfg.git "--git" ++ cfg.extraOptions);
 
     aliases = {
       exa = cmd;
--- a/modules/programs/exa.nix
+++ b/modules/programs/exa.nix
@@ -2,19 +2,7 @@
 
 with lib;
 
-let
-
-  cfg = config.programs.exa;
-
-  aliases = {
-    ls = "${pkgs.exa}/bin/exa";
-    ll = "${pkgs.exa}/bin/exa -l";
-    la = "${pkgs.exa}/bin/exa -a";
-    lt = "${pkgs.exa}/bin/exa --tree";
-    lla = "${pkgs.exa}/bin/exa -la";
-  };
-
-in {
+{
   meta.maintainers = [ hm.maintainers.kalhauge ];
 
   options.programs.exa = {
@@ -23,10 +11,50 @@ in {
 
     enableAliases = mkEnableOption "recommended exa aliases";
 
+    extraOptions = mkOption {
+      type = types.listOf types.str;
+      default = [ ];
+      example = [ "--group-directories-first" "--header" ];
+      description = ''
+        Extra command line options passed to exa.
+      '';
+    };
+
+    icons = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Display icons next to file names (<option>--icons</option> argument).
+      '';
+    };
+
+    git = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        List each file's Git status if tracked or ignored (<option>--git</option> argument).
+      '';
+    };
+
     package = mkPackageOption pkgs "exa" { };
   };
 
-  config = mkIf cfg.enable {
+  config = let
+    cfg = config.programs.exa;
+
+    cmd = concatStringsSep " " ([ "${cfg.package}/bin/exa" ]
+      ++ optional cfg.icons "--icons" ++ optional cfg.git "--git"
+      ++ cfg.extraOptions);
+
+    aliases = {
+      exa = cmd;
+      ls = cmd;
+      ll = "${cmd} -l";
+      la = "${cmd} -a";
+      lt = "${cmd} --tree";
+      lla = "${cmd} -la";
+    };
+  in mkIf cfg.enable {
     home.packages = [ cfg.package ];
 
     programs.bash.shellAliases = mkIf cfg.enableAliases aliases;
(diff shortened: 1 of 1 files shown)


teto (approved):

arjan-s (approved):
This is great! Glad I searched first, I was just planning on working on this!

modules/programs/exa.nix
```diff
@@ -23,10 +11,49 @@ in {
...
+    };
+
     package = mkPackageOption pkgs "exa" { };
   };
 
-  config = mkIf cfg.enable {
+  config = let
+    cfg = config.programs.exa;
+
+    baseCmd = concatStringsSep " " ([ "${pkgs.exa}/bin/exa" ]
```
rycee, line 45: Perhaps this should be
```suggestion
    baseCmd = concatStringsSep " " ([ "${cfg.package}/bin/exa" ]
```
?

modules/programs/exa.nix
```diff
@@ -23,10 +11,49 @@ in {
...
+      description = ''
+        Extra command line options passed to exa.
+      '';
+    };
+
+    icons = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Display icons next to file names (`--icons` argument).
```
rycee, line 27:
```suggestion
        Display icons next to file names (<option>--icons</option> argument).
```

modules/programs/exa.nix
```diff
@@ -23,10 +11,49 @@ in {
...
+      '';
+    };
+
+    git = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        List each file's Git status, if tracked or ignored (`--git` argument).
```
rycee, line 35:
```suggestion
        List each file's Git status, if tracked or ignored (<option>--git</option> argument).
```

modules/programs/exa.nix
```diff
@@ -23,10 +11,49 @@ in {
 
     enableAliases = mkEnableOption "recommended exa aliases";
 
+    extraOptions = mkOption {
+      type = types.listOf types.str;
+      default = [ ];
+      example = [ "--group-directories-first" "--header" ];
+      description = ''
+        Extra command line options passed to exa.
```
rycee, line 19: Maybe worth mentioning that this only applies to aliases defined in the module? Would it make sense to make the options apply in all cases? E.g., by wrapping the original `exa` executable by one that adds the options.

athre0z: Hmm. There's a chance for that to mess with e.g. scripts that use `exa` internally and expect it to work like vanilla `exa`. But then again, I don't think it's commonly used in scripts since there is basically no upside to that compared to just using `ls` in that use-case. I added wrapping -- let's see if anyone complains about it later. :)

stasjok: I think you don't want to hard-code a binary with options. I think usually it's done like this:

```bash
alias ls="exa ${extra_options}"
alias ll="ls -l"
alias la="ls -la"
```

This way user (or script) can always run `\ls`, `/bin/ls`, `command ls`  or whatever to ignore any potential aliases or options.

arjan-s: I agree. As a user I expect aliases to be made, not the 'binary' to be changed.

athre0z: There seems to be precedent for both variants within home-manager. I agree that the variant with just adding a shadowing alias is probably preferable. I force-pushed again with an amended version that implements this instead.

arjan-s:
Would still love to have this. @rycee any more feedback? Or is it ready to merge?

modules/programs/exa.nix
```diff
@@ -23,10 +11,50 @@ in {
...
+    };
+
     package = mkPackageOption pkgs "exa" { };
   };
 
-  config = mkIf cfg.enable {
+  config = let
+    cfg = config.programs.exa;
+
+    cmd = concatStringsSep " " ([ "${cfg.package}/bin/exa" ]
```
ncfavier, line 45:
```suggestion
    cmd = escapeShellArgs ([ "${cfg.package}/bin/exa" ]
```

stasjok:
I would also consider not using absolute path to exa in aliases. Use simple `exa`, because it already should be in `PATH`. Why? For example when I perform a garbage collection on a Nix store, some aliases can become invalid (I have many shells opened in tmux long time ago, so I already bumped on this).

ncfavier:
Good call.

ncfavier:
Looks like we have a problem: if `exa` is defined as `exa ${options}` then fish complains that the function is recursive; making it `command exa ${options}` fixes fish, but then ion doesn't know what `command` is... too bad, that's a [standard POSIX built-in](https://pubs.opengroup.org/onlinepubs/007904975/utilities/command.html).

I expect fish to have way more users than ion anyway, so we might just drop the latter?

ncfavier:
Found https://gitlab.redox-os.org/redox-os/ion/-/issues/409, was about to comment that `command` should be implemented but I first need to request an account and then *send them an email explaining why my account request should be approved*... This is not worth my time, dropping ion.

ncfavier:
Re-added ion in order not to break existing functionality. Will merge in a while if no other issues come up.

stasjok:
> Looks like we have a problem: if `exa` is defined as `exa ${options}` then fish complains that the function is recursive; making it `command exa ${options}` fixes fish, but then ion doesn't know what `command` is... too bad, that's a [standard POSIX built-in](https://pubs.opengroup.org/onlinepubs/007904975/utilities/command.html).
> 
> I expect fish to have way more users than ion anyway, so we might just drop the latter?

What config is generated in fish? What version of fish? I tested it and seems to working:

```
$ alias exa="exa --icons"

$ functions exa
# Defined via `source`
function exa --description 'alias exa=exa --icons'
 command exa --icons $argv

end

$ alias ls=exa

$ functions ls
# Defined via `source`
function ls --wraps=/nix/store/5kszpmj7j5dwzxjwzxp9r29vv644fry9-exa-0.10.1/bin/exa --wraps=exa --wraps='exa --icons' --description 'alias ls=exa'
  exa $argv

end

$ alias ll="ls -l"

$ functions ll
# Defined via `source`
function ll --wraps='/nix/store/5kszpmj7j5dwzxjwzxp9r29vv644fry9-exa-0.10.1/bin/exa -l' --wraps='ls -la' --wraps='ls -l' --description 'alias ll=ls -l'
  ls -l $argv

end

$ alias lt="exa --tree"

$ functions lt
# Defined via `source`
function lt --wraps='/nix/store/5kszpmj7j5dwzxjwzxp9r29vv644fry9-exa-0.10.1/bin/exa --tree' --wraps='exa --tree' --description 'alias lt=exa --tree'
  exa --tree $argv

end
```
Here it automatically added command to first alias, but not to any other.

**Do not** add `[exa]` to `escapeShellArgs`. Just use `exa` in the alias. Escape only user provided args. When doing `alias exa="'exa' '-l'"` (note, that `exa` is quoted) fish doesn't recognize that there will be recursion and doesn't prepend `command`. See example https://github.com/nix-community/home-manager/pull/3505#discussion_r1059963162.

modules/programs/exa.nix
```diff
@@ -23,10 +11,51 @@ in {
...
+    aliases = {
+      # Use `command` instead of hardcoding the path to exa so that aliases don't
+      # go stale after a system update.
+      exa = "command ${cmd}";
+      ls = "exa";
+      ll = "exa -l";
+      la = "exa -a";
+      lt = "exa --tree";
+      lla = "exa -la";
+    };
```
stasjok, line 57:
```suggestion
    exaArgs = optional cfg.icons "--icons"
      ++ optional cfg.git "--git" ++ cfg.extraOptions);

    aliases = {
      # Use `command` instead of hardcoding the path to exa so that aliases don't
      # go stale after a system update.
      exa = "exa ${exaArgs}";
      ls = "exa";
      ll = "exa -l";
      la = "exa -a";
      lt = "exa --tree";
      lla = "exa -la";
    };
```
Try this. Not tested.

ncfavier:
Seems to work!

ncfavier (approved):

loicreynier:
Shouldn't `alias exa="exa ${extra_options}"` be set even if `enableAliases = false`? At the moment the options have no effect without `enableAliases = true`.

ncfavier:
https://github.com/nix-community/home-manager/pull/3721

PR #3522, opened 2022-12-24
direnv: enable nushell integration

### LINKED ISSUE #3520: programs.direnv.enableNushellIntegration
enable direnv for nushell

autophagy:
### Description

This enables nushell integration by default for direnv, similar to bash/zsh/fish. The slightly verbose way of setting this is to ensure that peoples' existing nushell configuration isn't overwritten, only appended to, as would be the case if we just used the integration example from the nushell docs (https://www.nushell.sh/cookbook/direnv.html) to preserve backwards compatibility.

Closes #3520

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/direnv.nix
+++ b/modules/programs/direnv.nix
@@ -80,6 +80,15 @@ in {
       '';
     };
 
+    enableNushellIntegration = mkOption {
+      default = true;
+      type = types.bool;
+      readOnly = true;
+      description = ''
+        Whether to enable Nushell integration.
+      '';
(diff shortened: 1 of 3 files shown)


rycee:
Thanks for the contribution! Merged to master now.

happysalada:
Just a small feedback on this, after every command I type in nushell, I have
```
direnv: unloading
```
not sure if that is intended.
that is with the latest nushell 0.78

PR #3529, opened 2022-12-27
nushell: Add shellAliases option

### LINKED ISSUE #3525: home.shellAliases nushell support
### Description
Shell aliases defined in [`home.shellAliases`](https://rycee.gitlab.io/home-manager/options.html#opt-home.shellAliases) are not available in nushell sessions, even if they are valid in nushell.

autophagy:
### Description

This PR adds a `programs.nushell.shellAliases` option to allow nushell users to define shell aliases which are inserted into nushell's config.nu. It also integrates it with the general `home.shellAliases` configuration option, similar to bash/zsh/fish.

Closes #3525

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/nushell.nix
+++ b/modules/programs/nushell.nix
@@ -111,17 +111,36 @@ in {
         Additional configuration to add to the nushell environment variables file.
       '';
     };
+
+    shellAliases = mkOption {
+      type = types.attrsOf types.str;
+      default = { };
+      example = { ll = "ls -l"; };
+      description = ''
+        An attribute set that maps aliases (the top level attribute names in
+        this option) to command strings or directly to build outputs.
+      '';
+    };
   };
 
   config = mkIf cfg.enable {
     home.packages = [ cfg.package ];
+
     home.file = mkMerge [
-      (mkIf (cfg.configFile != null || cfg.extraConfig != "") {
+      (let
+        writeConfig = cfg.configFile != null || cfg.extraConfig != ""
+          || aliasesStr != "";
+
+        aliasesStr = concatStringsSep "\n"
+          (mapAttrsToList (k: v: "alias ${k} = ${v}") cfg.shellAliases);
+      in mkIf writeConfig {
         "${configDir}/config.nu".text = mkMerge [
           (mkIf (cfg.configFile != null) cfg.configFile.text)
           cfg.extraConfig
+          aliasesStr
         ];
       })
+
       (mkIf (cfg.envFile != null || cfg.extraEnv != "") {
         "${configDir}/env.nu".text = mkMerge [
           (mkIf (cfg.envFile != null) cfg.envFile.text)
--- a/tests/modules/programs/nushell/config-expected.nu
+++ b/tests/modules/programs/nushell/config-expected.nu
@@ -4,3 +4,6 @@ let $config = {
   use_ls_colors: true
 }
 
+
+alias ll = ls -a
+alias lsname = (ls | get name)
--- a/tests/modules/programs/nushell/example-settings.nix
+++ b/tests/modules/programs/nushell/example-settings.nix
@@ -15,6 +15,11 @@
     envFile.text = ''
       let-env FOO = 'BAR'
     '';
+
+    shellAliases = {
+      "lsname" = "(ls | get name)";
+      "ll" = "ls -a";
+    };
   };
 
   test.stubs.nushell = { };


teto:
seems like there might be some issues with escaping ? 
```
  × Statement used in pipeline.
    ╭─[/home/teto/.config/nushell/config.nu:14:1]
 14 │ alias c = cat
 15 │ alias dmesg = dmesg --color=always|less
    · ──┬──
    ·   ╰── not allowed in pipeline
 16 │ alias doctor = bash -c "$(curl -sSL https://doctor.novadiscovery.net/doctor)"
    ╰────
  help: 'alias' keyword is not allowed in pipeline. Use 'alias' by itself, outside of a pipeline.
  ```
  You can look how it's done in bash/zsh shells for instance

**Comment by autophagy** [comment]
> seems like there might be some issues with escaping ?
> 
> ```
>   × Statement used in pipeline.
>     ╭─[/home/teto/.config/nushell/config.nu:14:1]
>  14 │ alias c = cat
>  15 │ alias dmesg = dmesg --color=always|less
>     · ──┬──
>     ·   ╰── not allowed in pipeline
>  16 │ alias doctor = bash -c "$(curl -sSL https://doctor.novadiscovery.net/doctor)"
>     ╰────
>   help: 'alias' keyword is not allowed in pipeline. Use 'alias' by itself, outside of a pipeline.
> ```
> 
> You can look how it's done in bash/zsh shells for instance

Ah! Thank you for the catch - using `escapeShellArg` doesn't work for nushell since it basically just assigns the string to the alias:

```
heorot〉alias lsname = 'ls | get name'                              
heorot〉lsname                                          
ls | get name
```

Wrapping it in brackets [as per the docs](https://www.nushell.sh/book/aliases.html) works:
```
heorot〉alias lsname = (ls | get name)
heorot〉lsname                                          
╭────┬──────────────╮
│  0 │ LICENSE      │
│  1 │ Makefile     │
│  2 │ README.md    │
│  3 │ default.nix  │
│  4 │ docs         │
│  5 │ flake.lock   │
│  6 │ flake.nix    │
│  7 │ format       │
│  8 │ home-manager │
│  9 │ lib          │
│ 10 │ modules      │
│ 11 │ nix-darwin   │
│ 12 │ nixos        │
│ 13 │ overlay.nix  │
│ 14 │ templates    │
│ 15 │ tests        │
│ 16 │ xgettext     │
╰────┴──────────────╯
```

Which also works for aliases without pipes in them. I *don't think* special characters need to be escaped, but I'll look into it :female_detective:

**Comment by teto** [comment]
looks like the test fail https://github.com/nix-community/home-manager/actions/runs/3799465736/jobs/6462015802#step:9:2555

**Comment by autophagy** [comment]
Erk, thanks for the heads up - I forgot to update the expected config :sweat: Have updated it and added another test case with pipes, just in case.

**Comment by rycee** [comment]
I squashed this and prepared it for merge but while testing it I discovered that wrapping command by `()` doesn't seem to work as I would expect w.r.t command line arguments. For example,

``` console
/home/rycee〉alias h256 = sha256sum
/home/rycee〉h256 test.png
efbe5e25d3893639adda68f7d2471c77c6e2061f5e75646ac993f803f28d70ba  test.png
/home/rycee〉alias h256 = (sha256sum)
/home/rycee〉h256 test.png
Error: nu::parser::parse_mismatch (link)

  × Parse mismatch during operation.
   ╭─[entry #29:1:1]
 1 │ h256 test.png
   ·      ────┬───
   ·          ╰── expected operator
   ╰────

/home/rycee〉
```

autophagy:
> Maybe worth checking with some nushell people to see what they would recommend?

Yep, totally - I'll ask in the nushell discord what the recommended approach is for this and amend this PR when I have an answer. Thank you for your patience and diligence!

modules/home-environment.nix
```diff
@@ -513,6 +513,7 @@ in
     programs.bash.shellAliases = cfg.shellAliases;
     programs.zsh.shellAliases = cfg.shellAliases;
     programs.fish.shellAliases = cfg.shellAliases;
+    programs.nushell.shellAliases = cfg.shellAliases;
```
bobvanderlinden, line 516: It seems the `home.shellAliases` aren't always compatible with those from nushell.

Would it be possible to remove this and keep `programs.nushell.shellAliases`?

bobvanderlinden (commented):
Maybe revert the `(...)` change and the let the user decide whether to use `(...)` in their `programs.nushell.shellAliases`.

`home.shellAliases` might include aliases that are compatible with bash-like shells, but when using pipes it doesn't seem it can be made compatible with nushell at this time. Maybe it's better to remove `programs.nushell.shellAliases = cfg.shellAliases;` for now while still allowing `programs.nushell.shellAliases`?

modules/home-environment.nix
autophagy: Yeah, I think that'd probably make the most sense! I still didnt find a nice way of wrapping the `cfg.shellAliases` to be generally compatible. I'll make that change, and push up a commit, and ping the reviewers

rycee:
Thanks! I think having only nushell specific shell aliases sounds good. It can always be changed in the future if the situation changes. In any case, merged to master now.

bobvanderlinden:
🥳 thanks!

happysalada:
hey, this is great!
In a similar spirit, I added an attrset "envVars" to make it easier to define envVars
https://github.com/nix-community/home-manager/pull/3930
Curious to get feedback!
For me the motivating force was having some CLI apps needing some API keys that need to be secret.

PR #3531, opened 2022-12-27
mpd-mpris: init

olmokramer:
### Description

Adds a module for [mpd-mpris](https://github.com/natsukagami/mpd-mpris).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/mpd-mpris.nix
+++ b/modules/services/mpd-mpris.nix
@@ -1,0 +1,110 @@
+{ config, lib, pkgs, ... }:
+let
+  cfg = config.services.mpd-mpris;
+
+  ignoreIfLocalMpd = value: if cfg.mpd.useLocal then null else value;
+
+  renderArg = name: value:
+    if lib.isBool value && value then
+      "-${name}"
+    else if lib.isInt value then
+      "-${name} ${toString value}"
+    else if lib.isString value then
+      "-${name} ${lib.escapeShellArg value}"
+    else
+      "";
+
+  concatArgs = strings:
+    lib.concatStringsSep " " (lib.filter (s: s != "") strings);
+
+  renderArgs = args: concatArgs (lib.mapAttrsToList renderArg args);
+
+  renderCmd = pkg: args: "${pkg}/bin/mpd-mpris ${renderArgs args}";
+in {
+  meta.maintainers = [ lib.hm.maintainers.olmokramer ];
+
+  options.services.mpd-mpris = {
+    enable = lib.mkEnableOption
+      "mpd-mpris: An implementation of the MPRIS protocol for MPD";
+
+    package = lib.mkPackageOption pkgs "mpd-mpris" { };
+
+    mpd = {
+      useLocal = lib.mkOption {
+        type = lib.types.bool;
+        default = config.services.mpd.enable;
+        defaultText = lib.literalExpression "config.services.mpd.enable";
+        description = ''
+          Whether to configure for the local MPD daemon. If
+          <literal>true</literal> the <literal>network</literal>,
+          <literal>host</literal>, and <literal>port</literal>
+          settings are ignored.
+        '';
+      };
+
+      network = lib.mkOption {
+        type = with lib.types; nullOr str;
+        default = null;
+        description = ''
+          The network used to dial to the MPD server. Check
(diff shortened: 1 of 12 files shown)


modules/services/mpd-mpris.nix
```diff
@@ -0,0 +1,88 @@
...
+        default = if mpdCfg.enable then mpdCfg.network.listenAddress else null;
+        defaultText = "config.services.mpd.network.listenAddress";
+        example = "192.168.1.1";
+        description = "The address where MPD is listening for connections.";
+      };
+
+      port = lib.mkOption {
+        type = with lib.types; nullOr port;
+        default = if mpdCfg.enable then mpdCfg.network.port else null;
+        defaultText = "config.services.mpd.network.port";
```
rycee, line 44:
```suggestion
        defaultText = literalExpression "config.services.mpd.network.port";
```

modules/services/mpd-mpris.nix
```diff
@@ -0,0 +1,88 @@
...
+    enable = lib.mkEnableOption
+      "mpd-mpris: An implementation of the MPRIS protocol for MPD";
+
+    package = lib.mkPackageOption pkgs "mpd-mpris" { };
+
+    mpd = {
+      host = lib.mkOption {
+        type = with lib.types; nullOr str;
+        default = if mpdCfg.enable then mpdCfg.network.listenAddress else null;
+        defaultText = "config.services.mpd.network.listenAddress";
```
rycee, line 36:
```suggestion
        defaultText = literalExpression "config.services.mpd.network.listenAddress";
```

modules/services/mpd-mpris.nix
```diff
@@ -0,0 +1,88 @@
...
+        lib.platforms.linux)
+    ];
+
+    systemd.user.services.mpd-mpris = {
+      Install = { WantedBy = [ "default.target" ]; };
+
+      Unit = {
+        Description =
+          "mpd-mpris: An implementation of the MPRIS protocol for MPD";
+        After = [ "mpd.service" ];
```
rycee, line 72:
```suggestion
        After = [ "mpd.service" ];
        Requires = [ "mpd.service" ];
```

modules/services/mpd-mpris.nix
```diff
@@ -0,0 +1,88 @@
...
+  options.services.mpd-mpris = {
+    enable = lib.mkEnableOption
+      "mpd-mpris: An implementation of the MPRIS protocol for MPD";
+
+    package = lib.mkPackageOption pkgs "mpd-mpris" { };
+
+    mpd = {
+      host = lib.mkOption {
+        type = with lib.types; nullOr str;
+        default = if mpdCfg.enable then mpdCfg.network.listenAddress else null;
```
rycee, line 35: I think when `services.mpd` is enabled then it is probably best to simply use `localhost` as the listen address. Doing that seems to make mpd-mpris go for the MPD unix socket: https://github.com/natsukagami/mpd-mpris/blob/11fb84eb87a19965c6f1329d8552ca0ac6439f3d/cmd/mpd-mpris/main.go#L32-L43

rycee: Alternatively, perhaps the explicit options `-network unix -host %t/mpd/socket` would work?

When having `services.mpd.enabled` it should always listen to the unix socket but may not listen on a TCP socket so this should be more reliable.

olmokramer: As far as I can see that function is only called if the `-host` flag is left empty. So maybe the default values for `services.mpd-mpris.mpd.{host,config}` should just be `null`, then no `-host` or `-port` flag is generated and the automatic detection will do its thing. I tried this locally and it seems to work fine.

modules/services/mpd-mpris.nix
olmokramer: I left the `Requires` out because MPD might run on another host. Maybe it should only be added if the `services.mpd.enable = true` and `services.mpd-mpris.mpd.{host,port}` are both `null`?

olmokramer: I've added a `services.mpd-mpris.mpd.useLocal` option that defaults to `config.services.mpd.enable` and, when `true`, adds the `Required` field.

rycee:

PR #3533, opened 2022-12-28
neovim: fix extraLuaPackages type.

R-VdP:
### Description

Assigning to `programs.neovim.extraLuaPackages` a function taking a lua package set as input and returning a list of packages, as described in the documentation, threw an error because the rest of the code assumed that the value was always a plain list.
Using `lib.types.coercedTo`, we can accept such functions, as per the documentation, as well as plain lists, which we then convert to a function ignoring its input argument.
We print a warning when a plain list is assigned, since the function form is preferred, as it ensures that the right lua package set is used.

For the lua packages, we also get the lua package set from the finalPackage, to make sure that we are always using the same package set as the actual unwrapped neovim package being built.

For `programs.neovim.extraPythonPackages` I did the same.

I updated the test case so that we test both ways of setting these options.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/neovim.nix
+++ b/modules/programs/neovim.nix
@@ -13,25 +13,6 @@ let
 
   jsonFormat = pkgs.formats.json { };
 
-  extraPython3PackageType = mkOptionType {
-    name = "extra-python3-packages";
-    description = "python3 packages in python.withPackages format";
-    check = with types;
-      (x: if isFunction x then isList (x pkgs.python3Packages) else false);
-    merge = mergeOneOption;
-  };
-
-  # Currently, upstream Neovim is pinned on Lua 5.1 for LuaJIT support.
-  # This will need to be updated if Neovim ever migrates to a newer
-  # version of Lua.
-  extraLua51PackageType = mkOptionType {
-    name = "extra-lua51-packages";
-    description = "lua5.1 packages in lua5_1.withPackages format";
-    check = with types;
-      (x: if isFunction x then isList (x pkgs.lua51Packages) else false);
-    merge = mergeOneOption;
-  };
-
   pluginWithConfigType = types.submodule {
     options = {
       config = mkOption {
@@ -80,19 +61,23 @@ let
     optional = false;
   };
 
+  luaPackages = cfg.finalPackage.unwrapped.lua.pkgs;
+  resolvedExtraLuaPackages = cfg.extraLuaPackages luaPackages;
+
   extraMakeWrapperArgs = lib.optionalString (cfg.extraPackages != [ ])
     ''--suffix PATH : "${lib.makeBinPath cfg.extraPackages}"'';
-  extraMakeWrapperLuaCArgs = lib.optionalString (cfg.extraLuaPackages != [ ]) ''
-    --suffix LUA_CPATH ";" "${
-      lib.concatMapStringsSep ";" pkgs.lua51Packages.getLuaCPath
-      cfg.extraLuaPackages
-    }"'';
-  extraMakeWrapperLuaArgs = lib.optionalString (cfg.extraLuaPackages != [ ]) ''
-    --suffix LUA_PATH ";" "${
-      lib.concatMapStringsSep ";" pkgs.lua51Packages.getLuaPath
-      cfg.extraLuaPackages
-    }"'';
-
+  extraMakeWrapperLuaCArgs =
+    lib.optionalString (resolvedExtraLuaPackages != [ ]) ''
+      --suffix LUA_CPATH ";" "${
+        lib.concatMapStringsSep ";" luaPackages.getLuaCPath
+        resolvedExtraLuaPackages
+      }"'';
+  extraMakeWrapperLuaArgs = lib.optionalString (resolvedExtraLuaPackages != [ ])
+    ''
+      --suffix LUA_PATH ";" "${
+        lib.concatMapStringsSep ";" luaPackages.getLuaPath
+        resolvedExtraLuaPackages
+      }"'';
 in {
   imports = [
     (mkRemovedOptionModule [ "programs" "neovim" "withPython" ]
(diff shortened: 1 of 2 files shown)


modules/programs/neovim.nix
```diff
@@ -174,8 +159,14 @@ in {
         '';
       };
 
+      # Currently, upstream Neovim is pinned on Lua 5.1 for LuaJIT support.
+      # This will need to be updated if Neovim ever migrates to a newer
+      # version of Lua.
       extraLuaPackages = mkOption {
-        type = with types; either extraLua51PackageType (listOf package);
+        type = with types;
```
teto, line 166: rather than silently doing it, I would prefer to warn the user, make it work but with a deprecation notice.

modules/programs/neovim.nix
```diff
@@ -351,10 +342,10 @@ in {
     programs.neovim.generatedConfigViml = neovimConfig.neovimRcContent;
 
     programs.neovim.generatedConfigs = let
-      grouped = lib.lists.groupBy (x: x.type) pluginsNormalized;
+      grouped = builtins.groupBy (x: x.type) pluginsNormalized;
```
teto, line 345: lib.lists allows to target older nix as well so should be better to keep it.

modules/programs/neovim.nix
```diff
@@ -164,8 +144,13 @@ in {
       };
 
       extraPython3Packages = mkOption {
-        type = with types; either extraPython3PackageType (listOf package);
-        default = (_: [ ]);
+        # In case we get a plain list, we need to turn it into a function,
+        # as expected by the function in nixpkgs.
+        # The only way to do so is to call `const`, which will ignore its input.
+        type = with types;
```
teto, line 150: rather than silently doing it, I would prefer to warn the user, make it work but with a deprecation notice.

R-VdP:
@teto is there a consensus on the preferred type then? Do we prefer the function style to be in line with nixpkgs?

modules/programs/neovim.nix
```diff
@@ -174,8 +159,14 @@ in {
         '';
       };
 
+      # Currently, upstream Neovim is pinned on Lua 5.1 for LuaJIT support.
+      # This will need to be updated if Neovim ever migrates to a newer
+      # version of Lua.
       extraLuaPackages = mkOption {
-        type = with types; either extraLua51PackageType (listOf package);
+        type = with types;
+          let finalType = listOf package;
```
teto, line 167: There is a reason we prefer functions, it allows to work with different sets. Actually we should use the same lua as neovim-unwrapped. It should be gettable via `programs.neovim.finalPackage.neovim.lua` so one could do 
`programs.neovim.finalPackage.neovim.lua.withPackages(ls: extraLuaPackages)`

teto:
yes we prefer functions over lists. It allows to work with different sets, override the lua version. That's how it is in nixpkgs so let's do the same here.

modules/programs/neovim.nix
R-VdP: Good point, I changed this to show a warning.

modules/programs/neovim.nix
R-VdP: Ok. The one in `builtins` is a lot more efficient though, so it would be good to switch to it at some point.

modules/programs/neovim.nix
R-VdP: Yeah, agreed. I made this change now.

R-VdP:
@teto I made the suggested changes, thanks for reviewing!

modules/programs/neovim.nix
teto: it is the one used if available see https://github.com/NixOS/nixpkgs/blob/2671a557503f7462a5e9147ba728caf9dce4bff4/lib/lists.nix#L346

modules/programs/neovim.nix
```diff
@@ -80,19 +61,29 @@ let
...
+    ''
+      --suffix LUA_PATH ";" "${
+        lib.concatMapStringsSep ";" luaPackages.getLuaPath
+        resolvedExtraLuaPackages
+      }"'';
+
+  convertListToFunctionWithWarning = optionName:
+    flip warn const ''
+      Assigning a plain list to ${optionName} is deprecated.
+      Please assign a function taking a package set as argument instead.
```
teto, line 85: Can you improve the warning with an example please ? 
aka
`extraLuaPackages = [ pkgs.lua51Packages.basexx ]; should become extraLuaPackages = ls: [ ls.basexx ];`

tests/modules/programs/neovim/runtime.nix
```diff
@@ -19,13 +19,20 @@ with lib;
...
         }
       ];
+    } // lib.mkMerge [
+      {
+        extraPython3Packages = ps: with ps; [ jedi pynvim ];
+        extraLuaPackages = ps: with ps; [ luacheck ];
+      }
+      {
+        extraPython3Packages = with pkgs.python3Packages; [ jedi pynvim ];
+        extraLuaPackages = with pkgs.lua51Packages; [ luacheck ];
```
teto, line 29: what if you use a function here too ?

R-VdP: I explicitly added both ways of using the option to make sure that as long as we accept them, they both work.
To avoid a regression as the one this PR is fixing where one of the documented types is not actually accepted.

The variant where we use a function is just above.

modules/programs/neovim.nix
R-VdP: Ok, I added extra details in the description of the option, referring also to the example already provided there, and I referred to the option's documentation in the warning message, so that we do not duplicate the same information multiple times.
I think the option's description is the right place for this, so that people will see this info as soon as they look at the option, instead of discovering it only after having used the wrong type.

What do you think?

R-VdP:
Sorry, I was expecting a big red warning in my terminal if the tests failed, and didn't realise that there was a failure all the way up in the output when running all tests.
I fixed it now, the problem was just in the test code, not in the actual module.

modules/programs/neovim.nix
```diff
@@ -164,24 +156,56 @@ in {
...
-        example = literalExpression "(ps: with ps; [ luautf8 ])";
+        type = with types;
+          let fromType = listOf package;
+          in coercedTo fromType
+          (convertListToFunctionWithWarning "extraLuaPackages")
+          (functionTo fromType);
+        default = _: [ ];
+        defaultText = literalExpression "ps: [ ]";
+        example =
+          literalExpression "luaPackages: with luaPackages; [ luautf8 ]";
```
teto, line 197: the `luaPackages` may be confused with the top level one. You can use `ps` instead for `package set` or `lp` as `luaPackages` / `ls` (`lua set`)

modules/programs/neovim.nix
teto: I think it's easier if the example jumps at the user's face in the warning. Most of them dont want to look into the manual and just want to go on with their lives. Also your changes make the description quite verbose, no need to explain something that's going to be removed in the future.
e.g. I would remove
````
          For backwards compatibility, this option also accepts a plain list of
          packages, but this functionality is deprecated and might be removed in
          a future version, because doing so might cause problems when the
          specified packages are not compatible with the Python 3 version used
          by Neovim.
          ```

teto:
No problem. I was planning to merge as it but since it failed I've made a comment on the documentation. I would prefer to see the migration step in the warning rather than in the option description (it's less tedious). Also the descriptions were too long.

modules/programs/neovim.nix
R-VdP: Better like this? `pyPkgs`/`luaPkgs`

modules/programs/neovim.nix
R-VdP: Done.

modules/programs/neovim.nix
teto: great, let's merge !

teto:
thanks for the contribution and keeping up with reviews

PR #3537, opened 2022-12-29
kakoune: use str for hooks and check configuration

lobre:
### Description

This is a follow-up on https://github.com/nix-community/home-manager/pull/3418.

The aim is to turn kakoune hooks into a string instead of an enum to avoid having to update them when they change in kakoune.

And to validate that the string value is correct, the idea was to add a configuration checker in the checkPhase.

However, I struggled to correctly implement the checker as kakoune does not have a simple way to validate the configuration.

See my comment for more details: https://github.com/nix-community/home-manager/pull/3418#issuecomment-1346397799

As I am not using kakoune anymore, I am simply creating this PR to reference the WIP that has been done, in case someone wants to continue working on this.

I will be directly closing the PR, but feel free to comment if you are interested.

### DIFF
--- a/modules/programs/kakoune.nix
+++ b/modules/programs/kakoune.nix
@@ -9,49 +9,8 @@ let
   hook = types.submodule {
     options = {
       name = mkOption {
-        type = types.enum [
-          "NormalBegin"
-          "NormalIdle"
-          "NormalEnd"
-          "NormalKey"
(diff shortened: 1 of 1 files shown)


lobre:
As said above, closing. Feel free to comment if you want to pick it up.

PR #3538, opened 2022-12-29
man: use configured man-db package for generating database cache

stasjok:
### Description

**mandb** can be compiled with support for various databases. **mandb** in _nixpkgs_ is compiled with Berkeley db, but **mandb** in most distributions is compiled with gdbm. Database cache generated by _home-manager_ is `index.bt`, but all linux distributions I know are using `index.db`.

It results in a situation when **apropos** from _nixpkgs_ doesn't see a database from host OS, and **apropos** from host OS doesn't see a database from _home-manager_. **fish** man completion for example doesn't work properly (completes only for man pages from home-manager)

Before this change if you override a man package with

```nix
{
  programs.man.package = pkgs.man.override { db = pkgs.gdbm; };
}
```

you'll get an **apropos** command incompatible with a generated database (because they are compiled with support of different databases). If you override **man** package globally with an overlay, you'll get a massive amount of rebuilds.

After this change I can override a **man** package, it doesn't cause many rebuilds, and man-completion in **fish** works correctly (it uses apropos under the hood).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


rycee:
Thanks! Merged to master now.

PR #3539, opened 2022-12-29
fish: source hm-session-vars.sh only for interactive shells

stasjok:
### Description

Unlike bash **fish** executes all commands from `config.fish`. Bash executes `.bash_profile` or `.profile` for login shells and `.bashrc` for interactive shells. In my opinion environment variables should be set only for login shells. Simple `fish` command shouldn't change global variables like `PATH`. Right now in bash `hm-session-vars.sh` is sourced in `.profile` (login shells). Also it is sourced in `.bashrc`, but only when `targets.genericLinux.enable` is enabled:

https://github.com/nix-community/home-manager/blob/cbacdaba3c7b361defb36e1cdfa03ae4e74eb4a8/modules/targets/generic-linux.nix#L66-L71

The linked comment mentions a tmux issue, but it's not the case at least for modern tmux: in tmux by default shell is a login shell.

If you are wary that it won't be sourced in some cases, we can change condition to `status is-interactive` instead.

I'm pretty sure that one of the conditions (is-login or is-interactive) are desirable, because commands like `fish -c '<something>'` should be as quick as possible, because many tools may need to run some shell commands and they are using `$SHELL` variable for it. Right now in _home-manager_ `fish -c exit` is pretty slow (30 ms vs 8 ms for me) because **fenv** is slow (it runs external commands like bash/sed/tr multiple times).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -354,9 +354,11 @@ in {
         set -q __fish_home_manager_config_sourced; and exit
         set -g __fish_home_manager_config_sourced 1
 
-        set --prepend fish_function_path ${pkgs.fishPlugins.foreign-env}/share/fish/vendor_functions.d
-        fenv source ${config.home.profileDirectory}/etc/profile.d/hm-session-vars.sh > /dev/null
-        set -e fish_function_path[1]
+        if status is-interactive
+          set --prepend fish_function_path ${pkgs.fishPlugins.foreign-env}/share/fish/vendor_functions.d
+          fenv source ${config.home.profileDirectory}/etc/profile.d/hm-session-vars.sh > /dev/null
+          set -e fish_function_path[1]
+        end
 
         ${cfg.shellInit}
 


stasjok:
Even though it works perfectly well for me only for login shells, I changed it to interactive shells instead. I think running shell in graphical environment often will not be a login shell. With `is-interactive` it should be backwards compatible with previous behavior in practical scenarios.

jian-lin:
IIUC, the issue you deal with is the slowdown of setting env vars at shell initialization caused by `fenv`.

Your approach is to only set those env vars for interactive or login shells, which may cause those env vars not being set when needed, e.g. non-login and interactive shells started by gnome-terminal and non-login and non-interactive shells for running commands through ssh as described in my related [pr][1] for zsh.

I propose a better approach: replace `fenv` with `babelfish`. `babelfish` can translate shell scripts to native fish scripts, which can then be sourced. See its [pr][2] in nixpkgs for more info. In this way, the slowdown issue is solved and commands setting env var need not be guarded by `is-login` or `is-interactive` because those commands will [immediately return if not needed][3].

[1]: https://github.com/nix-community/home-manager/pull/2708
[2]: https://github.com/NixOS/nixpkgs/pull/108947
[3]: https://github.com/nix-community/home-manager/blob/275ab728912006eecb549338a50f24f294a7cfb7/modules/home-environment.nix#L553

stasjok:
OK, then.

PR #1769, opened 2021-01-31
home-environment: make sessionVariables a freeform module

rycee:
### Description

Also introduce explicit options for `PATH` and `NIX_PATH`.

### Checklist

- [X] Change is backwards compatible: Mostly, it does limit session variable values to `package`, `str`, `int`, and `float` but I think this covers pretty much all use-cases.

- [X] Code formatted with `./format`.

- [X] Code tested through `nix-shell --pure tests -A run.all`.

- [X] Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).

- [X] Commit messages are formatted like

    ```
    {component}: {description}

    {long description}
    ```

    See [CONTRIBUTING](https://github.com/nix-community/home-manager/blob/master/doc/contributing.adoc#sec-commit-style) for more information and [recent commit messages](https://github.com/nix-community/home-manager/commits/master) for examples.

PR #3541, opened 2022-12-29
home-environment: use `lazyAttrsOf` for `home.sessionVariables`

ncfavier:
`attrs` has unreasonable merge semantics and is deprecated. `attrsOf` doesn't support variables depending on each other as is recommended in the option's description.

Context:
- https://github.com/NixOS/nixpkgs/pull/97119
- https://github.com/NixOS/nixpkgs/pull/70138
- https://github.com/nix-community/home-manager/pull/1595 (cc @berbiche)
- https://github.com/nix-community/home-manager/pull/1769 
- https://github.com/nix-community/home-manager/pull/3496#issuecomment-1367534054

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/home-environment.nix
+++ b/modules/home-environment.nix
@@ -256,7 +256,7 @@ in
 
     home.sessionVariables = mkOption {
       default = {};
-      type = types.lazyAttrsOf types.anything;
+      type = with types; lazyAttrsOf (oneOf [ str path int float ]);
       example = { EDITOR = "emacs"; GS_OPTIONS = "-sPAPERSIZE=a4"; };
       description = ''
         Environment variables to always set at login.
--- a/modules/home-environment.nix
+++ b/modules/home-environment.nix
@@ -256,7 +256,7 @@ in
 
     home.sessionVariables = mkOption {
       default = {};
-      type = types.attrs;
+      type = types.lazyAttrsOf types.anything;
       example = { EDITOR = "emacs"; GS_OPTIONS = "-sPAPERSIZE=a4"; };
       description = ''
         Environment variables to always set at login.


modules/home-environment.nix
```diff
@@ -256,7 +256,7 @@ in
 
     home.sessionVariables = mkOption {
       default = {};
-      type = types.attrs;
+      type = types.lazyAttrsOf types.anything;
```
teto, line 259: types.anything ? not str/int ?  seems like types.attrs didn't enforce anything either so that's probably fine.

berbiche: Has `types.raw` or `types.unconditional` been merged? I believe they are substitutes for `types.anything`

ncfavier: `unconditional` has to do with moving the `lazy` to the element type instead of the attribute type, it's mostly an implementation detail.

`raw` is for when absolutely no processing is wanted. I think this would disable `mkOverride`, and would also not use `mergeEqualOption` so that multiple equal definitions would raise an error, which seems less nice.

But yes we really want a more precise type. Anything that can be `toString`ed was previously accepted, but maybe we should restrict that to `oneOf [ str path int float ]` or something.

ncfavier: The next step would be to make that the `freeformType` of a submodule, which is the core of https://github.com/nix-community/home-manager/pull/1769, so that more precise options (e.g. concatenating lists with `:` for PATH) can be added later.

teto (approved):

modules/home-environment.nix
teto: sounds awesome.

davidsierradz:
Hi, according to https://github.com/direnv/direnv/issues/203, previously I could declare an empty environment variable as:

```nix
  home.sessionVariables = {
    DIRENV_LOG_FORMAT = null;
  };
```

Right now, when trying to build my configuration, the following error happens:

```
error: A definition for option `home-manager.users.neuromante.home.sessionVariables.DIRENV_LOG_FORMAT' is not of type `string or path or signed integer or floating point number'.
```

More generally, how can I express an empty variable, so we can test it with `-n STRING`?

Thanks.

ncfavier:
`DIRENV_LOG_FORMAT = "";`

davidsierradz:
Thanks, that makes sense! :joy:

willemml:
Is there any reason lists of strings are not allowed? (e.g. for the `EDITOR` variable I had set `[ "emacsclient" "-nw" "-c" ]` which no longer works.)

ncfavier:
Lists of strings are ambiguous because different variables may need to be merged differently (e.g. for `PATH` you'd want colons). Instead you can just do `EDITOR = "emacsclient -nw -c";` which means the same thing.

willemml:
Hmm, I thought you could do argument lists in bash as well, I did end up just using `EDITOR = "emacsclient -nw -c";` though.

ncfavier:
Both things get translated to the same bash code because we use `toString`.

willemml:
Oh ok, thanks for the clarification 👍

Issue #3543, opened 2022-12-30
`mkAliasOptionModule` descriptions are not shown in the manual


phpcitizen:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

xsession.pointerCursor options definitions needs to be corrected at https://rycee.gitlab.io/home-manager/options.html.

### System information

```markdown
Not important.
```

ncfavier:
They're [aliases](https://github.com/nix-community/home-manager/blob/dd99675ee81fef051809bc87d67eb07f5ba022e8/modules/config/home-cursor.nix#L59-L79) to the new `home.pointerCursor` options.

The underlying issue here is that nmd does not support [Markdown descriptions](https://github.com/NixOS/nixpkgs/issues/175586), which are used by `mkAliasOptionModule` and friends.

rycee:
In this case I think these options need `visible = false` to make them not show up in the manual at all.

ncfavier:
Beware the difference between `mkRenamedOptionModule` and `mkAliasOptionModule`: the former have `visible = false` while the latter have a helpful description. Maybe we just want to use `mkRenamedOptionModule` everywhere?

ncfavier:
Fixed by https://github.com/nix-community/home-manager/pull/3570

PR #3545, opened 2022-12-31
home-cursor: use `mkRenamedOptionModule` for `xsession.pointerCursor`

ncfavier:
This hides the obsolete options from the manual, at the cost of displaying two warnings: the one from `mkRenamedOptionModule` and the generic migration instructions for `xsession.pointerCursor`.

Fixes (superficially) https://github.com/nix-community/home-manager/issues/3543, though it does not address the underlying [issue](https://github.com/NixOS/nixpkgs/issues/175586#issuecomment-1368012288) with `mkAliasOptionModule`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/config/home-cursor.nix
+++ b/modules/config/home-cursor.nix
@@ -30,7 +30,10 @@ let
       x11 = {
         enable = mkEnableOption ''
           x11 config generation for <option>home.pointerCursor</option>
-        '';
+        '' // {
+          default = config.xsession.enable;
+          defaultText = literalExpression "config.xsession.enable";
+        };
 
         defaultCursor = mkOption {
           type = types.str;
@@ -43,7 +46,10 @@ let
       gtk = {
         enable = mkEnableOption ''
           gtk config generation for <option>home.pointerCursor</option>
-        '';
+        '' // {
+          default = config.gtk.enable;
+          defaultText = literalExpression "config.gtk.enable";
+        };
       };
     };
   };
@@ -56,38 +62,27 @@ in {
   meta.maintainers = [ maintainers.polykernel maintainers.league ];
 
   imports = [
-    (mkAliasOptionModule [ "xsession" "pointerCursor" "package" ] [
+    (mkRenamedOptionModule [ "xsession" "pointerCursor" "package" ] [
       "home"
       "pointerCursor"
       "package"
     ])
-    (mkAliasOptionModule [ "xsession" "pointerCursor" "name" ] [
+    (mkRenamedOptionModule [ "xsession" "pointerCursor" "name" ] [
       "home"
       "pointerCursor"
       "name"
     ])
-    (mkAliasOptionModule [ "xsession" "pointerCursor" "size" ] [
+    (mkRenamedOptionModule [ "xsession" "pointerCursor" "size" ] [
       "home"
       "pointerCursor"
       "size"
     ])
-    (mkAliasOptionModule [ "xsession" "pointerCursor" "defaultCursor" ] [
+    (mkRenamedOptionModule [ "xsession" "pointerCursor" "defaultCursor" ] [
       "home"
       "pointerCursor"
       "x11"
       "defaultCursor"
     ])
-
-    ({ ... }: {
-      warnings = optional (any (x:
-        getAttrFromPath
-        ([ "xsession" "pointerCursor" ] ++ [ x ] ++ [ "isDefined" ])
-        options) [ "package" "name" "size" "defaultCursor" ]) ''
-          The option `xsession.pointerCursor` has been merged into `home.pointerCursor` and will be removed
-          in the future. Please change to set `home.pointerCursor` directly and enable `home.pointerCursor.x11.enable`
-          to generate x11 specific cursor configurations. You can refer to the documentation for more details.
-        '';
-    })
   ];
 
   options = {


modules/config/home-cursor.nix
```diff
@@ -56,22 +56,22 @@ in {
...
+    (mkRenamedOptionModule [ "xsession" "pointerCursor" "size" ] [
       "home"
       "pointerCursor"
       "size"
     ])
-    (mkAliasOptionModule [ "xsession" "pointerCursor" "defaultCursor" ] [
+    (mkRenamedOptionModule [ "xsession" "pointerCursor" "defaultCursor" ] [
       "home"
       "pointerCursor"
       "x11"
```
rycee, line 77: Unfortunately, in this case `mkAliasOptionModule` was used for a reason. If you look at the list entry below the fold in GitHub (click the ⬇) you'll notice that there is actually already a more helpful warning printed. The main difference is that it includes a reference to `home.pointerCursor.x11.enable`, which would be missing if we use `mkRenamedOptionModule`.

That said, I think we could get away by changing the warning text below slightly. Something simple like

> To enable X11 specific cursor configurations, please enable 'home.pointerCursor.x11.enable'.

may be sufficient. The other options will get their own warnings through `mkRenamedOptionModule`.

rycee: That said², I'm not certain why the `home.pointerCursor.x11.enable` option is there at all. Presumably the user would like to use the cursor settings whenever `config.xsession.enable` is set? @polykernel Can you recall why that is?

ncfavier: I did see that, that's why I said "at the cost of getting two warnings". I don't think it's too bad, although it might need a change in wording.

Probably a better solution would be to make `mkRenamedOptionModule` accept a "replacement instructions" message, just like `mkRemovedOptionModule` does.

But if we can just remove `home.pointerCursor.x11.enable` that's even simpler.

ncfavier: Or (assuming we do need `home.pointerCursor.x11.enable`) maybe we should indeed use `mkRemovedOptionModule` and fail instead of accepting a configuration that suddenly does not have X cursors anymore.

polykernel: > That said², I'm not certain why the `home.pointerCursor.x11.enable` option is there at all. Presumably the user would like to use the cursor settings whenever `config.xsession.enable` is set? @polykernel Can you recall why that is?

IIRC, the `enable` option was introduced as a toggle for each backend in case more options are added for the backend in the future.

After some thought, I think each `home.pointerCursor.<backend>` can also be a nullable submodule which achieves the same functionality as the `enable` option.

ncfavier: That seems orthogonal to what we're discussing here: if we used a nullable submodule, we would still have to translate the old `xsession.pointerCursor.*` into two separate settings: `home.pointerCursor.*` and `home.pointerCursor.x11 = whatever`.

The question is whether we can instead just enable the X backend whenever X itself (in this case xsession) is enabled.

polykernel: Sorry, I misread the comment thread and thought the point of discussion was around the double warning. In which case, using a nullable submodule would allow `mkRenamedOptionModule` to be used as is for `home.pointerCursor.x11.defaultCursor`, and the warning defined below can be removed.

> we would still have to translate the old `xsession.pointerCursor.*` into two separate settings: `home.pointerCursor.*` and `home.pointerCursor.x11 = whatever`.

I see your point, `defaultCursor` is defined under the X backend because I initially thought it was X specific, but the line is somewhat blurred with some Wayland application also using X configurations. If needed, `defaultCursor` can be moved under `home.pointerCursor` so the old `xsession.pointerCursor.*` directly maps to `home.pointerCursor.*`.

That said, enabling the X backend based on whether xsession is enabled is a better idea. This can be done by either removing `home.pointerCursor.x11.enable` entirely and using `xsession.enable` in the mkIf or setting the default value of `home.pointerCursor.x11.enable` to `xsession.enable`.

ncfavier: If you can submit a PR for that, that would be great! Similarly, I think the GTK backend can simply be enabled if `gtk.enable` is true. Then both `enable` options should get a `mkRemovedOptionModule` explaining why they've been removed.

polykernel: When drafting the PR, I noticed that it is not possible to maintain compatibility with the old `xsession.pointerCursor` module after the removal of `x11.enable`. Since the `xsession.initExtra` and `xresources.properties` settings become dependent on `xsession.enable`, this means it is not possible to enable the full X cursor configuration in isolation anymore which was supported by `xsession.pointerCursor`. With the `gtk.enable` option, the same type of issue arise because of coupled dependencies. For example, if a user wants to enable X cursor configurations via `home.pointerCursor` and has `gtk.enable` set to true only for gtk theme configuration, then the gtk cursor configuration also gets set by the `home.pointerCursor` unbeknownst to the user.

Is the incompatibility introduced acceptable for the module? In any case, I have submitted a draft PR as a work in progress.

ncfavier: Maybe it would be much simpler to keep the options, but make `x11.enable` *default* to `config.xsession.enable` and `gtk.enable` *default* to `config.gtk.enable`. That way we don't lose backwards compatibility (although those use cases seem questionable to me) and we don't have to bother with the whole "`mkRemovedOptionModule` in submodules" situation.

polykernel: I agree. Would you be able to append the default value changes to this PR or should I make another PR?

ncfavier: Done

polykernel (approved):

ncfavier:
I guess we can abandon this now that markdown descriptions are rendered fine.

Issue #3551, opened 2023-01-01
bug: error: The option `home.sessionVariables.XCURSOR_PATH' has conflicting definition values:


DanielSiepmann:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

I've run `nix-channel --update` after the holidays followed by `home-manager switch`. This ends in the following output:

```
~ > home-manager switch
error: The option `home.sessionVariables.XCURSOR_PATH' has conflicting definition values:
       - In `/home/daniels/.nix-defexpr/channels/home-manager/modules/targets/generic-linux.nix': "$XCURSOR_PATH\${XCURSOR_PATH:+:}/home/daniels/.nix-profile/share/icons:/usr/share/icons:/usr/share/pixmaps"
       - In `/home/daniels/.nix-defexpr/channels/home-manager/modules/config/home-cursor.nix': "$XCURSOR_PATH\${XCURSOR_PATH:+:}/home/daniels/.nix-profile/share/icons"
(use '--show-trace' to show detailed location information)
```

### Maintainer CC

@polykernel, @league

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 5.4.0-135-generic, Ubuntu, 20.04.5 LTS (Focal Fossa), nobuild`
 - multi-user?: `no`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.12.0`
 - channels(daniels): `"home-manager, nixpkgs, phps"`
 - nixpkgs: `/home/daniels/.nix-defexpr/channels/nixpkgs`
```

polykernel:
My apologies, this is caused by https://github.com/nix-community/home-manager/pull/2902 where I forgot to set `home.sessionVariables` with a higher priority.

DanielSiepmann:
Thanks @polykernel, I'm still new to nix and the ecosystem. Don't know why this appears now as those changes are already old.
Thanks for fast response and providing a fix.

teto:
the merging strategy for sessionVariables has recently been changed to detect this kind of conflict.

PR #3553, opened 2023-01-02
home.pointerCursor: use mkDefault to set XCURSOR_PATH

polykernel:
This commit fixes conflicting definitions of the XCURSOR_PATH environment variable with the same priority under `home.sessionVariables` introduced in https://github.com/nix-community/home-manager/pull/2902. Rather than raising the priority of the XCURSOR_PATH set in `targets.genericLinux`, the priority of the environment variable set in `home.pointerCursor` is lowered via mkDefault. This is done to avoid the `target.genericLinux` module overriding user set values silently.

### Description

fixes https://github.com/nix-community/home-manager/issues/3551

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/config/home-cursor.nix
+++ b/modules/config/home-cursor.nix
@@ -131,8 +131,8 @@ in {
       # https://github.com/nix-community/home-manager/issues/2812
       # https://wiki.archlinux.org/title/Cursor_themes#Environment_variable
       home.sessionVariables = {
-        XCURSOR_PATH = "$XCURSOR_PATH\${XCURSOR_PATH:+:}"
-          + "${config.home.profileDirectory}/share/icons";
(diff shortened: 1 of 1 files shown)


DanielSiepmann:
I've tested this change by changing the channel: `https://github.com/polykernel/home-manager/archive/refs/heads/pointercursor-xcursor-path-patch1.tar.gz home-manager`. It works.

Is there a better approach for testing? E.g. can I point a channel to a local folder? Couldn't find anything in that regard.

ncfavier:
I guess in this case it's fine for the `genericLinux` value to override the `pointerCursor` one since it's a strict superset.

@DanielSiepmann can't you just do `-I home-manager=/path/to/home-manager` ? (Take this with a grain of salt, I don't use channels)

ncfavier (approved):

PR #3558, opened 2023-01-03
home.pointerCursor: remove backend specific `enable` options

polykernel:
This PR removes the `home.pointerCursor.<backend>.enable` options for each backend in favour of using global options to toggle backend specific cursor configurations. This change was motivated by https://github.com/nix-community/home-manager/pull/3545 and improves integration of `home.pointerCursor` with other home-manager modules.

### Description

cc: @rycee @ncfavier 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/config/home-cursor.nix
+++ b/modules/config/home-cursor.nix
@@ -28,22 +28,12 @@ let
       };
 
       x11 = {
-        enable = mkEnableOption ''
-          x11 config generation for <option>home.pointerCursor</option>
-        '';
-
         defaultCursor = mkOption {
           type = types.str;
           default = "left_ptr";
           example = "X_cursor";
           description = "The default cursor file to use within the package.";
         };
-      };
-
-      gtk = {
-        enable = mkEnableOption ''
-          gtk config generation for <option>home.pointerCursor</option>
-        '';
       };
     };
   };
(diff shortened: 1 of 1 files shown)


polykernel:
I wasn't able to use `mkRemovedOptionModule` to add warnings for `home.pointerCursor.x11.enable` and `home.pointerCursor.gtk.enable` as per https://github.com/nix-community/home-manager/pull/3545#discussion_r1059930951 because `mkRemovedOptionModule` sets a pseudo option as an option tree via `setAttrByPath` but part of the attr path (i.e `x11.enable` and `gtk.enable`) is inside the submodule which causes a conflict definition error.

I am not really sure how to resolve this error, perhaps using `getSubOptions` to get an option tree representation to operate on?

ncfavier:
The type of `home.pointerCursor` isn't `submodule`, it's `null or submodule`, so [this check](https://github.com/NixOS/nixpkgs/blob/d4bad6a35a73cad2d7f93a1255b5933305ac499a/lib/modules.nix#L597) fails, so the option doesn't get merged into the submodule.

Ideally we could add `imports = [ (mkRemovedOption ["x11" "enable"] "...") ];` to the submodule itself, but that won't work yet because we don't have [submodule assertions](https://github.com/NixOS/nixpkgs/pull/97023) (or https://github.com/NixOS/nixpkgs/pull/207187). cc @infinisil @roberth

In the meantime, we'll probably have to use the same hack as in [`xdg-desktop-entries.nix`](https://github.com/nix-community/home-manager/blob/master/modules/misc/xdg-desktop-entries.nix): add an `assertions` option to the submodule and extract the assertions into the top-level assertions.

polykernel:
I am closing this PR in favor of the approach taken in https://github.com/nix-community/home-manager/pull/3545 which avoids introducing backward incompatibility and the `mkRemovedOptionModule` issue across submodule boundaries (see the linked PR for more details).

PR #3563, opened 2023-01-04
i3: remove i3/i3-gaps distinction

GaetanLepage:
### Description

[i3 4.22 has been released](https://i3wm.org/downloads/RELEASE-NOTES-4.22.txt). The main change is the merge of i3-gaps features.
This (now merged) [PR](https://github.com/NixOS/nixpkgs/pull/208861) brings this update to nixpkgs/NixOS.

Hence, we should update the `services.window-managers.i3` module by removing all the "`i3` vs `i3-gaps`" distinctions.

My only interrogation is the default value of the [`i3.config.floating.titlebar`](https://nix-community.github.io/home-manager/options.html#opt-xsession.windowManager.i3.config.floating.titlebar) and [`i3.config.window.titlebar`](https://nix-community.github.io/home-manager/options.html#opt-xsession.windowManager.i3.config.window.titlebar) options.
Before, there were set to `false` by default when using `i3-gaps`. The i3 documentation suggests that setting them to `true` is still valid so this is what I have done... **TO BE TESTED**
I didn't updated the test cases in consequence. Hence, some are failing because of this change of default value.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/window-managers/i3-sway/lib/options.nix
+++ b/modules/services/window-managers/i3-sway/lib/options.nix
@@ -1,11 +1,12 @@
-{ config, lib, moduleName, cfg, pkgs, capitalModuleName ? moduleName
-, isGaps ? true }:
+{ config, lib, moduleName, cfg, pkgs, capitalModuleName ? moduleName }:
 
 with lib;
 
 let
   isI3 = moduleName == "i3";
   isSway = !isI3;
+
+  inherit (config.home) stateVersion;
 
   fontOptions = types.submodule {
     options = {
@@ -77,7 +78,7 @@ let
 
   barModule = types.submodule {
     options = let
-      versionAtLeast2009 = versionAtLeast config.home.stateVersion "20.09";
+      versionAtLeast2009 = versionAtLeast stateVersion "20.09";
       mkNullableOption = { type, default, ... }@args:
         mkOption (args // {
           type = types.nullOr type;
@@ -158,7 +159,7 @@ let
         defaultText = "i3bar";
         description = "Command that will be used to start a bar.";
         example = if isI3 then
-          "\${pkgs.i3-gaps}/bin/i3bar -t"
+          "\${pkgs.i3}/bin/i3bar -t"
         else
           "\${pkgs.waybar}/bin/waybar";
       };
@@ -389,11 +390,17 @@ in {
       options = {
         titlebar = mkOption {
           type = types.bool;
-          default = !isGaps;
-          defaultText = if isI3 then
-            "xsession.windowManager.i3.package != nixpkgs.i3-gaps (titlebar should be disabled for i3-gaps)"
+          default = if versionOlder stateVersion "23.05" then
+            (isI3 && (cfg.config.gaps == null))
           else
-            "false";
+            true;
+          defaultText = if isI3 then ''
+            true for state version ≥ 23.05
+            config.gaps == null for state version < 23.05
+          '' else ''
+            true for state version ≥ 23.05
+            false for state version < 23.05
+          '';
           description = "Whether to show window titlebars.";
         };
 
@@ -432,11 +439,17 @@ in {
       options = {
         titlebar = mkOption {
           type = types.bool;
-          default = !isGaps;
-          defaultText = if isI3 then
-            "xsession.windowManager.i3.package != nixpkgs.i3-gaps (titlebar should be disabled for i3-gaps)"
+          default = if versionOlder stateVersion "23.05" then
+            (isI3 && (cfg.config.gaps == null))
           else
-            "false";
+            true;
+          defaultText = if isI3 then ''
+            true for state version ≥ 23.05
+            config.gaps == null for state version < 23.05
+          '' else ''
+            true for state version ≥ 23.05
+            false for state version < 23.05
+          '';
           description = "Whether to show floating window titlebars.";
         };
 
@@ -670,7 +683,7 @@ in {
 
   bars = mkOption {
     type = types.listOf barModule;
-    default = if versionAtLeast config.home.stateVersion "20.09" then [{
+    default = if versionAtLeast stateVersion "20.09" then [{
       mode = "dock";
       hiddenState = "hide";
       position = "bottom";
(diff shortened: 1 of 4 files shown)


teto:
I can test once you fix the test

GaetanLepage:
> I can test once you fix the test

Thank you @teto for your proposition. Actually, I would like to be sure about what would be the best default to set.
Is there a technical reason to not use `default_[floating_]border normal` when using gaps ?
Or is it more of a sensible default ?

My guess would be to either set what i3 is defaulting to (i.e. `normal`) or set the option default value to `null` and in this case not write anything in the config.

I am not an i3 user anymore. It would be great to have feedback from actual experienced users :)

uvNikita:
I think I found the reason for these defaults on i3-gaps: https://github.com/Airblader/i3/blob/gaps-next/README.md#gaps

GaetanLepage:
Thank you for finding the reference !
I don't know what is the current situation since the 4.22 release... I will change the tests to let @teto test the behavior.

GaetanLepage:
Actually, the sway tests are failing, because initially, the `default_[floating_]border` options were set by default to:

- true if the package was i3
- false if the package was i3-gaps **or sway**

And I actually have no idea why. Do you know why this is the case @alexarice ?

alexarice:
Sway has always had the equivalent of the `i3-gaps` feature so the default option for sway was probably copied from the default option for `i3-gaps`. I can't remember a specific reason for it.

As a side note, I believe in the past when making changes like this we've hidden them behind a `stateVersion` change. I'm not sure if this is still the way of doing things

GaetanLepage:
All the tests are passing.
It could be nice if some i3/i3-gaps/sway users could test this update (@teto).
Anyway, I think it should be ok to merge.

@rycee

alexarice:
Changing the border should at least have a changelog/news item so that people can find out why they suddenly have titlebars

GaetanLepage:
> Changing the border should at least have a changelog/news item so that people can find out why they suddenly have titlebars

Yes, it seems logical to me. Where should I mention this change ?

alexarice:
I haven't changed anything in home manager for a while so I'm not sure how things are done at the moment, probably best to wait for someone who knows what they are talking about to input.

olmokramer:
@GaetanLepage news items go in [`modules/misc/news.nix`](https://github.com/nix-community/home-manager/blob/master/modules/misc/news.nix#L113).

modules/services/window-managers/i3-sway/lib/options.nix
```diff
@@ -389,11 +388,7 @@ in {
       options = {
         titlebar = mkOption {
           type = types.bool;
-          default = !isGaps;
-          defaultText = if isI3 then
-            "xsession.windowManager.i3.package != nixpkgs.i3-gaps (titlebar should be disabled for i3-gaps)"
-          else
-            "false";
+          default = true;
```
teto, line 391: this seems to change the default for sway as is shown in the test. Is that because the default in sway/i3 changed ? otherwise remove "default" probably ?

GaetanLepage: This was what I was wondering about previously.
After checking in the upstream projects code, the default value for those options is indeed `normal` (and not `pixel`) for both [`i3`](https://github.com/i3/i3/blob/9c8746c00fee18561e2a0a039d9b1c7387b848c5/src/config.c#L212-L213) and [`sway`](https://github.com/swaywm/sway/blob/88c17ece3b2617762c3f5f1c1fa79fe8f4df0082/sway/config.c#L294-L295).
The author of the sway module [confirmed that at the time he just wanted to copy what was done for `i3-gaps`](https://github.com/nix-community/home-manager/pull/3563#issuecomment-1373460575).

Setting `titlebar` to `true` is consistent with the default value used in `i3` and `sway`, i.e. `default_[floating_]border normal`.

GaetanLepage:
> @GaetanLepage news items go in [`modules/misc/news.nix`](https://github.com/nix-community/home-manager/blob/master/modules/misc/news.nix#L113).

I added an entry in `news.nix` :)

modules/services/window-managers/i3-sway/lib/options.nix
teto: I completely ignored your previous posts very sorry. One bad example of context switching :( if normal is the default then yes we should have that.  https://github.com/nix-community/home-manager/pull/3563#issuecomment-1373460575 comments on using stateVersion and this seems like a good idea as it's easy to do. Pair it with a docs/release-notes/rl-2305.adoc entry (which can replace/complete the news.nix change)

GaetanLepage: All right ! How should I handle the `stateVersion` check ?
If `stateVersion >= 23.05`, then put `*titlebar` settings default to `true` with `sway`, else `false` ?

teto: yes

GaetanLepage: Which one seems the best to you ?

1. `default = !((versionOlder stateVersion "23.05") and isSway);`
2. `default = (versionAtLeast stateVersion "23.05") or isI3;`

alexarice: Neither preserve the original behaviour for people who had gaps activated right?

GaetanLepage: You are right. If we go for preserving the previous behavior, let's do it seriously.
Latest push fixes that:

```
default = if versionOlder stateVersion "23.05"
  then (isI3 and (cfg.config.gaps == null))
  else true
;
```

GaetanLepage: The tests are failing. What is the `stateVersion` used in the tests ?

GaetanLepage: It seems that getting the stateVersion here:
```
  inherit (config.home) stateVersion;
```
Leads to the crash of `nix-shell . -A install`.
Other modules seem to do it succesfully...

It works if I hardcode it:
```
stateVersion = "20.09";
```

GaetanLepage: Wow, this error was not easy to fix because the message was really misleading.
What caused the problem was adding a dependence on the `stateVersion` without providing a `defaultText`.
It was not obvious to me that the lack of `defaultText` lead to this error during `nix-shell . -A install`.

Anyway, I think everything is now correct.

PR #3568, opened 2023-01-05
easyeffects: add package option

pltanton:
### Description

Add package option to `service.easyeffects'.

There is some bugs in package, that you may want to fix by overriding. F.e. https://github.com/NixOS/nixpkgs/issues/189481

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/easyeffects.nix
+++ b/modules/services/easyeffects.nix
@@ -20,6 +20,13 @@ in {
       </programlisting>
       to your system configuration for the daemon to work correctly'';
 
+    package = mkOption {
+      type = types.package;
+      default = pkgs.easyeffects;
+      defaultText = literalExpression "pkgs.easyeffects";
+      description = "The <literal>easyeffects</literal> package to use.";
+    };
+
     preset = mkOption {
       type = types.str;
       default = "";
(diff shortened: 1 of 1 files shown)


modules/services/easyeffects.nix
```diff
@@ -52,7 +59,7 @@ in {
 
       Service = {
         ExecStart =
-          "${pkgs.easyeffects}/bin/easyeffects --gapplication-service ${presetOpts}";
+          "${cfg.package}/bin/easyeffects --gapplication-service ${presetOpts}";
         ExecStop = "${pkgs.easyeffects}/bin/easyeffects --quit";
```
sumnerevans, line 63:
```suggestion
        ExecStop = "${cfg.package}/bin/easyeffects --quit";
```

sumnerevans (changes requested):

modules/services/easyeffects.nix
pltanton: Missed it. Fixed.

sumnerevans (approved):

felschr:
Can this be backported to 22.11? Would be useful to try the new major easyeffects version on the stable NixOS channel.

PR #3570, opened 2023-01-06
docs: bump nmd

rycee:
### Description

- Introduces support for Markdown documentation in modules.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -7,8 +7,8 @@ let
 
   nmdSrc = fetchTarball {
     url =
-      "https://gitlab.com/api/v4/projects/rycee%2Fnmd/repository/archive.tar.gz?sha=fb9cf8e991487c6923f3c654b8ae51b6f0f205ce";
-    sha256 = "05233fy7mnn1l7ylwdmmig4ny1kwk3a2cs740gnbs0d5p28kas2r";
+      "https://gitlab.com/api/v4/projects/rycee%2Fnmd/repository/archive.tar.gz?sha=0a41b6979f6ce238c5e3795d63fdaa15f650d839";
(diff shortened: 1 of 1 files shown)


oxalica:
This introduces IFD on https://gitlab.com/rycee/nmd/-/blob/master/lib/modules-doc.nix#L63, and breaks the evaluation where IFD is disabled.

ncfavier:
Also the conversion script doesn't seem to support RST roles, so alias options look awkward: https://nix-community.github.io/home-manager/options.html#opt-xsession.pointerCursor.package

Will try to fix this later

ncfavier:
> This introduces IFD on [gitlab.com/rycee/nmd/-/blob/master/lib/modules-doc.nix#L63](https://gitlab.com/rycee/nmd/-/blob/master/lib/modules-doc.nix#L63), and breaks the evaluation where IFD is disabled.

PR'd a fix https://gitlab.com/rycee/nmd/-/merge_requests/9

ncfavier:
> Also the conversion script doesn't seem to support MyST roles, so alias options look awkward: [nix-community.github.io/home-manager/options.html#opt-xsession.pointerCursor.package](https://nix-community.github.io/home-manager/options.html#opt-xsession.pointerCursor.package)

Actually it supports them fine, the problem is that https://github.com/NixOS/nixpkgs/pull/208407 changes `mkAliasOptionModule` to not use `mdDoc` without actually changing the description. cc @pennae, we should make `mkAliasOptionModule` emit an actual DocBook description.

But I guess Home Manager can use `mkAliasOptionModuleMD` now that we support Markdown descriptions.

PR #3580, opened 2023-01-08
firefox: support passing any json value to settings

kira-bruneau:
### Description

Firefox internally only supports bool, int, and string types for preferences, but often stores objects, arrays and floats as strings.

This change makes it nicer to specify those type of preferences in Nix, and also makes it possible to merge objects & arrays across multiple modules.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


ncfavier (approved):

modules/programs/firefox.nix
```diff
@@ -212,6 +219,10 @@ in {
                   "distribution.searchplugins.defaultLocale" = "en-GB";
                   "general.useragent.locale" = "en-GB";
                   "browser.bookmarks.showMobileBookmarks" = true;
+                  "browser.newtabpage.pinned" = [{
+                    title = "NixOS";
+                    url = "https://nixos.org";
+                  }];
```
sumnerevans, line 225: This is fairly unintuitive behavior. I think that we should at the very least update the description to explain what will happen.

kira-bruneau: That's a great point - I'll update the description!

Issue #3507, opened 2022-12-19
bug: vscode 1.74.1 loads without extensions


donferi:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Vscode got updated to 1.74.1 in the following PR:
- https://github.com/NixOS/nixpkgs/pull/206007

After that, vscode no longer loads with extensions. Seems like it may be the incorrect extension path for darwin?

### System information

```markdown
- system: `"aarch64-darwin"`
 - host os: `Darwin 21.6.0, macOS 12.6`
 - multi-user?: `yes`
 - sandbox: `no`
 - version: `nix-env (Nix) 2.12.0`
 - channels(donferi): `"home-manager, nixpkgs"`
 - channels(root): `"nixpkgs"`
 - nixpkgs: `/Users/donferi/.nix-defexpr/channels/nixpkgs`
```
```

manuelbb-upb:
I also face this issue, because apparently the new version needs an "extensions.json" file.
See [this commit](https://github.com/NixOS/nixpkgs/commit/66a6cd1d5191af8d3ad3dfd7c31f49d930696b68) for `vscode-with-extensions` in nixpkgs.
My current workaround is setting `programs.vscode.mutableExtensionsDir=true;`
I also had to delete the extensions dir for this to work. 
VSCode will then create an "extensions.json".

Sciencentistguy:
Setting `mutableExtensionsDir=true` appears to stop home-manager from linking any extensions into the directory. My workaround was to use `home.file` to set them manually, and generate the json using the code from nixpkgs. See https://github.com/Sciencentistguy/nixfiles/commit/e9eb884cce63ffc41147d1ecfd0569910a3909c9

AnthonyEnr1quez:
Confirmed extensions are working with vscodium 1.75.0 and changes from #3588

Thanks @Taneb!

pwaller:
I just edited my home configuration to add one extension to the list of extensions under `vscode.extensions`. Upon reloading, vscode now only has that one extension installed and all the others from the list are missing.

ncfavier:
Which extension is that?

pwaller:
`bungcip.better-toml`.

The extensions.json when I found it only contained this one plugin. I never installed an extension manually. I exited vscode entirely.

[...]

PR #3588, opened 2023-01-12
vscode: add extensions.json file in extensions dir

Taneb:
### Description

Add `extensions.json` file in `.share/vscode/extensions`. This allows extensions to be used with vscode 1.74.1. See #3507.

I haven't thoroughly tested behaviour with `mutableExtensionsDir = false`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/vscode.nix
+++ b/modules/programs/vscode.nix
@@ -220,23 +220,26 @@ in {
           else
             builtins.attrNames (builtins.readDir (ext + "/${subDir}")));
       in if cfg.mutableExtensionsDir then
-        mkMerge (concatMap toPaths cfg.extensions ++ lib.optional (lib.versionAtLeast vscodeVersion "1.74.0") [{
-          # Whenever our immutable extensions.json changes, force VSCode to regenerate
-          # extensions.json with both mutable and immutable extensions.
-          "${extensionPath}/.extensions-immutable.json" = {
-            text = extensionJson;
-            onChange = ''
-              $DRY_RUN_CMD rm $VERBOSE_ARG -f ${extensionPath}/{extensions.json,.init-default-profile-extensions}
-              $VERBOSE_ECHO "Regenerating VSCode extensions.json"
-              $DRY_RUN_CMD ${getExe cfg.package} --list-extensions > /dev/null
-            '';
-          };
-        }])
+        mkMerge (concatMap toPaths cfg.extensions
+          ++ lib.optional (lib.versionAtLeast vscodeVersion "1.74.0") [{
+            # Whenever our immutable extensions.json changes, force VSCode to regenerate
+            # extensions.json with both mutable and immutable extensions.
+            "${extensionPath}/.extensions-immutable.json" = {
+              text = extensionJson;
+              onChange = ''
+                $DRY_RUN_CMD rm $VERBOSE_ARG -f ${extensionPath}/{extensions.json,.init-default-profile-extensions}
+                $VERBOSE_ECHO "Regenerating VSCode extensions.json"
+                $DRY_RUN_CMD ${getExe cfg.package} --list-extensions > /dev/null
+              '';
+            };
+          }])
       else {
         "${extensionPath}".source = let
           combinedExtensionsDrv = pkgs.buildEnv {
             name = "vscode-extensions";
-            paths = cfg.extensions ++ lib.optional (lib.versionAtLeast vscodeVersion "1.74.0") [ extensionJsonFile ];
+            paths = cfg.extensions
+              ++ lib.optional (lib.versionAtLeast vscodeVersion "1.74.0")
+              [ extensionJsonFile ];
           };
         in "${combinedExtensionsDrv}/${subDir}";
       }))
--- a/modules/programs/vscode.nix
+++ b/modules/programs/vscode.nix
@@ -247,8 +247,8 @@ in {
           else
             builtins.attrNames (builtins.readDir (ext + "/${subDir}")));
       in if cfg.mutableExtensionsDir then
-        mkMerge (concatMap toPaths cfg.extensions)
-        ++ [{ ".vscode/extensions/extensions.json".text = extensionsJson; }]
+        mkMerge (concatMap toPaths cfg.extensions
+        ++ [{ "${extensionPath}/extensions.json".text = extensionJson; }])
       else {
         "${extensionPath}".source = let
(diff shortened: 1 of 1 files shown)


modules/programs/vscode.nix
```diff
@@ -212,11 +248,13 @@ in {
             builtins.attrNames (builtins.readDir (ext + "/${subDir}")));
       in if cfg.mutableExtensionsDir then
         mkMerge (concatMap toPaths cfg.extensions)
+        ++ [{ ".vscode/extensions/extensions.json".text = extensionsJson; }]
       else {
         "${extensionPath}".source = let
           combinedExtensionsDrv = pkgs.buildEnv {
             name = "vscode-extensions";
-            paths = cfg.extensions;
+            paths = builtins.trace extensionJsonFile cfg.extensions
```
ncfavier, line 256:
```suggestion
            paths = cfg.extensions
```

modules/programs/vscode.nix
```diff
@@ -212,11 +248,13 @@ in {
             builtins.attrNames (builtins.readDir (ext + "/${subDir}")));
       in if cfg.mutableExtensionsDir then
         mkMerge (concatMap toPaths cfg.extensions)
+        ++ [{ ".vscode/extensions/extensions.json".text = extensionsJson; }]
```
ncfavier, line 251:
```suggestion
        mkMerge (concatMap toPaths cfg.extensions
        ++ [{ ".vscode/extensions/extensions.json".text = extensionsJson; }])
```

Also I wonder if this prevents installing extensions manually?

Taneb: Why do you suggest this change? I don't think that will work.

Also yes it might prevent installing extensions manually. I'd like some input from someone who uses home manager and VS Code that way and knows what to do about it

ncfavier: > I'd like some input from someone who uses home manager and VS Code that way and knows what to do about it

Ask and you shall receive. I'm suggesting these changes because I'm actually testing your code and they're required to make it work.

More in detail: `mkMerge` takes a list of modules and produces a special attrset that can't be further modified, so you want to append to the input list of modules rather than the output.

Taneb: Oh, I misread the suggestion! I'm sorry, I think you're correct. If I accept your suggestion then GitHub will mark this conversation as resolved, so could you make your point about preventing manual extension installation again elsewhere?

modules/programs/vscode.nix
```diff
@@ -211,12 +247,14 @@ in {
           else
             builtins.attrNames (builtins.readDir (ext + "/${subDir}")));
       in if cfg.mutableExtensionsDir then
-        mkMerge (concatMap toPaths cfg.extensions)
+        mkMerge (concatMap toPaths cfg.extensions
+        ++ [{ "${extensionPath}/extensions.json".text = extensionJson; }])
```
ncfavier, line 251: I can confirm that this prevents vscode from installing extensions manually. Maybe this should be an `home.activation` script that writes `extensions.json` if it doesn't exist?

Taneb: Please do!

ncfavier: Alternatively, it seems like we can make vscode regenerate the file itself with

```
rm -f ${extensionPath}/{extensions.json,.init-default-profile-extensions}
${vscodeExecutable} --list-extensions > /dev/null
```

This is probably better since it will also remove extensions.

ncfavier: This seems to work well, I've pushed it to your branch.

modules/programs/vscode.nix
```diff
@@ -34,6 +34,42 @@ let
...
+      preRelease = false;
+    };
+  };
+
+  extensionJson = builtins.toJSON (map toExtensionJsonEntry cfg.extensions);
+  extensionJsonFile = pkgs.writeTextFile {
+    name = "extensions-json";
+    text = extensionJson;
+    destination = "/share/vscode/extensions/extensions.json";
+  };
```
ncfavier, line 71: Note to self: it would be nice to deduplicate all of this between here and nixpkgs, probably by exposing it in `pkgs.vscode-utils`.

Sciencentistguy:
Given that the linked nixpkgs PR is merged, is there anything stopping this being merged? I'd quite like to retire the hacky workaround i'm running atm :sweat_smile:

modules/programs/vscode.nix
ncfavier: That PR has landed in nixos-unstable, you should be able to use `text = pkgs.vscode-utils.toExtensionJson cfg.extensions;` now.

modules/programs/vscode.nix
```diff
@@ -211,12 +220,23 @@ in {
           else
             builtins.attrNames (builtins.readDir (ext + "/${subDir}")));
       in if cfg.mutableExtensionsDir then
-        mkMerge (concatMap toPaths cfg.extensions)
+        mkMerge (concatMap toPaths cfg.extensions ++ lib.optional (lib.version >= "1.74.0") [{
```
ncfavier, line 223:
```suggestion
        mkMerge (concatMap toPaths cfg.extensions ++ lib.optional (lib.versionAtLeast vscodeVersion "1.74.0") [{
```

modules/programs/vscode.nix
```diff
@@ -211,12 +220,23 @@ in {
...
+              $DRY_RUN_CMD ${getExe cfg.package} --list-extensions > /dev/null
+            '';
+          };
+        }])
       else {
         "${extensionPath}".source = let
           combinedExtensionsDrv = pkgs.buildEnv {
             name = "vscode-extensions";
-            paths = cfg.extensions;
+            paths = cfg.extensions ++ lib.optional (vscodeVersion >= "1.74.0") [ extensionJsonFile ];
```
ncfavier, line 239:
```suggestion
            paths = cfg.extensions ++ lib.optional (lib.versionAtLeast vscodeVersion "1.74.0") [ extensionJsonFile ];
```

modules/programs/vscode.nix
Taneb: I tried to find that function in `nixpkgs` and I just couldn't (and I couldn't remember its name). I didn't think to look in `strings.nix` at all!

ncfavier:
Needs reformatting I think

ncfavier (approved):

PR #3600, opened 2023-01-16
hstr: hstr module added

Dines97:
### Description
This adds options to configure hstr. bash and zsh shell history suggest box - easily view, navigate, search and manage your command history.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/hstr.nix
+++ b/modules/programs/hstr.nix
@@ -1,0 +1,39 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.programs.hstr;
+
+in {
+  meta.maintainers = [ hm.maintainers.Dines97 ];
+
+  options.programs.hstr = {
+    enable = mkEnableOption ''
+      Bash And Zsh shell history suggest box - easily view, navigate, search and
+      manage your command history'';
+
+    package = mkPackageOption pkgs "hstr" { };
+
+    enableBashIntegration = mkEnableOption "Bash integration" // {
+      default = true;
+    };
+
+    enableZshIntegration = mkEnableOption "Zsh integration" // {
+      default = true;
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    programs.bash.initExtra = mkIf cfg.enableBashIntegration ''
+      eval "$(${cfg.package}/bin/hstr --show-configuration)"
+    '';
+
+    programs.zsh.initExtra = mkIf cfg.enableZshIntegration ''
+      eval "$(${cfg.package}/bin/hstr --show-zsh-configuration)"
+    '';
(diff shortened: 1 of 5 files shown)


modules/programs/hstr.nix
```diff
@@ -0,0 +1,46 @@
...
+  options.programs.hstr = {
+    enable = mkEnableOption
+      "bash and zsh shell history suggest box - easily view, navigate, search and manage your command history";
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.hstr;
+      defaultText = literalExpression "pkgs.hstr";
+      description = "Package providing the <command>hstr</command> tool.";
+    };
```
rycee, line 16: Can simplify:

```suggestion
    package = mkPackageOption pkgs "hstr" { };
```

modules/programs/hstr.nix
```diff
@@ -0,0 +1,46 @@
+{ config, lib, pkgs, ... }:
+with lib;
+let cfg = config.programs.hstr;
+in {
+  meta.maintainers = [ maintainers.Dines97 ];
+
+  options.programs.hstr = {
+    enable = mkEnableOption
+      "bash and zsh shell history suggest box - easily view, navigate, search and manage your command history";
```
rycee, line 9: Avoid long lines (also capitalization):

```suggestion
    enable = mkEnableOption ''
      Bash And Zsh shell history suggest box - easily view, navigate, search and
      manage your command history
    '';
```

modules/programs/hstr.nix
```diff
@@ -0,0 +1,46 @@
+{ config, lib, pkgs, ... }:
+with lib;
+let cfg = config.programs.hstr;
+in {
+  meta.maintainers = [ maintainers.Dines97 ];
```
rycee, line 5: The CI job fails because you are not present in the maintainers attribute set. You can add yourself in `lib/maintainers.nix` and reference that

```suggestion
  meta.maintainers = [ hm.maintainers.Dines97 ];
```

rycee:

modules/programs/hstr.nix
Dines97: Just copy pasted from the hstr github page to match this.

modules/programs/hstr.nix
Dines97: Didn't know that fixing right now

Dines97: Why hm.maintainers instead of just maintainers?

rycee: Without the `hm` it refers to the [Nixpkgs maintainers list](https://github.com/NixOS/nixpkgs/blob/master/maintainers/maintainer-list.nix).

rycee:
Thanks! Merged to master now.

PR #3601, opened 2023-01-18
avizo: add module

pltanton:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/avizo.nix
+++ b/modules/services/avizo.nix
@@ -1,0 +1,70 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+let
+  cfg = config.services.avizo;
+  settingsFormat = pkgs.formats.ini { };
+in {
+
+  meta.maintainers = [ hm.maintainers.pltanton ];
+
+  options.services.avizo = {
+    enable = mkEnableOption "avizo, a simple notification daemon";
+    settings = mkOption {
+      type = (pkgs.formats.ini { }).type;
+      default = { };
+      example = {
+        default = {
+          time = 1.0;
+          y-offset = 0.5;
+          fade-in = 0.1;
+          fade-out = 0.2;
+          padding = 10;
+        };
+      };
+      description =
+        "Attrs that will be converted to .ini avizo configuration file.";
+    };
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.avizo;
+      defaultText = literalExpression "pkgs.avizo";
+      example = literalExpression ''
+        pkgs.avizo.overrideAttrs (final: prev: {
+          patchPhase = "cp ''${./images}/*.png data/images/";
+        })
+      '';
+      description = "The <literal>avizo</literal> package to use.";
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+    xdg.configFile."avizo/config.ini" = {
+      source = settingsFormat.generate "avizo-config.ini" cfg.settings;
+    };
+
+    systemd.user = {
+      services.avizo = {
+        Unit = {
+          Description = "Volume/backlight OSD indicator";
+          PartOf = [ "graphical-session.target" ];
+          After = [ "graphical-session.target" ];
+          ConditionEnvironment = "WAYLAND_DISPLAY";
+          Documentation = "man:avizo(1)";
+        };
+
+        Service = {
+          Type = "simple";
+          ExecStart = ''
+            ${cfg.package}/bin/avizo-service
+          '';
+          Restart = "always";
+        };
+
+        Install = { WantedBy = [ "graphical-session.target" ]; };
+      };
+    };
+  };
+}
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -355,6 +355,8 @@ Makefile                                              @thiagokokada
 
 /modules/programs/zsh/prezto.nix                      @NickHu
 
+/modules/services/avizo.nix                           @pltanton
+
 /modules/services/barrier.nix                         @Kritnich
 /tests/modules/services/barrier                       @Kritnich
 
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -886,6 +886,14 @@ in
           'wayland.windowManager.sway.config.[window|floating].titlebar' now default to 'true'.
         '';
       }
+
+      {
+        time = "2023-01-18T11:19:41+00:00";
+        condition = hostPlatform.isLinux;
+        message = ''
+          A new module is available: 'services.avizo'.
+        '';
+      }
     ];
   };
 }
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -200,6 +200,7 @@ let
     ./programs/zplug.nix
     ./programs/zsh.nix
     ./programs/zsh/prezto.nix
+    ./services/avizo.nix
     ./services/barrier.nix
     ./services/betterlockscreen.nix
     ./services/blueman-applet.nix


modules/services/avizo.nix
```diff
@@ -0,0 +1,74 @@
...
+          padding = 10;
+        };
+      };
+      description =
+        "Attrs that will be converted to .ini avizo configuration file.";
+    };
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.easyeffects;
```
ncfavier, line 35:
```suggestion
      default = pkgs.avizo;
```

modules/services/avizo.nix
```diff
@@ -0,0 +1,74 @@
...
+  options.services.avizo = {
+    enable = mkOption {
+      default = false;
+      description = ''
+        Whether to enable avizo.
+        Avizo is a simple notification daemon, mainly intended to be used for multimedia keys for example with Sway.
+      '';
+    };
+    config = mkOption {
+      type = types.attrs;
```
ncfavier, line 18: Please make this more specific using [pkgs.formats.ini](https://nixos.org/manual/nixos/unstable/#sec-settings-nix-representable)

modules/services/avizo.nix
```diff
@@ -0,0 +1,74 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+let cfg = config.services.avizo;
+in {
+
+  meta.maintainers = [ hm.maintainers.pltanton ];
+
+  options.services.avizo = {
+    enable = mkOption {
```
ncfavier, line 10: Needs a type. Or just `mkEnableOption "avizo, a simple notification daemon […]"`.

modules/services/avizo.nix
```diff
@@ -0,0 +1,74 @@
...
+    };
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.easyeffects;
+      defaultText = literalExpression "pkgs.avizo";
+      example = literalExpression ''
+        pkgs.avizo.overrideAttrs (final: prev: {
+          patchPhase = "cp ''${./images}/*.png data/images/
+        });
```
ncfavier, line 40: This is not a valid Nix expression (unfinished line, final semicolon)

modules/services/avizo.nix
```diff
@@ -0,0 +1,74 @@
...
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+    xdg.configFile."avizo/config.ini" = {
+      text = (lib.generators.toINI { } cfg.config);
+    };
+
+    systemd.user = {
+      services.avizo = with osConfig.lib.stylix.colors; {
```
ncfavier, line 53: ?

ncfavier (changes requested):

modules/services/avizo.nix
pltanton: Fixed

modules/services/avizo.nix
pltanton: Fixed

modules/services/avizo.nix
pltanton: Fixed

modules/services/avizo.nix
pltanton: Fixed

modules/services/avizo.nix
pltanton: fixed

modules/services/avizo.nix
```diff
@@ -0,0 +1,71 @@
...
+in {
+
+  meta.maintainers = [ hm.maintainers.pltanton ];
+
+  options.services.avizo = {
+    enable = mkEnableOption ''
+      Whether to enable avizo.
+      Avizo is a simple notification daemon, mainly intended to be used for multimedia keys for example with Sway.
+    '';
+    config = mkOption {
```
ncfavier, line 14: This should probably be called `settings` for consistency (and because `config` is quite overloaded)

modules/services/avizo.nix
```diff
@@ -0,0 +1,71 @@
...
+    };
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.avizo;
+      defaultText = literalExpression "pkgs.avizo";
+      example = literalExpression ''
+        pkgs.avizo.overrideAttrs (final: prev: {
+          patchPhase = "cp ''${./images}/*.png data/images/";
+        });
```
ncfavier, line 37:
```suggestion
        })
```

modules/services/avizo.nix
```diff
@@ -0,0 +1,71 @@
...
+with lib;
+let cfg = config.services.avizo;
+in {
+
+  meta.maintainers = [ hm.maintainers.pltanton ];
+
+  options.services.avizo = {
+    enable = mkEnableOption ''
+      Whether to enable avizo.
+      Avizo is a simple notification daemon, mainly intended to be used for multimedia keys for example with Sway.
```
ncfavier, line 12: This string is interpolated as `Whether to enable {}.`.

```suggestion
      avizo, a simple notification daemon, mainly intended to be used for multimedia keys for example with Sway
```

modules/services/avizo.nix
```diff
@@ -0,0 +1,71 @@
...
+          After = [ "graphical-session.target" ];
+          ConditionEnvironment = "WAYLAND_DISPLAY";
+          Documentation = "man:avizo(1)";
+        };
+
+        Service = {
+          Type = "simple";
+          ExecStart = ''
+            ${cfg.package}/bin/avizo-service
+          '';
```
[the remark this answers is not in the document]
pltanton: It do not accepts any flags

modules/services/avizo.nix
rycee: Yeah, definitely need to change the name to `settings`.

modules/services/avizo.nix
```diff
@@ -0,0 +1,70 @@
...
+with lib;
+let cfg = config.services.avizo;
+in {
+
+  meta.maintainers = [ hm.maintainers.pltanton ];
+
+  options.services.avizo = {
+    enable = mkEnableOption ''
+      avizo, a simple notification daemon, mainly intended to be used for multimedia keys for example with Sway
+    '';
```
rycee, line 12: Either split the line to make it a bit shorter or just simplify the description :-)

```suggestion
    enable = mkEnableOption "avizo, a simple notification daemon";
```

If you want to keep the longer description make sure you don't have a newline at the end of the text. Otherwise there will be a space between the last character and the period.

modules/services/avizo.nix
```diff
@@ -0,0 +1,70 @@
...
+      description =
+        "Attrs that will be converted to .ini avizo configuration file.";
+    };
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.avizo;
+      defaultText = literalExpression "pkgs.avizo";
+      example = literalExpression ''
+        pkgs.avizo.overrideAttrs (final: prev: {
```
rycee, line 34: Are you sure this is right? Typically the `overrideAttrs` argument should be a function that just takes a single argument. Something like `overrideAttrs (oldAttrs: { …})`.

pltanton: Yes, according to dots: https://nixos.org/manual/nixpkgs/stable/#sec-pkg-overrideAttrs it also works in my personal setup

rycee:
Thanks! Rebased to master in 4295fdfa6b0005c32f2e1f0b732faf5810c1bc7f.

modules/services/avizo.nix
```diff
@@ -0,0 +1,70 @@
...
+      source = settingsFormat.generate "avizo-config.ini" cfg.settings;
+    };
+
+    systemd.user = {
+      services.avizo = {
+        Unit = {
+          Description = "Volume/backlight OSD indicator";
+          PartOf = [ "graphical-session.target" ];
+          After = [ "graphical-session.target" ];
+          ConditionEnvironment = "WAYLAND_DISPLAY";
```
misuzu, line 54: Is there a reason for this condition? avizo works with Xorg too.

PR #3603, opened 2023-01-18
rbenv: add module

marsam:
### Description

Add module for [rbenv](https://github.com/rbenv/rbenv)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/rbenv.nix
+++ b/modules/programs/rbenv.nix
@@ -1,0 +1,93 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.programs.rbenv;
+
+  pluginModule = types.submodule {
+    options = {
+      src = mkOption {
+        type = types.path;
+        description = ''
+          Path to the plugin folder.
+        '';
+      };
+      name = mkOption {
+        type = types.str;
+        description = ''
+          Name of the plugin.
+        '';
+      };
+    };
+  };
+
+in {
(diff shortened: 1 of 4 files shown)


modules/programs/rbenv.nix
```diff
@@ -0,0 +1,93 @@
...
+    programs.bash.initExtra = mkIf cfg.enableBashIntegration ''
+      eval "$(${cfg.package}/bin/rbenv init - bash)"
+    '';
+
+    programs.zsh.initExtra = mkIf cfg.enableZshIntegration ''
+      eval "$(${cfg.package}/bin/rbenv init - zsh)"
+    '';
+
+    programs.fish.shellInit = mkIf cfg.enableFishIntegration ''
+      ${cfg.package}/bin/rbenv init - fish | source
```
ncfavier, line 90: Is this known to work in fish? In bash, every command in a pipe runs in a subshell by default.

marsam: I'm no fully familiar with fish, but that is the command described in the rbenv documentation https://github.com/rbenv/rbenv#basic-git-checkout, and it's also used by other modules e.g.
https://github.com/nix-community/home-manager/blob/cc94ac214a3108a198ccaae0849393861b505a05/modules/programs/oh-my-posh.nix#L90-L92

rycee:
Thanks! Merged to master now.

PR #3604, opened 2023-01-19
vim-vint: add module

tomodachi94:
### Description

Adds the [Vint](https://github.com/Vimjas/vint) Vimscript linter as a module.

This is my first PR to Home-Manager; if I've done something wrong please tell me!

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/vim-vint.nix
+++ b/modules/programs/vim-vint.nix
@@ -1,0 +1,36 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.programs.vim-vint;
+
+  yamlFormat = pkgs.formats.yaml { };
+
+in {
+  meta.maintainers = [ maintainers.tomodachi94 ];
+
+  options = {
+    programs.vim-vint = {
+      enable = mkEnableOption "the Vint linter for Vimscript";
+      package = mkPackageOption pkgs "vim-vint" { };
+
+      settings = mkOption {
+        type = yamlFormat.type;
+        default = { };
+        description = ''
+          Configuration written to
+          <filename>$XDG_CONFIG_HOME/.vintrc.yaml</filename>
+        '';
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    xdg.configFile.".vintrc.yaml".source =
+      yamlFormat.generate "vim-vint-config" cfg.settings;
+  };
+}
--- a/tests/modules/programs/vim-vint/basic-configuration.nix
+++ b/tests/modules/programs/vim-vint/basic-configuration.nix
@@ -1,0 +1,25 @@
+{ config, pkgs, lib, xdg, ... }:
+
+{
+  config.programs.vim-vint = {
+    enable = true;
+    settings = {
+      cmdargs = {
+        severity = "error";
+        color = true;
+        env = { neovim = true; };
+      };
+      policies = {
+        ProhibitEqualTildeOperator.enabled = false;
+        ProhibitUsingUndeclaredVariable.enabled = false;
+        ProhibitAbbreviationOption.enabled = false;
+        ProhibitImplicitScopeVariable.enabled = false;
+        ProhibitSetNoCompatible.enabled = false;
+      };
+    };
+    config.nmt.script = let configFile = xdg.configFile.".vintrc.yaml";
+    in "	  assertFileContent \\\n	  \"${configFile}\" \\\n	  ${
+         ./basic-configuration.yaml
+       }\n  ";
+  };
+}
--- a/tests/modules/programs/vim-vint/basic-configuration.nix
+++ b/tests/modules/programs/vim-vint/basic-configuration.nix
@@ -1,7 +1,7 @@
 { config, pkgs, lib, xdg, ... }:
 
 {
-  config.programs.vim-vint = {
+  programs.vim-vint = {
     enable = true;
     settings = {
       cmdargs = {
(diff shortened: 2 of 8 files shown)


tomodachi94:
In the `policies` section, the `enabled` keyword is in past tense, which contradicts most Nix configurations I've seen. It works fine with Vint, but should we just leave it?

modules/programs/vim-vint.nix
```diff
@@ -0,0 +1,53 @@
...
+  options = {
+    programs.vim-vint = {
+      enable = mkOption {
+        type = types.bool;
+        default = false;
+        defaultText = "false";
+        description = ''
+          Whether to enable the Vint linter for Vimscript.
+        '';
+      };
```
ncfavier, line 23:
```suggestion
      enable = mkEnableOption "the Vint linter for Vimscript";
```

modules/programs/vim-vint.nix
```diff
@@ -0,0 +1,53 @@
...
+      package = mkOption {
+        type = types.package;
+        default = pkgs.vim-vint;
+        defaultText = literalExpression "pkgs.vim-vint";
+        example =
+          literalExpression "(pkgs.vim-vint.override { enableCheck = true; })";
+        description = ''
+          The <literal>vim-vint</literal> package to use.
+        '';
+      };
```
ncfavier, line 34:
```suggestion
      package = mkPackageOption pkgs "vim-vint" { };
```

I don't think the example is very useful or specific to vim-vint?

modules/programs/vim-vint.nix
```diff
@@ -0,0 +1,53 @@
...
+          <filename>$HOME/.vintrc.yaml</filename>
+        '';
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    xdg.configFile.".vintrc.yaml".source = # Vint's configuration location isn't very smart. It checks $XDG_CONFIG_HOME for `.vintrc.yaml`, then it checks $HOME.
```
ncfavier, line 50: The comment should go on its own line (or two lines). I'd just leave it out though.

ncfavier:
> In the `policies` section, the `enabled` keyword is in past tense, which contradicts most Nix configurations I've seen. It works fine with Vint, but should we just leave it?

Yes, it's fine. Let's not rename settings, that would be surprising.

tests/modules/programs/vim-vint/basic-configuration.yaml
```diff
@@ -0,0 +1,16 @@
...
+  ProhibitAbbreviationOption:
+    enabled: false
+  ProhibitEqualTildeOperator:
+    enabled: false
+  ProhibitImplicitScopeVariable:
+    enabled: false
+  ProhibitSetNoCompatible:
+    enabled: false
+  ProhibitUsingUndeclaredVariable:
+    enabled: true
```
ncfavier, line 16: This is set to `false` in the config, why is it `true` here?

tomodachi94:
@ncfavier I believe I've fixed the issues you raised above.

~~Now to wait for the tests to pass...~~

tests/modules/programs/vim-vint/basic-configuration.nix
```diff
@@ -0,0 +1,21 @@
...
+      policies = {
+        ProhibitEqualTildeOperator.enabled = false;
+        ProhibitUsingUndeclaredVariable.enabled = false;
+        ProhibitAbbreviationOption.enabled = false;
+        ProhibitImplicitScopeVariable.enabled = false;
+        ProhibitSetNoCompatible.enabled = false;
+      };
+    };
+  };
+}
```
ncfavier, line 21: The reason that the tests were passing earlier even though the expected configuration was wrong is that you did not add a test script! This test isn't actually testing anything. See other files in `tests/` for examples.

tomodachi94: Fixed in the latest push.

ncfavier: You can run the test locally before pushing. `nix-shell --pure ./tests -A run.vim-vint-basic-configuration`

tomodachi94:
Fixed the tests yet again, it appears they work as expected.

tests/modules/programs/vim-vint/basic-configuration.nix
```diff
@@ -0,0 +1,31 @@
...
+        ProhibitEqualTildeOperator.enabled = false;
+        ProhibitUsingUndeclaredVariable.enabled = false;
+        ProhibitAbbreviationOption.enabled = false;
+        ProhibitImplicitScopeVariable.enabled = false;
+        ProhibitSetNoCompatible.enabled = false;
+      };
+    };
+    config.nmt.script = let
+		configDir = if pkgs.stdenv.isDarwin then
+		  "home-files/Library/Application Support"
```
ncfavier, line 22: This would only be needed if we had a similar test in the module, but we always use `xdg.configFile`.

tomodachi94:
@ncfavier I believe I've fixed it.

ncfavier:
Please format.

tomodachi94:
> Please format.

Done.

tests/modules/programs/vim-vint/basic-configuration.nix
```diff
@@ -0,0 +1,25 @@
...
+      policies = {
+        ProhibitEqualTildeOperator.enabled = false;
+        ProhibitUsingUndeclaredVariable.enabled = false;
+        ProhibitAbbreviationOption.enabled = false;
+        ProhibitImplicitScopeVariable.enabled = false;
+        ProhibitSetNoCompatible.enabled = false;
+      };
+    };
+    config.nmt.script = let configFile = xdg.configFile.".vintrc.yaml";
+    in "	  assertFileContent \\\n	  \"${configFile}\" \\\n	  ${
```
tomodachi94, line 21: The formatter screwed up the formatting here; should I leave it alone?

ncfavier: I think it was not happy with the tabs. You should configure your editor to use spaces for indentation

ncfavier (approved):

PR #3607, opened 2023-01-20
herbstluftwm: Fix default tag renaming

olmokramer:
### Description

The previous implementation tried to rename the tag named "default" to the first tag in `cfg.tags`. This was a wrong approach because if a tag with the same name already existed, the renaming failed and the default tag would continue to exist.

The looking up of the default tag also contained a bug (introduced in #3509) because it should have used `by-name` in the path.

### Checklist

- [x] Change is backwards compatible. **Technically it is not, because the tag named "default" was never removed, but that is considered a bug. And technically #3509 (which introduced the bug) was also not backwards compatible.**

    ```
    {component}: {description}

    {long description}
    ```


rycee:

Issue #2991, opened 2022-05-31
bug: home.sessionPath is broken with ZSH


sagikazarmark:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

#2708 moved session variable loading from `.zshrc` to `.zshenv`.

While it works with most session variables, it breaks `home.sessionPath` that modifies `PATH`.

It breaks, because zshenv is loaded first and subsequent PATH modifications (like Nix loading) override variables set in zshenv.

Prior to #2708 I submitted #2445 assessing the situation. In that (based on #183) I concluded that .zshenv is not the best place for session variables. Quoting the example [.zshenv](https://zsh.sourceforge.io/Contrib/startup/std/zshenv) file:

```
# .zshenv is sourced on ALL invocations of the shell, unless the -f option is
# set.  It should NOT normally contain commands to set the command search path,
# or other common environment variables unless you really know what you're
# doing.  E.g. running "PATH=/custom/path gdb program" sources this file (when
# gdb runs the program via $SHELL), so you want to be sure not to override a
# custom environment in such cases.  Note also that .zshenv should not contain
# commands that produce output or assume the shell is attached to a tty.
```

Quoting why it was still added to `.zshenv` from #2708:

> The reason why these session variables are not moved to .zprofile is
> that programs started by systemd user instances are not able to get
> variables defined in that file. For example, GNOME
> Terminal (gnome-terminal-server.service) is one of these programs and
> doesn't get variables defined in .zprofile. As a result, the shells it
> starts, which are interactive and non-login, do not get those
> variables.

Based on the available information, I believe `.zshenv` is not the right place for session variables by default. I understand that it's problematic for some graphical sessions, but the current solution still goes against how ZSH startup scripts are supposed to be used and requires workarounds for "normal" use cases.

I propose moving session variables to zprofile instead of zshenv. Anyone in need of loading them in zshenv as well can workaround the issue (or we can introduce a variable if that's absolutely necessary).

I'd also like to point out that bash also uses the equivalent of zprofile at the moment.

Temporary workaround:

```nix
{
    zsh.profileExtra = lib.optionalString (config.home.sessionPath != [ ]) ''
      export PATH="$PATH''${PATH:+:}${lib.concatStringsSep ":" config.home.sessionPath}"
    '';
}
```

### Maintainer CC

cc @jian-lin

### System information

```markdown
- system: `"x86_64-linux"`
[...]

PR #3610, opened 2023-01-21
fish: warn of shell init

jian-lin:
### Description
bash is always enabled in nixos (`programs.bash.enable` has been removed since 2014) and defaults to enable in nix-darwin, so it is not added in this pr. 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -1,4 +1,4 @@
-{ config, lib, pkgs, ... }:
+{ config, lib, pkgs, ... }@moduleArgs:
 
 with lib;
 
@@ -289,6 +289,9 @@ in {
 
   config = mkIf cfg.enable (mkMerge [
     {
+      warnings = optional (!(moduleArgs.osConfig.programs.fish.enable or true))
+        "Enable os-level programs.fish.enable for proper shell initialization.";
+
       home.packages = [ cfg.package ];
 
       # Support completion for `man` by building a cache for `apropos`.


jian-lin:
Fixes: https://github.com/nix-community/home-manager/issues/2991
Bug: https://github.com/nix-community/home-manager/issues/2751#issuecomment-1048682643

jian-lin:
CI is passed.

ncfavier:
Is there really not a solution to this on the NixOS side?

jian-lin:
> Is there really not a solution to this on the NixOS side?

What do you think the solution should look like?

IMO, the requirement of setting up system level shell config files e.g., /etc/zprofile and /etc/fish/config.fish, is pretty reasonable.

ncfavier:
> What do you think the solution should look like?

I don't know and to be fair I don't fully understand the problem, but I don't believe that the solution is to tell people to enable a NixOS module that might be out of their control. Seems like a clear layer violation.

Surely there's *some* way to organise zsh's system-wide and user initialisation files across NixOS and home-manager so that everything happens in the right order.

jian-lin:
> to be fair I don't fully understand the problem

[Here][1] is my explanation of the problem and the cause for zsh.

For fish, the problem is a bit different: if system-level module is not enabled, system-level environment variables will not be set.  I do not known if there are issues opened about this, but I have seen several people encountering this.

> I don't believe that the solution is to tell people to enable a NixOS module that might be out of their control.

Ah, makes sense.  I did not see this use case.

FYI, I have created a [PR][2] for zsh in nixpkgs.  

For fish, I guess the use case of not enabling the system-level module is just [not supported][7].  However, a possible fix is to use [`fishEnvPreInit`][3] to set system-level environment variables and drop the need for [the][4] [module][5].  However, I do not like the current default converting method, which is `fenv`, so I will not make that PR.  I prefer `babelfish`, but making it default will [not][8] be [easy][6].

> Seems like a clear layer violation.

I agree and I do not like that neither. However, the fact is that some violation is necessary for now.  For example, the description of home-manager's option `programs.zsh.enableCompletion` says
> Don't forget to add `environment.pathsToLink = [ "/share/zsh" ];` to your system configuration to get completion for system packages (e.g. systemd).

For this pr, I can remove the zsh part if that is welcome. Otherwise, feel free to close it.

[1]: https://github.com/nix-community/home-manager/issues/3681#issuecomment-1436054233
[2]: https://github.com/NixOS/nixpkgs/pull/217205
[3]: https://github.com/NixOS/nixpkgs/blob/ddccab6ccc5bc6af359ceb635f20c8d923f42f45/pkgs/shells/fish/default.nix#L38
[4]: https://github.com/NixOS/nixpkgs/blob/ddccab6ccc5bc6af359ceb635f20c8d923f42f45/pkgs/shells/fish/default.nix#L89
[5]: https://github.com/NixOS/nixpkgs/blob/ddccab6ccc5bc6af359ceb635f20c8d923f42f45/nixos/modules/programs/fish.nix#L172
[6]: https://github.com/NixOS/nixpkgs/pull/108947#issuecomment-758182608
[7]: https://github.com/NixOS/nixpkgs/commit/3f6d21bafc50d788d570a74dad957e792cea855d#diff-8c4c874e0d59af17b669edc4add81be8026e72fb8b070036955cdbfa22aaecfcR51
[8]: https://github.com/bouk/babelfish/issues

jian-lin:
Since https://github.com/NixOS/nixpkgs/pull/217205 for zsh is merged in nixpkgs, I changed this pr to only warn for fish.

jian-lin:
@ncfavier 
> I don't believe that the solution is to tell people to enable a NixOS module that might be out of their control

Yeah, this is just a warning instead of a solution, which I have mentioned in https://github.com/nix-community/home-manager/pull/3610#issuecomment-1436113813.

IMHO, this warning is still valuable even if you cannot enable the system level fish module because it saves you a lot of time when debugging for environment variable issues.

Issue #3615, opened 2023-01-23
Add `formatter` flake output


cyntheticfox:
Nix has supported a `formatter` flake output for a little while now, and I think that adding a derivation to the flake to call `./format` could be helpful for developers new to the project, and would better integrate with tools that may try to look for a standard flake output

ncfavier:
https://github.com/nix-community/home-manager/pull/3620

PR #3620, opened 2023-01-25
flake: add formatter

ncfavier:
Closes https://github.com/nix-community/home-manager/issues/3615

### DIFF
--- a/format
+++ b/format
@@ -1,7 +1,8 @@
 #! /usr/bin/env nix-shell
-#! nix-shell -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/6616de389ed55fba6eeba60377fc04732d5a207c.tar.gz -i bash -p findutils nixfmt
+#! nix-shell -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/6616de389ed55fba6eeba60377fc04732d5a207c.tar.gz -i bash -p git gnugrep gnused findutils nixfmt
 
 nixfmt_args=()
+files=()
 
 for arg do
     case $arg in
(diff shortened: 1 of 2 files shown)


houstdav000 (commented):
Looks to work this way from my testing, though looking at the changes to the "format" script, it might be good in the future to rewrite to allow for running on distinct files, or to use something like `git rev-parse HEAD` to resolve the repository root.

ncfavier:
I hadn't thought of using git. We can list the files relative to the repo root with `git ls-files`, filter them, and reconstruct absolute paths by prepending `git rev-parse --show-toplevel`.

format
```diff
@@ -1,27 +1,51 @@
...
+    exclude_args+=(-e "$e")
+done
+
+git_root=$(git rev-parse --show-toplevel)
+
+git ls-files --full-name -- "${files[@]}" |
+    grep '\.nix$' |
+    grep -v "${exclude_args[@]}" |
+    sed "s|^|$git_root/|" |
+    xargs -d '\n' nixfmt "${nixfmt_args[@]}"
```
houstdav000, line 51: Probably best to check that `sed` command. I don't think that's the right syntax.

Either way, I think `find` was the right approach anyways, as `git ls-files` will only work on tracked files. Now, that may actually be desirable behavior, but I doubt it. If it is, there are options to `git ls-files` for exclusions, null-termination.

It'd make much more sense afaict to keep the default behavior of formatting all files in the repository if no file names are provided, but to run the formatter absolutely if given any file names.

That being said, the `find` function could well leverage `-execdir` instead of `exec`, and ensuring `-type f`. It might actually be faster though to leverage the ability of `nixfmt` to work on multiple files with `find $GITROOT -mindepth 1 -name '.git' -prune -o -type f $FINDEXPR[@] -print0 | xargs -0 nixfmt ${nixfmt_args[@]}`. Where $FINDEXPR is a statement like `! -path $FILEPATH`.

GNU find is quite a beast...

ncfavier: > Probably best to check that `sed` command. I don't think that's the right syntax.

> :sparkles: Always read the manual! You never know what you might learn! :sparkles:

> as `git ls-files` will only work on tracked files

We can add `-co`, that would list untracked files as well.

Sadly `git ls-files --exclude` only filters untracked files, so we have to do the filtering ourselves. (We can't use [pathspecs](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefpathspecapathspec) for this either: they're understood as a disjunction, so we can't say "list the files under those directories AND that match `*.nix`").

> null-termination

I'm normally in favour of that but that's obviously never going to be an issue with home-manager and I can't be bothered.

> `find ... -print0 | xargs -0 nixfmt ...`

That's equivalent to `-exec nixfmt {} +`, which is what we had previously.

ncfavier: > can't be bothered.

Did it anyway.

houstdav000: > > Probably best to check that `sed` command. I don't think that's the right syntax.
> 
> > sparkles Always read the manual! You never know what you might learn! sparkles

Guess that's what I get for not reading every possible manpage out there and expecting GNU's `sed` manpage to actually detail the command, not just implementation-specifics...

ncfavier: Yeah GNU man pages suck. This is documented in the POSIX man page `sed(1p)` and in the actual GNU sed manual.

cyntheticfox (approved):
Having re-looked at it, looks good to me.

somasis:
Anything holding back this from being merged? It'd be nice to have for entirely selfish reasons (like being able to reliably use `nix fmt` as my editor's default Nix file formatting command).

PR #3622, opened 2023-01-25
files: allow disabling creation of a file

BurNiinTRee:
### Description
Sometimes it might be necessary to disable a file created by some other module. A similar option exists in NixOS.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/files.nix
+++ b/modules/files.nix
@@ -4,7 +4,7 @@ with lib;
 
 let
 
-  cfg = config.home.file;
+  cfg = filterAttrs (n: f: f.enable) config.home.file;
 
   homeDirectory = config.home.homeDirectory;
 
--- a/modules/lib/file-type.nix
+++ b/modules/lib/file-type.nix
@@ -15,6 +15,14 @@ in
   fileType = basePathDesc: basePath: types.attrsOf (types.submodule (
     { name, config, ... }: {
       options = {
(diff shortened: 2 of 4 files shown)


modules/files.nix
```diff
@@ -5,6 +5,7 @@ with lib;
 let
 
   cfg = config.home.file;
+  cfg' = filterAttrs (n: f: f.enable) cfg;
```
ncfavier, line 8:
```suggestion
  cfg = filterAttrs (n: f: f.enable) config.home.file;
```

modules/lib/file-type.nix
```diff
@@ -15,6 +15,14 @@ in
   fileType = basePathDesc: basePath: types.attrsOf (types.submodule (
     { name, config, ... }: {
       options = {
+        enable = mkOption {
+          type = types.bool;
+          default = true;
+          description = ''
+            Wether this file should be generated. This
```
ncfavier, line 22:
```suggestion
            Whether this file should be generated. This
```

ncfavier (commented):
Seems reasonable

ncfavier (approved):

rycee:
Thanks! Merged to master in 7efca2ca18062791cbaa838f4d3e23d629bb1473.

PR #3623, opened 2023-01-26
services.autorandr: add module

GaetanLepage:
### Description

[autorandr](https://github.com/phillipberndt/autorandr) already has a home-manager module: [`programs.autorandr`](https://github.com/nix-community/home-manager/blob/master/modules/programs/autorandr.nix).
The latter configures the profiles and adds the package (`pkgs.autorandr`) to `home.packages`. This is meant for a CLI usage.

Meanwhile, the NixOS has a (unique) module for autorandr: [`services.autorandr`](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/nixos/modules/services/misc/autorandr.nix) which does the same as the home-manager `programs.autorandr` module as well as adding a `systemd` service to run autorandr automatically.

This PR adds a second home-manager autorandr module: `service.autorandr` which simply adds a `systemd` user service that starts `autorandr` automatically.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/autorandr.nix
+++ b/modules/services/autorandr.nix
@@ -1,0 +1,53 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let cfg = config.services.autorandr;
+
+in {
+
+  meta.maintainers = [ maintainers.GaetanLepage ];
+
+  options = {
+    services.autorandr = {
+      enable = mkEnableOption "" // {
+        description = ''
+          Whether to enable the Autorandr systemd service.
+          This module is complementary to <code>programs.autorandr</code> which handles the
+          configuration (profiles).
+        '';
+      };
+
+      ignoreLid = mkOption {
+        default = false;
+        type = types.bool;
+        description =
+          "Treat outputs as connected even if their lids are closed.";
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+    assertions = [
+      (lib.hm.assertions.assertPlatform "services.autorandr" pkgs
+        lib.platforms.linux)
+    ];
+
+    systemd.user.services.autorandr = {
+      Unit = {
+        Description = "autorandr";
+        After = [ "graphical-session-pre.target" ];
+        PartOf = [ "graphical-session.target" ];
+      };
+
+      Service = {
+        Type = "oneshot";
+        ExecStart = "${pkgs.autorandr}/bin/autorandr --change ${
+            optionalString cfg.ignoreLid "--ignore-lid"
+          }";
+      };
+
+      Install.WantedBy = [ "graphical-session.target" ];
+    };
+  };
+}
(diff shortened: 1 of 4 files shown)


modules/services/autorandr.nix
```diff
@@ -0,0 +1,47 @@
...
+          This module is complementary to <code>programs.autorandr</code> which handles the
+          configuration (profiles).
+        '';
+      };
+
+      ignoreLid = mkOption {
+        default = false;
+        type = types.bool;
+        description =
+          "Treat outputs as connected even if their lids are closed";
```
rycee, line 22:
```suggestion
          "Treat outputs as connected even if their lids are closed.";
```

modules/services/autorandr.nix
```diff
@@ -0,0 +1,47 @@
...
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+    systemd.user.services.autorandr = {
```
rycee, line 28: Seems this will only run once, when starting the graphical session? Shouldn't it somehow be hooked up to run after the screen setup changes?

modules/services/autorandr.nix
```diff
@@ -0,0 +1,47 @@
...
+        ExecStart = "${pkgs.autorandr}/bin/autorandr --change ${
+            optionalString cfg.ignoreLid "--ignore-lid"
+          }";
+      };
+
+      Install.WantedBy = [ "graphical-session.target" ];
+    };
+  };
+
+  meta.maintainers = [ maintainers.GaetanLepage ];
```
rycee, line 46: By convention, the `meta` attribute goes on the top, before `option`.

modules/services/autorandr.nix
```diff
@@ -0,0 +1,47 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let cfg = config.services.autorandr;
+
+in {
+  options = {
```
rycee, line 8:
```suggestion
  meta.maintainers = [ maintainers.GaetanLepage ];

  options = {
```

rycee:
Thanks! I've added a few comments.

modules/services/autorandr.nix
GaetanLepage: I agree, it would be even better. Do you know what would be the corresponding `systemd` target ?

modules/services/autorandr.nix
GaetanLepage: Fixed

modules/services/autorandr.nix
GaetanLepage: Done :)

modules/services/autorandr.nix
rycee: Sorry, I have no idea how to hook that up. I always use the builtin laptop screen.

rycee:
Thanks! Merged to master now.

PR #3630, opened 2023-01-30
mbsync: make passwordCommand escaping consistent

IWYpPtD49k1OHyy3AqwT9vOBJEaZ4S:
### Description

`mbsync` treats option `accounts.email.accounts.<name>.passwordCommand` like a shell command, whereas other command like `himalaya`, `getmail` and `aerc` treat it like a command to be executed directly. This PR makes `mbsync`'s behavior consistent with other mail clients.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/mbsync.nix
+++ b/modules/programs/mbsync.nix
@@ -84,7 +84,7 @@ let
     genSection "IMAPAccount ${name}" ({
       Host = imap.host;
       User = userName;
-      PassCmd = lib.concatMapStringsSep " " lib.escapeShellArg passwordCommand;
+      PassCmd = lib.escapeShellArgs passwordCommand;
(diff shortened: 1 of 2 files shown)


modules/programs/mbsync.nix
```diff
@@ -84,7 +84,7 @@ let
     genSection "IMAPAccount ${name}" ({
       Host = imap.host;
       User = userName;
-      PassCmd = toString passwordCommand;
+      PassCmd = lib.concatMapStringsSep " " lib.escapeShellArg passwordCommand;
```
ncfavier, line 87:
```suggestion
      PassCmd = lib.escapeShellArgs passwordCommand;
```

ncfavier (approved):

PR #3636, opened 2023-02-01
xsession: cleanup systemd variables

arcnmx:
### Description

The xsession module imports variables into the systemd user session, but does not clean up after itself once the session ends. This can lead to issues where a `DISPLAY` variable remains set for services while X is not available or running anymore.

In one particular case when using xresources, [it can cause unnecessary switch failures](https://github.com/nix-community/home-manager/blob/782cb855b2f23c485011a196c593e2d7e4fce746/modules/xresources.nix#L96).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/xsession.nix
+++ b/modules/xsession.nix
@@ -78,6 +78,7 @@ in {
 
       importedVariables = mkOption {
         type = types.listOf (types.strMatching "[a-zA-Z_][a-zA-Z0-9_]*");
+        apply = unique;
         example = [ "GDK_PIXBUF_ICON_LOADER" ];
         visible = false;
         description = ''
@@ -184,7 +185,7 @@ in {
 
       ${optionalString (cfg.importedVariables != [ ])
       ("systemctl --user import-environment "
-        + toString (unique cfg.importedVariables))}
+        + escapeShellArgs cfg.importedVariables)}
 
       ${cfg.profileExtra}
 
@@ -212,6 +213,10 @@ in {
         while [ -n "$(systemctl --user --no-legend --state=deactivating list-units)" ]; do
           sleep 0.5
         done
+
+        ${optionalString (cfg.importedVariables != [ ])
+        ("systemctl --user unset-environment "
+          + escapeShellArgs cfg.importedVariables)}
       '';
     };
   };
--- a/tests/modules/misc/xsession/basic-xprofile-expected.txt
+++ b/tests/modules/misc/xsession/basic-xprofile-expected.txt
@@ -9,7 +9,7 @@ fi
 # script starts up graphical-session.target.
 systemctl --user stop graphical-session.target graphical-session-pre.target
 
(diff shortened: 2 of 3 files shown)


modules/xsession.nix
```diff
@@ -212,6 +212,10 @@ in {
         while [ -n "$(systemctl --user --no-legend --state=deactivating list-units)" ]; do
           sleep 0.5
         done
+
+        ${optionalString (cfg.importedVariables != [ ])
+        ("systemctl --user unset-environment "
+          + toString (unique cfg.importedVariables))}
```
ncfavier, line 218:
```suggestion
          + escapeShellArgs (unique cfg.importedVariables))}
```

ncfavier: I also don't think we need that `unique`. (Remember it has quadratic complexity.)

arcnmx: This was copied from the same file above when it uses `import-environment`, so I'll apply the change to both places it's used.

arcnmx: > I also don't think we need that `unique`. (Remember it has quadratic complexity.)

surprisingly, it is somewhat required:

```bash
:;  AAA=5 systemctl --user import-environment AAA AAA
Failed to import environment: Invalid environment assignments

:; systemctl --user unset-environment AAA AAA
Failed to set environment: Invalid environment variable names or assignments
```

ncfavier: Then maybe we can at least share the `unique` call. (Maybe even move it to an `apply` in the option?)

arcnmx: `apply` seems sane to me, done!

(if you're really worried about performance that much, something dumb like `uniqueStrings = flip pipe [ (flip genAttrs id) attrNames ]` is also an option)

ncfavier (approved):

ncfavier:
Does this interact well with logging out and back in? I don't have a clear view of when `.xprofile` and `.xsession` are typically run.

arcnmx:
So it does, yeah? xsession is a blocking script that, assuming a graceful logout, cleans up after itself by stopping services and (now) unsetting the variables it sets before starting the window manager. Login managers will execute xsession for the user, and log out when it exits. More manual setups using `startx` typically involve xinitrc just sourcing xsession, so the same occurs...

As far as I can tell, xsession in h-m is responsible for managing the state of `graphical-session.target`. h-m does not separate xsession and xprofile, so they cannot be enabled separately such that it would import the variables but not unset them, unless a login manager were to only source xprofile and not xsession - then this cleanup logic wouldn't run. The cleanup could be done as part of a systemd user service on ExecStop that binds to `graphical-session.target`, however you'd still run into the same issue - the target is unlikely to be stopped if xsession wasn't run to begin with.

Personally I've only tested (and experienced the issue that prompted these changes) on my systems that use lightdm and startx. I don't know if there are setups that only use xprofile without xsession, but at the least they're no worse off now than they were before this change ><

PR #3638, opened 2023-02-03
modules: java: fix setting JAVA_HOME

spikespaz:
### Description

Some JVMs pass through `home` as a derivation rather than as a string, as `openjdk` does. Since the module option for session variables expects a string, this is a type error. I suspect that this incorrect, and have changed the assignment here to coerce the `cfg.package.home` attribute to a string to be safe.

After discussing with @NobbZ, we have decided it is best to mitigate this problem in HM rather than to make potentially breaking changes to Nixpkgs.

Please do mention if you think we ought to propose a change to Nixpkgs instead.

### Checklist

- [ ] ~~Code formatted with `./format`.~~

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/java.nix
+++ b/modules/programs/java.nix
@@ -36,6 +36,8 @@ in {
   config = mkIf cfg.enable {
     home.packages = [ cfg.package ];
 
-    home.sessionVariables.JAVA_HOME = cfg.package.home;
+    # some instances of `jdk-linux-base.nix` pass through `result` without turning it onto a path-string.
+    # while I suspect this is incorrect, the documentation is unclear.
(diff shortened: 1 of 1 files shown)


ncfavier (approved):

rycee:
@spikespaz Thanks for the contribution! For future reference, please wrap the commit message body at 72 characters as per [rule 6](https://cbea.ms/git-commit/#wrap-72) 🙂

spikespaz:
Sorry, GitHub did that. Because the commit was made on the website (green check) and I filled out the issue description at the same time, it assigned the same message to the commit body.

PR #3646, opened 2023-02-07
home-manager: Use `path:` URI type for flake default

dongcarl:
```
Nix interprets a path-like URI as a git repository if any of the path's parents is a git repository. Since home-manager uses a default flake URI of ~/.config/nixpkgs/flake.nix, if you have a git repository as your home directory and a '*' .gitignore it leads to the following problems:

    evaluating derivation 'git+file:///Users/dongcarl?dir=.config%2fnixpkgs#homeConfigurations."dongcarl".activationPackage'
    The following paths are ignored by one of your .gitignore files:
    .config

This is solved by explicitly specifying the `path:` URI type prefix for the default flake URI argument.
```

I hope that this fix gets backported to release branches.

### DIFF
--- a/home-manager/home-manager
+++ b/home-manager/home-manager
@@ -156,7 +156,7 @@ function setFlakeAttribute() {
         fi
 
         if [[ -v configFlake ]]; then
-            FLAKE_ARG="$(dirname "$(readlink -f "$configFlake")")"
+            FLAKE_ARG="path:$(dirname "$(readlink -f "$configFlake")")"
         fi
     fi
 


ncfavier:
Not sure I understand. The directory with your `flake.nix` in it should be a git repository anyway, otherwise Nix commands won't work?

> if you have a git repository as your home directory

I don't think Home Manager should support this use case, since it's kind of meant to avoid it...

dongcarl:
> Not sure I understand. The directory with your `flake.nix` in it should be a git repository anyway, otherwise Nix commands won't work?

Flake URIs seem to support mercurial, tarballs, paths, plain files, etc. So it doesn't need to be a git repository, in home-manager's default case, we only know that the `flake.nix` file exists, but not if it's in a mercurial, git, or just a plain path, so I think _a_ sane default is `path:` since that's the common denominator.

However, if `home-manager` wants to be opinionated in favor of git, an alternative solution is to at least print a more helpful message if it's detected that `$HOME/.config/nixpkgs` is not a git repository and prompt the user to not rely on the default and specify their own flake URI.

ncfavier (approved):

berbiche (approved):

rycee:
Seems reasonable. Will this cause any difference in behavior if you have `~/.config/nixpkgs` as a Git repository?

dongcarl:
> Seems reasonable. Will this cause any difference in behavior if you have `~/.config/nixpkgs` as a Git repository?

Just tried, no difference in behavior!

PR #3662, opened 2023-02-09
mcfly: don't install widget on limited terminals

marsam:
### Description

Otherwise we get the following warning for terminals with limited capabilities:

    bash: bind: warning: line editing not enabled
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/mcfly.nix
+++ b/modules/programs/mcfly.nix
@@ -78,11 +78,15 @@ in {
       home.packages = [ pkgs.mcfly ];
 
       programs.bash.initExtra = mkIf cfg.enableBashIntegration ''
-        eval "$(${pkgs.mcfly}/bin/mcfly init bash)"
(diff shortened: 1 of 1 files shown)


ncfavier:
Shouldn't this be fixed upstream?

marsam:
> Shouldn't this be fixed upstream?

I believe here is the right place to address it, since `eval` evaluates the contents regardless whether we wanted it to or not.

ncfavier:
But the thing we `eval` could perform that check itself, for example here:

https://github.com/cantino/mcfly/blob/8633cf2e9daf6bd40bfe50eee4c146c1a4725e4b/mcfly.bash#L3-L5

ncfavier:
Or here https://github.com/cantino/mcfly/blob/8633cf2e9daf6bd40bfe50eee4c146c1a4725e4b/mcfly.bash#L64-L65

PR #3664, opened 2023-02-09
himalaya: improve derivation for v0.7.X

soywod:
### Description

I tried to make the derivation safer and closer to the configuration file. I added missing options, added two services and added more tests. It should be ready for the new release v0.7.0.

Note: I am the author of the tool, and I feel confortable enough now with Nix to be part of maintainers in order to reduce delay between releases.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

cc @toastal


ambroisie:
You should make a separate commit to add yourself to the maintainers file.

soywod:
> You should make a separate commit to add yourself to the maintainers file.

Sure, I do it now!

ambroisie:
As for adding more options to the module, unless it provides more integration with the email infrastructure in home-manager, I believe it is frowned-upon to simply mirror every option into the module system, if instead a free-form type could suffice.

I have yet to take a deeper look at your changes, but just letting you know in advance :-).

soywod:
> As for adding more options to the module, unless it provides more integration with the email infrastructure in home-manager, I believe it is frowned-upon to simply mirror every option into the module system, if instead a free-form type could suffice.

Ow, I thought that explicitly writing down all options would make the module more self-documented and safer, but I also understand the fact to forward everything. Anyway I can rollback if needed, it is not a big deal.

toastal:
@rycee is this true? Would it be better to have the options _in_ Nix+home-manager when the upstream project maintainer is now doing and keeping a close eye on the home-manager project? The frowned-upon nature seemed particularly important for those not near the project and the configs break.

soywod:
Just for information, `v0.7.X` changes have been merged on [nixpkgs](https://github.com/NixOS/nixpkgs/pull/215330).

j0xaf:
I'd like to have this merged; one reason is that I need to set `folder-aliases`. The current implementation does not allow to do this, since it uses `//` as merge operation for user supplied settings and not `lib.recursiveUpdate`.

tests/modules/programs/himalaya/default.nix
```diff
@@ -1 +1,5 @@
-{ himalaya = ./himalaya.nix; }
```
ncfavier, line 1: `himalaya.nix` is now unused

modules/programs/himalaya.nix
```diff
@@ -1,140 +1,643 @@
...
+        standard output.
+      '';
+    };
+
+    imap-notify-cmd = mkOption {
+      type = nullOr str;
+      default = null;
+      example = literalExpression ''
+        notify-send "📫 <sender>" "\<subject>"
+      '';
```
ncfavier, line 302: That's not a `literalExpression`, that's a string.

modules/programs/himalaya.nix
```diff
@@ -1,140 +1,643 @@
...
+      description = literalExpression ''
+        Command used for real-time notifications. Available
+        placeholders: <literal><id></literal>,
+        <literal><sender></literal> or <literal><subject></literal>.
+      '';
```
ncfavier, line 307: `literalExpression` doesn't make sense in option descriptions.

modules/programs/himalaya.nix
```diff
@@ -1,140 +1,643 @@
...
+        notify-cmd.
+      '';
+    };
+
+    imap-watch-cmds = mkOption {
+      type = nullOr (listOf str);
+      default = null;
+      example = literalExpression ''
+        [ "mbsync -a" ]
+      '';
```
ncfavier, line 325:
```suggestion
      example = [ "mbsync -a" ];
```

ncfavier (changes requested):
I've only done a very superficial review, but the main concern is that this adds [too many options](https://nix-community.github.io/home-manager/index.html#sec-guidelines-valuable-options). Please see [RFC 42](https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md) and add freeform types as needed instead.

soywod:
@ncfavier I just push your changes. I removed all the options and used the `freeformType` instead.

@j0xaf I also replaced `//` by `lib.recursiveUpdate` and add a test to check that `folder-aliases` are correctly overridden.

The PR is ready for another review.

modules/programs/himalaya.nix
```diff
@@ -1,140 +1,263 @@
...
 
-  options = with lib; {
+  options = {
     programs.himalaya = {
-      enable = mkEnableOption "himalaya mail client";
+      enable = lib.mkEnableOption "Enable the Himalaya email client.";
 
-      package = mkOption {
-        type = types.package;
+      package = lib.mkOption {
```
ncfavier, line 105: Could use `mkPackageOption pkgs "himalaya" { }`.

ncfavier (approved):

tests/modules/programs/himalaya/basic.nix
```diff
@@ -0,0 +1,28 @@
...
+{
+  imports = [ ../../accounts/email-test-accounts.nix ];
+
+  accounts.email.accounts = {
+    "hm@example.com" = {
+      imap.port = 993;
+      smtp.port = 465;
+      himalaya.enable = true;
+      himalaya.backend = "deprecated";
+      himalaya.sender = "deprecated";
```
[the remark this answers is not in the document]
soywod: I will modify this for the next release

PR #3665, opened 2023-02-09
dircolors: Implement `terminalTypes` option (fix #3658)

### LINKED ISSUE #3658: request: dircolors terminal specific sections support
### Description

Currently, the `dircolors` module can't specify terminal specific sections. This makes it impossible to fully replicate the default configuration of dircolors with home-manager. In the default `.dir_colors` generated with `dircolors -p`, the `LS_COLORS` environment variable is set only when using these terminals:
```
TERM Eterm
TERM ansi
TERM *color*
TERM con[0-9]*x[0-9]*
TERM cons25
TERM console
TERM cygwin
TERM dtterm
TERM gnome
TERM hurd
TERM jfbterm
TERM konsole
TERM kterm
TERM linux
TERM linux-c
TERM mlterm
TERM putty
TERM rxvt*
TERM screen*
TERM st
TERM terminator
TERM tmux*
TERM vt100
TERM xterm*
```
However, in home-manager generated `.dir_colors` this list is omitted, and `LS_COLORS` is set regardless of the `TERM` environment variable. Please provide a way to replicate the default behavior of dircolors.

bsd-department:
### Description

In the default `dircolors -p` configuration, `LS_COLORS` is set only on specific terminals.

This will allow the user to set which terminals should have `LS_COLORS` set. If not set, the default list of supported terminals taken from `dircolors -p` will be used. Because the supported terminals list has to precede settings, this isn't possible to set through `extraConfig`. 

Fixes #3658

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/dircolors.nix
+++ b/modules/programs/dircolors.nix
@@ -43,6 +43,42 @@ in {
       '';
     };
 
+    terminalTypes = mkOption {
+      type = with types; listOf str;
+      # Default terminal types from `dircolors --print-database`.
+      default = [
+        "Eterm"
+        "ansi"
+        "*color*"
+        "con[0-9]*x[0-9]*"
+        "cons25"
+        "console"
+        "cygwin"
+        "dtterm"
+        "gnome"
+        "hurd"
+        "jfbterm"
+        "konsole"
+        "kterm"
+        "linux"
+        "linux-c"
+        "mlterm"
+        "putty"
+        "rxvt*"
+        "screen*"
+        "st"
+        "terminator"
+        "tmux*"
+        "vt100"
+        "xterm*"
+      ];
+      description = ''
+        List of terminal types for which
+        <xref linkend="opt-programs.dircolors.settings"/> should be applied.
+        If empty, settings are applied regardless of terminal type.
+      '';
+    };
+
     settings = mkOption {
       type = with types; attrsOf str;
       default = { };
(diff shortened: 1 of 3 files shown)


modules/programs/dircolors.nix
```diff
@@ -43,6 +43,42 @@ in {
...
+        "mlterm"
+        "putty"
+        "rxvt*"
+        "screen*"
+        "st"
+        "terminator"
+        "tmux*"
+        "vt100"
+        "xterm*"
+      ];
```
ambroisie, line 74: Shouldn't the default be empty, instead of having to synchronize with `dircolors`?

bsd-department:
To fully replicate the previous behavior - yes, but I doubt anyone would 
notice that if they didn't notice anything when porting their dircolors 
to home manager. I think setting this as the new default would be less 
surprising to new users. I can change it to be empty if such change is 
undesired.

modules/programs/dircolors.nix
```diff
@@ -43,6 +43,42 @@ in {
...
+      description = ''
+        List of terminal types for which
+        <option>programs.dircolors.settings</option> should be applied.
```
rycee, line 77:
```suggestion
        <xref linkend="opt-programs.dircolors.settings"/> should be applied.
```

rycee:
Thanks! Looks OK. Can avoid the unnecessary "fix" in the commit header, though

> dircolors: implement terminalTypes option
>
> Fixes #3658

Issue #3667, opened 2023-02-10
Add programs.ssh.package option


gmacon:
### Description

I'm interested in adding the option `program.ssh.package` so that I can more easily adjust what ssh client I'm using (`pkgs.openssh` most of the time, but `pkgs.openssh_gssapi` in a few places). I think the right thing to do is to make the default value `pkgs.openssh`, but I think that would be a backwards-breaking change, so I wanted to see what folks thought about this before I work on it.

sumnerevans:
It looks like the module doesn't even add the `ssh` package. I think it would be fine to add it and default to `pkgs.openssh`.

gmacon:
There's an open PR, #3668 that fixes this issue.

Issue #3375, opened 2022-10-31
vscode: manage extensions/settings without installing package


kubukoz:
### Description

I'm using the vscode module of HM to install extensions and configure settings, but I don't want to manage the program's installation itself with Nix (I want it to be self-updating, so I got it from brew). Could this be allowed by the module?

For now, I use this as a workaround:

```nix
{
  programs.vscode.package = pkgs.runCommand "dummy" { } "mkdir $out" // { pname = pkgs.vscode.pname; };
}
```

eclairevoyant:
This isn't specific to the vscode module. If you want to change how a program installs, you'll inevitably be using something like you showed or overlays (see https://github.com/nix-community/home-manager/issues/1375#issuecomment-653747701). To me this works as intended

ncfavier:
`pkgs.emptyDirectory` should work, but yeah, we should probably allow `package` options to be `null` in general. That would have to be coordinated with nixpkgs' `mkPackageOption` though.

ncfavier:
https://github.com/NixOS/nixpkgs/pull/232808

ncfavier:
I just realised there's a problem with this: the vscode module relies on `cfg.package` to determine the "flavour" (code/insiders/codium) and version of the package, which influences the location of the configuration and extension folders and the way the extensions are managed (a change in 1.74.0 required special handling). So simply allowing the option to be `null` wouldn't work.

You worked around this by adding a `pname` to your dummy package, and we could possibly add options `vscodeName` and `vscodeVersion` to allow specifying this independently from `package`, but I'm not sure this is worth it.

kubukoz:
fair enough, this isn't a blocker for me - the current state happens to work. It's a "nice to have" :)

eclairevoyant:
> I just realised there's a problem with this: the vscode module relies on `cfg.package` to determine the "flavour" (code/insiders/codium) and version of the package

Which is a design worth fixing as it should be handled by e.g. an option to specify the flavour

OliverJAsh:
I'm using the workaround mentioned above but I get this error. @kubukoz Did you find a solution for this?:

```
Activating installPackages
Activating onFilesChange
/nix/store/zhrjg6wxrxmdlpn6iapzpp2z2vylpvw5-home-manager.sh: line 125: /nix/store/jply1gcrcc4q48055849ai4w4harw28q-dummy/bin/vscode: No such file or directory
```

kubukoz:
[...]

PR #3668, opened 2023-02-10
ssh: install an ssh client

gmacon:
Fixes: #3667

### Description

Add `programs.ssh.package` option and install the ssh client. It's not clear to me if this change is backwards-compatible or not, but at least we have something concrete to look at now.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -103,7 +103,7 @@ in
       entries = mkOption {
         internal = true;
         type = types.listOf entryModule;
-        default = [];
+        default = [ ];
         description = "News entries.";
       };
     };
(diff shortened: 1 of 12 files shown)


somasis (approved):

rycee:
Thanks! Merged to master now.

thefloweringash:
Is there am easy way to turn this off? I want to use the version provided by Apple that provides integration with the macOS keychain. I tried setting `programs.ssh.package = null`, but that "is not of type `package'".

For now I'm using
```nix
{
  programs.ssh = { 
    package = pkgs.runCommand "home-manager-3668" {} ''
      mkdir $out
    '';
  };
}
```

ncfavier:
There should be, and I think we shouldn't merge this until there is. See also https://github.com/nix-community/home-manager/issues/3375

rycee:
I think using `programs.ssh.package = pkgs.emptyDirectory` is the best solution. You could always alias the `emptyDirectory` to something shorter or more descriptive, if you want. E.g., `programs.ssh.package = pkgs.nil` or `programs.ssh.package = pkgs.intentionallyLeftEmpty`.

ncfavier:
https://github.com/NixOS/nixpkgs/pull/232808

lilyball:
Personally, I think this should just be reverted. SSH is usually installed at the system level, not the user level. And if I, as a user, want to use a different SSH client, I can just add it to `home.packages` myself.

The `programs.ssh` module is all about the user-level configuration for SSH, not about making SSH available to the system. Just because I want to configure SSH for my user shouldn't implicitly change the SSH client I'm using.

ncfavier:
Suggesting a middle ground: https://github.com/nix-community/home-manager/pull/4016

PR #3669, opened 2023-02-11
thunderbird: support aliases

alois31:
### Description

The main email address, as well as all aliases, each get mapped to one identity in Thunderbird. All these identities are then linked to their account.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


alois31:
I'm not quite sure about how to best implement the backwards compatibility. Currently, I assume that all user settings are actually per-identity, like in the example in the documentation. Per-account settings could be implemented (in a not-quite-compatible way) by adding a separate configuration option for this, or perhaps by using a dummy id and, sniffing the key and then inserting the actual id (yuck).

rycee:
Thanks! This looks nice. It would be good to also expand the test case to check the new code.

Issue #506, opened 2018-12-20
feature request: programs.firejail


bbigras:
I guess we can't use wrappedBinaries with apps installed using home-manager.

https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/programs/firejail.nix

```
programs.firejail = {
    enable = true;
    wrappedBinaries = {
      dropbox = "${pkgs.dropbox}/bin/dropbox";
    };
  };
```

bbigras:
I wonder if we should give up this.

I'm not 100% sure but I think I heard that firejail wasn't that good. Anyone can confirm this?

Baitinq:
I think it'd be really cool if we could manage firejail through home-manager

tobiasBora:
Any news on this? It may be of help to solve this issue https://discourse.nixos.org/t/automatic-firejail-of-home-managers-librewolf-does-not-work/22291 Could someone reopen the ticket?

KaiserCalm:
Why is this a closed issue? This was never resolved. Stalebots are so horrible: https://nostalebots.xyz/

PR #3675, opened 2023-02-13
firejail: add module

VAWVAW:
Add the option to generate firejail wrappers in home-manager using the global firejail binary. (addresses #506)

### Description

This module wrapps binaries for firejail and adds them to `home.packages` with a slightly higher priority.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


jian-lin (commented):
Not familiar with firejail. Just some general suggestions.

modules/programs/firejail.nix
```diff
@@ -0,0 +1,99 @@
...
+
+# based on https://github.com/NixOS/nixpkgs/blob/nixos-22.11/nixos/modules/programs/firejail.nix
+with lib;
+
+let
+  cfg = config.programs.firejail;
+
+  wrappedBins = pkgs.runCommand "firejail-wrapped-binaries" {
+    preferLocalBuild = true;
+    allowSubstitutes = false;
```
jian-lin, line 11: `runCommandLocal` sets these two for you

modules/programs/firejail.nix
```diff
@@ -0,0 +1,99 @@
...
+    # override packages added from other modules in PATH
+    meta.priority = -5;
+  } ''
+    mkdir -p $out/bin
+    ${lib.concatStringsSep "\n" (lib.mapAttrsToList (command: value:
```
jian-lin, line 16: All `lib.` can be removed since you have `with lib;` at the beginning.

modules/programs/firejail.nix
```diff
@@ -0,0 +1,99 @@
...
+      let
+        opts = if builtins.isAttrs value then
```
jian-lin, line 18: `builtins.` can be removed since `isAttrs` is also in `lib`

VAWVAW:
I updated the PR to fix the firejail error in https://github.com/NixOS/nixpkgs/pull/211245#issuecomment-1407447976.

versedwildcat:
I have been using this module in my NixOS setup for the past week, works great.

However, is there any way to override the desktop icons (when it uses an absolute path) similar to the Firejail NixOS module?
I have tried merging the relevant code from nixpkgs's `firejail.nix` but the build fails, stating the .desktop file doesn't exist.

VAWVAW:
It should work now. But most desktop entries rely on the PATH anyway.

versedwildcat (approved):
Looks great! I have tested this module extensively these past weeks, no issues.

modules/programs/firejail.nix
```diff
@@ -0,0 +1,106 @@
...
+        You still need to install firejail globally using your package manager.
+        You will get file collisions if you put the actual application binary in
+        the environment (such as by adding the application package to
+        `home.packages`), and applications started via .desktop files are not
+        wrapped if they specify the absolute path to the binary.
+      '';
+    };
+  };
+
+  config = mkIf (cfg.wrappedBinaries != { }) {
```
WolfangAukang, line 101: Wouldn't it be a good idea to perform an assertion verifying that the path for the firejail binary exists (maybe with the `pathExists` builtin that should work)

ncfavier: That would be IFD. If we want an assertion, we could do something like [what the NixOS security wrappers module does](https://github.com/NixOS/nixpkgs/blob/1e2590679d0ed2cee2736e8b80373178d085d263/nixos/modules/security/wrappers/default.nix#L285-L286) to check this at build time.

WolfangAukang: @ncfavier what does IFD mean?

WolfangAukang: Got it, makes sense now. Yes, the idea is that as long the firejailBinary path is verified (currently it isn't), then that would work fine.

TheDragon44:
I came across this and see people have been running it successfully and then it went quiet - is it likely to be merged at some point?

KaiserCalm:
Really sad this didn't get merged. Having any kind of extra security on NixOS would have been really cool.

KaiserCalm:
@jian-lin is anything blocking this module from being merged, if so what coule be done to fix that?

KaiserCalm:
I have tested the module and it seems to be fully working as long as the nixos module is set to enabled.

dasehak:
@jian-lin, there is any reason why this module isn't merged?

s0me1newithhand7s:
> @jian-lin, there is any reason why this module isn't merged?

+

dasehak:
@ncfavier, @WolfangAukang, there is any reason why this module isn't merged? I see the discussion above, but I can't see how it's preventing this from being merged.

Goorzhel:
@VAWVAW, may you resolve the conflicts?

VAWVAW:
I am no longer using this module and have written a [wrapper](https://codeberg.org/vaw/nixos-lib/src/commit/419b511f3eda654c1fea15b8711f64c09bcef19c/lib/firejail.nix) instead. This allows the wrapped program to be passed to `.package` options and can handle overrides. So I'm closing this but feel free to rebase this yourself if you want to.

PR #3676, opened 2023-02-14
keychain: add nushell integration

CardboardTurkey:
### Description

There is [no equivalent](https://www.nushell.sh/book/thinking_in_nu.html#think-of-nushell-as-a-compiled-language) of `eval` in nushell. Instead, this commit parses the output of keychain and pipes it directly into `load-env`, inspired by the recommendation for [`ssh-agent`](https://www.nushell.sh/cookbook/misc.html#manage-ssh-passphrases).

I also haven't set the `SHELL` variable as I don't see why? `keychain` only seems to [check](https://github.com/funtoo/keychain/blob/master/keychain.sh#L437-L439) for `fish` and `csh`.

I've tested the functionality of this commit on my own computer and it works.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/keychain.nix
+++ b/modules/programs/keychain.nix
@@ -87,6 +87,14 @@ in {
       '';
     };
 
+    enableNushellIntegration = mkOption {
+      default = true;
+      type = types.bool;
+      description = ''
+        Whether to enable Nushell integration.
+      '';
+    };
+
     enableXsessionIntegration = mkOption {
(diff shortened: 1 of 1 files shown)


berbiche (approved):
Thanks!

Can't test but LGTM

rycee:
Merged to master now 🙂

PR #3679, opened 2023-02-17
docs: set default option goPath correctly

tmeijn:
In line 82:

```nix
goPath = if cfg.goPath != null then cfg.goPath else "go";
```

So the default is go, but this was not documented.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/go.nix
+++ b/modules/programs/go.nix
@@ -34,7 +34,7 @@ in {
 
       goPath = mkOption {
         type = with types; nullOr str;
-        default = null;
+        default = "go";
         example = "go";
         description = ''
           Primary <envar>GOPATH</envar> relative to


ncfavier:
Seems like the option is used in two ways:

- to determine where to add packages, this uses `go` by default
- to set the GOPATH environment variable, unless it's `null`

So changing the default would set GOPATH by default, which might be undesirable? ping @rvolosatovs

tmeijn:
@ncfavier Ahh I think you're right! I'm sorry, I thought it would just affect the documentation. I saw that in the docs it said default `null` while the "default" is actually `go`.

image

So thought I would make a quick contribution. Sorry for the inconvenience!

PR #3686, opened 2023-02-19
neovim: separate init.lua parts with newlines

stasjok:
Parts of init.lua are now separated with newlines.

### Description

Previously all parts of `init.lua` was joined without newlines.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/neovim.nix
+++ b/modules/programs/neovim.nix
@@ -384,7 +384,6 @@ in {
 
     programs.neovim.generatedConfigs = let
       grouped = lib.lists.groupBy (x: x.type) pluginsNormalized;
-      concatConfigs = lib.concatMapStrings (p: p.config);
       configsOnly = lib.foldl
         (acc: p: if p.config != null then acc ++ [ p.config ] else acc) [ ];
     in mapAttrs (name: vals: lib.concatStringsSep "\n" (configsOnly vals))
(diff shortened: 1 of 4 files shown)


modules/programs/neovim.nix
```diff
@@ -394,26 +393,32 @@ in {
...
+            '';
+          luaPluginsConfig =
+            lib.optionalString (hasAttr "lua" cfg.generatedConfigs)
+            cfg.generatedConfigs.lua;
+          luaRcContent = builtins.concatStringsSep "\n"
+            (builtins.filter (str: str != "") [
+              vimlConfig
+              cfg.extraLuaConfig
+              luaPluginsConfig
+            ]);
```
ncfavier, line 415: If we're going to do this, maybe we can even add headers as comments above each section?

```nix
''
  -- Legacy vim configuration
  ${optionalString ...}

  -- Lua configuration
  ...
''
```

stasjok: Usually I don't like comments with empty sections or more than two consecutive newlines (which is often a result from optional strings written this way). If I skip comments for empty strings, then it would be strange to have a comment with only one section, no? Also it is very important to not create any configs if all sections are empty.

It you think that this is a right way, I can try to do it.

teto:
This approach forces styling on the user. Doing nothing (as on master) grants more freedom. 
Also generating a comment block also can cause various issues down the line  depending on `type = "lua";` and the nixpkgs vimrc generator. This MR implements what used to exist for viml and I've removed in lua. 

The following snippet could instead be added to the wikii f you want to easily identify the blocks in your config:
```
  genBlockLua = title: content:
    ''
      -- ${title} {{{
      ${content}
      -- }}}
    '';

  luaPlugin = attrs: attrs // {
    type = "lua";
     config = lib.optionalString
       (attrs ? config && attrs.config != null) 
       (genBlockLua attrs.plugin.pname attrs.config)
     ;
  };
  
    defaultCompletionPlugins = with pkgs.vimPlugins; [
    (luaPlugin { plugin = nvim-cmp; })
    (luaPlugin { plugin = cmp-nvim-lsp; })
    (luaPlugin { plugin = cmp-nvim-lua; config = "-- this will generate a folded block "; })
 
```

stasjok:
This PR is not about style, it's about bugfix. With the following config:

```nix
{
  config = {
    programs.neovim = {
      enable = true;
      extraConfig = ''
        set sw=4
      '';
      extraLuaConfig = "vim.opt.expandtab = false";
      plugins = [{
        plugin = pkgs.emptyDirectory;
        type = "lua";
        config = ''
          vim.g.test_plugin2 = 1
        '';
      }];
    };
}
```

generated `init.lua` would be:

```lua
vim.cmd [[source /nix/store/00000000000000000000000000000000-nvim-init-home-manager.vim]]vim.opt.expandtab = falsevim.g.test_plugin2 = 1
```

I consider it an error.

It is working after `vim.cmd[[...]]`, because space is optional here in lua. But it most likely won't work without separator between `extraLuaConfig` and `generatedConfigs.lua`. User must add trailing newline, or space, or semicolon in `extraLuaConfig`. So what I did is:
1. Added two newlines after `vim.cmd[[...]]` (so it is separated with empty line, not configurable)
2. Added one newline between `extraLuaConfig` and `generatedConfigs.lua` (if user ends its `extraLuaConfig` with a trailing newline, it results in an empty line between config sections)

What I think should also be considered (I didn't do any of that so that this PR would be backward compatible):

1. Move `extraLuaConfig` first
2. Add another extra config at the end
3. Add `runtime` configuration similar to plugin's, but in top level

I think that user should be able to provide its own configuration before all plugin configurations (hence move `extraLuaConfig` to the top, because `vim.cmd [[...]]` contains plugin configs written in vimscript). Also it should be able to add config after all plugin configurations (in order to override something). To achieve all of that in my config I specified dummy empty plugins (one firsts with lua, one last with lua, and one with runtime files not linked to any plugin).

stasjok:
The intention of this PR was to force a newline between config section. This was addressed in #7219.

PR #3688, opened 2023-02-19
carapace: add module

bobvanderlinden:
This introduces the programs.carapace.* options to use carapace-bin as the way for bash, zsh and fish to handle shell autocompletion. This aligns completion support for different executables across the shells.

The integration of carapace-bin into the shells in this PR is based on the instructions from the carapace-bin readme:
https://github.com/rsteube/carapace-bin#getting-started

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/carapace.nix
+++ b/modules/programs/carapace.nix
@@ -1,0 +1,98 @@
+{ config, pkgs, lib, ... }:
+
+let
+
+  inherit (lib)
+    mkEnableOption mkPackageOption mkIf pipe fileContents splitString;
+  cfg = config.programs.carapace;
+  bin = cfg.package + "/bin/carapace";
+
+in {
+  meta.maintainers = with lib.maintainers; [ weathercold bobvanderlinden ];
+
+  options.programs.carapace = {
+    enable =
+      mkEnableOption "carapace, a multi-shell multi-command argument completer";
+
+    package = mkPackageOption pkgs "carapace" { };
+
+    enableBashIntegration = mkEnableOption "Bash integration" // {
+      default = true;
+    };
+
+    enableZshIntegration = mkEnableOption "Zsh integration" // {
+      default = true;
+    };
+
+    enableFishIntegration = mkEnableOption "Fish integration" // {
+      default = true;
+    };
+
+    enableNushellIntegration = mkEnableOption "Nushell integration" // {
+      default = true;
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    programs = {
+      bash.initExtra = mkIf cfg.enableBashIntegration ''
+        source <(${bin} _carapace bash)
+      '';
+
+      zsh.initExtra = mkIf cfg.enableZshIntegration ''
+        source <(${bin} _carapace zsh)
+      '';
+
+      fish.interactiveShellInit = mkIf cfg.enableFishIntegration ''
+        ${bin} _carapace fish | source
+      '';
+
+      nushell = mkIf cfg.enableNushellIntegration {
+        extraEnv = ''
+          let carapace_cache = "${config.xdg.cacheHome}/carapace"
+          if not ($carapace_cache | path exists) {
+            mkdir $carapace_cache
+          }
+          ${bin} _carapace nushell | save -f $"($carapace_cache)/init.nu"
+        '';
+        extraConfig = ''
+          source ${config.xdg.cacheHome}/carapace/init.nu
+        '';
+      };
+    };
+
+    xdg.configFile =
+      mkIf (config.programs.fish.enable && cfg.enableFishIntegration) (
+        # Convert the entries from `carapace --list` to empty
+        # xdg.configFile."fish/completions/NAME.fish" entries.
+        #
+        # This is to disable fish builtin completion for each of the
+        # carapace-supported completions It is in line with the instructions from
+        # carapace-bin:
+        #
+        #   carapace --list | awk '{print $1}' | xargs -I{} touch ~/.config/fish/completions/{}.fish
+        #
+        # See https://github.com/rsteube/carapace-bin#getting-started
+        let
+          carapaceListFile = pkgs.runCommandLocal "carapace-list" {
+            buildInputs = [ cfg.package ];
+          } ''
+            ${bin} --list > $out
+          '';
+        in pipe carapaceListFile [
+          fileContents
+          (splitString "\n")
+          (map (builtins.match "^([a-z0-9-]+) .*"))
+          (builtins.filter
+            (match: match != null && (builtins.length match) > 0))
+          (map (match: builtins.head match))
+          (map (name: {
+            name = "fish/completions/${name}.fish";
+            value = { text = ""; };
+          }))
+          builtins.listToAttrs
+        ]);
+  };
+}
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -1207,6 +1207,13 @@ in
           A new module is available: 'programs.yazi'.
         '';
       }
+
+      {
+        time = "2023-09-05T06:38:05+00:00";
+        message = ''
+          A new module is available: 'programs.carapace'.
+        '';
+      }
     ];
   };
 }
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -67,6 +67,7 @@ let
     ./programs/broot.nix
     ./programs/browserpass.nix
     ./programs/btop.nix
+    ./programs/carapace.nix
     ./programs/chromium.nix
     ./programs/command-not-found/command-not-found.nix
     ./programs/comodoro.nix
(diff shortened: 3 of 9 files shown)


bobvanderlinden:
I tried to add a test that actually checks for `completion -p gh`, as bash doesn't shit ``gh` auto-completion by default, but carapace does. But it wasn't clear to me how to run bash commands within the test. I falled back to checking the bashrc file contents.

modules/programs/carapace.nix
```diff
@@ -0,0 +1,62 @@
...
+      source <(${cfg.package}/bin/carapace _carapace)
+    '';
+
+    programs.zsh.initExtra = mkIf cfg.enableZshIntegration ''
+      eval (${cfg.package}/bin/carapace _carapace|slurp)
+    '';
+
+    programs.fish.shellInit = mkIf cfg.enableFishIntegration ''
+      mkdir -p ~/.config/fish/completions
+      ${cfg.package}/bin/carapace --list | awk '{print $1}' | xargs -I{} touch ~/.config/fish/completions/{}.fish # disable auto-loaded completions (#185)
```
berbiche, line 58: This should probably be done in an activation script or by using import-from-derivation (`home.file.something.source = pkgs.runCommand ...`).

tests/modules/programs/carapace/basic-configuration.nix
```diff
@@ -0,0 +1,13 @@
+{ config, pkgs, ... }:
+
+{
+  config = {
+    programs.bash.enable = true;
+    programs.carapace.enable = true;
```
berbiche, line 6: The package needs to be stubbed so as not to be downloaded.
See other tests for examples of using `nmt.stubs`

berbiche (changes requested):
Thanks!

modules/programs/carapace.nix
bobvanderlinden: Hmm I have now implemented this, but it now conflicts with  `.config/fish/completions` as a whole, instead of individual files.

Since this mechanism is merely to disable builtin autocompletion from fish itself and use carapace instead, it doesn't seem that nice.

I have rewritten it a bit so it now creates individual home.file entries based on the output of `carapace --list`.

bobvanderlinden:
@berbiche could you have another look?

tests/modules/programs/carapace/basic-configuration.nix
```diff
@@ -0,0 +1,16 @@
+{ ... }:
+
+{
+  programs.bash.enable = true;
+  programs.carapace.enable = true;
+
+  # Avoid testing fish integration, because it actually
+  # needs the carapace package to be available.
+  programs.carapace.enableFishIntegration = false;
```
rycee, line 9: Could remove this if adopting the suggestion above.

```suggestion
```

modules/programs/carapace.nix
```diff
@@ -0,0 +1,87 @@
...
+        } ''
+          carapace --list > $out
+        '';
+      in pipe carapaceListFile [
+        fileContents
+        (splitString "\n")
+        (builtins.map (builtins.match "^([a-z0-9-]+) .*"))
+        (builtins.filter (match: match != null && (builtins.length match) > 0))
+        (builtins.map (match: builtins.head match))
+        (builtins.map (name: {
```
rycee, line 80: At least map should be possible to use directly:

```suggestion
        (map (builtins.match "^([a-z0-9-]+) .*"))
        (builtins.filter (match: match != null && (builtins.length match) > 0))
        (map (match: builtins.head match))
        (map (name: {
```

loicreynier:
@bobvanderlinden are you still working on this?

Weathercold:
@bobvanderlinden You can add nushell integration and the tests I wrote from my PR https://github.com/nix-community/home-manager/pull/4374

Weathercold:
Also I suggest adding a separate option to prefer fish's builtin completions. Native completions are usually better, an example being the `nix` command.

bobvanderlinden:
> Also I suggest adding a separate option to prefer fish's builtin completions. Native completions are usually better, an example being the `nix` command.

How is that different from `enableFishIntegration`? Or are you suggesting making it possible to enable/disable individual completions?

Weathercold:
What I meant was to add an option to not override Fish's native completions since they are usually better. For example, carapace's `nix` completion is incomplete and lacks flake attribute evaluation.

On second thought, it's far more flexible to allow disabling individual completions when someone eventually implements settings configuration.

modules/programs/carapace.nix
```diff
@@ -0,0 +1,88 @@
...
+          }
+          ${bin} _carapace nushell | save -f $"($carapace_cache)/init.nu"
+        '';
+        extraConfig = ''
+          source ${config.xdg.cacheHome}/carapace/init.nu
+        '';
+      };
+    };
+
+    home.file = mkIf (config.programs.fish.enable && cfg.enableFishIntegration) (
```
Weathercold, line 58: put it in `config`, otherwise it will be enabled for all fish users

use `xdg.configFile` instead and remove `.config/` below

modules/programs/carapace.nix
```diff
@@ -0,0 +1,88 @@
...
+      # Convert the entries from `carapace --list` to empty
+      # home.file.".config/fish/completions/NAME.fish" entries.
+      #
+      # This is to disable fish buildin completion for each of the
```
Weathercold, line 62: > buildin

small typo here

Weathercold (changes requested):

modules/programs/carapace.nix
bobvanderlinden: > put it in `config`, otherwise it will be enabled for all fish users

I'm not exactly sure what you mean by this? Both `config.programs.fish.enable` and `cfg.enableFishIntegration` need to be enabled for this to be the case?

> use `xdg.configFile` instead and remove `.config/` below

Good suggestion :+1:. Updated the PR.

bobvanderlinden:
> What I meant was to add an option to not override Fish's native completions since they are usually better. For example, carapace's `nix` completion is incomplete and lacks flake attribute evaluation.

What would be the difference from `enableFishIntegration = false`?

> On second thought, it's far more flexible to allow disabling individual completions when someone eventually implements settings configuration.

I'd like to look into this in a separate PR. Currently it's already quite a lot and I'm guessing it's not clear yet how the options should be named or where they should reside (`programs.fish.completions.git.enable = false`? `programs.carapace.fishCompletion.git = false`?).

modules/programs/carapace.nix
Weathercold: `cfg.enableFishIntegration` is enabled by default, so these empty files will be generated just by enabling fish. You need to put `xdg.configFile` in the `config` attrset above which also checks for `cfg.enable`.

Also, there's one mention of `home.file` left in the comments

bobvanderlinden:
I've now added `overrideFishCompletions` and `overrideFishCompletion.${completion}` so that it is possible to override all fish completions with carapace and be able to specify which ones specifically.

I'm unsure about the naming and logic to explain a user what it is it is doing. Suggestions are welcome. To me it also sounds good to remove this functionality for now and move the discussion to a separate PR, so that the carapace module itself can already be merged.

modules/programs/carapace.nix
bobvanderlinden: > `cfg.enableFishIntegration` is enabled by default, so these empty files will be generated just by enabling fish. You need to put `xdg.configFile` in the `config` attrset above which also checks for `cfg.enable`.

https://github.com/bobvanderlinden/home-manager/blob/361cf75cc6eb55c0e215927b313401631727c440/modules/programs/carapace.nix#L79

is already inside

https://github.com/bobvanderlinden/home-manager/blob/361cf75cc6eb55c0e215927b313401631727c440/modules/programs/carapace.nix#L44

I'm probably misunderstanding :sweat_smile:, can you make a code suggestion to explain?

> Also, there's one mention of `home.file` left in the comments

Good catch :+1: I've updated the PR.

Weathercold: It's all good now 👍

Weathercold:
I agree with moving `overrideFishCompletions` and `overrideFishCompletion.${completion}` to a different PR

bobvanderlinden:
Thanks for the quick replies :smile: I removed the 2 commits regarding fish completion overrides.

Weathercold (approved):
Tested Fish & Nushell

rycee:
I interpret the last few comments as meaning the PR is complete and it looks good to me so I merged, hope you don't mind 🙂

bobvanderlinden:
Awesome! Thanks 👍

adamcstephens:
It would be nice to have the targeted completion overrides, as right now the fish integration has IFD...

bobvanderlinden:
> It would be nice to have the targeted completion overrides, as right now the fish integration has IFD...

The proposed completion overrides would still rely on IFD. There were other methods that didn't rely on IFD, but those would block changes to the fish completion directory.

Do you have a good alternative?

PR #3697, opened 2023-02-22
[Backport release-22.11] starship: add nushell integration #3701

Lord-Valen:
### Description

Backport #3347, #3519, #3528, #3701 to `release-22.11`.
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/starship.nix
+++ b/modules/programs/starship.nix
@@ -58,44 +58,24 @@ in {
       '';
     };
 
-    enableBashIntegration = mkOption {
+    enableBashIntegration = mkEnableOption "Bash integration" // {
       default = true;
-      type = types.bool;
-      description = ''
-        Whether to enable Bash integration.
-      '';
     };
 
-    enableZshIntegration = mkOption {
+    enableZshIntegration = mkEnableOption "Zsh integration" // {
       default = true;
-      type = types.bool;
-      description = ''
-        Whether to enable Zsh integration.
-      '';
     };
 
-    enableFishIntegration = mkOption {
+    enableFishIntegration = mkEnableOption "Fish integration" // {
       default = true;
-      type = types.bool;
-      description = ''
-        Whether to enable Fish integration.
-      '';
     };
 
-    enableIonIntegration = mkOption {
+    enableIonIntegration = mkEnableOption "Ion integration" // {
       default = true;
-      type = types.bool;
-      description = ''
-        Whether to enable Ion integration.
-      '';
     };
 
-    enableNushellIntegration = mkOption {
+    enableNushellIntegration = mkEnableOption "Nushell integration" // {
       default = true;
-      type = types.bool;
-      description = ''
-        Whether to enable Nushell integration.
-      '';
     };
   };
 
--- a/modules/programs/starship.nix
+++ b/modules/programs/starship.nix
@@ -124,6 +124,12 @@ in {
       end
     '';
 
+    programs.ion.initExtra = mkIf cfg.enableIonIntegration ''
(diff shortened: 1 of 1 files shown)


modules/programs/starship.nix
```diff
@@ -89,6 +89,14 @@ in {
...
       '';
     };
+
+    enableNushellIntegration = mkOption {
+      default = true;
+      type = types.bool;
+      description = ''
+        Whether to enable Nushell integration.
+      '';
+    };
```
sumnerevans, line 99:
```suggestion

    enableNushellIntegration = mkEnableOption "Nushell integration" // {
      default = true;
    };
```

sumnerevans (changes requested):

sumnerevans:
Please also squash these commits together (probably good to add `Co-authored-by` lines for all of the contributors to this PR).

modules/programs/starship.nix
rycee: Will need to be fixed in the master branch first :-)

rycee:
For cherry-picks I think it is best to merge individual commits, That allows to track the individual commits.

@Lord-Valen From what I can tell the ion integration was never removed from the 22.11 branch? This PR only seems to add the nushell support?

Lord-Valen:
> @Lord-Valen From what I can tell the ion integration was never removed from the 22.11 branch? This PR only seems to add the nushell support?

Yes, that's right. The ion integration was removed in master while fixing the (newly added) nushell integration, thus it was necessary to include the commit which adds it to have both the ion integration and the (fixed) nushell integration. This isn't apparent from the diff and will be less apparent with the commits squashed, thus the PR title should probably be changed.

Lord-Valen:
> For cherry-picks I think it is best to merge individual commits, That allows to track the individual commits.

Am I correct in interpreting this to mean that you would not recommend squashing?

Lord-Valen:
I imagine #3678 is intended to fix whatever is causing CI to fail?

ncfavier:
Should be fixed now, if you rebase

modules/programs/starship.nix
Lord-Valen: Resolved by https://github.com/nix-community/home-manager/commit/0b0986f0c054b5e1358ab8ce2825b00897eb7da8.

felschr:
This doesn't work for me on nixos-22.11 & home-manager release-22.11:

```console
❯ nu                       
Error: nu::parser::unknown_flag (link)

  × The `save` command doesn't have flag `force`.
   ╭─[/home/felschr/.config/nushell/env.nu:4:1]
 4 │ }
 5 │ /etc/profiles/per-user/felschr/bin/starship init nu | save --force /home/felschr/.cache/starship/init.nu
   ·                                                            ───┬───
   ·                                                               ╰── unknown flag
   ╰────
  help: use save --help for a list of flags

Error: nu::parser::sourced_file_not_found (link)

  × File not found
   ╭─[/home/felschr/.config/nushell/config.nu:7:1]
 7 │ 
 8 │ source /home/felschr/.cache/starship/init.nu
   ·        ──────────────────┬──────────────────
   ·                          ╰── File not found: /home/felschr/.cache/starship/init.nu
   ╰────
  help: sourced files need to be available before your script is run
```

As described in https://github.com/nix-community/home-manager/pull/3519#pullrequestreview-1231637445, `--force` was added to `save` due to a change in nushell >= 0.73 But nushell in nixos-22.11 is still on 0.71.

ncfavier:
Could you try out https://github.com/nix-community/home-manager/pull/3777 please?

PR #3699, opened 2023-02-24
rofi: allow to actually override package

n0emis:
allow to override the rofi-package and not strip the override

### Description

allow to override the rofi-package and not strip the override.
before, when a custom package with override was set, the own override was added, but not when no override was set

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/rofi.nix
+++ b/modules/programs/rofi.nix
@@ -266,9 +266,9 @@ in {
       rofiWithPlugins = cfg.package.override
         (old: rec { plugins = (old.plugins or [ ]) ++ cfg.plugins; });
     in if builtins.hasAttr "override" cfg.package then
-      rofiWithPlugins
+      cfg.package
     else
-      cfg.package;
+      rofiWithPlugins;
 
     home.packages = [ cfg.finalPackage ];
 


modules/programs/rofi.nix
```diff
@@ -266,9 +266,9 @@ in {
       rofiWithPlugins = cfg.package.override
         (old: rec { plugins = (old.plugins or [ ]) ++ cfg.plugins; });
     in if builtins.hasAttr "override" cfg.package then
-      rofiWithPlugins
+      cfg.package
     else
-      cfg.package;
+      rofiWithPlugins;
```
ncfavier, line 271: I think the order was correct before? We can only use `cfg.package.override` if `cfg.package` has an `override` attribute...

ncfavier:
Could you give more detail about the problem you encountered? AFAICT there's nothing wrong with the module.

n0emis:
Ah, okay. Im sorry. I misunderstood this as: When a override is set on the package, then apply the custom override. When no override has been specified on cfg.package. apply no override

ncfavier:
Right, `override` is a function that returns a new package with the applied override. It doesn't contain the override itself.

PR #3700, opened 2023-02-24
fix: swaylock wait option removed

ikerurda:
### Description

According to the manual:

>  -w      Wait for command to finish executing
>            before continuing, helpful for ensur‐
>            ing that a before-sleep command has
>            finished before the system goes to
>            sleep.

So this option doesn't allow execution of swaylock and dpms off as it will wait for the swaylock command to finish

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/swayidle.nix
+++ b/modules/services/swayidle.nix
@@ -124,7 +124,7 @@ in {
         # swayidle executes commands using "sh -c", so the PATH needs to contain a shell.
         Environment = [ "PATH=${makeBinPath [ pkgs.bash ]}" ];
         ExecStart =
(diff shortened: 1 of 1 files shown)


ncfavier:
I'm sure the flag was added for a reason, cc @c0deaddict. We should provide an option to disable it if that is needed for your use case.

Note that the [example](https://github.com/swaywm/swayidle/blob/master/swayidle.1.scd#example) from the docs uses both `-w` and `swaylock`, so I'm a bit confused.

ikerurda:
I am very sorry, you are right. I forgot the `-f` flag in swaylock

PR #3704, opened 2023-02-25
modules/git: make options passed to `less(1)` for `diff-so-fancy` configurable

Ma27:
### Description
The `-X` prevents that screen is cleared when showing a diff that's
larger than my screen.
    
I.e. when running `git diff` and press `q`, the last thing I want to see
is the prompt with `git diff` and *not* the part of the diff I browsed,
to be clear
    
    $ git diff
    $ <cursor>
    
Considering that this is somewhat opinionated, I decided to build an
option which allows you to pass arbitrary commands to the less
invocation.

cc @rycee 

### Checklist

- [ ] Code formatted with `./format`. (It's a one-liner)

- [x] Code tested through `nix-shell --pure tests -A run.all`. (`broot` failed, but that doesn't seem related).

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/git.nix
+++ b/modules/programs/git.nix
@@ -340,6 +340,14 @@ in {
           '';
         };
 
+        pagerOpts = mkOption {
+          type = types.listOf types.str;
+          default = [ "--tabs=4" "-RFX" ];
+          description = ''
+            Arguments to be passed to <command>less</command>.
+          '';
+        };
+
         markEmptyLines = mkOption {
           type = types.bool;
           default = true;
@@ -547,7 +555,9 @@ in {
       programs.git.iniContent =
         let dsfCommand = "${pkgs.diff-so-fancy}/bin/diff-so-fancy";
         in {
-          core.pager = "${dsfCommand} | ${pkgs.less}/bin/less --tabs=4 -RFX";
+          core.pager = "${dsfCommand} | ${pkgs.less}/bin/less ${
+              escapeShellArgs cfg.diff-so-fancy.pagerOpts
+            }";
           interactive.diffFilter = "${dsfCommand} --patch";
           diff-so-fancy = {
             markEmptyLines = cfg.diff-so-fancy.markEmptyLines;


ncfavier:
> The `-X` (a.k.a. `--no-init`) disables any additional settings

Are you sure? `less -FX` behaves as expected for me.

Ma27:
I have `less(1)` configured using the env-var `LESS` with the value `-r --quit-if-one-screen`.

`cat .zsh_history | less -FX` then `q` leaves the contents on the screen, whereas `cat .zsh_history | less -F` doesn't when hitting `q`.

ncfavier:
You realise that `-F` and `--quit-if-one-screen` are the same flag, yes?

Could you describe your issue more precisely: how do you want the pager to behave, and how is it behaving currently?

Ma27:
@ncfavier sorry for the confusion! While the change did what I wanted, I completely got the behavior of the options in use wrong. Will update the PR's description accordingly.

Basically `-X` doesn't clear the entire screen when browsing a diff that's larger than my screen and I want that behavior. Since this is somewhat opinionated (and other people may have even more wishes), I decided to make an option for that now.

modules/programs/git.nix
```diff
@@ -340,6 +340,15 @@ in {
           '';
         };
 
+        pagerOpts = mkOption {
+          type = types.listOf types.str;
+          default = [ "--tabs=4" "-RFX" ];
+          apply = concatMapStringsSep " " escapeShellArg;
```
ncfavier, line 346:
```suggestion
          apply = escapeShellArgs;
```

ncfavier (commented):
OK, this is reasonable.

Don't forget to `./format`.

modules/programs/git.nix
ncfavier: Well, actually I'd rather put this at the use site rather than in `apply`, so that pager options can still be inspected as a list.

Ma27:
> Don't forget to ./format.

On it :)

modules/programs/git.nix
Ma27: Done.

Ma27:
@ncfavier fixed both things :) 

(and sorry again for the confusion at first!)

ncfavier (approved):

Luflosi:
Does this qualify for backporting? I'd like to have this in release-22.11.

ncfavier:
https://github.com/nix-community/home-manager/pull/3751

Luflosi:
Awesome, thank you!

PR #3709, opened 2023-02-26
home-manager: move install command into main tool

rycee:
### Description

This simplifies things slightly since we can use the code infrastructure that is already present in the home manager tool.

We will name the sub-command to `init` and only do the activation if the user specifies a `--switch` option.

It is mandatory to specify the output directory, mainly intended as a way to move away from putting files in `~/.config/nixpkgs`.

We will automatically generate a `flake.nix` if Flake support is detected.

Finally, we update the installation instructions for the Nix Flakes standalone setup to use the new install sub-command.

Note, I didn't dare to update the Zsh and Fish completion files. @axionl, @anund Perhaps you could help with this? Basically, the init command takes either `--switch` or a directory path as argument.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


anund:
```git
diff --git a/home-manager/completion.zsh b/home-manager/completion.zsh
index c5d3b493..52bcce90 100644
--- a/home-manager/completion.zsh
+++ b/home-manager/completion.zsh
@@ -35,6 +35,7 @@ case "$state" in
       'edit[edit]' \
       'option[inspect option]' \
       'build[build]' \
+      'init[init]' \
       'switch[switch]' \
       'generations[list generations]' \
       'remove-generations[remove generations]' \
@@ -69,6 +70,11 @@ case "$state" in
           '--experimental-features[set experimental Nix features]:VALUE:()' \
           '--extra-experimental-features:[append to experimental Nix features]:VALUE:()'
         ;;
+      init)
+        _arguments \
+          '--switch[switch]' \
+          ':PATH:_files -/'
+        ;;
     esac
 esac
```

```zsh
❯ home-manager
-- command --
build               -- build
edit                -- edit
expire-generations  -- expire generations
generations         -- list generations
help                -- help
init                -- init
news                -- read the news
option              -- inspect option
packages            -- managed packages
remove-generations  -- remove generations
switch              -- switch
uninstall           -- uninstall
```

```zsh
❯ home-manager init -
-- option --
--switch  -- switch
-- PATH --
_home-manager     home-manager      home-manager.nix
```

This allows `home-manager init --switch <path>` or `home-manager init <path> --switch` in case that's relevant.

Notes for future testing. A way, perhaps not the best way, to test following the simplified instructions [zsh-completions](https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org)

```zsh
cp completion.zsh _home-manager # make the completion exist with the relevant filename convention
fpath=(<source-path>/home-manager $fpath) # set zsh fpath
unfunction _home-manager # unload existing handler if it's already there
autoload -U _home-manager # reload handler from _home-manager file
```

rycee:
Thanks @anund! I've incorporated that patch in the PR.

PR #3710, opened 2023-02-27
modules/redshift-gammastep: install package into the profile

Ma27:
### Description
This is needed to make sure that the tray icon is actually found. Otherwise an empty icon indicating that it wasn't found is shown.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/redshift-gammastep/lib/options.nix
+++ b/modules/services/redshift-gammastep/lib/options.nix
@@ -183,6 +183,8 @@ in {
     xdg.configFile.${xdgConfigFilePath}.source =
       settingsFormat.generate xdgConfigFilePath cfg.settings;
 
+    home.packages = [ cfg.package ];
+
     systemd.user.services.${moduleName} = {
       Unit = let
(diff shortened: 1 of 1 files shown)


modules/services/redshift-gammastep/lib/options.nix
```diff
@@ -183,6 +183,8 @@ in {
     xdg.configFile.${xdgConfigFilePath}.source =
       settingsFormat.generate xdgConfigFilePath cfg.settings;
 
+    home.packages = mkIf cfg.tray [ cfg.package ];
```
ncfavier, line 186:
```suggestion
    home.packages = [ cfg.package ];
```

Let's install it unconditionally, it seems to contain useful stuff.

cc @thiagokokada

Ma27: Also fine by me ;-)

thiagokokada (approved):

ncfavier (approved):

PR #3715, opened 2023-02-28
exa: added --icons to aliasses

realsnick:
added support to --icons


loicreynier:
IMO this should be an option rather than a default behavior since most default fonts do not support icons.

ncfavier:
I think https://github.com/nix-community/home-manager/pull/3505 is a much better approach to this.

PR #3717, opened 2023-02-28
programs.neovim: link packpath dir in XDG_DATA_HOME

teto:
link the packpath in expected folder so that even unwrapped neovim can pick home-manager's plugins.

I sometimes need to run neovim not wrapped/configured by nix (when developing neovim or when other projects bring their own neovim in PATH). Currently they dont find plugins installed by home-manager in the cases where packpath is not set to the generated nix packpath directory. With this change, neovim can discover HM-installed plugins by itself.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


teto:
Ok I'll merge as I think it's more representative of how people want neovim to work (including myself). Remember that you may need to change ~/.local/share/nvim/ away so that HM can deal with it along with your package manager (I use lazy for instance).

svrana:
I'm using a wrapped neovim and this change appears to cause one of my plugins? to be loaded twice. This results in getting an 'Already loaded' error message when I start neovim. To debug, I ran neovim like this `neovim -V2nvim.log` and searched for the 'Already loaded' string which I found at line 341

```
340 sourcing "/nix/store/r6h5damgnb72b2a9gf0hm0mlc09a0rn7-packdir-start/pack/myNeovimPackages/start/vim-latex-live-preview/plugin/latexlivepreview.vim"
Already loaded
```
(this is from the plugin named: vim-latex-live-preview)

If I search up the file for locations where it was previous loaded, I see it again at line 269. 

If I comment out the plugin or remove the site/ directory from XDG_STATE_HOME the error goes away. I've placed the log file at https://gist.github.com/svrana/5c68f834a88eb52fd11e726c0a089906

teto:
I think this is because your neovim (like everyone else) is wrapped with a `--cmd "set packpath..."` . so it loads the one at that path plus the one in the standard folder. If you remove this line in nixpkgs https://github.com/nixos/nixpkgs/blob/4bb072f0a8b267613c127684e099a70e1f6ff106/pkgs/applications/editors/neovim/utils.nix#L111 it should fix the issue (it was in my plan to make it configurable in nixpkgs). Until I fix this in nixpkgs, maybe rollback your HM version ? sry I am not sure how to fix cleanly in HM.

PatrickDaG:
I do not use home manager to manage my plugins, instead adding and configuring them directly in my vim config. This change forced the site directory to be a symlink to the nix store even though it is empty, which in turn prevents my vimconfig from being able to write it and add my plugins. I would suggest only doing this if there are plugins declared in the nix config, or making it optional.

adrianord:
I am facing the same issue as @PatrickDaG. AstroNvim is managing my plugins with Packer, not using nix for plugin management at all, and this symlink is causing my configuration to fail.

ncfavier:
I'm going to [revert](https://github.com/nix-community/home-manager/pull/3817) this for now until we figure this out

ncfavier:
https://github.com/nix-community/home-manager/commit/deb2f59b5c1fd11bec00600517ba7f51984c3090#commitcomment-106637065:

> Hi,
> I manage my plugins via [Packer.nvim](https://github.com/wbthomason/packer.nvim) and don't plan to migrate it to Home Manager. This way I can use my neovim config independently and just reference it from Home Manager to avoid duplication.
> 
> This change breaks my nix rebuild, since the site/pack directory generated by Packer is in the way of the new paths that are to be linked.
> 
> I'm not sure if there's a workaround to this or if it's just an issue in my setup. Has anyone an idea how to handle this w/o moving my neovim plugin management to Home Manager? Any help is much appreciated.

teto:
home-manager should create the fullpath `site/pack/HMpack` instead so it doesn't conflict. This can be done adjusting some stuff in nixpkgs, but we could have worked around it in HM.
```
        "nvim/site/pack/HmPackages" = {
          # text = "toto";
          source = "${pkgs.vimUtils.packDir neovimConfig.packpathDirs}/pack/HmPackages";
        };
```

Issue #3716, opened 2023-02-28
bug: Specialization does not work with nixos modules


zoriya:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Hello, I am trying to set up specializations for light/dark theme switching, but creating a specialization result in a build error.
```
error: The option `home-manager.users.zoriya.specialization.test.configuration._module.args.name' is defined multiple times while it's expected to be unique.

       Definition values:
       - In `/nix/store/3jhxmigz3b15xdcdhj0pryq2bg6qyhcv-source/modules/misc/specialization.nix': "zoriya"
       - In `<unknown-file>': "configuration"
       Use `lib.mkForce value` or `lib.mkDefault value` to change the priority on any of these definitions.
(use '--show-trace' to show detailed location information)`
```

The snippets that make my build crash is the specialization test"
```nix
nixpkgs.lib.nixosSystem {
  inherit system;
  specialArgs = inputs;
  modules = [
	{
	  home-manager = {
	    useGlobalPkgs = true;
	    useUserPackages = true;
	    extraSpecialArgs = inputs;
	    users.${user} = {
		    specialization.test.configuration = {
		      home.file.testfile.text = "very special";
		    };
		    home.stateVersion = "22.11";
	    };
	  };
	}
  ];
}
```

Link to the [full config](https://github.com/zoriya/flake/blob/059ff0118d539e5602f1319101234c0f221c2219/modules/home/colors/default.nix#L29)

### Maintainer CC

@rycee 

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.2.0, NixOS, 23.05 (Stoat), 23.05.20230222.988cc95`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.13.2`
 - channels(root): `"nixos"`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`
```

ncfavier:
That's a fun one! cc @roberth @infinisil

Minimal reproducer:

```nix
with import <nixpkgs/lib>;
let eval = evalModules {
  modules = [ {
    options.hm = mkOption {
      type = types.submoduleWith {
        modules = [ ({ moduleType, ... }: {
          options.foo = mkOption {
            type = types.str;
          };
          options.special = mkOption {
            type = moduleType;
          };
[...]

PR #3718, opened 2023-02-28
specialization: fix `name` conflict inside NixOS module

ncfavier:
If used inside the NixOS/nix-darwin module, we get conflicting definitions of `name` inside the specialization: one is the user name coming from the NixOS module definition and the other is `configuration`, the name of this option. Thus we need to explicitly wire the former into the module arguments.

Fixes https://github.com/nix-community/home-manager/issues/3716

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/specialization.nix
+++ b/modules/misc/specialization.nix
@@ -1,4 +1,4 @@
-{ config, extendModules, lib, ... }:
+{ config, name, extendModules, lib, ... }:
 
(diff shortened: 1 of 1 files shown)


roberth (approved):
I'm not a maintainer here, but as the author of `extendModules` I think this is a good solution.

PR #3729, opened 2023-03-02
home: Add buildEnvWithNoChroot to help avoid darwin sandbox failures

amarshall:
### Description

Sets `__noChroot = true` on select `buildEnv` derivations that assemble
large numbers of paths. This may be used to avoid sandbox failures on
darwin, see https://github.com/NixOS/nix/issues/4119 and the `sandbox`
option in `man nix.conf`.

I wish there was a way to do something akin to overlays for config, alas
there is not afaik, so the only way is to add an option. Since this is
opt-in, anyone enabling it thus understands the “risks” of disabling the
sandbox, however the risk for these derivations should be fairly low,
and this allows enabling the sandbox more generally on Darwin, which is
beneficial.

I have only added to the derivations that started giving me problems,
others may suffer from others but these are definitely likely to have
huge dependency lists therefore exposing the problem.

I’m open to a different name for the option, hard to come up with something representative but concise. I have not added any tests since it’s a rather cumbersome scenario to test, and may also be flaky. The default path (option disabled) is I expect already well-covered.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/home-environment.nix
+++ b/modules/home-environment.nix
@@ -511,6 +511,14 @@ in
       '';
     };
 
+    home.buildEnvWithNoChroot = lib.mkEnableOption ''
+      Sets <code>__noChroot = true</code> on select <code>buildEnv</code>
+      derivations that assemble large numbers of paths, as well the activation
+      script derivations. This may be used to avoid sandbox failures on Darwin,
+      see https://github.com/NixOS/nix/issues/4119 and the <code>sandbox</code>
+      option in <command>man nix.conf</command>.
+    '';
+
     home.preferXdgDirectories = lib.mkEnableOption "" // {
       description = ''
         Whether to make programs use XDG directories whenever supported.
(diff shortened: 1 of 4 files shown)


khaneliman:
I don't believe this is still needed, correct? I was able to enable my sandbox with "relaxed" again after the nix fix https://github.com/NixOS/nix/pull/12570 . I only haven't been able to fully enable due to a rust cargo fetcher bug that blocks builds in sandbox on darwin.

khaneliman:
> FYI the only thing `relaxed` does vs. `true` is disable sandbox for `__noChroot = true` derivations, so if you don’t have any of those, you can set `sandbox = true`. If that doesn’t work then, well, perhaps this patch is still useful :)

This is the issue I was referring to https://github.com/NixOS/nixpkgs/issues/394972

Issue #3640, opened 2023-02-04
bug: When leveraging the (documented) default path, clashes with nixos nixpkgs documentation -> causes infinite recursion


x10an14:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

When reading `man home-manager`, [it's specified that default path](https://github.com/nix-community/home-manager/commit/3a16ebdf726a7ba5abb073bb069fbe8344826e88#diff-d2e7efecbf081fdb2595288e62d73267b0b86b293b8c9eedbc7d8f0e06023776L163-R167) for my root `flake.nix` (which configures my `homeConfigurations`) is `~/.config/nixpkgs`. This is also referenced/hinted at (albeit without flakes being mentioned) in the official documentation online: https://nix-community.github.io/home-manager/index.html#ch-usage

It's therefore reasonable (I claim) that any user might have a folder structure somewhat like this (some might even say recommended for any beginners looking for templates to follow):
```
-> $ tree ~/.config/
~/.config/
└── nixpkgs
    ├── configs
    │   └── default.nix
    ├── flake.lock
    ├── flake.nix
    ├── modules
    │   └── default.nix
    ├── overlays
    │   └── default.nix
    └── pkgs
        └── default.nix

5 directories, 6 files
```

So, what happens when you add `~/.config/nixpkgs/overlays/default.nix` and import it from somewhere within your flakes config?
You hit this: 
```
nix-repl> :lf .
Added 17 variables.

nix-repl> homeConfigurations."x10an14@home-desktop".activationPackage
error: infinite recursion encountered

       at /nix/store/p63h4q5k9rn9ds37gqzdzz179yk3flpv-source/lib/fixed-points.nix:69:67:

           68|   #
           69|   extends = f: rattrs: self: let super = rattrs self; in super // f self super;
             |                                                                   ^
           70|

nix-repl>
```

Caused by this https://nixos.org/manual/nixpkgs/stable/#sec-overlays-lookup.

---

This breaks just about every `nix-` (read: non-flake non-nix-command style) command.
It's impossible to e.g. review nixpkgs PRs, etc.

`home-manager` command still work though - hence my surprise when this was the caused for all my `infinite recursion` issues.

<details>
<summary>Full stack-trace for recursion</summary>

```
-> $ nix repl
Welcome to Nix 2.12.0. Type :? for help.

nix-repl> :te
showing error traces

[...]

PR #3737, opened 2023-03-03
docs: simplify flake-specific instructions in the documentation

cbarrete:
- install home-manager via `nix run` rather than `nix-build` then running the activation script out of the store manually
- suggest simply using `home-manager switch` when the configuration lives in `~/.config/nixpkgs`

### Description

The flake setup was needlessly complicated and a bit overwhelming for beginners whom home-manager might be their entry point in the world of Nix.

### Checklist

- [ ] Code tested through `nix-shell --pure tests -A run.all`. (I don't use channels so this failed)

### DIFF
--- a/docs/nix-flakes.adoc
+++ b/docs/nix-flakes.adoc
@@ -126,10 +126,7 @@ $ nix flake new ~/.config/nixpkgs -t github:nix-community/home-manager
 2. Install Home Manager and apply the configuration by
 +
 [source,console]
-----
-$ nix build --no-link <flake-uri>#homeConfigurations.jdoe.activationPackage
-$ "$(nix path-info <flake-uri>#homeConfigurations.jdoe.activationPackage)"/activate
-----
+$ nix run <flake-uri>#homeConfigurations.jdoe.activationPackage
 +
 Substitute `<flake-uri>` with the flake URI of the configuration flake.
 If `flake.nix` resides in `~/.config/nixpkgs`,


docs/nix-flakes.adoc
```diff
@@ -126,23 +126,22 @@ $ nix flake new ~/.config/nixpkgs -t github:nix-community/home-manager
 2. Install Home Manager and apply the configuration by
 +
 [source,console]
-----
-$ nix build --no-link <flake-uri>#homeConfigurations.jdoe.activationPackage
-$ "$(nix path-info <flake-uri>#homeConfigurations.jdoe.activationPackage)"/activate
-----
+$ nix run <flake-uri>#homeConfigurations.jdoe.activationPackage
```
ncfavier, line 129: Have you tested this? Wouldn't we need `nix shell ... -c activate`?

docs/nix-flakes.adoc
```diff
@@ -126,23 +126,22 @@ $ nix flake new ~/.config/nixpkgs -t github:nix-community/home-manager
...
 +
 [source,console]
-$ home-manager switch --flake '<flake-uri>#jdoe'
+$ home-manager switch
 +
-once home-manager is installed.
+If your `flake.nix` does not reside in `~/.config/nixpkgs`, you must instead use
++
+[source,console]
+$ home-manager switch --flake '<flake-uri>#jdoe'
```
ncfavier, line 144: See https://github.com/nix-community/home-manager/issues/3640#issuecomment-1416851735: the default path is not great, probably better to keep `--flake` explicitly

cbarrete: I wasn't aware, thanks!
From the look of things, there are no concrete plans on electing a better default, right?

docs/nix-flakes.adoc
cbarrete: I have, though not on a blank machine/VM, but my understanding is that it doesn't matter if I already have an installation.

```
nixpkgs > pwd
/home/user/.config/nixpkgs
nixpkgs > ls
flake.lock  flake.nix  home.nix
nixpkgs > nix run .#homeConfigurations.user.activationPackage
warning: Git tree '/home/user/.config/nixpkgs' is dirty
Activating checkFilesChanged
Activating checkLinkTargets
Activating writeBoundary
Activating installPackages
nix profile remove /nix/store/76nsa0ljc3akw1gz7xiby4ygvj1rg396-home-manager-path
removing 'home-manager-path'
Activating linkGeneration
Cleaning up orphan links from /home/user
No change so reusing latest profile generation 205
Creating home file links in /home/user
Activating onFilesChange
Activating reloadSystemd
```

ncfavier: Right, it runs `home-manager-activation` because that's the derivation's `name`

rycee:
Also see #3709.

cbarrete:
I've removed the bit about `~/.config/nixpkgs` and not specifying the flake. Feel free to merge what remains (about using `nix run` rather than `nix build` + running the script by hand) or to close the PR if https://github.com/nix-community/home-manager/pull/3709 makes it obsolete soon enough.

PR #3740, opened 2023-03-04
lf: sort and update supported options

cbarrete:
Options taken from the output of `lf -doc` as of version `r28`.

### Description

The lf module was missing quite a few recent options, this PR adds support for them.

### Checklist

- [ ] Change is backwards compatible. (I've just removed an option, `color256`, which is not supported anymore by lf)

- [ ] Code tested through `nix-shell --pure tests -A run.all`. (I don't use channels so this doesn't run)

- [ ] Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef). (not required according to what the current tests look like)

    ```
    {component}: {description}

    {long description}
    ```


ncfavier:
This is bad. We should use an RFC 42-style type instead of keeping options in sync with upstream and ignoring unknown options. See https://nix-community.github.io/home-manager/#sec-guidelines-valuable-options

cbarrete:
So this means generating the config from a generic attr set, right?

ncfavier:
Yes

PR #3742, opened 2023-03-05
home-manager: handle missing per-user profiles directory

rycee:
### Description

Specifically, if the global per-user profiles path do not exist and we cannot create it during the activation, then place our profile in the Home Manager data directory. We prefer to use the global location, though, since it makes it visible to `nix-collect-garbage`.

This is intended to improve compatibility with Nix version 2.14 and later, which no longer creates the per-user directories.

Also, use the Home Manager data directory to manage the gcroot for the current generation. It does not have to sit in the global per-user gcroots directory since it should never be eligible for GC.

### Checklist

- [x] Change is backwards compatible. Assuming no reliance on location of gcroot and profile.

    ```
    {component}: {description}

    {long description}
    ```


ncfavier:
Note there are still references to the `per-user` profile dir in `home-manager/home-manager`, not sure if that's intentional.

rycee:
Unfortunately, I didn't have time to finish the changes but figured I'd create this draft PR to show that there is some action on the issue. Unfortunately we'll need to support both profile locations for the time being.

jisantuc:
Thanks for the fast fix here ❤️

Ericson2314:
N.B. you will want to use `${XDG_STATE_HOME-$HOME/.local/state/nix` not `.../home-manager` for `nix-collect-garbage` to work.`

PR #3745, opened 2023-03-06
aerc: integrate signature options

lukasngl:
### Description

Integrates the options defined in `accounts.email.accounts.<name>.signature`.

The options are Integrated as following:

1. If `signature.showSignature` is `none` don't generate signature config.
   (The `attach` variant does not have a corresponding option in aerc,
   so its treated like `append`)
2. If `signature.command` is not `null`, `signature-cmd` is set to a script, that:
   1. Prints the `signature.delimiter`
   2. Executes `signature.command`
3. Else `signature-file` is set to a file, where the content is the concatenation
    of `signature.delimiter` and `signature.text`.

*Note: This is based on #3150, as the upstream module is utterly broken.*

### Checklist

    ```
    {component}: {description}
    {long description}
    ```


rycee:
Thanks for the contribution! Unfortunately the size will make it very difficult to review in a reasonable way as a whole. I would suggest creating separate PRs for each commit, rather than moving this one to ready for review.

The "add space after definitions" commit should either be first (and merged before opening the other PRs) or last, otherwise there will definitely be various conflicts.

mdietrich16:
Hey, I'd be interested in this feature. Any movement since the last message?

lukasngl:
ups, sorry lost track of it. Have now rebased it on master.

lukasngl:
@teto would you mind taking a look at it?

lukasngl:
Hi, don't know why github marked this as having merge conflicts,
but anyways, I have now rebased on master.

I also figured out how to add test for the signature files,
and removed some trailing emtpy lines from the generated scripts.

I am not 100% sure on the format of the signature,
this PR uses `signature.delimiter` followed by `signature.text` or `signature.command`.

From what I have seen, neomutt and alot seem to ignore the `signature.delimiter`
and `signature.command` option and only use the `signature.text`.
Himalaya on the other hand uses the `signature.delimiter` but not the `signature.command`.

@mdietrich16 any opinions on this?

lukasngl:
If the `showSignature` option is set to `attach`, it is treated like `append` and a warning is shown.

This could be annoying, should we instead silently treat is like `attach` (as himalaya does) or only add the signature if the option is `append`?

mdietrich16:
I don't have a strong opinion on this, but I would say we use the delimiter option, as it is more confusing if it doesn't work if explicitly set than if it works but can be unset.

Regarding warning vs. silent, I'd have a slight preference towards silent, as otherwise users who would like to use attach with other clients would have to see the warning on every build.
An option would be to show the warning but add a toggle in the aerc account config to disable the warning to signal explicit notice of the warning, but IMO, that would be overkill

khaneliman:
>Regarding warning vs. silent, I'd have a slight preference towards silent, as otherwise users who would like to use attach with other clients would have to see the warning on every build.
An option would be to show the warning but add a toggle in the aerc account config to disable the warning to signal explicit notice of the warning, but IMO, that would be overkill

We had implemented a warning you can `silence` in firefox that you could take inspiration from. https://github.com/nix-community/home-manager/blob/master/modules/programs/firefox/mkFirefoxModule.nix#L752-L752

lukasngl:
Thanks @khaneliman, for the suggestion, but I also think the warning and an option would be overkill, especially since no one asked for it (yet).

So I removed the warning and just ignore the `attach` option for now until someone needs it and opens an issue.

modules/programs/aerc/accounts.nix
```diff
@@ -288,11 +290,40 @@ in
...
         };
 
+      signatureCfg =
+        account:
+        # TODO: aerc does not support attaching signatures yet.
+        # Until someone needs it, we will just ignore it for now.
+        if account.signature.showSignature == "append" then
+          if account.signature.command != null then
+            {
+              signature-cmd = writeShellScript "aerc-signature.sh" (
```
teto, line 300: it's fine like this but for this sake of the argument would it make sense to pass the delimiter to the command ? does aerc call `signature-cmd` with some arguments already ?

lukasngl: Hi, thanks for the review, but I'm not 100% sure what you mean exactly. According to [aerc-accounts(5)] *signature-cmd* is just the argument to "sh - c" and aerc doesn't have any notion of a *signature-delimiter* itself.

Do you mean, that we should set *signature-cmd* to "aerc-signature.sh ${delimiter}"?

I haven't seen that in the other mail modules. The idea was to introduce a similar behaviour as other mail clients so the config is portable (I.e. insert the delimiter before the signature). Although I don't know how other mail clients handle it as I personally haven't used signatures yet, so feedback is appreciated. 

[aerc-accounts(5)]: https://git.sr.ht/~rjarry/aerc/tree/master/item/doc/aerc-accounts.5.scd#L294

concatime:
I would really like this feature. Is there any deal-breaker left?

lukasngl:
hi, i just rebased on master again and requested reviews.

In the meantime you could test drive it as follows and maybe report back:

<details>
<summary>
As a flake user:
</summary>

```nix
  {
    inputs = {
      # Add this fork as an input
      home-manager-aerc-signature.url = "github:lukasngl/home-manager/aerc-signature";
    };

    outputs = { home-manager, home-manager-aerc-signature, ... }: {
      homeConfigurations."user" = home-manager.lib.homeManagerConfiguration {
        # ...
        modules = [
          {
            # Disable the upstrean aerc module
            disabledModules = [ "programs/aerc" ];
            # ... and replace it with the fork
            imports = [
              "${home-manager-aerc-signature}/modules/programs/aerc"
            ];
          }
          # ... rest of config
        ];
      };
    };
  }

```

</details>

<details>
<summary>
Educated guess for non-flake users:
</summary>

```nix
  { pkgs, ... }:

  let
    aerc-signature = pkgs.fetchFromGitHub {
      repo = "home-manager";
      owner = "lukasngl";
      ref = "aerc-signature";
      sha256 = ""; # Insert correct has here
    };
  in {
    disabledModules = [ "programs/aerc" ];
    imports = [
      "${aerc-signature}/modules/programs/aerc"
    ];

    # ... rest of config
  }

```

</details>

modules/programs/aerc/accounts.nix
```diff
@@ -288,11 +290,40 @@ in
...
+            {
+              signature-file = writeText "aerc-signature.txt" (
+                lib.concatStringsSep "\n" [
+                  account.signature.delimiter
+                  account.signature.text
+                ]
+              );
+            }
+        else
+          { };
```
khaneliman, line 317:
```suggestion
    signatureCfg = account:
      # TODO: aerc does not support attaching signatures yet.
      # Until someone needs it, we will just ignore it for now.
      lib.optionalAttrs (account.signature.showSignature == "append") (
        if account.signature.command != null then {
          signature-cmd = writeShellScript "aerc-signature.sh" ''
            printf '%s\n' "${account.signature.delimiter}"
            ${account.signature.command}
          '';
        } else {
          signature-file = writeText "aerc-signature.txt" ''
            ${account.signature.delimiter}
            ${account.signature.text}
          '';
        }
      );
```

concatime:
I would vote to ignore `account.signature.delimiter` if `signature-file` is used because aerc will [prepend its own delimiter](https://man.archlinux.org/man/aerc-accounts.5.en#:~:text=signature%2Dfile) (`-- `).
> Aerc will add that delimiter if it is not already present.

PR #3753, opened 2023-03-09
zsh: add `package` option to `oh-my-zsh`

nekowinston:
### Description

Adds a `package` option to `programs.zsh.oh-my-zsh`, which allows omz to be overridden with a custom package. My motivation was being able to use `pkgs.unstable.oh-my-zsh` with it, without doing so via another overlay.

This is my first small contribution to the nix ecosystem, and I'm unsure how I would add tests for this, since I couldn't find existing tests for `programs.zsh.oh-my-zsh`.

Maybe the same option should also be added to `programs.zsh.prezto`?

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/zsh.nix
+++ b/modules/programs/zsh.nix
@@ -142,6 +142,8 @@ let
     options = {
       enable = mkEnableOption "oh-my-zsh";
 
+      package = mkPackageOption pkgs "oh-my-zsh" { };
+
       plugins = mkOption {
         default = [];
         example = [ "git" "sudo" ];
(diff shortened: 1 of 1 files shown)


modules/programs/zsh.nix
```diff
@@ -142,6 +142,16 @@ let
...
 
+      package = mkOption {
+        type = types.package;
+        default = pkgs.oh-my-zsh;
+        defaultText = literalExpression "pkgs.oh-my-zsh";
+        example = literalExpression "pkgs.oh-my-zsh";
+        description = ''
+          The oh-my-zsh package to install.
+        '';
+      };
```
ncfavier, line 153:
```suggestion
      package = mkPackageOption pkgs "oh-my-zsh" { };
```

nekowinston: Oh interesting, I took reference from [`zoxide`](https://github.com/nix-community/home-manager/blob/36999b8d19eb6eebb41983ef017d7e0095316af2/modules/programs/zoxide.nix#L15-L22), I didn't know about this helper function. Thanks!

rycee:

PR #3773, opened 2023-03-15
i3status-rust: update it to handle 0.30.x releases

### LINKED ISSUE #3771: bug: programs.i3status-rust will need to comply with v0.30+

### Issue description

PR https://github.com/NixOS/nixpkgs/pull/217546, includes new version with breaking changes that need to be addressed in the module.

These changes are listed in https://github.com/greshake/i3status-rust/blob/master/NEWS.md#i3status-rust-0300 

From my use-case, I can see that `programs.i3status-rust.bars.*.theme` and `.icons` needs adjusts.

### Maintainer CC

@workflow

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.2.6-lqx1, NixOS, 23.05 (Stoat), 23.05.20230315.6bf4086`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.13.3`
 - nixpkgs: `/nix/store/7clfm3qiharfn7z44h1dxbrhnakhmfm0-source`
```

thiagokokada:
### Description

0.30.0 is a major release that brings many breaking changes to the configuration file. See:
https://github.com/greshake/i3status-rust/blob/master/NEWS.md#i3status-rust-0300

The only one that actually affects the module though is the change in how the theme/icons are defined. Other changes are mostly on how to specify formatting/blocks, and since we just generate the TOML as-is, it needs changes in the user side.

So most changes in this commit are documentation updates, having up-to-date examples from things that changed, e.g.: the new `click` attribute that now can be applied to any block.

Fixes #3771.
Related nixpkgs PR: https://github.com/NixOS/nixpkgs/pull/217546

### Checklist

- [x] Change is backwards compatible. *As long someone stays in `stateVersion = 22.05`*.

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/i3status-rust.nix
+++ b/modules/programs/i3status-rust.nix
@@ -21,36 +21,32 @@ in {
           blocks = mkOption {
             type = settingsFormat.type;
             default = [
+              { block = "cpu"; }
               {
                 block = "disk_space";
                 path = "/";
-                alias = "/";
                 info_type = "available";
-                unit = "GB";
-                interval = 60;
+                interval = 20;
                 warning = 20.0;
                 alert = 10.0;
+                format = " $icon root: $available.eng(w:2) ";
               }
               {
                 block = "memory";
-                display_type = "memory";
-                format_mem = "{mem_used_percents}";
-                format_swap = "{swap_used_percents}";
+                format = " $icon $mem_total_used_percents.eng(w:2) ";
+                format_alt = " $icon_swap $swap_used_percents.eng(w:2) ";
               }
               {
-                block = "cpu";
-                interval = 1;
+                block = "sound";
+                click = [{
+                  button = "left";
+                  cmd = "pavucontrol";
+                }];
               }
-              {
-                block = "load";
-                interval = 1;
-                format = "{1m}";
-              }
-              { block = "sound"; }
               {
                 block = "time";
-                interval = 60;
-                format = "%a %d/%m %R";
+                interval = 5;
+                format = " $timestamp.datetime(f:'%a %d/%m %R') ";
               }
             ];
             description = ''
@@ -64,20 +60,23 @@ in {
                 {
                   block = "disk_space";
                   path = "/";
-                  alias = "/";
                   info_type = "available";
-                  unit = "GB";
                   interval = 60;
                   warning = 20.0;
                   alert = 10.0;
                 }
                 {
                   block = "sound";
-                  format = "{output_name} {volume}%";
-                  on_click = "pavucontrol --tab=3";
+                  format = " $icon $output_name {$volume.eng(w:2) |}";
+                  click = [
+                    {
+                      button = "left";
+                      cmd = "pavucontrol --tab=3";
+                    }
+                  ];
                   mappings = {
-                   "alsa_output.pci-0000_00_1f.3.analog-stereo" = "";
-                   "bluez_sink.70_26_05_DA_27_A4.a2dp_sink" = "";
+                    "alsa_output.pci-0000_00_1f.3.analog-stereo" = "";
+                    "bluez_sink.70_26_05_DA_27_A4.a2dp_sink" = "";
                   };
                 }
               ];
@@ -94,7 +93,7 @@ in {
             example = literalExpression ''
               {
                 theme =  {
-                  name = "solarized-dark";
+                  theme = "solarized-dark";
                   overrides = {
                     idle_bg = "#123456";
                     idle_fg = "#abcdef";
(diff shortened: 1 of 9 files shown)


modules/programs/i3status-rust.nix
```diff
@@ -258,8 +251,8 @@ in {
         '';
 
         source = settingsFormat.generate ("config-${cfgFileSuffix}.toml") ({
-          theme = cfg.theme;
-          icons = cfg.icons;
+          theme = { theme = cfg.theme; };
+          icons = { icons = cfg.icons; };
```
thiagokokada, line 255: We could make this change backwards compatible if we check the version of `i3status-rust` here.

thiagokokada: Maybe check for the `home.stateVersion` here?

thiagokokada: Done in fb3dff74d14c9536317d83854545df6556244a44.

thiagokokada:

modules/programs/i3status-rust.nix
```diff
@@ -258,8 +251,14 @@ in {
         '';
 
         source = settingsFormat.generate ("config-${cfgFileSuffix}.toml") ({
-          theme = cfg.theme;
-          icons = cfg.icons;
+          theme = if lib.versionAtLeast config.home.stateVersion "23.05" then {
```
ncfavier, line 254: Is there a reason not to condition this on `cfg.package.version` instead?

ncfavier:
@PedroHLC could you test this?

modules/programs/i3status-rust.nix
thiagokokada: No really a strong reason, but I did this way:

- Because it was easier.
- This is just a small part of the migration process anyway. Most users will need to update their configuration, so I thought having the state version dictating this would make users more aware of the changes.

But if you think it is better, I can change the check to use the version instead.

modules/programs/i3status-rust.nix
```diff
@@ -21,36 +21,32 @@ in {
...
-                interval = 1;
-                format = "{1m}";
-              }
-              { block = "sound"; }
               {
                 block = "time";
-                interval = 60;
-                format = "%a %d/%m %R";
+                interval = 5;
+                format = " $timestamp.datetime(f:'%a %d/%m %R') ";
```
AndersonTorres, line 49: Are the spaces needed?

modules/programs/i3status-rust.nix
AndersonTorres: +1 for version.

tests/modules/programs/i3status-rust/with-old-state-version.nix
```diff
@@ -0,0 +1,53 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
```
AndersonTorres, line 3:
```suggestion
```

Not being used.

AndersonTorres (changes requested):
Besides, LGTM.

modules/programs/i3status-rust.nix
ncfavier: IMO this is not what stateVersion is for: users are not supposed to have to change it, ever (unless they're reinstalling their whole system). Configuration isn't state, we actually know statically which format to generate, so I would prefer using `lib.versionAtLeast cfg.package.version "0.30"`.

modules/programs/i3status-rust.nix
thiagokokada: Yes.

thiagokokada: You can check the release notes: https://github.com/greshake/i3status-rust/blob/master/NEWS.md#i3status-rust-0300

> Icons are now part of the format string option as a placeholder in blocks where format is customisable. If you have modified format and would like to keep the same behaviour (icon, whitespace) you need to update the value. For example,

```
[[block]]
block = "cpu"
format = "{utilization}"
```

needs to be changed to:

```
[[block]]
block = "cpu"
format = " $icon $utilization "
```

modules/programs/i3status-rust.nix
thiagokokada: Done.

tests/modules/programs/i3status-rust/with-old-state-version.nix
thiagokokada: Removed it from all tests since they're not being used too.

PedroHLC:
> @PedroHLC could you test this?

Tested 4f349932b7aee93e9b97a35955d3917f94196640 rebased over current master, and it worked, generated the expected config file.

thiagokokada:
I just added myself as maintainer of the module. Should be ready to merge now.

mightyiam:
Thank you.

PR #3776, opened 2023-03-17
starship: condition nushell integration on nushell 0.73+

ncfavier:
Older versions are not supported, see https://github.com/nix-community/home-manager/pull/3697#issuecomment-1473739415, https://starship.rs/#nushell

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/starship.nix
+++ b/modules/programs/starship.nix
@@ -80,6 +80,14 @@ in {
   };
 
   config = mkIf cfg.enable {
+    assertions = [{
+      assertion = cfg.enableNushellIntegration
+        -> versionAtLeast config.programs.nushell.package.version "0.73";
+      message = ''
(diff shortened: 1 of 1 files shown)


houstdav000:
So, bit of a problem with this: I get the error even if I don't have NuShell control enabled. Could we make the integration enable option default to if NuShell is managed by home-manager (e.g. `default = config.nushell.enable;`)? 

Might also be a good idea for the other supported shells as well, but that can be a separate thing.

ncfavier:
Oh, I totally missed that `enableNushellIntegration` is enabled by default... See https://github.com/nix-community/home-manager/pull/3777 for more discussion.

ncfavier:
I'll just revert for now.

PR #3777, opened 2023-03-17
[backport 22.11] starship: condition nushell integration on nushell 0.73+

ncfavier:
Backporting https://github.com/nix-community/home-manager/pull/3776

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/starship.nix
+++ b/modules/programs/starship.nix
@@ -80,6 +80,14 @@ in {
   };
 
   config = mkIf cfg.enable {
+    assertions = [{
+      assertion = cfg.enableNushellIntegration
+        -> versionAtLeast config.programs.nushell.package.version "0.73";
+      message = ''
+        Nushell integration for starship requires nushell 0.73 or later.
+      '';
+    }];
+
     home.packages = [ cfg.package ];
 
     xdg.configFile."starship.toml" = mkIf (cfg.settings != { }) {


ncfavier:
I don't know anything about nushell but that doesn't look like it's coming from home-manager. cc @Philipp-M

rycee:
Looks like nushell 0.71 is the default in Nixpkgs 22.11? Doesn't this nushell integration therefore make Home Manager version 22.11 incompatible with Nixpkgs 22.11? If so, that's a bit embarrassing. Would be better if the integration could be made to work with nushell 0.71…

ncfavier:
Yeah, I'm not sure why it was backported. It looks like this may not even be enough: https://github.com/starship/starship/pull/4477 is needed to avoid the above error with nushell 0.70+, but that PR is only in starship 1.12 while nixos-22.11 is on 1.11...

@Lord-Valen did you get this to work at all on 22.11? What versions of starship and nushell are you using?

Philipp-M:
Yeah I agree with @rycee, maybe at least check the version of nushell and enable the integration for it based on the version of nushell instead of an assertion? And maybe add a warning that tells the user that starship doesn't work with nushell versions < 0.73 (if `programs.nushell.enable = true`) ?

ncfavier:
I've reverted the PR on both branches. Not really in the right mindset to think about this, sorry.

PR #3787, opened 2023-03-19
borgmatic: Optionally exclude home-manager symlinks from backup

DamienCassou:
-------

### Description

Make it easy to generate a borgmatic's exclusion list containing all
symlinks generated by home manager.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/borgmatic.nix
+++ b/modules/programs/borgmatic.nix
@@ -43,7 +43,10 @@ let
     };
   };
 
-  configModule = types.submodule {
+  configModule = types.submodule ({ config, ... }: {
+    config.location.extraConfig.exclude_from =
+      mkIf config.location.excludeHomeManagerSymlinks
+      (mkAfter [ (toString hmExcludeFile) ]);
     options = {
       location = {
         sourceDirectories = mkOption {
@@ -57,6 +60,18 @@ let
           description = "Paths to repositories.";
           example =
             literalExpression ''["ssh://myuser@myrepo.myserver.com/./repo"]'';
+        };
+
+        excludeHomeManagerSymlinks = mkOption {
+          type = types.bool;
+          description = ''
+            Whether to exclude Home Manager generated symbolic links from
+            the backups. This facilitates restoring the whole home
+            directory when the Nix store doesn't contain the latest
+            Home Manager generation.
+          '';
+          default = false;
+          example = true;
         };
 
         extraConfig = extraConfigOption;
@@ -120,9 +135,17 @@ let
         extraConfig = extraConfigOption;
       };
     };
-  };
+  });
 
   removeNullValues = attrSet: filterAttrs (key: value: value != null) attrSet;
+
+  hmFiles = builtins.attrValues config.home.file;
+  hmSymlinks = (lib.filter (file: !file.recursive) hmFiles);
+  hmExcludePattern = file: ''
+    ${config.home.homeDirectory}/${file.target}
+  '';
+  hmExcludePatterns = lib.concatMapStrings hmExcludePattern hmSymlinks;
+  hmExcludeFile = pkgs.writeText "hm-symlinks.txt" hmExcludePatterns;
 
   writeConfig = config:
     generators.toYAML { } {
--- a/modules/programs/borgmatic.nix
+++ b/modules/programs/borgmatic.nix
@@ -5,6 +5,8 @@ with lib;
 let
   cfg = config.programs.borgmatic;
 
+  yamlFormat = pkgs.formats.yaml { };
+
   mkNullableOption = args:
     lib.mkOption (args // {
       type = lib.types.nullOr args.type;
@@ -20,7 +22,7 @@ let
     };
 
   extraConfigOption = mkOption {
-    type = with types; attrsOf (oneOf [ str bool path int (listOf str) ]);
+    type = yamlFormat.type;
     default = { };
     description = "Extra settings.";
   };
--- a/tests/modules/programs/borgmatic/exclude-hm-symlinks-nothing-else.nix
+++ b/tests/modules/programs/borgmatic/exclude-hm-symlinks-nothing-else.nix
@@ -1,0 +1,40 @@
+{ config, pkgs, ... }:
+
+let
+  backups = config.programs.borgmatic.backups;
+  excludeFile = pkgs.writeText "excludeFile.txt" "/foo/bar";
+in {
+  config = {
+    programs.borgmatic = {
+      enable = true;
+      backups = {
+        main = {
+          location = {
+            sourceDirectories = [ "/my-stuff-to-backup" ];
+            repositories = [ "/mnt/disk1" ];
+            excludeHomeManagerSymlinks = true;
+          };
+        };
+      };
+    };
+
+    test.stubs.borgmatic = { };
+
+    nmt.script = ''
+      config_file=$TESTED/home-files/.config/borgmatic.d/main.yaml
+      assertFileExists $config_file
+
+      yq=${pkgs.yq-go}/bin/yq
+
+      hmExclusionsFile=$($yq '.location.exclude_from[0]' $config_file)
+      expected_content='/home/hm-user/.config/borgmatic.d/main.yaml'
+
+      grep --quiet "$expected_content" "$hmExclusionsFile"
+
+      if [[ $? -ne 0 ]]; then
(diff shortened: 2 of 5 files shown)


modules/programs/borgmatic.nix
```diff
@@ -57,6 +57,18 @@ let
             literalExpression ''["ssh://myuser@myrepo.myserver.com/./repo"]'';
         };
 
+        excludeHomeManagerSymlinks = mkOption {
+          type = types.bool;
+          description = ''
+            Whether to exclude home-manager generated symlinks from
+            the backups. This facilitates restoring the whole home
+            directory when the Nix store doesn't contain the latest
+            home-manager generation.
```
rycee, line 66:
```suggestion
            Whether to exclude Home Manager generated symbolic links from
            the backups. This facilitates restoring the whole home
            directory when the Nix store doesn't contain the latest
            Home Manager generation.
```

tests/modules/programs/borgmatic/exclude-hm-symlinks-nothing-else.nix
```diff
@@ -0,0 +1,43 @@
...
+
+    test.stubs.borgmatic = { };
+
+    nmt.script = ''
+      config_file=$TESTED/home-files/.config/borgmatic.d/main.yaml
+      assertFileExists $config_file
+
+      declare -A expectations
+
+      expectations[location.exclude_from[0]]="/nix/store/18sk4zmxs1h73nbc0r2ar9z60c65mwfm-borgmatic-home-manager-exclude-symlinks.txt"
```
rycee, line 29: Will need to read this path dynamically to avoid issues when the hash changes.

modules/programs/borgmatic.nix
```diff
@@ -122,12 +134,26 @@ let
 
   removeNullValues = attrSet: filterAttrs (key: value: value != null) attrSet;
 
+  hmFiles = lib.mapAttrsToList (_: file: file) config.home.file;
+  hmSymlinks = (lib.filter (file: !file.recursive) hmFiles);
+  hmExcludePatterns = map (file: ''
+    ${config.home.homeDirectory}/${file.target}
+  '') hmSymlinks;
+  hmExcludeFile = pkgs.writeText "borgmatic-home-manager-exclude-symlinks.txt"
+    (lib.concatStrings hmExcludePatterns);
```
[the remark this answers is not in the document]
ncfavier: This will not append a final newline to the file, which might be fine in this instance but will almost certainly bite us one day when we don't expect it. Lines should be newline-terminated, not -separated.

```nix
hmExcludePatterns = lib.concatMapStrings (file: "${config.home.homeDirectory}/${file.target}\n") hmSymlinks;
```

modules/programs/borgmatic.nix
```diff
@@ -122,12 +134,24 @@ let
...
+  hmFiles = lib.mapAttrsToList (_: file: file) config.home.file;
+  hmSymlinks = (lib.filter (file: !file.recursive) hmFiles);
+  hmExcludePattern = file: "${config.home.homeDirectory}/${file.target}";
+  hmExcludePatterns = lib.concatMapStringsSep "\n" hmExcludePattern hmSymlinks;
+  hmExcludeFile = pkgs.writeText "hm-symlinks.txt" hmExcludePatterns;
+  hmSymlinksExclusions = config:
+    lib.optionalAttrs config.location.excludeHomeManagerSymlinks {
+      exclude_from =
+        (lib.optionals (config.location.extraConfig ? "exclude_from")
+          config.location.extraConfig.exclude_from) ++ [ hmExcludeFile ];
```
ncfavier, line 146: The ad hoc merging here could be avoided by moving this whole thing into the `configModule` submodule.

tests/modules/programs/borgmatic/exclude-hm-symlinks-nothing-else.nix
DamienCassou: I tried a different solution. Is it better?

modules/programs/borgmatic.nix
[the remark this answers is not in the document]
DamienCassou: I agree with @ncfavier: I changed my code to look like @rycee's structure but making sure all lines are terminated with a newline character.

modules/programs/borgmatic.nix
```diff
@@ -122,12 +134,26 @@ let
 
   removeNullValues = attrSet: filterAttrs (key: value: value != null) attrSet;
 
+  hmFiles = lib.mapAttrsToList (_: file: file) config.home.file;
```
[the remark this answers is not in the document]
DamienCassou: Done

modules/programs/borgmatic.nix
DamienCassou: I [tried to do that](https://gist.github.com/DamienCassou/8b35027f35ad8f81b9657cf78d55c859) but it fails because `location.extraConfig.exclude_from` can be used in user configuration files as well:

```
error: The option `programs.borgmatic.backups.main.location.extraConfig.exclude_from' has conflicting definition values:
       - In `/home/cassou/Documents/projects/nix-system/home-manager/modules/programs/borgmatic.nix':
           [
             <derivation /nix/store/h010kdss4pd2lb0j2zl4fmnl57k4p2yg-hm-symlinks.txt.drv>
           ]
       - In `/home/cassou/Documents/projects/nix-system/home-manager/tests/modules/programs/borgmatic/exclude-hm-symlinks.nix':
           [
             <derivation /nix/store/a7az4qq2b9w5cknk6jnhacxm9nizwzyx-excludeFile.txt.drv>
           ]
```

DamienCassou: I feel the solution could be to make `exclude_from` a declared option instead of an extra one. 

What do you think?

ncfavier: I think that error is because the type of `extraConfig` is `attrsOf (oneOf [ str bool path int (listOf str) ])`, but you're trying to merge two lists of *derivations*, which aren't strings, so it doesn't know how to merge them. `listOf (either str path)` should fix it.

ncfavier: Well, since this is YAML we might as well just use the YAML type from [`pkgs.formats`](https://nixos.org/manual/nixos/unstable/#sec-settings-nix-representable).

DamienCassou: I [tried your suggestion](https://gist.github.com/DamienCassou/8b35027f35ad8f81b9657cf78d55c859) but it still shows the same error message. Can we merge this PR and try that in a separate one?

DamienCassou:
@rycee @ncfavier I think I'm done applying your suggestions (except one from @ncfavier that I can't make work). Can you please have another look?

modules/programs/borgmatic.nix
ncfavier: Applying that patch on top of the current revision I get no eval error, only a test error that is fixed by swapping 0 and 1 in the index to `exclude_from`. We might want to make the order more deterministic using `mkBefore` or `mkAfter`.

DamienCassou: Ok, I applied the patch and the CI is happy so I am happy.

For some reason, the test still fails locally for me when using `nix-shell`:

```
$ nix-shell --pure tests -A run.borgmatic-program-exclude-hm-symlinks
error: The option `programs.borgmatic.backups.main.location.extraConfig.exclude_from' has conflicting definition values:
       - In `/home/cassou/Documents/projects/nix-system/home-manager/tests/modules/programs/borgmatic/exclude-hm-symlinks.nix':
           [
             "/nix/store/7vvpsb5d589890x125agz60d5i7jsxav-excludeFile.txt"
           ]
       - In `/home/cassou/Documents/projects/nix-system/home-manager/modules/programs/borgmatic.nix':
           [
             "/nix/store/ix4zmh6b02az1lwm98q1wzsn2rbarlsz-hm-symlinks.txt"
           ]
```

The problem also appears when doing `home-manage switch`.

But everything is fine even locally when using `nix develop`:

```
$ nix develop --ignore-environment .#tests.borgmatic-program-exclude-hm-symlinks
borgmatic-program-exclude-hm-symlinks: OK
```

Could there be a difference between nixos-unstable and release-22.11?

I'm happy merging the PR like it is though.

DamienCassou:
@ncfavier I applied your final suggestion.

rycee:

PR #3791, opened 2023-03-19
services.batsignal: init

Kranzes:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -376,6 +376,8 @@ Makefile                                              @thiagokokada
 /modules/services/barrier.nix                         @Kritnich
 /tests/modules/services/barrier                       @Kritnich
 
+/modules/services/batsignal.nix                       @Kranzes
+
 /modules/services/betterlockscreen.nix                @SebTM
 
 /modules/programs/borgmatic.nix                       @DamienCassou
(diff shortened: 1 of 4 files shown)


ncfavier:
Dupe of https://github.com/nix-community/home-manager/pull/3243

Kranzes:
> Dupe of #3243

I like my implementation much more personally, imo https://github.com/nix-community/home-manager/pull/3243 is too intrusive, and tries to handle too much.

loicreynier:
I'm fine with both implementations. I personally prefer having the specific options available in the HM module but it's true that it would be better to let the program handle its defaults. I could set to `null` the defaults in my implementation to let the program handle the defaults. Concerning the maintainability, since the program is pretty mature, I don't think it will really be an issue.

Kranzes:
If you're fine with both then I definitely want mine merge instead ;)

ncfavier:
I'm inclined towards this implementation, though ideally for a program whose main mode of configuration is command-line arguments we'd want a middle ground between a "stringly-typed" `extraArgs` and hardcoding every flag as an option. Something like an [RFC 42](https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md)-style `pkgs.formats.cli` powered by [`toGNUCommandLine`](https://github.com/NixOS/nixpkgs/blob/master/lib/cli.nix), but this isn't a thing yet :shrug:

rycee:
Thanks! Merged to master now.

Issue #3760, opened 2023-03-12
Borgmatic Hooks?


mentalblock:
cc: @DamienCassou 

It seems like the [current nix abstraction of borgmatic](https://github.com/nix-community/home-manager/blob/e412025fffdcd6219ddd21c65d9a1b90005ce508/modules/programs/borgmatic.nix) does not allow specifying hooks as outlined in the [official borgmatic configuration](https://torsion.org/borgmatic/docs/reference/configuration/). Can someone verify this is the case? I am questioning whether I understand the current home-manager options correctly.

I do not believe I have the skillset to add the appropriate options to the current module, but I will give it a shot if someone can confirm that the current state of the module does not support hooks.

I look forward to hearing from other users regarding this use case. Thanks in advance.

ncfavier:
That is correct. We should add a YAML `freeformType` to the `configModule` so that any options can be configured before the module knows about them, following [RFC 42](https://github.com/Infinisil/rfcs/blob/config-option/rfcs/0042-config-option.md).

mentalblock:
It looks like we can basically use [the NixOS module](https://github.com/NixOS/nixpkgs/blob/919f2b2b6292436570613e7f7166eb992b12255f/nixos/modules/services/backup/borgmatic.nix) as a guideline as it is more flexible and uses the pattern you describe.

As a general question, what is the proper direction when a NixOS module and a home-manager module mostly overlap? Is the home-manager module still maintained for those who run home-manager but not NixOS?

DamienCassou:
I think https://github.com/nix-community/home-manager/pull/3793 will fix this issue.

DamienCassou:
I implemented a different fix for this issue in https://github.com/nix-community/home-manager/pull/4108.

mentalblock:
Thank you so much @DamienCassou , works perfectly.

PR #3793, opened 2023-03-20
borgmatic: Allow top-level extra options

DamienCassou:
### Description

Fix https://github.com/nix-community/home-manager/issues/3760

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/borgmatic.nix
+++ b/modules/programs/borgmatic.nix
@@ -44,7 +44,7 @@ let
   };
 
   configModule = types.submodule ({ config, ... }: {
-    config.location.extraConfig.exclude_from =
+    config.extraConfig.location.exclude_from =
       mkIf config.location.excludeHomeManagerSymlinks
       (mkAfter [ (toString hmExcludeFile) ]);
     options = {
@@ -73,8 +73,6 @@ let
           default = false;
           example = true;
         };
-
-        extraConfig = extraConfigOption;
       };
 
       storage = {
@@ -84,7 +82,6 @@ let
           example =
             literalExpression ''"''${pkgs.password-store}/bin/pass borg-repo"'';
         };
-        extraConfig = extraConfigOption;
       };
 
       retention = {
@@ -101,8 +98,6 @@ let
         keepWeekly = mkRetentionOption "weekly";
         keepMonthly = mkRetentionOption "monthly";
         keepYearly = mkRetentionOption "yearly";
-
-        extraConfig = extraConfigOption;
       };
 
       consistency = {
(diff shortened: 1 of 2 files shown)


modules/programs/borgmatic.nix
```diff
@@ -168,7 +170,7 @@ let
       } // config.retention.extraConfig;
       consistency = removeNullValues { checks = config.consistency.checks; }
         // config.consistency.extraConfig;
-    };
+    } // config.extraConfig);
```
rycee, line 173: It may be a bit unfortunate that this doesn't do a recursive update so `consistency.extraConfig = …` would give a different result compared to `extraConfig.consistency = …`.

DamienCassou: I think it is a very good point. Would you mind giving me a clue on how to implement that please?

DamienCassou: after discussing on Matrix, we decided to:

1. get rid of all the `extraConfig` options and only keep the top-level one
2. use `mkRenamedOptionModule` to warn the user about the deprecated API

Unfortunately, I've been unable to make it work. I pushed a WIP commit to show my latest try. Can someone please help?

DamienCassou: The latest issue I have is

```
error: You're trying to declare a value of type `list'
       rather than an attribute-set for the option
       `programs.borgmatic.backups.<name>.imports'!

       This usually happens if `programs.borgmatic.backups.<name>.imports' has option
       definitions inside that are not matched. Please check how to properly define
       this option by e.g. referring to `man 5 configuration.nix'!
```

ncfavier: The immediate problem is that you're defining `imports` inside `options`, but it should be at the same level as `config` and `options`.

The real problem is that these `mkRemovedOptionModule`s will not work because of https://github.com/NixOS/nixpkgs/issues/96006. The most promising attempt to fix that is currently https://github.com/NixOS/nixpkgs/pull/207187.

DamienCassou: If I understand what @ncfavier just said, it seems impossible to deprecate the non-top-level `extraConfig` options for now. Having both a top-level `extraConfig` option and having non-top-level ones is messy. 

The only solution I see is to have no top-level `extraConfig` option. As a result, I suggest we introduce a top-level option for each of borgmatic's setting sections and add one `extraConfig` for each. If I'm not mistaken after reading the [example configuration file](https://torsion.org/borgmatic/docs/reference/config.yaml), this just means adding `output` and `hooks` top-level options.

What do you think @ncfavier and @rycee?

DamienCassou:
Can someone please answer my question?

modules/programs/borgmatic.nix
DamienCassou: I implemented this approach in https://github.com/nix-community/home-manager/pull/4108.

DamienCassou:
Closing this PR in favor of https://github.com/nix-community/home-manager/pull/4108.

PR #3796, opened 2023-03-20
nushell: adds login.nu configuration option

SaiProton:
### Description
Nushell has the option to source from the login.nu file upon logging in, in the case that nushell is used as a login shell.
See https://www.nushell.sh/book/configuration.html#configuring-nu-as-a-login-shell for more information.

This change adds the login file alongside the existing config and env files as another configuration option.

### Checklist

       (tested with `nix develop --ignore-environment .#tests.nushell-example-settings`)

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/nushell.nix
+++ b/modules/programs/nushell.nix
@@ -96,6 +96,23 @@ in {
       '';
     };
 
+    loginFile = mkOption {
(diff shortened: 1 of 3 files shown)


rycee:

PR #3797, opened 2023-03-21
zathura: add documentation for mode-specific mappings

GaetanLepage:
### Description

As [this comment](https://github.com/nix-community/home-manager/pull/3134#issuecomment-1475944069) points out, the current zathura module does not support mode-specific mappings.
This PR allows the user to provide for a mapping an attrs that contains both the action and the mode.
Of course, the user can still provide a simple string which ensures backward compatibility.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/zathura.nix
+++ b/modules/programs/zathura.nix
@@ -58,10 +58,14 @@ in {
           <manvolnum>5</manvolnum>
         </citerefentry>
         for the full list of possible mappings.
+
+        You can create a mode-specific mapping by specifying the mode before the key:
+        <literal>"[normal] &lt;C-b&gt;" = "scroll left";</literal>
       '';
       example = {
         D = "toggle_page_mode";
         "<Right>" = "navigate next";
+        "[fullscreen] <C-i>" = "zoom in";
       };
     };
 


GaetanLepage:
@rycee  @ncfavier @ne9z

ncfavier:
Doesn't this prevent from binding the same key in different modes?

Also, sadly the descriptions won't appear in the manual until https://github.com/NixOS/nixpkgs/issues/207962 is solved, and I've lost all motivation to work on that issue for now.

ne9z:
No need to rush this.  I'm content with the solution from yesterday.

https://github.com/ne9z/dotfiles-flake/blob/cc206b1a40e2774826b33390a07c8fcc9efaf976/modules/per-user/yc/modules/home-manager/default.nix#L115

GaetanLepage:
@ncfavier you are right...
Maybe we should just keep it like this and update the documentation.

modules/programs/zathura.nix
```diff
@@ -58,10 +58,14 @@ in {
           <manvolnum>5</manvolnum>
         </citerefentry>
         for the full list of possible mappings.
+
+        You can create a mode-specific mapping by specifying the mode before the key:
+        <literal>"[normal] <C-b>" = "scroll left";</literal>
```
ncfavier, line 63:
```suggestion
        <literal>"[normal] &lt;C-b&gt;" = "scroll left";</literal>
```

GaetanLepage: Thank you, done !

PR #3800, opened 2023-03-22
home-manager: set the correct NIX_PATH when running from flakes

midchildan:
### Description

`nix run home-manager` failed to work properly because it attempts to find the Home Manager source from `NIX_PATH`. The problem is, Home Manager wouldn't be in `NIX_PATH` in most cases if flakes are being used.

Before:

```console
❯ nix run home-manager -- -f home.nix news | head
error: file 'home-manager/home-manager/home-manager.nix' was not found in the Nix search path (add it using $NIX_PATH or -I)
/nix/store/pxmr8qxkhlc5r9ny9iz4i14saih39l4p-home-manager/bin/home-manager: line 712: /tmp/home-manager-build.fS7mCjQ1Tw/news-info.sh: No such file or directory
```

After:

```console
❯ nix run '.#' -- -f home.nix news | head
* 2023-01-31 22:08:41 [read]

  A new module is available: 'programs.rbenv'.



* 2023-01-30 10:39:11 [read]

  A new module is available: 'programs.wlogout'.

```

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -89,7 +89,7 @@
         pkgs = nixpkgs.legacyPackages.${system};
         docs = import ./docs { inherit pkgs; };
         tests = import ./tests { inherit pkgs; };
-        hmPkg = pkgs.callPackage ./home-manager { };
+        hmPkg = pkgs.callPackage ./home-manager { path = ./.; };
       in {
         devShells.tests = tests.run;
         packages = {


rycee:
Had a quick look at this and I think the old behavior is intentional. With flakes we should rely on a different mechanism for this evaluation. Basically we cannot rely on there being a `home.nix` file and need to evaluate the flake output instead.

I created #3807 to show a more useful error when running `home-manager news` in flake setup. Will also have a look at getting proper support

midchildan:
With this change, I also hoped to allow `nix run home-manager -- build` to build non-flake Home Manager configurations. The main use case for this is quick and dirty testing of configurations. Using a non-flake configuration would allow me to skip the process of creating a new flake each time when I want to test something out. Since I use flakes for my actual configuration, I don't have home-manager present in my Nix channels. This lead me to create this PR.

On the other hand, I see that this isn't exactly a popular use case though. Do you think a feature like this should be supported in Home Manager?

rycee:
I've now merged a reworking of the news functionality in `home-manager`. The `path` parameter is also set so hopefully your use case is now covered. Please give it a try and see if it works for you.

midchildan:
Thanks! It worked for me.

Issue #3555, opened 2023-01-02
bug: order of `plugins` and `extraConfig` sections in generated `tmux.conf`


alex35mil:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

I feel that it's not a bug, and I misuse something, b/c nobody ever reported it, but the case should be common. Anyway.

In generated `tmux.conf`, the `plugins` section goes before `extraConfig`. If I configure a status bar in the `extraConfig` and want to use values from the plugins, it doesn't work since the status bar must be configured before `run-shell` calls.

Should these sections be swapped in the output?

### Maintainer CC

@rycee (mentioning b/c of git blame. couldn't find the maintainers list)

### System information

```markdown
- system: `"aarch64-darwin"`
 - host os: `Darwin 21.6.0, macOS 12.6`
 - multi-user?: `yes`
 - sandbox: `no`
 - version: `nix-env (Nix) 2.11.0`
 - channels(Alex): `"home-manager"`
 - channels(root): `"nixpkgs"`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixpkgs`
```

Pegasust:
What's the status on this issue? Is there anyone working on a fix?

I think adding a `configBeforePlugins` field makes sense as an escape hatch for me.

sumnerevans:
I think that a new config option sounds like a good idea. We need to try and preserve backwards compatibility, and moving the `extraConfig` location would cause a breaking change.

D3vil0p3r:
Any news?

podocarp:
Can I propose to have a extraConfigPre and extraConfigPost section?
Either that, or specify in the docs that this is extraConfigPre.

mahyarmirrashed:
Any news on this?

podocarp:
> This is still needed. Some tmux plugins need to set some configurations before the plugin is loaded (i.e.: tmux-catppuccin)

For catppuccin I found a fix (I'm using their flakes but I assume the same would work with the plugin), it seems like setting plugin specific extraConfigs is the way https://nix.catppuccin.com/search/v1.2/?query=tmux&option_scope=1&option=catppuccin.tmux.extraConfig

PR #3801, opened 2023-03-24
draft: tmux: add configBeforePlugin option

Pegasust:
### Description

- Adds `programs.tmux.configBeforePlugin` option. 
- This complements `programs.tmux.extraConfig`, but should allow `tmux` config for `status-right` and `status-left` to take in variables exposed by plugins
- Requested by https://github.com/nix-community/home-manager/issues/3555

### Checklist

  - Unrelated tests failing: https://github.com/nix-community/home-manager/issues/3803

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/tmux.nix
+++ b/modules/programs/tmux.nix
@@ -100,7 +100,7 @@ let
     set  -s escape-time       ${toString cfg.escapeTime}
     set  -g history-limit     ${toString cfg.historyLimit}
 
-    ${optionalString ((builtins.stringLength cfg.configBeforePlugin) > 0) ''
+    ${optionalString (cfg.configBeforePlugin != "") ''
     # ============================================= #
     # Extra configs before plugins                  #
     # --------------------------------------------- #
--- a/modules/programs/tmux.nix
+++ b/modules/programs/tmux.nix
@@ -159,7 +159,7 @@ in {
           Additional configuration in <filename>tmux.conf</filename>
           before plugins run-shell session. This is useful to configure
           status-right or status-left that uses variables exposed by plugins.
-        ''
+        '';
       };
 
       clock24 = mkOption {
--- a/modules/programs/tmux.nix
+++ b/modules/programs/tmux.nix
@@ -100,11 +100,11 @@ let
     set  -s escape-time       ${toString cfg.escapeTime}
     set  -g history-limit     ${toString cfg.historyLimit}
 
-    ${optionalString (cfg.configBeforePlugin != "") ''
+    ${optionalString (cfg.extraConfigBeforePlugin != "") ''
     # ============================================= #
     # Extra configs before plugins                  #
     # --------------------------------------------- #
-    ${cfg.configBeforePlugin}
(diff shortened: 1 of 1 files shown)


modules/programs/tmux.nix
```diff
@@ -99,6 +99,13 @@ let
     setw -g clock-mode-style  ${if cfg.clock24 then "24" else "12"}
     set  -s escape-time       ${toString cfg.escapeTime}
     set  -g history-limit     ${toString cfg.historyLimit}
+
+    ${optionalString ((builtins.stringLength cfg.configBeforePlugin) > 0) ''
```
rycee, line 103:
```suggestion
    ${optionalString (cfg.configBeforePlugin != "") ''
```

modules/programs/tmux.nix
```diff
@@ -147,6 +154,16 @@ in {
         description = "Base index for windows and panes.";
       };
 
+      configBeforePlugin = mkOption {
```
rycee, line 157:
```suggestion
      extraConfigBeforePlugin = mkOption { 
```

rycee:
Thanks for the contribution!

An alternative is to do something like

```
xdg.configFile."tmux/tmux.conf".text = lib.mkOrder 600 "…";
```

in your configuration.

Pegasust:
> ```
> xdg.configFile."tmux/tmux.conf".text = lib.mkOrder 600 "…";
> ```
I'm not too familiar with `lib.mkBefore` and `lib.mkAfter`. I know `lib.mkOrder` is something between providing the actual value and `lib.mkDefault`.

Does `lib.mkOrder` override any of the fields or does it correctly place the section between the end of `lib.mkBefore tmuxConfig` and the beginning of `lib.mkAfter cfg.extraConfig`?

modules/programs/tmux.nix
Pegasust: Done, along with changing all the respective cfg counterparts.

ncfavier:
There are two different things here: `mkDefault` and `mkForce` are defined in terms of `mkOverride` and determine the *override priority* of definitions. `mkBefore` and `mkAfter` are defined in terms of `mkOrder` and determine the *order priority* of definitions.

Only the definitions with the highest override priority are used, and, if there are several, they are ordered according to their order priority.

Here all the definitions have the same override priority, so let's focus on order priorities. The definitions in the module are as follows:

- `mkBefore tmuxConf` (priority = 500)
- `pluginsConf` (default priority = 1000)
- `mkAfter cfg.extraConfig` (priority = 1500)

So, if you add a definition with `mkOrder 600` (or any number between 500 and 1000), it will get ordered between the general tmux config and the plugins config.

Pegasust:
Makes so much sense now! I was confused between order priority vs override priority. Thanks for the clarification!

Pegasust:
Just to clean up home-manager's PR queue, I eventually took @rycee 's recommendation and implemented something similar. Thanks all for helping me out on this one.

PR #3804, opened 2023-03-25
im/fcitx: drop as fcitx 4 has been removed from nixpkgs

NickCao:
Reference: https://github.com/NixOS/nixpkgs/pull/220776

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/i18n/input-method/default.nix
```diff
@@ -22,16 +22,15 @@ let
...
-    [ ./fcitx.nix ./fcitx5.nix ./hime.nix ./kime.nix ./nabi.nix ./uim.nix ];
+  imports = [ ./fcitx5.nix ./hime.nix ./kime.nix ./nabi.nix ./uim.nix ];
 
   options.i18n = {
     inputMethod = {
       enabled = mkOption {
-        type = types.nullOr
-          (types.enum [ "fcitx" "fcitx5" "nabi" "uim" "hime" "kime" ]);
+        type =
+          types.nullOr (types.enum [ "fcitx5" "nabi" "uim" "hime" "kime" ]);
```
ncfavier, line 31: Maybe instead of removing it, add an assertion to tell users to move to fcitx5?

modules/i18n/input-method/default.nix
```diff
@@ -42,14 +41,6 @@ in {
           Currently the following input methods are available in Home Manager:
 
           <variablelist>
-          <varlistentry>
-            <term><literal>fcitx</literal></term>
-            <listitem><para>
-              A customizable lightweight input method
```
ncfavier, line 48: Move this to the description for fcitx5?

ncfavier (approved):

marsam (approved):

PR #548, opened 2019-01-23
fish: use global for abbr

Jomik:
Makes fish use global scope for abbreviations.
This makes it so that they don't stick across config changes.
Before, an abbreviation would still exist even if removed from the config.

PR #3806, opened 2023-03-25
fish: support flexible abbreviations

### LINKED ISSUE #3706: Support new abbr features for `programs.fish.shellAbbrs`
### Description
[Fish 3.6.0](https://fishshell.com/docs/current/relnotes.html#fish-3-6-0-released-january-7-2023) brought more flexibility to abbreviations, like enabling them anywhere in the command line (for sth like expanding `sudo !!` to `sudo <last command>`) or regexp matching. Home-manager's `programs.fish.shellAbbrs` [currently only supports](https://github.com/nix-community/home-manager/blob/693d76eeb84124cc3110793ff127aeab3832f95c/modules/programs/fish.nix#L142) plain alias-like abbreviations. It would be nice to see this support enhanced for the new capabilities.

Also, `--global`, used in the current implementation, is removed and now just ignored for backwards compatibility.

natsukium:
### Description

resolves #3706 
https://github.com/fish-shell/fish-shell/releases/tag/3.6.0

I added abbrModule with referring to the functionModule implementation. 
abbrModule follows the "add" subcommand. (https://fishshell.com/docs/current/cmds/abbr.html)

The backward compatibility is preserved by setting the type of fish.shellAbbrs to attrsOf (either str abbrModule). I have confirmed this in the tests added this time.
Also, the --global option is no longer needed. I have removed it from the abbrsStr configuration.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -138,9 +138,76 @@ let
     };
   };
 
-  abbrsStr = concatStringsSep "\n"
-    (mapAttrsToList (k: v: "abbr --add --global -- ${k} ${escapeShellArg v}")
-      cfg.shellAbbrs);
+  abbrModule = types.submodule {
+    options = {
+      expansion = mkOption {
+        type = with types; nullOr str;
+        default = null;
+        description = ''
+          The command expanded by an abbreviation.
+        '';
+      };
+
+      position = mkOption {
+        type = with types; nullOr str;
+        default = null;
+        example = "anywhere";
+        description = ''
+          If the position is "command", the abbreviation expands only if
+          the position is a command. If it is "anywhere", the abbreviation
+          expands anywhere.
+        '';
+      };
+
+      regex = mkOption {
+        type = with types; nullOr str;
+        default = null;
+        description = ''
+          The regular expression pattern matched instead of the literal name.
+        '';
+      };
+
+      setCursor = mkOption {
+        type = with types; (either bool str);
+        default = false;
+        description = ''
+          The marker indicates the position of the cursor when the abbreviation
+          is expanded. When setCursor is true, the marker is set with a default
+          value of "%".
+        '';
+      };
+
+      function = mkOption {
+        type = with types; nullOr str;
+        default = null;
+        description = ''
+          The fish function expanded instead of a literal string.
+        '';
+      };
+    };
+  };
+
+  abbrsStr = concatStringsSep "\n" (mapAttrsToList (name: def:
+    let
+      mods = with def;
+        cli.toGNUCommandLineShell {
+          mkOption = k: v:
+            if v == null then
+              [ ]
+            else if k == "set-cursor" then
+              [ "--${k}=${lib.generators.mkValueStringDefault { } v}" ]
+            else [
+              "--${k}"
+              (lib.generators.mkValueStringDefault { } v)
+            ];
+        } {
+          inherit position regex function;
+          set-cursor = setCursor;
+        };
+      modifiers = if isAttrs def then mods else "";
+      expansion = if isAttrs def then def.expansion else def;
+    in "abbr --add ${modifiers} -- ${name}"
+    + optionalString (expansion != null) " \"${expansion}\"") cfg.shellAbbrs);
 
   aliasesStr = concatStringsSep "\n"
     (mapAttrsToList (k: v: "alias ${k} ${escapeShellArg v}") cfg.shellAliases);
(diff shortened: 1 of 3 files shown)


bhankas:
Bad bot

berbiche (approved):
LGTM!
I am deferring merging to active contributors.

tests/modules/programs/fish/abbrs.nix
```diff
@@ -0,0 +1,62 @@
...
+          expansion = "--color";
+        };
+        L = {
+          position = "anywhere";
+          setCursor = true;
+          expansion = "% | less";
+        };
+        "!!" = {
+          position = "anywhere";
+          function = "last_histroy_item";
```
berbiche, line 20: There's a typo here, not that it matters

modules/programs/fish.nix
```diff
@@ -138,9 +138,72 @@ let
...
+        else
+          optional (v != null) "--${n}=${toString v}";
+      mods = with def;
+        modifierStr "position" position ++ modifierStr "regex" regex
+        ++ modifierBoolOrStr "set-cursor" setCursor
+        ++ modifierStr "function" function;
+      modifiers = if isAttrs def then toString mods else "";
+      expansion = if isAttrs def then def.expansion else def;
+    in "abbr --add ${modifiers} -- ${name}"
+    + optionalString (expansion != null) " \"${expansion}\"") cfg.shellAbbrs);
```
ncfavier, line 206: This is reinventing too many wheels. At least `escapeShellArg`(`s`), maybe [`toGNUCommandLine`](https://github.com/NixOS/nixpkgs/blob/master/lib/cli.nix).

ncfavier: Right so `toGNUCommandLine` is perhaps not the right tool here, but please at least generate a list of words and use `escapeShellArgs` on it.

modules/programs/fish.nix
```diff
@@ -138,9 +138,72 @@ let
     };
   };
 
-  abbrsStr = concatStringsSep "\n"
-    (mapAttrsToList (k: v: "abbr --add --global -- ${k} ${escapeShellArg v}")
```
ncfavier, line 142: Why are we removing `--global`? Is that not supported any more, I can't find it in the docs?

ncfavier: Introduced in https://github.com/nix-community/home-manager/pull/548

ncfavier: Looks like this is the [default behaviour](https://github.com/fish-shell/fish-shell/blob/c7b43b3abfb24af324100667932b168f0ef566e3/fish-rust/src/builtins/abbr.rs#L545-L547) now, so should be safe to remove.

ncfavier: Oops, you said that in the OP.

tests/modules/programs/fish/abbrs.nix
natsukium: Thanks, I fixed it.

modules/programs/fish.nix
natsukium: Thanks for the suggestion.
I rewrote the `modifiers` using `toGNUCommandLineShell`.
On the other hand, I left `expansion` as is because, in [some cases](https://github.com/nix-community/home-manager/pull/3806/files#diff-ab9e6378ee2cecd492130a366ed79158c145115d81928d86c93a2285fef1a2e1R27-R31), it did not work correctly when using `escapeShellArg`.

ncfavier: You'd be reverting https://github.com/nix-community/home-manager/commit/a08dabf015081666f50ad1538f85d69e2217585e and maybe breaking @cole-h 's configuration.

In the case you're pointing to, it seems like

```nix
expansion = ''
  for dir in */
    cd $dir
    !
    cd ..
  end'';
```

would do. Can you actually think of use cases for allowing definition-time expansion in the `expansion`?

pvsr:
I tested all of the new options and didn't run into any issues. Code looks good to me too. It would be nice to make `expansion` and `function` mutually exclusive but it's not crucial since it doesn't really make sense to set both anyway. I'm in favor of merging.

modules/programs/fish.nix
pvsr: I agree that definition-time expansion isn't necessary. `function` is a more flexible way to build abbrs from arbitrary code
```fish
functions.__4DIRS = "string join \\n -- 'for dir in */' 'cd $dir' '!' 'cd ..' 'end'";
shellAbbrs.4DIRS = {
  setCursor = "!";
  function = "__4DIRS";
};
```

If the function is slow and you don't mind slowing down shell startup instead you can still throw it in `interactiveShellInit` instead of using the abbr module. That's an inconvenience but imo it's outweighed by the convenience of not having to manually escape abbr bodies, and of course backwards compatibility.

pvsr (commented):
I tested all of the new options and didn't run into any issues. Aside from the open thread about shell-escaping the abbr body the code looks good to me. It would be nice to make `expansion` and `function` mutually exclusive but it's not crucial since it doesn't really make sense to set both anyway.

rycee:
Thanks! Merged to master now. Sorry about the delay!

Issue #3724, opened 2023-03-01
Support fish theme plugins


d12bb:
[Fish 3.4](https://fishshell.com/docs/current/relnotes.html#fish-3-4-0-released-march-12-2022) added support for themes, so theme plugins (like [Catppuccin](https://github.com/catppuccin/fish)) now often ship as `themes/foo.theme`, but `programs.fish.plugins` completely ignores this directory and files, so one can't use it for adding a theme.

NovaViper:
Sadly had to revert the commit that closed this issue 😔 Working on another one that will *hopefully* get the feature added properly

PR #3809, opened 2023-03-26
feat(fish): theme support

caarlos0:
### Description

Adds very rudimentary support for themes in fish.

Right now it is linking on every shell, which is not ideal... I'm new to nix, is there a way to do the linking just when "installing" the plugin? Once I figure that out (or if someone help me :)) I'll fix the remaining checklist items. :) 

refs #3724
refs https://github.com/fish-shell/fish-shell/issues/9456

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -416,8 +416,17 @@ in {
     # in the paths and any initialization scripts.
     (mkIf (length cfg.plugins > 0) {
       xdg.configFile = mkMerge ((map (plugin: {
-        "fish/conf.d/plugin-${plugin.name}.fish".source =
-          fishIndent "${plugin.name}.fish" ''
+        "fish/conf.d/plugin-${plugin.name}.fish" = {
+          onChange = fishIndent ''
+            if test -d $plugin_dir/themes
+              # https://github.com/fish-shell/fish-shell/issues/9456
+              mkdir -p $__fish_config_dir/themes
+              for f in $plugin_dir/themes/*.theme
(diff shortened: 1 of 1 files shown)


modules/programs/fish.nix
```diff
@@ -430,6 +430,14 @@ in {
               set fish_complete_path $fish_complete_path[1] $plugin_dir/completions $fish_complete_path[2..-1]
             end
 
+            if test -d $plugin_dir/themes
+              # https://github.com/fish-shell/fish-shell/issues/9456
+              mkdir -p $__fish_config_dir/themes
+              for f in $plugin_dir/themes/*.theme
+                ln -s $f $__fish_config_dir/themes/(basename $f)
```
rycee, line 437: If the link exists from a previous run, wouldn't this cause an error message? Perhaps need to force the link creation?

```suggestion
                ln -sf $f $__fish_config_dir/themes/(basename $f)
```

caarlos0: indeed, thanks for the review!

stelcodes:
Just want to note that this method would leave uninstalled theme files around

stelcodes:
Is there a reason why this shouldn't be in an `onChange` hook script?

caarlos0:
didn't know about onChange.

So, instead of doing it where I'm doing now, do the linking in the `onChange` hook of `"fish/conf.d/plugin-${plugin.name}.fish"`?

shirok1:
Hello, is there any news?

PR #3811, opened 2023-03-26
`git-workspace`: add module and service

aciceri:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```


aciceri:
Didn't add any new test, the problem is that to reasonably test `git-workspace` we should include a Github or Gitlab token in the test. Also I'm not sure how `home-manager` tests are executed, if they can access internet to fetch repositories or not.

As soon as I'm able to build the docs to check how new options are rendered (and probably I'll have to fix them) I'll mark this PR as ready but I think that, except this, should be already ready.

aciceri:
It took a bit but this is finally ready. I'm using it since March.

.github/CODEOWNERS
sumnerevans, line 1: We got rid of this file a few weeks ago. Please delete this.

modules/programs/git-workspace.nix
```diff
@@ -0,0 +1,15 @@
...
+in {
+  meta.maintainers = [ lib.maintainers.aciceri ];
+  options.programs.git-workspace = {
+    enable = lib.mkEnableOption "git-workspace";
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.git-workspace;
+      defaultText = "pkgs.git-workspace";
+      description = "The git-workspace to use";
+    };
```
sumnerevans, line 12:
```suggestion
    package = mkPackageOption pkgs "git-workspace" { };
```

modules/services/git-workspace.nix
```diff
@@ -0,0 +1,153 @@
...
+        more information on the syntax.
+      '';
+    };
+    environmentFile = lib.mkOption {
+      type = lib.types.path;
+      description = ''
+        Path to the systemd <literal>EnvironmentFile</literal> to keep tokens,
+        do not put a derivation output here since it would expose secrets in
+        the world readable Nix store.
+
```
sumnerevans, line 61:
```suggestion
            </para><para>
```

modules/services/git-workspace.nix
```diff
@@ -0,0 +1,153 @@
...
+        It contains lines like
+        <literal>
+        GITHUB_TOKEN=ghp_...
+        </literal>
+
```
sumnerevans, line 66:
```suggestion
            </para><para>
```

sumnerevans (changes requested):

modules/programs/git-workspace.nix
```diff
@@ -0,0 +1,10 @@
+{ config, lib, pkgs, ... }:
+let cfg = config.programs.git-workspace;
+in {
+  meta.maintainers = [ lib.maintainers.aciceri ];
+  options.programs.git-workspace = {
+    enable = lib.mkEnableOption "git-workspace";
+    package = lib.mkPackageOption pkgs "git-workspace" { };
+  };
+  config = lib.mkIf cfg.enable { home.packages = [ pkgs.git-workspace ]; };
```
rycee, line 9: I assume you meant to use

```suggestion
  config = lib.mkIf cfg.enable { home.packages = [ cfg.package ]; };
```

here.

But I would suggest simply adding

to the service module and removing this file entirely.

modules/services/git-workspace.nix
```diff
@@ -0,0 +1,153 @@
...
+    freeformType = tomlFormat.type;
+    options.provider = lib.mkOption {
+      type = lib.types.listOf (lib.types.submodule {
+        freeformType = tomlFormat.type;
+        options = {
+          provider = lib.mkOption {
+            type = lib.types.str;
+            example = "github";
+          };
+          name = lib.mkOption {
```
rycee, line 15: Please separate options by empty lines to help make things a bit more readable:

```suggestion

          name = lib.mkOption {
```

modules/services/git-workspace.nix
```diff
@@ -0,0 +1,153 @@
+{ config, lib, pkgs, ... }:
+let
+  cfg = config.services.git-workspace;
+  tomlFormat = pkgs.formats.toml { };
+  workspaceType = lib.types.submodule {
+    freeformType = tomlFormat.type;
+    options.provider = lib.mkOption {
```
rycee, line 7: All the `lib.`s are a bit overwhelming. If you don't want to use `with lib` then can put something like

``` nix
inherit (lib) mkOption literalExpression …;
```

in the top-level `let` bindings.

modules/services/git-workspace.nix
```diff
@@ -0,0 +1,153 @@
...
+      type = lib.types.listOf (lib.types.submodule {
+        freeformType = tomlFormat.type;
+        options = {
+          provider = lib.mkOption {
+            type = lib.types.str;
+            example = "github";
```
rycee, line 13: I think some brief description would be nice for each of these options.

modules/services/git-workspace.nix
```diff
@@ -0,0 +1,153 @@
...
+          };
+        };
+      });
+    };
+  };
+in {
+  meta.maintainers = [ lib.maintainers.aciceri ];
+  options.services.git-workspace = {
+    enable = lib.mkEnableOption "git-workspace systemd timer";
+    package = lib.mkOption {
```
rycee, line 32: Again some empty lines help when reading this, I think. For example

```suggestion
  meta.maintainers = [ lib.maintainers.aciceri ];
  
  options.services.git-workspace = {
    enable = lib.mkEnableOption "git-workspace systemd timer";
    
    package = lib.mkOption {
```

and so on…

modules/services/git-workspace.nix
```diff
@@ -0,0 +1,153 @@
...
+      default = if config.programs.git-workspace.enable then
+        config.programs.git-workspace.package
+      else
+        pkgs.git-workspace;
+      defaultText = lib.literalExpression ''
+        if config.programs.git-workspace.enable then
+          config.programs.git-workspace.package
+        else
+          pkgs.git-workspace;
+      '';
```
rycee, line 43: With removing the programs module this could be simplified, indeed the entire option could be defined using `mkPackageOption`.

modules/services/git-workspace.nix
```diff
@@ -0,0 +1,153 @@
...
+      description = ''
+        Path to the systemd <literal>EnvironmentFile</literal> to keep tokens,
+        do not put a derivation output here since it would expose secrets in
+        the world readable Nix store.
+            </para><para>
+        It contains lines like
+        <literal>
+        GITHUB_TOKEN=ghp_...
+        </literal>
+            </para><para>
```
rycee, line 66: Indentation
```suggestion
        </para><para>
        It contains lines like
        <literal>
        GITHUB_TOKEN=ghp_...
        </literal>
        </para><para>
```

modules/services/git-workspace.nix
```diff
@@ -0,0 +1,153 @@
...
+              provider = "github"
+              name = "github_username"
+              path = "''${config.home.homeDirectory}/personal_projects"
+              skip_forks = false;
+            }
+            {
+              provider = "gitlab"
+              name = "gitlab_username"
+              # it's possible to use the same path as previous provider
+              # if there are not name clashes
```
rycee, line 92:
```suggestion
              # It's possible to use the same path as a previous provider
              # if there are no name clashes.
```

modules/services/git-workspace.nix
```diff
@@ -0,0 +1,153 @@
...
+        };
+      '';
+    };
+  };
+  config = lib.mkIf cfg.enable {
+    assertions = [
+      (lib.hm.assertions.assertPlatform "services.git-workspace" pkgs
+        lib.platforms.linux)
+    ];
+    xdg.configFile = lib.mapAttrs' (workspaceName: workspace: {
```
rycee, line 112: Can separate the top-level attributes in the config section, i.e.,
```suggestion

    xdg.configFile = lib.mapAttrs' (workspaceName: workspace: {
```

etc.

modules/services/git-workspace.nix
```diff
@@ -0,0 +1,153 @@
...
+      name = "git-workspace/${workspaceName}/workspace.toml";
+      value.source = (tomlFormat.generate "${workspaceName}-workspace.toml"
+        workspace).outPath;
+    }) cfg.workspaces;
+    systemd.user.services = lib.mapAttrs' (workspaceName: workspace: rec {
+      name = "git-workspace-${workspaceName}-update";
```
rycee, line 118: Could perhaps avoid the `rec` by doing a let binding:
```suggestion
    systemd.user.services = lib.mapAttrs' (workspaceName: workspace:
      let name = "git-workspace-${workspaceName}-update"; in {
        inherit name;
```

modules/services/git-workspace.nix
```diff
@@ -0,0 +1,153 @@
...
+      value = {
+        Unit.Description = "Runs `git-workspace update` for ${workspaceName}";
```
rycee, line 120: Could simplify the description a bit

```suggestion
        Unit.Description = "git-workspace update for ${workspaceName}";
```

Then it should also look a bit nicer in the systemd journal:

> Starting git-workspace update for blah...

modules/services/git-workspace.nix
```diff
@@ -0,0 +1,153 @@
...
+        Service = {
+          EnvironmentFile = cfg.environmentFile;
+          ExecStart = let
+            script = pkgs.writeShellApplication {
+              name = "${name}-launcher";
+              text = ''
+                ${cfg.package}/bin/git-workspace \
```
rycee, line 127: Maybe
```suggestion
                ${getExe cfg.package} \
```
would work?

modules/services/git-workspace.nix
```diff
@@ -0,0 +1,153 @@
...
+                  --workspace ${config.xdg.configHome}/git-workspace/${workspaceName} \
+                  update
```
rycee, line 129: I think this may have problems if `configHome` or `workspaceName` contains, e.g., whitespaces. Perhaps can do

```suggestion
                  ${escapeShellArgs [ "--workspace" "${config.xdg.configHome}/git-workspace/${workspaceName}"
                  "update"]}
```

modules/services/git-workspace.nix
```diff
@@ -0,0 +1,153 @@
...
+      value.source = (tomlFormat.generate "${workspaceName}-workspace.toml"
+        workspace).outPath;
+    }) cfg.workspaces;
+    systemd.user.services = lib.mapAttrs' (workspaceName: workspace: rec {
+      name = "git-workspace-${workspaceName}-update";
+      value = {
+        Unit.Description = "Runs `git-workspace update` for ${workspaceName}";
+        Service = {
+          EnvironmentFile = cfg.environmentFile;
+          ExecStart = let
```
rycee, line 123: I don't think you need to wrap this in a shell script, should be possible to immediately call the application. The runtime dependencies could be added through PATH:

```suggestion
          Environment = [ "PATH=${with pkgs; makeBinPath [busybox openssh git]}" ];
          ExecStart = "${cfg.package}/bin/git-workspace …"
```

It does seem a bit odd to me, through, that the package itself does not add these as dependencies…

modules/services/git-workspace.nix
```diff
@@ -0,0 +1,153 @@
...
+          in "${script}/bin/${name}-launcher";
+        };
+      };
+    }) cfg.workspaces;
+    systemd.user.timers = lib.mapAttrs' (workspaceName: workspace: {
+      name = "git-workspace-${workspaceName}-update";
+      value = {
+        Unit = {
+          Description =
+            "Automatically runs `git-workspace update` for ${workspaceName}";
```
rycee, line 142: Can simplify here as well
```suggestion
            "git-workspace update for ${workspaceName}";
```

modules/services/git-workspace.nix
```diff
@@ -0,0 +1,153 @@
...
+    xdg.configFile = lib.mapAttrs' (workspaceName: workspace: {
+      name = "git-workspace/${workspaceName}/workspace.toml";
+      value.source = (tomlFormat.generate "${workspaceName}-workspace.toml"
+        workspace).outPath;
+    }) cfg.workspaces;
+    systemd.user.services = lib.mapAttrs' (workspaceName: workspace: rec {
+      name = "git-workspace-${workspaceName}-update";
+      value = {
+        Unit.Description = "Runs `git-workspace update` for ${workspaceName}";
+        Service = {
```
rycee, line 121: I wonder if, perhaps, this service shouldn't be marked as a oneshot one? I.e.,
```suggestion
        Service = {
          Type = "oneshot";
```

modules/services/git-workspace.nix
rycee: Looking at it a bit more, I'm not sure how this would interact with the user adding PATH to their environment file… But I suppose, that is not something they should be doing? 🙂

rycee:
Thanks for the contribution! I've added a bunch of comments, in general eveything looks just fine so the comments are mostly stylistic ones.

modules/programs/git-workspace.nix
aciceri: Yes, having a `progams` module like this is pretty insignificant indeed, I'm removing it.

modules/services/git-workspace.nix
aciceri: Doing that, but only for things that appear multiple times.

aciceri: Doing that, but not with single types.

modules/services/git-workspace.nix
aciceri: if the user does it he deserves potential consequences 😈

modules/services/git-workspace.nix
aciceri: Oh you are right

aciceri:
@rycee Could you please trigger the workflow?

PR #3831, opened 2023-04-02
nushell: introduce settings option

bobvanderlinden:
Currently it is convoluted to set basic configuration options for Nushell. It always requires "let-env config =". Because multiple sources need to set these options from home-maanger, the config needs to be updated as well, resulting in multiple "let-env config" statements that update potentially existing config values.

When combined with other home-manager options that add to the Nushell configuration it is error-prone to set options yourself. New users will want to set "show_banner = false", but it is non-trivial when direnv integration is enabled.

This change makes sure all settings can be set at once. The settings are serialized to Nu code.

In addition, the settings also support Nu expressions, which allows referring to variables set in env.nu as well as using more complex expressions.

Lastly, the test examples have been updated with new (nested) configuration options. See
https://www.nushell.sh/blog/2022-11-29-nushell-0.72.html#config-options-have-been-grouped-together

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/nushell.nix
+++ b/modules/programs/nushell.nix
@@ -38,18 +38,49 @@ let
         };
       };
     });
+
+  formatNuValue = let
+    indentation = "  ";
+    indent = str:
+      let
+        lines = splitString "\n" str;
+        indentedLines = map (line: "${indentation}${line}") lines;
+      in lib.concatStringsSep "\n" indentedLines;
+  in value:
+  {
+    bool = v: if v then "true" else "false";
+    int = toString;
+    float = toString;
+    string = builtins.toJSON;
+    null = v: "null";
+    path = v: formatNuValue (toString v);
+    list = v: ''
+      [${
+        lib.concatStrings (map (v: ''
+
+          ${indent (formatNuValue v)}'') v)
+      }
+      ]'';
+    set = v:
+      if nuExpressionType.check v then
+        v.__nu
+      else ''
+        {${
+          lib.concatStrings (mapAttrsToList (k: v: ''
+
+            ${indent "${k}: ${formatNuValue v}"}'') v)
+        }
+        }'';
+  }.${builtins.typeOf value} value;
+
+  nuExpressionType = mkOptionType {
+    name = "nu";
+    description = "Nu expression";
+    check = x: isAttrs x && x ? __nu && isString x.__nu;
+    merge = mergeEqualOption;
+  };
 in {
   meta.maintainers = [ maintainers.Philipp-M ];
-
-  imports = [
-    (mkRemovedOptionModule [ "programs" "nushell" "settings" ] ''
-      Please use
-
-        'programs.nushell.configFile' and 'programs.nushell.envFile'
-
-      instead.
-    '')
-  ];
 
   options.programs.nushell = {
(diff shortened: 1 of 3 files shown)


bobvanderlinden:
@Philipp-M Could you have a look?

bobvanderlinden:
Thanks for the thoughtful comment, suggestions and shameless pings :smile:

I get why this wouldn't be accepted, but with my current configuration file becoming quite messy I thought it was worth a shot. With something like this PR the config.nu becomes a lot more like what users would manually write in their config file. Easier to grasp what is going on, easier for newcomers to add `show_banner = false` and easier for integrations to interop with each-other (especially when introducing conflicting settings that would otherwise silently overwrite eachother depending on order :sweat_smile:).

I've incorporated an assertion for nushell >= 0.73 and added myself as maintainer. :+1:

Nyabinary:
Is this uh still dead?

bobvanderlinden:
It's here if we want such a feature. I'm currently not actively using nushell anymore, so it's fine to leave this open or close it. When I was using nushell this helped me disable the banner more easily.

Maybe it helps if you want this feature to give some motivation/reason/usecase to adopt this.

Nyabinary:
> It's here if we want such a feature. I'm currently not actively using nushell anymore, so it's fine to leave this open or close it. When I was using nushell this helped me disable the banner more easily.
> 
> Maybe it helps if you want this feature to give some motivation/reason/usecase to adopt this.

Well I want to use it :3 I think that valid usecase enough? @rycee This is good to merge right?

nyabinary:
Merge conflict btw :3

Philipp-M:
I'm happy to review an updated PR, though I'm not able to maintain it effectively, since I'm not using nushell actively currently (new maintainers welcome), so I would rely on issues and/or PRs.

Lord-Valen:
This won't work with >= 0.83. [`let-env` was removed](https://www.nushell.sh/blog/2023-07-25-nushell_0_83.html#breaking-changes). We must use `$env.config =` instead. That should actually simplify the implementation for us, thankfully.

nyabinary:
@Philipp-M What does this need to be merged :P

Philipp-M:
Well my last comment is still accurate.

I'm not using nushell currently, and have not a deep insight what needs to be done exactly (and what has changed in nushell since I last used it), but I'm happy to review an updated PR, which someone needs to do (I'm not having the time for that currently).
AFAIK this is a little bit out of date, so if you want to do it, go for it.
I would also be happy, if someone else would (additionally) maintain the nushell module, #5500 may be also relevant.

nyabinary:
> Well my last comment is still accurate.
> 
> I'm not using nushell currently, and have not a deep insight what needs to be done exactly (and what has changed in nushell since I last used it), but I'm happy to review an updated PR, which someone needs to do (I'm not having the time for that currently). AFAIK this is a little bit out of date, so if you want to do it, go for it. I would also be happy, if someone else would (additionally) maintain the nushell module, #5500 may be also relevant.

Commented this because with the new update nushell broke for me :P https://github.com/nix-community/home-manager/issues/5527

nyabinary:
Superseded by #6184

PR #3832, opened 2023-04-02
nushell: add support for shellAliases

bobvanderlinden:
Currently extraConfig needs to be used for defining aliases for Nushell. Other shells inherit from home.shellAliases. It would be nice if home-manager did this for Nushell as well.

This change adds a shellAliases option that creates `alias k = v` entries in `config.nu`.

Nushell's `shellAliases` is set to `home.shellAlises` in the same way fish' shellAliases option does.

### Checklist

I got an (seemingly) unrelated error:

```
error: The option `programs.borgmatic.backups.main.location.extraConfig.exclude_from' has conflicting definition values:
```

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/home-environment.nix
+++ b/modules/home-environment.nix
@@ -513,6 +513,7 @@ in
     programs.bash.shellAliases = cfg.shellAliases;
     programs.zsh.shellAliases = cfg.shellAliases;
     programs.fish.shellAliases = cfg.shellAliases;
+    programs.nushell.shellAliases = cfg.shellAliases;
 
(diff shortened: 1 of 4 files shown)


rycee:
Thanks for the contribution!

There is an existing PR #3529 adding the same feature. It is not merged, though, due to open questions related to escaping and handling of command line arguments. I'll close this in favor of the old one. If you have any suggestion how to advance the old PR then please add a comment on that PR 🙂

PR #3837, opened 2023-04-03
hyprland: init

fufexan:
### Description

Add the Hyprland module, ported from
https://github.com/hyprwm/Hyprland/blob/main/nix/hm-module.nix,
which was adapted from the sway module.

CC: @Mic92

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -1144,6 +1144,14 @@ in
           A new module is available: 'services.swayosd'
         '';
       }
+
+      {
+        time = "2023-07-20T21:56:49+00:00";
+        condition = hostPlatform.isLinux;
+        message = ''
+          A new module is available: 'wayland.windowManager.hyprland'
+        '';
+      }
     ];
   };
 }
--- a/modules/services/window-managers/hyprland.nix
+++ b/modules/services/window-managers/hyprland.nix
@@ -1,0 +1,216 @@
+{ config, lib, pkgs, ... }:
+let
+
+  cfg = config.wayland.windowManager.hyprland;
+
+in {
+  meta.maintainers = [ lib.maintainers.fufexan ];
+
+  # A few option removals and renames to aid those migrating from the upstream
+  # module.
+  imports = [
+    (lib.mkRemovedOptionModule # \
+      [ "wayland" "windowManager" "hyprland" "disableAutoreload" ]
+      "Autoreloading now always happen")
+
+    (lib.mkRemovedOptionModule # \
+      [ "wayland" "windowManager" "hyprland" "recommendedEnvironment" ]
+      "Recommended environment variables are now always set")
+
+    (lib.mkRenamedOptionModule # \
+      [ "wayland" "windowManager" "hyprland" "nvidiaPatches" ] # \
+      [ "wayland" "windowManager" "hyprland" "enableNvidiaPatches" ])
+  ];
+
+  options.wayland.windowManager.hyprland = {
+    enable = lib.mkEnableOption "Hyprland wayland compositor";
+
+    package = lib.mkPackageOption pkgs "hyprland" { };
+
+    finalPackage = lib.mkOption {
+      type = lib.types.package;
+      readOnly = true;
+      default = cfg.package.override {
+        enableXWayland = cfg.xwayland.enable;
+        hidpiXWayland = cfg.xwayland.hidpi;
+        nvidiaPatches = cfg.enableNvidiaPatches;
+      };
+      defaultText = lib.literalMD
+        "`wayland.windowManager.hyprland.package` with applied configuration";
+      description = ''
+        The Hyprland package after applying configuration.
+      '';
+    };
+
+    plugins = lib.mkOption {
+      type = with lib.types; listOf (either package path);
+      default = [ ];
+      description = ''
+        List of Hyprland plugins to use. Can either be packages or
+        absolute plugin paths.
+      '';
+    };
+
+    systemdIntegration = lib.mkOption {
+      type = lib.types.bool;
+      default = pkgs.stdenv.isLinux;
+      description = ''
+        Whether to enable {file}`hyprland-session.target` on
+        hyprland startup. This links to `graphical-session.target`.
+        Some important environment variables will be imported to systemd
+        and dbus user environment before reaching the target, including
+        - `DISPLAY`
+        - `HYPRLAND_INSTANCE_SIGNATURE`
+        - `WAYLAND_DISPLAY`
+        - `XDG_CURRENT_DESKTOP`
+      '';
+    };
+
+    xwayland = {
+      enable = lib.mkEnableOption "XWayland" // { default = true; };
+      hidpi = lib.mkEnableOption null // {
+        description = ''
+          Enable HiDPI XWayland, based on [XWayland MR 733](https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/733).
+          See <https://wiki.hyprland.org/Nix/Options-Overrides/#xwayland-hidpi> for more info.
+        '';
+      };
+    };
+
+    enableNvidiaPatches =
+      lib.mkEnableOption "patching wlroots for better Nvidia support";
+
+    settings = lib.mkOption {
+      type = with lib.types;
+        let
+          valueType = nullOr (oneOf [
+            bool
+            int
+            float
+            str
+            path
+            (attrsOf valueType)
+            (listOf valueType)
+          ]) // {
+            description = "Hyprland configuration value";
+          };
+        in valueType;
+      default = { };
+      description = ''
+        Hyprland configuration written in Nix. Entries with the same key
+        should be written as lists. Variables' and colors' names should be
+        quoted. See <https://wiki.hyprland.org> for more examples.
+
+        ::: {.note}
+        Use the [](#opt-wayland.windowManager.hyprland.plugins) option to
+        declare plugins.
+        :::
+
+      '';
+      example = lib.literalExpression ''
+        {
+          decoration = {
+            shadow_offset = "0 5";
+            "col.shadow" = "rgba(00000099)";
+          };
+
+          "$mod" = "SUPER";
+
+          bindm = [
+            # mouse movements
+            "$mod, mouse:272, movewindow"
+            "$mod, mouse:273, resizewindow"
+            "$mod ALT, mouse:272, resizewindow"
+          ];
+        }
+      '';
+    };
+
+    extraConfig = lib.mkOption {
+      type = lib.types.lines;
+      default = "";
+      example = ''
+        # window resize
+        bind = $mod, S, submap, resize
+
+        submap = resize
+        binde = , right, resizeactive, 10 0
+        binde = , left, resizeactive, -10 0
+        binde = , up, resizeactive, 0 -10
+        binde = , down, resizeactive, 0 10
+        bind = , escape, submap, reset
+        submap = reset
+      '';
+      description = ''
+        Extra configuration lines to add to `~/.config/hypr/hyprland.conf`.
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    assertions = [
+      (lib.hm.assertions.assertPlatform "wayland.windowManager.hyprland" pkgs
+        lib.platforms.linux)
+    ];
+
+    warnings = let
+      inconsistent = (cfg.systemdIntegration || cfg.plugins != [ ])
+        && cfg.extraConfig == "" && cfg.settings == { };
+      warning =
+        "You have enabled hyprland.systemdIntegration or listed plugins in hyprland.plugins but do not have any configuration in hyprland.settings or hyprland.extraConfig. This is almost certainly a mistake.";
+    in lib.optional inconsistent warning;
+
+    home.packages = lib.optional (cfg.package != null) cfg.finalPackage;
+
+    xdg.configFile."hypr/hyprland.conf" = let
+      combinedSettings = cfg.settings // {
+        plugin = let
+          mkEntry = entry:
+            if lib.types.package.check entry then
+              "${entry}/lib/lib${entry.pname}.so"
+            else
+              entry;
+        in map mkEntry cfg.plugins;
+      };
+
+      shouldGenerate = cfg.systemdIntegration || cfg.extraConfig != ""
+        || combinedSettings != { };
+
+      toHyprconf = attrs:
+        let
+          mkSection = n: attrs: ''
+            ${n} {
+              ${toHyprconf attrs}}
+          '';
+          mkFields = lib.generators.toKeyValue { listsAsDuplicateKeys = true; };
+          sections = lib.filterAttrs (n: v: lib.isAttrs v) attrs;
+          fields = lib.filterAttrs (n: v: !(lib.isAttrs v)) attrs;
+        in lib.concatStringsSep "\n" (lib.mapAttrsToList mkSection sections)
+        + mkFields fields;
+    in lib.mkIf shouldGenerate {
+      text = lib.optionalString cfg.systemdIntegration ''
+        exec-once = ${pkgs.dbus}/bin/dbus-update-activation-environment --systemd DISPLAY WAYLAND_DISPLAY HYPRLAND_INSTANCE_SIGNATURE XDG_CURRENT_DESKTOP && systemctl --user start hyprland-session.target
+      '' + lib.optionalString (combinedSettings != { })
+        (toHyprconf combinedSettings)
+        + lib.optionalString (cfg.extraConfig != "") cfg.extraConfig;
+      onChange = lib.mkIf (cfg.package != null) ''
+        (  # execute in subshell so that `shopt` won't affect other scripts
+          shopt -s nullglob  # so that nothing is done if /tmp/hypr/ does not exist or is empty
+          for instance in /tmp/hypr/*; do
+            HYPRLAND_INSTANCE_SIGNATURE=''${instance##*/} ${cfg.finalPackage}/bin/hyprctl reload config-only \
+              || true  # ignore dead instance(s)
+          done
+        )
+      '';
+    };
+
+    systemd.user.targets.hyprland-session = lib.mkIf cfg.systemdIntegration {
+      Unit = {
+        Description = "Hyprland compositor session";
+        Documentation = [ "man:systemd.special(7)" ];
+        BindsTo = [ "graphical-session.target" ];
+        Wants = [ "graphical-session-pre.target" ];
+        After = [ "graphical-session-pre.target" ];
+      };
+    };
+  };
+}
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -330,6 +330,7 @@ let
     ./services/window-managers/bspwm/default.nix
     ./services/window-managers/fluxbox.nix
     ./services/window-managers/herbstluftwm.nix
+    ./services/window-managers/hyprland.nix
     ./services/window-managers/i3-sway/i3.nix
     ./services/window-managers/i3-sway/sway.nix
     ./services/window-managers/i3-sway/swaynag.nix
--- a/tests/default.nix
+++ b/tests/default.nix
@@ -232,6 +232,7 @@ import nmt {
     ./modules/services/udiskie
     ./modules/services/window-managers/bspwm
     ./modules/services/window-managers/herbstluftwm
+    ./modules/services/window-managers/hyprland
     ./modules/services/window-managers/i3
     ./modules/services/window-managers/spectrwm
     ./modules/services/window-managers/sway
--- a/tests/modules/services/window-managers/hyprland/default.nix
+++ b/tests/modules/services/window-managers/hyprland/default.nix
@@ -1,0 +1,4 @@
+{
+  hyprland-simple-config = ./simple-config.nix;
+  hyprland-inconsistent-config = ./inconsistent-config.nix;
+}
--- a/tests/modules/services/window-managers/hyprland/inconsistent-config.nix
+++ b/tests/modules/services/window-managers/hyprland/inconsistent-config.nix
@@ -1,0 +1,21 @@
+{ config, lib, ... }:
+
+{
+  wayland.windowManager.hyprland = {
+    enable = true;
+    package = lib.makeOverridable
+      (attrs: config.lib.test.mkStubPackage { name = "hyprland"; }) { };
+    plugins =
+      [ "/path/to/plugin1" (config.lib.test.mkStubPackage { name = "foo"; }) ];
+  };
+
+  test.asserts.warnings.expected = [
+    "You have enabled hyprland.systemdIntegration or listed plugins in hyprland.plugins but do not have any configuration in hyprland.settings or hyprland.extraConfig. This is almost certainly a mistake."
+  ];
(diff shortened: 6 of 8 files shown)


ncfavier:
I'm not sure we want to maintain the module in home-manager if upstream has its own?

fufexan:
Alright, fair. I was just following up on [this](https://github.com/hyprwm/Hyprland/pull/870#issuecomment-1435879673).

I can close the PR if having it in two places is not suitable.

FlorianFranzen:
@ncfavier So am I expected to track all the upstream repositories of software with nix support to get home-manager integration? That does not really sound feasible to me either.

ncfavier:
I guess? Not sure why that wouldn't be feasible.

ncfavier:
To be clear, I'm not opposed to adding a hyprland module at all, *if that's what upstream wants*. As long as they maintain [their own out-of-tree module](https://wiki.hyprland.org/Nix/Hyprland-on-Home-Manager/), I don't see the point in duplicating code and work. Closing for now.

Mic92:
One advantage of the home-manager module would be that it than uses the nixpkgs package while upstream will use its own (which means user have to compile the window manager themself).

ncfavier:
(Just realised @fufexan *is* upstream... 🤦 guess I'll reopen.)

fufexan:
I'm unsure why the tests fail on a module that I haven't touched.

ncfavier:
Should be fixed by rebasing.

modules/services/window-managers/hyprland.nix
```diff
@@ -0,0 +1,120 @@
...
+      default = "";
+      description = lib.mdDoc ''
+        Extra configuration lines to add to `~/.config/hypr/hyprland.conf`.
+      '';
+    };
+
+    recommendedEnvironment = lib.mkEnableOption null // {
+      description = lib.mdDoc ''
+        Whether to set the recommended environment variables.
+      '';
```
FlorianFranzen, line 80: There are a lot of "recommended" variables for wayland (e.g. [here](https://github.com/FlorianFranzen/nixrc/blob/d3d57da19d29e9f91bd2076127b2a56a13f61a5f/pkgs/sway.nix#L15-L62) is a list of mine) so it would help to add some more info to the description.

```suggestion
      description = lib.mdDoc ''
        Whether to set the recommended environment variables (i.e. `NIXOS_OZONE_WL`).
      '';
```

FlorianFranzen (changes requested):
Found a few mostly minor issues.

modules/services/window-managers/hyprland.nix
```diff
@@ -0,0 +1,120 @@
...
+  };
+
+  config = lib.mkIf cfg.enable {
+    home.packages = lib.optional (cfg.package != null) cfg.package
+      ++ lib.optional cfg.xwayland.enable pkgs.xwayland;
+
+    home.sessionVariables =
+      lib.mkIf cfg.recommendedEnvironment { NIXOS_OZONE_WL = "1"; };
+
+    xdg.configFile."hypr/hyprland.conf" = lib.mkIf (cfg.extraConfig != null) {
```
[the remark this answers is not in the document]
FlorianFranzen: Just noticed that also the [warning](https://github.com/hyprwm/Hyprland/blob/main/nix/hm-module.nix#L119-L124) was removed, so this makes even less sense now.

modules/services/window-managers/hyprland.nix
ncfavier: If these aren't specific to Hyprland, maybe it would make sense to expose that option as `wayland.recommendedEnvironment` instead (in a separate module).

fufexan: This option does not exist in the NixOS module (discussion [here](https://github.com/NixOS/nixpkgs/pull/221730#discussion_r1157860002)), so I will also remove it from here and follow up with a PR as per @ncfavier's suggestion.

modules/services/window-managers/hyprland.nix
```diff
@@ -0,0 +1,120 @@
...
+    recommendedEnvironment = lib.mkEnableOption null // {
+      description = lib.mdDoc ''
+        Whether to set the recommended environment variables.
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    home.packages = lib.optional (cfg.package != null) cfg.package
+      ++ lib.optional cfg.xwayland.enable pkgs.xwayland;
```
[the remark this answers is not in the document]
fufexan: `wlroots-hyprland` is always being used regardless of the hidpi option. The hidpi option does not bring any additional dependencies.
Am I missing something?

modules/services/window-managers/hyprland.nix
ncfavier: Also note the following paragraph in the NixOS 22.05 release notes:

> If you are using Wayland you can choose to use the Ozone Wayland support
  in Chrome and several Electron apps by setting the environment variable
  `NIXOS_OZONE_WL=1` (for example via
  `environment.sessionVariables.NIXOS_OZONE_WL = "1"`).
  This is not enabled by default because Ozone Wayland is
  still under heavy development and behavior is not always flawless.
  Furthermore, not all Electron apps use the latest Electron versions.

I don't know if this is still relevant.

fufexan:
> Any reason why plugin support was removed?

This PR predates the addition of plugin support. I will add it here as well.

modules/services/window-managers/hyprland.nix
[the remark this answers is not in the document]
FlorianFranzen: Here you install XWayland from pkgs, not the patched on inside the wlroots expression.

modules/services/window-managers/hyprland.nix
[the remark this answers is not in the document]
FlorianFranzen: Resolved, thanks.

modules/services/window-managers/hyprland.nix
```diff
@@ -0,0 +1,120 @@
...
+    package = lib.mkOption {
+      type = with lib.types; nullOr package;
+      default = defaultHyprlandPackage;
+      description = lib.mdDoc ''
+        Hyprland package to use. Will override the 'xwayland' and 'nvidiaPatches'
+        options.
+
+        Defaults to the one provided by the flake. Set it to
+        {package}`pkgs.hyprland` to use the one provided by nixpkgs or
+        if you have an overlay.
```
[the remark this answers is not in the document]
FlorianFranzen: Resolved, thanks.

modules/services/window-managers/hyprland.nix
FlorianFranzen: Resolved, thanks.

modules/services/window-managers/hyprland.nix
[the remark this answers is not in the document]
fufexan: It is the same in the Hyprland repo's HM module, and I haven't noticed anything amiss. I'll test it more though. It may not even be necessary to install it at all.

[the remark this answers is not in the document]
fufexan: I have removed `programs.xwayland.enable` from the Hyprland NixOS module, and `xwayland`'s listing in `home.packages` from the HM module, and it seems to have no negative effect. I'll remove it from here as well.

modules/services/window-managers/hyprland.nix
```diff
@@ -0,0 +1,129 @@
...
+        Some important environment variables will be imported to systemd
+        and dbus user environment before reaching the target, including
+        - `DISPLAY`
+        - `HYPRLAND_INSTANCE_SIGNATURE`
+        - `WAYLAND_DISPLAY`
+        - `XDG_CURRENT_DESKTOP`
+      '';
+    };
+
+    disableAutoreload = lib.mkOption {
```
rycee, line 48: Best to avoid confusion with double negation and such. I would suggest

```suggestion
    enableAutoreload = lib.mkOption {
```

with the default being `true`. That said, is there any occasion where you would not want automatic reload?

modules/services/window-managers/hyprland.nix
```diff
@@ -0,0 +1,129 @@
...
+
+  options.wayland.windowManager.hyprland = {
+    enable = lib.mkEnableOption (lib.mdDoc "Hyprland wayland compositor");
+
+    package = lib.mkOption {
+      type = with lib.types; nullOr package;
+      default = defaultHyprlandPackage;
+      description = lib.mdDoc ''
+        Hyprland package to use. Will override the 'xwayland' and 'nvidiaPatches'
+        options.
```
rycee, line 20: Seems a bit confusing, why not have the default be `pkgs.hyprland` and then add a read-only `finalPackage` option (like in the emacs module, for example).

In the config section set

    wayland.windowManager.hyprland.finalPackage = cfg.package.override { … }

modules/services/window-managers/hyprland.nix
```diff
@@ -0,0 +1,129 @@
...
+    plugins = lib.mkOption {
+      type = with lib.types; listOf (either package path);
+      default = [ ];
+      description = lib.mdDoc ''
+        List of Hyprland plugins to use. Can either be packages or
+        absolute plugin paths.
+      '';
+    };
+
+    systemdIntegration = lib.mkOption {
```
rycee, line 33: When would you not want to use systemd integration? Btw, when enabling the systemd integration I don't see any attempt at shutting down the graphical session when Hyprland exits. Will that happen automatically?

modules/services/window-managers/hyprland.nix
```diff
@@ -0,0 +1,129 @@
...
+      enable = lib.mkEnableOption (lib.mdDoc "XWayland") // { default = true; };
+      hidpi = lib.mkEnableOption null // {
+        description = lib.mdDoc ''
+          Enable HiDPI XWayland, based on [XWayland MR 733](https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/733).
+          See <https://wiki.hyprland.org/Nix/Options-Overrides/#xwayland-hidpi> for more info.
+        '';
+      };
+    };
+
+    nvidiaPatches = lib.mkEnableOption
```
rycee, line 68: Maybe introduce the enable naming here for consistency?

```suggestion
    enableNvidiaPatches = lib.mkEnableOption
```

modules/misc/news.nix
```diff
@@ -1039,6 +1039,13 @@ in
           A new module is available: 'programs.script-directory'.
         '';
       }
+
+      {
+        time = "2023-05-24T16:10:00+00:00";
+        message = ''
```
rycee, line 1045: I imagine this is Linux only 🙂

```suggestion
        condition = hostPlatform.isLinux;
        message = ''
```

modules/services/window-managers/hyprland.nix
```diff
@@ -0,0 +1,129 @@
...
+    extraConfig = lib.mkOption {
+      type = lib.types.nullOr lib.types.lines;
+      default = "";
+      description = lib.mdDoc ''
+        Extra configuration lines to add to `~/.config/hypr/hyprland.conf`.
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
```
rycee, line 80:
```suggestion
  config = lib.mkIf cfg.enable {
    assertions = [
      (lib.hm.assertions.assertPlatform "wayland.windowManager.hyprland" pkgs lib.platforms.linux)
    ];
```

modules/services/window-managers/hyprland.nix
```diff
@@ -0,0 +1,129 @@
...
+          See <https://wiki.hyprland.org/Nix/Options-Overrides/#xwayland-hidpi> for more info.
+        '';
+      };
+    };
+
+    nvidiaPatches = lib.mkEnableOption
+      (lib.mdDoc "patching wlroots for better Nvidia support.");
+
+    extraConfig = lib.mkOption {
+      type = lib.types.nullOr lib.types.lines;
```
rycee, line 72: I think you could omit the `null` case.

```suggestion
      type = lib.types.lines;
```

That said, to me it seems like it would be possible to relatively easily create an RFC-42 style `settings` option instead of `extraConfig`?

modules/services/window-managers/hyprland.nix
fufexan: I added it because of [this issue](https://github.com/hyprwm/Hyprland/issues/1255), however I think it's safe to remove it.

modules/services/window-managers/hyprland.nix
fufexan: Oh, I didn't even think of this. Thanks, will fix.

modules/services/window-managers/hyprland.nix
fufexan: Perhaps it makes sense for users that are on a distro without systemd. But they'd have a hard time using Nix by itself, let alone HM.

No, there's no attempt to shut down the graphical session. I don't know how to do this, but I'm willing to include it if you have clues. Also, it does not automatically shut down.

modules/services/window-managers/hyprland.nix
fufexan: I think it could work. Currently there's https://github.com/hyprwm/Hyprland/pull/870 which is *very* big and intricate.

Perhaps adapting `lib.generators.toINI` could work, but I'm unsure how to actually do it. I've tried 
```nix
lib: let
  libStr = lib.strings;
  libAttr = lib.attrsets;
in {
  toHyprconf = {
    # format a setting line from key and value
    mkKeyValue ? lib.generators.mkKeyValueDefault {} "=",
    # allow lists as values for duplicate keys
    listsAsDuplicateKeys ? false,
  }: attrsOfAttrs: let
    # map function to string for each key val
    mapAttrsToStringsSep = sep: mapFn: attrs:
      libStr.concatStringsSep sep
      (libAttr.mapAttrsToList mapFn attrs);
    mkSection = sectName: sectValues:
      ''
        ${sectName} {
      ''
      + (lib.generators.toKeyValue {inherit mkKeyValue listsAsDuplicateKeys;} sectValues)
      + ''
        }
      '';
  in
    # map input to hyprconf sections
    mapAttrsToStringsSep "\n" mkSection attrsOfAttrs;
}
```
but couldn't test it in a repl.

modules/services/window-managers/hyprland.nix
fufexan: Removed.

modules/services/window-managers/hyprland.nix
fufexan: Updated.

modules/services/window-managers/hyprland.nix
rycee: Here is a rough and mostly untested attempt:

```nix
toHyprconf = { }: attrs: let
  mkSection = n: attrs: ''
    ${n} {
    ${toKeyValue { listsAsDuplicateKeys = true; } attrs}
    }
  '';

  mkField = mkKeyValueDefault {} "=";

  sections = filterAttrs (n: v: isAttrs v) attrs;
  fields = filterAttrs (n: v: !(isAttrs v)) attrs;
in
  concatStringsSep "\n" (
    mapAttrsToList mkSection sections ++ mapAttrsToList mkField fields
  );
```

This does assume that Hyprland does not care about field order, but I think that is a correct assumption?

modules/services/window-managers/hyprland.nix
```diff
@@ -0,0 +1,113 @@
...
+      hidpi = lib.mkEnableOption null // {
+        description = lib.mdDoc ''
+          Enable HiDPI XWayland, based on [XWayland MR 733](https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/733).
+          See <https://wiki.hyprland.org/Nix/Options-Overrides/#xwayland-hidpi> for more info.
+        '';
+      };
+    };
+
+    enableNvidiaPatches = lib.mkEnableOption
+      (lib.mdDoc "patching wlroots for better Nvidia support.");
```
rycee, line 52: The default description of `mkEnableOption` already puts a period at the end.

```suggestion
      (lib.mdDoc "patching wlroots for better Nvidia support");
```

modules/services/window-managers/hyprland.nix
```diff
@@ -0,0 +1,113 @@
...
+  options.wayland.windowManager.hyprland = {
+    enable = lib.mkEnableOption (lib.mdDoc "Hyprland wayland compositor");
+
+    package = lib.mkPackageOptionMD pkgs "hyprland" { };
+
+    finalPackage = cfg.package.override {
+      enableXWayland = cfg.xwayland.enable;
+      hidpiXWayland = cfg.xwayland.hidpi;
+      nvidiaPatches = cfg.enableNvidiaPatches;
+    };
```
rycee, line 15: Unfortunately, it is a bit more complicated. Need something like

```suggestion
    finalPackage = lib.mkOption {
      type = lib.types.package;
      readOnly = true;
      default = cfg.package.override {
        enableXWayland = cfg.xwayland.enable;
        hidpiXWayland = cfg.xwayland.hidpi;
        nvidiaPatches = cfg.enableNvidiaPatches;
      };
      defaultText = lib.literalMD
        "`wayland.windowManager.hyprland.package` with applied configuration";
      description = ''
        The Hyprland package after applying configuration.
      '';
    };
```

modules/services/window-managers/hyprland.nix
rycee: Without something like `exec-exit`, which would run a command on shutdown it would be tricky to solve the shutdown entirely within Hyprland. I think it would require wrapping the Hyprland executable and in the wrapper do something like https://github.com/nix-community/home-manager/blob/6a1922568337e7cf21175213d3aafd1ac79c9a2e/modules/xsession.nix#L209-L219 on shutdown.

modules/services/window-managers/hyprland.nix
fufexan: That gets us closer, but there are instances of nested sections, for example
```ini
input {
  sensitivity = 0.5
  touchpad {
    natural_scroll = true
  }
}

File: modules/services/window-managers/hyprland.nix (hunk 13 above)
Comment (fufexan) [reply modules/services/window-managers/hyprland.nix:15]: Oops, forgot about that.

File: modules/services/window-managers/hyprland.nix (hunk 11 above)
Comment (rycee) [reply modules/services/window-managers/hyprland.nix:64]: Maybe something like

```nix
toHyprconf = { }: attrs: let
  mkSection = n: attrs: ''
    ${n} {
    ${toHyprconf { } attrs}}
  '';

  mkFields = generators.toKeyValue { listsAsDuplicateKeys = true; };

  sections = filterAttrs (n: v: isAttrs v) attrs;
  fields = filterAttrs (n: v: !(isAttrs v)) attrs;
in
  concatStringsSep "\n" (mapAttrsToList mkSection sections)
  + mkFields fields;
```

Won't do indentation or anything fancy but I guess good enough to start with.

fufexan: That looks promising. I'll play around to see whether I can get it to indent properly. Thanks!

fufexan: The hyprland config allows a key to have multiple values. For example
```ini
exec-once=eww open bar
exec-once=firefox

bind=SUPER, Return, exec, kitty
bind=SUPER, Return, exec, notify-send "Kitty was started!"
```

How could we make this work?

ncfavier: That's what `listsAsDuplicateKeys` does

fufexan: That's really handy then!

fufexan: Sorry for the long delay, I've been having exams for the past month.

I just realized we have to handle submaps (nested keybinds) as a special case.
```ini
    # window resize
    bind = $mod, S, submap, resize

    submap = resize
    binde = , right, resizeactive, 10 0
    binde = , left, resizeactive, -10 0
    binde = , up, resizeactive, 0 -10
    binde = , down, resizeactive, 0 10
    bind = , escape, submap, reset
    submap = reset
```

Here, the order of binds matters. Any binds between `submap = ...`. Can we simply pass this as `key=value` (I'm thinking multiline string) or is there a better way?

fufexan: It seems the multiline string way worked just fine, although it looks a bit ugly.

fufexan:
I've tried testing this with my configuration but it does not seem to work in any way. I have the testing branch available [here](https://github.com/fufexan/dotfiles/tree/hyprland-module). If anyone wants to test building it (`nixos-rebuild build --flake github:fufexan/dotfiles/hyprland-module#io`).

Alternatively, here's the rough config file I've used for testing the `toHyprconf` function (which worked fine). [hyprlang2.txt](https://github.com/nix-community/home-manager/files/11779768/hyprlang2.txt)
I load it in a repl and write it in a file with
```nix
> :lf nixpkgs
> config = import ./hyprconf2.nix lib
> :b legacyPackages.x86_64-linux.writeText "hyprland.conf" config
```

fufexan:
We may be able to remove the `xwayland.hidpi` option (and the patching), as Hyprland has a new option that does the same thing, called `xwayland:force_zero_scaling`. Paired with toolkit-specific scaling options, it should achieve even better support for HiDPI. Currently not everything works perfectly, so I will wait for fixes before removing the `hidpi` option from this module.

modules/services/window-managers/hyprland.nix
```diff
@@ -0,0 +1,194 @@
...
+          Enable HiDPI XWayland, based on [XWayland MR 733](https://gitlab.freedesktop.org/xorg/xserver/-/merge_requests/733).
+          See <https://wiki.hyprland.org/Nix/Options-Overrides/#xwayland-hidpi> for more info.
+        '';
+      };
+    };
+
+    enableNvidiaPatches =
+      lib.mkEnableOption "patching wlroots for better Nvidia support";
+
+    settings = lib.mkOption {
```
rycee, line 63: Note, name changed to `settings` in line with RFC 42.

rycee:
Looks pretty good! I rebased and added a fixup commit with various cleanups (and added tests). If you can, then please give it a try.

I'll try to finish it all up tomorrow.

fufexan:
Thank you! I'll test this right now.

fufexan:
I got this error
```nix
       error: A definition for option `home-manager.users.mihai.wayland.windowManager.hyprland.settings.input.touchpad' is not of type `atom (null, bool, int, float or string) or a list of them for duplicate keys'. Definition values:
       - In `/nix/store/ml8sy5pirc2yi9d2kzkpw6hw9b87qn6i-source/home/wayland/hyprland/settings.nix':
           {
             scroll_factor = 0.3;
           }
```

This is the nix code that triggered it
```nix
    input = {
      kb_layout = "ro";

      # focus change on cursor move
      follow_mouse = 1;
      accel_profile = "flat";
      touchpad = {
        scroll_factor = 0.3;
      };
    };
```

rycee:
@fufexan Aha, I wasn't aware that the nesting could be more than one level deep. I've force-pushed an update now that allows arbitrary nesting. Please give it a try.

fufexan:
Thank you, works just fine now!

rycee:
Great! I've merged it now 🙂

fufexan:
Thank you for all the help, Robert!

RaySlash:
Can this module be backported to `23.05` so that we can use hyprland on nixos stable release with hm.? Thank you

PR #3838, opened 2023-04-03
maintainers: rename houstdav000 → cyntheticfox

ncfavier:
Renamed in nixpkgs https://github.com/NixOS/nixpkgs/pull/223976, so now our tests are [failing](https://github.com/nix-community/home-manager/actions/runs/4599709875/jobs/8125413653?pr=3837#step:6:7)... @rycee we need to find a way to avoid that, but I'm not sure how.

cc @cyntheticfox 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


cyntheticfox:
Ah, didnt realize home-manager would be affected. Sorry about that 😅

PR #3840, opened 2023-04-03
services.gpg-agent: darwin compatibility

elohmeier:
### Description

Added compatibility for Darwin to services.gpg-agent.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/gpg-agent.nix
+++ b/modules/services/gpg-agent.nix
@@ -256,13 +256,25 @@ in {
       '') cfg.sshKeys;
     })
 
+    (mkIf (pkgs.stdenv.isDarwin && cfg.enableSshSupport) {
+      launchd.agents.gpg-agent = {
+        enable = true;
+        config = {
+          ProgramArguments = [ "${gpgPkg}/bin/gpgconf" "--launch" "gpg-agent" ];
+          RunAtLoad = true;
+          KeepAlive.SuccessfulExit = false;
+          EnvironmentVariables.GNUPGHOME = homedir;
+        };
+      };
+    })
+
     # The systemd units below are direct translations of the
     # descriptions in the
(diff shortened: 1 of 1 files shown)


modules/services/gpg-agent.nix
```diff
@@ -256,13 +256,25 @@ in {
       '') cfg.sshKeys;
     })
 
+    (mkIf (pkgs.stdenv.isDarwin && cfg.enableSshSupport) {
```
ncfavier, line 259: Since https://github.com/nix-community/home-manager/pull/2915, there's no need for the platform conditions any more.

modules/services/gpg-agent.nix
```diff
@@ -256,13 +256,25 @@ in {
...
     # descriptions in the
     #
     #   ${gpgPkg}/share/doc/gnupg/examples/systemd-user
     #
     # directory.
-    {
+    (mkIf pkgs.stdenv.isLinux {
       assertions = [
         (hm.assertions.assertPlatform "services.gpg-agent" pkgs platforms.linux)
       ];
```
ncfavier, line 280: We should remove this assertion as well.

cmacrae:
@elohmeier Thanks for raising this! Great timing, I was just about to open a PR myself, so was delighted to see this.

Not sure how much testing you did on various systems, but with your changes, it doesn't work on macOS Ventura. There are some slight nuances with this module's original intention only being for Linux.

I have an updated version of this that is fully working, also with ssh-agent support working. I also addressed @ncfavier's comments.

If you'd be interested in these contributions, how would you like me to submit them? A PR to your branch?

elohmeier:
Closing in favor of #2964

PR #3847, opened 2023-04-04
thunderbird: added extraConfig option

VwCSXg:
### Description

The PR adds the option `extraConfig` which includes custom strings at the end of the generated `user.js`.

This is useful for importing `user.js`-templates such as [HorlogeSkynet/thunderbird-user.js](https://github.com/HorlogeSkynet/thunderbird-user.js).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


d-dervishi:
Looks good! If the intended use case is to easily import external templates, then we should make sure that the extra config does not override locally defined preferences. Do you know whether Thunderbird gives priority to the first or last occurrence of a key?

VwCSXg:
Thanks!

Thunderbird gives preference to the last occurance which means that you need to overwrite templates within the extraConfig option with the suggested implementation. I chose to implement it this way because its analogous to how the [firefox module](https://github.com/nix-community/home-manager/blob/master/modules/programs/firefox.nix#L59) handles it, but I see your point and either way works for me.

rycee:
Thanks! Merged to master now.

I think keeping it on the bottom should be good enough for now. At least, I haven't heard complaints about this position in the Firefox module.

PR #3851, opened 2023-04-06
home-cursor: follow xdg spec for icons folder

GaetanLepage:
### Description

Currently, the `home.pointerCursor` module creates a file in the `~/.icons/` folder.
~~According to the [XDG spec for _Icon Theme Specification_](https://www.freedesktop.org/wiki/Specifications/icon-theme-spec/), it seems that [this folder can be located in one of the `$XDG_DATA_DIRS` folders](https://specifications.freedesktop.org/icon-theme-spec/latest/ar01s03.html).~~
This PR moves the `icons/` folder to this location.

EDIT:  According to [the Arch wiki](https://wiki.archlinux.org/title/Cursor_themes#Manually) and my testing, it seems to work better when using `$XDG_DATA_HOME/icons/`.
Maybe it is better to put it in both places.
For information, the implementation of `$XDG_DATA_DIRS` is the following:
```nix
xdg.systemDirs.data = [
  (toString (pkgs.writeTextDir "icons/default/index.theme" ''
    [icon theme]
    Name=Default
    Comment=Default Cursor Theme
    Inherits=${cfg.name}
  ''))
];
```

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/config/home-cursor.nix
+++ b/modules/config/home-cursor.nix
@@ -119,7 +119,7 @@ in {
       # Set name in icons theme, for compatibility with AwesomeWM etc. See:
       # https://github.com/nix-community/home-manager/issues/2081
       # https://wiki.archlinux.org/title/Cursor_themes#XDG_specification
-      home.file.".icons/default/index.theme".text = ''
+      xdg.dataFile."icons/default/index.theme".text = ''
         [icon theme]
         Name=Default
         Comment=Default Cursor Theme


GaetanLepage:
cc @ncfavier @rycee

ncfavier (approved):
Haven't tested but from a quick look at the source code it looks like awesome respects this. cc @SeraphyBR

GaetanLepage:
> Haven't tested but from a quick look at the source code it looks like awesome respects this. cc @SeraphyBR

Ok thank you for the review !
What could I do for this to get merged ? Is more testing needed ?

PR #3852, opened 2023-04-07
ssh: made the ssh config file path modular

Neon-44:
made a new option programs.ssh.userConfigFile with default ".ssh/config" and replaced home.file.".ssh/config".text with home.file.${cfg.userConfigFile}.text

### Description

adds an Option programs.ssh.userConfigFile which expects a String returned (default .ssh/config).
replaces home.file.".ssh/config".txt with home.file.${cfg.userConfigFile}.txt
can be used together with an Alias (ssh = ssh -F /path/to/file) to declutter the home directory
default option for userConfigFile is still ".ssh/config"

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

this PR doesn't add a new Module.

### DIFF
--- a/modules/programs/ssh.nix
+++ b/modules/programs/ssh.nix
@@ -392,6 +392,21 @@ in
       '';
     };
 
+    userConfigFile = mkOption {
+      type = types.str;
+      default = ".ssh/config";
+      description = ''
+      tells Home-Manager where in your home you want your ssh config to be.
+      Example: <filename>.ssh/config<filename> will put it into 
+      <filename>~/.ssh/config<filename>
+      while <filename>.config/ssh/config<filename> will put it into 
+      <filename>~/.config/ssh/config<filename>
+      file is put into <filename>~/.ssh/config</filename> by default.
+      can be paired with a alias 'ssh = "ssh -F"' to declutter your 
+      home folder.
+      '';
+    };
+
     userKnownHostsFile = mkOption {
       type = types.str;
       default = "~/.ssh/known_hosts";
@@ -510,7 +525,7 @@ in
       }
     ];
 
-    home.file.".ssh/config".text =
+    home.file.${cfg.userConfigFile}.text =
       let
         sortedMatchBlocks = hm.dag.topoSort cfg.matchBlocks;
         sortedMatchBlocksStr = builtins.toJSON sortedMatchBlocks;


Nehctarg:
Shouldn't the generation of the alias be another option included with this?

Nehctarg:
Okay, I think that can be put aside for another pr. Having this here merged would be huge

poelzi (approved):

Aleksanaa:
Still important

modules/programs/ssh.nix
```diff
@@ -392,6 +392,21 @@ in
       '';
     };
 
+    userConfigFile = mkOption {
```
khaneliman, line 395:
```suggestion
    configDirectory = mkOption {
```

modules/programs/ssh.nix
```diff
@@ -392,6 +392,21 @@ in
...
+      default = ".ssh/config";
+      description = ''
+      tells Home-Manager where in your home you want your ssh config to be.
+      Example: <filename>.ssh/config<filename> will put it into 
+      <filename>~/.ssh/config<filename>
+      while <filename>.config/ssh/config<filename> will put it into 
+      <filename>~/.config/ssh/config<filename>
+      file is put into <filename>~/.ssh/config</filename> by default.
+      can be paired with a alias 'ssh = "ssh -F"' to declutter your 
+      home folder.
```
khaneliman, line 406:
```suggestion
        Set ssh config directory.

        Example: `.ssh/config` will put it into `~/.ssh/config`
        while `.config/ssh/config` will put it into `~/.config/ssh/config`.

        File is put into `~/.ssh/config` by default.

        Can be paired with an alias `ssh = "ssh -F"` to declutter your
        home folder.
```

modules/programs/ssh.nix
```diff
@@ -510,7 +525,7 @@ in
       }
     ];
 
-    home.file.".ssh/config".text =
+    home.file.${cfg.userConfigFile}.text =
```
khaneliman, line 528:
```suggestion
    home.file.${cfg.configDirectory}.text =
```

teto:
rather than adding yet another config item, you could disable the writing of the file (e.g. `  xdg.configFile."nix/nix.conf".enable = false;`) and write its output somewhere else. I am not absolutely sure you can copy the content of the file but worth a try, it should still be available even when disabled.
`

Neon-44:
the wording @khaneliman suggested is a lot better, let's go with that instead ^^

PR #3855, opened 2023-04-08
rcm: add module

SpringerJack:
### Description

This module provides basic functionality to easily bridge the gap between declarative and ad-hoc home configuration for users of [rcm](https://github.com/thoughtbot/rcm). 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


SpringerJack:
By the way, my current implementation of this is a bit quirky. All settings in the `.rcrc` are uppercase which I hate to type, so I change all lowercase letters to uppercase. This is probably not what a user would first expect and I am willing to remove this if need be, however I consider it to be convenient and in no way harmful. Opinions welcome.

modules/programs/rcm.nix
```diff
@@ -0,0 +1,36 @@
+{ config, pkgs, user, lib, ... }:
+
+with lib;
+with lib.strings;
+let
+  cfg = config.programs.rcm;
+
+  stringifyList = l:
+    if isList l then ''"${concatStringsSep " " l}"'' else stringifyList [ l ];
```
ambroisie, line 9: `toString <list>` already does that.

ambroisie: Ah sorry, I had missed the `"` wrapping.

I think at that point you could just inline it into `mkKeyValue`.

modules/programs/rcm.nix
```diff
@@ -0,0 +1,36 @@
...
+          Submodule defining the .rcrc file. Mapping of the key value pairs for rcms config.
+        '';
+      };
+    };
+
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = with pkgs; [ rcm ];
+    home.file.".rcrc" = { text = toConfigFile cfg.settings; };
```
ambroisie, line 32: Arguably could be set to an arbitrary location with the `RCRC`  environment variable (which is done in other modules, e.g: GnuPG).

ambroisie:
For my own curiosity, what is your use-case for this module that can't be fulfilled by home-manager itself?

modules/programs/rcm.nix
SpringerJack: Ha, never used that but thats a great idea. Thanks for the hint.

SpringerJack:
I don't think it does anything home-manager in itself can't do, it just makes some things more *comfortable* - at least for me. Parts of my configuration (e.g. my emacs setup) may change frequently and I never liked 
a) having to rebuild for every change
or b) creating writable symlinks to my collection of configs by hand. 

Also rcm works on systems without home-manager.

modules/programs/rcm.nix
SpringerJack: I'd rather inline only the branching part for readability and less repeated quote confusion.

modules/programs/rcm.nix
```diff
@@ -5,17 +5,24 @@ with lib.strings;
...
   options = {
     programs.rcm = {
       enable = mkEnableOption "rcm";
+      rcrc = mkOption {
+        type = types.path;
+        default = "${config.home.homeDirectory}.rcrc";
+        description = lib.mdDoc ''
+          Location of the RCRC file.
+        '';
+        example = "${home.homeDirectory}/.config/rcm/rcrc";
```
ambroisie, line 24: Escape the example value, your documentation should not depend on the configuration.

modules/programs/rcm.nix
```diff
@@ -5,17 +5,24 @@ with lib.strings;
...
+      "${toUpper k}=${if isList v then quotedList v else quotedList [ v ]}";
     listsAsDuplicateKeys = false;
   };
 in {
   options = {
     programs.rcm = {
       enable = mkEnableOption "rcm";
+      rcrc = mkOption {
+        type = types.path;
+        default = "${config.home.homeDirectory}.rcrc";
```
ambroisie, line 20:
```suggestion
        default = "${config.home.homeDirectory}/.rcrc";
        defaultText = ''"''${config.home.homeDirectory}/.rcrc"'';
```

ambroisie (changes requested):

ambroisie:
Also, you should rebase your changes to have two commits, in order:

* adding yourself to the list of maintainers.
* adding the module.

modules/programs/rcm.nix
ambroisie:
```suggestion
        example = literalExpression ''"''${home.homeDirectory}/.config/rcm/rcrc"'';
```

ambroisie:
I just noticed that your `rcrc` option documents itself as accepting an absolute path, when it in fact expects a path relative to `$HOME`.

So remove the `${config.home.homeDirectory}` from the options, and disregard my `literalExpression` suggestions.

modules/programs/rcm.nix
```diff
@@ -0,0 +1,44 @@
...
+        '';
+      };
+    };
+  };
+  config = mkIf cfg.enable {
+    home.packages = with pkgs; [ rcm ];
+    home.file.${cfg.rcrc} = { text = toConfigFile cfg.settings; };
+    home.sessionVariables = {
+      RCRC = "${config.home.homeDirectory}/${cfg.rcrc}";
+    };
```
ambroisie, line 40: Add `options` to module argument list and apply the following:

```suggestion
    home.sessionVariables = lib.mkIf cfg.rcrc != options.programs.rcm.rcrc.default {
      RCRC = "${config.home.homeDirectory}/${cfg.rcrc}";
    };
```

That way we avoid overriding `RCRC` if there is no need for it.

modules/programs/rcm.nix
```diff
@@ -0,0 +1,44 @@
...
+  };
+in {
+  options = {
+    programs.rcm = {
+      enable = mkEnableOption "rcm";
+      rcrc = mkOption {
+        type = types.path;
+        default = ".rcrc";
+        description = lib.mdDoc ''
+          Location of the RCRC file.
```
ambroisie, line 22: It should mention that the path is relative to `$HOME`.

ambroisie:
Sorry I missed those points in my previous review. Should be fine after that.

ncfavier:
> b) creating writable symlinks to my collection of configs by hand

You can do that declaratively with [`config.lib.file.mkOutOfStoreSymlink`](https://github.com/nix-community/home-manager/blob/53bd74f786934997e7f6a5ed9741b226e511e508/modules/files.nix#L64-L69).

PR #3856, opened 2023-04-08
firefox: Support paths for userChrome & userContent

amarshall:
### Description

A path may be preferred for some uses, and allowing it avoids the user needing to `builtins.readFile`, thus creating duplicates and making it more difficult to determine the actual store path.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/firefox/mkFirefoxModule.nix
+++ b/modules/programs/firefox/mkFirefoxModule.nix
@@ -401,7 +401,7 @@ in {
           };
 
           userChrome = mkOption {
-            type = types.lines;
+            type = types.oneOf [ types.lines types.path ];
             default = "";
             description = "Custom ${appName} user chrome CSS.";
             example = ''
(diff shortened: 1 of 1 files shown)


kira-bruneau (approved):
Oh I missed this before. I don't have commit access, but this change looks good to me!

nyabinary:
Bump does this still work with current firefox?

kira-bruneau:
I'll test this again, but from just the diff I don't think there's anything wrong.

CI was only failing on darwin because of some upstream issues in nixpkgs.

kira-bruneau (approved):
I just tested this commit ontop of bd65bc3cde04c16755955630b344bc9e35272c56 (the `HEAD` of `release-24.11`), and everything looks good to me!

kira-bruneau:
Oh wait actually, I think it only worked because I have this set in my settings:

```nix
toolkit.legacyUserProfileCustomizations.stylesheets = true
```

kira-bruneau:
@teto Oh this should be good to merge now!

PR #3861, opened 2023-04-09
home-manager: migrate profiles to Nix state directory

rycee:
### Description

If the user runs a recent Nix version that places per-user profiles in `$XDG_STATE_DIR/nix/profiles`, then migrate the home-manager profile there.

Also clean up `setupVars` a bit.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/home-manager/home-manager
+++ b/home-manager/home-manager
@@ -117,18 +117,25 @@ function setHomeManagerNixPath() {
 
 # Sets some useful Home Manager related paths as global read-only variables.
 function setHomeManagerPathVariables() {
-    declare -r nixStateDir="${NIX_STATE_DIR:-/nix/var/nix}"
-    declare -r globalProfilesDir="$nixStateDir/profiles/per-user/$USER"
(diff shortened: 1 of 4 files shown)


inclyc:
Should we update / check the default value here according to these changes?

https://github.com/nix-community/home-manager/blob/28698126bd825aff21cae9ffd15cf83e169051b0/modules/home-environment.nix#L209

inclyc:
> The link target changes in newer Nix versions, however. 

Recently I encountered an error while having a fresh-installed NixOS. There is no such symlinks at all. Nix put all files into $XDG_STATE_HOME.  

```
/home/lyc/.config/zsh/.zshenv:.:2: no such file or directory: /home/lyc/.nix-profile/etc/profile.d/hm-session-vars.sh
```

Looks like #3876 is addressing this issue, but not yet merged. I currently ad-hoc my environment by create the symbolic links manually:

```
ln -s ~/.local/state/nix/profile ~/.nix-profile
```

```
nix --version
nix (Nix) 2.15.0
```

PR #3872, opened 2023-04-13
activation-init.sh: create the Nix profiles path before accessing it

midchildan:
### Description

The Nix profiles path may not exist right after installing Nix. In that case, it is created on demand by the Nix CLI tools. But after #3861, Home Manager assumes it exists and fails if it doesn't. This caused problems when testing Home Manager configurations on GitHub Actions and cachix/install-nix-action was used to install Nix.

This change makes sure to trigger the creation of the Nix profiles path before attempting to access it.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


rycee:
Thanks, good catch! Merged to master now 🙂

PR #3875, opened 2023-04-14
lazygit: use xdg.configHome on darwin too if xdg.enable

### LINKED ISSUE #3634: Change location of lazygit's config on Darwin
## The issue

Hi,
Lazygit in the in the [configuration docs][1] says that, by default on Darwin, it will look for a configuration file under `~/Library/Application Support/lazygit`. The problem is that, at least on MacOS Ventura (Darwin 22.3), it looks instead under `~/.config/lazygit`.

The way to check this is by running lazygit with the flag `cd`. Which prints the default config directory.
```bash
$ uname -rom
22.3.0 x86_64 Darwin

$ lazygit -cd
/Users/sacca/.config/lazygit
```

## Possible Solutions

_Referring to_ `modules/programs/lazygit.nix`

1. Removing check in `!Darwin`
2. Removing the hole `home.file` expression

[1]: https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md

utkarshgupta137:
### Description

Use `xdg.configHome` if `xdg.enable` is true, like most other tools on darwin.
Lazygit supports XDG path if XDG_CONFIG_HOME is defined: https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md

Fixes #3634.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/lazygit.nix
+++ b/modules/programs/lazygit.nix
@@ -34,8 +34,10 @@ in {
       '';
       description = ''
         Configuration written to
-        <filename>$XDG_CONFIG_HOME/lazygit/config.yml</filename> on Linux
-        or <filename>~/Library/Application Support/lazygit/config.yml</filename> on Darwin. See
+        <filename>$XDG_CONFIG_HOME/lazygit/config.yml</filename>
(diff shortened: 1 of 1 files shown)


rycee:

PR #3876, opened 2023-04-14
home: fix profileDirectory if use-xdg-base-directories

utkarshgupta137:
### Description

When `nix.settings.use-xdg-base-directories` is true, the `home.profileDirectory` should be updated so that the environment is sourced correctly.
Related: https://github.com/NixOS/nix/pull/5588.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/home-environment.nix
+++ b/modules/home-environment.nix
@@ -508,6 +508,8 @@ in
       if config.submoduleSupport.enable
         && config.submoduleSupport.externalPackageInstall
       then "/etc/profiles/per-user/${cfg.username}"
+      else if config.nix.enable && (config.nix.settings.use-xdg-base-directories or false)
+      then "${config.xdg.stateHome}/nix/profile"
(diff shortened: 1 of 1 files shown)


modules/home-environment.nix
```diff
@@ -508,6 +508,8 @@ in
       if config.submoduleSupport.enable
         && config.submoduleSupport.externalPackageInstall
       then "/etc/profiles/per-user/${cfg.username}"
+      else if config.nix.settings.use-xdg-base-directories
```
rycee, line 511: I think you'll need something like

```suggestion
      else if config.nix.enable && (config.nix.settings.use-xdg-base-directories or false)
```

utkarshgupta137: Done

utkarshgupta137:
@rycee Should be good to merge now?

rycee:
Thanks! Merged to master now.

PR #3894, opened 2023-04-19
mako: add systemd integration

### LINKED ISSUE #3893: Request: add Mako systemd user service option
### Description
Currently there is no systemd user service defined in `modules/services/mako.nix`. This is strange because service modules for long running processes usually include a systemd service option (see waybar, swayidle, wlsunset, etc) but the only thing in the Mako module currently that starts Mako is the `xdg.configFile."mako/config".onChange` trigger  via `${cfg.package}/bin/makoctl reload || true`.

This config file trigger is not a good solution because it is in no way bound to a Wayland session and users may want to change their Mako config without starting Mako (i.e. while inside an Xorg session). The omission of a systemd service option has already confused some users (#2028).

A systemd user service option should be added, just like wlsunset for example.

stelcodes:
### Description
```
    This change adds a systemd user service to the mako module and enables
    it by default. It also removes the xdg.configFile."mako/config".onChange
    trigger which was the sole method of starting mako from the module
    previously.
```

Inspiration mostly came from the wlsunset module.

### Checklist

Not completely. `services.mako.enable = true` now enables a systemd service which starts mako when previously it did not start mako at all. Also,  the config file onChange trigger is removed.

    ```
    {component}: {description}

    {long description}
    ```

Fixes #3893

### DIFF
--- a/modules/services/mako.nix
+++ b/modules/services/mako.nix
@@ -25,6 +25,20 @@ in {
         description = "The mako package to use.";
       };
 
+      systemd.target = mkOption {
+        type = types.str;
+        default = "graphical-session.target";
+        example = "sway-session.target";
+        description = ''
+          The systemd target that will automatically start the Mako service.
+          </para>
+          <para>
+          When setting this value to <literal>"sway-session.target"</literal>,
+          make sure to also enable <option>wayland.windowManager.sway.systemdIntegration</option>,
+          otherwise the service may never be started.
+        '';
+      };
+
       maxVisible = mkOption {
         default = 5;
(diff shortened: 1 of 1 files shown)


stelcodes:
Don't merge, need to fix systemd service definition to be more like `fnott`.

ncfavier:
We'll also want to make the systemd integration optional to avoid conflicting with [other ways](https://github.com/emersion/mako#running) of starting the program (if a systemd service is needed at all).

stelcodes:
I'm not confident this PR is necessary, closing for now.

PR #3899, opened 2023-04-20
fnott: add DBus service file

christoph-heiss:
### Description

This is needed to make fnott bus-activatable. Otherwise it won't be started when a bus request is made [0].
Done in the same way as e.g. dunst sets that up [1,2].

Fixes notifications through fnott on my machine.
Tested using a simple `nix-shell -p libnotify --run 'notify-send test'`.

[0] https://www.freedesktop.org/software/systemd/man/systemd.service.html#id-1.10.6
[1] https://github.com/nix-community/home-manager/blob/master/modules/services/dunst.nix#L139
[2] https://github.com/dunst-project/dunst/blob/master/org.knopwob.dunst.service.in

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/fnott.nix
+++ b/modules/services/fnott.nix
@@ -107,6 +107,13 @@ in {
       };
     };
 
+    xdg.dataFile."dbus-1/services/fnott.service".text = ''
+      [D-BUS Service]
+      Name=org.freedesktop.Notifications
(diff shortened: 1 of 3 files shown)


rycee:

PR #3904, opened 2023-04-22
swappy: add module

eclairevoyant:
### Description

Adds HM module for swappy, a simple GTK-based screenshot editor for Wayland.

See [project homepage](https://github.com/jtheoof/swappy) and [nixpkgs derivation](https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/misc/swappy/default.nix).

Empty config is used by default to allow the application to use its internal defaults.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -1143,6 +1143,13 @@ in
           A new module is available: 'services.ssh-agent'
         '';
       }
+
+      {
+        time = "2023-07-07T06:00:00+00:00";
+        message = ''
+          A new module is available: 'programs.swappy'.
+        '';
+      }
     ];
   };
 }
--- a/modules/programs/swappy.nix
+++ b/modules/programs/swappy.nix
@@ -1,0 +1,47 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.programs.swappy;
+
+  iniFormat = pkgs.formats.ini { };
+
+  iniFile = iniFormat.generate "config" cfg.settings;
+
+in {
+  meta.maintainers = [ hm.maintainers.eclairevoyant ];
+
+  options.programs.swappy = {
+    enable = mkEnableOption "Swappy, a GTK-based screenshot editor for Wayland";
+
+    package = mkPackageOption pkgs "swappy" { };
+
+    settings = mkOption {
+      type = iniFormat.type;
+      default = { };
+      example = ''
+        {
+          Default = {
+            show_panel = true;
+          };
+        }'';
+      description = ''
+        Configuration to use for Swappy. See
+        <citerefentry>
+          <refentrytitle>swappy</refentrytitle>
+          <manvolnum>1</manvolnum>
+        </citerefentry>
+        for available options.
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    xdg.configFile =
+      mkIf (cfg.settings != { }) { "swappy/config".source = iniFile; };
+  };
+}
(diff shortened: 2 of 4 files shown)


teto:
what is swappy ? helps reviewer not to have to look for it

tomodachi94 (approved):
Lgtm.

tomodachi94:
To speed up merging a bit, maybe add some tests? See <https://github.com/nix-community/home-manager/tree/master/tests/modules/programs/vim-vint> for an example of how I did this a few months ago.

eclairevoyant:
@tomodachi94 In this case I originally didn't add tests because it used the INI formatter (I assume if something broke in that module then we would have bigger problems), but if you still think tests would be helpful I can add in.

teto:
I've just checked out and enabled the module but starting it (nixos-unstable) gives:
```
$ swappy       

** (swappy:97788): CRITICAL **: 17:15:09.540: original image not loaded
```
any idea ? I am running sway with nvidia drivers

eclairevoyant:
> I've just checked out and enabled the module but starting it (nixos-unstable) gives:
> 
> ```
> $ swappy       
> 
> ** (swappy:97788): CRITICAL **: 17:15:09.540: original image not loaded
> ```

* any error like that wouldn't be related to this HM module, and
* you will probably want to read the manpages first, see https://github.com/jtheoof/swappy/issues/137

modules/programs/swappy.nix
```diff
@@ -0,0 +1,64 @@
...
+          save_filename_format = "swappy-%Y%m%d-%H%M%S.png";
+          show_panel = false;
+          line_size = 5;
+          text_size = 20;
+          text_font = "sans-serif";
+          paint_mode = "brush";
+          early_exit = false;
+          fill_shape = false;
+        };
+      };
```
rycee, line 40: Is there a strong reason why the default is not `{ }`? I think it would be best to have it empty to use the true Swappy defaults. For example, hopefully the true default would handle `save_dir = "$HOME/Desktop"` better for people who don't actually have that directory.

rycee: Yeah, I had a look and this does seem to be the case. See https://github.com/jtheoof/swappy/blob/596b9a8b17234303a9d4d2ea1d3107cb5442253e/src/config.c#L27

So I would strongly suggest trimming the default value to

```suggestion
      default = { };
```

modules/programs/swappy.nix
```diff
@@ -0,0 +1,64 @@
...
+
+  options.programs.swappy = {
+    enable = mkEnableOption "Swappy";
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.swappy;
+      defaultText = literalExpression "pkgs.swappy";
+      description = "Package providing <command>swappy</command>.";
+    };
```
rycee, line 24: Can simplify to
```suggestion
    package = mkPackageOption pkgs "swappy { };
```

modules/misc/news.nix
```diff
@@ -1133,8 +1133,17 @@ in
         condition = config.programs.zsh.enable;
         message = ''
           A new modules is available: 'programs.zsh.antidote'
+
```
rycee, line 1136:
```suggestion
```

modules/misc/news.nix
```diff
@@ -1133,8 +1133,17 @@ in
...
         '';
       }
+
+      {
+        time = "2023-06-24T19:31:35+00:00";
+        message = ''
+          A new module is available: 'programs.swappy'.
+        '';
+      }
+
```
rycee, line 1146:
```suggestion
```

rycee:

eclairevoyant:
Thanks for the feedback; PR updated.

modules/programs/swappy.nix
eclairevoyant: Thanks, I made an incorrect assumption about the behavior of the app with no config provided. I tested with no config and it works as expected.

PR #3906, opened 2023-04-22
neovim: Enable tests on Darwin

YorikSar:
### Description

Checked on Apple Silicon laptop with both `aarch64-darwin` and `x86-64-darwin`. All tests succeeded on both.

### Checklist

- [ ] ~Code formatted with `./format`.~ `tests/default.nix` is not formatted with `nixfmt`, but I didn't mess with formatting.

- [ ] ~Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).~ N/A

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/tests/default.nix
+++ b/tests/default.nix
@@ -96,6 +96,7 @@ import nmt {
     ./modules/programs/ncmpcpp
     ./modules/programs/ne
     ./modules/programs/neomutt
+    ./modules/programs/neovim
     ./modules/programs/newsboat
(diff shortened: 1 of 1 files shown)


rycee:
Thanks, very nice! Merged to master now.

PR #3907, opened 2023-04-22
neovim: Add nvim-treesitter configuration

YorikSar:
### Description

Allows to use only necessary grammars and compose their list acrossmodules.

I had to implement this for my dotflies where I configure different languages in different modules, for example: 
https://github.com/YorikSar/dotfiles/commit/c591a5aa7d47ddce5616316298af1ff8b508bcf1

Hopefully this would save others this headache.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/neovim.nix
+++ b/modules/programs/neovim.nix
@@ -59,6 +59,27 @@ let
     plugin = cfg.coc.package;
     config = cfg.coc.pluginConfig;
     optional = false;
+  } ++ optional cfg.treesitter.enable {
+    plugin = cfg.treesitter.package.withPlugins
+      config.programs.neovim.treesitter.grammars;
+    type = "lua";
+    config = ''
+      require('nvim-treesitter.configs').setup {
+        -- auto_install is not reproducible
+        auto_install = false,
+        highlight = {
+          enable = ${
+            if cfg.treesitter.highlight.enable then "true" else "false"
+          },
+          disable = { ${
+            concatStringsSep ", "
+            (map (s: ''"${s}"'') cfg.treesitter.highlight.disableForGrammars)
+          } },
+          -- docs claim that this slows down the editor and mostly not needed
+          additional_vim_regex_highlighting = false,
+        };
+      }
+    '';
   };
 
   luaPackages = cfg.finalPackage.unwrapped.lua.pkgs;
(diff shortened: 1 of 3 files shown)


YorikSar:
Created a draft, because it's stacked on top of #3906 and #3906, will rebase when those are merged. Other than that it's ready for merge.

YorikSar:
In fact, why wait. Removed those commits from this branch.

teto:
I would like to keep the complexity of the module low. What you propose is already possible with:
```
    programs.neovim.plugins = [ {
      plugin = (nvim-treesitter.withPlugins (
        plugins: with plugins; [
          tree-sitter-query
        ]
      ))};
   ];
```

also nvim-treesitter is a moving target likely to become obsolete. Maybe you can propose this to https://github.com/pta2002/nixvim ?

YorikSar:
> I would like to keep the complexity of the module low.

I don't think this addition greatly increases complexity: it's basically 1 "main" option + couple of supporting ones.

> What you propose is already possible with:
> 
> ```
>     programs.neovim.plugins = [ {
>       plugin = (nvim-treesitter.withPlugins (
>         plugins: with plugins; [
>           tree-sitter-query
>         ]
>       ))};
>    ];
> ```

Unfortunately this is not compostable/extendable. The main point here is too allow to spread list of plugins over a number of modules.

> also nvim-treesitter is a moving target likely to become obsolete.

Could you please expand?

> Maybe you can propose this to https://github.com/pta2002/nixvim ?

I didn't know about that project, but isn't it tangential to home-manager? I'd prefer to stick to home-manager if possible.

teto:
> Could you please expand?

nvim-treesitter was a way to experiment with treesitter without polluting the main neovim repository. nvim-treesitter utilities are bit by bit upstreamed in neovim core until it just becomes a list of grammars (like lspconfig). 
On nixos, the list of grammars is not needed since all grammars are packaged already (or trivial to add).

Plugins can specify grammar as dependencies
 https://github.com/NixOS/nixpkgs/blob/103897b4fcd1af998d70f0af599b7ae8ccdd1082/pkgs/applications/editors/vim/plugins/overrides.nix#L775
 and you can have several such plugins so it is composable otherwise it's a nixpkgs bug.
 
If you want to install grammars manually it makes more sense to add nvim-treesitter with the grammars as a plugin than adding a new unneeded way to do this that needs to be maintained on top of that.
 
Closing.

YorikSar:
@teto Do you mean that you can specify multiple nvim-treesitter instances in the "plugins" list? So, for example, if I specify:

```
programs.neovim.plugins = with pkgs.vimPlugins; [
  (nvim-treesitter.withPlugins (p: [p.go]))
  (nvim-treesitter.withPlugins (p: [p.rust]))
]
```

it should enable both grammars just fine without conflict?

If that's the case, then indeed my usecase is covered and options in this PR are not needed.

teto:
yes it should be the case if not that's a bug and let me know

YorikSar:
> yes it should be the case if not that's a bug and let me know

Great! I've rewrote my config replacing usage of these options with multiple instances of plugin here: https://github.com/YorikSar/dotfiles/commit/5a80f5283e9c23e75b393b0eed79baa3b49fe663 - and it works just fine.
Thank you very much for showing me the right way!

teto:
maybe the nixpkgs doc needs updating if that was not clear ! Thanks for trying to improve home-manager ;) hopefully next PR will get merged

PR #3911, opened 2023-04-24
helix: Support `[[grammar]]` in languages.toml

### LINKED ISSUE #2871: Helix: support configuring grammars
### Support configuring tree-sitter grammars in `languages.toml` for Helix.

Docs: https://docs.helix-editor.com/languages.html#tree-sitter-grammars

```toml
[[grammar]]
name = "rust"
source = { git = "https://github.com/tree-sitter/tree-sitter-rust", rev = "a250c4582510ff34767ec3b7dcdd3c24e8c8aa68" }
```

The `use-grammars` key should also be supported, which controls which grammars are fetched and built (if omitted all grammars are used).

```toml
use-grammars = { only = [ "rust", "c", "cpp" ] }
# or
use-grammars = { except = [ "yaml", "json" ] }
```

matthewess:
### Description

Add support for the `[[grammar]]` key of `languages.toml`: https://docs.helix-editor.com/languages.html#tree-sitter-grammar-configuration

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

edit: Should fix #2871

### DIFF
--- a/modules/programs/helix.nix
+++ b/modules/programs/helix.nix
@@ -57,6 +57,43 @@ in {
       '';
     };
 
+    grammars = mkOption {
+      type = types.listOf tomlFormat.type;
+      default = [ ];
+      example = [{
+        name = "lalrpop";
+        source = {
+          git = "https://github.com/traxys/tree-sitter-lalrpop";
+          rev = "7744b56f03ac1e5643fad23c9dd90837fe97291e";
+        };
+      }];
+      description = ''
+        Language specific tree-sitter grammars at
+        <filename>$XDG_CONFIG_HOME/helix/languages.toml</filename>.
+        </para><para>
+        See <link xlink:href="https://docs.helix-editor.com/languages.html#tree-sitter-grammar-configuration" />
+        for more information.
+      '';
+    };
+
+    use-grammars = mkOption {
+      type = tomlFormat.type;
+      default = { };
+      example = literalExpression ''
+        {
+          only = [ "rust" "c" "cpp" ];
+        }
+      '';
+      description = ''
+        Controls which grammars are fetched and built when using
+        <literal>hx --grammar fetch</literal> and
+        <literal>hx --grammar build</literal>.
+        </para><para>
+        See <link xlink:href="https://docs.helix-editor.com/languages.html#choosing-grammars" />
+        for more information.
+      '';
+    };
+
     themes = mkOption {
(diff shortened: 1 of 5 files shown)


tests/modules/programs/helix/languages-expected-use-grammars.toml
```diff
@@ -0,0 +1,12 @@
+[use-grammars]
+except = ["yaml", "json"]
+[[grammar]]
```
matthewess, line 3: i dont love that using `concatTextFile` gives us this very slightly ugly result without an additional break between sections, unsure if there is a workaround

matthewess:
cc @Philipp-M in case you have any thoughts

modules/programs/helix.nix
```diff
@@ -136,10 +173,29 @@ in {
...
-          source =
-            tomlFormat.generate "helix-config" { language = cfg.languages; };
-        };
+        "helix/languages.toml" = mkIf (cfg.languages != [ ] || cfg.grammars
+          != [ ] || cfg.use-grammars != { }) {
+            source =
+              # NB: helix requires that `use-grammars` be the first key in languages.toml if present
+              # pkgs.formats.toml relies on piping generated json through
+              # [remarshal](https://github.com/remarshal-project/remarshal),
+              # so we have to concatenate in order to preserve this ordering
```
Philipp-M, line 182: Are you sure about that? AFAIK helix uses the `serde_derive` implementation for it which should respect the TOML standard, no weird TOML breaking custom deserializer (in the way that it relies on ordered tables/objects).

Philipp-M:
Sorry almost forgot about it.

I'm not sure about the PR (see comment in code), I think it's better to just serialize the whole languages.toml.

I have implemented this some time back in https://github.com/Philipp-M/home-manager/commit/8bdd7cc552acb2f78391d82cccdfc0b306fbc623.
Actually I have planned to upstream this when a new helix release is out that has merged this: https://github.com/helix-editor/helix/pull/2507

But it doesn't solve the grammars issue see the discussion in #2871

modules/programs/helix.nix
ncfavier: The docs say
> `# Note: this key must come **before** the [[language]] and [[grammar]] sections`
but indeed this just means that it should be *outside* of the `language` and `grammar` sections, so just regular TOML.

Issue #3858, opened 2023-04-09
bug: Fish shell error /etc/profiles/per-user/mudrii/bin/command-not-found


mudrii:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Getting an error post install if using fish as primary on nix-darwin
Using fish as shell and if typing command that doesn't exist getting errors below

```sh
/etc/profiles/per-user/mudrii/bin/command-not-found: line 3: use: command not found
/etc/profiles/per-user/mudrii/bin/command-not-found: line 4: use: command not found
/etc/profiles/per-user/mudrii/bin/command-not-found: line 5: use: command not found
/etc/profiles/per-user/mudrii/bin/command-not-found: line 6: use: command not found
/etc/profiles/per-user/mudrii/bin/command-not-found: line 7: use: command not found
/etc/profiles/per-user/mudrii/bin/command-not-found: line 9: my: command not found
/etc/profiles/per-user/mudrii/bin/command-not-found: line 11: my: command not found
/etc/profiles/per-user/mudrii/bin/command-not-found: line 13: syntax error near unexpected token `('
/etc/profiles/per-user/mudrii/bin/command-not-found: line 13: `my $dbh = DBI->connect("dbi:SQLite:dbname=$dbPath", "", "")'
```

```sh
la /etc/profiles/per-user/mudrii/bin/command-not-found
Permissions Size User Group Date Modified Name
lrwxr-xr-x    83 root wheel  1 Jan  1970   /etc/profiles/per-user/mudrii/bin/command-not-found -> /nix/store/7b864y0pncqy8wqplnwnirpfcygy67sn-home-manager-path/bin/command-not-found
```

```sh
cat /nix/store/7b864y0pncqy8wqplnwnirpfcygy67sn-home-manager-path/bin/command-not-found                                                              2818ms
#! /nix/store/2ivbs4wdlv799hmdhlwq4vpbdhr7xxn7-perl-5.36.0-env/bin/perl -w

use strict;
use DBI;
use DBD::SQLite;
use String::ShellQuote;
use Config;

my $program = $ARGV[0];

my $dbPath = "/nix/var/nix/profiles/per-user/root/channels/nixos/programs.sqlite";

my $dbh = DBI->connect("dbi:SQLite:dbname=$dbPath", "", "")
    or die "cannot open database `$dbPath'";
[...]

PR #3914, opened 2023-04-24
command-not-found: make script portable

ryuheechul:
Without `/usr/bin/env -S`, this breaks with bash on Darwin

### Description

I discovered that shebang being a pure script (not binary) seems to be problematic with

[bashInteractive package](https://search.nixos.org/packages?channel=unstable&show=bashInteractive) on Darwin (I imagine the same for 
[bash](https://search.nixos.org/packages?channel=unstable&show=bash))

But works with [zsh](https://search.nixos.org/packages?channel=unstable&show=zsh)

So I found and tested the solution that works for both bash and zsh on Darwin as well as NixOS.

### Checklist

	- not sure if any updates on test cases are required for this change

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/command-not-found/command-not-found.pl
+++ b/modules/programs/command-not-found/command-not-found.pl
@@ -1,4 +1,4 @@
-#! @perl@/bin/perl -w
+#!/usr/bin/env -S @perl@/bin/perl -w
 
 use strict;
 use DBI;


ncfavier:
I guess that explains https://github.com/nix-community/home-manager/issues/3858.

This should probably be fixed in the spirit of https://github.com/NixOS/nixpkgs/pull/124556 by using `makeBinaryWrapper` in https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/interpreters/perl/wrapper.nix.

ryuheechul:
@ncfavier Yeah, I think there is a high chance this would fix #3858 too.
Also as you suggested, fixing the perl wrapper seems to be better idea although I don't have a clue on how to do that.

modules/programs/command-not-found/command-not-found.pl
```diff
@@ -1,4 +1,4 @@
-#! @perl@/bin/perl -w
+#!/usr/bin/env -S @perl@/bin/perl -w
 
 use strict;
```
rycee, line 3: Another alternative, that I think may be better:

```suggestion
#!@perl@/bin/perl

use strict;
use warnings;
```

rycee:
Unfortunately, I think `makeBinaryWrapper` would probably not be the best for our use case since it would require most people to do a compile step.

Some time ago I worked on an alternative wrapper tool that attempts to be something of an intermediate between the standard Bash wrapper and the binary wrapper. That style of wrapper would be more suitable for our purposes, I think.

That said. I think in this case the best solution would be use remove the `-w` option and use `use warnings`.

ncfavier:
> it would require most people to do a compile step

What's wrong with that? Compiling the wrapper is very fast.

> remove the `-w` option

I don't see how this fixes the issue at hand, which is that Darwin doesn't support recursive shebangs, and `@perl@/bin/perl` is a shebang script.

Svep is a neat idea though, short of proper [declarative wrappers](https://github.com/NixOS/rfcs/pull/75)...

ryuheechul:
@ncfavier I see. Hopefully that can be addressed at upstream but I don't think I'm stepping into it today though.

Also I just learned that even if this shebang thing is fixed, the experience with `command-not-found` on darwin is just broken as adding `nixos` channel to root doesn't actually work properly.

However, I also learned that using `nix-index` instead works similar to the experience of `command-not-found` on nixos. So I settled with my workaround like this one, https://github.com/ryuheechul/dotfiles/commit/333f22e7de1197acf476dc68a748ec06d7537b54.

So I'm closing this PR for now.

Thanks for both @ncfavier and @rycee, I learned more about the subject thanks to your comments :)

PR #3916, opened 2023-04-25
zplug: Add options to allow override the package and the root dir of …

15cm:
…the package

The nixpkgs zplug package puts everything under `$out/` that pollutes the nix profile dir. Ideally the package output should be in a place like `$out/share/zplug`. Modifying the package itself will break backward compatibility, like here HM hardcodes the init script path to `${pkgs.zplug}/init.zsh`.

The zplugPackageRootDir option in this change allow using a zplug overlay that changes the output path, e.g.
```
zplug = super.zplug.overrideAttrs (old: {
  installPhase = ''
    outdir=$out/share/zplug
    mkdir -p $outdir
    cp -r $src/* $outdir/
  '';
});
```
and then in HM:
```
zplugPackageRootDir = "${cfg.package}/share/zplug"
```

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/zplug.nix
+++ b/modules/programs/zplug.nix
@@ -26,6 +26,21 @@ in {
   options.programs.zsh.zplug = {
     enable = mkEnableOption "zplug - a zsh plugin manager";
 
+    package = mkOption {
+      type = types.package;
+      default = pkgs.zplug;
+      defaultText = literalExpression "pkgs.zplug";
+      description = "The zplug package to install.";
+    };
+
+    zplugPackageRootDir = mkOption {
+      type = with types; either str path;
+      default = "${cfg.package}";
+      defaultText = literalExpression "cfg.package";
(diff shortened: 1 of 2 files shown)


ncfavier:
I would prefer a coordinated change between nixpkgs and home-manager to move things to `/share`. We could merge the HM change just when the nixpkgs change hits nixos-unstable, that way risks of breakage are minimal (and would be eventually be fixed by updating anyway, so not really a BC break).

15cm:
SGTM. My main concern was that a nixpkgs change will break other users who directly use the zplug package. But I think anyway it needs to be fixed, and as you said, the change will sit in unstable first. I'll prepare a PR in the nixpkgs and update this HM PR to reflect the location change of the `init.zsh`. Let's see how it goes on the nixpkgs side.

PR #3917, opened 2023-04-25
atuin: Use flags option with nushell integration

eboskma:
### Description

The nushell integration for atuin was not using the `flags` option. 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/atuin.nix
+++ b/modules/programs/atuin.nix
@@ -129,7 +129,9 @@ in {
         if not ($atuin_cache | path exists) {
           mkdir $atuin_cache
         }
-        ${cfg.package}/bin/atuin init nu ${flagsStr} | save --force ${config.xdg.cacheHome}/atuin/init.nu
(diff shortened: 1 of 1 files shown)


modules/programs/atuin.nix
```diff
@@ -129,7 +129,9 @@ in {
         if not ($atuin_cache | path exists) {
           mkdir $atuin_cache
         }
-        ${cfg.package}/bin/atuin init nu | save --force ${config.xdg.cacheHome}/atuin/init.nu
+        ${cfg.package}/bin/atuin init nu ${
+          builtins.concatStringsSep " " cfg.flags
```
ncfavier, line 133:
```suggestion
          flagsStr
```

eboskma: Ah, I thought I tested that and found it did not work. Just tested it again, and I guess I did something wrong initially.

ncfavier (approved):

PR #3920, opened 2023-04-25
home-manager-auto-upgrade: add Flake support

shikanime:
### Description

The use of Flake have increased quite a lot, supporting auto upgrade for Flake seems reasonable.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/services/home-manager-auto-upgrade.nix
```diff
@@ -61,7 +89,7 @@ in {
             echo "Update Nix's channels"
             ${pkgs.nix}/bin/nix-channel --update
             echo "Upgrade Home Manager"
-            ${homeManagerPackage}/bin/home-manager switch
+            ${homeManagerPackage}/bin/home-manager switch ${toString cfg.flags}
```
ncfavier, line 92:
```suggestion
            ${homeManagerPackage}/bin/home-manager switch ${lib.escapeShellArgs cfg.flags}
```

shikanime: It is intentional as `lib.escapeShellArgs` would produce
```
/nix/store/xxx-home-manager/bin/home-manager switch '--flake github:alice/dotfiles'
```
Which doesn't work, the intent is to produce:
```
/nix/store/xxx-home-manager/bin/home-manager switch --flake github:alice/dotfiles
```

ncfavier: So you should use `[ "--flake" (cfg.flake) ]`.

shikanime: On its own, it won't make much difference, but if we combine the two, we'll get the desired results 😊
```
[ "--flake" (lib.escapeShellArg cfg.flake) ] 
```

tests/modules/services/home-manager-auto-upgrade/basic-configuration.nix
eclairevoyant, line 1: shouldn't a separate test be added instead? also a proper test should be done to verify that the flake was in fact used, as is this points to a nonexistent github repo

modules/services/home-manager-auto-upgrade.nix
```diff
@@ -37,6 +62,9 @@ in {
         lib.platforms.linux)
     ];
 
+    services.home-manager.autoUpgrade.flags =
+      if cfg.flake != null then [ "--flake" cfg.flake ] else [ ];
```
eclairevoyant, line 66:
```suggestion
      lib.optionals (cfg.flake != null) [ "--flake" cfg.flake ];
```

eclairevoyant (changes requested):

tests/modules/services/home-manager-auto-upgrade/basic-configuration.nix
shikanime: Tbh, I haven't really found a practical way of testing this separately.

modules/services/home-manager-auto-upgrade.nix
shikanime: Why not, I mean there are many way to do it, I was inspired by the Nixpkgs' own implementation of the feature.

tests/modules/services/home-manager-auto-upgrade/basic-configuration.nix
eclairevoyant: create a minimal flake within this repo, don't depend on something external

shikanime: The main problem I had was to find a way to test

shikanime: The main problem I had was to find a way to test the generated content from the writeShellScript:

```nix
{
  config = {
    nmt.script = ''
      assertFileContent \
        home-files/.config/systemd/user/home-manager-auto-upgrade.service \
        ${./example-flake.service}
    '';
  };
}
```

```
[Service]
ExecStart=/nix/store/jz0qgdhbxv7cdgr371b7rgvlav5pn37f-home-manager-auto-upgrade
```

```shell
#!/nix/store/51sszqz1d9kpx480scb1vllc00kxlx79-bash-5.2-p15/bin/bash
echo "Update Nix's channels"
/nix/store/q6r9kwmidiy6wx1w1nf3ff0q40sfq4dg-nix-2.15.1/bin/nix-channel --update
echo "Upgrade Home Manager"
/nix/store/w4p00g8igncrz5h74nmxyhgfnh6s89a0-home-manager/bin/home-manager switch '--flake' 'github:alice/dotfiles'
```

But so far, I didn't found example in the repository to take inspiration

thefossguy:
Would it be better to detect if `flake.nix` existed and just run `nix flake update`? As of now, `pkgs.nix.version` evaluates to `"2.18.1"` with the `23.11` release of nixpkgs so we cannot do a `${pkgs.nix}/bin/nix flake update --flake $HOME/.config/home-manager/flake.nix` so we resort to `pushd`-ing and `popd`-ing.

```diff
diff --git a/modules/services/home-manager-auto-upgrade.nix b/modules/services/home-manager-auto-upgrade.nix
index 659f911..4531427 100644
--- a/modules/services/home-manager-auto-upgrade.nix
+++ b/modules/services/home-manager-auto-upgrade.nix
@@ -55,8 +55,16 @@ in {

         Service.ExecStart = toString
           (pkgs.writeShellScript "home-manager-auto-upgrade" ''
-            echo "Update Nix's channels"
-            ${pkgs.nix}/bin/nix-channel --update
+            if [ -f "$HOME/.config/home-manager/flake.nix" ]; then
+                pushd "$HOME/.config/home-manager"
+                    echo 'Running `nix flake update` for the Home Manager flake'
+                    ${pkgs.nix}/bin/nix flake update
+                popd
+            else
+                echo "Update Nix's channels"
+                ${pkgs.nix}/bin/nix-channel --update
+            fi
+
             echo "Upgrade Home Manager"
             ${homeManagerPackage}/bin/home-manager switch
           '');
```

PR #3922, opened 2023-04-26
zplug: Update the path of init.zsh

15cm:
### Description

The current zplug nixpkgs puts everything under `$out/`. It pollutes the nix
profile dir.

This is a breaking change. It depends on an change of the output path in the
nixpkgs zplug package.

Please see https://github.com/nix-community/home-manager/pull/3916 for the reason that we decided to go with a breaking change, and the alternative we considered there. @ncfavier FYI I create a new PR here as I failed to get Github branch rename work for the old PR.

### Checklist

- [x] ~~Change is backwards compatible.~~

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/zplug.nix
+++ b/modules/programs/zplug.nix
@@ -47,7 +47,7 @@ in {
     programs.zsh.initExtraBeforeCompInit = ''
       export ZPLUG_HOME=${cfg.zplugHome}
 
-      source ${pkgs.zplug}/init.zsh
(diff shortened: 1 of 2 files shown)


15cm:
I'll get back here once I create the PR in nixpkgs.

ncfavier:
Merge when https://nixpk.gs/pr-tracker.html?pr=228284 is all green.

PR #3926, opened 2023-04-28
zellij: adds options to integrate with zsh, bash and fish shells

erikkrieg:
### Description
Zellij is commonly [integrated with a shell to automatically launch sessions](https://zellij.dev/documentation/integration.html). When using `programs.zellij` you currently need to update your shell of choice to using something like `initExtra` to append the Zellij integration command. Not a big problem, but since many other packages include options like `enableZshIntegration` I figured it was appropriate to add the option to the Home Managers' Zellij package.

Integrations based on Zellij documentation have been added for the following shells:
- bash
- fish
- zsh

Options are defaulted to `false`. While similar options for other home-manager programs seem to default to `true` (like `fzf`) I opted for false for backwards compatibility.

As part of testing I used this branch as the source for home-manager and installed zellij along with each of the three shells, trying various permutations: true, false and undefined.

For automated tests, I referenced the tests for `programs.scmpuff`, which tests similar shell integrations.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


erikkrieg:
There weren't any tests for Zellij as far as I could tell, so I decided to hold off on tests until I got some feedback.

erikkrieg:
I find the manual to be a bit unclear for fish. There are two options for fish presented in this section of the doc: https://zellij.dev/documentation/integration.html#autostart-on-shell-creation

First, they show the option I used:

```fish
if set -q ZELLIJ
else
  zellij
end
```

Then they show:
```fish
if status is-interactive
    ...
    eval (zellij setup --generate-auto-start fish | string collect)
end
```

But for the second option the manual suggest that it might not work depending on the version of `fish`.

I opted for the first option because it sounded like there weren't any issues with it. Now, I am not a `fish` user, so I'm not personally sure what the preference is. But I also don't see an obvious alternative presented in the case where `eval (zellij setup --generate-auto-start fish | string collect)` doesn't work. So, even if we detect different `fish` versions, I'm not sure what to do with that.

When I have some time I'll take a look at a few different versions of `fish` and see what happens.

ambroisie:
@erikkrieg I think it is okay to use what upstream suggests, given that it should work on the latest version of `fish` (IIUC).

We could be extra thorough and print a warning if the version is older than X.Y, and point to the documentation. But that's going the extra mile, and not absolutely necessary IMO.

erikkrieg:
Anything I'm missing here? Someone I should ping to snare a PR reviewer?

modules/programs/zellij.nix
```diff
@@ -39,6 +39,30 @@ in {
...
       '';
     };
+
+    enableBashIntegration = mkOption {
+      default = false;
+      type = types.bool;
+      description = ''
+        Whether to enable Bash integration.
+      '';
+    };
```
ncfavier, line 49: We usually make shell integration options default to `true`, using something like `mkEnableOption "Foo integration" // { default = true; }`.

ncfavier: Ok, maybe `false` is better for backwards compatibility. I'd still go with `mkEnableOption`.

ambroisie: Could the default be based on `stateVersion` (`false` if `stateVersion` < 23.05)?

erikkrieg: We can assume most configs that are enabling the zellij package will be configuring the shell integrations themselves. So if we enable these shell integrations as the default there will be some redundant code in their respective shells' rc files.

While I haven't tested it yet, I do think that these shell integration evals are idempotent, so it could be relatively safe to enable them by default. There could be the cases, though, where a users with more bespoke integrations encounter conflicts. Not sure how common it is to deviate from the documented approach, though.

So there is a decent case for either option, imo. I'm happy to go with whatever y'all prefer.

erikkrieg:
I see the formatter CI check failed. I'll take a look at that once GitHub isn't on fire anymore.

modules/programs/zellij.nix
ncfavier: Let's go with `true`.

erikkrieg:
@ncfavier I've changed the options to `mkEnableOption` that default to `true` (also fixed some formatting errors that I missed earlier, by bad). Thanks!

tests/modules/programs/zellij/default.nix
```diff
@@ -0,0 +1,9 @@
+{
+  zellij-bash = ./bash.nix;
+  zellij-no-bash = ./no-bash.nix;
+  zellij-no-shell = ./no-shell.nix;
```
ncfavier, line 4: Where's `no-shell.nix`?

erikkrieg: Good catch with the `no-shell`. I referenced another package's tests when creating these ones.

I can merge these into two tests: one where all shells are enabled, and another where they are all disabled.

tests/modules/programs/zellij/disable-shells.nix
```diff
@@ -0,0 +1,34 @@
...
+  };
+
+  nmt.script = ''
+    assertFileNotRegex home-files/.bashrc \
+      'eval "$(zellij setup --generate-auto-start bash)"'
+
+    assertFileNotRegex home-files/.zshrc \
+      'eval "$(zellij setup --generate-auto-start zsh)"'
+
+    assertFileNotRegex home-files/.config/fish/config.fish '@zellij@'
```
ncfavier, line 32: I still feel like this isn't a very useful test... If the integration snippets ever change, this test won't fail, so it probably won't be updated, and then it won't be testing anything any more.

Just remove it IMO.

tests/modules/programs/zellij/default.nix
erikkrieg: Tests cases have been reduced to just two, now.

tests/modules/programs/zellij/disable-shells.nix
erikkrieg: Talking about the fish one in particular? Or the lot of them? Both the tests for enabling and disabling? 

We can also make the fish regex look for `eval (zellij setup --generate-auto-start fish | string collect)`.

erikkrieg: As eluded to in a past comment, I ripped these tests off of another package doing this exact thing. Not at all married to them at all. Just let me know if by getting rid of the tests, you mean all of the tests introduced by this PR or just some subset (like the disable fish one?)

ncfavier: I'm talking about the `disable-shells.nix` test.

erikkrieg: Sounds good. I'll drop those.

ncfavier:
Needs formatting

K900:
Should this really be on by default? I got a bit spooked when it just ... showed up, to be honets.

erikkrieg:
> Should this really be on by default? I got a bit spooked when it just ... showed up, to be honets.

Originally I had it disabled by default. It was requested to enable by default in this resolved convo: https://github.com/nix-community/home-manager/pull/3926#discussion_r1188346338

ncfavier:
We can still flip it to `false` if that's better. PRs welcome.

erikkrieg:
@ncfavier I can get a PR open that changes the default to `false` so that it is available in the case where you decide to go in that direction.

erikkrieg:
@ncfavier Started a PR with the integrations disabled by default here: https://github.com/nix-community/home-manager/pull/3981

PR #3927, opened 2023-04-28
README: move releases section higher

ncfavier:
This is important information that people still have [trouble finding](https://github.com/nix-community/home-manager/pull/3454#issuecomment-1527173043), let's make it more obvious.


rycee:
Thanks! Merged to master now.

PR #3930, opened 2023-04-29
nushell: add envVars attribute

happysalada:
### Description

Much like what was done with shellAliases, I wanted a convenient way to set environment variables.
Admittedly that can be done in extraEnv. The one use case where this becomes very convenient is with the appearance of those CLI tools that require an API key in your env. That is env var should be a secret and ideally read from a path. This is just a more convenient way of having an env var be a string or read from a path.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/nushell.nix
+++ b/modules/programs/nushell.nix
@@ -121,6 +121,15 @@ in {
         this option) to command strings or directly to build outputs.
       '';
     };
+
+    environmentVariables = mkOption {
+      type = types.attrsOf types.str;
+      default = { };
+      example = { FOO = "BAR"; };
+      description = ''
+        An attribute set that maps an environment variable to a shell interpreted string.
+      '';
+    };
   };
 
   config = mkIf cfg.enable {
@@ -141,10 +150,15 @@ in {
         ];
       })
 
-      (mkIf (cfg.envFile != null || cfg.extraEnv != "") {
+      (let
+        envVarsStr = concatStringsSep "\n"
+          (mapAttrsToList (k: v: "let-env ${k} = ${v}")
+            cfg.environmentVariables);
+      in mkIf (cfg.envFile != null || cfg.extraEnv != "" || envVarsStr != "") {
         "${configDir}/env.nu".text = mkMerge [
           (mkIf (cfg.envFile != null) cfg.envFile.text)
           cfg.extraEnv
+          envVarsStr
         ];
       })
     ];
--- a/tests/modules/programs/nushell/env-expected.nu
+++ b/tests/modules/programs/nushell/env-expected.nu
@@ -1,2 +1,4 @@
 let-env FOO = 'BAR'
 
+
+let-env BAR = $'(echo BAZ)'
--- a/tests/modules/programs/nushell/example-settings.nix
+++ b/tests/modules/programs/nushell/example-settings.nix
@@ -20,6 +20,8 @@
       "lsname" = "(ls | get name)";
       "ll" = "ls -a";
     };
+
+    environmentVariables = { BAR = "$'(echo BAZ)'"; };
   };
 
   test.stubs.nushell = { };


happysalada:
After a small discussion in https://github.com/ryantm/agenix/pull/180
The best way for secrets is probably not to accept a source attribute since those need to be shell expanded.
I changed the design to just be an attribute set where values will be shell expanded.

modules/programs/nushell.nix
```diff
@@ -121,6 +121,16 @@ in {
         this option) to command strings or directly to build outputs.
       '';
     };
+
+    envVars = mkOption {
```
bobvanderlinden, line 125: Personally I'd write this out as environmentVariables. Just to be in line with options like sessionVariables.

modules/programs/nushell.nix
```diff
@@ -121,6 +121,16 @@ in {
...
+      type = types.attrsOf types.str;
+      default = { };
+      example = { FOO.text = "$'(echo BAR)'"; };
```
bobvanderlinden, line 128:
```suggestion
      example = { FOO = "$'(echo BAR)'"; };
```
I'd also expect a more straightforward example without command substitution.

modules/programs/nushell.nix
```diff
@@ -141,10 +151,14 @@ in {
         ];
       })
 
-      (mkIf (cfg.envFile != null || cfg.extraEnv != "") {
+      (let
+        envVarsStr = concatStringsSep "\n"
+          (mapAttrsToList (k: v: "let-env ${k} = ${v}") cfg.envVars);
```
bobvanderlinden, line 156: This is a bit confusing. You'd presume you can write something like:

```
EDITOR = "${pkgs.vim}/bin/vim"
```
But you'll have to do something like:

```
EDITOR = ''
  "${pkgs.vim}/bin/vim"
''
```
Or use escaping:

```
EDITOR = "\"${pkgs.vim}/bin/vim\""
```

It would be nice if the option would do this by default.

This is in line with an earlier PR that provides arbitrary settings. By default the literal value is used, but you can use .__nu to write Nu code instead. See https://github.com/nix-community/home-manager/pull/3831

happysalada: I've just tested to make sure with `    EDITOR = "${pkgs.helix}/bin/hx";`
and it works like a charm.

modules/programs/nushell.nix
happysalada: you suggestion is much better indeed.

modules/programs/nushell.nix
happysalada: I wanted to convey the fact it was shell expanded. I'm happy to go with something simpler though.

modules/programs/nushell.nix
bobvanderlinden: Oh right, without spaces it is fine of course. Good point 👍

happysalada:
@rycee @ncfavier when you have a moment.

modules/programs/nushell.nix
```diff
@@ -121,6 +121,15 @@ in {
         this option) to command strings or directly to build outputs.
       '';
     };
+
+    environmentVariables = mkOption {
+      type = types.attrsOf types.str;
+      default = { };
+      example = { FOO = "BAR"; };
```
ncfavier, line 128: Wouldn't `BAR` need to be quoted, if this is interpreted by the shell?

I think it would make more sense to escape the values here so that strings are treated literally, and you can still use `extraEnv` for variables that require arbitrary code.

happysalada: I personally like that both are possible here.
You can use it for arbitrary code if you want to or use it with just bare strings if you prefer.
Do you feel strongly about this ?
For me extraEnv would be for stuff that require intermediary variables or more complicated code. One liners would still fit in the environmentVariables category.

ncfavier: > Do you feel strongly about this ?

Not at all. I don't even use nushell.

ncfavier (approved):

PR #3933, opened 2023-04-29
sapling: add module

pbar1:
sapling: add module
Adds a program module for [Sapling](https://sapling-scm.com/).

Since Sapling itself is very similar in nature to Mercurial,
`modules/programs/mercurial.nix` was copied to make this module with the
ignore pieces removed (Sapling respects gitignore).

### DIFF
--- a/modules/programs/sapling.nix
+++ b/modules/programs/sapling.nix
@@ -1,0 +1,83 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.programs.sapling;
+
+  iniFormat = pkgs.formats.ini { };
+
+in {
+  meta.maintainers = [ maintainers.pbar ];
+
+  options = {
+    programs.sapling = {
+      enable = mkEnableOption "Sapling";
+
+      package = mkPackageOption pkgs "sapling" { };
+
+      userName = mkOption {
+        type = types.str;
+        description = "Default user name to use.";
+      };
+
+      userEmail = mkOption {
+        type = types.str;
+        description = "Default user email to use.";
+      };
+
+      aliases = mkOption {
+        type = types.attrsOf types.str;
+        default = { };
+        description = "Sapling aliases to define.";
+      };
+
+      extraConfig = mkOption {
+        type = types.attrsOf types.anything;
+        default = { };
+        description = "Additional configuration to add.";
+      };
+
+      iniContent = mkOption {
+        type = iniFormat.type;
+        internal = true;
+      };
+    };
+  };
+
+  config = mkIf cfg.enable (mkMerge [
+    {
+      home.packages = [ cfg.package ];
+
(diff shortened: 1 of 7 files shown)


modules/programs/sapling.nix
```diff
@@ -0,0 +1,81 @@
...
+      };
+
+      aliases = mkOption {
+        type = types.attrsOf types.anything;
+        default = { };
+        description = "Sapling aliases to define.";
+      };
+
+      extraConfig = mkOption {
+        type = types.either (types.attrsOf types.anything) types.lines;
```
rycee, line 38: Is there any option that cannot be represented through the attribute set alternative? Ideal would be to avoid the need for `types.lines` here.

modules/programs/sapling.nix
```diff
@@ -0,0 +1,81 @@
...
+        description = "Default user name to use.";
+      };
+
+      userEmail = mkOption {
+        type = types.str;
+        description = "Default user email to use.";
+      };
+
+      aliases = mkOption {
+        type = types.attrsOf types.anything;
```
rycee, line 32: Is it really possible to have anything as an alias? I would have guessed that `types.attrsOf types.str` would be sufficient.

rycee:
Thanks for the contribution, I added a few comments. Please also add a test case and wrap the commit message body at [72 characters](https://cbea.ms/git-commit/#wrap-72).

For future reference, the description template with the checklist is for the benefit of the reviewer so please take the time to fill it out.

pbar1:
> Thanks for the contribution, I added a few comments. Please also add a test case and wrap the commit message body at [72 characters](https://cbea.ms/git-commit/#wrap-72).
> 
> For future reference, the description template with the checklist is for the benefit of the reviewer so please take the time to fill it out.

Thanks for the comments! I've addressed them; added some tests, and you're right about the typing that we can get away with in both cases.

Ah, my apologies - I didn't even see the description template (ironically) due to Sapling itself 😅. I submitted this with `sl pr submit --draft` initially, which I think bypassed the templated description and just uses the commit message.

For some more context, if you see my branch name being this PR number as well as frequent force pushes, those are also artifacts of the "Sapling flow" of `sl commit` -> `sl pr submit --draft`  -> `sl amend` -> `sl pr submit`.

shikanime:
A small improvement is to configure Sapling to utilize the .gitignore file located in the `$HOME/.config/git/ignore` configured coming from the [git](https://github.com/nix-community/home-manager/blob/e8aaced73ebaf6bfa8e3c6ab0a19cb184bc4d798/modules/programs/git.nix#L376-L378) module, by default.

Similarly to:

```nix
ui."ignore.gitconfig" = "${config.home.homeDirectory}/.config/git/ignore";
```

rycee:
@pbar1 Sorry about the long delay. Merged to master now 🙂

@shikanime Sounds like a nice idea. It could be added as a new PR.

PR #3935, opened 2023-05-01
ssh: add setEnv option

GaetanLepage:
### Description

Add support for the `SetEnv` option in the `programs.ssh` module.
It allows to set environment variables and send them to the server.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/ssh.nix
+++ b/modules/programs/ssh.nix
@@ -15,6 +15,12 @@ let
 
   unwords = builtins.concatStringsSep " ";
 
+  mkSetEnvStr = envStr: unwords
+    (mapAttrsToList
+      (name: value: ''${name}="${escape [ "\"" "\\" ] (toString value)}"'')
+      envStr
+    );
+
   bindOptions = {
     address = mkOption {
       type = types.str;
@@ -189,6 +195,14 @@ let
         '';
       };
 
+      setEnv = mkOption {
+        type = with types; attrsOf (oneOf [ str path int float ]);
+        default = {};
+        description = ''
+          Environment variables and their value to send to the server.
+        '';
+      };
+
       compression = mkOption {
         type = types.nullOr types.bool;
         default = null;
@@ -322,6 +336,7 @@ let
     ++ optional (cf.hostname != null)        "  HostName ${cf.hostname}"
     ++ optional (cf.addressFamily != null)   "  AddressFamily ${cf.addressFamily}"
     ++ optional (cf.sendEnv != [])           "  SendEnv ${unwords cf.sendEnv}"
+    ++ optional (cf.setEnv != {})            "  SetEnv ${mkSetEnvStr cf.setEnv}"
     ++ optional (cf.serverAliveInterval != 0)
       "  ServerAliveInterval ${toString cf.serverAliveInterval}"
     ++ optional (cf.serverAliveCountMax != 3)
--- a/tests/modules/programs/ssh/match-blocks-attrs-expected.conf
+++ b/tests/modules/programs/ssh/match-blocks-attrs-expected.conf
@@ -5,6 +5,7 @@ Host * !github.com
 Host abc
   ProxyJump jump-host
 Host xyz
+  SetEnv BAR="_bar_ 42" FOO="foo12"
   ServerAliveInterval 60
   ServerAliveCountMax 10
   IdentityFile file
--- a/tests/modules/programs/ssh/match-blocks-attrs.nix
+++ b/tests/modules/programs/ssh/match-blocks-attrs.nix
@@ -35,6 +35,10 @@ with lib;
             }
           ];
           dynamicForwards = [{ port = 2839; }];
+          setEnv = {
+            FOO = "foo12";
+            BAR = "_bar_ 42";
+          };
         };
 
         "* !github.com" = {


ncfavier (approved):

modules/programs/ssh.nix
```diff
@@ -15,6 +15,12 @@ let
 
   unwords = builtins.concatStringsSep " ";
 
+  mkSetEnvStr = envStr: unwords
+    (mapAttrsToList
+      (name: value: "${name}=${value}")
```
sumnerevans, line 20: `value` looks like it needs to be escaped.

ncfavier: The man page is unclear about what escaping is possible. Testing would be needed.

teto:
Rewriting the ssh option spec in nix looks daunting, wouldn't it be simpler to support freeform types with some adequate submodules ? see nixpkgs ssh server for a basic example.

modules/programs/ssh.nix
GaetanLepage: I have wrapped the `value` in `"` (double quotes). Is that what you were thinking about ?

GaetanLepage:
> Rewriting the ssh option spec in nix looks daunting, wouldn't it be simpler to support freeform types with some adequate submodules ? see nixpkgs ssh server for a basic example.

I agree. In the long term that would be very appreciated.
However this would require a bit more work than this simple and straightforward PR...

modules/programs/ssh.nix
```diff
@@ -15,6 +15,12 @@ let
 
   unwords = builtins.concatStringsSep " ";
 
+  mkSetEnvStr = envStr: unwords
+    (mapAttrsToList
+      (name: value: ''${name}="${value}"'')
```
rycee, line 20: I think maybe the value needs to be escaped, e.g., in case it contains a `"` character?

GaetanLepage: with `escapeShellArg` ?

GaetanLepage: `(name: value: ''${name}="${escape [ "\"" ] value}"'')` ?

ncfavier: Who knows. Read the source code, or experiment until you're confident with how the string is processed. Or just leave `value` raw and let the user figure it out.

teto: ssh -t checks the config (though it doesnt warn of everything), it's used in nixpkgs.

GaetanLepage: I did some testing:
```
        setEnv.LC_EASY = "abc123";
        setEnv.LC_WITH_SPACE = "abc 123";
        setEnv.LC_WITH_1_QUOTE = ''abc" 123'';
        setEnv.LC_WITH_2_QUOTES = ''abc"" 123'';
```
I now have an implementation that works for all of the above test cases:
`(name: value: ''${name}="${escape [ "\"" ] value}"'')`

Result (on the remote host):
```
auriga% env | grep LC
LC_WITH_SPACE=abc 123
LC_WITH_1_QUOTE=abc" 123
LC_WITH_2_QUOTES=abc"" 123
LC_EASY=abc123
```

ncfavier: How about `setEnv.LC_BACKSLASH = ''a\b'';`?

GaetanLepage: > How about `setEnv.LC_BACKSLASH = ''a\b'';`?

It is rendered as this in the ssh config:
`  SetEnv LC_TEST="a\b"`

And, when evaluated on the host, it gives:
```
auriga% echo $LC_TEST 
a
```

I guess that it is the correct behavior... I mean that if the user wants to set `LC_FOO = ''a\b''`, he wants to have `SetEnv LC_FOO="a\b"` in his ssh config. What do you think ?

ncfavier: If I set that I'd expect `LC_TEST` to contain `a\b` on the host. Try adding `\` to the list of escaped characters?

GaetanLepage: Escaping it does produce `LC_TEST="a\\b"` in the ssh config but on the host it doesn't prevent `LC_TEST` from being just `a`.

More precisely, it seems that `LC_TEST` does contain the backslash but it cannot be "echo-ed"

rycee: Glad `\"` works. I think it also makes sense to escape `\` since that is a pretty common character one might want to pass on. Don't think it's worth checking more characters, though. If somebody runs into any further ones then they can raise an issue and we'll fix it then.

Btw, please add a test case to [`tests/modules/programs/ssh`](https://github.com/nix-community/home-manager/tree/master/tests/modules/programs/ssh) for this to avoid regressions :-)

GaetanLepage: Actually, the `home.sessionVariables` implementation does not escape anything:
https://github.com/nix-community/home-manager/blob/6abb775e75a7f25451781dc704cfe03f27ad7496/modules/lib/shell.nix#L5

GaetanLepage: > In this regard, do you think that we should keep going with the escaping strategy or replicating the same thing as for `home.sessionVariables` (i.e. a simple `toString`) ?

@rycee @teto @ncfavier What is your opinion on this ?

ncfavier: `sessionVariables` doesn't use `escapeShellArg` in order to allow variable expansion, but it should probably escape `"` and `\` to be honest.

In any case, I think this is fine; please add a test.

GaetanLepage: Thank you for the feedback :)
I added a test.

ncfavier (approved):

PR #3936, opened 2023-05-01
vim: Allow plugins to specify a config

stelth:
### Description

Currently, in order to specify config options for plugins, you have to add the configuration to extraConfig. If you break up your vim definitions across multiple files, the ordering of the extraConfig options are not guaranteed.

Example:
default.nix:
```
{...}: {
  import = [./other-options.nix];
  programs.vim = {
    enable = true;
    extraConfig = ''
    let mapleader = ' '
    '';
  }
}
```

other-options.nix:
```
{...}: {
  programs.vim = {
    plugins = with pkgs.vimPlugins [pluginThatNeedsKeymap];

    extraConfig = ''
      nmap <leader>k <Plug>doPluginThing<CR>
    '';
  };
}
```

If other-option's extraConfig comes first, then, it will be using the wrong mapleader.

After this change (using the same default.nix from above):
other-options.nix:
```
{...}: {
  programs.vim = {
    plugins = with pkgs.vimPlugins [
    {
      plugin = pluginThatNeedsKeymap;
      config = ''
	nmap <leader>k <Plug>doPluginThing<CR>
      '';
    ];
  };
}
```

Now, all of the plugin specific configuration options are guaranteed to come after all of the extraConfig options and you don't have to worry about extraConfig's getting rearranged or refactoring your vim config.

The neovim module allows tacking configuration on to the plugin definition, this PR mostly lifts that portion out and tacks it on to the existing vim module without breaking backwards compatibility.

### Checklist

- [NA] Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).

    ```
    {component}: {description}

    {long description}
    ```

  - [NA] Added myself as module maintainer. See [example](https://github.com/nix-community/home-manager/blob/068ff76a10e95820f886ac46957edcff4e44621d/modules/programs/lesspipe.nix#L6).

  - [NA] Added myself and the module files to `.github/CODEOWNERS`.

### DIFF
--- a/modules/programs/vim.nix
+++ b/modules/programs/vim.nix
@@ -7,6 +7,21 @@ let
   cfg = config.programs.vim;
   defaultPlugins = [ pkgs.vimPlugins.vim-sensible ];
 
+  pluginWithConfigType = types.submodule {
+    options = {
+      config = mkOption {
+        type = types.nullOr types.lines;
+        description = "Script to configure this plugin.";
+        default = null;
+      };
+
+      plugin = mkOption {
+        type = types.package;
+        description = "vim plugin";
+      };
+    };
+  };
+
   knownSettings = {
     background = types.enum [ "dark" "light" ];
     backupdir = types.listOf types.str;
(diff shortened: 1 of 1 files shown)


modules/programs/vim.nix
```diff
@@ -54,28 +69,41 @@ let
...
       plugins = mkOption {
-        type = with types; listOf (either str package);
-        default = defaultPlugins;
-        example = literalExpression "[ pkgs.vimPlugins.YankRing ]";
+        type = with types;
+          listOf (oneOf [
+            package
+            pluginWithConfigType
+            str
+            (listOf (oneOf [ package pluginWithConfigType str ]))
```
ncfavier, line 88: Why would we allow lists of lists?

PR #3938, opened 2023-05-01
i3status-rust: satisfy new 0.31 TOML output requirements

amesgen:
### Description

From the [0.31.0 news](https://github.com/greshake/i3status-rust/blob/49499c1ca624228c4b9e88b62b36cd0fa5e2dce5/NEWS.md#i3status-rust-0310):

> `block = "..."` is now required to be the first field of block configs.

The approach taken here normalizes numbers like `10.0` to `10`, but I think this should be inconsequential.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/i3status-rust.nix
+++ b/modules/programs/i3status-rust.nix
@@ -6,7 +6,18 @@ let
 
   cfg = config.programs.i3status-rust;
 
-  settingsFormat = pkgs.formats.toml { };
+  settingsFormat = pkgs.formats.toml { } // {
+    # Since 0.31, the "block" key has to be first in the TOML output.
+    generate = name: value:
+      pkgs.runCommand name {
+        nativeBuildInputs = [ pkgs.jq pkgs.remarshal ];
+        value = builtins.toJSON value;
+        passAsFile = [ "value" ];
+      } ''
+        jq '.block |= map({block: .block} + del(.block))' "$valuePath" \
+          | json2toml --preserve-key-order > "$out"
+      '';
+  };
 
 in {
   meta.maintainers = with lib.maintainers; [ farlion thiagokokada ];
--- a/tests/modules/programs/i3status-rust/with-custom.nix
+++ b/tests/modules/programs/i3status-rust/with-custom.nix
@@ -105,12 +105,12 @@
         ${
           pkgs.writeText "i3status-rust-expected-config" ''
             [[block]]
-            alert = 10.0
             block = "disk_space"
+            alert = 10
             info_type = "available"
             interval = 60
             path = "/"
-            warning = 20.0
+            warning = 20
 
             [[block]]
             block = "memory"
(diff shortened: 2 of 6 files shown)


modules/programs/i3status-rust.nix
```diff
@@ -8,6 +8,16 @@ let
...
   settingsFormat = pkgs.formats.toml { };
 
+  # Since 0.31, the "block" key has to be first in the TOML output.
+  prefixBlockType = toml:
+    toml.overrideAttrs (oldAttrs: {
+      nativeBuildInputs = oldAttrs.nativeBuildInputs ++ [ pkgs.jq ];
+      buildCommand = ''
+        cat "$valuePath" \
+          | jq '.block |= map({block: .block} + del(.block))' \
+          | json2toml -p > "$out"
```
ncfavier, line 18: The use of `cat` is unnecessary here.

Also, rather than using `overrideAttrs` and coupling the module to the internals of `pkgs.formats.toml.generate`, I think the derivation is short enough that it's better to recreate it from scratch.

ncfavier (commented):
I think this change is very unfortunate and I've [complained](https://github.com/greshake/i3status-rust/commit/b49f9a8fb7a296b52ca3336cb8a3f0648002cae9#commitcomment-111367444) to upstream about it. We are now essentially depending on the undocumented, accidental facts that both `jq` and `json2toml` preserve key order...

amesgen:
> I think this change is very unfortunate and I've [complained](https://github.com/greshake/i3status-rust/commit/b49f9a8fb7a296b52ca3336cb8a3f0648002cae9#commitcomment-111367444) to upstream about it. We are now essentially depending on the undocumented, accidental facts that both `jq` and `json2toml` preserve key order...

Fully agreed that this is quite unfortunate, thanks for reporting it upstream! Also, we could hold off merging this PR if it becomes clear that upstream is open to reverting this change.

Note that `json2toml`'s `-p` flag is specifically about preserving the key order:
```
  -p, --preserve-key-order
                        preserve the order of dictionary/mapping keys
```
For `jq`, it is indeed less explicit (e.g. https://github.com/stedolan/jq/issues/1364#issuecomment-290883930); other implementations like [`gojq`](https://github.com/itchyny/gojq) don't preserve key order.

ncfavier:
> we could hold off merging this PR

The release is already out, we might as well support it...

> `-p`

Neat.

ncfavier (approved):

ncfavier:
Upstream heard my plea! https://github.com/greshake/i3status-rust/commit/b8eb8d03ca8249a2b8af255f6403e718581c65e8

rycee:
Good job with raising this with upstream @ncfavier, it really seems like the best place to fix this and IMHO their code looks better after the fix as well 🙂

In any case, I think reverting would be preferred once the release reaches the unstable Nixpkgs channels + grace period of at least a week or so.

PR #3945, opened 2023-05-04
zsh: add `package` option

christoph-heiss:
### Description

Much like it is already available for many other modules, add it here too.

[ Also, I noticed there is no explicitly assigned maintainer/codeowner for the `zsh` module. I'd be willing to take over and actively maintain if that would be desired, as I'm using it since many years now and care about it. ] 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/zsh.nix
+++ b/modules/programs/zsh.nix
@@ -215,6 +215,8 @@ in
     programs.zsh = {
       enable = mkEnableOption "Z shell (Zsh)";
 
+      package = mkPackageOption pkgs "zsh" { };
+
       autocd = mkOption {
         default = null;
         description = ''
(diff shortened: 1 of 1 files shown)


cyntheticfox:
Would probably make sense to extend to the other package parts like prezto, but would take a lot of work to ensure it works. This is definitely good to start with.

modules/programs/zsh.nix
```diff
@@ -482,7 +484,7 @@ in
     }
 
     {
-      home.packages = with pkgs; [ zsh ]
+      home.packages = with pkgs; [ cfg.package ]
         ++ optional cfg.enableCompletion nix-zsh-completions
         ++ optional cfg.oh-my-zsh.enable cfg.oh-my-zsh.package;
```
sumnerevans, line 489: Seems cleaner to get rid of the `with` because it's only being used by one of thing here.

```suggestion
      home.packages = [ cfg.package ]
        ++ optional cfg.enableCompletion pkgs.nix-zsh-completions
        ++ optional cfg.oh-my-zsh.enable cfg.oh-my-zsh.package;
```

sumnerevans (changes requested):
Looks good overall!

sumnerevans (approved):

PR #3946, opened 2023-05-04
xfconf: support uint

gou4shi1:
#### The Problem
The xfconf module of home-manager can't distinguish between `int` and `uint`, all exported integer types in xml are `int`, e.g. the result of this module is `<property name="dpms-on-ac-sleep" type="int" value="15"/>`, while the correct config in xml should be `<property name="dpms-on-ac-sleep" type="uint" value="15"/>`.
Some xfconf consumers use `xfconf_channel_get_uint()` to get config, and this function discards values with wrong types, e.g. it will discard `15` from `<property name="dpms-on-ac-sleep" type="int" value="15"/>`.

#### The Solution
change from
```
"xfce4-power-manager/dpms-on-ac-sleep" = 15;
```
to
```
"xfce4-power-manager/dpms-on-ac-sleep" = { type = "uint"; value = 15; };
```

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/xfconf.nix
+++ b/modules/misc/xfconf.nix
@@ -6,8 +6,31 @@ let
 
   cfg = config.xfconf;
 
+  xfIntVariant = types.submodule {
+    options = {
+      type = mkOption {
+        type = types.enum [ "int" "uint" "uint64" ];
+        description = ''
+          To distinguish between int, uint and uint64 in xfconf,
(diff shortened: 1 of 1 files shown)


ncfavier (approved):

gou4shi1:
Thanks!

Issue #3611, opened 2023-01-21
bug: readline quoting breaks "keyname: function-name or macro" style


n8henrie:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

https://www.gnu.org/software/bash/manual/html_node/Readline-Init-File-Syntax.html

The (IMO) much more readable alternative to the "keyseq" style as used in the example:
https://github.com/nix-community/home-manager/blob/275ab728912006eecb549338a50f24f294a7cfb7/modules/programs/readline.nix#L34

would be:  `{ Control-h = "backward-kill-word"; }`

Unfortunately the current implementation quotes `Control-h` and outputs `"Control-h"`, which doesn't work. Obviously one could still put it in `extraConfig`, but it would be nice if it didn't default to adding quotes that I didn't ask for.

I think it makes sense to expect users to explicitly escape quotes if they want them, since it seems equally valid and more readable to *not* have them in many cases.

https://github.com/nix-community/home-manager/blob/275ab728912006eecb549338a50f24f294a7cfb7/modules/programs/readline.nix#L24

It seems like the most straightforward backward-compatible way to go about this would be adding a new option (`keynames`? `unquotedBindings`?) happy to make a PR if it seems reasonable. 

### Maintainer CC

@vojta001 
@rycee 

### System information

```markdown
- system: `"aarch64-darwin"`
 - host os: `Darwin 22.2.0, macOS 13.1`
 - multi-user?: `yes`
 - sandbox: `no`
 - version: `nix-env (Nix) 2.11.1`
 - channels(n8henrie): `"darwin, nixpkgs-22.11"`
 - channels(root): `""`
 - nixpkgs: `/Users/n8henrie/.nix-defexpr/channels/nixpkgs`
```

ncfavier:
A simple heuristic would be to consider that a key is a keyseq if it contains a backslash, and a keyname otherwise. I'm pretty confident that keynames can't contain backslashes, less confident about the other way around unfortunately. Does it make sense at all to define a keyseq without any modifiers?

If this can't be done I'd go with adding a new option `keynameBindings`.

n8henrie:
Don't think that will always work. This is a valid `keyseq`:

```
"ab": "asdf"
```

hpfr:
[...]

PR #3947, opened 2023-05-04
readline: Add support for keynames

n8henrie:
[Kenames](Kenames) like `Control-n` or `meta-p` should not be quoted or they don't work.

Keyseqs like `\C-p` or `ab` must continue to be quoted.

See also: https://www.gnu.org/software/bash/manual/html_node/Readline-Init-File-Syntax.html

Fixes https://github.com/nix-community/home-manager/issues/3611

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


ncfavier (approved):
Not great but I don't have a better idea.

n8henrie:
@ncfavier I'm also fine with the `keynameBindings` option we discussed, but it seemed like this would probably work for the 99% use case (with others needing the escape hatch if they want a binding for the literal keystrokes <kbd>c</kbd><kbd>o</kbd><kbd>n</kbd><kbd>t</kbd><kbd>r</kbd><kbd>o</kbd><kbd>l</kbd><kbd>-</kbd><kbd>p</kbd>)

tests/modules/programs/readline/using-all-options.txt
```diff
@@ -4,8 +4,8 @@ $include /etc/inputrc
 set bell-style audible
 set completion-map-case on
 set completion-prefix-display-length 2
+Control-p: "whups"
 "\C-h": backward-kill-word
 $if mode=emacs
 "\e[1~": beginning-of-line
 $endif
-
```
ncfavier, line 11: !

n8henrie:
Restored the surplus newline at the end of the test file, tests should pass now.

PR #3953, opened 2023-05-05
pass-secret-service: Add dbus file, assert

cyntheticfox:
### Description

Add the dbus service file in the package folder to XDG_DATA_HOME, as well as adding an assertion to ensure both it and `gnome-keyring` aren't enabled at the same time.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -489,7 +489,8 @@ Makefile                                              @thiagokokada
 
 /modules/services/parcellite.nix                      @gleber
 
-/modules/services/pass-secret-service.nix             @cab404
+/modules/services/pass-secret-service.nix             @cab404 @cyntheticfox
+/tests/modules/services/pass-secret-service.nix       @cyntheticfox
 
 /modules/services/password-store-sync.nix             @pacien
 
--- a/modules/services/gnome-keyring.nix
+++ b/modules/services/gnome-keyring.nix
@@ -30,7 +30,11 @@ in {
         lib.platforms.linux)
       {
         assertion = !config.services.pass-secret-store.enable;
-        message = "Only one libsecret service can be enabled at a time.";
+        message = ''
+          Only one secrets service per user can be enabled at a time.
+          Other services enabled:
+          - pass-secret-store
+        '';
       }
     ];
 
--- a/modules/services/gnome-keyring.nix
+++ b/modules/services/gnome-keyring.nix
@@ -28,6 +28,10 @@ in {
     assertions = [
       (lib.hm.assertions.assertPlatform "services.gnome-keyring" pkgs
         lib.platforms.linux)
+      {
+        assertion = !config.services.pass-secret-store.enable;
+        message = "Only one libsecret service can be enabled at a time.";
+      }
     ];
 
     systemd.user.services.gnome-keyring = {
--- a/modules/services/pass-secret-service.nix
+++ b/modules/services/pass-secret-service.nix
@@ -15,8 +15,8 @@ in {
     package = mkPackageOption pkgs "pass-secret-service" { };
 
     storePath = mkOption {
-      type = types.str;
-      default = "";
+      type = with types; nullOr str;
+      default = null;
       defaultText = "$HOME/.password-store";
       example = "/home/user/.local/share/password-store";
       description = ''
@@ -52,7 +52,7 @@ in {
         Service = {
           Type = "dbus";
           ExecStart = "${binPath} ${
-              optionalString (cfg.storePath != "") "--path ${cfg.storePath}"
+              optionalString (cfg.storePath != null) "--path ${cfg.storePath}"
             }";
           BusName = busName;
         };
--- a/modules/services/pass-secret-service.nix
+++ b/modules/services/pass-secret-service.nix
@@ -20,7 +20,10 @@ in {
       defaultText = "$HOME/.password-store";
       example = "/home/user/.local/share/password-store";
       description = ''
-        Absolute path to password store.
+      Absolute path to password store. Defaults to
+      <filename>$HOME/.password-store</filename> if the
+      <option>programs.password-store</option> module is not enabled, and
+      <option>programs.password-store.PASSWORD_STORE_DIR</option> if it is.
       '';
     };
   };
--- a/modules/services/pass-secret-service.nix
+++ b/modules/services/pass-secret-service.nix
@@ -5,8 +5,7 @@ with lib;
 let
   cfg = config.services.pass-secret-service;
 
-  serviceArgs =
-    optionalString (cfg.storePath != null) "--path ${cfg.storePath}";
+  busName = "org.freedesktop.secrets";
 in {
   meta.maintainers = with maintainers; [ cab404 cyntheticfox ];
 
@@ -16,11 +15,13 @@ in {
     package = mkPackageOption pkgs "pass-secret-service" { };
 
     storePath = mkOption {
-      type = with types; nullOr str;
-      default = null;
-      defaultText = "~/.password-store";
+      type = types.str;
+      default = "";
+      defaultText = "$HOME/.password-store";
       example = "/home/user/.local/share/password-store";
-      description = "Absolute path to password store.";
+      description = ''
+        Absolute path to password store.
+      '';
     };
   };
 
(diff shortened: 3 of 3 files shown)


modules/services/pass-secret-service.nix
```diff
@@ -16,33 +15,48 @@ in {
...
     };
   };
 
   config = mkIf cfg.enable {
     assertions = [
       (hm.assertions.assertPlatform "services.pass-secret-service" pkgs
         platforms.linux)
+      {
+        assertion = !config.services.gnome-keyring.enable;
+        message = "Only one secrets service per user can be enabled at a time";
```
ncfavier, line 34: Maybe mention the conflicting modules

modules/services/pass-secret-service.nix
```diff
@@ -16,33 +15,48 @@ in {
     package = mkPackageOption pkgs "pass-secret-service" { };
 
     storePath = mkOption {
-      type = with types; nullOr str;
-      default = null;
```
ncfavier, line 20: Why the change in type?

modules/services/pass-secret-service.nix
```diff
@@ -16,33 +15,48 @@ in {
...
-      defaultText = "~/.password-store";
+      type = types.str;
+      default = "";
+      defaultText = "$HOME/.password-store";
```
ncfavier, line 20: `defaultText` is for reflecting the option's `default` value. If the software uses its own default when the option is `null` or empty, this should be documented in the `description`, or we should just `default` to it to avoid the indirection.

The actual default is also determined by the [password-store module](https://github.com/nix-community/home-manager/blob/669669fcb403e3137dfe599bbcc26e60502c3543/modules/programs/password-store.nix#L62-L63), so maybe this would be best explained with a `literalDocBook`.

modules/services/pass-secret-service.nix
cyntheticfox: Yeah, I forgot to. I went and double-checked there weren't other services with modules supporting `org.freedesktop.secrets`, but forgot to go back and list them. I'll do so.

modules/services/pass-secret-service.nix
cyntheticfox: Yeah, I never could decide how best to set the default. I want it to be simple where if they don't define it, it uses the application default, but doesn't forcibly couple itself to the `password-store` module if the person doesn't want to either use `password-store` or wants to have a separate store for the `secrets` service. For example, I don't actually use `password-store` currently outside of the secrets service.

Additionally, while `pass` itself uses the `PASSWORD_STORE_DIR` environment variable, `pass-secret-service` will not respect it, as the underlying project only checks the environment for that variable if you use the CLI (see <https://github.com/aviau/python-pass/blob/7b633e0034ad409872d45f09e965ffbe9ebb9700/pypass/passwordstore.py#L61>), and instead defaults to `$HOME/.password-store`. Previous behavior just configured it with the `password-store` module, which is why we added the `mkDefault` choice.

At the same time, it doesn't make sense to define a `null` choice separately from the empty string `""` IMO, since both would indicate "no path given", and the latter might actually cause things to break.

See previous discussion in <https://github.com/nix-community/home-manager/pull/2969#discussion_r877575663>.

modules/services/pass-secret-service.nix
cyntheticfox: See below comment

modules/services/pass-secret-service.nix
ncfavier: > At the same time, it doesn't make sense to define a `null` choice separately from the empty string `""` IMO, since both would indicate "no path given", and the latter might actually cause things to break.

What do you mean? `null` is the standard "no value given" value. The fact that the empty string doesn't make sense as a path (hell, sometimes it does) is accidental.

ncfavier: > but doesn't forcibly couple itself to the password-store module if the person doesn't want to either use password-store or wants to have a separate store for the secrets service.

It's a default value. Most people will use a single password store, so it's a good default. Those who don't can change it.

modules/services/pass-secret-service.nix
cyntheticfox: Added in b1bfb4dd

modules/services/pass-secret-service.nix
cyntheticfox: > `null` is the standard "no value given" value.

Really? There are quite a few other modules that contradict that. A good many modules will use an empty list `[ ]`, an empty attrset `{ }` or the empty string `""` rather than `null`, including ones like `kakoune`, `gpg`, and `bspwm`. Heck, Nix even has an entire challenge around attribute paths that may or may not exist, but those don't resolve to `null` when accessed, they throw an error. I'd argue since `""` likely doesn't work due to the provision of a path flag in this context, `""` makes for an adequate "undefined" value, while making the system simpler.

cyntheticfox: > > but doesn't forcibly couple itself to the password-store module if the person doesn't want to either use password-store or wants to have a separate store for the secrets service.
> 
> It's a default value. Most people will use a single password store, so it's a good default. Those who don't can change it.

So here's another challenge if we do that: you now _have_ to enable the separate password store, in order for the default configuration to be evaluated. I don't use `password-store` for literally anything that isn't the secrets-service, and while I don't consider myself "most people", it seems awfully bold to presume what most people will do in any case. In any case, that's at best a reason to leave the defaults for both the same value separately, not to couple them together -- and "those who don't can change it"

ncfavier: > Really?

In Nix, yes. In languages with a better type system, you would express this type as `Maybe String` or `Option String`, and the default value would be `Nothing` or `None`. `null` is a cheap approximation of that.

In types that have merging semantics, like `lines`, `listOf t` or `attrsOf t`, it makes sense for the default value to be the [neutral element](https://en.wikipedia.org/wiki/Identity_element) of the merging operation (`""`, `[]` and `{}` respectively). This isn't the case here since we can't combine two different paths.

Using `""` as a "default" or "unset" string value is common in shells, for lack of an alternative: everything is a string. Nix has at least an ersatz of a decent type system, let's use it.

ncfavier: > do that

Wait, do what? I'm not sure what we're talking about; I'm not suggesting any changes to the modules, only to the documentation.

ncfavier: Just revert the type as it was.

modules/services/pass-secret-service.nix
cyntheticfox: Reverted in [93d20b1](https://github.com/nix-community/home-manager/pull/3953/commits/93d20b13a018f3fdb5c8e37102b09552c31432b1)

modules/services/pass-secret-service.nix
cyntheticfox: Should be good with 851df4f, but still waiting on CI to make sure

ncfavier:
The assertion in pass-secret-service is broken in at least two ways. Keep it simple please, this will do:

```nix
      {
        assertion = !config.services.gnome-keyring.enable;
        message = ''
          Only one secrets service per user can be enabled at a time.
          Other services enabled:
          - gnome-keyring
        '';
      }
```

ncfavier (approved):

modules/services/gnome-keyring.nix
```diff
@@ -28,6 +28,14 @@ in {
     assertions = [
       (lib.hm.assertions.assertPlatform "services.gnome-keyring" pkgs
         lib.platforms.linux)
+      {
+        assertion = !config.services.pass-secret-store.enable;
```
matejc, line 32: pass-secret-store does not exist, should be pass-secret-service?

arjan-s: Looks like it. This merged PR causes the following error on master currently:
```
error: attribute 'pass-secret-store' missing

       at /nix/store/bn8h6s75bca34kilcmxgx9kpwz9d4xji-source/modules/services/gnome-keyring.nix:32:22:

           31|       {
           32|         assertion = !config.services.pass-secret-store.enable;
             |                      ^
           33|         message = ''
```

ncfavier: *grumble* how did the tests not catch this

ncfavier: Oh we don't have a test for gnome-keyring. That would be why.

.github/CODEOWNERS
```diff
@@ -489,7 +489,8 @@ Makefile                                              @thiagokokada
 
 /modules/services/parcellite.nix                      @gleber
 
-/modules/services/pass-secret-service.nix             @cab404
+/modules/services/pass-secret-service.nix             @cab404 @cyntheticfox
+/tests/modules/services/pass-secret-service.nix       @cyntheticfox
```
rycee, line 493: What file is this?

``` console
$ ls tests/modules/services/pass-secret-service.nix
No such file or directory
```

rycee: Maybe you meant

    /tests/modules/services/pass-secret-service       @cyntheticfox

PR #3954, opened 2023-05-06
remmina: Init module

cyntheticfox:
### Description

Create a module to enable managing Remmina, an RDP client, with a home-manager module, providing a systemd service and mimetype integration that can be disabled if so desired.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/remmina.nix
+++ b/modules/services/remmina.nix
@@ -1,0 +1,74 @@
+{ config, lib, pkgs, ... }:
+
+let
+
+  inherit (lib) mkIf mkMerge mkEnableOption mkPackageOption mkOption;
+
+  cfg = config.services.remmina;
+
+in {
+  meta.maintainers = with lib.maintainers; [ cyntheticfox ];
+
+  options.services.remmina = {
+    enable = mkEnableOption "Remmina";
+
+    package = mkPackageOption pkgs "remmina" { };
+
+    addRdpMimeTypeAssoc = mkEnableOption "Remmina RDP file open option" // {
+      default = true;
+    };
+
+    systemdService = {
+      enable = mkEnableOption "systemd Remmina service" // { default = true; };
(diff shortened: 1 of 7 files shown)


modules/services/remmina.nix
```diff
@@ -0,0 +1,72 @@
...
+        Unit = {
+          Description = "Remmina remote desktop client";
+          Documentation = "man:remmina(1)";
+          Requires = [ "graphical-session-pre.target" ];
+        };
+
+        Service = {
+          Type = "simple";
+          ExecStart = "${lib.getExe cfg.package} ${
+              lib.concatStringsSep " " cfg.systemdService.startupFlags
```
sumnerevans, line 43:
```suggestion
              escapeShellArgs cfg.systemdService.startupFlags
```

modules/services/remmina.nix
```diff
@@ -0,0 +1,72 @@
...
+            }";
+          Restart = "on-failure";
+        };
+
+        Install.WantedBy = [ "graphical-session.target" ];
+      };
+
```
sumnerevans, line 50: super nit:
```suggestion
```

sumnerevans (changes requested):

modules/services/remmina.nix
cyntheticfox: Does it work properly with `escapeShellArgs`? I've had a lot of issues getting that to work thanks to how Systemd parses it

cyntheticfox: replaced in d371fc9

sumnerevans (approved):

thibaultmol:
I was just looking up if someone had made a module for Remmina.
Thanks for your work! hope it gets merged

rycee:

PR #3969, opened 2023-05-08
home-manager: take extraSpecialArgs.lib into consideration

msteen:
### Description

I have an extended lib that I pass through my configurations, including Home Manager configurations, via the special arguments. It being the only place to safely override the default lib, considering that you might use lib in defining imports. However, since Home Manager also extends lib, these extensions get lost due to my special argument lib overriding the default Home Manager extended lib. I made a few changes to make sure the special argument lib is favored and extended instead of the one passed as an argument, such that Home Manager extensions to lib don't get lost.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -1,6 +1,6 @@
 { configuration
 , pkgs
-, lib ? pkgs.lib
+, lib ? extraSpecialArgs.lib or pkgs.lib
 
   # Whether to check that each option has a matching declaration.
 , check ? true
@@ -21,7 +21,7 @@ let
     in
       fold f res res.config.warnings;
 
-  extendedLib = import ./lib/stdlib-extended.nix lib;
+  extendedLib = import ./lib/stdlib-extended.nix extraSpecialArgs.lib or lib;
 
   hmModules =
     import ./modules.nix {
@@ -33,7 +33,9 @@ let
     modules = [ configuration ] ++ hmModules;
     specialArgs = {
       modulesPath = builtins.toString ./.;
-    } // extraSpecialArgs;
+    } // extraSpecialArgs // optionalAttrs (extraSpecialArgs ? lib) {
+      lib = extendedLib;
+    };
   };
 
   module = showWarnings (


modules/default.nix
```diff
@@ -1,6 +1,6 @@
 { configuration
 , pkgs
-, lib ? pkgs.lib
+, lib ? extraSpecialArgs.lib or pkgs.lib
```
ncfavier, line 3: Why is this needed? Can't you just pass `lib` directly as an argument to this file?

modules/default.nix
```diff
@@ -21,7 +21,7 @@ let
     in
       fold f res res.config.warnings;
 
-  extendedLib = import ./lib/stdlib-extended.nix lib;
+  extendedLib = import ./lib/stdlib-extended.nix extraSpecialArgs.lib or lib;
```
ncfavier, line 24: This seems redundant with the line above, since `lib` is already equal to `extraSpecialArgs.lib or lib`.

modules/default.nix
```diff
@@ -33,7 +33,9 @@ let
     modules = [ configuration ] ++ hmModules;
     specialArgs = {
       modulesPath = builtins.toString ./.;
-    } // extraSpecialArgs;
+    } // extraSpecialArgs // optionalAttrs (extraSpecialArgs ? lib) {
+      lib = extendedLib;
+    };
```
ncfavier, line 38: I'm also not sure why this is needed: the module system should take care of passing the correct `lib` as a "special" argument to every module.

ncfavier (commented):
It would help if you could share a minimal test case: a file showing the behaviour you want that isn't happening right now.

msteen:
Pity my description apparently wasn't enough to make you understand the problem. Let me rephrase it:

**Bug:** Passing `lib` through `extraSpecialArgs` breaks HM. This is unexpected behavior, and thus I, at least, consider it a bug.
Example:
```nix
inputs.home-manager.lib.homeManagerConfiguration {
  inherit pkgs;
  extraSpecialArgs = specialArgs; # contains an attribute named lib
  modules = homeModules;
  check = true;
}
```

Are there workarounds? Yes. Just make sure `extraSpecialArgs` don't contain `lib` and pass it through the optional `lib` argument. As in my setup `specialArgs` is reused across various configuration builds, I have to work around it like this:
```nix
inputs.home-manager.lib.homeManagerConfiguration {
  inherit pkgs;
  inherit (specialArgs) lib;
  extraSpecialArgs = removeAttrs specialArgs [ "lib" ]; # contains an attribute named lib
  modules = homeModules;
  check = true;
}
```

modules/default.nix
msteen: See my response to your issue comment.

modules/default.nix
msteen: Only if the default were used. Not if another lib was explicitly passed as an argument. Yet the lib in the `extraSpecialArgs` would overwrite it in the end. To prevent having to extend it again on line 37, I did this.

modules/default.nix
msteen: Redundant question, comes down to the same question in your issue comment, thus see my response to that.

martin-braun:
This is what I tried to do, intentionally. I've set `home-manager.extraSpecialArgs = specialArgs;` in my home-server configuration entry that gets called from the flake entry and suddenly you get weird issues, like that it can't find attribute `program.mako` or `targets`. It's good to know that I can bypass it by only setting new stuff to `home-manager.extraSpecialArgs`, but I can totally see that it would be useful to favor a custom incoming `lib`, although I still don't have this requirement. Could it be also useful for `config` and `pkgs`?

Issue #3971, opened 2023-05-09
bug: borgmatic conflicting values


cyntheticfox:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

On master, the `borgmatic-program-exclude-hm-symlinks` test fails with:

```txt
error: The option `programs.borgmatic.backups.main.location.extraConfig.exclude_from' has conflicting definition values:
       - In `/home/cynthia/repos/home-manager/tests/modules/programs/borgmatic/exclude-hm-symlinks.nix':
           [
             "/nix/store/pmq27zqfly93wvjgy6y5n67f457c0gb3-excludeFile.txt"
           ]
       - In `/home/cynthia/repos/home-manager/modules/programs/borgmatic.nix':
           [
             "/nix/store/va42wvgk47rf1bac3p7gly00pybv0jy8-hm-symlinks.txt"
           ]
(use '--show-trace' to show detailed location information)
```

Run via `nix-shell --pure tests -A run.borgmatic-program-exclude-hm-symlinks`

### Maintainer CC

@DamienCassou

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.2.13, NixOS, 22.11 (Raccoon), 22.11.20230504.cc45a3f`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.11.1`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`
```

cyntheticfox:
it's this exact error from <https://github.com/nix-community/home-manager/pull/3787#discussion_r1142665947>, but surely the `home-manager` test system doesn't use your own system packages, right?

ncfavier:
The `nix-shell` interface just imports `<nixpkgs>`, so you can set that with `-I nixpkgs=` or use `nix develop .#all`.

PR #3905, opened 2023-04-22
flake: Move tests back to devShells.*.tests

YorikSar:
### Description

I assume it was erroneously lost during flake-utils removal in https://github.com/nix-community/home-manager/pull/3860

### Checklist

- [x] Change is backwards compatible.

- [x] Code formatted with `./format`.

- [x] Code tested through `nix-shell --pure tests -A run.all`.

- [ ] ~Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).~ N/A

- [x] Commit messages are formatted like

    ```
    {component}: {description}

    {long description}
    ```

    See [CONTRIBUTING](https://github.com/nix-community/home-manager/blob/master/docs/contributing.adoc#sec-commit-style) for more information and [recent commit messages](https://github.com/nix-community/home-manager/commits/master) for examples.

PR #3082, opened 2022-07-10
flake.nix: provide checks (tests) and docs and related fixes (cont'd)

ShamrockLee:
Continuation of #3024, which is accidentally closed after a failed rebase.

### Description

This patch provides the Home Manager Module test cases from flake through `outputs.checks.${system}` and `outputs.legacyPackages.${system}`. This allows developers to run the tests against the master branch without having to update their environment.

If applied, developers can run the tests by
```
nix develop --override-input nixpkgs github:NixOS/nixpkgs/nixos-unstable .#tests.run.all
```
~or simply~
```
nix flake check --override-input nixpkgs github:NixOS/nixpkgs/nixos-unstable .
```

Update: As reported by https://github.com/nix-community/home-manager/pull/3024#issuecomment-1156281495 , the multi-platform failure would revive on 22.05 even after the IFD fix. We should probably add the test suite into the `checks` until https://github.com/NixOS/nix/issues/4265 got fixed.

(`nix develop` is the Flake analogy of `nix-shell --pure -A`. The `--override-input` above makes Nix build against the Nixpkgs nixos-unstable branch instead of the master branch to avoid rebuilding the world.)

Fixes #2722 when running the tests for flake users.

This patch also contains two fixes found when testing the above features:
*   Fix impurities when evaluating `tests` introduced in `modules/modules.nix` by the appended element `pkgsModule` that injects `_module.args.pkgsPath` as `<nixpkgs>` .
*   Get `pkgs` through `inputs.nixpkgs.legacyPackages.${system}` whenever available instead of  `import nixpkgs { inherit "${system}"; }`
    *    ~This avoids IFD where possible, working around https://github.com/NixOS/nix/issues/4265 on platforms covered by `eachDefaultSystem`.~ Update: not on NixOS 22.05.
    *     This prevents re-evaluation of Nixpkgs. See [1000 instances of nixpkgs](https://discourse.nixos.org/t/1000-instances-of-nixpkgs/17347) for detail.

(I don't know why the last sub-bullet is rendered as a code block. It's probably a GitHub bug.

Regarding #3056:
This PR does not mean to revert the decisions made in the PR made by rycee. Rycee's PR cleared up the hack to use flake.lock content on non-flake system, with the side effect of removing the flake support to access the documents. This PR adds back the flake-based document support plus the tests support, while paying respect to the non-flake infrastructure.

### Checklist

- [X] Change is backwards compatible. (I think so, but the impurity fix requires experienced people to double check.)

[...]

PR #3972, opened 2023-05-09
PR_TEMPLATE: Note nix3 test method in checklist

cyntheticfox:
### Description

Allows for running with the `nixpkgs` from the lock file, as well as using a more familiar interface for users of the experimental CLI, which would've avoided me opening #3971.

Also updates the corresponding note in `docs/contributing.adoc` to have the correct invocation path.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -23,7 +23,7 @@ Also make sure to read the guidelines found at
 
 - [ ] Code formatted with `./format`.
 
-- [ ] Code tested through `nix-shell --pure tests -A run.all`.
+- [ ] Code tested through `nix-shell --pure tests -A run.all` or `nix develop --ignore-environment .#all` using the experimental CLI.
 
 - [ ] Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).
 
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -23,7 +23,7 @@ Also make sure to read the guidelines found at
 
 - [ ] Code formatted with `./format`.
 
-- [ ] Code tested through `nix-shell --pure tests -A run.all` or `nix develop --ignore-environment .#all` using the experimental CLI.
+- [ ] Code tested through `nix-shell --pure tests -A run.all` or `nix develop --ignore-environment .#all` using Flakes.
 
 - [ ] Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).
 
--- a/docs/contributing.adoc
+++ b/docs/contributing.adoc
@@ -265,4 +265,4 @@ $ nix-shell --pure tests -A run.alacritty-empty-settings
 However, those invocations will impurely source the system’s nixpkgs, and may cause failures. To run against the nixpkgs from the flake.lock, use instead e.g.
 
 [source,console]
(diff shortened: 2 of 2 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -23,7 +23,7 @@ Also make sure to read the guidelines found at
 
 - [ ] Code formatted with `./format`.
 
-- [ ] Code tested through `nix-shell --pure tests -A run.all`.
+- [ ] Code tested through `nix-shell --pure tests -A run.all` or `nix develop --ignore-environment .#all` using the experimental CLI.
```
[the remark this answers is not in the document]
ncfavier: Yes, but Flakes are likely to remain experimental for longer than the new CLI, and you have to use the new CLI to use Flakes anyway.

cyntheticfox:
Should I edit the other command to be something like

```shell
nix-shell -I "nixpkgs=https://github.com/nixos/nixpkgs/archive/$(nix-shell -p 'jq' --cmd "jq -r '.nodes.nixpkgs.locked.rev' flake.lock").tar.gz" --pure tests -A run.all
```

to let standard users use the locked version?

.github/PULL_REQUEST_TEMPLATE.md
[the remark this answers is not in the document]
cyntheticfox: Fixed in a71f8d7

ncfavier:
That's a bit verbose, I'd rather use something like [flake-compat](https://github.com/edolstra/flake-compat) to use the locked inputs in a way that's transparent to the user. In fact I have done that before, but it was undone in https://github.com/nix-community/home-manager/pull/3056/files#diff-1cc580de297308d93d82f7b72446ae4b98832a8aae3378e9e134102519a0e33a, so you'll have to convince @rycee.

cyntheticfox:
I see... well, either way, this pr is probably good, though I suppose I could list that big command in the `docs/contributing.adoc`

docs/contributing.adoc
```diff
@@ -265,4 +265,4 @@ $ nix-shell --pure tests -A run.alacritty-empty-settings
 However, those invocations will impurely source the system’s nixpkgs, and may cause failures. To run against the nixpkgs from the flake.lock, use instead e.g.
 
 [source,console]
-$ nix develop --ignore-environment .#tests.all
+$ nix develop --ignore-environment .#all
```
ncfavier, line 268: I'm actually not a fan of this lack of namespacing. I think it would make more sense to resuscitate https://github.com/nix-community/home-manager/pull/3082 and expose the tests as `legacyPackages` instead of `packages`, so you could do `nix develop .#tests.run.all` (we can't do that with `packages` without angering `nix flake check`, cf. https://github.com/nix-community/home-manager/pull/3905).

cyntheticfox: Is there a particular reason we use `devShells` instead of `checks`?

ncfavier: I think the problem is that `checks` expect to be built, while our test suite (based on [nmt](https://git.sr.ht/~rycee/nmt)) must be run in a shell (hence `nix develop`).

cyntheticfox: ...aren't most common Nix builders written to use shells? `stdenv.mkDerivation` is literally a Nix framework to write a shell-script builder, which is then invoked as the build function.

TBH, I don't even see anything in `nmt` that would likely be problematic -- just how the test framework creates dummy packages, but that's in `/tests/stubs.nix`, not `nmt` itself, where pulling out that function into a separate top-level function (whether under `lib` in `flake.nix` or as itself in `default.nix`

ncfavier: Yeah, disregard that. We can actually just use `tests.build.all` once the space leak is fixed (https://gitlab.com/rycee/nmt/-/merge_requests/6)

cyntheticfox: In that case, is this good to merge as-is for now?

PR #3974, opened 2023-05-10
systemd: accept derivations as values for systemd options

R-VdP:
### Description

Allow for values of type derivation to be used as values for systemd options. This allows for instance to assign the result of `pkgs.writeShellScript` to `ExecStart` directly, without needing to call `toString` every time.
It is also closer to what people might be used to from NixOS.

~~Note: I don't really like the formatting of having one `attrsOf` on a separate line, but the auto-formatter seems to insist on doing it that way...~~

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/systemd.nix
+++ b/modules/systemd.nix
@@ -51,7 +51,7 @@ let
 
   unitType = unitKind:
     with types;
-    let primitive = either bool (either int str);
+    let primitive = oneOf [ bool int str path ];
     in attrsOf (attrsOf (attrsOf (either primitive (listOf primitive)))) // {
       description = "systemd ${unitKind} unit configuration";
     };
--- a/tests/modules/systemd/services.nix
+++ b/tests/modules/systemd/services.nix
@@ -1,14 +1,17 @@
-{ config, lib, pkgs, ... }:
+{ pkgs, ... }:
 
-with lib;
-
-{
+let
+  drvScript = pkgs.writeShellScript "drv-script.sh" ''
+    echo "Just a test"
+  '';
+in {
   config = {
     systemd.user.services."test-service@" = {
       Unit = { Description = "A basic test service"; };
 
       Service = {
         Environment = [ "VAR1=1" "VAR2=2" ];
+        ExecStartPre = drvScript;
         ExecStart = ''/some/exec/start/command --with-arguments "%i"'';
       };
     };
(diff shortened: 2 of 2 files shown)


modules/systemd.nix
```diff
@@ -52,7 +52,8 @@ let
   unitType = unitKind:
     with types;
     let primitive = either bool (either int str);
-    in attrsOf (attrsOf (attrsOf (either primitive (listOf primitive)))) // {
+    in attrsOf
+    (attrsOf (attrsOf (oneOf [ primitive package (listOf primitive) ]))) // {
```
ncfavier, line 56: I don't like `package` and don't think it's needed here. What I'd do is simply make `primitive = oneOf [ bool int str path ]` and leave this line unchanged.

ncfavier (changes requested):

modules/systemd.nix
R-VdP: Fair enough!
I actually didn't notice that primitive was defined just above.

ncfavier:
Could you add a test for this in `tests/modules/systemd/services.nix`?

R-VdP:
> Could you add a test for this in `tests/modules/systemd/services.nix`?

Yup, done!

ncfavier (approved):

tests/modules/systemd/services.nix
```diff
@@ -23,6 +26,9 @@ with lib;
             Environment=VAR1=1
             Environment=VAR2=2
             ExecStart=/some/exec/start/command --with-arguments "%i"
+            ExecStartPre=${
+              builtins.unsafeDiscardStringContext drvScript.outPath
```
ambroisie, line 30: Is `unsafeDiscardStringContext` necessary here?

I'm not well versed in the dark arts of those low-level operations, if you have an explanation for why you need it I'd gladly hear it.

ncfavier: Could probably use `pkgs.writeText` instead of `builtins.toFile`, that way you wouldn't need it.

ambroisie: The script is written using `pkgs.writeShellScript` which used `pkgs.writeTextFile` (just like `pkgs.writeText`) unless I am mistaken.

R-VdP: Good point, it's actually nicer to use `pkgs.writeText` indeed, I made the change now.

@ambroisie `builtins.toFile` was used to create the file with the expected output, so that's the one who was complaining about a store path (e.g. a string with context, in nix parlance) being included in its output, nix forbids this in general.

tests/modules/systemd/services.nix
```diff
@@ -18,11 +21,12 @@ with lib;
       assertFileExists $serviceFile
       assertFileContent $serviceFile \
         ${
-          builtins.toFile "services-expected.conf" ''
+          pkgs.writeText "services-expected.conf" ''
             [Service]
             Environment=VAR1=1
             Environment=VAR2=2
             ExecStart=/some/exec/start/command --with-arguments "%i"
+            ExecStartPre=${drvScript.outPath}
```
ncfavier, line 29:
```suggestion
            ExecStartPre=${drvScript}
```

PR #3975, opened 2023-05-10
starship: add enable_transience for fish

will:
### Description

Starship has an advanced, experimental feature where fancy stuff in the prompt can be replaced with something more simple after the command is ran. This is very helpful for copy and pasting shell history somewhere else.

image

docs: https://starship.rs/advanced-config/#transientprompt-and-transientrightprompt-in-fish

Fish is currently the only shell as far as I can tell that both home-manager and starship support for this feature. Since the function has to be called after starship is loaded, this seems like the best place to put it. Without putting it inside home-manager itself, you would have to forego the provided `programs.fish.interactiveShellInit` entirely just so you can be sure to only call the new function after starship is loaded, which seems like a shame.

### Checklist

I did run just the starship-settings test, but there weren't any tests for the shell init hooks.

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/starship.nix
+++ b/modules/programs/starship.nix
@@ -77,6 +77,19 @@ in {
     enableNushellIntegration = mkEnableOption "Nushell integration" // {
       default = true;
     };
+
+    enableTransience = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        The TransientPrompt feature of Starship replaces previous prompts with a
+        custom string. This is only a valid option for the Fish shell.
+
+        For documentation on how to change the default replacement string and
+        for more information visit
+        https://starship.rs/advanced-config/#transientprompt-and-transientrightprompt-in-cmd
+      '';
+    };
   };
 
   config = mkIf cfg.enable {
@@ -101,6 +114,7 @@ in {
     programs.fish.interactiveShellInit = mkIf cfg.enableFishIntegration ''
       if test "$TERM" != "dumb"  -a \( -z "$INSIDE_EMACS"  -o "$INSIDE_EMACS" = "vterm" \)
         eval (${starshipCmd} init fish)
+        ${lib.optionalString cfg.enableTransience "enable_transience"}
       end
     '';
 
--- a/tests/modules/programs/starship/default.nix
+++ b/tests/modules/programs/starship/default.nix
@@ -1,1 +1,5 @@
-{ starship-settings = ./settings.nix; }
+{
+  starship-settings = ./settings.nix;
+  starship-fish-with-transience = ./fish_with_transience.nix;
+  starship-fish-without-transience = ./fish_without_transience.nix;
+}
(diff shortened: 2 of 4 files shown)


caarlos0 (approved):

modules/programs/starship.nix
```diff
@@ -101,6 +103,7 @@ in {
     programs.fish.interactiveShellInit = mkIf cfg.enableFishIntegration ''
       if test "$TERM" != "dumb"  -a \( -z "$INSIDE_EMACS"  -o "$INSIDE_EMACS" = "vterm" \)
         eval (${starshipCmd} init fish)
+        ${lib.optionals cfg.enableTransience "enable_transience"}
```
ncfavier, line 106: Have you tested this? I'd expect it to fail since `optionals` produces a list and you're interpolating a string...

Please add a test

modules/programs/starship.nix
```diff
@@ -77,6 +77,8 @@ in {
     enableNushellIntegration = mkEnableOption "Nushell integration" // {
       default = true;
     };
+
+    enableTransience = mkEnableOption "TransientPrompt" // { default = false; };
```
ncfavier, line 81: This should link to the docs and make it clear that only fish supports this.

ncfavier (changes requested):

modules/programs/starship.nix
will: > Have you tested this? 

Yes, I'm running currently off of my branch, and the screenshot in the PR description is from using it. The bottom half where it has just `>` instead of the colors is the feature. 

```nix
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
    stable.url = "github:NixOS/nixpkgs/22.11";
    #home-manager = { url = "github:nix-community/home-manager"; inputs.nixpkgs.follows = "nixpkgs"; };
    home-manager = { url = "github:will/home-manager/starship-transience"; inputs.nixpkgs.follows = "nixpkgs"; };
```

However I'll switch the conditional to a more suitable one.

> Please add a test

I'm all for adding tests, but haven't yet written any tests in nix. And the other related tests for starship only have tests for the generated toml file, and not for the hooks it puts into shells. Do you have an example of another project that has a test for the `interactiveShellInit` hooks that I can learn from?

will:
Thanks for the review @ncfavier. For updates to the PR, do you prefer new fixup commits to this, or force pushing over the existing commit, or something else?

modules/programs/starship.nix
ncfavier: Right, it only fails when `enableTransience` is `false` since `optionals false ...` yields `[]`. Use `optionalString` instead.

will:
Ok @ncfavier, I've addressed both of your feedback. I've used `optionalString` properly and added tests for both fish with and without transience enabled, and I've added documentation that it's fish only with the url linking to the starship docs. 

I think that's everything, please let me know if there are other issues though

ncfavier (approved):

PR #3976, opened 2023-05-11
i3-sway: fix indentation

hellodword:
### Description

Currently the generated config file will be:
```
bar {
  position top
  status_command i3status
i3bar_command /nix/store/dyxib7xhhplkbvz07vw6mhc6islhgz2s-i3-4.21.1/bin/i3bar
}
```

What we need:
```
bar {
  position top
  status_command i3status
  i3bar_command /nix/store/dyxib7xhhplkbvz07vw6mhc6islhgz2s-i3-4.21.1/bin/i3bar
}
```

### DIFF
--- a/modules/services/window-managers/i3-sway/lib/functions.nix
+++ b/modules/services/window-managers/i3-sway/lib/functions.nix
@@ -85,7 +85,7 @@ rec {
               (optionalString (position != null) "position ${position}")
               (optionalString (statusCommand != null)
                 "status_command ${statusCommand}")
-              "${moduleName}bar_command ${command}"
+              "  ${moduleName}bar_command ${command}"
               (optionalString (workspaceButtons != null)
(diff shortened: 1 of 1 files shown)


modules/services/window-managers/i3-sway/lib/functions.nix
```diff
@@ -85,7 +85,7 @@ rec {
               (optionalString (position != null) "position ${position}")
               (optionalString (statusCommand != null)
                 "status_command ${statusCommand}")
-              "${moduleName}bar_command ${command}"
+                "${moduleName}bar_command ${command}"
```
ncfavier, line 88: Have you actually tested this? It looks like these spaces are in a Nix expression context, so they will just be ignored.

Do you actually need this for the configuration to be valid or is it just cosmetic?

ncfavier:
Proposed a proper fix for this in https://github.com/nix-community/home-manager/pull/3978

PR #3977, opened 2023-05-11
river: add module

GaetanLepage:
### Description

Add the `wayland.windowManager.river` module to configure the [`river` wayland compositor](https://github.com/riverwm/river)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/window-managers/river.nix
+++ b/modules/services/window-managers/river.nix
@@ -1,0 +1,212 @@
+{ config, lib, pkgs, ... }:
+
+let
+  inherit (lib) types;
+  cfg = config.wayland.windowManager.river;
+
+  # Systemd integration
+  variables = builtins.concatStringsSep " " cfg.systemd.variables;
+  extraCommands = builtins.concatStringsSep " "
+    (map (f: "&& ${f}") cfg.systemd.extraCommands);
+  systemdActivation = ''
+    exec "${pkgs.dbus}/bin/dbus-update-activation-environment --systemd ${variables} ${extraCommands}"
+  '';
+
+  toValue = val:
+    if lib.isString val || lib.isDerivation val then
+      toString val
+    else if true == val then
+      "enabled"
+    else if false == val then
+      "disabled"
+    else if lib.isInt val then
+      toString val
+    else if lib.isFloat val then
+      lib.strings.floatToString val
+    else
+      abort "unsupported type ${builtins.typeOf val}";
+
+  # Intermediary function that converts some value (attrs, str, ...) to one or several commands.
+  toArgs = path: value:
+
+    let
+      stringValue = lib.concatStringsSep " " (path ++ [ (toValue value) ]);
+      finalValue = if lib.isAttrs value then
+        toCommand path value
+      else if lib.isList value then
+        lib.lists.flatten (map (x: toArgs path x) value)
+      else if value == null then
+        [ ]
+      else
+        [ stringValue ];
+    in finalValue;
+
+  # toCommand :: [string] -> attrs -> [string]
+  # Recursive function that converts an attrs to a list of commands that can be written to the
+  # config file.
+  toCommand = basePath: attrs:
+    lib.concatLists (lib.mapAttrsToList
+      (key: value: let path = basePath ++ [ key ]; in toArgs path value) attrs);
+
+in {
+  meta.maintainers = [ lib.maintainers.GaetanLepage ];
+
+  options.wayland.windowManager.river = {
+    enable = lib.mkEnableOption "the river window manager";
+
+    package = lib.mkPackageOption pkgs "river" {
+      nullable = true;
+      extraDescription = ''
+        Set to `null` to not add any river package to your path.
+        This should be done if you want to use the NixOS river module to install river.
+      '';
+    };
+
+    xwayland.enable = lib.mkEnableOption "XWayland" // { default = true; };
+
+    systemd = {
+      enable = lib.mkEnableOption null // {
+        default = true;
+        description = ''
+          Whether to enable {file}`river-session.target` on
+          river startup. This links to {file}`graphical-session.target`}.
+          Some important environment variables will be imported to systemd
+          and D-Bus user environment before reaching the target, including
+          - `DISPLAY`
+          - `WAYLAND_DISPLAY`
+          - `XDG_CURRENT_DESKTOP`
+          - `NIXOS_OZONE_WL`
+          - `XCURSOR_THEME`
+          - `XCURSOR_SIZE`
+        '';
+      };
+
+      variables = lib.mkOption {
+        type = types.listOf types.str;
+        default = [
+          "DISPLAY"
+          "WAYLAND_DISPLAY"
+          "XDG_CURRENT_DESKTOP"
+          "NIXOS_OZONE_WL"
+          "XCURSOR_THEME"
+          "XCURSOR_SIZE"
+        ];
+        example = [ "-all" ];
+        description = ''
+          Environment variables to be imported in the systemd & D-Bus user
+          environment.
+        '';
+      };
+
+      extraCommands = lib.mkOption {
+        type = types.listOf types.str;
+        default = [
+          "systemctl --user stop river-session.target"
+          "systemctl --user start river-session.target"
+        ];
+        description = "Extra commands to be run after D-Bus activation.";
+      };
+    };
+
+    extraSessionVariables = lib.mkOption {
+      type = types.attrs;
+      default = { };
+      description = "Extra session variables set when running the compositor.";
+      example = { MOZ_ENABLE_WAYLAND = "1"; };
+    };
+
+    settings = lib.mkOption {
+      type = let
+        valueType = with types;
+          nullOr (oneOf [
+            bool
+            int
+            float
+            str
+            path
+            (attrsOf valueType)
+            (listOf valueType)
+          ]) // {
+            description = "River configuration value";
+          };
+      in valueType;
+      default = { };
+      description = "General settings given to `riverctl`.";
+      example = {
+        border-width = 2;
+        declare-mode = [ "locked" "normal" "passthrough" ];
+        map.normal."Alt Q" = "close";
+        input.pointer-foo-bar = {
+          accel-profile = "flat";
+          events = true;
+          pointer-accel = -0.3;
+          tap = false;
+        };
+        rule-add."-app-id" = {
+          "'bar'" = "csd";
+          "'float*'"."-title"."'foo'" = "float";
+        };
+        set-cursor-warp = "on-output-change";
+        set-repeat = "50 300";
+        xcursor-theme = "someGreatTheme 12";
+        spawn = [ "firefox" "'foot -a terminal'" ];
+      };
+    };
+
+    extraConfig = lib.mkOption {
+      type = types.lines;
+      default = "";
+      example = ''
+        rivertile -view-padding 6 -outer-padding 6 &
+      '';
+      description =
+        "Extra lines appended to {file}`$XDG_CONFIG_HOME/river/init`.";
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    assertions = [
+      (lib.hm.assertions.assertPlatform "wayland.windowManager.river" pkgs
+        lib.platforms.linux)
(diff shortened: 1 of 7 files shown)


JustinLovinger:
Environment variables do not seem to be set for systemd as documentation implies:

```
 wayland.windowManager.river.systemdIntegration
     Whether to enable river-session.target on river startup. This links to graphical-session.target. Some
     important environment variables will be imported to systemd and dbus user environment before reaching the
     target, including

     •   DISPLAY

     •   WAYLAND_DISPLAY

     •   XDG_CURRENT_DESKTOP
```

berbiche:
Good work.

The module is complex and I think it could be made much simpler with a different design. 

River's config is basically a shell script where `riverctl` or `rivertile` is called with different sets of attributes.
The configuration could thus be represented as a nested attribute set where the leaves are the configuration value and the keys represent the path of the value to set.
This would remove a lot of boilerplate from the code generation and make it simpler to maintain if `riverctl` adds support for new commands (it would basically require no changes).

I've got the following script which seems to work:

``` nix
with lib;
let
  toValue = val:
    if isString val || isDerivation val then ''"${val}"'' 
    else if true == val then ''"enabled"''
    else if false == val then ''"disabled"''
    else if isInt val then toString val
    else if isFloat val then strings.floatToString val
    else if isList val then concatMapStringsSep " " toValue val
    else abort "unsupported type ${builtins.typeOf val}";

  # toCommand :: [string] -> attrs -> [string]
  toCommand = basePath: attrs:
   concatLists (mapAttrsToList (key: value:
      let
        path = basePath ++ [key];
        stringValue = concatStringsSep " " (path ++ [(toValue value)]);
        finalValue =
          if isAttrs value
          then toCommand path value
          else [stringValue];
      in finalValue
    ) attrs);
in
  toCommand
```

Here's an example with the nix expression and the expected generated commands.

```nix
toCommand ["riverctl"] {
  input.pointer-foo-bar = {
    accel-profile = "flat";
    pointer-accel = -0.3;
    events = "enabled";
    tap = "disabled";
  };
}
```
results in
```
riverctl input pointer-foo-bar accel-profile "flat"
riverctl input pointer-foo-bar pointer-accel -0.3
riverctl input pointer-foo-bar events "enabled"
riverctl input pointer-foo-bar tap "disabled"
```

Lists of primitive types would be supported but not lists of anything else. 

The caveat of the suggested syntax is that it won't support something that produces a result like this without more complex logic:

```
riverctl foo bar
riverctl foo bar baz
```

modules/services/window-managers/river/default.nix
```diff
@@ -0,0 +1,225 @@
...
+      ${inputs}
+
+      ### EXTRA CONFIGURATION ###
+      ${cfg.extraConfig}
+
+      ### STARTUP PROGRAMS ###
+      ${concatMapStringsSep "\n" formatStartupProgram cfg.startupPrograms}
+
+      ### LAYOUT ###
+      riverctl default-layout ${cfg.config.layoutGenerator.name}
```
JustinLovinger, line 204: Note, the name of the layout is not guaranteed to be equal to the filename of the executable, see <https://github.com/riverwm/river/wiki#my-windows-are-not-tiled>.

GaetanLepage: Thanks for this remark !

GaetanLepage:
Thank you very much for the feedback @berbiche and @JustinLovinger !
I am working on a rewrite and will push it once it will be ready !

modules/services/window-managers/river/default.nix
```diff
@@ -0,0 +1,225 @@
...
+      ${cfg.config.layoutGenerator.name} ${cfg.config.layoutGenerator.arguments} &
```
JustinLovinger, line 205: systemd integration should follow:

```
### SYSTEMD ###
${pkgs.dbus}/bin/dbus-update-activation-environment --systemd DISPLAY WAYLAND_DISPLAY XDG_CURRENT_DESKTOP XDG_SESSION_TYPE XDG_VTNR NIXOS_OZONE_WL
systemctl --user --no-block start river-session.target
```

with a conditional for `cfg.systemdIntegration` of course.

I based this on the Sway systemd integration, <https://github.com/nix-community/home-manager/blob/0ee5ab611dc1fbb5180bd7d88d2aeb7841a4d179/modules/services/window-managers/i3-sway/sway.nix#L320>, and tested it on my machine.

berbiche: We should probably make that bit generic across Wayland window managers in a different PR

GaetanLepage:
@JustinLovinger @berbiche, thanks to your very helpful comments, I have been able to set up a new implementation inspired by the [`toCommand` helper](https://github.com/nix-community/home-manager/pull/3977#issuecomment-1605882905) which I changed a little bit.

I will be glad to hear what you think about it now.

modules/services/window-managers/river/default.nix
GaetanLepage: I have created an option to eventually specify a name for the executable if it is not the same as the layout generator itself.

modules/services/window-managers/river/default.nix
GaetanLepage: I added the proper systemd `dbus` call.

modules/services/window-managers/river/options.nix
```diff
@@ -0,0 +1,492 @@
...
+
+        whenTyping = mkOption {
+          type = types.bool;
+          default = false;
+          description = ''
+            Hide the cursor when pressing any non-modifier key.
+            Show the cursor again on any movement.
+          '';
+        };
+      };
```
berbiche, line 254: IMO all these camelCase options are redundant.

There is no advantage having these options documented in Home Manager's options page over the official documentation.

berbiche (changes requested):
Thanks for your hard work but the module is still too complex.

I believe the design I recommended in my initial comment is still much simpler.

The set of settings could be described in less than 30 lines.

``` nix
let
  primitives = types.oneOf [ types.str types.int types.float ... ];
  settingsType = types.attrsOf (types.attrsOf primitives)
in
{
  settings = mkOption {
    type = types.submodule {
      freeformType = settingsType;
      # An option that is explicitly described and typed, if required
      options.something = ...;
    };
  };

  # Same thing I guess
  # If it can be nested under `settings` in some way it would be great
  layouts = ...;
}
```

and in the config section, you could generate the configuration file with `toCommand "riverctl" cfg.settings`.

1. Does the order of settings matter in the configuration?

If yes, you can probably just call `toCommand` per key/settings in the order they should appear in the final configuration as nix stores attribute sets ordered by the key name.

JustinLovinger:
@GaetanLepage, To add on to berbiche, take a look at RFC 42, <https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md>. The style of module you are writing is deprecated, and Nix is moving towards the style of module berbiche suggested.

GaetanLepage:
@berbiche @justinlovinger Hi !
After several months I decided to come back to this module.
Indeed, adopting the RFC42-style was a better idea.

Please, tell me if the module better suits your expectation now :)

modules/services/window-managers/river.nix
```diff
@@ -0,0 +1,213 @@
...
+            int
+            float
+            str
+            path
+            (attrsOf valueType)
+            (listOf valueType)
+          ]) // {
+            description = "River configuration value";
+          };
+        in valueType;
```
GaetanLepage, line 133: In the end, I haven't felt that it was necessary to use a `types.submodule` with `freeformType`.
Indeed, I haven't found necessary to explicitly declare any option in this case.

By the way, this is the same thing that is done in the [hyprland module](https://github.com/nix-community/home-manager/blob/3d6791b3897b526c82920a2ab5f61d71985b3cf8/modules/services/window-managers/hyprland.nix#L113-L127).

modules/services/window-managers/river.nix
```diff
@@ -0,0 +1,213 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
```
drupol, line 3: This is deprecated. Just inherit only what you need out of it.

GaetanLepage: Done :)

modules/services/window-managers/river.nix
```diff
@@ -0,0 +1,212 @@
...
+          - `WAYLAND_DISPLAY`
+          - `XDG_CURRENT_DESKTOP`
+          - `NIXOS_OZONE_WL`
+          - `XCURSOR_THEME`
+          - `XCURSOR_SIZE`
+        '';
+      };
+
+      variables = lib.mkOption {
+        type = with types; listOf str;
```
drupol, line 85: Most of those with are not needed, how about using types.listOf instead?

modules/services/window-managers/river.nix
drupol: Merci

modules/services/window-managers/river.nix
GaetanLepage: You would rather have `type = types.listOf types.str;` ?

drupol: Yeah I personally prefer avoid using with in some cases.
I have to admit that it is an opinionated feedback and I wouldn't mind if you ignore it.

GaetanLepage: I am fine with this as well.
I have only left the one for the complex `valueType` (`settings` option). Is that fine ?

drupol: Yeah I think it's cool now, thanks 👍

GaetanLepage:
Thanks for the review @drupol !

modules/services/window-managers/river.nix
```diff
@@ -0,0 +1,212 @@
...
+    };
+
+    extraConfig = lib.mkOption {
+      type = types.lines;
+      default = "";
+      example = ''
+        rivertile -view-padding 6 -outer-padding 6 &
+      '';
+      description =
+        "Extra lines appended to <filename>$XDG_CONFIG_HOME/river/init</filename>";
```
rycee, line 163:
```suggestion
        "Extra lines appended to {file}`$XDG_CONFIG_HOME/river/init`";
```

modules/services/window-managers/river.nix
```diff
@@ -0,0 +1,212 @@
...
+    };
+
+    xwayland.enable = lib.mkEnableOption "XWayland" // { default = true; };
+
+    systemd = {
+      enable = lib.mkEnableOption null // {
+        default = true;
+        description = ''
+          Whether to enable {file}`river-session.target` on
+          river startup. This links to `graphical-session.target`.
```
rycee, line 72:
```suggestion
          river startup. This links to {file}`graphical-session.target`.
```

tests/modules/services/window-managers/river/configuration.nix
```diff
@@ -0,0 +1,98 @@
...
+    test.stubs.river = { };
+
+    nmt.script = ''
+      riverInit=home-files/.config/river/init
+      assertFileExists "$riverInit"
+      assertFileIsExecutable "$riverInit"
+
+      normalizedConfig=$(normalizeStorePaths "$riverInit")
+      referenceConfig=$(normalizeStorePaths "${
+        pkgs.writeShellScript "river-init-expected" (readFile ./init)
```
rycee, line 93: Isn't the `writeShellScript` just adding the shebang? If so I think you could do
```suggestion
        ./init
```
and add the normalized shebang to the init file.

tests/modules/services/window-managers/river/configuration.nix
```diff
@@ -0,0 +1,98 @@
...
+      };
+
+      extraConfig = ''
+        rivertile -view-padding 6 -outer-padding 6 &
+        some
+        extra config
+      '';
+    };
+
+    test.stubs.river = { };
```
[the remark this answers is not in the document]
GaetanLepage: Done

modules/services/window-managers/river.nix
GaetanLepage: Done

modules/services/window-managers/river.nix
GaetanLepage: Done

rycee:

PR #3981, opened 2023-05-12
zellij: disables shell integrations by default

erikkrieg:
### Description

I added the shell integrations back in this PR: https://github.com/nix-community/home-manager/pull/3926

There was some early feedback suggesting that shell integrations shouldn't be enabled by default, so I'm opening this PR to make it more convenient for the maintainers to pivot to this options if desired.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/tests/modules/programs/zellij/enable-shells.nix
+++ b/tests/modules/programs/zellij/enable-shells.nix
@@ -3,7 +3,12 @@
 {
   programs = {
     # All shell integratons are enabled by default.
-    zellij.enable = true;
+    zellij = {
+      enable = true;
+      enableBashIntegration = true;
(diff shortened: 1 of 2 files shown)


tests/modules/programs/zellij/enable-shells.nix
```diff
@@ -3,7 +3,12 @@
 {
   programs = {
     # All shell integratons are enabled by default.
```
ncfavier, line 5:
```suggestion
```

erikkrieg: Thanks. It is still early in the morning where I'm at XD

PR #3987, opened 2023-05-14
aerc: add notmuch support

kchibisov:
This doesn't touch query-map option which could be used to create virtual mailboxes with imap. The maildir backend creates virtual mailboxes automatically.

### Checklist

- [ ] Change is backwards compatible. (It depends how you look at it, given that notmuch was ignored with aerc the generation will break for `notmuch.enable = true` and `mbsync.enable.true` setups, because it'll be notmuch now). Maybe the `aerc.notmuch` option should be added? 
 
### DIFF
--- a/modules/programs/aerc-accounts.nix
+++ b/modules/programs/aerc-accounts.nix
@@ -133,6 +133,16 @@ in {
           source =
             "maildir://${config.accounts.email.maildirBasePath}/${cfg.maildir.path}";
         };
+        notmuch = cfg:
+          let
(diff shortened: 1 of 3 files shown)


kchibisov:
The macOS failure looks unrelated to my patch.

onemoresuza:
The tests seem now to have passed. Perhaps it could be merged after the conflicts are solved?

kchibisov:
You could take over the patch if you want, I'm not that interest in it anymore.

PR #4003, opened 2023-05-19
helix: update languages.toml generation (support every option in languages.toml)

Philipp-M:
### Description

Updates the generation of `languages.toml` by supporting every option in it with `programs.helix.languages`. This is a breaking change (which IMHO is justified, as it should be more future-proof).

It closes #3911 as well, as it achieves basically the same (support serializing `grammar` and `use-grammars`) with less code (but a breaking change as mentioned above).

But #2871 is not fixed by this, as this seems to be a little bit more tricky (see discussion in this issue).

Motivation for this is support for multiple language servers (https://github.com/helix-editor/helix/pull/2507)
I've included a comment in the example of this option, which I will remove once a new stable helix version is out (but this will likely take some time, as just a new version came out).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/helix.nix
+++ b/modules/programs/helix.nix
@@ -46,12 +46,31 @@ in {
     };
 
     languages = mkOption {
-      type = types.listOf tomlFormat.type;
-      default = [ ];
-      example = [{
-        name = "rust";
-        auto-format = false;
-      }];
+      type = with types;
+        coercedTo (listOf tomlFormat.type) (language:
+          lib.warn ''
+            The syntax of programs.helix.languages has changed.
+            It now generates the whole languages.toml file instead of just the language array in that file.
+
+            Use
+            { language = <languages list>; }
+            instead.
+          '' { inherit language; }) (addCheck tomlFormat.type builtins.isAttrs);
+      default = { };
+      example = literalExpression ''
+        {
+          # the language-server option currently requires helix from the master branch at https://github.com/helix-editor/helix/
+          language-server.typescript-language-server = with pkgs.nodePackages; {
+            command = "''${typescript-language-server}/bin/typescript-language-server";
+            args = [ "--stdio" "--tsserver-path=''${typescript}/lib/node_modules/typescript/lib" ];
+          };
+
+          language = [{
+            name = "rust";
+            auto-format = false;
+          }];
+        }
+      '';
       description = ''
         Language specific configuration at
         <filename>$XDG_CONFIG_HOME/helix/languages.toml</filename>.
@@ -140,9 +159,8 @@ in {
         "helix/config.toml" = mkIf (cfg.settings != { }) {
           source = tomlFormat.generate "helix-config" cfg.settings;
         };
-        "helix/languages.toml" = mkIf (cfg.languages != [ ]) {
-          source =
-            tomlFormat.generate "helix-config" { language = cfg.languages; };
+        "helix/languages.toml" = mkIf (cfg.languages != { }) {
+          source = tomlFormat.generate "helix-languages-config" cfg.languages;
         };
       };
 
--- a/tests/modules/programs/helix/example-settings.nix
+++ b/tests/modules/programs/helix/example-settings.nix
@@ -21,10 +21,27 @@ with lib;
         };
       };
 
-      languages = [{
-        name = "rust";
-        auto-format = false;
-      }];
+      languages = {
+        language-server.typescript-language-server = let
+          typescript-language-server = config.lib.test.mkStubPackage {
+            outPath = "@typescript-language-server@";
+          };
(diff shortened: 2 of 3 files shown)


happysalada:
hey, I've got one suggestion regarding helix.
Would it make sense to merge the original language.toml configuration with the nix-defined ones ? 
The reason  ask is that if currently you want to specify a new language server for a language, then you have to redefine the whole language options (sources, files...) and add the one change to language servers. It seems it would be amazing if you could just say that you want to override the language-servers option and it took the rest from the original configuration file.
Just a thought.

ncfavier:
It seems like something we could do to make the migration easier is to add an assertion that `cfg.languages ? language`.

Philipp-M:
> Would it make sense to merge the original language.toml configuration with the nix-defined ones ?

This happens already, the merge behavior is described/defined in https://github.com/helix-editor/helix/blob/ff442eecd054e67081158ef2f0eeab0ed9156ca5/helix-loader/src/config.rs#L26

The home-manager version just generates the (user-) TOML files. Actually this comment/commit resulted from a PR that was motivated to support exactly this behavior, previously (a long time ago, before the relevant [PR](https://github.com/helix-editor/helix/pull/3080) was merged) only the `language` array was merged.

> It seems like something we could do to make the migration easier is to add an assertion that `cfg.languages ? language`.

That's a very good point, though I'm asserting now that the option is a set (instead of previously a list).
It will pretty much always have to be a set, we could maybe even restrict this in the type, but an assertion with a helpful message may be better for migration.

modules/programs/helix.nix
```diff
@@ -133,16 +143,27 @@ in {
   };
 
   config = mkIf cfg.enable {
+    assertions = [{
+      assertion = builtins.typeOf cfg.languages == "set";
```
ncfavier, line 147:
```suggestion
      assertion = builtins.isAttrs cfg.languages;
```

ncfavier:
An alternative would be to use something like

```nix
with types; coercedTo (listOf tomlFormat.type)
                      (language: lib.warn "boo" { inherit language; })
                      (addCheck tomlFormat.type builtins.isAttrs)
```

Philipp-M:
Nice nix type-fu, I like it. Better than some trace that doesn't immediately show what's wrong, and keeps working, even when the user hasn't updated it (so he can do it at a later time...), I've updated it accordingly, thanks.

tests/modules/programs/helix/example-settings.nix
```diff
@@ -21,10 +21,22 @@ with lib;
         };
       };
 
-      languages = [{
-        name = "rust";
-        auto-format = false;
-      }];
+      languages = {
+        language-server.typescript-language-server = {
+          # depending on derivation outputs will change the generated settings everytime nixpkgs changes...
```
ncfavier, line 26: We use [package stubs](https://github.com/nix-community/home-manager/blob/master/tests/stubs.nix) for this; you can look at the other tests for inspiration

Philipp-M: Thanks for the tip, fixed it in the latest force-push

ncfavier (approved):

KiruyaMomochi:
For me this warning message is a little confusing:
```plain
warning: The syntax of programs.helix.languages has changed.
It now generates the whole languages.toml file instead of just the language array in that file.

Use
{ language = <languages list>; }
instead.
```

According to the warning message, I have tried to change my config to
```patch
 programs.helix = {
   enable = true;
-  languages = [
+  language = [
     ...
   ]
 };
```

But it does not work. I don't know I should use `languages.language` until I looked into the code.

Therefore, I suggest to change the warning message to
```patch
 Use
-{ language = <languages list>; }
+programs.helix.languages = { language = <languages list>; }
 instead.
```

ncfavier:
Feel free to open a PR

mschwaig:
I was having some trouble with writing a working config that uses this, so I am sharing my code here for setting up `nixd` as a nix language server in case it is useful to others:

```nix
helix = {
  enable = true;
  package = inputs.helix.packages.${pkgs.system}.default;
  languages = {
    language-server.nixd = {
      command = "${pkgs.nixd}/bin/nixd";
    };
    language = [{
      name = "nix";
      auto-format = false;
      language-servers = [ "nixd" "nil" ];
    }];
  };
};
```
which leads to the following `languages.toml` file that the current helix version from master branch is happy with:
```toml
[[language]]
auto-format = false
language-servers = ["nixd", "nil"]
name = "nix"

[language-server.nixd]
command = "/nix/store/v0809ax8pn1glv0vzg5j2dw4zi828mci-nixd-1.2.2/bin/nixd"
```

PR #4008, opened 2023-05-21
Add `systemdTarget` options

Enzime:
### Description

Makes it easier to have multiple sessions from the same `home-manager` configuration

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/polybar.nix
+++ b/modules/services/polybar.nix
@@ -196,6 +196,14 @@ in {
         '';
         example = "polybar bar &";
       };
+
+      systemdTarget = mkOption {
+        type = types.str;
+        default = "tray.target";
+        description = ''
+          Systemd target to bind to.
+        '';
+      };
     };
   };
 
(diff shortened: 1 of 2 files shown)


rycee:
I'm not sure what you mean by multiple sessions? Something like having an X and Wayland session open simultaneously?

rycee:
@Enzime Aha. I think the proper solution would be to use specializations for that. You could have one specialization for X11 and one for Sway. For now, you'll need to activate the specialization manually, though, or set up a systemd service to run the activation script on login. Alternatively, set some environment variable and set `ConditionEnvironment` in the different services depending on which you want to run.

Eventually I think HM will move to applying the current generation through a systemd service by default and then you should be able to have multiple choices in the DM for the different specializations out of the box.

In any case, I'm not fond of the idea of letting people easily mess around with the systemd targets. You can always do that anyway but `mkForce`.

Enzime:
@rycee The specialization stuff sounds quite interesting but sounds too manual for use at the moment. I noticed that there was already some services that define the `systemdTarget` usually for Sway/Wayland sessions, which I find is quite convenient to use with `sway-session.target`, however there's no equivalent target for `i3`/X11 at the moment, so I thought it would be a small win to make it easier to set the systemd target for some services so users can roll their own target.

PR #4011, opened 2023-05-22
Fix type of programs.helix.languages

RGBCube:
### Description

Since it now generates the whole TOML, the type should be `tomlFormat.type` instead of `listOf tomlFormat.type`

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/helix.nix
+++ b/modules/programs/helix.nix
@@ -47,7 +47,7 @@ in {
 
     languages = mkOption {
       type = with types;
-        coercedTo (listOf tomlFormat.type) (language:
+        coercedTo tomlFormat.type (language:
           lib.warn ''
             The syntax of programs.helix.languages has changed.
             It now generates the whole languages.toml file instead of just the language array in that file.


ncfavier:
`coercedTo` is a bit confusing: the first argument is the type we're coercing *from*. So this seems correct, unless you encountered a concrete problem?

RGBCube:
> `coercedTo` is a bit confusing: the first argument is the type we're coercing _from_. So this seems correct, unless you encountered a concrete problem?



> `coercedTo` is a bit confusing: the first argument is the type we're coercing _from_. So this seems correct, unless you encountered a concrete problem?

I don't really know what that does, but I get an error when i give a attrset to the `languages` field with the current version:
```nix
 > ./rebuild.sh -c asus
[sudo] password for nixos: 
error: A definition for option `home-manager.users.nixos.programs.helix.languages' is not of type `list of (TOML value)'. Definition values:
       - In `/nix/store/4xjzm05dkf5q3y33qi2593z7hbj2hqcs-source/machines/asus/helix':
           {
             language = [
               {
                 language-server = {
                   command = "/nix/store/wrmxyjh8jy7x4wi1aqvi6xpglbk9lpnd-bash-language-server-4.9.1/bin/bash-language-server";
```

PR #4012, opened 2023-05-23
fish: add babelfish translation option

emilazy:
### Description

See https://github.com/NixOS/nixpkgs/pull/108947, https://github.com/LnL7/nix-darwin/pull/270. When enabled this brings down fish startup time from "noticeable delay" to "imperceptibly fast" for me.

The NixOS and nix-darwin versions of this construct derivations for the bash sources and place the results in `/etc/fish`. I couldn't figure out if there was any point to that kind of thing so I just used the derivation directly and embedded the bash in the build script. I also didn't abstract it out at all because there's nothing else to translate at present.

I wanted to port `programs.nix-index.enableFishIntegration` to use this and forego the additional shell when enabled, but babelfish [isn't quite ready for it yet](https://github.com/bouk/babelfish/issues/23).

I didn't add a test because I'm not sure what kind of test case would be useful within the Home Manager test framework. I guess you could check the file contains correct code for setting a specific variable, but that might break if babelfish translates it differently (but equally correctly) in the future.

I'd be tempted to forego the option and just use babelfish unconditionally, since it should handle this basic variable-setting use case perfectly, but theoretically people could have added arbitrarily complicated bash code to `home.sessionVariablesExtra` and in that case this could break their configurations if they use edge-case features that babelfish doesn't yet correctly translate. If that's an acceptable tradeoff I'd be happy to amend the PR to default the option on or skip it altogether.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/home-environment.nix
```diff
@@ -291,6 +291,15 @@ in
       '';
     };
 
+    home.sessionVariablesInit = mkOption {
```
ncfavier, line 294: I'd make this `internal`, otherwise it's not clear whether to use this or `sessionVariablesExtra`.

ncfavier:
Maybe we can make this the default conditionally based on `stateVersion`?

modules/home-environment.nix
emilazy: Yeah I was thinking maybe you might want to override it for whatever reason, but probably not.

If it's going to be internal then another alternative would be to expose the derivation that creates `<profile>/etc/profile.d/hm-session-vars.sh` rather than the text, which might generalize more elegantly if there's more things to translate in the future (e.g. `command-not-found.sh`)?

emilazy:
Yes, I had that thought. I would be happy to implement that if defaulting to it outright is a bridge too far. It'd be nice if NixOS and nix-darwin did that too; the speedup is quite considerable...

modules/home-environment.nix
ncfavier: Yes, that would be perfect.

emilazy:
Oh wait, `home.sessionVariablesExtra` is internal anyway; the only way anyone could cause any issue here would be `home.sessionVariables.EVIL = "$(extremely complex bash code)"`, which seems deeply unlikely. (Nothing stops you doing `home.sessionVariables.EVEN_MORE_EVIL = "\"; echo hi \""`, either, which it presumably wouldn't be considered a valid compatibility break if it stopped working.)

I have the `home.stateVersion` logic implemented but maybe worrying about breakage here is overkill for an option that I seriously doubt anyone is doing anything fancier than simple variable and command substitutions with?

ncfavier:
Can babelfish deal with the code that we know may be present in `sessionVariablesExtra`? (At a glance, https://github.com/nix-community/home-manager/blob/58eb968c21d309a6c2b020ea8d64e25c38ceebba/modules/services/gpg-agent.nix#L240-L244 and https://github.com/nix-community/home-manager/blob/58eb968c21d309a6c2b020ea8d64e25c38ceebba/modules/targets/generic-linux.nix#L59-L64)

emilazy:
Yeah, it gets both of these correct:

```fish
if test -z "$SSH_AUTH_SOCK"
  set -gx SSH_AUTH_SOCK (/path/to/gnupg/bin/gpgconf --list-dirs agent-ssh-socket | string collect; or echo)
end
```

and

```fish
/nix/store/lnggzp27m91d1niswv65jdr2kkg61my0-babelfish-1.1.0/bin/babelfish < '/path/to/nix/etc/profile.d/nix.sh' | source
# reset TERM with new TERMINFO available (if any)
set -gx TERM "$TERM"
```

respectively.

In general babelfish is quite sophisticated; e.g. its author said "I'd say babelfish will work fine for 100% of the use case we have in Nixpkgs. At some point I manually went though everything to make sure it does." on the [NixOS PR](https://github.com/NixOS/nixpkgs/pull/108947#issuecomment-757750046) and the only mistranslation I ran into on [this nontrivial script from nix-index](https://github.com/bennofs/nix-index/blob/a8ceb75b5f447ef5d0d5569dfd19401624b6c604/command-not-found.sh) was this comment parsing infelicity: https://github.com/bouk/babelfish/issues/23. It looks like nothing else adds to `home.sessionVariablesExtra` at present but anything added in the future would presumably have to be something more elaborate than anything in the NixOS or nix-darwin initialization scripts for it to trip up babelfish.

emilazy:
Alright, yolo version pushed :)

For what it's worth, Nix ships with an `/etc/profile/nix.fish` which would avoid a conversion step, but the babelfish-translated version is functionally identical other than not using `$fish_user_paths` (which doesn't happen with the `foreign-env` version either). Still, it could be worth conditionalizing and porting to `programs.fish.shellInit`, but not in this PR.

ncfavier:
Looks good. Since this is a potentially breaking change, I'll think we'll have to wait until the 23.05 branchoff (which is happening now for NixOS).

emilazy:
Sure; does that mean this will land in the release after 23.05 and I should move the changelog entry to that one after branch-off happens?

emilazy:
Just checking on this since I'm not familiar with the HM release process - it seems https://github.com/nix-community/home-manager/issues/4013 is done but the version on `master` hasn't been rolled over yet; I guess this should land after the next equivalent of https://github.com/nix-community/home-manager/pull/3440?

rycee:
@emilazy HM master is now 23.11 so I guess you are good to go 🙂

emilazy:
@ncfavier Thanks for the merge! I was just about to rebase this for the release turnover - you'll want to move the release notes to `rl-2311.adoc` :)

ncfavier:
Fixed

PR #4017, opened 2023-05-24
ripgrep: add module

pedorich-n:
### Description

Added a module for [ripgrep](https://github.com/BurntSushi/ripgrep) ([nixpkgs](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/tools/text/ripgrep/default.nix)). 
Added functionality to control a config file, described here: https://github.com/BurntSushi/ripgrep/blob/master/GUIDE.md#configuration-file.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

<s>  - [ ] Added myself and the module files to `.github/CODEOWNERS`.  </s>


modules/programs/ripgrep.nix
```diff
@@ -0,0 +1,58 @@
...
+      };
+
+      configDir = mkOption {
+        type = types.str;
+        default = "${config.xdg.configHome}/ripgrep";
+        defaultText = literalExpression "\${config.xdg.configHome}/ripgrep";
+        description = ''
+          Directory where the <filename>ripgreprc</filename> file will be stored (Inside $HOME).
+        '';
+      };
```
ambroisie, line 28: I believe this kind of option is only added to work around programs that _do not_ look into `XDG_CONFIG_DIR` unless explicitly configured through a `*_PATH` environment variable.

I'd advise that you drop this option and refrain from defining `RIPGREP_CONFIG_PATH`.

modules/programs/ripgrep.nix
```diff
@@ -0,0 +1,58 @@
...
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home = {
+      packages = [ cfg.package ];
+
+      file."${cfg.configDir}/ripgreprc" = mkIf (cfg.config != [ ]) {
+        text = (builtins.concatStringsSep "\n" cfg.config) + "\n";
```
ambroisie, line 50: Is this subject to potential quoting hazards?

modules/programs/ripgrep.nix
pedorich-n: Sadly this is impossible, as `ripgrep` by default doesn't look for a config file anywhere:
>Setting up a configuration file is simple. ripgrep will not look in any predetermined directory for a config file automatically. Instead, you need to set the `RIPGREP_CONFIG_PATH` environment variable to the file path of your config file. Once the environment variable is set, open the file and just type in the flags you want set automatically. 

Source: https://github.com/BurntSushi/ripgrep/blob/master/GUIDE.md#configuration-file

pedorich-n: We can just not expose this option and use `XDG_CONFIG_DIR`. But that would require adding some more `if-else` statements because of Mac, right?

modules/programs/ripgrep.nix
pedorich-n: Hm, I haven't thought of that. Let me see how this is implemented in other modules.

modules/programs/ripgrep.nix
ambroisie: > Sadly this is impossible, as ripgrep by default doesn't look for a config file anywhere:

Ah, in that case carry on.

> But that would require adding more if-else statements because of Mac, right? We'd have to use $HOME/Application Support/... or something there

I _believe_ `config_file` should do the right thing there? I don't use Mac so I'm not sure.

pedorich-n: It's just I saw in other modules check for `isDarwin`, for example:
https://github.com/nix-community/home-manager/blob/6a19225/modules/programs/tealdeer.nix#L9-L12
https://github.com/nix-community/home-manager/blob/6a19225/modules/programs/nushell.nix#L9-L12

But maybe they all copy each other.

modules/programs/ripgrep.nix
pedorich-n: How do you think this should be done instead? I looked through modules and couldn't find anything more "clever" than `concatStringsSep` used in similar cases.

ambroisie: I think that depends on whether or not `ripgrep` would have any quoting issues.

As an example, for shell quoting (`sh`, `bash`, `zsh`, shell scripts, etc...) one usually uses `escapeShellArgs`.

ncfavier: > Every line is a shell argument, after trimming whitespace.

So one argument per line is correct.

```suggestion
        text = lib.concatLines cfg.config;
```

modules/programs/ripgrep.nix
```diff
@@ -0,0 +1,58 @@
...
+    programs.ripgrep = {
+      enable = mkEnableOption "Ripgrep";
+
+      package = mkOption {
+        type = types.package;
+        default = pkgs.ripgrep;
+        defaultText = literalExpression "pkgs.ripgrep";
+        description =
+          "Package containing the <command>ripgrep</command> program.";
+      };
```
ncfavier, line 19:
```suggestion
      package = mkPackageOption pkgs "ripgrep" { };
```

modules/programs/ripgrep.nix
```diff
@@ -0,0 +1,58 @@
...
+        '';
+      };
+
+      config = mkOption {
+        type = with types; listOf str;
+        default = [ ];
+        defaultText = literalExpression "[]";
+        example = [ "--max-columns-preview" "--colors=line:style:bold" ];
+        description = ''
+          List of arguments to pass to ripgrep. Each line is given to ripgrep as a single command line argument verbatim.
```
ncfavier, line 36:
```suggestion
          List of arguments to pass to ripgrep. Each item is given to ripgrep as a single command line argument verbatim.
```

modules/programs/ripgrep.nix
```diff
@@ -0,0 +1,58 @@
...
+        defaultText = literalExpression "\${config.xdg.configHome}/ripgrep";
+        description = ''
+          Directory where the <filename>ripgreprc</filename> file will be stored (Inside $HOME).
+        '';
+      };
+
+      config = mkOption {
+        type = with types; listOf str;
+        default = [ ];
+        defaultText = literalExpression "[]";
```
ncfavier, line 33:
```suggestion
```

modules/programs/ripgrep.nix
ncfavier:
```suggestion
      configDir = mkOption {
        type = types.str;
        default = "${config.xdg.configHome}/ripgrep";
        defaultText = literalExpression "\"\${config.xdg.configHome}/ripgrep\"";
        description = ''
          Directory where the <filename>ripgreprc</filename> file will be stored (Inside $HOME).
        '';
      };
```

Why `Inside $HOME`? This looks like an absolute path.

pedorich-n: I assumed that `home.file.<path>` only can edit files under `$HOME`.

modules/programs/ripgrep.nix
pedorich-n: Nice! Didn't know about that.

modules/programs/ripgrep.nix
```diff
@@ -0,0 +1,51 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let cfg = config.programs.ripgrep;
+in {
+  meta.maintainers = [ maintainers.pedorich-n ];
```
ncfavier, line 7:
```suggestion
  meta.maintainers = [ hm.maintainers.pedorich-n ];
```

ncfavier (approved):

modules/programs/ripgrep.nix
```diff
@@ -0,0 +1,51 @@
...
+        type = types.str;
+        default = "${config.xdg.configHome}/ripgrep";
+        defaultText =
+          literalExpression ''"''${config.xdg.configHome}/ripgrep"'';
+        description = ''
+          Directory where the <filename>ripgreprc</filename> file will be stored.
+        '';
+      };
+
+      config = mkOption {
```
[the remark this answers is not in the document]
pedorich-n: Done

modules/programs/ripgrep.nix
```diff
@@ -0,0 +1,51 @@
...
+
+      configDir = mkOption {
+        type = types.str;
+        default = "${config.xdg.configHome}/ripgrep";
+        defaultText =
+          literalExpression ''"''${config.xdg.configHome}/ripgrep"'';
+        description = ''
+          Directory where the <filename>ripgreprc</filename> file will be stored.
+        '';
+      };
```
rycee, line 23: I would suggest avoiding this type of option in the future and IMHO I would prefer removing this option before too many use the module.

Basically, we should use the default configuration location as much as possible. We have tried to force programs to use XDG configuration directories before and it tends to cause issues. Especially when one has to involve environment variables, which causes the program not to find the configuration unless the user relogs.

Instead please convince upstream to support the XDG configuration directory.

ncfavier: Discussed in https://github.com/nix-community/home-manager/pull/4017#discussion_r1206968471

rycee: Aha, ok. I'll move there.

modules/programs/ripgrep.nix
rycee: Copied from a comment I made elsewhere:

> I would suggest avoiding this type of option in the future and IMHO I would prefer removing this option before too many use the module.
>
> Basically, we should use the default configuration location as much as possible. We have tried to force programs to use XDG configuration directories before and it tends to cause issues. Especially when one has to involve environment variables, which causes the program not to find the configuration unless the user relogs.
>
> Instead please convince upstream to support the XDG configuration directory.

To moderate my comment somewhat, though. I think it is fine to allow using the XDG directories if the program supports it but we should not make "force XDG use" the default since it invariably leads to problems. I also don't think we should clutter each such program with an option to support XDG directories, or in this case any directory in general. Instead perhaps add a Boolean option `home.preferXdgDirectories` or something that other modules, such as this one, can check.

pedorich-n: I agree with what you're saying, but this program is an exception.
Because, as I said previously: there's no default configuration file location. It expects you to set an env variable with a path.

So I had to either force a config file location (like `$XDG`) or expose a parameter to let the users set it. I saw some other modules exposing parameter like that, so I thought it was fine.

>ripgrep will not look in any predetermined directory for a config file automatically. Instead, you need to set the `RIPGREP_CONFIG_PATH` environment variable to the file path of your config file.

Source: https://github.com/BurntSushi/ripgrep/blob/master/GUIDE.md#configuration-file

Let me know what you think is better to do in this case, and I will open a new PR.

rycee: Aha, ok. Yeah, I didn't read the previous discussions so missed that. So in that case I would say to always use a fixed path, e.g., the XDG one, and not make it configurable.

pedorich-n: Okay, I will open a PR to fix that.

pedorich-n: Here it is: https://github.com/nix-community/home-manager/pull/4053

PR #4022, opened 2023-05-26
home-manager: verify username and home directory

### LINKED ISSUE #4019: bug: Running switch with sudo puts symbolic links into `/root` eventhough `homeDirectory` is defined

### Issue description

I was running `home-manager switch` as sudo, mistakenly

```sh
sudo home-manager switch --flake '.#'
```

and it turns out the command updates symlinked files defined with `home.file` at `/root`.
Eventhough my `home.nix` contains:

```nix
  home = {
    inherit username;
    homeDirectory = "/home/${username}";
  };
```

Is it expected? I definitely didn't.

### Maintainer CC

_No response_

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 5.15.94, NixOS, 22.11 (Raccoon), 22.11.20230219.0cf4274`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.11.1`
 - nixpkgs: `/nix/store/hbg5gjcr6bzya30advl28k4ynafhclhn-source`
```

rycee:
### Description

The generation activation script should be run by the user specified in `home.username` and `home.homeDirectory`. If some other user runs the activation script, then files may end up in the wrong place or with the wrong owner.

This commits adds a check early in the activation script that verifies that the running user match the user in the configuration.

Fixes #4019

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/lib-bash/activation-init.sh
```diff
@@ -88,6 +88,24 @@ function setupVars() {
...
+  if [[ $USER != $expectedUser ]]; then
+    _iError 'Error: USER is set to "%s" but we expect "%s"' "$expectedUser" "$USER"
+    exit 1
+  fi
+}
+
+function checkHomeDirectory() {
+  local expectedHome="$1"
+
+  if [[ $HOME != $expectedHome ]]; then
```
ncfavier, line 103:
```suggestion
  if [[ $HOME -ef $expectedHome ]]; then
```

modules/lib-bash/activation-init.sh
```diff
@@ -88,6 +88,24 @@ function setupVars() {
     fi
 }
 
+function checkUsername() {
+  local expectedUser="$1"
+
+  if [[ $USER != $expectedUser ]]; then
```
ncfavier, line 94:
```suggestion
  if [[ $USER != "$expectedUser" ]]; then
```

RHS of `!=` is subject to globbing

modules/lib-bash/activation-init.sh
rycee: Thanks 🙂

modules/lib-bash/activation-init.sh
rycee: Thanks 🙂

PR #4027, opened 2023-05-27
antidote:  add module  , a zsh plug manager

hitsmaxft:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/antidote.nix
+++ b/modules/programs/antidote.nix
@@ -1,0 +1,53 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+let
+  cfg = config.programs.zsh.antidote;
+
+  relToDotDir = file:
+    (optionalString (config.programs.zsh.dotDir != null)
+      (config.programs.zsh.dotDir + "/")) + file;
+
+  zPluginStr = with lib;
+    (pluginNames:
+      optionalString (pluginNames != [ ]) "${concatStrings (map (name: ''
+        ${name}
+      '') pluginNames)}");
+in {
+  meta.maintainers = [ maintainers.hitsmaxft ];
+
+  options.programs.zsh.antidote = {
+    enable = mkEnableOption "antidote - a zsh plugin manager";
+
+    plugins = mkOption {
+      type = types.listOf types.str;
+      default = [ ];
+      example = [ "zsh-users/zsh-autosuggestions" ];
+      description = "List of antidote plugins.";
+    };
+
+    useFriendlyNames = mkEnableOption "friendly names";
+
+    package = mkPackageOption pkgs "antidote" { };
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    home.file."${relToDotDir ".zsh_plugins.txt"}".text = zPluginStr cfg.plugins;
+
+    ### move zsh_plugins.txt
+    programs.zsh.initExtraBeforeCompInit = ''
+      ## home-manager/antidote begin :
+      source ${cfg.package}/antidote.zsh
+      ${optionalString cfg.useFriendlyNames
+      "zstyle ':antidote:bundle' use-friendly-names 'yes'"}
+      bundlefile=${relToDotDir ".zsh_plugins.txt"}
+      zstyle ':antidote:bundle' file $bundlefile
+      staticfile=${relToDotDir ".zsh_plugins.zsh"}
+      zstyle ':antidote:static' file $staticfile
+      antidote load $bundlefile $staticfile
+      ## home-manager/antidote end
+    '';
+  };
+}
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -1125,6 +1125,14 @@ in
           can control it by using the `qt5ct` and `qt6ct` applications;
           - `qt.style.name = "kvantum"`: override the style by using themes
           written in SVG. Supports many popular themes.
+          '';
+      }
+
+      {
+        time = "2023-06-17T22:18:22+00:00";
+        condition = config.programs.zsh.enable;
+        message = ''
+          A new modules is available: 'programs.zsh.antidote'
         '';
       }
     ];
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -51,6 +51,7 @@ let
     ./programs/afew.nix
     ./programs/alacritty.nix
     ./programs/alot.nix
+    ./programs/antidote.nix
     ./programs/aria2.nix
     ./programs/astroid.nix
     ./programs/atuin.nix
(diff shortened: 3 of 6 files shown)


hitsmaxft:
@rycee  wil this PR be review? code is done, any suggestions?

modules/programs/antidote.nix
```diff
@@ -0,0 +1,71 @@
...
+  trimLines = (text:
+    concatLines (map (p: removePrefix "     " p) (splitString "\n" text)));
+in {
+  options.programs.zsh.antidote = {
+    enable = mkEnableOption "antidote - a zsh plugin manager";
+
+    plugins = mkOption {
+      default = [ ];
+      type = types.listOf types.str;
+      description = "List of antidote plugins, strings.";
```
rycee, line 22: No need to say that they are string, the option type already makes that clear.

```suggestion
      description = "List of antidote plugin.";
```

modules/programs/antidote.nix
```diff
@@ -0,0 +1,71 @@
...
+      '') pluginNames)}");
+  trimLines = (text:
+    concatLines (map (p: removePrefix "     " p) (splitString "\n" text)));
+in {
+  options.programs.zsh.antidote = {
+    enable = mkEnableOption "antidote - a zsh plugin manager";
+
+    plugins = mkOption {
+      default = [ ];
+      type = types.listOf types.str;
```
rycee, line 21: The conventional field order is type, default, example, and description.

```suggestion
      type = types.listOf types.str;
      default = [ ];
```

modules/programs/antidote.nix
```diff
@@ -0,0 +1,71 @@
+{ config, lib, pkgs, ... }:
+
+with builtins;
+with lib;
+let
+  module_config = config.programs.zsh.antidote;
```
rycee, line 6: The convention is to use `cfg` for the configuration associated with the module:

```suggestion
  cfg = config.programs.zsh.antidote;
```

modules/programs/antidote.nix
```diff
@@ -0,0 +1,71 @@
+{ config, lib, pkgs, ... }:
+
+with builtins;
```
rycee, line 3: Typically it should be enough with `with lib`:

```suggestion
```

modules/programs/antidote.nix
```diff
@@ -0,0 +1,71 @@
...
+      default = [ ];
+      type = types.listOf types.str;
+      description = "List of antidote plugins, strings.";
+    };
+
+    useFriendlyNames = mkOption {
+      default = false;
+      type = types.bool;
+      description = "use friendlyPath.";
+    };
```
rycee, line 29: Not sure what this actually does but I think something like this would be simpler:

```suggestion
    useFriendlyNames = mkEnableOption "friendly names";
```

modules/programs/antidote.nix
```diff
@@ -0,0 +1,71 @@
...
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.antidote;
+      description = "custom antidote pacakge.";
+    };
```
rycee, line 35: Could use something like:

```suggestion
    package = mkPackageOption pkgs "antidote" { };
```

modules/programs/antidote.nix
```diff
@@ -0,0 +1,71 @@
...
+
+    antidoteHome = mkOption {
+      type = types.nullOr types.path;
+      default = null;
+      defaultText = "~/.antidote or ~/Library/Caches/antidote(darwin).";
+      apply = toString;
+      description = "Path to antidote home directory, cache contents.";
+    };
```
rycee, line 43: In which case would it be necessary to set a custom value here? My spontaneous thought would be to removing this option entirely.

modules/programs/antidote.nix
```diff
@@ -0,0 +1,71 @@
...
+  };
+
+  config = mkIf module_config.enable {
+
+    home.packages = [ module_config.package ];
+
+    home.file.".zsh_plugins.txt".text = zPluginStr (module_config.plugins);
```
rycee, line 50: Maybe this has to be something like
```suggestion
    home.file."${relToDotDir ".zsh_plugins.txt"}".text = zPluginStr (module_config.plugins);
```
? In other words, if the user has configured a special `ZDOTDIR`, will antidote also look there?

modules/lib/maintainers.nix
```diff
@@ -418,4 +418,10 @@
     github = "liyangau";
     githubId = 71299093;
   };
+  hitsmaxft = {
+    name = "Bhe Hongtyu";
+    email = "mfthits@gmail.com";
+    github = "hitsmaxft";
+    githubId = 352727;
+  };
```
rycee, line 426: You seem to already be in the Nixpkgs maintainer list so no need to add yourself here.
```suggestion
```

modules/programs/antidote.nix
```diff
@@ -0,0 +1,71 @@
...
+
+  zPluginStr = with lib;
+    (pluginNames:
+      optionalString (pluginNames != [ ]) "${concatStrings (map (name: ''
+        ${name}
+      '') pluginNames)}");
+  trimLines = (text:
+    concatLines (map (p: removePrefix "     " p) (splitString "\n" text)));
+in {
+  options.programs.zsh.antidote = {
```
rycee, line 16:
```suggestion
  meta.maintainers = [ maintainers.hitsmaxft ];

  options.programs.zsh.antidote = {
```

rycee:
Thanks for the contribution! I've added some comments.

modules/programs/antidote.nix
hitsmaxft: Cache path is not important to custom.  since it is a undocumented option, removed.

modules/programs/antidote.nix
hitsmaxft: such a good advice ,  i've change to follow zsh.dotdir option

rycee:
Thanks! Merged to master now.

modules/programs/antidote.nix
```diff
@@ -0,0 +1,53 @@
...
+
+    ### move zsh_plugins.txt
+    programs.zsh.initExtraBeforeCompInit = ''
+      ## home-manager/antidote begin :
+      source ${cfg.package}/antidote.zsh
+      ${optionalString cfg.useFriendlyNames
+      "zstyle ':antidote:bundle' use-friendly-names 'yes'"}
+      bundlefile=${relToDotDir ".zsh_plugins.txt"}
+      zstyle ':antidote:bundle' file $bundlefile
+      staticfile=${relToDotDir ".zsh_plugins.zsh"}
```
seh, line 47: Should this package create the bundle file automatically? If not, are you expecting that the user will run [_antidote bundle .zsh_plugins.txt_](https://github.com/mattmc3/antidote#performance)?

PR #4031, opened 2023-05-29
nix: add a declarative alternative to Nix channels

midchildan:
### Description
This adds a new option, `nix.channels`. It's the Nix channels equivalent of the `nix.registry` option, and compatible with pre-Flake Nix tooling including nix-env and nix-shell. Like `nix.registry`, this option is useful for pinning Nix channels.

Channels defined in the new option can coexist with channels introduced through the `nix-channel` command. If the same channel exists in both, the one from Home Manager will be prioritized.

This PR builds on #2677 with minor changes and additional tests.

Closes #2677.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/nix.nix
+++ b/modules/misc/nix.nix
@@ -10,6 +10,20 @@ let
 
   isNixAtLeast = versionAtLeast (getVersion nixPackage);
 
+  nixPath = lib.concatStringsSep ":" cfg.nixPath;
+
+  # The deploy path for declarative channels. The directory name is prefixed
+  # with a number to make it easier for files in ~/.nix-defexpr to control the
+  # order they'll be read relative to each other.
+  channelPath = ".nix-defexpr/50-home-manager";
+
+  channelsDrv = let
+    mkEntry = name: drv: {
+      inherit name;
+      path = toString drv;
+    };
+  in pkgs.linkFarm "channels" (lib.mapAttrsToList mkEntry cfg.channels);
+
   nixConf = assert isNixAtLeast "2.2";
     let
 
@@ -127,6 +141,22 @@ in {
       '';
     };
 
+    channels = lib.mkOption {
+      type = with lib.types; attrsOf package;
+      default = { };
+      example = lib.literalExpression "{ inherit nixpkgs; }";
+      description = lib.mdDoc ''
+        A declarative alternative to Nix channels. Whereas with stock channels,
+        you would register URLs and fetch them into the Nix store with
+        {manpage}`nix-channel(1)`, this option allows you to register the store
+        path directly. One particularly useful example is registering flake
+        inputs as channels.
+
+        This option can coexist with stock Nix channels. If the same channel is
+        defined in both, this option takes precedence.
+      '';
+    };
+
     registry = mkOption {
       type = types.attrsOf (types.submodule (let
         inputAttrs = types.attrsOf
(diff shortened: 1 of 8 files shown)


modules/misc/nix.nix
```diff
@@ -10,6 +10,20 @@ let
 
   isNixAtLeast = versionAtLeast (getVersion nixPackage);
 
+  nixPath = lib.concatStringsSep ":" cfg.nixPath;
+
+  # The deploy path for declarative channels. The directory name is prefixed
+  # with a number to make it easier for files in ~/.nix-defexpr to control the
+  # order they'll be read relative to each other.
+  channelPath = ".nix-defexpr/50-home-manager";
```
ambroisie, line 18: What about the path one should use with an XDG-compliant `nix` (like e.g: #3876)?

ambroisie (approved):

modules/misc/nix.nix
```diff
@@ -10,6 +10,27 @@ let
 
   isNixAtLeast = versionAtLeast (getVersion nixPackage);
 
+  nixPath = lib.concatStringsSep ":" cfg.nixPath;
+
+  useXdg = config.nix.enable
+    && (config.nix.settings.use-xdg-base-directories or false);
+  defexprDir = if useXdg then
+    "${config.xdg.dataHome}/nix/defexpr"
```
ambroisie, line 18: Isn't it supposed to be `XDG_STATE_HOME` (according to [the documentation](https://nixos.org/manual/nix/stable/command-ref/files/default-nix-expression))?.

modules/misc/nix.nix
```diff
@@ -10,6 +10,27 @@ let
...
+  else
+    "${config.home.homeDirectory}/.nix-defexpr";
+
+  # The deploy path for declarative channels. The directory name is prefixed
+  # with a number to make it easier for files in defexprDir to control the order
+  # they'll be read relative to each other.
+  channelPath = "${defexprDir}/50-home-manager";
```
ambroisie, line 25: Is that actually needed? Given that `NIX_PATH` is explicitly set in the module.

midchildan: This is needed to avoid having to specify `-f '<nixpkgs>'` to nix-env every time. nix-env looks for expressions in `~/.nix-defexpr` instead of `NIX_PATH`.

midchildan: This is also how actual Nix channels work.

modules/misc/nix.nix
midchildan: Thanks, fixed. I was looking at the outdated PR description in https://github.com/NixOS/nix/pull/5588.

modules/misc/nix.nix
ambroisie: This should be taken care of by `~/.nix-defexpr/channels` linking to `$XDG_STATE_HOME/profiles/channels` according to the manual.

midchildan: `~/.nix-defexpr/channels` is managed by nix-channels. Following the symlink eventually leads to the Nix store, so I can't repurpose it in any way.

ambroisie:
Still relevant

ambroisie:
Still relevant.

alexandru0-dev:
any updates?

ambroisie:
Bad bot.

midchildan:
Rebased to fix conflicts. The new test failures seem to identical to #5450, but I haven't been able to reproduce it locally. Possibly due to me pinning Nixpkgs to an older commit locally.

midchildan:
Opened #5463 to fix the Darwin failure.

SebTM (approved):
LGTM - I personally don't use it but it does not break existing stuff and looks good/I understand the motivation of the change 🙏🏻 

(I will put it on my re-submission list for thueseday - if no concerns/issues raised I would go ahead and merge this 😃)

PR #4039, opened 2023-05-31
programs.sway: always wrap the provided package

teto:
when swapping the sway package with another (e.g., "swayfx") we lose HM's wrapping like session variables, command line arguments etc for no reason. It's inconsistent with how the neovim module behaves too.

If someone wants to bypass the HM wrapping, it's still possible to wrap the package with a function .

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/services/window-managers/i3-sway/sway.nix
```diff
@@ -454,17 +454,25 @@ in {
           platforms.linux)
       ];
 
-      home.packages = optional (cfg.package != null) cfg.package
+      wayland.windowManager.sway.finalPackage = let 
+        ourSway = if cfg.package != null then pkgs.sway-unwrapped else cfg.package;
```
ncfavier, line 458: Got this backward?

modules/services/window-managers/i3-sway/sway.nix
```diff
@@ -335,7 +328,7 @@ in {
 
     package = mkOption {
       type = with types; nullOr package;
-      default = defaultSwayPackage;
+      default = pkgs.sway;
       defaultText = literalExpression "${pkgs.sway}";
```
ncfavier, line 332: While you're at it
```suggestion
      defaultText = literalExpression "pkgs.sway";
```

modules/services/window-managers/i3-sway/sway.nix
teto: indeed

modules/services/window-managers/i3-sway/sway.nix
teto: I've used `swayfx` to show something different than the default.

teto: also it should have been sway-unwrapped anyway

ncfavier: The `defaultText` should show what the `default` value is. Use `example` for an example.

teto:
so I lost a few minutes before discovering the tests ran with a mock/stub "sway". It's pretty smart but the module needs to override sway which broke with the stub. It was simpler to remove the stub so that's what I have done. My changes to the tests fixed `make TEST=sway-bar-focused-colors test` but before I proceed, I would like confirmation that my changes are OK. 
I am a bit confused by the `status_command @i3status@/bin/i3status` that dont seem to get replaced/normalized. Is that due to the change in stubs/is it expected ? cc @rycee @sigprof

teto:
this requires some changes tin nixpkgs to work, so I will try to merge https://github.com/NixOS/nixpkgs/pull/237044 first

rycee:
The sway-stubs.nix file contains `i3status = { };`, which I think would explain the `@i3status@`.

Does this change mean that the CI will download Sway and its dependencies?

PR #4046, opened 2023-06-01
xsession: add option xsession.initBeforeSystemd

wrvsrx:
### Description

This commit add an option called `xsession.initBeforeSystemd` to setup xsession related options such as merging xresources and doing other x11 related setup before `hm-graphical-session.target` is started.

For example, before this commit, if we set dpi using xresources, it won't work for user graphical systemd services since xresources is merged after `hm-graphical-session.target` is started.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/config/home-cursor.nix
+++ b/modules/config/home-cursor.nix
@@ -139,7 +139,7 @@ in {
     }
 
     (mkIf cfg.x11.enable {
-      xsession.initExtra = ''
+      xsession.initBeforeSystemd = ''
         ${pkgs.xorg.xsetroot}/bin/xsetroot -xcf ${cursorPath} ${
           toString cfg.size
         }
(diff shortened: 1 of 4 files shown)


wrvsrx:
I accidentally closed the PR, now I fixed the bug in the previous test and met all commit requirements.

wrvsrx:
maintainers CC: @rycee @polykernel @league

wrvsrx:
This PR is still valid. Maintainers CC: @rycee @polykernel @league. Would you like to review this PR?

olmokramer:
This seems like a good change to me. I have to repeat the `xsetroot` command to set the primary cursor in my window manager's startup script and it looks like this will fix that.

PR #4050, opened 2023-06-02
tests: fix `getOutput` on stubs

ncfavier:
Currently, stub packages can't be used with `getExe`, see https://github.com/nix-community/home-manager/pull/3291#issuecomment-1572756632.

Setting `outputSpecified` prevents `getOutput` from descending into outputs, which don't have an overridden `outPath`.

Alternatives:
- removing all outputs makes things fail
- also overriding `outPath` in outputs is more code

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/tests/modules/programs/autojump/default-settings.nix
+++ b/tests/modules/programs/autojump/default-settings.nix
@@ -8,6 +8,7 @@ with lib;
 
     test.stubs.autojump = {
       buildScript = "mkdir -p $out/bin; touch $out/bin/autojump";
+      outPath = null;
     };
 
     nmt.script = ''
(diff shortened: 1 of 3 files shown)


ncfavier:
Another fix was necessary on top of that because the i3 tests have a stub for `i3` but still expect it to be used in a derivation [here](https://github.com/nix-community/home-manager/blob/3876cc613ac3983078964ffb5a0c01d00028139e/modules/services/window-managers/i3-sway/i3.nix#L191). This used to work by accident because `out` was different from the base derivation.

rycee:
Looks good to me! Merged to master now.

PR #4061, opened 2023-06-05
comodoro: add module

soywod:
### Description

**Comodoro, a CLI to manage your time**. Key features:

- Centralized server timer controllable by multiple clients at the same time
- Multi protocols (only TCP for now, but you can build your own)
- Cycles customizable via config file (Pomodoro style, 52/17 style, custom)
- Server and timer hooks customizable via config file

Website & doc: https://pimalaya.org/comodoro/

Sources: https://github.com/soywod/comodoro

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

@toastal

### DIFF
--- a/modules/programs/comodoro.nix
+++ b/modules/programs/comodoro.nix
@@ -1,0 +1,31 @@
+{ config, lib, pkgs, ... }:
+
+let
+  cfg = config.programs.comodoro;
+  tomlFormat = pkgs.formats.toml { };
+
+in {
+  meta.maintainers = with lib.hm.maintainers; [ soywod ];
+
+  options.programs.comodoro = {
+    enable = lib.mkEnableOption "Comodoro, a CLI to manage your time";
+
+    package = lib.mkPackageOption pkgs "comodoro" { };
+
+    settings = lib.mkOption {
+      type = lib.types.submodule { freeformType = tomlFormat.type; };
+      default = { };
+      description = ''
+        Comodoro configuration.
+        See <link xlink:href="https://pimalaya.org/comodoro/cli/configuration/"/> for supported values.
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    xdg.configFile."comodoro/config.toml".source =
+      tomlFormat.generate "comodoro-config.toml" cfg.settings;
+  };
+}
--- a/modules/services/comodoro.nix
+++ b/modules/services/comodoro.nix
@@ -1,0 +1,70 @@
+{ config, lib, pkgs, ... }:
+
+let
+  cfg = config.services.comodoro;
+
+  args = with cfg; {
+    inherit preset;
+    protocols = if lib.isList protocols then
+      lib.concatStringsSep " " protocols
+    else
+      protocols;
+  };
+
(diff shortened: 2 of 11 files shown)


modules/programs/comodoro.nix
```diff
@@ -0,0 +1,91 @@
...
+      protocols;
+  };
+
+in {
+  meta.maintainers = with lib.hm.maintainers; [ soywod ];
+
+  options = {
+    programs.comodoro = {
+      enable =
+        lib.mkEnableOption "Enable Comodoro, the CLI to manage your time.";
```
sumnerevans, line 22: `mkEnableOption` automatically adds "Enable " to the front and "." to the end.

```suggestion
        lib.mkEnableOption "Comodoro, the CLI to manage your time";
```

modules/programs/comodoro.nix
```diff
@@ -0,0 +1,91 @@
...
+        default = { };
+        description = ''
+          Comodoro configuration.
+          See <link xlink:href="https://pimalaya.org/comodoro/cli/configuration/"/> for supported values.
+        '';
+      };
+    };
+
+    services.comodoro = {
+      enable = lib.mkEnableOption "Enable the Comodoro server.";
```
sumnerevans, line 35:
```suggestion
      enable = lib.mkEnableOption "the Comodoro server";
```

modules/programs/comodoro.nix
```diff
@@ -0,0 +1,91 @@
...
+            Use configuration from the given preset as defined in the configuration file.
+          '';
+        };
+
+        protocols = lib.mkOption {
+          type = with lib.types; nonEmptyListOf nonEmptyStr;
+          description = ''
+            Define protocols the server should use to accept requests.
+          '';
+        };
```
sumnerevans, line 63: Are there any good defaults for this or will users have to specify both of these values to run the server?

sumnerevans (changes requested):

modules/programs/comodoro.nix
soywod: These values are mandatory ones, hence the `nonEmpty*`.

modules/services/comodoro.nix
```diff
@@ -0,0 +1,72 @@
...
+          "PASSWORD_STORE_DIR" = "~/.password-store";
+        }
+      '';
+      description = ''
+        Extra environment variables to be exported in the service.
+      '';
+    };
+
+    settings = {
+      preset = lib.mkOption {
```
rycee, line 36: I don't think  `settings` is necessary here. It is not describing an RFC-42 style configuration like in the programs module.

```suggestion
    preset = lib.mkOption {
```

tests/modules/programs/comodoro/default.nix
```diff
@@ -0,0 +1 @@
+{ comodoro = ./comodoro.nix; }
```
rycee, line 1:
```suggestion
{ comodoro-program = ./comodoro.nix; }
```

rycee:
Thanks! Added some additional comments. Also please fix the commit  message. Should be

> comodoro: add module

rycee:
Thanks! Merged to master now.

PR #4063, opened 2023-06-07
doc: show how to writeShellScriptBin from a file

ldicarlo:
### Description

I added explicitly how to import a shell from a file. 
See https://discourse.nixos.org/t/cannot-run-basic-shell-using-writeshellscriptbin/28835

### Checklist

This PR contains only comments so I guess there is no need for other checks.

### DIFF
--- a/home-manager/home-manager
+++ b/home-manager/home-manager
@@ -337,6 +337,8 @@ $xdgVars
     # (pkgs.writeShellScriptBin "my-hello" ''
     #   echo "Hello, \${config.home.username}!"
     # '')
+    # # or you can load it from a local file: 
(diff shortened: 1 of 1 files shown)


teto:
this is like basic nix, I dont think we have to duplicate the doc and describe all the ways one can create a package.

ncfavier:
Yeah, I'm vaguely -1 on this. Documentation for `writeShellScriptBin` belongs in the documentation for `writeShellScriptBin`.

ldicarlo:
No problem, thank you

teto:
I closed without a message sry but thanks for trying to improve home-manager ;)

PR #4065, opened 2023-06-07
aerc: add package option

janydoe:
### Description

Add package option to the aerc program

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/aerc.nix
+++ b/modules/programs/aerc.nix
@@ -25,6 +25,8 @@ in {
 
     enable = mkEnableOption "aerc";
 
+    package = mkPackageOption pkgs "aerc" { };
+
     extraAccounts = mkOption {
       type = sectionsOrLines;
       default = { };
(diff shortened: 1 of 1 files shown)


ncfavier:
Quoting the [NixOS manual](https://nixos.org/manual/nixos/unstable/#sec-option-declarations-util-mkPackageOption):

> Note: You shouldn’t necessarily make package options for all of your modules. You can always overwrite a specific package throughout nixpkgs by using [nixpkgs overlays](https://nixos.org/manual/nixpkgs/stable/#chap-overlays).

I don't really understand your justification: if you're using home-manager as a NixOS module, then use `nixpkgs.overlays` from the NixOS modules as well?

```nix
nixpkgs.overlays = [ (self: super: {
  aerc = pkgs.unstable.aerc;
}) ];
```

Why would there be a compilation error?

ncfavier:
That's because your `unstable` package set inherits the same `overlays`. So to evaluate `pkgs.aerc`, you need `pkgs.unstable.aerc`, but then that's equal to `pkgs.unstable.unstable.aerc`, which is `pkgs.unstable.unstable.unstable.aerc`, etc.

Do you actually need the `unstable` packages to have the same overlays? I think it's probably a good idea to separate them, since they are after all targetting different package sets.

PR #4066, opened 2023-06-07
Espanso: Fix broken module to be compatible with Espanso version 2.x

liyangau:
### Description

Espanso module does not work with 2.x version.  In the new version configuration and matches are put into different folders as below.
```
config/
  default.yml
match/
  base.yml
```
The updated module supports creating multiple files for both config and matches. Users have the full flexibility to create multiple files for different apps.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/espanso.nix
+++ b/modules/services/espanso.nix
@@ -1,14 +1,20 @@
 { pkgs, config, lib, ... }:
 let
   inherit (lib)
-    mkOption mkEnableOption mkIf maintainers literalExpression types platforms;
+    mkOption mkEnableOption mkIf maintainers literalExpression types platforms
+    mkRemovedOptionModule versionAtLeast;
 
   inherit (lib.hm.assertions) assertPlatform;
 
   cfg = config.services.espanso;
+  espansoVersion = cfg.package.version;
 
   yaml = pkgs.formats.yaml { };
 in {
+  imports = [
+    (mkRemovedOptionModule [ "services" "espanso" "settings" ]
+      "Use services.espanso.configs and services.espanso.matches instead.")
+  ];
   meta.maintainers = with maintainers; [ lucasew ];
   options = {
     services.espanso = {
@@ -89,7 +95,15 @@ in {
   };
 
   config = mkIf cfg.enable {
-    assertions = [ (assertPlatform "services.espanso" pkgs platforms.linux) ];
+    assertions = [
+      (assertPlatform "services.espanso" pkgs platforms.linux)
+      {
+        assertion = versionAtLeast espansoVersion "2";
+        message = ''
+          Only support version Espanso version 2
+        '';
+      }
+    ];
 
     home.packages = [ cfg.package ];
 
@@ -97,11 +111,11 @@ in {
       configFiles = lib.mapAttrs' (name: value: {
         name = "espanso/config/${name}.yml";
         value = { source = yaml.generate "${name}.yml" value; };
-      }) config.services.espanso.configs;
+      }) cfg.configs;
       matchesFiles = lib.mapAttrs' (name: value: {
         name = "espanso/match/${name}.yml";
         value = { source = yaml.generate "${name}.yml" value; };
-      }) config.services.espanso.matches;
+      }) cfg.matches;
     in configFiles // matchesFiles;
 
     systemd.user.services.espanso = {
--- a/modules/services/espanso.nix
+++ b/modules/services/espanso.nix
@@ -32,7 +32,7 @@ in {
         type = yaml.type;
         default = { default = { }; };
         example = literalExpression ''
-          configs = {
+          {
             default = {
               show_notifications = false;
             };
@@ -53,7 +53,7 @@ in {
         type = yaml.type;
         default = { default.matches = [ ]; };
         example = literalExpression ''
-          matches = {
+          {
             base = {
               matches = [
                 {
--- a/modules/services/espanso.nix
+++ b/modules/services/espanso.nix
@@ -35,6 +35,11 @@ in {
             };
           };
         '';
+        description = ''
+          The Espanso configuration to use. See
+          <link xlink:href="https://espanso.org/docs/configuration/basics/"/>
+          for a description of available options.
+        '';
       };
 
       matches = mkOption {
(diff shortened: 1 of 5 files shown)


modules/services/espanso.nix
```diff
@@ -25,41 +21,58 @@ in {
         defaultText = literalExpression "pkgs.espanso";
       };
 
-      settings = mkOption {
+      configs = mkOption {
         type = yaml.type;
-        default = { matches = [ ]; };
+        default = { default = { }; };
```
ncfavier, line 26: Do we need a default?

modules/services/espanso.nix
```diff
@@ -25,41 +21,58 @@ in {
...
         '';
-        description = ''
-          The Espanso configuration to use. See
-          <link xlink:href="https://espanso.org/docs/configuration/"/>
-          for a description of available options.
+      };
+
+      matches = mkOption {
+        type = yaml.type;
+        default = { default.matches = [ ]; };
```
ncfavier, line 47: Ditto

modules/services/espanso.nix
```diff
@@ -70,8 +83,16 @@ in {
...
-      yaml.generate "espanso-default.yml" cfg.settings;
+    xdg.configFile = let
+      configFiles = lib.mapAttrs' (name: value: {
+        name = "espanso/config/${name}.yml";
+        value = { source = yaml.generate "${name}.yml" value; };
+      }) config.services.espanso.configs;
+      matchesFiles = lib.mapAttrs' (name: value: {
+        name = "espanso/match/${name}.yml";
+        value = { source = yaml.generate "${name}.yml" value; };
+      }) config.services.espanso.matches;
```
ncfavier, line 104:
```suggestion
      }) cfg.configs;
      matchesFiles = lib.mapAttrs' (name: value: {
        name = "espanso/match/${name}.yml";
        value = { source = yaml.generate "${name}.yml" value; };
      }) cfg.matches;
```

modules/services/espanso.nix
liyangau: I am following previous config here. I guess there is no harm to set an empty default when the users do not provide anything.

modules/services/espanso.nix
liyangau: same as above

modules/services/espanso.nix
```diff
@@ -25,53 +27,96 @@ in {
...
-                replace = "{{output}}";
-                vars = [{
-                  name = "output";
-                  type = "shell";
-                  params = { cmd = "echo Hello from your shell"; };
-                }];
-              }
-            ];
-          }
+          configs = {
```
bobvanderlinden, line 34: The example of configs shouldnt include the configs key.

bobvanderlinden (commented):
Haven't tried to run this, but code looks good overal 👍 small suggestion.

modules/services/espanso.nix
ncfavier: Oh, sorry, I was unclear. I meant: do we need a `default` that's not just `{}`? I.e., why the `default` inside the `default`?

liyangau: I use `default = { }; inside default is to create an empty `~/.config/espanso/config/default.yml` when it is not set.

Without it, Espanso will throw below error.

```
[ERROR] unable to load config

Caused by:
    missing config directory
```

ncfavier (approved):

liyangau:
Looks like _pkgs.espanso_ does not `passthru.test` version. Shall I remove the checks of `versionAtLeast espansoVersion "2";` to move this forward?

ncfavier:
The stubs don't have versions by default. I'm [working on it](https://github.com/nix-community/home-manager/pull/4069).

ncfavier:
Should be fixed now if you rebase on master.

modules/services/espanso.nix
```diff
@@ -1,19 +1,22 @@
...
-
 in {
-  meta.maintainers = with maintainers; [ lucasew ];
-
+  imports = [
+    (mkRemovedOptionModule [ "services" "espanso" "settings" ]
+      "Use services.espanso.configs and services.espanso.matches instead.")
+  ];
+  meta.maintainers =
+    [ lib.hm.maintainers.liyangau maintainers.bobvanderlinden ];
```
ncfavier, line 19: Why the change of maintainer?

modules/services/espanso.nix
```diff
@@ -25,53 +28,96 @@ in {
...
   };
 
   config = mkIf cfg.enable {
-    assertions = [ (assertPlatform "services.espanso" pkgs platforms.linux) ];
+    assertions = [
+      (assertPlatform "services.espanso" pkgs platforms.linux)
+      {
+        assertion = versionAtLeast espansoVersion "2";
+        message = ''
+          Only support version Espanso version 2
```
ncfavier, line 104:
```suggestion
          The services.espanso module only supports Espanso version 2 or later.
```

modules/services/espanso.nix
liyangau: I copied the maintainer part from another file.  I fix it with the correct ones...

liyangau: Speaking of maintainer, do we keep adding new maintainer in a list or in this case, override it so people know who writes the current one?

ncfavier: Presumably @lucasew still wants to be a maintainer for the module?

modules/services/espanso.nix
```diff
@@ -1,19 +1,25 @@
...
 in {
-  meta.maintainers = with maintainers; [ lucasew ];
-
+  imports = [
+    (mkRemovedOptionModule [ "services" "espanso" "settings" ]
+      "Use services.espanso.configs and services.espanso.matches instead.")
+  ];
+  meta.maintainers = [
+    maintainers.lucasew
+    maintainers.bobvanderlinden
```
ncfavier, line 20: Does @bobvanderlinden want to be added?

liyangau: Yes,  he said it was fine. https://discourse.nixos.org/t/how-to-override-service-espanso-module-for-home-manager/28551/12

ncfavier (approved):

PR #4069, opened 2023-06-08
tests/stubs: inherit default versions from pkgs

ncfavier:
Makes version assertions work out of the box.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


ncfavier:
The failure of the browserpass test on darwin is due to a [nixpkgs issue](https://github.com/NixOS/nixpkgs/pull/236258#issuecomment-1583450593) and is not triggered by this PR. I've disabled the test on darwin for now.

PR #4075, opened 2023-06-09
boxxy: add module

nikp123:
### Description

Added a HM module for generating the [boxxy](https://github.com/queer/boxxy) configuration file.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


nikp123:
I'm not sure what's meant by ``./format`` and backwards compatibility, so I left those as questions.

nikp123:
Going to force push a bit on my own repo to make sure all tests are passing.

(can't catch everything on my first try😅)

modules/programs/boxxy.nix
```diff
@@ -0,0 +1,126 @@
+{ config, lib, pkgs, ... }:
+with lib;
+let
+  cfg = config.programs.boxxy;
+
+  boxxyRulesOpts = types.submodule {
+    options = {
```
ncfavier, line 7: We should add a `freeformType`, see [RFC 42](https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md#part-1-1)

modules/programs/boxxy.nix
```diff
@@ -0,0 +1,126 @@
...
+    };
+  };
+in {
+  options.programs.boxxy = {
+    configPath = mkOption {
+      default = "${config.xdg.configHome}/boxxy/boxxy.yaml";
+      defaultText = "$XDG_CONFIG_HOME/boxxy/boxxy.yaml";
+      type = types.str;
+      description = "Path where to put the generated configuration file";
+    };
```
ncfavier, line 90: Do we need this option? cf https://github.com/nix-community/home-manager/pull/4017#discussion_r1215390589

modules/programs/boxxy.nix
nikp123: Am on it, just going to run a nix test beforehand to make sure the change doesn't break anything.

modules/programs/boxxy.nix
ncfavier: I think it would be better to remove it then

nikp123:
is there any way to execute one specific test, because doing ``#all`` for flakes is painfully slow

ncfavier:
`#boxxy-whatever` should work

modules/programs/boxxy.nix
nikp123: [This commit](https://github.com/nix-community/home-manager/commit/a1782d051c26e7d833bb1a91ab7dc1e302fa826d) solves this.

nikp123:
whoops had a git moment there.

nikp123:
Anyway that should be it. Sorry for the quick commits, typos and failed commands get missed easily.

modules/programs/boxxy.nix
```diff
@@ -0,0 +1,117 @@
...
+    enable = mkEnableOption "boxxy: Boxes in badly behaving applications";
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.boxxy;
+      defaultText = literalExpression "pkgs.boxxy";
+      description = ''
+        Package containing the <command>boxxy</command> program.
+      '';
+    };
```
ncfavier, line 94:
```suggestion
    package = mkPackageOption pkgs "boxxy" { };
```

modules/programs/boxxy.nix
```diff
@@ -0,0 +1,117 @@
+{ config, lib, pkgs, ... }:
+with lib;
+let
+  cfg = config.programs.boxxy;
+
+  configPath = "${config.xdg.configHome}/boxxy/boxxy.yaml";
+  settingsFormat = pkgs.formats.json { };
```
ncfavier, line 7:
```suggestion
  settingsFormat = pkgs.formats.yaml { };
```

nikp123:
Should I squash these commits into a single one, or leave them as is.

modules/programs/boxxy.nix
```diff
@@ -0,0 +1,110 @@
...
+        type = types.str;
+        description = ''
+          Unique identifier of the boxxy rule. This can be any single-line string.
+        '';
+      };
+
+      target = mkOption {
+        type = types.str;
+        default = "";
+        defaultText = "~/.ssh";
```
ncfavier, line 23: Did you mean to use `example` here? Also, does tilde expansion actually work?

modules/programs/boxxy.nix
```diff
@@ -0,0 +1,110 @@
...
+        description = ''
+          What directory/file to redirect.
+        '';
+      };
+
+      rewrite = mkOption {
+        type = types.str;
+        default = "";
+        defaultText = "$XDG_CONFIG_HOME/ssh";
```
ncfavier, line 32: Did you mean to use `example` here? Also, does variable expansion actually work?

modules/programs/boxxy.nix
```diff
@@ -0,0 +1,110 @@
...
+        type = types.listOf (types.str);
+        default = [ ];
+        example = literalExpression ''
+          [
+            "bash"
+            "/usr/bin/sh"
+          ]
+        '';
+        description = ''
+          Apply redirection ONLY to specified executable name.
```
ncfavier, line 56:
```suggestion
          Apply redirection ONLY to specified executable names.
```

modules/programs/boxxy.nix
```diff
@@ -0,0 +1,110 @@
...
+      mode = mkOption {
+        type = types.enum [ "file" "directory" ];
+        default = "directory";
+        description = ''
+          Does the current path redirect a file or a directory?
+        '';
+      };
+
+      only = mkOption {
+        type = types.listOf (types.str);
```
ncfavier, line 47: Nit: remove redundant parentheses (same below)

modules/programs/boxxy.nix
```diff
@@ -0,0 +1,110 @@
...
+        default = [ ];
+        example = [ "/home/example/Projects/my-project" ];
+        description = ''
+          Apply redirection ONLY when in a certain directory.
+        '';
+      };
+
+      env = mkOption {
+        type = types.attrs;
```
ncfavier, line 70: `types.attrs` is deprecated (cf. NixOS manual), something like `attrsOf str` should do (maybe `oneOf [ str ... ]`)

ncfavier (commented):
It's fine, we'll squash on merge.

modules/programs/boxxy.nix
nikp123: Yeah, that was a mistake, fixing that in an another commit.

The tilde expansion works because boxxy launches a shell.

modules/programs/boxxy.nix
nikp123: Yes, that's the same case as above. I kinda mixed up example and defaultText there.

modules/programs/boxxy.nix
nikp123: Updating this also in commit

modules/programs/boxxy.nix
nikp123: This as well

modules/programs/boxxy.nix
nikp123: I hope its processed the same in the json function. Going to test this change before pushing.

modules/programs/boxxy.nix
ncfavier: Are you sure? If I'm reading the right part of the source code (https://github.com/queer/boxxy/blob/dd40d5e967c01f4b639a41a1e12c2089b3ab56d5/src/enclosure/fs.rs#L92) it only does tilde expansion, not variable expansion.

nikp123:
For yaml tree's does it not matter if different keys have the starting dash for a tree?
image

nikp123:
also the strings aren't properly quoted, is there any way around that

nikp123:
> For yaml tree's does it not matter if different keys have the starting dash for a tree? image

This does not seem to matter, it's ok to ignore.

modules/programs/boxxy.nix
nikp123: Oh, apologies. I thought it launches a shell process for each. I should've checked the code as well to be sure.

nikp123:
It should be fine now (I hope).

modules/programs/boxxy.nix
```diff
@@ -0,0 +1,110 @@
...
+        example = "~/.ssh";
+        description = ''
+          What directory/file to redirect.
+        '';
+      };
+
+      rewrite = mkOption {
+        type = types.str;
+        default = "";
+        example = "${xdg.configHome}/ssh";
```
ncfavier, line 32:
```suggestion
        example = literalExpression ''"''${config.xdg.configHome}/ssh"'';
```

nikp123: Whats up with the quotations?

nikp123: Also going to ask: Does it matter in that case for docs generation, because I didn't think that code was parsed anyway?

ncfavier: If you write

```nix
example = "${xdg.configHome}/ssh";
```

you're asking nix to expand `xdg.configHome` (I guess you meant `config.xdg.configHome`) into the string, and making that the example. This will fail because you can't access `config` when building the documentation.

Instead, you want the example to be a string containing a literal nix expression, hence the weird syntax is necessary so that nix doesn't expand the `${}`. (As an alternative, this would also work: `example = "\"\${config.xdg.configHome}/ssh\"";`.)

ncfavier: The code is indeed not *parsed*, but it is *evaluated* to generate the documentation (that's how you can see literal examples like `pkgs.foo` in https://nix-community.github.io/home-manager/options.html).

nikp123: So regardless where the string content is, nix was going to parse that as equal? I should've thought of that, whoops.
But yes, you're right. The end user should use ``config`` instead of the top-level variable to refer to their ``xdg.configHome``.

ncfavier: > So regardless where the string content is, nix was going to parse that as equal?

Not sure what you mean by that.

ncfavier (approved):

nikp123:
Thanks for the help.

PR #4076, opened 2023-06-09
Revert "maintainers: add rasmus-kirk as a maintainer"

ncfavier:
Revert accidental push to master instead of https://github.com/nix-community/home-manager/pull/4004

I can't push to master directly because it's protected...


rycee:
Thanks for sorting it out!

PR #4078, opened 2023-06-10
Add infrastructure for contacts and calendars

teto:
This also adds the modules

  - programs.vdirsyncer,
  - programs.khal, and
  - services.vdirsyncer

that integrate with the new infrastructure.

Co-authored-by: Andrew Scott <3648487+ayyjayess@users.noreply.github.com>
Co-authored-by: Sebastian Zivota <sebastian.zivota@mailbox.org>

### Description

This is a rebased version of the longstanding PR ( > 3 years ?) made to add contact and calendar support.
I've commented out "locale" support for khal since it was causing issues.

With it in its current state I was able to configure vdirsyncer and khal.
My proposition is to merge this wiht all its faults otherwise we can still wait 4 years before having  anything working.
We have 6 months before release to iron things out and until then we make it clear that the module is experimental.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/accounts/calendar.nix
+++ b/modules/accounts/calendar.nix
@@ -6,36 +6,37 @@ let
 
   cfg = config.accounts.calendar;
 
-  localModule = name: types.submodule {
-    options = {
-      path = mkOption {
-        type = types.str;
-        default = "${cfg.basePath}/${name}";
-        defaultText = "‹accounts.contact.basePath›/‹name›";
-        description = "The path of the storage.";
-      };
+  localModule = name:
+    types.submodule {
+      options = {
+        path = mkOption {
+          type = types.str;
+          default = "${cfg.basePath}/${name}";
+          defaultText = "‹accounts.contact.basePath›/‹name›";
+          description = "The path of the storage.";
+        };
 
-      type = mkOption {
-        type = types.enum [ "filesystem" "singlefile" ];
-        description = "The type of the storage.";
-      };
+        type = mkOption {
+          type = types.enum [ "filesystem" "singlefile" ];
+          description = "The type of the storage.";
+        };
 
-      fileExt = mkOption {
-        type = types.nullOr types.str;
-        default = null;
-        description = "The file extension to use.";
-      };
+        fileExt = mkOption {
+          type = types.nullOr types.str;
+          default = null;
+          description = "The file extension to use.";
+        };
 
-      encoding = mkOption {
-        type = types.nullOr types.str;
-        default = null;
-        description = ''
-          File encoding for items, both content and file name.
-          Defaults to UTF-8.
-        '';
+        encoding = mkOption {
+          type = types.nullOr types.str;
+          default = null;
+          description = ''
+            File encoding for items, both content and file name.
+            Defaults to UTF-8.
+          '';
+        };
       };
     };
-  };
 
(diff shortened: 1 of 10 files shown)


ncfavier:
What about https://github.com/nix-community/home-manager/pull/3912?

teto:
#3912 is nowhere close to this old contact branch. This is much better as it generates the config for you.

teto:
also just in case you are not on the chat, I actually asked on matrix before merging. To sum up it's not perfect but let's merge and make it as good as the mail infra https://github.com/nix-community/home-manager/issues/2335#issuecomment-1588127066

DanielSiepmann:
Is there a way to exclude the new WIP modules in my local setup?
I already have a working vdirsync module for myself. This now conflicts with the new upstream provided by this PR.
And I can't migrate to upstream, due to still incomplete accounts modules.

That's why I would like to exclude the upstream vdirsync until its done.

ncfavier:
`disabledModules = [ "programs/vdirsyncer.nix" ];` should do it, at the same level as `imports`, `config`, `options`...

teto:
@DanielSiepmann how does your module compare ? anything we could merge ?

> And I can't migrate to upstream, due to still incomplete accounts modules.

what part do you miss, especially compared to yours ? Note that I messed up when merging. Does https://github.com/nix-community/home-manager/pull/4087 fix your situation ?

ncfavier:
`disabledModules` is supposed to look for relative paths in `modulesPath`, which is `<home-manager/modules>`, so your module should be left alone.

teto:
you might be better off renaming your module, so you can try both.

DanielSiepmann:
@teto
> Does #4087 fix your situation ?
I guess yes, as I end up in `error: The option `accounts.calendar' does not exist. Definition values`, will give it a try (later) and see whether I can migrate to the upstream :) And will provide feedback if I miss anything.

@ncfavier Thanks, it indeed works.

Janik-Haag:
@teto do you mind sharing your config I can't get it running in the current state ^^

my current config is this but the accounts module dosn't create a config for vdirsyncer.
```nix
{ ... }:
{
  services.vdirsyncer.enable = true;
  accounts.contact.basePath = ".con";
  accounts.calendar.basePath = ".cal";
  accounts.calendar.accounts.nerdberg = {
    khal = {
      enable = true;
      color = null;
      readOnly = true;
      type = "calendar";
    };
    local.type = "filesystem";
    local.fileExt = ".ics";
    name = "Nerdberg";
    remote = {
      type = "http";
      url = "https://kalender.nerdberg.de/events.ics";
    };
    vdirsyncer = {
      enable = true;
      collections = null;
      conflictResolution = "remote wins";
    };
  };
}
```

Janik-Haag:
also on another note:  we have some stuff that is using local and other stuff locale which seems like it should be called the same every where, and I think there is a copy paste error:
image

teto:
@Janik-Haag I would recommand first crafting a working configuration manually and once you have it working convert it to home-manager format so you can spot what is missing. My config is here https://github.com/teto/home/blob/main/hm/profiles/vdirsyncer.nix .

could you point in the code the typo ? I am not sure if you refer to `locale` as in LANG or LC_MONETARY variables

PR #4080, opened 2023-06-12
git-credential-oauth: init

tomodachi94:
### Description

Adds a module to enable the Git OAuth credential helper, which is a much more secure alternative to the `store` method that is common practice.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/git-credential-oauth.nix
+++ b/modules/programs/git-credential-oauth.nix
@@ -1,0 +1,24 @@
+{ config, lib, pkgs, ... }:
+
+let
+
+  cfg = config.programs.git-credential-oauth;
+
+in {
+  meta.maintainers = [ lib.maintainers.tomodachi94 ];
+
+  options = {
+    programs.git-credential-oauth = {
+      enable = lib.mkEnableOption "Git authentication handler for OAuth";
+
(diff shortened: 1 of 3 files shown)


modules/programs/git-credential-oauth.nix
```diff
@@ -0,0 +1,33 @@
...
+
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    programs.git.extraConfig.credential.helper =
+      mkIf cfg.enableGitCredentialHelper
+      [ "${cfg.package}/bin/git-credential-oauth" ];
```
rycee, line 31: Is there any situation where you would _not_ want to do this? That is, maybe `enableGitCredentialHelper` could be removed?

tomodachi94: Actually, good point; the only real reason for enabling this module is for the credential helper. (If they want the package in their `$PATH`, they can just add it to `home.packages`. I'll update this when I have a few minutes :)

tomodachi94: Added in latest push.

rycee:
Thanks! Merged to master now.

PR #4086, opened 2023-06-13
senpai: use SCFG format

### LINKED ISSUE #2534: bug: senpai new config format

### Issue description

https://github.com/NixOS/nixpkgs/pull/148667#event-5716391903

I merged a new version of [senpai](https://git.sr.ht/~taiite/senpai). Big changes include migrating from YAML to SCFG and supporting a password cmd option.

From their docs:

```shell-session
$ cat <<EOF >~/.config/senpai/senpai.scfg
address chat.sr.ht
nickname senpai
password "my password can't be this cute (2010)"
# alternatively, specify a command to fetch your password:
# password-cmd  gopass show irc/<username>
EOF
```

### Maintainer CC

@malte-v 

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 5.15.5-zen1, NixOS, 22.05 (Quokka)`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.5pre20211126_55275fc`
 - channels(root): `"home-manager, nixos-22.05pre335103.6daa4a5c045, nixpkgs-mozilla"`
 - channels(toastal): `""`
```

jleightcap:
### Description

Closes #2534.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/senpai.nix
```diff
@@ -41,32 +43,70 @@ in {
...
+        else
+          abort "toSCFG: type ${builtins.typeOf v} is unsupported";
+      toSCFG = v:
+        concatStringsSep "\n"
+        (mapAttrsToList (key: val: "${key} ${toSCFG' val}")
+          (filterAttrs (_: val: val != null && val != { }) v));
+    in toSCFG (cfg.config // cfg.extraConfig);
   };
 
   meta.maintainers = [ hm.maintainers.malvo ];
```
jleightcap, line 112: I can open a PR to add myself as a maintainer -- cc @malte-v

malte-v: please do! you can also remove me as a maintainer since i clearly haven't done any maintenence work😅

ncfavier: You can do it in this PR

jleightcap: Ah, I can just use my nixpkgs entry. Thanks for the pointer 🙂

tests/modules/programs/senpai/example-settings-expected.conf
```diff
@@ -0,0 +1,15 @@
...
+  prompt 2
+}
+
+highlight guest senpai lenon
+nickname Guest123456
+pane-widths {
+  nicknames 16
+}
+
+password-cmd gopass show irc/guest
```
ncfavier, line 13: Shouldn't this be `password-cmd "gopass show irc/guest"`?

ncfavier:
As far as I can tell, `toKDL` works fine at least for the test configuration. We just need to filter out `null` values.

Maybe just add `toSCFG = toKDL;` in `generators.nix` for now?

modules/programs/senpai.nix
```diff
@@ -41,33 +43,71 @@ in {
...
+      Configuration for senpai. For a complete list of options, see
+      <citerefentry><refentrytitle>senpai</refentrytitle>
+      <manvolnum>5</manvolnum></citerefentry>.
+    '';
   };
 
   config = mkIf cfg.enable {
+    warnings = with cfg.config;
+      if (password-cmd != null && password != null) then
+        [ "senpai: password-command overrides password!" ]
```
ncfavier, line 87: This should probably be an assertion.

tests/modules/programs/senpai/example-settings-expected.conf
jleightcap: No, SCFG doesn't quote its strings. See: https://git.sr.ht/~taiite/senpai/tree/master/item/doc/senpai.5.scd#L199

modules/programs/senpai.nix
```diff
@@ -41,32 +43,70 @@ in {
...
+      else
+        [ ];
     home.packages = [ cfg.package ];
-    xdg.configFile."senpai/senpai.yaml".source =
-      cfgFmt.generate "senpai.yaml" cfg.config;
+    xdg.configFile."senpai/senpai.scfg".text = let
+      toSCFG' = v:
```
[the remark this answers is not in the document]
jleightcap: Added in this PR.

jleightcap:
> As far as I can tell, `toKDL` works fine at least for the test configuration. We just need to filter out `null` values.
> 
> Maybe just add `toSCFG = toKDL;` in `generators.nix` for now?

Unfortunately existing `toKDL` isn't close enough to SCFG, mostly related to string handling, but thanks for the pointer. Instead I've included a de novo generator here.

Let me know if more testing is warranted -- since `senpai` seems to be the only package in-tree that supports SCFG I didn't aim for a fully general generator.

tests/modules/programs/senpai/example-settings-expected.conf
jleightcap: Ah, I understand. Thank you for the well reasoned response -- an option wrapping `execve` as a list of strings is much more correct :slightly_smiling_face:

modules/programs/senpai.nix
```diff
@@ -41,17 +57,37 @@ in {
...
+            default = null;
+            description = ''
+              Specifying 'senpai.no-tls' is deprecated, Set
+              <literal>senpai.extraConfig = { tls = false; }</literal>
+              instead.
+            '';
+          };
+          password-cmd = mkOption {
+            type = types.nullOr (types.listOf types.str);
+            default = null;
```
ncfavier, line 72: Probably needs an `example`

modules/programs/senpai.nix
```diff
@@ -61,13 +97,34 @@ in {
...
 
   config = mkIf cfg.enable {
+    assertions = with cfg.config; [
+      {
+        assertion = !isNull password-cmd -> isNull password;
+        message = "senpai: password-command overrides password!";
+      }
+      {
+        # TODO: remove me for the next release
+        assertion = isNull no-tls;
```
ncfavier, line 117: If setting the option at all causes an error, we might as well remove it. Look into `mkRemovedOptionModule` or `mkChangedOptionModule`

modules/programs/senpai.nix
```diff
@@ -61,13 +97,34 @@ in {
...
+    ];
     home.packages = [ cfg.package ];
-    xdg.configFile."senpai/senpai.yaml".source =
-      cfgFmt.generate "senpai.yaml" cfg.config;
+    xdg.configFile."senpai/senpai.scfg".text =
+      lib.hm.generators.toSCFG (cfg.config // cfg.extraConfig);
   };
 
-  meta.maintainers = [ hm.maintainers.malvo ];
+  meta.maintainers = [ maintainers.jleightcap ];
```
ncfavier, line 129: Does @malte-v not want to be a maintainer any more?

modules/lib/generators.nix
```diff
@@ -99,4 +99,31 @@
...
+      else if isInt v || isFloat v || isBool v then
+        toJSON v
+      else
+        abort "toSCFG: type ${typeOf v} is unsupported";
+
+    filterSCFG = filterAttrs (_: val: !isNull val && val != { });
+
+    sectionSCFG = mapAttrsToList (key: val: "${key} ${toSCFG' val}");
+
+    toSCFG = v: concatStringsSep "\n" (sectionSCFG (filterSCFG v));
```
ncfavier, line 127:
```suggestion
    toSCFG = v: concatLines (sectionSCFG (filterSCFG v));
```

(This also takes care of adding a final newline)

modules/programs/senpai.nix
```diff
@@ -61,13 +97,34 @@ in {
         <manvolnum>5</manvolnum></citerefentry>.
       '';
     };
+    extraConfig = mkOption {
+      type = types.attrs;
+      default = { };
+      description = ''
+        Options that should be appended to the senpai configuration file.
+      '';
+    };
```
ncfavier, line 106: I am still not sure why we need this, instead of simply keeping the `freeformType`? This seems to be moving away from [RFC 42](https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md).

modules/programs/senpai.nix
malte-v: that is correct.

modules/lib/generators.nix
jleightcap: Adds a blank line at the end of a block due to the mutual recursion. Looks fine to me, but can reformat if needed.

modules/programs/senpai.nix
jleightcap: Unfamiliarity. Thanks for the link, fixed.

modules/programs/senpai.nix
```diff
@@ -4,7 +4,23 @@ with lib;
...
+      "programs"
+      "senpai"
+      "nickname"
+    ])
+    (mkChangedOptionModule [ "programs" "senpai" "no-tls" ] [
+      "programs"
+      "senpai"
+      "tls"
+    ] (config: !config.programs.senpai.no-tls))
+  ];
```
ncfavier, line 23: Sadly this is not going to work because of https://github.com/NixOS/nixpkgs/issues/96006. (Well, and also because this is defined in a `let in` but not used)

I think the best we can do for now is add assertions that the old options aren't used...

Sorry for leading you to a wrong track, this corner of the module system isn't quite ready yet.

PR #4091, opened 2023-06-14
firefox: improve handling of bookmark toolbar

rembo10:
### Description
Bookmarks weren't being put in folders in the toolbar in Firefox. The if/then condition was basically "either it is a toolbar item or has a directory name". This PR fixes that (just not sure about the indentation). It also allows single bookmark items in the toolbar

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/firefox.nix
+++ b/modules/programs/firefox.nix
@@ -77,12 +77,8 @@ let
           }>${escapeXML bookmark.name}</A>'';
 
       directoryToHTML = indentLevel: directory: ''
-        ${indent indentLevel}<DT>${
-          if directory.toolbar then
-            ''<H3 PERSONAL_TOOLBAR_FOLDER="true">Bookmarks Toolbar''
-          else
-            "<H3>${escapeXML directory.name}"
-        }</H3>
+        ${indent indentLevel}<DT>
+          <H3>${escapeXML directory.name}</H3>
         ${indent indentLevel}<DL><p>
         ${allItemsToHTML (indentLevel + 1) directory.bookmarks}
         ${indent indentLevel}</p></DL>'';
@@ -97,8 +93,9 @@ let
         lib.concatStringsSep "\n"
         (map (itemToHTMLOrRecurse indentLevel) bookmarks);
 
-      bookmarkEntries = allItemsToHTML 1 bookmarks;
-    in pkgs.writeText "firefox-bookmarks.html" ''
+      splitEntries = partition (b: b.toolbar) bookmarks;
+
+    in pkgs.writeText "firefox-bookmarks.html" (''
       <!DOCTYPE NETSCAPE-Bookmark-file-1>
       <!-- This is an automatically generated file.
         It will be read and overwritten.
@@ -106,10 +103,17 @@ let
       <META HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=UTF-8">
       <TITLE>Bookmarks</TITLE>
       <H1>Bookmarks Menu</H1>
+
       <DL><p>
-      ${bookmarkEntries}
+        ${allItemsToHTML 1 splitEntries.wrong}
+    '' + (lib.optionalString (length splitEntries.right > 0) ''
+      <DT><H3 PERSONAL_TOOLBAR_FOLDER="true">Bookmarks Toolbar</H3>
+      <DL><p>
+        ${allItemsToHTML 1 splitEntries.right}
       </p></DL>
-    '';
+    '') + ''
+      </p></DL>
+    '');
 
 in {
   meta.maintainers = [ maintainers.rycee maintainers.kira-bruneau ];
@@ -270,6 +274,13 @@ in {
                       type = types.str;
                       description = "Bookmark url, use %s for search terms.";
                     };
+
+                    toolbar = mkOption {
+                      type = types.bool;
+                      default = false;
+                      description = "If bookmark should be shown in toolbar.";
+                    };
+
                   };
                 }) // {
                   description = "bookmark submodule";


modules/programs/firefox.nix
```diff
@@ -106,8 +107,13 @@ let
...
       <TITLE>Bookmarks</TITLE>
       <H1>Bookmarks Menu</H1>
+
       <DL><p>
-      ${bookmarkEntries}
+        ${bookmarkEntries}
+        <DT><H3 PERSONAL_TOOLBAR_FOLDER="true">Bookmarks Toolbar</H3>
+        <DL><p>
+          ${toolbarEntries}
+        </p></DL>
```
ncfavier, line 116: You're missing a closing `</DT>` tag. Also, shouldn't this section be conditional on there being at least one toolbar entry?

modules/programs/firefox.nix
```diff
@@ -97,7 +93,12 @@ let
         lib.concatStringsSep "\n"
         (map (itemToHTMLOrRecurse indentLevel) bookmarks);
 
-      bookmarkEntries = allItemsToHTML 1 bookmarks;
+      isToolbar = b: allItemsToHTML 1 (filter (bookmark: bookmark.toolbar == b) bookmarks);
```
ncfavier, line 96: Use `lib.partition` instead of two passes of `filter`

ncfavier:
cc @kira-bruneau

modules/programs/firefox.nix
rembo10: Hey thanks for the feedback. I was just going off of what was being done for the individual bookmarks, but looking into it, it seems that even Firefox doesn't include the closing DT tags.... so not sure what to do here. I've updated to make the toolbar section optional

modules/programs/firefox.nix
rembo10: Done

kira-bruneau:
Oh weird, I thought CI would run the tests?

It looks like [tests/modules/programs/firefox/profile-settings-expected-bookmarks.html](https://github.com/nix-community/home-manager/blob/e72f98143ae0ea520050339eb71b138ba5130647/tests/modules/programs/firefox/profile-settings-expected-bookmarks.html) also needs to be updated with this change:
```diff
firefox-profile-settings: FAILED
Expected /nix/store/yz35p03maqwkff0rv0fdpk56svfqm1jd-firefox-bookmarks.html to be same as /nix/store/62s3ps2rqxm4pg9j1blj31gbdnmc9qkj-profile-settings-expected-bookmarks.html but were different:
--- actual
+++ expected
@@ -5,28 +5,21 @@
 <META HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=UTF-8">
 <TITLE>Bookmarks</TITLE>
 <H1>Bookmarks Menu</H1>
-
 <DL><p>
-    <DT><A HREF="https://en.wikipedia.org/wiki/Special:Search?search=%s&amp;go=Go" ADD_DATE="0" LAST_MODIFIED="0" SHORTCUTURL="wiki" TAGS="wiki">wikipedia</A>
+  <DT><H3 PERSONAL_TOOLBAR_FOLDER="true">Bookmarks Toolbar</H3>
+  <DL><p>
+    <DT><A HREF="https://nixos.wiki/wiki/Home_Manager" ADD_DATE="0" LAST_MODIFIED="0">Home Manager</A>
+  </p></DL>
+  <DT><A HREF="https://en.wikipedia.org/wiki/Special:Search?search=%s&amp;go=Go" ADD_DATE="0" LAST_MODIFIED="0" SHORTCUTURL="wiki" TAGS="wiki">wikipedia</A>
   <DT><A HREF="https://www.kernel.org" ADD_DATE="0" LAST_MODIFIED="0">kernel.org</A>
-  <DT>
-  <H3>Nix sites</H3>
+  <DT><H3>Nix sites</H3>
   <DL><p>
     <DT><A HREF="https://nixos.org/" ADD_DATE="0" LAST_MODIFIED="0">homepage</A>
     <DT><A HREF="https://nixos.wiki/" ADD_DATE="0" LAST_MODIFIED="0" TAGS="wiki,nix">wiki</A>
-    <DT>
-  <H3>Nix sites</H3>
+    <DT><H3>Nix sites</H3>
     <DL><p>
       <DT><A HREF="https://nixos.org/" ADD_DATE="0" LAST_MODIFIED="0">homepage</A>
       <DT><A HREF="https://nixos.wiki/" ADD_DATE="0" LAST_MODIFIED="0">wiki</A>
     </p></DL>
   </p></DL>
-<DT><H3 PERSONAL_TOOLBAR_FOLDER="true">Bookmarks Toolbar</H3>
-<DL><p>
-    <DT>
-  <H3>[definition 1-entry 1]</H3>
-  <DL><p>
-    <DT><A HREF="https://nixos.wiki/wiki/Home_Manager" ADD_DATE="0" LAST_MODIFIED="0">Home Manager</A>
-  </p></DL>
-</p></DL>
 </p></DL>
For further reference please introspect /nix/store/cxi4vrbbzg2il190yv49l4s4fni83cm1-nmt-report-firefox-profile-settings
```

kira-bruneau:
I think it would also be good to add to the test that you can add a directory to the toolbar now too!

kira-bruneau:
Hmm actually looking into this in more detail, I'm not sure if this is the right approach. @rembo10 what was the main problem you were trying to solve? It looks like this change will complicate the separation between toolbar & non-toolbar bookmarks and break existing configurations. I think the original idea was that you would have one directory marked as "the toolbar directory", and then you could add bookmarks and directories within that one toolbar folder.

With this change, now the toolbar option for directories will no longer take effect (which will be a breaking change), and you'll have to mark every bookmark you want to put into the toolbar explicitly, rather than just group them together all under "the toolbar folder".

It almost seems weird that the toolbar folder was distinguished by a flag in the first place, since there should only ever be one (I think that's what might be making things confusing?). I feel like it would be more obvious if we had a toolbar option separate from the bookmarks option.

rembo10:
Yeah it's kind of counterintuitive, but I don't know what the best solution is

PR #4095, opened 2023-06-15
qt: bunch of fixes and updates

thiagokokada:
### Description

- Always apply `qt.style.package`. Before this was only applying when `qt.platformTheme = "gnome"`, but this is wrong
- Only set `~/.config/Trolltech.conf` when using `qt.platformTheme = "gtk"` or `qt.platformTheme = "gnome"`
- Add support for Qt6 (untested)
- Add myself as maintainer

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/qt.nix
+++ b/modules/misc/qt.nix
@@ -31,7 +31,7 @@ let
   };
 
 in {
-  meta.maintainers = [ maintainers.rycee ];
+  meta.maintainers = with maintainers; [ rycee thiagokokada ];
 
   imports = [
     (mkChangedOptionModule [ "qt" "useGtkTheme" ] [ "qt" "platformTheme" ]


modules/misc/qt.nix
```diff
@@ -173,26 +175,27 @@ in {
...
     xsession.importedVariables = [ "QT_QPA_PLATFORMTHEME" ]
       ++ lib.optionals (cfg.style.name != null) [ "QT_STYLE_OVERRIDE" ];
 
-    # Enable GTK+ style for Qt4 in either case.
+    # Enable GTK+ style for Qt4 in Gtk/GNOME.
     # It doesn’t support the platform theme packages.
-    home.activation.useGtkThemeInQt4 = hm.dag.entryAfter [ "writeBoundary" ] ''
-      $DRY_RUN_CMD ${pkgs.crudini}/bin/crudini $VERBOSE_ARG \
-        --set "${config.xdg.configHome}/Trolltech.conf" Qt style GTK+
-    '';
```
thiagokokada, line 196: This Crudini project only exists to manipulate INI files: https://github.com/pixelb/crudini

We have way better options now to manage those in Nix.

thiagokokada:
macOS tests are failing because of an unrelated problem (Nix install action is failing).

modules/misc/qt.nix
```diff
@@ -173,26 +179,27 @@ in {
...
       pkgs.libsForQt5.qt5ct
       pkgs.qt6Packages.qt6ct
     ] else if cfg.platformTheme == "kde" then [
       pkgs.libsForQt5.plasma-integration
       pkgs.libsForQt5.systemsettings
     ] else
-      [ pkgs.libsForQt5.qtstyleplugins ];
+      [ pkgs.libsForQt5.qtstyleplugins ])
+      ++ lib.optionals (cfg.style.package != null)
+      (lib.flatten [ cfg.style.package ]);
```
ncfavier, line 193:
```suggestion
      (lib.toList cfg.style.package);
```

modules/misc/qt.nix
ncfavier: We could perhaps condition this on `stateVersion`.

thiagokokada: > This might break existing configurations. Shouldn't Trolltech.conf stay writable? What other applications might modify it?
I never saw other applications writing this file.

thiagokokada: @ncfavier Changed my mind. Will keep the activation script because while I don't think this file is written by other applications, I am not 100% sure and I just realize I don't have any Qt4 applications to test.

So for now the only change I did is to only apply this file when `qt.platformTheme` is either `gtk` or `gnome`.

ncfavier (approved):

PR #4106, opened 2023-06-17
home-manager-tool: Depricate home-manager paths

Rosuavio:
### Description
Depricate hard-coded fallback home-manager paths. Specificly "${XDG_CONFIG_HOME:-$HOME/.config}/nixpkgs/home-manager" and "$HOME/.nixpkgs/home-manager".

Use @HOME_MANAGER_PATH@ if it has been provided and points to something that exisits. Warn the user if it does not point to something.

If we have not been provided with a @HOME_MANAGER_PATH@ that exisits, then for both hard-coded paths show a warning if something exisits where the paths are pointing.

This no longer attempts to use either of the hard-coded paths as fallbacks for the home-manager path.

Alternative to https://github.com/nix-community/home-manager/pull/4084

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/home-manager/home-manager
+++ b/home-manager/home-manager
@@ -104,13 +104,41 @@ function setConfigFile() {
 }
 
 function setHomeManagerNixPath() {
-    local path
-    for path in "@HOME_MANAGER_PATH@" \
-                "${XDG_CONFIG_HOME:-$HOME/.config}/nixpkgs/home-manager" \
-                "$HOME/.nixpkgs/home-manager" ; do
+    local path="@HOME_MANAGER_PATH@"
+
+    if [[ -n "$path" ]] ; then
         if [[ -e "$path" || "$path" =~ ^https?:// ]] ; then
             EXTRA_NIX_PATH+=("home-manager=$path")
             return
+        else
+            _iWarn 'Home Manager not found at %s.' "$path"
+        fi
+    fi
+
+    for p in "${XDG_CONFIG_HOME:-$HOME/.config}/nixpkgs/home-manager" \
+             "$HOME/.nixpkgs/home-manager" ; do
+        if [[ -e "$p" ]] ; then
+            # translators: This message will be seen by very few users that likely are familiar with English. So feel free to leave this untranslated.
+            _iWarn $'The fallback Home Manager path %s has been deprecated and a file/directory was found there.' \
+                   "$p"
+            # translators: This message will be seen by very few users that likely are familiar with English. So feel free to leave this untranslated.
+            _i $'To remove this warning, do one of the fallowing.
+
+1. Explicitly tell Home Manager to use the path, for example by adding
+
+     { programs.home-manager.path = "%s"; }
+
+   to your configuration.
+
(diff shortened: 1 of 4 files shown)


home-manager/home-manager
```diff
@@ -105,12 +105,36 @@ function setConfigFile() {
...
+              { programs.home-manager.path = "%s"; }
+            in %s.
+
+            If you import home-manager directly, you can use the `path` parameter
+              pkgs.callPackage /path/to/home-manager-package { path = "%s"; }
+            when calling the home-manager package.
+
+         b) Remove the deprecated path.
+              $ rm -r "%s"' \
+            "$p" "$p" "${HOME_MANAGER_CONFIG:-your home config}" "$p" "$p"
```
Rosuavio, line 137: I'm not great with bash, so repeating `"$p"` a bunch was the best I could figure out

home-manager/home-manager
```diff
@@ -105,12 +105,36 @@ function setConfigFile() {
...
+            If home-manager comes from your home config, you can set
+              { programs.home-manager.path = "%s"; }
+            in %s.
+
+            If you import home-manager directly, you can use the `path` parameter
+              pkgs.callPackage /path/to/home-manager-package { path = "%s"; }
+            when calling the home-manager package.
+
+         b) Remove the deprecated path.
+              $ rm -r "%s"' \
```
Rosuavio, line 136: I used the nixpkgs unfree message to model this off of. Shooting in the dark if this is the right things to do here.

Rosuavio, line 136: Having all of this be in a warning message ends up being a lot of colored text, maybe its too much. Does only having the first line by in the warning color make more sense?

Rosuavio, line 136: Not sure if we should output the builtin paths in there resolved versions in every case. Maybe it should print the actual string from the code in some if not all the places in the message.

Rosuavio (commented):
@rycee Is this something like what you had in mind in https://github.com/nix-community/home-manager/pull/4084#discussion_r1231909332 ? Also will it be okay to completely stop using those paths immediately or should we still fallback to those paths, but with the warning?

home-manager/home-manager
rycee: You could perhaps use `_iWarn` for only the first line. And then make a second, longer, printing with `_i`.

rycee: I think this is pretty good 🙂 I'll think a bit and see if I come up with anything better.

home-manager/home-manager
rycee: I don't think you can do better with the `printf` command, so this is fine.

home-manager/home-manager
rycee: I'm not sure, will ponder it.

Rosuavio: Updated, looks better to me, but I might never see this again anyways.

rycee:
I've done a few tweaks here and there and merged to master now. Thanks for the contribution!

PR #4108, opened 2023-06-18
borgmatic: Add missing support for output and hooks

### LINKED ISSUE #3760: Borgmatic Hooks?

DamienCassou:
### Description

Fix #3760 by adding output and hooks top-level configuration groups.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/borgmatic.nix
+++ b/modules/programs/borgmatic.nix
@@ -134,6 +134,10 @@ let
 
         extraConfig = extraConfigOption;
       };
+
+      output = { extraConfig = extraConfigOption; };
+
+      hooks = { extraConfig = extraConfigOption; };
     };
   });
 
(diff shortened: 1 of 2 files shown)


rycee:
Thanks! Merged to master now.

PR #4110, opened 2023-06-18
Re-enable nixpkgs release check

ncfavier:
Was disabled in https://github.com/nix-community/home-manager/commit/dc2a4e4146d1626bc6a916cb57c9e12be2399ca3#diff-5c980e4c769c045b39cdf1cf9597e9cb2d45929f77e106ee8d9abbcdd7b501a0 and never re-enabled. No wonder we're getting so many bug reports caused by out of sync nixpkgs!

This should be backported to 23.05, and maybe even to 22.11.

It might make sense to have a short "grace period" around branch-off, so that you can keep using home-manager master and nixpkgs master together even if one of them has bumped its version. This should happen automatically, as turning the release check off and on is prone to forgetting and if we're ready to do that we might as well just bump `.version`. I'm not sure how to best implement this; I think one way would be to ensure that master always "knows" it's master and not a release branch.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/home-environment.nix
+++ b/modules/home-environment.nix
@@ -455,12 +455,12 @@ in
 
     home.enableNixpkgsReleaseCheck = mkOption {
       type = types.bool;
-      default = false;          # Temporarily disabled until release stabilizes.
+      default = true;
       description = ''
         Determines whether to check for release version mismatch between Home
         Manager and Nixpkgs. Using mismatched versions is likely to cause errors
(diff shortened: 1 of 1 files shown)


rycee:
Thanks! Yeah, we'll have to think of some nice way to make this a bit more automated.

rycee:
Or, at least make it part of the normal release process. I'll need to write down the necessary steps in the manual and should also add this then as well.

PR #4122, opened 2023-06-20
zsh: allow setting custom syntax highlighting styles

ThinkChaos:
### Description

Commit message:

> Custom styles allow overriding the default colors.
> Example:
> ```nix
> {
>   zsh.syntaxHighlighting.styles.comment = "fg=#6c6c6c";
> }
> ```
> 
> See https://github.com/zsh-users/zsh-syntax-highlighting/blob/master/docs/highlighters.md

The motivation for adding this is to workaround https://github.com/zsh-users/zsh-syntax-highlighting/issues/510

I also made the package configurable since we have a module now.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/zsh.nix
+++ b/modules/programs/zsh.nix
@@ -212,6 +212,8 @@ let
     options = {
       enable = mkEnableOption "zsh syntax highlighting";
 
+      package = mkPackageOption pkgs "zsh-syntax-highlighting" { };
+
       styles = mkOption {
         type = types.attrsOf types.str;
         default = {};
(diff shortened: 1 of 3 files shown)


ThinkChaos:
Small feedback for contributions:
I think it'd be nice to have a link to the `nmt` repo or bash-lib in the issue template. I had to go digging a bit to find what assertions exist. Or maybe link a wiki page in this repo instead of the commit so you can expand a bit more and link the commit and bash-lib from there.

modules/programs/zsh.nix
```diff
@@ -584,11 +606,17 @@ in
...
+        (optionalString cfg.syntaxHighlighting.enable
           # Load zsh-syntax-highlighting after all custom widgets have been created
           # https://github.com/zsh-users/zsh-syntax-highlighting#faq
-          "source ${pkgs.zsh-syntax-highlighting}/share/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh"
-        )
+        ''
+          source ${cfg.syntaxHighlighting.package}/share/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh
+          ${lib.concatStringsSep "\n" (
+              lib.mapAttrsToList
+                (name: value: "ZSH_HIGHLIGHT_STYLES['${name}']='${value}'")
```
ncfavier, line 616:
```suggestion
                (name: value: "ZSH_HIGHLIGHT_STYLES[${lib.escapeShellArg name}]=${lib.escapeShellArg value}")
```

Maybe?

ncfavier:
> Small feedback for contributions: I think it'd be nice to have a link to the `nmt` repo or bash-lib in the issue template. I had to go digging a bit to find what assertions exist. Or maybe link a wiki page in this repo instead of the commit so you can expand a bit more and link the commit and bash-lib from there.

https://github.com/nix-community/home-manager/pull/4125

ThinkChaos:
Thanks for the quick review and updating the contribution guide!  
I applied your suggestion and squashed the commit.

PR #4131, opened 2023-06-21
Speed up copyFonts

apeschar:
### Description

When a lot of fonts are used, the `copyFonts` step can take multiple seconds. It currently deletes and again copies the fonts from the Nix store, even if the files are identical.

This PR makes two changes:

1. Skip font installation if the derivation hasn't changed: the path to the build in the Nix store is written to a file. As long as the Nix store path doesn't change (ie, the font build is the same), the copyFonts step is skipped.
1. Use `rsync` instead of `install` to copy font files. Rsync will only copy files that are new or have actually changed.

With these changes the `copyFonts` step is instant when no fonts are changed, and quite a lot faster when there are changes.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/targets/darwin/fonts.nix
+++ b/modules/targets/darwin/fonts.nix
@@ -10,20 +10,17 @@ let
     pathsToLink = "/share/fonts";
   };
   fonts = "${fontsEnv}/share/fonts";
+  installDir = "${homeDir}/Library/Fonts/HomeManager";
 in {
   # macOS won't recognize symlinked fonts
   config = mkIf pkgs.stdenv.hostPlatform.isDarwin {
-    home.activation.copyFonts = hm.dag.entryAfter [ "writeBoundary" ] ''
-      copyFonts() {
-        rm -rf ${homeDir}/Library/Fonts/HomeManager || :
-
-        local f
-        find -L "${fonts}" -type f -printf '%P\0' | while IFS= read -rd "" f; do
-          run install $VERBOSE_ARG -Dm644 -T \
-            "${fonts}/$f" "${homeDir}/Library/Fonts/HomeManager/$f"
-        done
-      }
-      copyFonts
-    '';
+    home.file."Library/Fonts/.home-manager-fonts-version" = {
+      text = "${fontsEnv}";
+      onChange = ''
+        run mkdir -p ${escapeShellArg installDir}
+        run ${pkgs.rsync}/bin/rsync $VERBOSE_ARG -acL --chmod=u+w --delete \
+          ${escapeShellArgs [ "${fonts}/" installDir ]}
+      '';
+    };
   };
 }


modules/targets/darwin/fonts.nix
```diff
@@ -10,18 +10,24 @@ let
...
   fonts = "${fontsEnv}/share/fonts";
+  installDir = "${homeDir}/Library/Fonts/HomeManager";
+  versionFile = "${installDir}/derivation";
 in {
   # macOS won't recognize symlinked fonts
   config = mkIf pkgs.stdenv.hostPlatform.isDarwin {
     home.activation.copyFonts = hm.dag.entryAfter [ "writeBoundary" ] ''
       copyFonts() {
-        rm -rf ${homeDir}/Library/Fonts/HomeManager || :
+        if [[ $(cat ${versionFile} 2>/dev/null) = ${fontsEnv} ]]; then
```
berbiche, line 20:
```suggestion
        if [[ $(cat ${versionFile} 2>/dev/null) = ${escapeShellArg fontsEnv} ]]; then
```

modules/targets/darwin/fonts.nix
```diff
@@ -10,18 +10,24 @@ let
...
-        find -L "${fonts}" -type f -printf '%P\0' | while IFS= read -rd "" f; do
-          $DRY_RUN_CMD install $VERBOSE_ARG -Dm644 -T \
-            "${fonts}/$f" "${homeDir}/Library/Fonts/HomeManager/$f"
-        done
+        $DRY_RUN_CMD mkdir -p ${installDir}
+        $DRY_RUN_CMD ${pkgs.rsync}/bin/rsync $VERBOSE_ARG -acL --chmod=u+w --delete ${fonts}/ ${installDir}
+
+        if [[ ! -v DRY_RUN ]]; then
+          echo ${fontsEnv} > ${versionFile}
+        fi
```
berbiche, line 30:
```suggestion
        $DRY_RUN_CMD echo ${escapeShellArg fontsEnv} > ${versionFile}
```
`fontsEnv` declaration could also be updated to use `escapeShellArg` instead.

modules/targets/darwin/fonts.nix
```diff
@@ -10,18 +10,24 @@ let
...
+          return 0
+        fi
 
-        local f
-        find -L "${fonts}" -type f -printf '%P\0' | while IFS= read -rd "" f; do
-          $DRY_RUN_CMD install $VERBOSE_ARG -Dm644 -T \
-            "${fonts}/$f" "${homeDir}/Library/Fonts/HomeManager/$f"
-        done
+        $DRY_RUN_CMD mkdir -p ${installDir}
+        $DRY_RUN_CMD ${pkgs.rsync}/bin/rsync $VERBOSE_ARG -acL --chmod=u+w --delete ${fonts}/ ${installDir}
```
berbiche, line 26:
```suggestion
        $DRY_RUN_CMD ${pkgs.rsync}/bin/rsync $VERBOSE_ARG -acL --chmod=u+w --delete ${escapeShellArg fonts}/ ${installDir}
```

berbiche (commented):
Nice improvement 👍 
Thanks for the contribution.

apeschar:
Thanks for your review, @berbiche. I think I've fixed your concerns.

How do you feel about merging this? :)

apeschar:
@ncfavier, would you perhaps like to have a look at this PR?

I think I've addressed @berbiche's comments, and this will be a big performance boost for those on macOS with a lot of fonts or with large fonts (eg, Nerd Fonts). I've been using this for the past months without issues.

ncfavier:
Do we need step 1, isn't step 2 enough?

apeschar:
Need no, but it speeds things up significantly:

`rsync` still walks the entire tree and compares the content of all the files because it cannot rely on timestamps (since all timestamps are 0 in Nix).

So although that is faster than rewriting all the files, it still involves reading all files, which can take a few seconds.

ncfavier:
Fair enough. Have you considered something like

```nix
home.file.${versionFile} = {
  text = fontsEnv;
  onChange = ''
    rsync ...
  '';
};
```

?

apeschar:
I had not considered that; I was unaware of this possibility. I've implemented the same logic using onChange. Have another look, if you like.

ncfavier (approved):

apeschar:
Rebased against master. I think this can safely be merged.

modules/targets/darwin/fonts.nix
```diff
@@ -10,20 +10,17 @@ let
...
-      }
-      copyFonts
-    '';
+    home.file."Library/Fonts/.home-manager-fonts-version" = {
+      text = "${fontsEnv}";
+      onChange = ''
+        mkdir -p ${escapeShellArg installDir}
+        ${pkgs.rsync}/bin/rsync $VERBOSE_ARG -acL --chmod=u+w --delete \
+          ${escapeShellArgs ["${fonts}/" installDir]}
+      '';
```
sumnerevans, line 23: These probably still need to be protected by the `run` shell function.

https://nix-community.github.io/home-manager/release-notes.xhtml#sec-release-24.05-highlights

sumnerevans (changes requested):

apeschar:
This PR is still applicable. Just rebased on master.

modules/targets/darwin/fonts.nix
apeschar: Never mind, I see that `run` is also necessary to respect the `silence` flag. I've added it. :)

sumnerevans (approved):

rycee:

andreykaipov:
Another awesome part of this PR is it fixes copyFonts on macOS for me. No idea how it ever worked for others before!

I used to get the following issue since BSD `find` doesn't have `-printf`: 

```
Activating copyFonts
find: -printf: unknown primary or operator
```

PR #4135, opened 2023-06-21
starship: Remove INSIDE_EMACS checks when enabling shell integration

IvanMalison:
### Description

Many of the terminals supported inside emacs work perfectly fine with STARSHIP.
The TERM=dumb case already handles the tramp and eterm cases, so as far as I can
tell, this is basically just a check for the benefit of OLD versions of
term-mode (see
https://www.emacswiki.org/emacs/AnsiTerm#:~:text=Historically%2C%20'M%2Dx%20ansi%2Dterm,the%20older%20'C%2Dc'%20binding.,
which indicates that it also now handles colors).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/starship.nix
+++ b/modules/programs/starship.nix
@@ -100,26 +100,26 @@ in {
     };
 
     programs.bash.initExtra = mkIf cfg.enableBashIntegration ''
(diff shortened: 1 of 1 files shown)


ncfavier (approved):
This seems to be the better alternative: https://github.com/nix-community/home-manager/pull/4116#discussion_r1237907253

PR #4138, opened 2023-06-22
Fixing some minor issues with calendar infra.

Janik-Haag:
### Description

I fixed a small issues in khal because it excepts a certain setting only in the default section and I fixed a small docs error in the calendar infra. 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/khal.nix
+++ b/modules/programs/khal.nix
@@ -25,7 +25,6 @@ let
   genCalendarStr = name: value:
     concatStringsSep "\n" ([
       "[[${name}]]"
-      "highlight_event_days = True"
       "path = ${
         value.local.path + "/"
         + (optionalString (value.khal.type == "discover") value.khal.glob)
@@ -158,6 +157,7 @@ in {
           # locale = definedAttrs (cfg.locale // { _module = null; });
 
           default = optionalAttrs (!isNull primaryAccount) {
+            highlight_event_days = true;
             default_calendar = if isNull primaryAccount.primaryCollection then
               primaryAccount.name
             else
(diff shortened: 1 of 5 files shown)


modules/programs/khal.nix
```diff
@@ -158,6 +157,7 @@ in {
           # locale = definedAttrs (cfg.locale // { _module = null; });
 
           default = optionalAttrs (!isNull primaryAccount) {
+            highlight_event_days = True;
```
ncfavier, line 160:
```suggestion
            highlight_event_days = true;
```

Could you add a test so that this would have been caught?

Janik-Haag: I will take a look at writing tests.

Janik-Haag:
The test currently fails and I suspect it's something with the test stubs but I'm don't know.

ncfavier:
You can click on [Details](https://github.com/nix-community/home-manager/actions/runs/5430144422/jobs/9875678168?pr=4138) and see which step is failing

Janik-Haag:
The test it self, it already fails locally but I don't know why.

tests/modules/programs/khal/default.nix
ncfavier, line 1: This file should contain an attribute set mapping test names to test files. See neighbouring [`kitty/default.nix`](https://github.com/nix-community/home-manager/blob/master/tests/modules/programs/kitty/default.nix) for example.

Janik-Haag: Argh sorry, I should have just taken a closer look, builds fine now and catches the error I made above. A better test would probably to compare if the config that gets generated is the same as the expected one but better this then nothing I guess :shrug:

PR #4144, opened 2023-06-22
home-cursor.nix: enable gtk module when enabling gtk config generation

Maroka-chan:
The gtk configurations are not generated unless config.gtk is enabled. This is a point of confusion because config.home.pointerCursor.gtk can essentially be disabled, despite having it enabled.

### Description

I changed the home.pointerCursor.gtk option such that it also enables the gtk module. The pointerCursor.gtk option essentially does nothing when the gtk module is disabled, so I think it is safe to assume that you would want to enable gtk when using this option. My cursor size was not consistent between applications, and it admittedly took me way too long to figure out that this was the problem..

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/config/home-cursor.nix
+++ b/modules/config/home-cursor.nix
@@ -42,7 +42,8 @@ let
 
       gtk = {
         enable = mkEnableOption ''
-          gtk config generation for <option>home.pointerCursor</option>
+          gtk config generation for <option>home.pointerCursor</option>.
+	  Note that <option>gtk</option> has to be enabled for the configs to take effect
         '';
       };
     };
@@ -152,7 +153,6 @@ in {
     })
 
     (mkIf cfg.gtk.enable {
-      gtk.enable = true;
       gtk.cursorTheme = mkDefault { inherit (cfg) package name size; };
     })
   ]);
--- a/modules/config/home-cursor.nix
+++ b/modules/config/home-cursor.nix
@@ -42,8 +42,7 @@ let
 
       gtk = {
         enable = mkEnableOption ''
-          gtk config generation for <option>home.pointerCursor</option>.
-	  Note that <option>gtk</option> has to be enabled for the configs to take effect
+          gtk config generation for <option>home.pointerCursor</option>
         '';
       };
     };
@@ -105,6 +104,12 @@ in {
         backends, you can toggle them via the enable option. For example,
         <xref linkend="opt-home.pointerCursor.x11.enable"/>
         will enable x11 cursor configurations.
+	</para><para>
+        Note that this will merely generate the cursor configurations.
+	To apply the configurations, the relevant subsytems must also be configured.
+	For example, <xref linkend="opt-home.pointerCursor.gtk.enable"/> will generate
+        the gtk cursor configuration, but <xref linkend="opt-gtk.enable"/> needs
+        to be set for it to be applied.
       '';
     };
   };
--- a/modules/config/home-cursor.nix
+++ b/modules/config/home-cursor.nix
@@ -152,6 +152,7 @@ in {
     })
 
     (mkIf cfg.gtk.enable {
+      gtk.enable = true;
       gtk.cursorTheme = mkDefault { inherit (cfg) package name size; };
     })
   ]);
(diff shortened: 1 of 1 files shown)


ncfavier:
@polykernel what do you think?

polykernel:
Sorry for the extremely late response, some real life stuff got in the way.

> The gtk configurations are not generated unless config.gtk is enabled. This is a point of confusion because config.home.pointerCursor.gtk can essentially be disabled, despite having it enabled.

I agree that this behavior can be confusing for users, but enabling the gtk module implictly extends the scope of home.pointerCursor. The sole responsibility of the `home.pointerCursor` module is to generate cursor configurations. When the option `home.pointerCursor.<backend>` is enabled, the outcome should be that the cursor configuration for that particular backend is generated. The idea is to separate generation of the configuration from activation and thereby make activation explicit.

For example, suppose an user has the following block in their home-manager configuration
```nix
...
gkt.enable = true;
home.pointerCursor.gtk.enable = true;
...
```
If the user wants to disable all gtk configurations, then commenting out gtk.enable suffice. However, if home.pointerCursor.gtk also contained logic for activating the gtk module, then gtk.enable no longer serves as the master switch and this can potentially leads to some subtle bugs(one can argue using mkOverride would solve the problem but this just shifts the level where it occurs since then the mkOverride priority essentially becomes the default priority).

Therefore I think it is important to preserve the separation between generating and activating the configuration even though it is inconvenient here. The documentation for the module should be updated to indicate this.

rycee:
Agreed, can just add some text in the `home.pointerCursor.gtk.enable` to indicate that it only applies if `gkt.enable` is enabled.

Maroka-chan:
That makes sense. I added a note to the option description instead. It's hopefully clear that it refers to the config.gtk option.

modules/config/home-cursor.nix
```diff
@@ -42,7 +42,8 @@ let
 
       gtk = {
         enable = mkEnableOption ''
-          gtk config generation for <option>home.pointerCursor</option>
+          gtk config generation for <option>home.pointerCursor</option>.
+	  Note that <option>gtk</option> has to be enabled for the configs to take effect
```
polykernel, line 46: It may be worthwhile to add a clause under the main `home.pointerCursor` submodule's description since the same requirement holds for the x11 backend and `xsession.enable`.

Perhaps something akin to
```nix
description = ''
  Cursor configuration. Set to <literal>null</literal> to disable.
  </para><para>
  Top-level options declared under this submodule are backend independent
  options. Options declared under namespaces such as <literal>x11</literal>
  are backend specific options. By default, only backend independent cursor
  configurations are generated. If you need configurations for specific
  backends, you can toggle them via the enable option. For example,
  <xref linkend="opt-home.pointerCursor.x11.enable"/>
  will enable x11 cursor configurations.
  </para><para>
  Note that the cursor configurations are only generated, the relevant
  subsytems must be configured for them to take effect. For example,
  <xref linkend="opt-home.pointerCursor.gtk.enable"/> will enable
  gtk cursor configuration but <xref linkend="opt-gtk.enable"/> needs
  to be set for it to be applied.
'';
```

Maroka-chan: Good point. I've now removed the note on the gtk option and added the more general note under the main submodule description :)

polykernel (approved):

modules/config/home-cursor.nix
```diff
@@ -104,6 +104,12 @@ in {
         backends, you can toggle them via the enable option. For example,
         <xref linkend="opt-home.pointerCursor.x11.enable"/>
         will enable x11 cursor configurations.
+	</para><para>
+        Note that this will merely generate the cursor configurations.
+	To apply the configurations, the relevant subsytems must also be configured.
+	For example, <xref linkend="opt-home.pointerCursor.gtk.enable"/> will generate
+        the gtk cursor configuration, but <xref linkend="opt-gtk.enable"/> needs
+        to be set for it to be applied.
```
ncfavier, line 112: You seem to have mixed spaces and tabs here

Maroka-chan: oh shoot, you're right. I haven't gotten around to configure vim on this machine yet, so I didn't consider that it made tabs. In any case, It should be consistent now.

ncfavier (approved):
macos tests should succeed once [nixpk.gs/pr-tracker.html?pr=242954](https://nixpk.gs/pr-tracker.html?pr=242954) turns all green

PR #4146, opened 2023-06-23
fix: antidote zsh path changed

hitsmaxft:
### Description

fix bug in latest added module antidote.  

since the package submit into nixpkgs is different from my local overlays, relocating zsh path

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/antidote.nix
+++ b/modules/programs/antidote.nix
@@ -39,7 +39,7 @@ in {
     ### move zsh_plugins.txt
     programs.zsh.initExtraBeforeCompInit = ''
       ## home-manager/antidote begin :
-      source ${cfg.package}/antidote.zsh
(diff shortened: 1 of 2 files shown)


hitsmaxft:
Is there a way to performing a pkg file check in nmt.script ?

rycee:
@hitsmaxft It is possible but it will introduce a dependency on the package, which we typically try to avoid to keep the number of downloads down in the CI job.

rycee:
In any case, that's for the fix! I've merged it to master now.

hitsmaxft:
> @hitsmaxft It is possible but it will introduce a dependency on the package, which we typically try to avoid to keep the number of downloads down in the CI job.

Got it. Slowing tests is harmful, anyway,  I've find a way to  testing with   `home-manager switch -I home-manager=xxx'.

Issue #4054, opened 2023-06-03
parse zellij keybinds KDL


ymgyt:
### Description

I am trying to output the following zellij config

```kdl
keybinds {
    normal {
        bind "Ctrl l" { SwitchToMode "Session"; }
        bind "Ctrl r" { SwitchToMode "Resize"; }
    }
}
```

my nix file 

```nix
{ ... }: {
  programs.zellij = {
    enable = true;
    settings = {
      keybinds = {
        normal = {
          bind = [
            {
              _args = [ "Ctrl l" ];
              SwitchToMode = "Session";
            }
            { 
              _args = [ "Ctrl r" ];
              SwitchToMode = "Resize";
            }
          ];
        };
      };
    };
  };
}
```

But in fact, the following kdl is generated

```kdl
keybinds {
    normal {
        bind {
            - "Ctrl l" {
                SwitchToMode "Session"
            }
            - "Ctrl r" {
                SwitchToMode "Resize"
            }
        }
    }
}
```

zellij failed to parse it

```
> zellij                                                                                                                            06/04/2023 03:51:20 AM

  × Failed to parse Zellij configuration
   ╭─[/home/ymgyt/.config/zellij/config.kdl:3:1]
 3 │     		bind {
 4 │ ╭─▶ 			- "Ctrl l" {
 5 │ │   				SwitchToMode "Session"
 6 │ ├─▶ 			}
   · ╰──── Unsupported action: -
 7 │     			- "Ctrl r" {
   ╰────
  help: For more information, please see our configuration guide: https://zellij.dev/documentation/configuration.html
```

ncfavier:
@h7x4

antholeole:
fix is otw: https://github.com/nix-community/home-manager/pull/4160/files

gotcha:
I have added documentation: https://github.com/nix-community/home-manager/pull/4665

Feedback welcome !

PR #4160, opened 2023-06-25
toKDL: added new _repeatedKey keyword to kdl to simulate repeated KDL keys

antholeole:
### Description

KDL syntax allows us to repeat keys. This is impossible in nix, so there has to be a workaround. There was already the notion of "special keys" in nix2kdl, so I extended it one bit further and added a special key. Now, one key will repeat for all children in _repeatKeyFor. For instance:

```nix
myKey = {
   _repeatKeyFor = ["a" "b"];
};
```
translates to:
```kdl
myKey "a"
myKey "b"
```

I also documented this in the zellij (currently the only package to use KDL) package. Also documented the old features because they were hard to discover (i.e. https://github.com/nix-community/home-manager/issues/4054)

### Checklist

- [x] Change is backwards compatible. (unless someone was using the _repeatKeyFor keyword - which is incredibly ulikely)

    ```
    {component}: {description}

    {long description}
    ```


ncfavier:
I'm not sure I like having to pollute one's configuration with `_repeatedKey`s to define something as basic as key bindings; also this leads to invalid things being expressible, like `bind = { _repeatedKeys = [ ... ]; otherStuffThatWillBeIgnored = ...; }`.

I think I'd prefer a `_repeatableKeys` magic key in the *parent* attribute set, used like this:

```nix
keybinds.normal = {
  _repeatableKeys = [ "bind" ];
  bind = [ { ... } { ... } ];
};
```

That way we can set it once in the module and the user doesn't have to care about it. Would you be willing to implement that?

antholeole:
Your suggestion does look much nicer. It doesn't solve the first point, though.

I could just add an "extraConfig" option that takes a string. This solves the problem for Zellij, but going forward, if anyone else chooses to use KDL then we again have to work this problem out. 

I'd be happy to implement either approach: _repeatedKeys or extraConfig. Let me know

ncfavier:
> It doesn't solve the first point, though.

I was suggesting setting `_repeatableKeys` directly in the home-manager module so that the user doesn't have to do it.

We can do better, actually: the encoding of lists using `-` rather than repeated keys is not part of the KDL specification, but is rather a convention for representing [JSON in KDL](https://github.com/kdl-org/kdl/blob/main/JSON-IN-KDL.md). Zellij does not use that convention, so I think it makes sense to add a parameter to the `toKDL` generator to disable it entirely: `toKDL { repeatKeys = true; } cfg.settings`.

antholeole:
That would be very clean. So the current implementation does the following:

This nix:
```
flatItems = [ 1 2 "asdf" true null ];
```

gets transformed to this kdl:
```
flatItems 1 2 "asdf" true null
```

after this PR, if passing the "repeatKeys" param, it should be transformed to:
```
flatItems 1 
flatItems 2 
flatItems "asdf" 
flatItems true 
flatItems null
```

If this is correct I'll begin implementing. One last thing though: I think it makes more sense for toKDL to adhere to the raw KDL spec, and then if the user passes "jsonInKdl = true" then do the current behavior. While this isn't backwards compatible, it doesn't seem like anyone is using the toKdl function besides zellij (after a github search). Should I make the toKDL the default behavior, and then have the json spec be opt-in?

ncfavier:
I was thinking it should only affect lists-of-non-primitives, i.e. the behaviour of [`convertListOfNonFlatAttrsToKDL`](https://github.com/nix-community/home-manager/blob/0ee5ab611dc1fbb5180bd7d88d2aeb7841a4d179/modules/lib/generators.nix#L71-L74). So your nix would be translated to

```kdl
flatItems 1 2 "asdf" true null
nonPrimitives 1 2
nonPrimitives 3 4
```

> Should I make the toKDL the default behavior, and then have the json spec be opt-in?

I guess?

antholeole:
If it only works for non-primatives, that just doesn't work for zellij anymore. Zellij wants:

```
unbind "Ctrl k"
unbind "Ctrl f"
```

which is unmodelable if we do it for only non-primitives. I could solve this at the zellij level, but again, it does not solve the general case

ncfavier:
Oh, then that idea is dead in the water because zellij also seems to need flat-lists-of-primitives in things like [themes](https://github.com/zellij-org/zellij/blob/fda5923d64dc1388f063e78d9087c236d38555ff/example/config.kdl#L208-L218). We need to somehow discriminate between `bind`/`unbind` and `themes.*`. I guess that puts us back to my previous suggestion with hardcoded `_repeatableKeys = [ "bind" "unbind" ];` in the module.

> guess how worried are we that the general KDL case works? I can just drop this whole crusade and just implement a specific Zellij keybinds configuration and be happy

TBF that's probably better anyway: `_args` is not very user-friendly. We can maybe expose some of the internal functions from `toKDL` to avoid duplicating work.

cbr9:
Has any progress been done on this?

antholeole:
You can point to my fork of the repo. I have it working and the example I
posted works properly. It’ll take more work to upstream it so I’m not sure
if I’ll get around to it or not.

cgahr:
Isn't it easier to do the following:
```nix
{
  "myKey \"a\"" = [];
  "myKey \"b\"" = [];
}
```
converts to
```kdl
myKey "a"
myKey "b"
```
In the best case, we could convert `'` to `"` if it appears in the name (and only in the name):
```nix
{
  "myKey 'a'" = [];
  "myKey 'b'" = "doesn't convert";
}
```
converts to
```kdl
{
myKey "a"
myKey "b" "doesn't convert"
}
```
This would make writing kdl configurations much cleaner.

ElSargo:
You can already do something like:
```nix
"bind \"Alt 9\"" = { GoToTab = 9; };
```
To get: 
```kdl
bind "Alt 9" {
    GoToTab 9
}
```
Or 
```nix
"bind \"Alt s\"" = { 
    "LaunchOrFocusPlugin \"zellij:session-manager\"" = { 
        floating = true; 
    }; 
} ;
```
to get: 
```kdl
bind "a" {
    LaunchOrFocusPlugin "zellij:session-manager" {
        floating true
    }
}
```
The point is that this sort of string escaping nonsense, or worse having to do something like
```nix
"bind \"Ctrl d\"" = { "" = "Detach"; };
```
is very ugly and not at all in line with what nix code is supposed to look like.

cgahr:
Btw, I don't think this change is enough. Consider the following layout definition:
```kdl
layout {
  pane
  pane
  pane
}
```
This is a perfectly valid layout definition for zellij, however, there is no way of expressing this layout in nix using the old `toKDL` function or @antholeole newer version.

ElSargo:
I think the fundamental problem here is that kdl is based in lists of nodes whereas idiomatic nix is based on sets. Therefore we could create a list based scheme that would perfectly express kdl. However since this would not be idiomatic, I think we should try to create a set based scheme that covers 95% of use cases in a convenient way, and allows inserting lists of nodes as a kind of escape hatch for any edge cases.

ElSargo:
Take for example this kdl, I have annotaed the structure so it becomes clear how we can express this in nix:
```kdl
// Node A
// args Childern
layout {
   // Node B, no childern
    pane
    // Node C
    // arg1  arg2                          childern
    pane split_direction="vertical" {
        // Node D
        pane
        // Node E
        // arg1 arg2
        pane command="htop"
    }
}
```
As you can see each node has a list of args which can be either string ( quoted or not ) or a key value pair.
As such this could be expressed in nix with the help of a mkNode function:
```nix
mkNode [ "layout" ] [
   ( mkNode [ "pane" ]  [ ] )
   ( mkNode [ "pane" { split_direction = "vertical"; } ] [
      ( mkNode [ "pane" ] [ ]  )
      ( mkNode [  "pane" ] [ { command = "htop"; } ] )
   ]
]
```
This of course is very ugly but can express any kdl. We could then use helper functions to take idiomatic nix and transform it into this structure, while still allowing any value to be of type Node in order to still express any valid config. For example:
```nix
mkNodesWithArgs [ "bind" ] {
  "Alt 1" = [ "GoToTab" 1];
  "Alt d" = "Detach";
}
```
Could output
```kdl
bind "Alt 1" {
  "GoToTab" 1
}
bind "Alt d" {
   "Detach"
}

We could then pass the various sections of the settings attrSet to an appropriate function to handle this in a convenient and idiomatic way. I don't like how verbose the raw nix nested lists are but I don't see how we could express all valid kdl as nix.

antholeole:
Unfortunately my take here was not good enough to make this work :(. I see some other people had some much better takes on this. Feel free to open a new one!

PR #4173, opened 2023-06-27
lsd: use -A instead of -a in aliases

nim65s:
Hi,

### Description

The current `lla` alias, together with the `total-size` option try to get the size of the `..` directory, and for this has to recursively open all sibling folders. This may be super slow if some of those siblings contain too many files, and raise a ton of useless errors if some of those siblings contains non-readable files.

I'm suggesting to use `-A` instead, which will skip the obvious `.` and `..` folders.

While here, I think we could also add `llt`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


rycee:
Thanks, looks good to me. Merged now.

PR #4174, opened 2023-06-27
chromium: add support for ungoogled-chromium

networkException:
### Description

This pull request adds support for using `programs.ungoogled-chromium` and various options like `commandLineArgs`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


networkException:
Alternatively #4175 also solves the `commandLineArgs` issue with `ungoogled-chromium`

sumnerevans (changes requested):
`darwinDirs` needs to be updated as well (line 139)

networkException:
Isn't that covered by the expression in line 136? Also I don't think there are darwin builds for ungoogled-chromium

sumnerevans:
Oh, I didn't notice the `or browser` expression. Looks like ungoogled-chromium in nixpkgs is in fact only built for Linux: https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/applications/networking/browsers/chromium/browser.nix#L91

Though, there appears to be darwin builds for ungoogled-chromium at least here: https://github.com/ungoogled-software/ungoogled-chromium/releases/tag/114.0.5735.198-1, but I guess nobody on Darwin cares enough to package it.

sumnerevans:
I installed Ungoogled Chromium on my Mac test machine and it looks like the path is just `Library/Application Support/Chromium`. Let's go ahead and add that for completeness sake.

ThinkChaos (approved):
Tested this locally and it works for me!

ThinkChaos (commented):
I realized that I could configure extensions using `programs.chromium` just fine which was my goal with using this change. I had misread the existing code and thought the `extensions` options was disabled.  
Now I'm less convinced adding `programs.ungoogled-chromium` is good:
- using `programs.chromium` with `package = pkgs.ungoogled-chromium` allows configuring everything this does
- `programs.chromium` and `programs.ungoogled-chromium` both use the same config directory, which different options in HM doesn't model properly

Is the goal here to make `ungoogled-chromium` more discoverable?
Making `ungoogled-chromium` the example for `options.chromium.program` could be enough to point towards that.

networkException:
No my intention was the same as your pr (I marked it as an alternative because of that). I don't mind closing this either

folliehiyuki:
Hi! Is there anything preventing this PR from being merged?

It works and I don't want to keep a patched `home-manager` URL in my flake input forever to configure ungoogled-chromium.

merrkry:
It seems no extra work is required to support ungoogled-chromium. Would be very nice if this got merged.

stackptr:
I think this should be closed for reasons mentioned in https://github.com/nix-community/home-manager/pull/4174#pullrequestreview-1513157695

PR #4175, opened 2023-06-28
chromium: fix `commandLineArgs` to use the user specified package

ThinkChaos:
### Description

`commandLineArgs` were only being applied when the user didn't override the `package` option.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/chromium.nix
+++ b/modules/programs/chromium.nix
@@ -42,9 +42,6 @@ let
         description = ''
           List of command-line arguments to be passed to ${name}.
           </para><para>
-          Note this option does not have any effect when using a
-          custom package for <option>programs.${browser}.package</option>.
-          </para><para>
           For a list of common switches, see
           <link xlink:href="https://chromium.googlesource.com/chromium/src/+/refs/heads/main/chrome/common/chrome_switches.cc">Chrome switches</link>.
           </para><para>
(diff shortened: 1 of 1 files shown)


networkException (approved):
The note in

https://github.com/nix-community/home-manager/blob/a6c284dadef9103e3279416bd1e52a9321422fb1/modules/programs/chromium.nix#L45-L46

does not apply anymore with this patch.

ThinkChaos:
Good catch, thanks! Force pushed to remove that.

As for your PR, my goal was pretty much the same, so either solution is fine by me!  
For reference, there's already somewhat explicit support via `browser = if drvName == "ungoogled-chromium" then "chromium" else drvName;` since `ungoogled-chromium` uses the same config dir as chromium.

sumnerevans:
I think that the approach in https://github.com/nix-community/home-manager/pull/4174 is better since it allows you to easily install and configure multiple versions of chromium-based browsers in a single HM install.

networkException:
Anything against also merging this? I don't really see why commandLineArgs would need to be limited to supported browsers

sumnerevans (approved):
For some reason I read this as removing options in `options.programs.*` but looking closer this is just changing it to pass `pkgs.*` instead of `browserPkgs.*` and then overriding within `browserModule`.

PR #4177, opened 2023-06-28
i18n: Use glibcLocales from NixOS if possible (#2333)

### LINKED ISSUE #2333: LOCALE_ARCHIVE is unconfigurable and unconditional

### Issue description

home-manager unconditionally inserts `pkgs.glibcLocales` into the closure without any mechanism for disabling, configuring, or trimming it down in the way that nixos does (see `i18n.supportedLocales` and related options). This adds something like 200MB to the closure size with no way to turn it off.

(to be fair there are workarounds like using package overlays or setting disabledModules but still, this should be an option!)

### Maintainer CC

_No response_

### System information

```markdown
host os: `Linux 5.14.2, NixOS, 21.11 (Porcupine)`
```

rvl:
### Description

This will reduce the system closure size by about 200MB under NixOS by sharing the `glibcLocales` package.

When home-manager is installed on Linux without the NixOS module, all glibc locales are installed, as before.

Resolves #2333

### Checklist

- [x] Code tested through `nix develop --ignore-environment .#all` ⇒ OK.

- [x] Test cases updated/added ⇒ added `i18n-custom-locales` and modified other tests referencing `pkgs.glibcLocales`.

- [x] Commit message is formatted as per docs.

- [x] Option docs built and inspected with `nix build .#docs-html` and `nix build .#docs-manpages`.

### DIFF
--- a/modules/config/i18n.nix
+++ b/modules/config/i18n.nix
@@ -15,15 +15,16 @@
 # below for changes:
 # https://github.com/NixOS/nixpkgs/blob/nixpkgs-unstable/pkgs/development/libraries/glibc/nix-locale-archive.patch
 
-{ lib, pkgs, ... }:
+{ lib, pkgs, config, ... }:
 
 with lib;
 
 let
+  inherit (config.i18n) glibcLocales;
 
-  inherit (pkgs.glibcLocales) version;
+  inherit (glibcLocales) version;
 
-  archivePath = "${pkgs.glibcLocales}/lib/locale/locale-archive";
+  archivePath = "${glibcLocales}/lib/locale/locale-archive";
 
   # lookup the version of glibcLocales and set the appropriate environment vars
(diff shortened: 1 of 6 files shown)


modules/config/i18n.nix
```diff
@@ -36,6 +37,27 @@ let
 in {
   meta.maintainers = with maintainers; [ midchildan ];
 
+  options = {
+    i18n.glibcLocales = mkOption {
+      type = types.path;
+      visible = false;
+      internal = true;
```
ncfavier, line 44: Why not make this visible?

ncfavier (commented):
Looking good!

modules/config/i18n.nix
rvl: OK @ncfavier - I have made the option visible and enhanced the doc texts.

ncfavier (approved):

PR #4178, opened 2023-06-28
ssh-agent: init module

### LINKED ISSUE #1087: Support for starting ssh-agent?
I cannot find anywhere in the home-manager codebase that supports starting `ssh-agent`.

I found options for configuring `gpg-agent` to emulate `ssh-agent`. For the past two days, I've experimented with this option, but for now I would rather use `ssh-agent` alone.

Does home-manager currently support users who want to use `ssh-agent` without `gpg-agent`?

lheckemann:
### Description

Add a simple module for ssh-agent, fixes #1087.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -1135,6 +1135,14 @@ in
           A new modules is available: 'programs.zsh.antidote'
         '';
       }
+
+      {
+        time = "2023-06-30T14:46:22+00:00";
+        condition = config.services.ssh-agent.enable;
+        message = ''
+          A new module is available: 'services.ssh-agent'
+        '';
+      }
     ];
   };
 }
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -1132,7 +1132,7 @@ in
         time = "2023-06-17T22:18:22+00:00";
         condition = config.programs.zsh.enable;
         message = ''
-          A new modules is available: 'programs.zsh.antidote'
+          A new module is available: 'programs.zsh.antidote'
         '';
       }
 
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -306,6 +306,7 @@ let
     ./services/screen-locker.nix
     ./services/sctd.nix
     ./services/spotifyd.nix
+    ./services/ssh-agent.nix
     ./services/stalonetray.nix
     ./services/status-notifier-watcher.nix
     ./services/swayidle.nix
--- a/modules/services/ssh-agent.nix
+++ b/modules/services/ssh-agent.nix
@@ -1,0 +1,36 @@
+{ config, options, lib, pkgs, ... }:
+
+let
+
+  cfg = config.services.ssh-agent;
+
+in {
+  meta.maintainers = [ lib.maintainers.lheckemann ];
+
+  options = {
+    services.ssh-agent = {
+      enable = lib.mkEnableOption "OpenSSH private key agent";
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    home.sessionVariablesExtra = ''
+      if [[ -z "$SSH_AUTH_SOCK" ]]; then
+        export SSH_AUTH_SOCK=$XDG_RUNTIME_DIR/ssh-agent
+      fi
+    '';
+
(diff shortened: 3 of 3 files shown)


modules/lib/maintainers.nix
sumnerevans, line 1: no need to add yourself here as a maintainer if you are already a nixpkgs maintanier

modules/modules.nix
```diff
@@ -306,6 +306,7 @@ let
     ./services/screen-locker.nix
     ./services/sctd.nix
     ./services/spotifyd.nix
+    ./services/ssh-agent.nix
```
sumnerevans, line 309: this needs a news entry

modules/services/ssh-agent.nix
```diff
@@ -0,0 +1,36 @@
...
+  options = {
+    services.ssh-agent = {
+      enable = lib.mkEnableOption "OpenSSH private key agent";
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    home.sessionVariablesExtra = ''
+      if [[ -z "$SSH_AUTH_SOCK" ]]; then
+        export SSH_AUTH_SOCK=$XDG_RUNTIME_DIR/ssh-agent.sock
```
sumnerevans, line 19: We should probably use the same path that is used by nixpkgs: https://github.com/NixOS/nixpkgs/blob/c57bc5f8129dd25337ad155f22fed3e3809dfe67/nixos/modules/programs/ssh.nix#L347-L352. That way, if people have `programs.ssh.startAgent` in their nixos config, it won't conflict with this option.

```suggestion
        export SSH_AUTH_SOCK=$XDG_RUNTIME_DIR/ssh-agent
```

sumnerevans (changes requested):

lheckemann:
Applied all your suggestions, thanks!

modules/misc/news.nix
```diff
@@ -1135,6 +1135,14 @@ in
           A new modules is available: 'programs.zsh.antidote'
         '';
       }
+
+      {
+        time = "2023-06-30T14:46:22+00:00";
+        condition = config.programs.zsh.enable;
+        message = ''
+          A new modules is available: 'programs.zsh.antidote'
```
sumnerevans, line 1143:
```suggestion
          A new module is available: 'services.ssh-agent'
```

sumnerevans (changes requested):

sumnerevans:
Is there any reason to keep this as a draft? I personally would like to get this in.

modules/misc/news.nix
lheckemann: Go figure that I'd forget to adjust the actual content of the news item :upside_down_face: thanks

modules/misc/news.nix
```diff
@@ -1132,7 +1132,15 @@ in
         time = "2023-06-17T22:18:22+00:00";
         condition = config.programs.zsh.enable;
         message = ''
-          A new modules is available: 'programs.zsh.antidote'
+          A new module is available: 'programs.zsh.antidote'
```
sumnerevans, line 1135: Hmm... this is definitely a typo, but for the sake of `git blame` I don't think it's great to have this change in this PR.

sumnerevans (approved):
Going to merge this manually, since both commits need to end up on `master` separate from one another, and creating a whole separate PR just for the typo change seems excessive.

modules/misc/news.nix
```diff
@@ -1132,7 +1132,15 @@ in
...
+        '';
+      }
+
+      {
+        time = "2023-06-30T14:46:22+00:00";
+        condition = config.services.ssh-agent.enable;
```
olmokramer, line 1141: If I understand correctly, this will now only render the news item for people that already have the service enabled.

rycee: Yeah, I assume this is a mistake. My guess is that it should be `condition = hostPlatform.isLinux;` and the module itself should contain

```
assertions = [
  (lib.hm.assertions.assertPlatform "services.ssh-agent" pkgs lib.platforms.linux)
];
```

since it is Linux-only.

PR #4180, opened 2023-06-29
nixos: increase TimeoutStartSec from 1m30s to 5m

bjornfor:
### Description

I'm experiencing timeouts for home-manager-$USER.service in my NixOS VM tests. The NixOS testing framework increases the default timeout from 1m30 to 5m, but since home-manager specifies a timeout, that value wins.

There's no point in having a tight timeout value, so increase it to 5m.

### Checklist

- [ ] Code tested through `nix-shell --pure tests -A run.all` or `nix develop --ignore-environment .#all` using Flakes. I got a ton of different failing tests; it looks like `home-files/asserts/warnings.actual` is an empty file? I cannot imagine how these failures are related to this trivial change.

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/nixos/default.nix
+++ b/nixos/default.nix
@@ -49,7 +49,7 @@ in {
             User = usercfg.home.username;
             Type = "oneshot";
             RemainAfterExit = "yes";
-            TimeoutStartSec = 90;
+            TimeoutStartSec = "5m";
             SyslogIdentifier = "hm-activate-${username}";
 
             ExecStart = let


bjornfor:
CC @arcnmx and @rycee due to bd747c5a53e6f0df368066788be26e9934620bcb ("nixos: add timeout to hm-activate service").

ambroisie:
Could we just remove the timeout instead?

bjornfor:
(I got this email notification from github, but the message doesn't appear in github web UI, so pasting below.)

> Could we just remove the timeout instead?

Perhaps, but I wanted to make as little as possible change for now, to ease backporting.

Note that https://github.com/nix-community/home-manager/commit/bd747c5a53e6f0df368066788be26e9934620bcb explicitly added a timeout to `prevent a system from ever booting`. The reason being that oneshot services don't have a timeout at all by default.

But how does the system behave when waiting for home-manager-USER.service? Are you able to log in or not? For VM tests a hanging service will be caught anyway through other timeouts.

bjornfor:
Ping @rycee?

arcnmx:
90sec was just chosen to match systemd's `DefaultTimeoutStartSec`. Since this should only be a problem in extenuating circumstances I'm personally fine with increasing it. As long as a timeout of some sort is set, a user can recover a broken system eventually if they wait long enough.

> But how does the system behave when waiting for home-manager-USER.service? Are you able to log in or not? For VM tests a hanging service will be caught anyway through other timeouts.

The problem indeed is that a system can stall and never reach `multi-user.target` at all if the service never ends up starting or failing. This can prevent vital services from running and prevent a user from ever logging in. Display managers in particular can have this problem because `graphical.target` has a `Requires=/After=` dependency on `multi-user.target`. Systemd has a timeout set for most services by default for a reason, it just can't do so reasonably for `oneshot` services without knowing what it's doing.

It's unfortunate that `systemd.extraConfig` isn't `systemd.settings` or something that could be used to populate the timeout by default. Perhaps an option in the h-m nixos module might make sense, but that seems like overkill. Though if this is only ever an issue in tests, a test closure config could even just set `home-manager-${user}.serviceConfig.TimeoutStartSec` itself if necessary.

bjornfor:
@arcnmx: Thanks for your input.

Merge please?

Suyashtnt:
Is there any other way to increase the length?

arcnmx:
> Is there any other way to increase the length?

if you're looking for a workaround or alternative, I believe you should be able to `mkForce` override the timeout for the service in a test/VM nixos configuration

an actual alternative to this PR and avoiding setting `TimeoutStartSec` at all would probably mean switching to a different service type like `notify` or `forking` or something - then it should just use the default instead

rycee:
Thanks! Merged to master now. But I'm somewhat confused why the activation would take over 1.5 minutes. The idea is that the activation script should run "quickly", utmost a second or so.

bjornfor:
> Thanks! Merged to master now.

Thanks!

> But I'm somewhat confused why the activation would take over 1.5 minutes. The idea is that the activation script should run "quickly", utmost a second or so.

I wondered about that too. My theory is that systemd starts a ton of services in parallel (which starts the timeout counters), but the system is too weak to run the processes in a timely manner and home-manager activation service gets sufficiently starved.

PR #4181, opened 2023-06-30
README: Remove the pills

roberth:
This removes the Nix Pills reference, because they are not a good introduction to the Nix ecosystem, but rather a thorough explanation of many disparate things. Reading through them might give some light bulb moments for an intermediate reader, but that does not mean that they're good for a beginner.

I've also removed the mention of infinite recursion without source location. That's an old meme for a problem that has been mostly solved. Mentioning it here has two effects
 - Propagate the outdated meme
 - Make users insensitive to bad errors. Learned helplessness. That kind of thing. What we really want is for them to report bad error messages, so that they can be fixed. And they can be fixed; just report them at `NixOS/nix`

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,11 +36,9 @@ Words of warning
 ----------------
 
 Unfortunately, it is quite possible to get difficult to understand
-errors when working with Home Manager, such as infinite loops with no
-clear source reference. You should therefore be comfortable using the
-Nix language and the various tools in the Nix ecosystem. Reading
-through the [Nix Pills][] document is a good way to familiarize
-yourself with them.
+errors when working with Home Manager. You should therefore be
+comfortable using the Nix language and the various tools in the Nix
+ecosystem.
 
 If you are not very familiar with Nix but still want to use Home
 Manager then you are strongly encouraged to start with a small and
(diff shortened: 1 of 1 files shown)


roberth:
Here's the most recent testimonial, fwiw https://ersei.net/en/blog/its-nixin-time#the-part-in-which-i-make-

ncfavier (approved):
It would be good to replace it with something better. Do you feel like a good introduction to Nix (at least the parts that are relevant for HM) exists at the moment?

roberth:
> It would be good to replace it with something better.

I don't know if that's good. I think the audience can already edit config files in arbitrary configuration languages, so they would know at what point they need to learn more.

Maybe [the one on nix.dev](https://nix.dev/tutorials/first-steps/nix-language#nix-language-basics) is at least better _than Nix Pills_ but I don't think home manager users should need it, and it's not geared towards a module system use case.

rycee:
I think it is fair enough to remove this, ideal would be to have some nice alternative, like @ncfavier mentioned.

About the infinite recursion, the thing I see people do over and over again is something along the lines of

``` nix
{
  imports = [ (if pkgs.stdenv.isLinux then ./foo.nix else ./bar.nix) ];
}
```

which inevitably will lead to an infinite recursion error that (at least in version 2.13.3) doesn't really give any clear hint of what actually is the problem. If there is one error case that I wished was handled better, it's this one. I've occasionally thought about whether it wouldn't be possible to catch this in the module system itself.

roberth:
@rycee seems feasible to improve on. Either
 - return a single detailed message https://github.com/NixOS/nix/issues/8629, or
 - make the trace readable, on point, and enabled https://github.com/NixOS/nix/issues/7553

Technically the information is in there, but I can hardly blame someone for not reading the trace, considering the length and verbosity. So yeah, the improvements aren't complete yet, but "infinite recursion at unknown location" specifically should already be a thing of the past. We do still have something like "stack overflow at unknown location", but that is for diverging computations, not lazy fixpoints, and it is far more rare and not caused by the module system to my knowledge.

rycee:
@roberth Thanks, yes something that can insert additional information in the trace would probably do quite nicely. As-is, while each trace entry does have a location, none of them actually seem to give you a hint about the actual error location. With this error context, one could perhaps, at least, add some reference to the module file containing the problematic `imports` line.

PR #4188, opened 2023-07-02
less: add `historyFile` option

cbarrete:
### Description

Adds an option for setting `LESSHIST` when using `less`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


ncfavier:
Honestly less has a billion environment variables, I'm not sure it's worth making them into options. Also the man page says it's `LESSHISTFILE`, not `LESSHIST`

cbarrete:
Thanks for catching the typo.

Regarding the many environment variables, fair enough. I'd like to have everything related to a single program live in the same place though, would giving an attribute set of the less-specific environment variables (which would just be forwarded to `home.sessionVariables`) be acceptable?

ambroisie:
Arguably obsoleted by #7612 (`programs.less.config = "#env\nLESSHISTFILE=path/to/file";`).

PR #4189, opened 2023-07-02
readline: optionally use the XDG config home instead of home dir

cbarrete:
### Description

The env var `INPUTRC` can be used to change where the configuration file should be located.
We can use it to move the file out of the home directory. 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/readline.nix
+++ b/modules/programs/readline.nix
@@ -58,6 +58,13 @@ in {
       description = "Whether to include the system-wide configuration.";
     };
 
+    useXdgConfigDir = mkOption {
+      type = types.bool;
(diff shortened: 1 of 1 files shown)


ambroisie:
@rycee I think it makes sense for _most_ modules to have such an option, however I can see why e.g: `programs.gpg` might be better-off _not_ being "XDG-enabled" (it's hack-ish and not always expected by tools that make use of `gpg`...): this could be addressed on a case-by-case basis by the user.

rycee:
Closed with the merge of #5196.

PR #4192, opened 2023-07-03
xfconf: remove properties with null values

pjstadig:
### Description

In some cases, I'd like to remove xfconf properties. For example, if you set a keyboard shortcut's value to an empty string:

```
xfconf.settings.xfce4-keyboard-shortcuts."commands/custom/<Alt><Super>s" = "";
```

Rather than removing the property, this will cause an error. A property must be removed with `xfconf-query -r ...`.

I had considered adding a separate list of properties for removal:

```
xfconf.unsettings.xfce4-keyboard-shortcuts = [ "commands/custom/<Alt><Super>s" ]
```

but that would allow for both setting and unsetting the same property and doesn't allow for defaults vs. overrides, etc.

### Checklist

- [ ] Test cases updated/added.

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/xfconf.nix
+++ b/modules/misc/xfconf.nix
@@ -76,7 +76,7 @@ in {
     settings = mkOption {
       type = with types;
       # xfIntVariant must come AFTER str; otherwise strings are treated as submodule imports...
-        let value = oneOf [ bool int float str xfIntVariant ];
+        let value = nullOr (oneOf [ bool int float str xfIntVariant ]);
         in attrsOf (attrsOf (either value (listOf value))) // {
           description = "xfconf settings";
         };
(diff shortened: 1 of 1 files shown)


modules/misc/xfconf.nix
```diff
@@ -108,8 +108,11 @@ in {
         mkCommand = channel: property: value: ''
           $DRY_RUN_CMD ${pkgs.xfce.xfconf}/bin/xfconf-query \
             ${
-              escapeShellArgs
-              ([ "-n" "-c" channel "-p" "/${property}" ] ++ withType value)
+              if value != null then
+                escapeShellArgs
+                ([ "-n" "-c" channel "-p" "/${property}" ] ++ withType value)
+              else
+                escapeShellArgs ([ "-r" "-c" channel "-p" "/${property}" ])
```
ncfavier, line 115: Let's avoid repetition:

```suggestion
              escapeShellArgs
                ([ "-c" channel "-p" "/${property}" ] ++ if value == null then [ "-r" ] else [ "-n" ] ++ withType value)
```

Also, should we use `-R` with `-r`? (cf https://docs.xfce.org/xfce/xfconf/xfconf-query)

ncfavier:
cc @chuangzhu @gou4shi1

gou4shi1:
LGTM

pjstadig:
I forgot to run `./format` after adding the change to remove repetition. I've squashed everything down to a single commit.

I guess it needs a maintainer to approve the test run?

PR #4194, opened 2023-07-03
khard: Add module

olmokramer:
### Description

Add support for [khard](https://github.com/lucc/khard), integrating it with the recently merged contact infrastructure.

/cc @teto as author of the calendar and contacts infra PR.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/khard.nix
+++ b/modules/programs/khard.nix
@@ -1,0 +1,88 @@
+{ config, lib, pkgs, ... }:
+let
+  cfg = config.programs.khard;
+
+  accounts =
+    lib.filterAttrs (_: acc: acc.khard.enable) config.accounts.contact.accounts;
+
+  renderSettings = with lib.generators;
+    toINI {
+      mkKeyValue = mkKeyValueDefault rec {
+        mkValueString = v:
+          if lib.isList v then
+            lib.concatStringsSep ", " v
+          else if lib.isBool v then
+            if v then "yes" else "no"
+          else
+            v;
+      } "=";
+    };
+in {
+  meta.maintainers =
+    [ lib.hm.maintainers.olmokramer lib.maintainers.antonmosich ];
+
+  options = {
+    programs.khard = {
+      enable = lib.mkEnableOption "Khard: an address book for the Unix console";
+
+      settings = lib.mkOption {
+        type = with lib.types;
+          submodule {
+            freeformType = let primOrList = oneOf [ bool str (listOf str) ];
+            in attrsOf (attrsOf primOrList);
+
+            options.general.default_action = lib.mkOption {
+              type = str;
+              default = "list";
+              description = "The default action to execute.";
+            };
+          };
+        default = { };
+        description = ''
+          Khard settings. See
+          <https://khard.readthedocs.io/en/latest/#configuration>
+          for more information.
+        '';
+        example = lib.literalExpression ''
+          {
+            general = {
+              default_action = "list";
+              editor = ["vim" "-i" "NONE"];
+            };
+
+            "contact table" = {
+              display = "formatted_name";
+              preferred_phone_number_type = ["pref" "cell" "home"];
+              preferred_email_address_type = ["pref" "work" "home"];
+            };
+
+            vcard = {
+              private_objects = ["Jabber" "Skype" "Twitter"];
+            };
+          }
+        '';
+      };
+    };
+
+    accounts.contact.accounts = lib.mkOption {
+      type = with lib.types;
+        attrsOf (submodule {
+          options.khard.enable = lib.mkEnableOption "khard access";
+        });
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    home.packages = [ pkgs.khard ];
+
+    xdg.configFile."khard/khard.conf".text = ''
+      [addressbooks]
+      ${lib.concatMapStringsSep "\n" (acc: ''
+        [[${acc.name}]]
+        path = ${acc.local.path}
+      '') (lib.attrValues accounts)}
+
+      ${renderSettings cfg.settings}
+    '';
+  };
+}
--- a/tests/modules/programs/khard/basic_config.nix
+++ b/tests/modules/programs/khard/basic_config.nix
@@ -1,0 +1,34 @@
+{
+  accounts.contact = {
+    basePath = ".contacts";
+    accounts.test = {
+      local.type = "filesystem";
+      khard.enable = true;
+    };
+  };
+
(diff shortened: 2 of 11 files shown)


modules/accounts/contacts.nix
```diff
@@ -126,6 +126,7 @@ in {
         contactOpts
         (import ../programs/vdirsyncer-accounts.nix)
         (import ../programs/khal-accounts.nix)
+        (import ../programs/khard-accounts.nix)
```
rycee, line 129: Hmm, these imports seem like a mistake to me? I think putting something like

```
options.accounts.contact.accounts = mkOption {
  type = with types; attrsOf (submodule (import ./khard-accounts.nix));
};
```

inside the `khard.nix` file would be more idiomatic. For example, the various email related modules such as `notmuch.nix`.

modules/programs/khard.nix
```diff
@@ -0,0 +1,74 @@
...
+        submodule {
+          freeformType = let
+            prim = either bool str;
+            primOrList = either prim (listOf prim);
+          in attrsOf (attrsOf primOrList);
+        };
+      default = { };
+      description = ''
+        Khard settings. See
+        <link xlink:href="https://khard.readthedocs.io/en/latest/#configuration"/>
```
rycee, line 37:
```suggestion
        <https://khard.readthedocs.io/en/latest/#configuration>
```

since option descriptions now are interpreted as Markdown.

tests/modules/programs/khard/basic_config.nix
```diff
@@ -0,0 +1,36 @@
+{
+  config = {
```
rycee, line 2: Can simplify a bit by unwrapping the `config` field.
```suggestion
```

rycee:
Thanks! Added a few comments 🙂

modules/accounts/contacts.nix
olmokramer: Btw, `notmuch.nix` doesn't seem to import any other file. There is a `notmuch-virtual-mailbox.nix` but it's imported in `neomutt.nix`. I don't use neomutt, notmuch or the mail infra on home-manager so I'm a bit confused about what's going on here or what you are referring to.

olmokramer:
@rycee Thanks for the review. I've applied your suggested changes. Let me know what you think about the imports :)

teto:
> /cc @teto as author of the calendar and contacts infra PR

I merely merged the work of other authors. @rycee is a much better reviewer than I am. I am a bit tight on time so feel free to go ahead without me.

olmokramer:
> > /cc @teto as author of the calendar and contacts infra PR
> 
> I merely merged the work of other authors. @rycee is a much better reviewer than I am. I am a bit tight on time so feel free to go ahead without me.

Ah I didn't know you did not write that code. Fair enough and thanks for replying :)

modules/programs/khard.nix
```diff
@@ -0,0 +1,74 @@
...
+
+  options.programs.khard = {
+    enable = lib.mkEnableOption "Khard: an address book for the Unix console";
+
+    settings = lib.mkOption {
+      type = with lib.types;
+        submodule {
+          freeformType = let
+            prim = either bool str;
+            primOrList = either prim (listOf prim);
```
antonmosich, line 31: I don't think khard has any config options which "allow" bools in them, listOf str would suffice.

modules/programs/khard.nix
```diff
@@ -0,0 +1,74 @@
...
+
+  accounts =
+    lib.filterAttrs (_: acc: acc.khard.enable) config.accounts.contact.accounts;
+
+  renderSettings = with lib.generators;
+    toINI {
+      mkKeyValue = mkKeyValueDefault rec {
+        mkValueString = v:
+          if lib.isList v then
+            lib.concatStringsSep ", " (map mkValueString v)
```
antonmosich, line 13: Bools do not need to be able to be in lists, so you can remove the recursion

antonmosich (commented):
I would also consider putting the most important option(s) in their own options outside of settings (I thought about `default_action`, and did that in #4400).

olmokramer:
@antonmosich Thanks for your review! I will take a look over the weekend.

I don't have a strong opinion on the `default_action` having its own separate option, but I would like to be able to not set it at all so that I just use `khard`'s default. Or will `khard` error if you don't set it? I forgot...

antonmosich:
That's what it looks like if you don't have `default_action` set
```shell
usage: khard [-h] [-c CONFIG] [--debug] [--skip-unparsable] [-v] SUBCOMMAND ...
khard: error: Missing subcommand on command line or default action parameter in config.
```

modules/programs/khard.nix
olmokramer: Good point. I've changed the type to `oneOf [bool str (listOf str)]` now.

modules/programs/khard.nix
olmokramer: :+1: Removed the recursion.

olmokramer:
@antonmosich Thanks for checking what happens without a `default_command`. In that case I think it makes sense to add the option, yes. I've copied over your definition of the `general.default_command` option over to the `settings` option, which in my implementation was a [freeform module](https://nixos.org/manual/nixos/stable/#sec-freeform-modules) so you still set it via `programs.khard.settings.general.default_command` but if that is not set it will use the default value of `list`. What I'm not so sure about, is the `enum` listing all _currently_ valid commands, so if an update to `khard` introduces a new command, the current implementation must be updated first before users can set it as the default. So I thought about changing the type to `either (enum [ ... ]) str` so that in that case users will not have to wait for a home manager update. What are your thoughts?

olmokramer:
PS @antonmosich Would it be OK if I add you as a maintainer for the khard module?

modules/accounts/contacts.nix
olmokramer: Ok I've moved the contents of `khard-accounts.nix` into `khard.nix` now.

antonmosich:
Thank you, I would be happy being a maintainer. The freeform module is very elegant, I like that and I think that `list` is a sensible default.
I am also not really sure about the `enum` listing, I would even consider changing to `str` instead of the `either` construction.

olmokramer:
Yeah just `str` is probably fine. Changed that type and added you as a maintainer now.

antonmosich (approved):

rycee:
Thanks for the nice work! Merged to master now 🙂

olmokramer:
Thanks @rycee!

PR #4196, opened 2023-07-04
aerc: add assertion to limit per-account extraConfig to UI config

dryya:
### Description

The aerc configuration file `aerc.conf` can contain 10 sections, but only the UI section supports what the aerc manual calls contextual configuration. This works by appending to the section heading either `:account=name` or `:folder=bar`. It's supported in `config.accounts.email.accounts.<name>.aerc.extraConfig.*.*`, as the aerc-accounts module applies a
function to add `<name>` to section headings.
However, this means home-manager will generate files like:
```
[general:account=default]
log-file=...
```
and the options will not be recognized by aerc.

This change only adds the account name to the UI section by adding a `attrsets.filterAttrs` check to `mkAccountConfig`. This could be confusing if a user only wants to apply their configuration to one account, and a possible alternative could be to change this option to something like `config.accounts.email.accounts.<name>.aerc.uiConfig.*`.
The error message about setting `unsafe-accounts-conf = true` has also been updated, to clarify it should be set
in the global `programs.aerc.extraConfig` section.

### Checklist

### DIFF
--- a/modules/programs/aerc-accounts.nix
+++ b/modules/programs/aerc-accounts.nix
@@ -53,9 +53,9 @@ in {
           example =
             literalExpression ''{ source = "maildir://~/Maildir/example"; }'';
           description = ''
-            Extra config added to the configuration of this account in
+            Extra config added to the configuration section for this account in
             <filename>$HOME/.config/aerc/accounts.conf</filename>.
-            See aerc-config(5).
+            See <citerefentry><refentrytitle>aerc-accounts</refentrytitle><manvolnum>5</manvolnum></citerefentry>.
           '';
         };
 
@@ -66,18 +66,20 @@ in {
             ''{ messages = { d = ":move ''${folder.trash}<Enter>"; }; }'';
           description = ''
             Extra bindings specific to this account, added to
-            <filename>$HOME/.config/aerc/accounts.conf</filename>.
-            See <citerefentry><refentrytitle>aerc-config</refentrytitle><manvolnum>5</manvolnum></citerefentry>.
+            <filename>$HOME/.config/aerc/binds.conf</filename>.
+            See <citerefentry><refentrytitle>aerc-binds</refentrytitle><manvolnum>5</manvolnum></citerefentry>.
           '';
         };
 
         extraConfig = mkOption {
           type = confSections;
           default = { };
-          example = literalExpression "{ ui = { sidebar-width = 42; }; }";
+          example = literalExpression "{ ui = { sidebar-width = 25; }; }";
           description = ''
-            Extra config specific to this account, added to
-            <filename>$HOME/.config/aerc/aerc.conf</filename>.
+            Config specific to this account, added to <filename>$HOME/.config/aerc/aerc.conf</filename>.
+            Aerc only supports per-account UI configuration.
+            For other sections of <filename>$HOME/.config/aerc/aerc.conf</filename>,
+            use <literal>.programs.aerc.extraConfig</literal>.
             See <citerefentry><refentrytitle>aerc-config</refentrytitle><manvolnum>5</manvolnum></citerefentry>.
           '';
         };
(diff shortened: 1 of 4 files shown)


modules/programs/aerc-accounts.nix
```diff
@@ -221,7 +221,9 @@ in {
     // account.aerc.extraAccounts;
 
   mkAccountConfig = name: account:
-    mapAttrNames (addAccountName name) account.aerc.extraConfig;
+    (mapAttrNames (addAccountName name)
+      (lib.attrsets.filterAttrs (k: v: k == "ui") account.aerc.extraConfig))
+    // lib.attrsets.filterAttrs (k: v: k != "ui") account.aerc.extraConfig;
```
ncfavier, line 226: For future reference: use `lib.partition` to do this in a single pass.

ncfavier (changes requested):
I don't know much about aerc, but it seems to me that if the only per-account configuration allowed is `ui`, then we should restrict the per-account `extraConfig` to that. Otherwise the other settings will basically end up in the global `extraConfig`, possibly causing conflicts.

So I would either add an assertion, or make the type of the per-account `extraConfig` a submodule with just an `ui` option.

ncfavier:
For example:

```nix
type = types.submodule {
  options.ui = mkOption {
    type = confSection;
    description = "Per-account UI configuration...";
  };
};
```

modules/programs/aerc-accounts.nix
ncfavier: Actually I read too fast and thought this was a list. If you're just extracting the `ui` key from an attrset, you can just do something like this

```nix
builtins.removeAttrs account.aerc.extraConfig [ "ui" ] //
optionalAttrs (account.aerc.extraConfig?ui) {
  ${addAccountName name "ui"} = account.aerc.extraConfig.ui;
}
```

ncfavier: or

```nix
mapAttrNames (n: if n == "ui" then addAccountName name n else n) account.aerc.extraConfig
```

ncfavier:
So it seems like we need a default value for both `extraConfig` and `extraConfig.ui`. Setting both to `{}` means that the per-account `ui` section will always be generated, which may or may not be a problem.

As alternatives we could make it default to `null` and only generate it if it's `!= null`, or fall back to the assertion I mentioned.

ncfavier:
Silently discarding non-`ui` sections doesn't seem great

modules/programs/aerc-accounts.nix
```diff
@@ -66,18 +66,20 @@ in {
...
           default = { };
-          example = literalExpression "{ ui = { sidebar-width = 42; }; }";
+          example = literalExpression "{ ui = { sidebar-width = 25; }; }";
           description = ''
-            Extra config specific to this account, added to
-            <filename>$HOME/.config/aerc/aerc.conf</filename>.
+            Config specific to this account, added to <filename>$HOME/.config/aerc/aerc.conf</filename>.
+            Aerc only supports per-account UI configuration.
+            For other sections of <filename>$HOME/.config/aerc/aerc.conf</filename>,
+            use <literal>.programs.aerc.extraConfig</literal>.
```
ncfavier, line 82:
```suggestion
            use <literal>programs.aerc.extraConfig</literal>.
```

modules/programs/aerc.nix
```diff
@@ -162,16 +162,28 @@ in {
...
 
+    assertions = [{
+      assertion = let
+        extraConfigSections = (unique (flatten
+          (mapAttrsToList (_: v: attrNames v.aerc.extraConfig) aerc-accounts)));
+      in extraConfigSections == [ ] || extraConfigSections == [ "ui" ];
+      message = ''
+        Only the UI section of <filename>$XDG_CONFIG_HOME/aerc/aerc.conf</filename>
+        supports contextual (per-account configuration. Please move any other
+        configuration to <literal>programs.aerc.extraConfig</literal>.
```
ncfavier, line 183: Assertion messages are plain text, sadly

```suggestion
        Only the ui section of $XDG_CONFIG_HOME/aerc/aerc.conf
        supports contextual (per-account) configuration. Please move any other
        configuration to programs.aerc.extraConfig.
```

ncfavier (approved):

PR #4208, opened 2023-07-07
 unison: Allow using same option multiple times

simfu:
### Description

 Unison supports the same option to be given several times as a command
 line argument (e.g. unison -path xxx -path yyy).

 This commit adds Home Manager support for this by allowing a list of
 strings to be given to services.unison.pairs.<name>.commandOptions values.

### Checklist

(new to Nix and first PR, did not manage to fill all the checklist yet)

- [ x] Commit messages are formatted like

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@pacien

### DIFF
--- a/modules/services/unison.nix
+++ b/modules/services/unison.nix
@@ -22,7 +22,7 @@ let
       };
 
       commandOptions = mkOption rec {
-        type = with types; attrsOf str;
+        type = with types; attrsOf (either str (listOf str));
         apply = mergeAttrs default;
         default = {
           repeat = "watch";
@@ -35,6 +35,8 @@ let
         description = ''
           Additional command line options as a dictionary to pass to the
           <literal>unison</literal> program.
+          </para><para>
+          Use a list of strings to declare the same option multiple times.
           </para><para>
           See
           <citerefentry>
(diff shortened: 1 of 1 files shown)


modules/services/unison.nix
```diff
@@ -36,6 +36,8 @@ let
           Additional command line options as a dictionary to pass to the
           <literal>unison</literal> program.
           </para><para>
+          Use a list of strings to declare the same option multiple time.
```
ncfavier, line 39:
```suggestion
          Use a list of strings to declare the same option multiple times.
```

ncfavier (approved):

modules/services/unison.nix
```diff
@@ -60,7 +62,7 @@ let
     };
   };
 
-  serialiseArg = key: val: escapeShellArg "-${key}=${escape [ "=" ] val}";
+  serialiseArg = key: val: concatStringsSep " " (forEach (toList val) (x: escapeShellArg "-${key}=${escape [ "=" ] x}"));
```
pacien, line 65: Please wrap that line so that it does not exceed 80 chars.

pacien (approved):
Patch looks good to me.

modules/services/unison.nix
ncfavier: (`./format` will do that for you)

PR #4224, opened 2023-07-11
dunst: add systemd target option

rhoriguchi:
### Description

Needed so it starts in sway environments.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/dunst.nix
+++ b/modules/services/dunst.nix
@@ -64,6 +64,15 @@ in {
         description = "Package providing {command}`dunst`.";
       };
 
+      systemdTarget = mkOption {
+        type = types.str;
+        default = "graphical-session.target";
+        example = "sway-session.target";
+        description = ''
+          Systemd target to bind to.
+        '';
+      };
+
       configFile = mkOption {
         type = with types; either str path;
         default = "${config.xdg.configHome}/dunst/dunstrc";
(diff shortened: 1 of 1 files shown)


rycee:
Not sure this should be necessary. Seems to me that the sway session should bind to `graphical-session.target`, similar to `hm-graphical-session.target`.

For example, I'm using Hyprland and Dunst at the moment and it's working just fine. I have

``` console
$ systemctl cat --user hyprland-session.target 
# /home/rycee/.config/systemd/user/hyprland-session.target
[Unit]
After=graphical-session-pre.target
BindsTo=graphical-session.target
Description=Hyprland compositor session
Documentation=man:systemd.special(7)
Wants=graphical-session-pre.target
```

Looking at `sway.nix` it seems it is already doing that?

https://github.com/nix-community/home-manager/blob/a6d1d954b81caf4c9291b8ac35452fef842f289b/modules/services/window-managers/i3-sway/sway.nix#L494-L505

So I'm not certain why you're not seeing dunst begin started, I'm guessing there might be some other issue at play?

rycee:
Ah, might be the dbus socket activation not working properly…

rhoriguchi:
mhh interesting, I'm using also hyprland but launched with sddm with autologin. <https://github.com/rhoriguchi/nixos-setup/blob/ad531aa4338e48541e751cb91af71e79779d5422/modules/profiles/hyprland.nix>. I don't have `wayland.windowManager.sway` enabled. So in order to get this to work I had to manually set `Install.WantedBy`, same for `waybar` and `swayidle`, but with the value `hyprland-session.target`.

rycee:
Quick question, does it work for you if you start hyprland from the Linux console? That is, not from sddm.

PR #4228, opened 2023-07-12
nixpkgs-disabled: Allow modules to extend package set

rcerc:
### Description

Allow Home Manager modules to extend the package set when using `modules/misc/nixpkgs-disabled.nix` by re-importing the original package set with the new configuration and overlays added. Since this is only done when `nixpkgs.{config,overlays}` is specified, the defaults incur no overhead.

This change has been working for me for a while when using the NixOS module with `useGlobalPkgs = true`. However, I have not tried cross-compilation, for example.

Please let me know if there is a better way to implement this than what I propose.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@thiagokokada

### DIFF
--- a/modules/misc/nixpkgs-disabled.nix
+++ b/modules/misc/nixpkgs-disabled.nix
@@ -2,6 +2,7 @@
   config,
   lib,
   pkgs,
+  superPkgs,
   ...
 }:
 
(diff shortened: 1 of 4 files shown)


emilazy:
This would seem to contradict the whole purpose of using `nixpkgs-disabled`, i.e. to ensure that the specified package set is used and Nixpkgs is not re-imported?

rcerc:
The idea was to allow each user configuration to extend the package set passed to the NixOS module instead of having to set `useGlobalPkgs = false` and reference the correct version of Nixpkgs within each configuration, or having to provide the combined overlays for all the users in the system-wide NixOS configuration. Maybe what I am trying to do is already possible, but I am not aware of how?

emilazy:
It looks like Home Manager imports Nixpkgs from the global `pkgsPath` module argument. You could probably set `useGlobalPkgs = false` and add a module like this to the Home Manager configs:

```nix
{ lib, ... }:

{
  _module.args.pkgsPath = correctNixpkgs.path;
  nixpkgs = { inherit (correctNixpkgs) config overlays system; };
}
```

rcerc:
Thank you. I'll consider using this personally. I will still leave the PR open in case giving the `nixpkgs.{config,overlays}` options this functionality is thought to be convenient.

emilazy:
I think that at least if this were implemented, the option would have to be renamed from `useGlobalPkgs` (since the Home Manager configurations would no longer be using the global `pkgs`), and the module from `nixpkgs-disabled` (since it's actually just reimplementing most of the `nixpkgs` module at that point). A toggle to set defaults in the `nixpkgs` module according to the host `pkgs` would make more sense to achieve this use case, I believe.

emilazy:
Maybe an alternative (not replacement) to `useGlobalPkgs` could be introduced called `inheritGlobalPkgs`, and the implementation moved to the `nixpkgs` module conditioned on that. (e.g., setting `inheritGlobalPkgs` at the entry point could inject `{ nixpkgs.superPkgs = pkgs; }`, and the `nixpkgs` module could look at that if it's non-null.)

I would personally be unhappy to see `useGlobalPkgs`/`nixpkgs-disabled` go. NixOS added a comparable module recently for good reasons: https://github.com/NixOS/nixpkgs/commit/e5db80ae487b59b4e9f950d68983ffb0575e26c6. I would like to preserve the behaviour that you can set one `pkgs` and ensure that Nixpkgs is not imported again. (It is not just performance; there are also annoying other things with mismatched Nixpkgs especially if you are injecting stuff or want accurate version information.)

rcerc:
OK, I merged the modules and provided an `inheritGlobalPkgs` option (as well as `readOnly`). @emilazy, if you have time, could you please check that I did not misunderstand anything or write something silly?

On a tangential note, I wonder why `useNixpkgsModule` is `true` by default; `home-manager/home-manager.nix` already imports Nixpkgs when using the Home Manager CLI, and then Nixpkgs is imported again in `modules/misc/nixpkgs.nix`. Does this not come with a performance penalty? If it does, had it been done to allow setting `nixpkgs.{config,overlays}`?

PR #4230, opened 2023-07-13
xplr: add module

NotAShelf:
### Description

Adds a home-manager module for [xplr](https://github.com/sayanarijit/xplr)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/xplr.nix
+++ b/modules/programs/xplr.nix
@@ -1,0 +1,72 @@
+{ config, lib, pkgs, ... }:
+let
+  inherit (lib)
+    concatStringsSep types mkIf mkOption mkEnableOption mkPackageOption
+    literalExpression;
+
+  cfg = config.programs.xplr;
+
+  initialConfig = ''
+    version = '${cfg.package.version}'
+  '';
+
+  # We provide a default version line within the configuration file, which is
+  # obtained from the package's attributes. Merge the initial configFile, a
+  # mapped list of plugins and then the user defined configuration to obtain the
+  # final configuration.
+  pluginPath = if cfg.plugins != [ ] then
+    (''
+      package.path=
+    '' + (concatStringsSep " ..\n"
+      (map (p: ''"${p}/init.lua;${p}/?.lua;"'') cfg.plugins)) + ''
+         ..
+        package.path
+      '')
+  else
+    "\n";
+
+  configFile = initialConfig + pluginPath + cfg.extraConfig;
+in {
+  meta.maintainers = [ lib.maintainers.NotAShelf ];
+
+  options.programs.xplr = {
+    enable = mkEnableOption "xplr, terminal UI based file explorer";
+
+    package = mkPackageOption pkgs "xplr" { };
+
+    plugins = mkOption {
+      type = with types; nullOr (listOf (either package str));
+      default = [ ];
+      defaultText = literalExpression "[]";
+      description = ''
+        Plugins to be added to your configuration file.
+
+        Must be a package, an absolute plugin path, or string to be recognized
+        by xplr. Paths will be relative to
+        {file}`$XDG_CONFIG_HOME/xplr/init.lua` unless they are absolute.
+      '';
+    };
+
+    extraConfig = mkOption {
+      type = types.lines;
+      default = "";
+      example = literalExpression ''
+        require("wl-clipboard").setup {
+          copy_command = "wl-copy -t text/uri-list",
+          paste_command = "wl-paste",
+          keep_selection = true,
+        }
+      '';
+      description = ''
+        Extra xplr configuration.
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    xdg.configFile."xplr/init.lua".source =
+      pkgs.writeText "init.lua" configFile;
+  };
+}
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -1179,6 +1179,13 @@ in
           A new module is available: 'services.git-sync'.
         '';
       }
+
+      {
+        time = "2023-08-15T15:45:45+00:00";
+        message = ''
+          A new module is available: 'programs.xplr'.
+        '';
+      }
     ];
   };
 }


modules/programs/xplr.nix
```diff
@@ -0,0 +1,68 @@
...
+      default = [ ];
+      defaultText = literalExpression "[]";
+      description = mdDoc ''
+        Plugins to be added to your configuration file. Must be a package, an absolute plugin path, or string
+        to be recognized by xplr. Paths will be relative to $XDG_CONFIG_HOME/xplr/init.lua unless they are absolute.
+      '';
+    };
+
+    # TODO: rename, this is the main configuration
+    config = mkOption {
```
fufexan, line 45:
```suggestion
    extraConfig = mkOption {
```
Since it isn't configured in nix but in lua, it should be named `extraConfig`.

fufexan (changes requested):

modules/programs/xplr.nix
```diff
@@ -0,0 +1,67 @@
...
+    version = '${cfg.package.version}'
+  '';
+  # we provide a default version line within the configuration file, which is obtained from the package's attributes
+  # merge the initial configFile, a mapped list of plugins and then the user defined configuration to obtain the final configuration
+  pluginPath = if cfg.plugins != [ ] then
+    (''
+      package.path=
+    '' + (concatStringsSep " ..\n" (map (p: ''"${p}/init.lua;"'') cfg.plugins))
+      + ''
+         ..
```
sayanarijit, line 19:
```suggestion
      package.path=
    '' + (concatStringsSep " ..\n" (map (p: ''"${p}/init.lua;${p}/?.lua;"'') cfg.plugins))
      + ''
         ..
```

Some plugins can contain multiple Lua packages.

sayanarijit (changes requested):

sayanarijit:
Also, xplr prefers the latest version of coreutils to be in $PATH on macOS (not a strict requirement though). Not sure how to add that here.

modules/programs/xplr.nix
sayanarijit: I think, if we allow string based extraConfig, we shouldn't add the global config version automatically, rather, we should let the users define it. Something like extraConfig.version (required).

This way, if there's a breaking change, xplr will redirect users to https://xplr.dev/en/upgrade-guide with a friendlier error message, and remind them to update the user config as well as the version.

It isn't required if we use `config` (nix), since, then we can go nix's rules on breaking changes.

sayanarijit: But then again we can use nix's notification system to do the same during system update. Not sure what'd be the best practice here.

NotAShelf:
> Also, xplr prefers the latest version of coreutils to be in $PATH on macOS (not a strict requirement though). Not sure how to add that here.

We can use wrapProgram to add coreutils to the package's path, should be an easy fix (though, shouldn't that be done on the package's end?)

sayanarijit:
> though, shouldn't that be done on the package's end?

Not sure what you mean by package's end, but xplr as a tool tries to remain as platform independent as possible. macOS's coreutils is painfully outdated, and it remains so for licensing and/or compatibility reasons. Developers are expected to install the latest version themselves and set up a custom $PATH to make the binaries available.

So, as a compromise, xplr won't fail miserably if coreutils is outdated, but there might be a few featured missing (e.g. creating files inside non-existent directories e.g. `some/non/existent/dir/file.txt` won't be supported).

sayanarijit:
For ref: see https://github.com/sayanarijit/xplr/blob/219ee6815284bcc732fe366a07414bf11af99613/src/init.lua#L1740

NotAShelf:
> Not sure what you mean by package's end

I mean the package derivation in nixpkgs, there should be a post-build phase in which the resulting xplr is wrapped with the latest version of coreutils. Wrapping a package, unless for configuration purposes, generally don't go into a module.

sayanarijit:
Ah, got it, thanks for clarifying. Makes sense.

n3oney (approved):

modules/lib/maintainers.nix
```diff
@@ -215,6 +214,13 @@
       fingerprint = "8F87 050B 0F9C B841 1515  7399 B784 3F82 3355 E9B9";
     }];
   };
+  NotAShelf = {
+    name = "NotAShelf";
+    email = "raf@notashelf.dev";
+    github = "NotAShelf";
+    githubId = 62766066;
+    matrix = "@raf:notashelf.dev";
+  };
```
eclairevoyant, line 223:
```suggestion
```

You're already a nixpkgs maintainer, and you're not using this entry here anyway

modules/lib/maintainers.nix
```diff
@@ -5,7 +5,6 @@
 # are expected to be follow the same format as described in [1].
 #
 # [1] https://github.com/NixOS/nixpkgs/blob/fca0d6e093c82b31103dc0dacc48da2a9b06e24b/maintainers/maintainer-list.nix#LC1
-
```
eclairevoyant, line 8: undelete

modules/programs/xplr.nix
```diff
@@ -0,0 +1,65 @@
...
+      '')
+  else
+    "\n";
+  configFile = initialConfig + pluginPath + cfg.extraConfig;
+in {
+  meta.maintainers = [ maintainers.notashelf ];
+  options.programs.xplr = {
+    enable = mkEnableOption "xplr, terminal UI based file explorer";
+
+    package = mkPackageOptionMD pkgs "xplr" { };
```
eclairevoyant, line 30:
```suggestion
    package = mkPackageOptionMD pkgs "xplr" { nullable = true; };
```

modules/programs/xplr.nix
```diff
@@ -0,0 +1,65 @@
...
+
+    plugins = mkOption {
+      type = with types; nullOr (listOf (either package str));
+      default = [ ];
+      defaultText = literalExpression "[]";
+      description = mdDoc ''
+        Plugins to be added to your configuration file. Must be a package, an absolute plugin path, or string
+        to be recognized by xplr. Paths will be relative to $XDG_CONFIG_HOME/xplr/init.lua unless they are absolute.
```
eclairevoyant, line 38: keep lines to 80 chars (https://nix-community.github.io/home-manager/index.html#sec-code-style)

eclairevoyant (changes requested):

modules/programs/xplr.nix
```diff
@@ -0,0 +1,65 @@
...
+    '' + (concatStringsSep " ..\n" (map (p: ''"${p}/init.lua;"'') cfg.plugins))
+      + ''
+         ..
+        package.path
+      '')
+  else
+    "\n";
+  configFile = initialConfig + pluginPath + cfg.extraConfig;
+in {
+  meta.maintainers = [ maintainers.notashelf ];
```
eclairevoyant, line 26:
```suggestion
  meta.maintainers = [ lib.maintainers.notashelf ];
```

modules/programs/xplr.nix
NotAShelf: the most I can truncate it to is roughly 130 characters

"Plugins to be added to your configuration file. Paths will be relative to $XDG_CONFIG_HOME/xplr/init.lua unless they are absolute" 

Any less and some meaning will be lost.

modules/programs/xplr.nix
NotAShelf: What for?

eclairevoyant: Not every HM user uses nix to manage packages, and this allows setting to `null` directly instead of having to create a dummy package

modules/programs/xplr.nix
eclairevoyant: Not asking you to shorten the desc, just add line breaks.

modules/programs/xplr.nix
eclairevoyant: TBH this should even be set treewide, but no one's got around to it yet - see #3375

modules/programs/xplr.nix
sayanarijit: @NotAShelf I think you missed this...

eclairevoyant:
@rycee why was nullable = true removed?

sayanarijit:
Unresolved: https://github.com/nix-community/home-manager/pull/4230#discussion_r1292904130

modules/programs/xplr.nix
NotAShelf: No, I just can't work with git. History got messed up, again.

NotAShelf:
Should be good now.

sayanarijit (approved):

rycee:
@NotAShelf Sorry, I was on the train and was hoping to merge before arriving but had to interrupt right before. The package cannot be `null` since we typically don't allow that and

```
  initialConfig = ''
    version = '${cfg.package.version}'
  '';
```

would not work if package is nulled.

NotAShelf:
> @NotAShelf Sorry, I was on the train and was hoping to merge before arriving but had to interrupt right before. The package cannot be `null` since we typically don't allow that and
> 
> ```
>   initialConfig = ''
>     version = '${cfg.package.version}'
>   '';
> ```
> 
> would not work if package is nulled.

Thanks for the clarification!

As @eclairevoyant previously stated, users might want to use the home-manager (and by extension, nix) to manage packages so maybe it would be a better idea to make the version a configuration option rather than trying to get it from the selected package?

eclairevoyant (approved):

rycee:
In any case, I've merged the PR now. Thanks for the contribution 🙂

trueNAHO:
As a Nix beginner, I am having issues installing this home-manager module.

Although, [`xplr`](https://github.com/nix-community/home-manager/commit/b84767a145e04d07ef699d266161cda605f48b11) was added before [`pqiv`](https://github.com/nix-community/home-manager/commit/9db5b89f40736943d017a761da044479044be182), only `pqiv` is mentioned at https://nix-community.github.io/home-manager/options.html. Furthermore,

```nix
programs.xplr.enable = true;
```

causes the following error upon running `home-manager switch`:

```text
error:
       … while evaluating a branch condition

         at /nix/store/rh9r2vzd2v4q5fz7dqp8vyp44dan6r7n-source/lib/lists.nix:57:9:

           56|       fold' = n:
           57|         if n == len
             |         ^
           58|         then nul

       … while calling the 'length' builtin

         at /nix/store/rh9r2vzd2v4q5fz7dqp8vyp44dan6r7n-source/lib/lists.nix:55:13:

           54|     let
           55|       len = length list;
             |             ^
           56|       fold' = n:

       error: The option `programs.xplr' does not exist. Definition values:
       - In `/nix/store/kmgrav0hxa4h1f2vp5jwxz96hv1c4n3r-source/home/programs/xplr':
           {
             enable = true;
           }
```

For reference, my `flake.lock` points at https://github.com/nix-community/home-manager/tree/19c6a4081b14443420358262f8416149bd79561a, which is several commits ahead of the [commit that added the `xplr` module to home-manager](https://github.com/nix-community/home-manager/commit/b84767a145e04d07ef699d266161cda605f48b11):

```json
{
  "nodes": {
    "home-manager": {
      "inputs": {
        "nixpkgs": [
          "nixpkgs"
        ]
      },
      "locked": {
        "lastModified": 1694134858,
        "narHash": "sha256-fG/ESauOGmiojKlpJG8gB62dJa5Wd+ZIuiDMKK/HD3g=",
        "owner": "nix-community",
        "repo": "home-manager",
        "rev": "19c6a4081b14443420358262f8416149bd79561a",
        "type": "github"
      },
      "original": {
        "owner": "nix-community",
        "repo": "home-manager",
        "type": "github"
      }
    },
    "nixpkgs": {
      "locked": {
        "lastModified": 1693985761,
        "narHash": "sha256-K5b+7j7Tt3+AqbWkcw+wMeqOAWyCD1MH26FPZyWXpdo=",
        "owner": "nixos",
        "repo": "nixpkgs",
        "rev": "0bffda19b8af722f8069d09d8b6a24594c80b352",
        "type": "github"
      },
      "original": {
        "owner": "nixos",
        "ref": "nixos-unstable",
        "repo": "nixpkgs",
        "type": "github"
      }
    },
    "root": {
      "inputs": {
        "home-manager": "home-manager",
        "nixpkgs": "nixpkgs",
      }
    }
  },
  "root": "root",
  "version": 7
}
```

eclairevoyant:
Looks like the entry was not added to https://github.com/nix-community/home-manager/blob/master/modules/modules.nix#L226

NotAShelf:
whoops

PR #4232, opened 2023-07-14
zellij: generate completions with shell integration options

DavSanchez:
### Description

Generate completions (and some aliases) for `zellij` with the available shell integration options (e.g. `enableZshIntegration` etc). [Link to documentation](https://zellij.dev/documentation/controlling-zellij-through-cli.html#completions).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@mainrs

### DIFF
--- a/modules/programs/zellij.nix
+++ b/modules/programs/zellij.nix
@@ -69,15 +69,18 @@ in {
       };
 
     programs.bash.initExtra = mkIf cfg.enableBashIntegration (mkOrder 200 ''
+      eval "$(zellij setup --generate-completion bash)"
       eval "$(zellij setup --generate-auto-start bash)"
     '');
 
     programs.zsh.initExtra = mkIf cfg.enableZshIntegration (mkOrder 200 ''
+      eval "$(zellij setup --generate-completion zsh)"
       eval "$(zellij setup --generate-auto-start zsh)"
     '');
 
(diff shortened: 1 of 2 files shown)


ambroisie:
Isn't this [already done in the package](https://github.com/NixOS/nixpkgs/blob/12303c652b881435065a98729eb7278313041e49/pkgs/tools/misc/zellij/default.nix#L51-L54), and should instead be activated through e.g: `programs.zsh.enableCompletion` and `environment.pathsToLink = [ "/share/zsh" ]`?

DavSanchez:
Hi @ambroisie , thanks for reviewing. I actually missed these lines when inspecting the derivation, and since the aliases do not seem to be added to my shell sessions I thought the completions weren't being actually added.

I'll troubleshoot this in case it's an issue with my config and update/close this accordingly.

DavSanchez:
Alright, I think I was getting confused by the zsh autoloading. Alias functions become available only when I try to autocomplete a full `zellij` command for the first time.

It would probably make sense to split the alias function definitions from the completions so they can be separately added to `.zshrc`, but this is to be done from Zellij's side before revisiting enabling the aliases via Nix. I'm closing this and will reopen when it gets sorted. Thanks!

PR #4235, opened 2023-07-15
git: add options for multiple signing methods (GPG, SSH, X.509)

### LINKED ISSUE #4221: Add option for SSH commit signing for Git config
### Description
Currently, the settings for Git in HM only supports GPG out of the box. This isn't super annoying for people who use SSH signing like me, but a dedicated option would be nice.

For example, here's how my config looks like right now:
```nix
{pkgs, ...}: {
  programs.git = {
    enable = true;
    userName = "Leah";
    userEmail = "hi@pluie.me";
    signing = {
      signByDefault = true;
      key = "ssh-ed25519 ...";
    };

    extraConfig = {
      gpg = {
        format = "ssh";
        ssh.program = "...";
      };
    };
  };
}
```

How it ideally should look like:
```nix
{pkgs, ...}: {
  programs.git = {
    enable = true;
    userName = "Leah";
    userEmail = "hi@pluie.me";
    signing = {
      signByDefault = true;
      key = "ssh-ed25519 ...";
      
      ssh = {
      	enable = true; # gpg.format → "ssh";
      	program = "..."; # gpg.ssh.program
      };
    };
  };
}
```

pluiedev:
### Description
Adds corresponding options to all signing methods Git current supports. As of this commit, that includes signing with GPG, SSH and X.509.

The existing options for configuring GPG, `programs.git.signing.gpgPath` and `programs.git.signing.key`, have been moved to `programs.git.signing.gpg.program` and `programs.git.signing.gpg.key` respectively, and the original options have been deprecated.

Fixes #4221

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@rycee

### DIFF
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -1165,6 +1165,19 @@ in
           A new module is available: 'services.swayosd'
         '';
       }
+
+      {
+        time = "2023-07-15T05:44:07+00:00";
+        condition = config.programs.git.enable && config.programs.git.signing != null;
+        message = ''
+          The options 'programs.git.signing.gpgPath' and 'programs.git.signing.key'
+          have been deprecated and replaced by new options specific to each signing method.
+
+          Supported signing methods include GnuPG, SSH and X.509, and they're 
+          configured and enabled under 'programs.git.signing.gpg',
+          'programs.git.signing.ssh', and 'programs.git.signing.x509' respectively.
+        '';
+      }
     ];
   };
 }
--- a/modules/programs/git.nix
+++ b/modules/programs/git.nix
@@ -59,28 +59,90 @@ let
     in attrsOf supersectionType;
 
   signModule = types.submodule {
-    options = {
-      key = mkOption {
-        type = types.nullOr types.str;
-        description = ''
-          The default GPG signing key fingerprint.
-          </para><para>
-          Set to <literal>null</literal> to let GnuPG decide what signing key
-          to use depending on commit’s author.
-        '';
-      };
+    imports = [
+      (mkRenamedOptionModule [ "gpgPath" ] [ "gpg" "program" ])
+      (mkRenamedOptionModule [ "key" ] [ "gpg" "key" ])
+    ];
 
+    options = {
       signByDefault = mkOption {
         type = types.bool;
         default = false;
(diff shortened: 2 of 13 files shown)


ncfavier:
This PR feels a bit over-engineered to me. In light of [RFC 42](https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md) I'd like to keep the number of extra options to a minimum, and the interface simple. This seems like it just adds too many combinations of options to keep track of. (Also would run into https://github.com/NixOS/nixpkgs/issues/96006)

I think really all that is needed for https://github.com/nix-community/home-manager/issues/4221 is maybe adding a `signing.format` alias for `extraConfig.gpg.format`.

pluiedev:
> tra options to a minimum, and the interface simple. This seems like it just adds too many combinations of options to keep

Makes sense - originally I had a much bigger list of options but then I sought to cut them back after reading RFC 42. I think keeping `programs.git.signing.key` where it originally belongs makes sense, and adding a new `format` option makes sense too.

The only question in my mind right now is whether we should rename the `gpgPath` option - I use SSH signing and so to me, I think specifying the signer binary with an option deliberately made for GPG is quite weird. I would probably carry out my current change, which is to rename `gpgPath` to just `program`, but maybe there's other things I've overlooked

ncfavier:
Yes, we should rename `gpgPath` to `program` and make it feed into `gpg.${signing.format}.program`. This won't be a clean migration because of https://github.com/NixOS/nixpkgs/issues/96006 though, so either we add a top-level assertion or users will just have to figure it out.

pluiedev:
> Yes, we should rename `gpgPath` to `program` and make it feed into `gpg.${signing.format}.program`. This won't be a clean migration because of [NixOS/nixpkgs#96006](https://github.com/NixOS/nixpkgs/issues/96006) though, so either we add a top-level assertion or users will just have to figure it out.

Given by the number of people using commit signing... I don't think it's good to break stuff and just let people "figure it out". I think the best way to approach it would be to add the top-level assertion and prompt users to migrate. (I'm not sure if they'll get the news before everything breaks but... who knows)

pluiedev:
Closing in favor of a new PR which should be *much* simpler than this

PR #4236, opened 2023-07-15
zsh: fix custom syntax highlighting styles

aaron-p1:
### Description

Previously, customizing a highlighting style with `programs.zsh.syntaxHighlighting.styles.comment = "fg=#6c6c6c"` did not work because in `.zshrc` it added `ZSH_HIGHLIGHT_STYLES['comment']='fg=#6c6c6c'` which changes the key `'comment'` instead of `comment`.

This pull request changes the syntax to `ZSH_HIGHLIGHT_STYLES+=('comment' 'fg=#6c6c6c')` which does not add the quotes to the key.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@ThinkChaos

### DIFF
--- a/modules/programs/zsh.nix
+++ b/modules/programs/zsh.nix
@@ -613,7 +613,7 @@ in
           source ${cfg.syntaxHighlighting.package}/share/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh
           ${lib.concatStringsSep "\n" (
               lib.mapAttrsToList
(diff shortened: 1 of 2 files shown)


ncfavier (approved):

ncfavier:
Thanks!

ThinkChaos:
Thanks to you both!  
I guess I messed up my test (or forgot to) after adding the escaping, and hadn't noticed since.

PR #4249, opened 2023-07-17
swaync: add module

rhoriguchi:
### Description

Add module `services.swaync`

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/swaync.nix
+++ b/modules/services/swaync.nix
@@ -1,0 +1,142 @@
+{ pkgs, lib, config, ... }:
+let
+  cfg = config.services.swaync;
+  settingsFormat = pkgs.formats.json { };
+
+  settings = cfg.settings // { "$schema" = cfg.schema; };
+  configFile =
+    (settingsFormat.generate "swaync/config.json" settings).overrideAttrs (_:
+      {
+        # TODO uncomment once version higher than 0.9.0
+        # checkPhase = "${pkgs.check-jsonschema}/bin/check-jsonschema --schemafile ${settings."$schema"} $out ";
+      });
+in {
+  meta.maintainers = [ lib.maintainers.rhoriguchi ];
+
+  options.services.swaync = {
+    enable = lib.mkEnableOption "Swaync notification daemon";
+
+    package = lib.mkPackageOption pkgs "swaynotificationcenter" { };
+
+    systemd = {
+      enable = lib.mkEnableOption "Systemd integration";
+
+      target = lib.mkOption {
+        type = lib.types.str;
+        default = "graphical-session.target";
+        example = "sway-session.target";
+        description = ''
+          Systemd target to bind to.
+        '';
+      };
+    };
+
+    style = lib.mkOption {
+      type = lib.types.nullOr (lib.types.either lib.types.path lib.types.lines);
+      default = null;
+      description = ''
+        CSS style of the bar. See https://github.com/ErikReider/SwayNotificationCenter/blob/main/src/style.css for
+        the documentation.
+
+        If the value is set to a path literal, then the path will be used as the CSS file.
+      '';
+      example = ''
+        .notification-row {
+          outline: none;
+        }
+
+        .notification-row:focus,
+        .notification-row:hover {
+          background: @noti-bg-focus;
+        }
+
+        .notification {
+          border-radius: 12px;
+          margin: 6px 12px;
+          box-shadow: 0 0 0 1px rgba(0, 0, 0, 0.3), 0 1px 3px 1px rgba(0, 0, 0, 0.7),
+            0 2px 6px 2px rgba(0, 0, 0, 0.3);
+          padding: 0;
+        }
+      '';
+    };
+
+    schema = lib.mkOption {
+      default = "${cfg.package}/etc/xdg/swaync/configSchema.json";
+      type = lib.types.path;
+      description = lib.mdDoc ''
+        Schema to validate the configuration.
+      '';
+    };
+
+    settings = lib.mkOption {
+      default = { };
+      type = settingsFormat.type;
+      description = ''
+        Configuration written to {file}`$XDG_CONFIG_HOME/swaync/config.json`.
+        See https://github.com/ErikReider/SwayNotificationCenter/blob/main/src/configSchema.json for the documentation.
+      '';
+      example = lib.literalExpression ''
+        {
+          positionX = "right";
+          positionY = "top";
+          layer = "overlay";
+          control-center-layer = "top";
+          layer-shell = true;
+          cssPriority = "application";
+          control-center-margin-top = 0;
+          control-center-margin-bottom = 0;
+          control-center-margin-right = 0;
+          control-center-margin-left = 0;
+          notification-2fa-action = true;
+          notification-inline-replies = false;
+          notification-icon-size = 64;
+          notification-body-image-height = 100;
+          notification-body-image-width = 200
+        };
+      '';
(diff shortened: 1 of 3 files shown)


modules/lib/maintainers.nix
```diff
@@ -363,6 +363,12 @@
       fingerprint = "D5D6 FD1F 0D9A 3284 FB9B  C26D 3F98 EC7E C2B8 7ED1";
     }];
   };
+  rhoriguchi = {
+    email = "ryan.horiguchi@gmail.com";
+    github = "rhoriguchi";
+    githubId = 6047658;
+    name = "Ryan Horiguchi";
+  };
```
rycee, line 371: No need to add yourself if you are already in Nixpkgs' maintainer list.

modules/services/swaync.nix
```diff
@@ -0,0 +1,154 @@
...
+    enable = lib.mkEnableOption "Swaync notification daemon";
+
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.swaynotificationcenter;
+      defaultText = lib.literalExpression "pkgs.swaynotificationcenter";
+      description = ''
+        Swaync package to use. Set to <code>null</code> to use the default package.
+      '';
+    };
```
rycee, line 30: Can simplify:

```suggestion
    package = lib.mkPackageOption pkgs "swaynotificationcenter" { };
```

modules/services/swaync.nix
```diff
@@ -0,0 +1,154 @@
...
+        BusName = "org.freedesktop.Notifications";
+        ExecStart = "${cfg.package}/bin/swaync";
+        ExecReload = [
+          "${cfg.package}/bin/swaync-client --reload-config"
+          "${cfg.package}/bin/swaync-client --reload-css"
+        ];
+        Restart = "on-failure";
+      };
+
+      Install.WantedBy = [ cfg.systemd.target ];
```
rycee, line 151: This should not be necessary, the service should get activated automatically by dbus:

```suggestion
```

modules/services/swaync.nix
```diff
@@ -0,0 +1,154 @@
...
+      target = lib.mkOption {
+        type = lib.types.str;
+        default = "graphical-session.target";
+        example = "sway-session.target";
+        description = ''
+          Systemd target to bind to.
+        '';
+      };
+    };
+
```
rycee, line 44: I would suggest always creating the systemd service.
```suggestion
```

modules/services/swaync.nix
```diff
@@ -0,0 +1,154 @@
...
+      default = null;
+      description = ''
+        CSS style of the bar.
+        </para>
+        <para>
+        See <link xlink:href="https://github.com/ErikReider/SwayNotificationCenter/blob/main/src/style.css"/>
+        for the documentation.
+        </para>
+        <para>
+        If the value is set to a path literal, then the path will be used as the css file.
```
rycee, line 56: Should be Markdown now.

```suggestion
        CSS style of the bar. See
        <https://github.com/ErikReider/SwayNotificationCenter/blob/main/src/style.css>
        for the documentation.

        If the value is set to a path literal, then the path will be used as the CSS file.
```

modules/services/swaync.nix
```diff
@@ -0,0 +1,154 @@
...
+    };
+
+    settings = lib.mkOption {
+      type = jsonFormat.type;
+      default = { };
+      description = ''
+        Configuration written to <filename>$XDG_CONFIG_HOME/swaync/config.json</filename>.
+        </para>
+        <para>
+        See <link xlink:href="https://github.com/ErikReider/SwayNotificationCenter/blob/main/src/configSchema.json"/>
```
rycee, line 85:
```suggestion
        Configuration written to {file}`$XDG_CONFIG_HOME/swaync/config.json`.

        See <https://github.com/ErikReider/SwayNotificationCenter/blob/main/src/configSchema.json>.
```

modules/services/swaync.nix
```diff
@@ -0,0 +1,154 @@
...
+          inherit onChange;
+        };
+      };
+
+    systemd.user.services.swaync = lib.mkIf cfg.systemd.enable {
+      Unit = {
+        Description = "Swaync notification daemon";
+        Documentation = "https://github.com/ErikReider/SwayNotificationCenter";
+        PartOf = [ "graphical-session.target" ];
+        After = [ "graphical-session.target" ];
```
rycee, line 136:
```suggestion
        After = [ "graphical-session-pre.target" ];
```

modules/services/swaync.nix
```diff
@@ -0,0 +1,154 @@
...
+          inherit onChange;
+        };
+        "swaync/style.css" = lib.mkIf (cfg.style != null) {
+          source =
+            if builtins.isPath cfg.style || lib.isStorePath cfg.style then
+              cfg.style
+            else
+              pkgs.writeText "swaync/style.css" cfg.style;
+          inherit onChange;
+        };
```
rycee, line 128: I think you can remove the onChange stuff. Just add

```
Unit.X-Restart-Triggers = [
  "${config.xdg.configFile."swaync/config.json".source}"
  "${config.xdg.configFile."swaync/style.css".source}"
]
```

to the service definition.

modules/services/swaync.nix
rhoriguchi: My thought process was, if you only want to use the notification shade but not the notification popups. This way you can configure it without starting it.

modules/services/swaync.nix
rhoriguchi: I need this for the same reason as I need it here. https://github.com/nix-community/home-manager/pull/4224#issuecomment-1631526382

modules/services/swaync.nix
rhoriguchi: I had to remove the `<` and `>` from the url to get it to build.

modules/services/swaync.nix
rhoriguchi: There's a couple of other places where this improvement could be made.

rhoriguchi:
Thx for the inputs @rycee

teto:
I am interested in this as a swaync user but this breaks my nixos-unstable build: 
```
    > /nix/store/lsg1z6y6k5jp5wz2w0yqf8zs92n62j1y-manual-combined/man
ual-combined.xml:51457: validity error : xml:id : attribute value opt-services.swaync.settings._\$schema_ is not an NCName
```
also could you rebase ? other than that looks ok

modules/services/swaync.nix
teto: One could still disable the systemd in its own config via systemd.user.services.swaync.enable = false or something similar (not sure how to disable the service) , i.e., it makes the module simpler.

teto:
I've been using this for a while and only now realized it was not merged yet. Could you rebase and address comments please ? it's a pretty cool software/module

rhoriguchi:
I'm currently not running hyperland, so I can't really test. Thank you @teto for rebasing.

hmajid2301:
hi guys any updates on this ?

teto:
@hmajid2301 feel free to open a new PR with this as a basis as long as you keep the commit authors intact

mobsenpai:
Looks like only formatting tests are failing, isn't it?

abayomi185:
I'd like to look into this. Will create a PR referencing this

rycee:
Merged #5345 so I'll close this. Thanks everybody for your input!

PR #4259, opened 2023-07-21
git-sync: darwin support

ryane:
### Description

`services.git-sync` support on darwin

- On darwin, creates a launch agent to run git-sync on an interval and when the `path` changes.
- The `uri` option is not used on Darwin. The auto-creation of the local git directory from the `uri` is a feature of the
[git-sync-on-inotify](https://github.com/simonthum/git-sync/blob/master/contrib/git-sync-on-inotify) wrapper (which won't work on Darwin afaik) and not `git-sync` itself.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@imalison  @cab404

_Disclaimer: new-ish to nix, first attempted home-manager contribution_

### DIFF
--- a/modules/services/git-sync.nix
+++ b/modules/services/git-sync.nix
@@ -24,9 +24,21 @@ let
     };
   };
 
+  mkAgent = name: repo: {
+    enable = true;
+    config = {
+      StartInterval = repo.interval;
+      ProcessType = "Background";
+      WorkingDirectory = "${repo.path}";
+      WatchPaths = [ "${repo.path}" ];
+      ProgramArguments = [ "${cfg.package}/bin/git-sync" ];
+    };
+  };
+
+  mkService = if pkgs.stdenv.isLinux then mkUnit else mkAgent;
   services = mapAttrs' (name: repo: {
     name = "git-sync-${name}";
-    value = mkUnit name repo;
+    value = mkService name repo;
   }) cfg.repositories;
 
   repositoryType = types.submodule ({ name, ... }: {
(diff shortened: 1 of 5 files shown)


cab404:
Can't test, but LGTM!

modules/services/git-sync.nix
```diff
@@ -90,12 +103,14 @@ in {
...
-  config = mkIf cfg.enable {
-    assertions = [
-      (lib.hm.assertions.assertPlatform "services.git-sync" pkgs
-        lib.platforms.linux)
-    ];
+  config = mkIf cfg.enable (mkMerge [
+    (mkIf pkgs.stdenv.isLinux { systemd.user.services = services; })
+    (mkIf pkgs.stdenv.isDarwin {
+      launchd = {
+        enable = true;
```
rycee, line 110: Don't think this is necessary, it should default to true on Darwin.

modules/lib/maintainers.nix
```diff
@@ -171,6 +171,14 @@
...
     githubId = 1224006;
   };
+  ryane = {
+    email = "ryanesc@gmail.com";
+    github = "ryane";
+    githubId = 7346;
+    name = "Ryan Eschinger";
+    keys =
+      [{ fingerprint = "E4F4 1EAB BF0F C785 06D8  62EF EF68 CF41 D42A 593D"; }];
+  };
```
rycee, line 181: No need to add yourself if you are already in Nixpkgs' maintainer list.

rycee:
Thanks! Merged now 🙂

rycee:
@ambroisie Fair enough, added now 🙂

PR #4269, opened 2023-07-23
Jetbrains remote dev

GenericNerdyUsername:
### Description

This adds a module that allows you to select the IDEs accessible using gateway and a ssh connection

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/lib/maintainers.nix
+++ b/modules/lib/maintainers.nix
@@ -223,6 +223,11 @@
     githubId = 33058747;
     name = "Gaetan Lepage";
   };
+  genericnerdyusername = {
+    email = "genericnerdyusername@proton.me";
+    github = "genericnerdyusername";
+    githubId = 111183546;
+  };
   maximsmol = {
     email = "maximsmol@gmail.com";
     github = "maximsmol";
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -1407,6 +1407,13 @@ in {
           A new module is available: 'services.arrpc'
         '';
       }
+
+      {
+        time = "2024-02-05T22:45:37+00:00";
+        message = ''
+          A new module is available: 'programs.jetbrains-remote'
+        '';
+      }
     ];
   };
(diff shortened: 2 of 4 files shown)


modules/programs/jb-remote.nix
```diff
@@ -0,0 +1,33 @@
...
+    "${ide}/bin/${ide.meta.mainProgram}-remote-dev-server registerBackendLocationForGateway";
+  lines = map mkLine cfg.ides;
+  linesStr = concatStringsSep "\n" lines;
+  dagEntry = hm.dag.entryAfter [ "writeBoundary" ] linesStr;
+
+in {
+  meta.maintainers = with lib.maintainers; [ genericnerdyusername ];
+
+  options.programs.jb-remote = {
+    ides = mkOption {
```
sumnerevans, line 19: should we be installing these IDEs into `home.packages` as well?

modules/programs/jb-remote.nix
```diff
@@ -0,0 +1,36 @@
...
+  lines = map mkLine cfg.ides;
+  linesStr = ''
+    rm $HOME/.cache/JetBrains/RemoteDev/userProvidedDist/_nix_store*
+  '' + (concatStringsSep "\n" lines);
+  dagEntry = hm.dag.entryAfter [ "writeBoundary" ] linesStr;
+
+in {
+  meta.maintainers = with lib.maintainers; [ genericnerdyusername ];
+
+  options.programs.jb-remote = {
```
sumnerevans, line 21: I don't think we have any abbreviations elsewhere, so we should probably say `jetbrains-remote` or similar instead of `jb-remote` .

sumnerevans (changes requested):

sumnerevans (approved):

rycee:
Thanks! Merged to master now. Note, I added an `enable` option that needs to be enabled for the module to take effect.

PR #4272, opened 2023-07-25
aerc: do not use smtp-starttls

mxkrsv:
### Description

This option was deprecated in 0.15.0, +insecure schema is suggested instead.
Thus, we can just copy imap logic, which already uses +insecure.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@lukasngl

### DIFF
--- a/modules/programs/aerc-accounts.nix
+++ b/modules/programs/aerc-accounts.nix
@@ -177,21 +177,20 @@ in {
               params = cfg.aerc.smtpOauth2Params;
             };
 
-            protocol = if smtp.tls.enable && !smtp.tls.useStartTls then
-              "smtps${loginMethod'}"
+            protocol = if smtp.tls.enable then
+              if smtp.tls.useStartTls then
+                "smtp${loginMethod'}"
+              else
+                "smtps${loginMethod'}"
             else
-              "smtp${loginMethod'}";
+              "smtp+insecure${loginMethod'}";
 
             port' = optPort smtp.port;
-
-            smtp-starttls =
-              if smtp.tls.enable && smtp.tls.useStartTls then "yes" else null;
 
           in {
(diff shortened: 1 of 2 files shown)


ncfavier:
cc @patwid

tests/modules/programs/aerc/extraAccounts.expected
```diff
@@ -47,12 +46,12 @@ outgoing-cred-cmd = echo PaSsWorD!
 
 [g_smtp-passcmd-notls-nostarttls]
 from = Foo Bar <addr@mail.invalid>
-outgoing = smtp+plain://foobar@smtp.host.invalid:42
+outgoing = smtp+plain+insecure://foobar@smtp.host.invalid:42
```
patwid, line 49: `outgoing = smtp+insecure+plain://foobar@smtp.host.invalid:42`

same below

modules/programs/aerc-accounts.nix
```diff
@@ -177,21 +177,20 @@ in {
...
-            protocol = if smtp.tls.enable && !smtp.tls.useStartTls then
-              "smtps${loginMethod'}"
+            protocol = if smtp.tls.enable then
+              if smtp.tls.useStartTls then
+                "smtp${loginMethod'}"
+              else
+                "smtps${loginMethod'}"
             else
-              "smtp${loginMethod'}";
+              "smtp${loginMethod'}+insecure";
```
patwid, line 186: not sure if it actually makes a difference, but i would probably swap the order here for clarity still, because `smtp+insecure` corresponds to the `scheme` and `plain` corresponds to the `auth` method (separated by `+`)

appart from that, this LGTM!

poztit (approved):

Issue #2812, opened 2022-03-21
Issue: Some apps don't find the cursor icons 


leiserfg:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Some apps (at least Firefox) don't find the xcursors, I make it work by setting
`XCURSOR_PATH=$HOME/.nix-profile/share/icons` and I think that's in fact the solution, but I don't know where to put it in a PR as it (in theory) should work for wayland as well (not only x11).

### Maintainer CC

@rycee 

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 5.15.25-1-MANJARO, Manjaro Linux, noversion`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.6.1`
 - channels(leiserfg): `"home-manager, nixpkgs"`
 - nixpkgs: `/home/leiserfg/.nix-defexpr/channels/nixpkgs`
```

berbiche:
@polykernel is this something your PR #2891 could fix?

polykernel:
@berbiche Yes, it can but I am unsure on where the variable should be set because there isn't a standard interface for graphic sessions in home-manager. I think it can be set in `home.sessionVariables` which may not be appropriate depending on the user's graphic session configurations, but as per https://github.com/nix-community/home-manager/issues/1011#issuecomment-850889575 this is the preferred way to manage variables for any login session. What do you think?

berbiche:
> I think it can be set in `home.sessionVariables` which may not be appropriate depending on the user's graphic session configurations

That's the best solution we currently have.

nim65s:
Hi,

I used `home.pointercursor` and correctly have `export XCURSOR_PATH="$XCURSOR_PATH${XCURSOR_PATH:+:}/home/gsaurel/.nix-profile/share/icons"` in my `.nix-profile/etc/profile.d/hm-session-vars.sh`, but eg. firefox & kitty still can't get the cursors on login.

My current workaround is to set `".icons".source = ~/.nix-profile/share/icons;` in `home.file`.

polykernel:
Thanks for the report. I have added a symlink from `$HOME/.icons` to the home-manager profile icons directory and moved the index.theme file for the default theme in https://github.com/nix-community/home-manager/pull/4281.

PR #4281, opened 2023-07-29
home-cursor: make icon configuration compatible with more applications

polykernel:
The current cursor configuration poorly accomodate applications not following the XDG icon theme spec. As a result, some applications fail to locate user defined icons and themes.

To address this, a symlink from $HOME/.icons to the home-manager profile icons directory was added for backwards compatibility with applications using hardcoded icons directories. In addition, the default index.theme file was moved from XDG_DATA_DIRS back to $HOME/.icons for greater compatibility. As per the XDG icon theme spec, applications are to search in $HOME/.icons first for icons and themes so the existing behavior should not break for XDG compliant applications. [1]

[1] - https://specifications.freedesktop.org/icon-theme-spec/latest/ar01s03.html

### Description

fixes https://github.com/nix-community/home-manager/issues/2812#issuecomment-1640252050

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@GaetanLepage as you change the location of the index.theme file in https://github.com/nix-community/home-manager/commit/013948ddf6576bbc02629e9b78e87b1364cd3dcf

### DIFF
--- a/modules/config/home-cursor.nix
+++ b/modules/config/home-cursor.nix
@@ -52,6 +52,27 @@ let
       escapeShellArg cfg.x11.defaultCursor
     }";
 
+  defaultIndexThemePackage = pkgs.writeTextFile {
+    name = "index.theme";
+    destination = "/share/icons/default/index.theme";
+    # Set name in icons theme, for compatibility with AwesomeWM etc. See:
+    # https://github.com/nix-community/home-manager/issues/2081
+    # https://wiki.archlinux.org/title/Cursor_themes#XDG_specification
+    text = ''
+      [Icon Theme]
+      Name=Default
+      Comment=Default Cursor Theme
+      Inherits=${cfg.name}
+    '';
+  };
+
+  iconMappings = let
+    knownFiles = filterAttrs (_: type: type != "unknown")
+      (builtins.readDir "${cfg.package}/share/icons");
+  in mapAttrs'
+  (n: _: nameValuePair ".icons/${n}" "${cfg.package}/share/icons/${n}")
+  knownFiles;
+
 in {
   meta.maintainers = [ maintainers.polykernel maintainers.league ];
 
@@ -120,17 +141,7 @@ in {
         (hm.assertions.assertPlatform "home.pointerCursor" pkgs platforms.linux)
       ];
 
-      home.packages = [ cfg.package ];
-
-      # Set name in icons theme, for compatibility with AwesomeWM etc. See:
-      # https://github.com/nix-community/home-manager/issues/2081
-      # https://wiki.archlinux.org/title/Cursor_themes#XDG_specification
-      xdg.dataFile."icons/default/index.theme".text = ''
-        [icon theme]
-        Name=Default
-        Comment=Default Cursor Theme
-        Inherits=${cfg.name}
-      '';
+      home.packages = [ cfg.package defaultIndexThemePackage ];
 
       # Set directory to look for cursors in, needed for some applications
       # that are unable to find cursors otherwise. See:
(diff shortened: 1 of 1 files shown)


GaetanLepage:
Ok, so if I understand well, before [_my PR_](https://github.com/nix-community/home-manager/pull/3851), the icon config file was placed at `~/.icons/default/index.theme`.
My PR moved it to `$XDG_DATA_HOME/icons/default/index.theme` (i.e. `~/.local/share/icons/default/index.theme`).
Then, this current PR proposes to put it back in the previous place.

-> I would then naively to suggest linking it to both places if some apps are not able to find it in `$XDG_DATA_HOME`.

PS: The [issue](https://github.com/nix-community/home-manager/issues/2812) mentions firefox, although my cursor works fine in Firefox. Maybe it works for me because I am using NixOS and/or wayland.

nim65s:
ok for both.
For the issue on firefox, I am on ubuntu and wayland.

GaetanLepage:
@ncfavier @rycee Do you also agree on linking the icon theme configuration to both places:
- `~/.local/share/icons`
- `~/.icons`
?

rycee:
@GaetanLepage If various programs look in both those locations then I think it is no problem to link to both places.

GaetanLepage:
> @GaetanLepage If various programs look in both those locations then I think it is no problem to link to both places.

I indeed agree.

modules/config/home-cursor.nix
```diff
@@ -142,6 +146,11 @@ in {
         XCURSOR_SIZE = mkDefault cfg.size;
         XCURSOR_THEME = mkDefault cfg.name;
       };
+
+      # Symlink $HOME/.icon to the home-manager profile icons directory, needed for
+      # backwards compatibility with some applications. See:
+      # https://specifications.freedesktop.org/icon-theme-spec/latest/ar01s03.html
+      home.file.".icons".source = "${config.home.profileDirectory}/share/icons";
```
rycee, line 153: Unfortunately this does not work in pure evaluation mode:

> error: access to absolute path '/home/rycee/.nix-profile/share/icons' is forbidden in pure eval mode (use '--impure' to override)

Maybe

```suggestion
      home.file.".icons".source = "${config.home.path}/share/icons";
```

instead?

rycee: A concern here may be that users may wish to place their own icons non-declaratively inside `~/.icons` and this would prevent that. Perhaps `recursive = true` is a possibility but it may be a bit slow if you have many icons installed…

polykernel: Given the scope of the home-pointecursor module pertains to cursor configurations, I opted to symlink only the leaf files from `cfg.package` rather than the entire icons directory to mitigate the performance impact incurred by `recursive = true`. This leaves somewhat more to be desired but I think a clear separation of concerns is appropriate here. On a technicality, the user can still pass a symlink farm to `cfg.package` and potentially experience slowdowns due to link creation, however I think in that case it can assumed that the user is able to wrangle and fine tune the directory structure manually to their needs.1

polykernel:
> I would then naively to suggest linking it to both places if some apps are not able to find it in `$XDG_DATA_HOME`.

The XDG spec actually prescribe that application check `$HOME/.icons` first before `$XDG_DATA_HOME/icons` so in theory, linking `$HOME/.icons` should be sufficient. However, in practices, some applications are partially/non compliant so I agree to linking to both locations for potentially better compatibility.

I have updated the PR to link cursor icons in `cfg.package` and the default `index.theme` to both `$HOME/.icons` and `$XDG_DATA_HOME/icons`.

rycee:
@polykernel Yes, that is a clever solution! Very nice, merged to master now 🙂

modules/config/home-cursor.nix
```diff
@@ -52,6 +52,27 @@ let
...
+      [Icon Theme]
+      Name=Default
+      Comment=Default Cursor Theme
+      Inherits=${cfg.name}
+    '';
+  };
+
+  iconMappings = let
+    knownFiles = filterAttrs (_: type: type != "unknown")
+      (builtins.readDir "${cfg.package}/share/icons");
```
R-VdP, line 71: Unfortunately, this line introduced an IFD into home-manager, causing eval errors in my CI system that does not allow for IFD.

Is there an existing policy on IFD in home-manager? In nixpkgs for instance IFD is not allowed.

I think we should at least retain the option of simply recursively linking the cursor dirs instead of relying on IFD, to account for situations where IFD is not allowed or desirable.

Any opinions?

polykernel: My apologies, I am not too familiar with IFD and thus overlooked the IFD being introduced. I think in this case the IFD is not strictly necessary and can be avoided by symlinking only the directory under the icon package matching `config.home.pointerCursor.name`. This should not result in any functional difference in behavior from the status quo as the module is only meant to generate configurations for the specified cursor. I have implementing the above change in https://github.com/nix-community/home-manager/pull/4369.

R-VdP: No need to apologise! IFDs are easily missed.

Thanks for the quick fix!

PR #4288, opened 2023-07-31
gh: option to enable helper for additional hosts

satoqz:
### Description

`gh` can also be used with github enterprise hosts, for which there exists no easy option to enable the credential helper except for directly working with `programs.git.extraConfig`.

Therefore I've added the option `programs.gh.extraGitCredentialHelperHosts`. The name is up for debate since it's maybe a bit verbose.

Not sure if this is a needed addition since it's somewhat niche, at the same time it's not very complex and makes the life of github enterprise users a little easier.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@Gerschtli @berbiche

### DIFF
--- a/modules/programs/gh.nix
+++ b/modules/programs/gh.nix
@@ -60,6 +60,12 @@ in {
         "settings"
         "git_protocol"
       ])
+      (mkRenamedOptionModule [ "programs" "gh" "enableGitCredentialHelper" ] [
+        "programs"
+        "gh"
+        "gitCredentialHelper"
+        "enable"
+      ])
     ];
 
   options.programs.gh = {
@@ -91,19 +97,19 @@ in {
       '';
     };
 
-    enableGitCredentialHelper =
-      mkEnableOption "the gh git credential helper for github.com" // {
+    gitCredentialHelper = {
+      enable = mkEnableOption "the gh git credential helper" // {
         default = true;
       };
 
-    extraGitCredentialHelperHosts = mkOption {
-      type = types.listOf types.string;
-      default = [ ];
-      description =
-        "Additional GitHub Enterprise hosts to enable the gh git credential helper for";
-      example = literalExpression ''
-        [ "https://github.example.com" ]
-      '';
+      hosts = mkOption {
+        type = types.listOf types.str;
+        default = [ "https://github.com" ];
+        description = "GitHub hosts to enable the gh git credential helper for";
+        example = literalExpression ''
+          [ "https://github.com" "https://github.example.com" ]
+        '';
+      };
     };
 
     extensions = mkOption {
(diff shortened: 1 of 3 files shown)


modules/programs/gh.nix
```diff
@@ -96,6 +96,16 @@ in {
         default = true;
       };
 
+    extraGitCredentialHelperHosts = mkOption {
+      type = types.listOf types.string;
```
ncfavier, line 100:
```suggestion
      type = types.listOf types.str;
```

modules/programs/gh.nix
```diff
@@ -112,9 +122,13 @@ in {
     xdg.configFile."gh/config.yml".source =
       yamlFormat.generate "gh-config.yml" cfg.settings;
 
-    programs.git.extraConfig.credential."https://github.com".helper =
-      mkIf cfg.enableGitCredentialHelper
-      "${cfg.package}/bin/gh auth git-credential";
+    programs.git.extraConfig.credential = let
+      hosts = lib.optional cfg.enableGitCredentialHelper "https://github.com"
+        ++ cfg.extraGitCredentialHelperHosts;
```
ncfavier, line 127: Not sure about those semantics. `enableGitCredentialHelper` should control whether `programs.git.extraConfig.credential` is set at all, and there should be an option listing all the hosts to enable the credential helper for, including the default `github.com`.

satoqz: That makes a lot of sense, any name suggestions for that option?

ncfavier: Maybe `gitCredentialHelper.enable` and `gitCredentialHelper.hosts`, and `mkRenamedOptionModule [ ... "enableGitCredentialHelper" ] [ ... "gitCredentialHelper" "enable" ]`.

modules/programs/gh.nix
```diff
@@ -91,11 +97,21 @@ in {
...
-    enableGitCredentialHelper =
-      mkEnableOption "the gh git credential helper for github.com" // {
+    gitCredentialHelper = {
+      enable = mkEnableOption "the gh git credential helper" // {
         default = true;
       };
 
+      hosts = mkOption {
+        type = types.listOf types.str;
+        default = [ "https://github.com" ];
```
ncfavier, line 107: Maybe move this default value to `config`, so that people can append to the list while keeping the default value (and still be able to `mkForce` if they want to remove github)

satoqz: On second thought I'm not sure about this, it creates ambiguity since anyone who sets the `hosts` option now intuitively can't be sure whether they also need to include `"https://github.com"` in it.

I'd argue that those people who do actually end up adding additional hosts (which is probably a small subset of users of the module) will probably think about ensuring the default is still there.

Additionally, I'd find it confusing if the option is called `hosts` but doesn't actually have to include all the hosts that you enable the option for.

ncfavier: Fair enough

Issue #4291, opened 2023-08-01
bug: direnv nushell integration needs to be updated for nushell 0.83


lilyball:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

`nushell` version 0.83 deprecated the `let-env` command in favor of just assigning to `$env.NAME` directly. This syntax was supported in at least version 0.82 as well (I haven't tried older versions). Unfortunately, the `direnv` module is still using `let-env` in its nushell integration.

https://github.com/nix-community/home-manager/blob/4fd794d3df88735dcf9662155d77b08a2e2dde29/modules/programs/direnv.nix#L127-L142

As nixpkgs now has nushell 0.83, this needs to be updated to assign to `$env.config` directly.

While you're at it, you can also simplify the code, it doesn't need to create each key separately, it can look something like

```nu
$env.config.hooks.pre_prompt = ($env.config?.hooks?.pre_prompt? | default [] | append {
    code: "…"
})
```

### Maintainer CC

@rycee

### System information

```markdown
- system: `"aarch64-darwin"`
 - host os: `Darwin 22.6.0, macOS 13.5`
 - multi-user?: `no`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.15.1`
 - channels(lily): `""`
 - channels(root): `""`
 - nixpkgs: `/etc/nix/inputs/nixpkgs`
```

ncfavier:
Is this fixed by https://github.com/nix-community/home-manager/pull/4292 ?

lilyball:
Yes, thanks.

PR #4292, opened 2023-08-01
nushell: deprecation of let-env

happysalada:
### Description

let-env has been deprecated in nushell's latest release.
This PR aims to fix the errors that this causes.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/direnv.nix
+++ b/modules/programs/direnv.nix
@@ -128,10 +128,10 @@ in {
       # Using mkAfter to make it more likely to appear after other
       # manipulations of the prompt.
       mkAfter ''
-        let-env config = ($env | default {} config).config
-        let-env config = ($env.config | default {} hooks)
-        let-env config = ($env.config | update hooks ($env.config.hooks | default [] pre_prompt))
-        let-env config = ($env.config | update hooks.pre_prompt ($env.config.hooks.pre_prompt | append {
+        $env.config = ($env | default {} config).config
+        $env.config = ($env.config | default {} hooks)
+        $env.config = ($env.config | update hooks ($env.config.hooks | default [] pre_prompt))
+        $env.config = ($env.config | update hooks.pre_prompt ($env.config.hooks.pre_prompt | append {
           code: "
             let direnv = (${pkgs.direnv}/bin/direnv export json | from json)
             let direnv = if ($direnv | length) == 1 { $direnv } else { {} }
--- a/modules/programs/nushell.nix
+++ b/modules/programs/nushell.nix
@@ -85,7 +85,7 @@ in {
       type = types.nullOr (linesOrSource "env.nu");
       default = null;
       example = ''
-        let-env FOO = 'BAR'
+        $env.FOO = 'BAR'
       '';
       description = ''
         The environment variables file to be used for nushell.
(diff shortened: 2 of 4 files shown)


happysalada:
Adding a bit of context here.
The code added for zoxide and atuin is just a kludge until those packages update their init script. It will be backward compatible, so I think we can leave that kludge for a while after the packages have been update to be sure no one has problems.

While these fixes will work for version 0.83+ , I don't think they will work with earlier version, so perhaps this shouldn't be merged ? I'm opening this PR more as a discussion and to help other people with the problem as well.

happysalada:
here is where I found the fix https://github.com/ajeetdsouza/zoxide/issues/599

ncfavier:
Does this fix https://github.com/nix-community/home-manager/issues/4291 ?

modules/programs/atuin.nix
```diff
@@ -130,7 +130,7 @@ in {
         if not ($atuin_cache | path exists) {
           mkdir $atuin_cache
         }
-        ${cfg.package}/bin/atuin init nu ${flagsStr} | save --force ${config.xdg.cacheHome}/atuin/init.nu
+        ${cfg.package}/bin/atuin init nu ${flagsStr} | str replace --string --all 'let-env ' '$env.' | save --force ${config.xdg.cacheHome}/atuin/init.nu
```
ncfavier, line 133: I'd rather this be fixed in nixpkgs by applying https://github.com/atuinsh/atuin/pull/1080

modules/programs/zoxide.nix
```diff
@@ -86,7 +86,7 @@ in {
         if not ($zoxide_cache | path exists) {
           mkdir $zoxide_cache
         }
-        ${cfg.package}/bin/zoxide init nushell ${cfgOptions} | save --force ${config.xdg.cacheHome}/zoxide/init.nu
+        ${cfg.package}/bin/zoxide init nushell ${cfgOptions} | str replace --string --all 'let-env ' '$env.' | save --force ${config.xdg.cacheHome}/zoxide/init.nu
```
ncfavier, line 89: I'd rather this be fixed in nixpkgs by applying https://github.com/ajeetdsouza/zoxide/pull/587

happysalada:
Yes this does fix the issue you referenced.

I agree with the two comments you made. What do we do in the meanwhile though ? My idea behind the fixes on zoxide and atuin is that they can always be removed after the packages have been updated, even after the update, the fix wont break anything. Another release of both packages might be weeks away, thats why i suggested those in the meanwhile.

Let me know what you would rather do. Personally I would tend to leave the two fixes in at least until a release is made for both packages.

ncfavier:
I meant applying both patches in nixpkgs so that we don't have to wait for releases.

BvnditGvngWill:
I agree

happysalada:
I've asked both of the upstream repos for a release (it might not happen, but I thought that it was worth trying at least).
I'm leaving this as is for now, so people can just use it if their environment is broken. If there are no answers in a week, it makes sense to apply the patches manually in nixpkgs.
Thank you for the quick review!

happysalada:
Alright, ive removed the fixes for zoxide and atuin. It should be ready to merge.

Btw the creator of atuin said she would cut a release this weekend,.

ncfavier (approved):

jcdickinson:
Thanks @happysalada!

PR #4300, opened 2023-08-02
nushell: add `home.shellAlises` support for nushell

ryan4yin:
### Description

This PR add `home.shellAlises` support for nushell.

It was once implemented by PR https://github.com/nix-community/home-manager/pull/3832, but was rejected due to the existing PR https://github.com/nix-community/home-manager/pull/3529, and this feature is not covered by the old PR.

### Checklist

- [ x Code tested through `nix-shell --pure tests -A run.all` or `nix develop --ignore-environment .#all` using Flakes.

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/home-environment.nix
+++ b/modules/home-environment.nix
@@ -516,6 +516,7 @@ in
     programs.bash.shellAliases = cfg.shellAliases;
     programs.zsh.shellAliases = cfg.shellAliases;
     programs.fish.shellAliases = cfg.shellAliases;
+    programs.nushell.shellAliases = cfg.shellAliases;
 
     home.sessionVariables =
       let


ryan4yin:
Sorry for that, `home.shellAliases` aren't always compatible with those from nushell, colsed.

rycee:
Hi, yeah unfortunately that is the case. The `home.shellAliases` option expects the more traditional style of shell aliases. Hopefully we can support them in nushell at some point in the future, but for now we'll have to skip it.

Thanks for giving it a try, though 🙂

jaredmontoya:
> Hi, yeah unfortunately that is the case. The `home.shellAliases` option expects the more traditional style of shell aliases. Hopefully we can support them in nushell at some point in the future, but for now we'll have to skip it.
> 
> Thanks for giving it a try, though 🙂

@rycee does that mean that users who know what they are doing have to always do this for NO reason?:
```nix
programs.nushell.shellAliases = config.home.shellAliases // {
  l = "ls -la";
  ll = "ls -l";
  la = "ls -a";
  other = "nu-only aliases";
};
```

Because it in fact works if you use `home.shellAliases` correctly.

doesn't `home.shellAliases` already expect ONLY ALISES THAT WORK IN ALL SHELLS that you are using by the nature of it being an attribute of home?

And it is not me who said this, here is quote from the docs:
> This option should only be used to manage simple aliases that are
compatible across **ALL(INCLUDING NUSHELL)** shells. If you need to use a shell specific
feature then make sure to use a shell specific option, for example
programs.bash.shellAliases for Bash.

What's the problem for the user to just NOT write things that only work in nushell in `home.shellAliases` and instead write complex nushell-only aliases in `programs.nushell.shellAliases` when they need to

And if some aliases work only in bash type shells the user must not put them in `home.shellAliases` expecting that it is made for bash aliases. The user must write bash-only aliases in `programs.bash.shellAliases` and then set `programs.zsh.shellAliases = config.programs.bash.shellAliases` to avoid duplicating them between bash type shells because that is how those attributes being used makes the MOST sense.

Issue #2954, opened 2022-05-11
Confusing way of setting nixpkgs options with lib.homeManagerConfiguration [Nix Flakes]


funketh:
The `nixpkgs.config` options are overridden here:
https://github.com/nix-community/home-manager/blob/5ac84ebeef5d2566ea458b81375cc6eafa19c225/flake.nix#L47-L56
This was very confusing and unintuitive since I did not know about the pkgs parameter (it isn't mentioned in the online documentation) and my `nixpkgs.config.allowUnfree = true;` in the actual configuration got ignored.

Personally, I would be against overriding these here.
(The NixOS equivalent `lib.nixosSystem` does not do this either afaik.)
At the very least the pkgs option (and its relevance for config options such as allowUnfree) should be mentioned in the docs.

cmm:
see also #2942 
FTR I still totally don't get how `nixpkgs.config.allowUnfree = true` worked before and (subsequently) how exactly did https://github.com/NixOS/nixpkgs/commit/1c49b81263858c69b932da05ae63a7767b308e74 break it

PaulGrandperrin:
I agree this interface is very confusing, but I would like `pkgs` to continue to be settable, otherwise `home-manager` becomes uncomposable.
In fact, I think it should be the normal way to set it, and stop using `follows` in the inputs.

Here's an example of `flake.nix`:
```nix
{
  inputs = {
    nixos-unstable.url = "github:nixos/nixpkgs/nixos-unstable";
    nixos-unstable-small.url = "github:nixos/nixpkgs/nixos-unstable-small";
    nixpkgs-darwin-unstable.url = "github:nixos/nixpkgs/nixpkgs-darwin-unstable";
 
	home-manager.url = "github:nix-community/home-manager";
    home-manager.inputs.nixpkgs.follows = "nixpkgs"; # using this method prevents composition and is confusing
  };

  outputs = inputs:
      homeConfigurations = {
	        aaa = inputs.home-manager.lib.homeManagerConfiguration {
	          pkgs = inputs.nixos-unstable.legacyPackages.x86_64-linux;
			  ...
	        };
		    bbb = inputs.home-manager.lib.homeManagerConfiguration {
	          pkgs = inputs.nixos-unstable-small.legacyPackages.x86_64-linux;
			  ...
	        };
			ccc = inputs.home-manager.lib.homeManagerConfiguration {
[...]

PR #4304, opened 2023-08-03
 flake: fix nixpkgs `config`

### LINKED ISSUE #2942: bug: Flake config cannot use unfree packages despite nixpkgs.config.allowUnfree = true

### Issue description

I have a flake config ([source](https://github.com/schuelermine/home-manager-configuration)) and as of nixpkgs c777cdf5c564015d5f63b09cc93bef4178b19b01 and home-manager 538343be863cb0b9e9f1471e6dc09e0e140c7b3d unfree packages refuse to evaluate despite a `nixpkgs.config.allowUnfree = true;` in my config.

### Maintainer CC

_No response_

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 5.17.5, NixOS, 22.05 (Quokka), 22.05.20220503.cbe587c`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.9.0pre20220428_4bb111c`
 - channels(root): `"nixos, nixpkgs"`
 - channels(anselmschueler): `""`
```

Gerg-L:
### Description
Closes #2942

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -80,7 +80,10 @@
             configuration = { ... }: {
               imports = modules
                 ++ [{ programs.home-manager.path = toString ./.; }];
-              nixpkgs = { inherit (pkgs) config overlays; };
+              nixpkgs = {
+                config = nixpkgs.lib.mkDefault pkgs.config;
+                inherit (pkgs) overlays;
+              };
             };
           });
       };


NotAShelf:
lol

justinlime:
Finally 👌

eclairevoyant (approved):
test case, adapted from the linked issue:

```nix
{
  inputs.home-manager = {
    url = "github:gerg-l/home-manager";
    inputs.nixpkgs.follows = "nixpkgs";
  };

  outputs = {
    self,
    nixpkgs,
    home-manager,
  }: let
    system = "x86_64-linux";
    allowUnfree = {nixpkgs.config.allowUnfree = true;};
  in {
    homeConfigurations.test = home-manager.lib.homeManagerConfiguration {
      pkgs = nixpkgs.legacyPackages.${system};
      modules = [
        {
          home = {
            username = "test";
            homeDirectory = "/home/test";
            stateVersion = "22.11";
          };
        }

        allowUnfree

        ({pkgs, ...}: {
          home.packages = [pkgs.unrar];
        })
      ];
    };
  };
}
```

`nix run github:nix-community/home-manager -- --flake .#test build` completes successfully

sumnerevans:
Looks promising. Please make sure to follow the checklist, especially the commit message.

cvoges12:
Finally!

Gerg-L:
> Looks promising. Please make sure to follow the checklist, especially the commit message.

yup, dealing with it now

Gerg-L:
alright ignore my struggling with naming all should be good now

flake.nix
```diff
@@ -80,7 +80,8 @@
             configuration = { ... }: {
               imports = modules
                 ++ [{ programs.home-manager.path = toString ./.; }];
-              nixpkgs = { inherit (pkgs) config overlays; };
+              nixpkgs =
+                nixpkgs.lib.mkDefault { inherit (pkgs) config overlays; };
```
ncfavier, line 84: Maybe push this down a bit? `nixpkgs.config = mkDefault pkgs.config; nixpkgs.overlays = mkDefault pkgs.overlays;`

Gerg-L: `config` and `overlays` are the only options under `nixpkgs`
this change does nothing

ncfavier: It would allow users to change `nixpkgs.overlays` while still having `config` inherited from `pkgs`, for example.

Gerg-L:
It should be noted that this fix is still a hack, and the actual idiomatic solution is change the `pkgs` argument in `lib.homeManagerConfiguration` to a `nixpkgs` argument

emilazy:
I'm curious why `homeManagerConfiguration` requires a `pkgs` option rather than having it set modularly like NixOS and nix-darwin do and just requiring `lib`? In particular I would expect passing `pkgs` to override the package set and disable the `nixpkgs` module, so defaulting the `nixpkgs.*` options based on it seems strange to me.

emilazy:
Linking https://github.com/LnL7/nix-darwin/commit/f9724c4543035d6190c00168ebfa93f0b2e927d0 just in case it's useful to anyone; Home Manager doesn't seem to have quite the brokenness that plagued nix-darwin at the time, but hopefully it offers some food for thought on how external users of the module system should handle specifying a package set and the kind of interface I ended up settling on.

Gerg-L:
that's basically what I'm saying, I could easily make a PR to switch it. but it'd obviously be a very breaking change

flake.nix
Gerg-L: > It would allow users to change `nixpkgs.overlays` while still having `config` inherited from `pkgs`, or vice versa.

```nix
{
  inputs = {
    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
    home-manager = {
      url = "github:gerg-l/home-manager";
      inputs.nixpkgs.follows = "nixpkgs";
    };
  };
  outputs = {
    nixpkgs,
    home-manager,
    ...
  }: {
    homeConfigurations."gerg" = home-manager.lib.homeManagerConfiguration {
      pkgs = import nixpkgs {
        system = "x86_64-linux";
        config.allowUnfree = true;
      };
      modules = [
        ({pkgs, ...}: {
          nixpkgs.overlays = [
            (
              _: prev: {
                foo = prev.hello;
              }
            )
          ];
          home = {
            packages = [
              pkgs.hello-unfree
            ];
            username = "gerg";
            homeDirectory = "/home/gerg";
            stateVersion = "23.05";
          };
          programs.home-manager.enable = true;
        })
      ];
    };
  };
}
```

ncfavier: Oh, of course, `nixpkgs` is not a submodule option so `pushDownProperties` does it for you. If it was, though, my comment would apply, so I don't like relying on this (you never know, it might become a submodule in the future).

eclairevoyant: Code should be written based on present state, not based on what might possibly happen at some point in the future. Over-engineering is less maintainable and doesn't save as much effort as you might assume. If there really needs to be a change, let it be handled at that time.

ncfavier: This isn't really about over-engineering, it's about being explicit rather than implicit. There's no need to rely on the "present state" when we could just say what we want explicitly.

cvoges12:
I think this leaves two options for what we could do:
- merge the PR as is and wait for the next breaking change to merge a better solution
- close the PR and wait patiently for a better solution in the next breaking change

I'm not sure what's the best decision to go with here. But in my head, that's kind of what it comes down to. We could get something working and clean it up after. Or we could wait it out before properly implementing it. So what do you guys think is the right move?

Gerg-L:
My opinion is merge this then I'll work on a more permanent solution. https://github.com/Gerg-L/home-manager/tree/testing already working on it on this branch

eclairevoyant:
Wait what out? The issue is over a year old. There's no reason to hold back a 1 LOC nonintrusive change for something that should never have been in HM to begin with.

As far as I know there is no "schedule" for breaking changes in HM either, only that it needs to follow [the guidelines in the manual](https://nix-community.github.io/home-manager/index.html#sec-guidelines-back-compat). The general consensus for such types of changes would be deprecate (warn) first whenever possible. If you want stability you would stick to the release branch, so this change would only be merged to the release branch in 23.11.

PS the thumbs down on your first comment is because +1 comments are basically spam and add nothing to the conversation.

emilazy:
There is no need to break compatibility to support specifying a package set in other ways.

Gerg-L:
> There is no need to break compatibility to support specifying a package set in other ways.

that's not really what I mean,


if you're using doing the current idiomatic solution and using
```nix
pkgs = inputs.nixpkgs.legacyPackages.<system>
```
 `lib.homeManagerConfiguration` it's just taking `pkgs.config` and `pkgs.overlays` then setting `config.nixpkgs.` to the default values (at the wrong priority)
 so if I inline everything it looks like this: 
```nix
_module.args.pkgs = import inputs.nixpkgs.legacyPackages.<system>.path {
  system = "<system>";
  inherit (config.nixpkgs) config overlays ; 
}
```
which is not that bad


but if you're using 
```nix
pkgs = import inputs.nixpkgs {
    system = "<system>";
    overlays = [...];
    config = {...};
}
```
then you're instantiating `nixpkgs` twice!
which inline looks somthing like this!
```nix
_module.args.pkgs = import (import inputs.nixpkgs {
    system = "<system>";
    overlays = [...];
    config = {...};
}).path {
  system = "<system>";
  inherit (config.nixpkgs) config overlays; 
};
```
I'm very active in the unofficial NixOS discord server and I see this ALL. THE. TIME.

The idiomatic solution IMO looks like this 
```nix
nixpkgs.system = "<system>";

_module.args.pkgs = import nixpkgs {
  inherit (config.nixpkgs) config overlays;
}
```

My thoughts are deprecate passing `pkgs` with warnings and add a `nixpkgs` argument

emilazy:
Sure, I agree the current behaviour is bad, just that a better behaviour can be supported without breaking existing use.

(That said, passing a package set in directly is useful if you want to immutably override the package set and use [`nixpkgs-disabled.nix`](https://github.com/nix-community/home-manager/blob/0a014a729cdd54d9919ff36b714d047909d7a4c8/modules/misc/nixpkgs-disabled.nix); otherwise you would need to accept `lib` or import Nixpkgs twice. This is how the Home Manager NixOS/nix-darwin module `useGlobalPkgs` works, as well as the `nix-darwin` `pkgs` argument. But this would indeed be a breaking change, unfortunately.)

Gerg-L:
> That said, passing a package set in directly is useful if you want to immutably override the package set
can easily be done with a `nixpkgs.pkgs` option exactly how `nixpkgs` does it

emilazy:
It's not quite that simple; Home Manager needs `lib.evalModules` from Nixpkgs and it's very bad for the `lib` and `pkgs` to mismatch. So you either need to take `lib`, or take `pkgs` and get `lib` from it. nix-darwin implements this: if `pkgs` is specified, it uses `pkgs.lib` and also injects `{ _module.args.pkgs = lib.mkForce pkgs; }`.

Gerg-L:
tracked down the commits to blame for this:
https://github.com/nix-community/home-manager/commit/0c236e13bc6b07a138a7a31fa408a73202cff28c
https://github.com/NixOS/nixpkgs/commit/1c49b81263858c69b932da05ae63a7767b308e74

flake.nix
Gerg-L: This is insane in my opinion but whatever, my next PR will fix all this

ambroisie:
Should a test be added, to check that it won't break in the future?

Gerg-L:
> Should a test be added, to check that it won't break in the future?

What would break? `nixpkgs.`, `config`, and `overlays` are no longer being overridden if they have default values set.

 The actual fix to how `pkgs`  is handled will require breaking change. I'm currently working on that.

Gerg-L:
Turns out overlays needs to not be set `mkDefault` as lists merge

schuelermine:
Is anybody looking to merge this?

Gerg-L:
> Is anybody looking to merge this?

I don't really know what else could be done...

rycee:
Thanks! Looks good to me so finally merged now. Apologies for the long delay.

PaulGrandperrin:
This might have closed https://github.com/nix-community/home-manager/issues/2954 too

PR #4306, opened 2023-08-04
Add QT_PLUGIN_PATH and QML2_IMPORT_PATH to make Qt6 themes and plugins work properly

Aleksanaa:
### Description

Add QT_PLUGIN_PATH and QML2_IMPORT_PATH environment variables to let Qt6 load plugins and qmls installed in home.packages, for example, fcitx5. There are also new options in case not convenient to add a path to home.packages.

#### Original

Fcitx5 needs QT_PLUGIN_PATH to grab input in Qt apps in wayland. Nixpkgs/NixOS module [does set this environment variable](https://github.com/NixOS/nixpkgs/blob/66aedfd010204949cb225cf749be08cb13ce1813/nixos/modules/i18n/input-method/fcitx5.nix#L65-L70), so it's long advised to enable this option together with Home Manager one, but it doesn't make sense to enable both of course.

In order to keep the possibility to reuse this environment variable for others, I've added a new option in Qt module. I'm new to Home Manager modules so I'm not 100% sure this is correct.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@Kranzes @ncfavier @rycee

### DIFF
--- a/modules/misc/qt.nix
+++ b/modules/misc/qt.nix
@@ -79,6 +79,32 @@ in {
         '';
       };
 
+      pluginPath = mkOption {
+        type = types.listOf (types.path);
+        default = [ ];
+        example = literalExpression
+          ''[ "''${pkgs.fcitx5-with-addons}/lib/qt-6-plugins" ]'';
+        description = ''
+          Configure the lookup path for Qt plugins.
+          This sets or prepends to QT_PLUGIN_PATH.
+          Note that home-manager profile path is already included.
+          See [the Qt documentation](https://doc.qt.io/qt-6/deployment-plugins.html) for further details.
+        '';
+      };
+
+      qmlImportPath = mkOption {
+        type = types.listOf (types.path);
+        default = [ ];
+        example =
+          literalExpression ''[ "''${pkgs.qt6.qtwayland}/lib/qt-5.15.9/qml" ]'';
+        description = ''
+          Configure the lookup path for QML modules.
+          This sets or prepends to QML2_IMPORT_PATH.
+          Note that home-manager profile path is already included.
+          See [the Qt documentation](https://doc.qt.io/qt-6/qtqml-syntax-imports.html) for further details.
+        '';
+      };
+
       style = {
         name = mkOption {
           type = types.nullOr types.str;
@@ -141,6 +167,18 @@ in {
     qt.style.package = mkIf (cfg.style.name != null)
       (mkDefault (stylePackages.${toLower cfg.style.name} or null));
 
+    qt.pluginPath =
+      builtins.map (prefix: "${config.home.profileDirectoty}/${prefix}") [
+        pkgs.qt6.qtbase.qtPluginPrefix
+        pkgs.qt5.qtbase.qtPluginPrefix
+      ];
+
+    qt.qmlImportPath =
+      builtins.map (prefix: "${config.home.profileDirectoty}/${prefix}") [
+        pkgs.qt6.qtbase.qtQmlPrefix
+        pkgs.qt5.qtbase.qtQmlPrefix
+      ];
+
     # Necessary because home.sessionVariables doesn't support mkIf
     home.sessionVariables = filterAttrs (n: v: v != null) {
       QT_QPA_PLATFORMTHEME = if cfg.platformTheme == "gtk" then
@@ -150,6 +188,12 @@ in {
       else
         cfg.platformTheme;
       QT_STYLE_OVERRIDE = cfg.style.name;
+      QT_PLUGIN_PATH = "${
+          concatStringsSep ":" cfg.pluginPath
+        }\${QT_PLUGIN_PATH:+:}$QT_PLUGIN_PATH";
+      QML2_IMPORT_PATH = "${
+          concatStringsSep ":" cfg.qmlImportPath
+        }\${QML2_IMPORT_PATH:+:}$QML2_IMPORT_PATH";
     };
 
     home.packages = (if cfg.platformTheme == "gnome" then
--- a/modules/misc/qt.nix
+++ b/modules/misc/qt.nix
@@ -128,7 +128,7 @@ in {
     };
   };
 
-  config = mkIf (cfg.enable && cfg.platformTheme != null) {
+  config = mkIf cfg.enable {
     assertions = [{
       assertion = cfg.platformTheme == "gnome" -> cfg.style.name != null
         && cfg.style.package != null;
@@ -160,9 +160,10 @@ in {
     ] else if cfg.platformTheme == "kde" then [
       pkgs.libsForQt5.plasma-integration
       pkgs.libsForQt5.systemsettings
-    ] else
-      [ pkgs.libsForQt5.qtstyleplugins ])
-      ++ lib.optionals (cfg.style.package != null)
+    ] else if cfg.platformTheme == "gtk2" then
+      [ pkgs.libsForQt5.qtstyleplugins ]
+    else
+      [ ]) ++ lib.optionals (cfg.style.package != null)
       (lib.toList cfg.style.package);
 
     xsession.importedVariables = [ "QT_QPA_PLATFORMTHEME" ]


modules/misc/qt.nix
```diff
@@ -79,6 +79,18 @@ in {
         '';
       };
 
+      pluginPath = mkOption {
+        type = types.nullOr (types.listOf (types.str));
+        default = null;
```
9ary, line 84: Why make this nullable when an empty list would have the same use?

modules/misc/qt.nix
```diff
@@ -150,6 +162,12 @@ in {
       else
         cfg.platformTheme;
       QT_STYLE_OVERRIDE = cfg.style.name;
+      QT_PLUGIN_PATH = if cfg.pluginPath != [ ] then
+        "${
+          concatStringsSep ":" cfg.pluginPath
+        }\${QT_PLUGIN_PATH:+:}$QT_PLUGIN_PATH"
+      else
+        null;
```
ncfavier, line 170: Can't be null, use mkIf

Aleksanaa: `mkIf (cfg.pluginPath != [])`? I don't know how to use mkIf here. And it said
https://github.com/nix-community/home-manager/blob/2014199bfdf72384800a6240e9b44d97a15425c5/modules/misc/qt.nix#L156-L157

ncfavier: You're right, sorry.

ncfavier:
cc @syphoxy for https://github.com/nix-community/home-manager/pull/4067

9ary:
For what it's worth, nixos sets QT_PLUGIN_PATH for qt4 as well (always enabled): https://github.com/NixOS/nixpkgs/blob/394beeefa6133b9645baa5a202f8c52a7b6577cc/nixos/modules/programs/environment.nix#L41
Not sure if there's a possibility for conflict there.

Aleksanaa:
> Not sure if there's a possibility for conflict there.

I think it won't, because it just appends the variables you set to already defined system-wide variables, the same as nixos/fcitx5 does. An example is `XCURSOR_PATH`, it's set like this:

https://github.com/nix-community/home-manager/blob/86dd48d70a2e2c17e84e747ba4faa92453e68d4a/modules/config/home-cursor.nix#L140-L141

And the result is:
```
=> cat /etc/profiles/per-user/aleksana/etc/profile.d/hm-session-vars.sh 
...
export XCURSOR_PATH="$XCURSOR_PATH${XCURSOR_PATH:+:}/etc/profiles/per-user/aleksana/share/icons"
...
```

It will read the already exist variable, and if it exists, append a `:` to it. I'm just doing prepending instead of appending.

9ary:
Ah wait I'm dumb and didn't realize you were prepending, just ignore me then.

Aleksanaa:
> notably, it's not necessary to inject the fcitx5 package plugin path directly. those paths all get merged together already so you merely need to add the global qt5/qt6 plugin path to QT_PLUGIN_PATH.

This doesn't work if user has specified other Qt plugins in nixpkgs config, or if they are not using NixOS at all.

Aleksanaa:
> notably, it's not necessary to inject the fcitx5 package plugin path directly. those paths all get merged together already so you merely need to add the global qt5/qt6 plugin path to QT_PLUGIN_PATH.

This doesn't work if user has specified other Qt plugins in nixpkgs config, or if they are not using NixOS at all. I think we should treat these variables just like PATH so no features get lost when our variables are set.

> those paths all get merged together already

What? Even if you've only installed fcitx5 in home-manager and hasn't configured it in nixos modules?

Aleksanaa:
> notably, it's not necessary to inject the fcitx5 package plugin path directly. those paths all get merged together already so you merely need to add the global qt5/qt6 plugin path to QT_PLUGIN_PATH.

Sorry, just misunderstood. I'm thinking about not hardcoding any packages in modules (so the users don't have to deal with overlays), but this module is already hardcoding it.

9ary:
I prefer the solution proposed by @syphoxy here for two reasons:
- this is what the nixos module does to configure themes (https://github.com/NixOS/nixpkgs/blob/12703229626d8622b5343d6ed513e8c18d02a50a/nixos/modules/config/qt.nix#L111-L116)
- From a user perspective, when adding a qt plugin to home.packages (or nixos system packages), I expect it to become available. This seems to already work out of the box for qt5, and I'm not sure why it is necessary to explicitly set QT_PLUGIN_PATH for qt6.

Considering the second point, it's possible that this is better fixed on the nixpkgs side. I think qt5 might be built with the profile paths hardcoded, so either qt6 should be built that way too, or qt5 should be changed to require the path variable to be set.

> Sorry, just misunderstood. I'm thinking about not hardcoding any packages in modules (so the users don't have to deal with overlays), but this module is already hardcoding it.

The appropriate way to deal with this is to add a package option to the module like all others.

Aleksanaa:
> Considering the second point, it's possible that this is better fixed on the nixpkgs side.

Yes, considering that the user is running NixOS as their system, it's better fixed on nixpkgs side. But there's a problem: what if a user has configured fcitx5 in home-manager, and installed a qt5 package system-wide by some package manager (for example, pacman)?

> From a user perspective, when adding a qt plugin to home.packages (or nixos system packages), I expect it to become available. This seems to already work out of the box for qt5, and I'm not sure why it is necessary to explicitly set QT_PLUGIN_PATH for qt6.

This is more feasible. I'll edit this pr later today.

9ary:
> Yes, considering that the user is running NixOS as their system, it's better fixed on nixpkgs side. But there's a problem: what if a user has configured fcitx5 in home-manager, and installed a qt5 package system-wide by some package manager (for example, pacman)?

I'm not familiar with h-m policy on this kind of setup but it seems a bit over-ambitious to me to try to support packages that are not in nixpkgs (even on non-nixos/nix-darwin systems). Would such a combination even be binary-compatible?

Aleksanaa:
Yes, I think you are right. But there seems to be a good reason for qt6 to remove that qt_plugin_path guessing.

9ary:
I figured out why qt5 finds plugins correctly. There is a patch which makes it scan PATH: https://github.com/NixOS/nixpkgs/blob/82326d676dcb51f262ee7e17b241385c97c646f9/pkgs/development/libraries/qt-5/5.15/qtbase.patch.d/0009-qtbase-qtpluginpath.patch.

This works not only for "installed" packages, but also in nix-shell. For example, the following allows you to run a pyside2 script on wayland without any qt packages in the ambient environment:
```
nix-shell --pure -p python3Packages.pyside2 python3 libsForQt5.qtwayland
```
Not specifying the last package causes qt to complain that the platform plugin is missing and the application does not start.

I could not find an equivalent for qt6. Time to find out why they didn't carry this over.

Aleksanaa:
I guess it's just the work around Qt6 has not been finished then. Qt6 qtwayland is also not included yet. But we can implement this to partly solve it for now.

9ary:
Ah here it is: https://github.com/NixOS/nixpkgs/pull/236770
> The patch was actually intentionally omitted, I felt that loading random plugins from the PATH is just, an incorrect behavior. And if you do need that to happen, you may just set appropriate `environment.profileRelativeEnvVars`, which is already done in `nixos/modules/config/qt.nix`

In that case, I'll settle my vote on @syphoxy's solution which is both simpler and consistent with the nixos module.

To address a concern I had previously, we don't need to worry about plugins installed by system configuration: if the latter needs to deploy system-wide qt plugins then it can set the path itself, so I don't think we'd end up in the same situations as shells which require the system module to function correctly.

> Qt6 qtwayland is also not included yet.

Wait what? It's there and it works perfectly fine for me.

> But we can implement this to partly solve it for now.

I wouldn't be a fan of adding a temporary workaround if we were going to pursue fixing this in the qt6 package, but now I'm convinced this is the correct solution anyway so let's go ahead with it.

Aleksanaa:
> Wait what? It's there and it works perfectly fine for me.

I mean when making packages. But that's not related here, just to mention that qt6 has not been finished probably.

Aleksanaa:
@9ary what do you think of this?

modules/misc/qt.nix
```diff
@@ -79,6 +79,18 @@ in {
         '';
       };
 
+      pluginPath = mkOption {
```
9ary, line 82: Not a fan of this, imo we should do exactly what the nixos module does here: if the qt module is enabled, then set QT_PLUGIN_PATH and QML2_IMPORT_PATH to the user profile. There is no need for an option, and no need to make it customizable.

modules/misc/qt.nix
```diff
@@ -141,6 +153,9 @@ in {
     qt.style.package = mkIf (cfg.style.name != null)
       (mkDefault (stylePackages.${toLower cfg.style.name} or null));
 
+    qt.pluginPath =
+      [ "${config.home.profileDirectory}/${pkgs.qt6.qtbase.qtPluginPrefix}" ];
```
9ary, line 157: Qt5 should also be included in spite of the patch included in nixpkgs.

9ary (changes requested):

modules/misc/qt.nix
Aleksanaa: Does it bring some side effect? It's not useless. We are occupying a environment variable without letting user to decide.

9ary: Hmm, it's true that there's no way to extend this in home-manager like there is in nixos.
Let's still add the other variable.

Aleksanaa: It's done 😉

modules/misc/qt.nix
```diff
@@ -141,6 +153,14 @@ in {
     qt.style.package = mkIf (cfg.style.name != null)
       (mkDefault (stylePackages.${toLower cfg.style.name} or null));
 
+    qt.pluginPath =
+      builtins.map (prefix: "${config.home.profileDirectoty}/${prefix}") [
+        pkgs.qt6.qtbase.qtPluginPrefix
+        pkgs.qt6.qtbase.qtQmlPrefix
+        pkgs.qt5.qtbase.qtPluginPrefix
+        pkgs.qt5.qtbase.qtQmlPrefix
+      ];
```
9ary, line 162: `pkgs.qt*.qtbase.qtQmlPrefix` goes in `QML2_IMPORT_PATH`, it needs to be a separate list.

Aleksanaa: Oh sorry

modules/misc/qt.nix
```diff
@@ -128,7 +156,7 @@ in {
     };
   };
 
-  config = mkIf (cfg.enable && cfg.platformTheme != null) {
+  config = mkIf cfg.enable {
```
9ary, line 159: This should be a separate commit, I've already done this in #4312 fwiw.

modules/misc/qt.nix
```diff
@@ -79,6 +79,34 @@ in {
...
+        type = types.listOf (types.path);
+        default = [ ];
+        example = literalExpression
+          ''[ "''${pkgs.fcitx5-with-addons}/lib/qt-6-plugins" ]'';
+        description = ''
+          Make Qt look for plugins in the paths specified in QT_PLUGIN_PATH
+          when programs start. These values will be separated by colons
+          and (if present) prepended to system-set environment variables.
+          Note that you don't have to specify the path here if it's included
+          in home-manager profile.
```
9ary, line 92: I would rephrase this to something like:
```
Configure the lookup path for Qt plugins. This sets the `QT_PLUGIN_PATH` environment variable.
Note that this is not necessary for any packages in [](#opt-home.packages).
See [the Qt documentation](https://doc.qt.io/qt-6/deployment-plugins.html) for further details.
```

Same for the other one.

9ary (approved):
Looks good to me. Documentation could be improved but this is the right approach.

modules/misc/qt.nix
Aleksanaa: But this pr will become a nonsense if not removed here. And #4312 depends on this pr.

9ary: Feel free to either split it into a commit of your own or cherry-pick mine. There is another change that should be made along with this one that I ought to have put in the same commit.

> And #4312 depends on this pr.

Not really. Only one of the changes depends on this to do anything, the rest doesn't (and there are already other qt6 themes in the module that currently don't work).

9ary:
Sorry if I'm being too pedantic but the commit message for the module gate isn't good. The condition isn't unused, it's a bug. It prevents proper use of the module as it is before this PR even comes into consideration, and the commit message needs to reflect that.

Also, once you remove that gate, it's actually possible for the module to be enabled without setting a platform theme, but as the code is right now, it will end up pulling qtstyleplugins unnecessarily when the platform theme is `null`.

See my version of this in #4312 for the full fix: https://github.com/nix-community/home-manager/pull/4312/commits/4f69a3eac01a4b73e07b898ec75475aa2e48c1ea.

modules/misc/qt.nix
```diff
@@ -79,6 +79,32 @@ in {
...
       };
 
+      pluginPath = mkOption {
+        type = types.listOf (types.path);
+        default = [ ];
+        example = literalExpression
+          ''[ "''${pkgs.fcitx5-with-addons}/lib/qt-6-plugins" ]'';
+        description = ''
+          Congigure the lookup path for Qt plugins.
+          This sets or pependeds to QT_PLUGIN_PATH.
```
9ary, line 89: Some typos:
- Congigure -> Configure
- pependeds -> prepends

Aleksanaa:
> I have one thing to provide as feedback: in the module help text the word prependes should be prepends.

Sorry, it was 3am here yesterday, and I was not so wide awake.

> they would have an incomplete environment for using fcitx with Qt applications.

You can use fcitx5 for non-qt apps without enabling this option I guess. And if the user doesn't issue `qt.enable`, Can we explain it as "don't configure qt for me"? But yeah, enabling it would also make sense.

modules/misc/qt.nix
```diff
@@ -79,6 +79,32 @@ in {
...
 
+      pluginPath = mkOption {
+        type = types.listOf (types.path);
+        default = [ ];
+        example = literalExpression
+          ''[ "''${pkgs.fcitx5-with-addons}/lib/qt-6-plugins" ]'';
+        description = ''
+          Configure the lookup path for Qt plugins.
+          This sets or prepends to QT_PLUGIN_PATH.
+          Note that home-manager profile path is already included.
```
rycee, line 90:
```suggestion
          Note that the Home Manager profile path is already included.
```

modules/misc/qt.nix
```diff
@@ -79,6 +79,32 @@ in {
...
       };
 
+      pluginPath = mkOption {
+        type = types.listOf (types.path);
+        default = [ ];
+        example = literalExpression
+          ''[ "''${pkgs.fcitx5-with-addons}/lib/qt-6-plugins" ]'';
+        description = ''
+          Configure the lookup path for Qt plugins.
+          This sets or prepends to QT_PLUGIN_PATH.
```
rycee, line 89:
```suggestion
          This sets or prepends to {env}`QT_PLUGIN_PATH`.
```

modules/misc/qt.nix
```diff
@@ -79,6 +79,32 @@ in {
         '';
       };
 
+      pluginPath = mkOption {
+        type = types.listOf (types.path);
```
rycee, line 83:
```suggestion
        type = types.listOf types.path;
```

modules/misc/qt.nix
rycee: Same for `qmlImportPath`.

modules/misc/qt.nix
rycee: Same for `qmlImportPath`.

modules/misc/qt.nix
rycee: Same for `qmlImportPath`.

modules/misc/qt.nix
```diff
@@ -141,6 +167,18 @@ in {
     qt.style.package = mkIf (cfg.style.name != null)
       (mkDefault (stylePackages.${toLower cfg.style.name} or null));
 
+    qt.pluginPath =
+      builtins.map (prefix: "${config.home.profileDirectoty}/${prefix}") [
```
rycee, line 171: The map function is always available:

```suggestion
      map (prefix: "${config.home.profileDirectoty}/${prefix}") [
```

modules/misc/qt.nix
```diff
@@ -141,6 +167,18 @@ in {
...
+        pkgs.qt6.qtbase.qtPluginPrefix
+        pkgs.qt5.qtbase.qtPluginPrefix
+      ];
+
+    qt.qmlImportPath =
+      builtins.map (prefix: "${config.home.profileDirectoty}/${prefix}") [
```
rycee, line 177:
```suggestion
      map (prefix: "${config.home.profileDirectoty}/${prefix}") [
```

rycee:
Thanks for the contribution! I haven't followed the discussion and just added some purely stylistic comments. Additionally, please make sure to wrap the commit message bodies at 72 characters per line, see https://cbea.ms/git-commit/#wrap-72

thiagokokada:
> If for some reason a user doesn't use `qt.enable = true` because they aren't explicitly deciding to use Qt but they are deciding to use fcitx, they would have an incomplete environment for using fcitx with Qt applications. That is to say, I can install and use a Qt application without `qt.enable = true` but I cannot install and use fcitx without `qt.enable = true`.

I didn't see this PR before and opened another one that also ends up fixing the same issue (see: #4579). Also, there is already a fix in `master` for the `fcitx5` issue (see #4468), so the approach of this PR would conflict with that fix.

I think the correct way to fix this issue with `fcitx5` and `qt` modules is to actually fix the root cause of this issue: a lack of way to compose environment variables.

I decided to create a PR to fix this for PATH-like variables: #4582. With #4582 merged we can now set `home.sessionSearchVariables.QT_PLUGIN_PATH` in multiple different places and they will all be merged in one entry.

All in all, I don't think the approach of this PR is bad, so I would be happy to see either of those 2 PRs merged. My PR also brings a bunch of refactors and some new features to the `qt` module, so my PR is not exactly a duplicate, but if this approach ended up merged I can change my PR to drop the environment variable changes. However, with #4582 done I am not sure why we would need to add a new configuration to set those 2 variables, since just using `home.sessionSearchVariables` would be enough.

9ary:
I'm partial to #4582 as well, sounds like the best option to me.

Aleksanaa:
I'm not in favor of a PR that's adding too much consideration beyond my needs. I just want the fcitx case to be fixed earlier.

PR #4322, opened 2023-08-07
swayr: add module

9p4:
### Description

Add module: swayr

https://sr.ht/~tsdh/swayr

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/swayr.nix
+++ b/modules/programs/swayr.nix
@@ -1,0 +1,135 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.programs.swayr;
+  tomlFormat = pkgs.formats.toml { };
+  configFile = tomlFormat.generate "config.toml" cfg.settings;
+  finalConfig = pkgs.writeText "swayr.toml"
+    ((builtins.readFile configFile) + cfg.extraConfig);
+in {
+  meta.maintainers = [ lib.hm.maintainers."9p4" ];
+
+  options.programs.swayr = {
+    enable = mkEnableOption "the swayr service";
+
+    settings = mkOption {
+      type = types.nullOr tomlFormat.type;
+      default = { };
+      example = literalExpression ''
+        menu = {
+          executable = "${pkgs.wofi}/bin/wofi";
+          args = [
+            "--show=dmenu"
+            "--allow-markup"
+            "--allow-images"
+            "--insensitive"
+            "--cache-file=/dev/null"
+            "--parse-search"
+            "--height=40%"
+            "--prompt={prompt}"
+          ];
+        };
+
+        format = {
+          output_format = "{indent}<b>Output {name}</b>    <span alpha=\"20000\">({id})</span>";
+          workspace_format = "{indent}<b>Workspace {name} [{layout}]</b> on output {output_name}    <span alpha=\"20000\">({id})</span>";
+          container_format = "{indent}<b>Container [{layout}]</b> <i>{marks}</i> on workspace {workspace_name}    <span alpha=\"20000\">({id})</span>";
+          window_format = "img:{app_icon}:text:{indent}<i>{app_name}</i> — {urgency_start}<b>“{title}”</b>{urgency_end} <i>{marks}</i> on workspace {workspace_name} / {output_name}    <span alpha=\"20000\">({id})</span>";
+          indent = "    ";
+          urgency_start = "<span background=\"darkred\" foreground=\"yellow\">";
+          urgency_end = "</span>";
+          html_escape = true;
+        };
+
+        layout = {
+          auto_tile = false;
+          auto_tile_min_window_width_per_output_width = [
+            [ 800 400 ]
+            [ 1024 500 ]
+            [ 1280 600 ]
+            [ 1400 680 ]
+            [ 1440 700 ]
+            [ 1600 780 ]
+            [ 1680 780 ]
+            [ 1920 920 ]
+            [ 2048 980 ]
+            [ 2560 1000 ]
+            [ 3440 1200 ]
+            [ 3840 1280 ]
+            [ 4096 1400 ]
+            [ 4480 1600 ]
+            [ 7680 2400 ]
+          ];
+        };
+
+        focus = {
+          lockin_delay = 750;
+        };
+
+        misc = {
+          seq_inhibit = false;
+        };
+      '';
+      description = ''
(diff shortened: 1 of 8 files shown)


9p4:
@rycee any update?

billksun:
Just found out about `swayr` and thought it would be nice if it can be configured via `home-manager` as well.

@rycee, would be great if you can take a look.

modules/programs/swayr.nix
```diff
@@ -0,0 +1,133 @@
...
+      description = "swayr package to use.";
+    };
+  };
+
+  config = mkIf cfg.enable (mkMerge [
+    {
+      home.packages = [ cfg.package ];
+
+      # Creating an empty file on empty configuration is desirable, otherwise swayrd will create the file on startup.
+      xdg.configFile."swayr/config.toml" = { source = finalConfig; };
```
teto, line 112: Can I disagree with that ? at least make it optional, I would prefer to manage the config.toml imperatively

teto:
the commit guidelines meiton that your commit should start with `programs.swayr`. 
I have started using swayr last week so this is good but late timing for me ;) 

I think it would be best if the config file could be optional because:
1/ I prefer to manage it imperatively
2/ is it that bad that swayr creates the file ? at least the defaults wont get outdated

teto:
also with the defaults for the config, HM generates an empty config.toml so swayr just crashes:
```
 swayr switch-window
thread 'main' panicked at 'Could not read response from swayrd: Error("EOF while parsing a value", line: 1, column: 0)', swayr/src/client.rs:28:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
. 
And I think I would prefer to enable the systemd service by default if swayr is enabled since it wouldn't work otherwise and it's best to optimize for a "work out of the box" experience.

modules/programs/swayr.nix
```diff
@@ -0,0 +1,133 @@
...
+        Unit = {
+          Description = "A window-switcher & more for sway";
+          Documentation = "https://sr.ht/~tsdh/swayr";
+          After = [ cfg.systemd.target ];
+          PartOf = [ cfg.systemd.target ];
+          X-Restart-Triggers =
+            [ "${config.xdg.configFile."swayr/config.toml".source}" ];
+        };
+        Service = {
+          ExecStart = "${cfg.package}/bin/swayrd";
```
teto, line 126: note that the upstream service sets RUST_BACKTRACE and I think we should do the same as it makes errors more helpful https://git.sr.ht/~tsdh/swayr/tree/main/item/swayr/etc/swayrd.service

9p4:
> And I think I would prefer to enable the systemd service by default if swayr is enabled since it wouldn't work otherwise and it's best to optimize for a "work out of the box" experience.

It does bind to sway-session.target by default, so it should be enabled by default already if sway is being used. I've switched it to graphical-session.target by default instead.

> the commit guidelines meiton that your commit should start with programs.swayr.

I don't see that anywhere. The docs refer to https://github.com/nix-community/home-manager/commit/69f8e47e9e74c8d3d060ca22e18246b7f7d988ef as an "ideal commit message", and this doesn't have the "programs" or "services" in front of it.

Furthermore:
> Finally, when adding a new module, say programs/foo.nix, we use the fixed commit format foo: add module. You can, of course, still include a long description if you wish.

modules/programs/swayr.nix
```diff
@@ -0,0 +1,135 @@
...
+        Unit = {
+          Description = "A window-switcher & more for sway";
+          Documentation = "https://sr.ht/~tsdh/swayr";
+          After = [ cfg.systemd.target ];
+          PartOf = [ cfg.systemd.target ];
+          X-Restart-Triggers = mkIf (cfg.settings != { })
+            [ "${config.xdg.configFile."swayr/config.toml".source}" ];
+        };
+        Service = {
+          Environment = "RUST_BACKTRACE=1";
```
teto, line 127: I tried to add fuzzel to the service PATH with 
```
  systemd.user.services.swayrd.Service = {
   Environment = [ "PATH=${lib.makeBinPath [ pkgs.fuzzel pkgs.wofi ]}" 
   ];
  };
```
which fails with
```
error: The option `home-manager.users.teto.systemd.user.services.swayrd.Service.Environment' has conflicting definition values:
       - In `/nix/store/2p117prwd40gzsnxg8823xskmgkpsaq5-source/hm/profiles/sway.nix': "PATH=/nix/store/1kg8yaz49vxy753dkshzlbcclx2v5g7l-fuzzel-1.9.2/bin:/nix/store/qhg92a8hd0gr8c5yjxw0nkn6xw402qj8-wofi-1.3/bin"
       - In `/nix/store/55y5rag38lyw9sb4d5wwdlcjh52ivkl0-source/modules/programs/swayr.nix': "RUST_BACKTRACE=1"
```
Can you set `Environment = [ "RUST_BACKTRACE=1" ];` instead so the module can merge the different values please ? tested with success locally.

modules/programs/swayr.nix
```diff
@@ -0,0 +1,135 @@
...
+  finalConfig = pkgs.writeText "swayr.toml"
+    ((builtins.readFile configFile) + cfg.extraConfig);
+in {
+  meta.maintainers = [ lib.hm.maintainers."9p4" ];
+
+  options.programs.swayr = {
+    enable = mkEnableOption "the swayr service";
+
+    settings = mkOption {
+      type = tomlFormat.type;
```
teto, line 18: I would prefer `type.nullOr tomlFormat.type;` so one can bypass nix handling of the config

modules/programs/swayr.nix
```diff
@@ -0,0 +1,135 @@
...
+      default = { };
```
teto, line 19: yoiu haven't address my comment about the default: this generates an empty config by default which breaks swayrd, please set better defaults.

teto: ha I see you've done `mkIf (cfg.settings != { }) { source = finalConfig; };` . I wonder if I can use `xdg.configFile."swayr/config.toml".source = mkForce null` or sthg along those lines so that you dont have to modify the module.

teto: nevermind that I can just do `xdg.configFile."swaync/config.json".enable = false;` and bypass the thing. Apologies for the noise.

teto (approved):
Thanks a lot !

PR #4331, opened 2023-08-11
container: init service for linux and mac

michaelCTS:
### Description

Evolved from the original PR based on https://github.com/nix-community/home-manager/pull/2630 by @MaeIsBad, this PR introduces 3 major options:

**`virtualisation.containers`**

From the original PR and nixpkgs, allows managing `regstristries.conf` (used by podman and others, but not docker) declaratively. `containers.conf` is modified by the aforementioned tools and managing by `home-manager` is considered out of scope of this PR.

**`virtualisation.oci-containers`**

Inspired by [`nixpkgs`](https://github.com/NixOS/nixpkgs/blob/db9208ab987cdeeedf78ad9b4cf3c55f5ebd269b/nixos/modules/virtualisation/oci-containers.nix#L335-L341), a simple method to enable containers which will pick the appropriate/supported oci-container implementation of your platform. ATM only linux and mac are supported by activating `virtualisation.podman`.

`virtualisation.oci-containers.enable = true` and you're off.

**`virtualisation.podman`**

Enables support for podman on linux and mac. Linux is more simple as it activates as systemd user service and that's that.

Mac is the difficult child in this relationship as it requires a VM to function. I opted to support basic, declarative VM management using `virtualisation.podman.machines.<name>`. `podman` stores its VM configs in JSON and outputs them using the CLI in JSON as well. Since python is my forté and not `bash` + `jq`, I wrote a helper called `podmactl` which is called at activation of a new home-manager generation.

It diffs the existing machines from requested machines (JSON generated by `nix`) and the required machines are generated and only one is active (podman doesn't allow multiple running machines).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -1414,6 +1414,16 @@ in {
           A new module is available: 'programs.jetbrains-remote'
         '';
       }
+
+      {
+        time = "2023-02-14T08:45:52+00:00";
+        message = ''
+          Three new modules are available:
+            'virtualisation.containers',
+            'virtualisation.oci-containers',
+            'virtualisation.podman'.
+        '';
+      }
     ];
   };
 }
--- a/modules/virtualisation/containers.nix
+++ b/modules/virtualisation/containers.nix
@@ -1,0 +1,76 @@
+{ config, lib, pkgs, ... }:
+
+let
+  cfg = config.virtualisation.containers;
+
+  inherit (lib) mkOption types;
+
+  toml = pkgs.formats.toml { };
+in {
+  meta.maintainers = [ lib.maintainers.michaelCTS ];
+
+  options.virtualisation.containers = {
+    enable = lib.mkEnableOption "the common containers configuration module";
+
+    ociSeccompBpfHook.enable = lib.mkEnableOption "the OCI seccomp BPF hook";
+
+    registries = {
+      search = mkOption {
+        type = types.listOf types.str;
+        default = [ "docker.io" "quay.io" ];
+        description = ''
+          List of repositories to search.
+        '';
+      };
+
+      insecure = mkOption {
+        type = types.listOf types.str;
+        default = [ ];
+        description = ''
+          List of insecure repositories.
+        '';
+      };
+
+      block = mkOption {
+        type = types.listOf types.str;
+        default = [ ];
+        description = ''
+          List of blocked repositories.
+        '';
+      };
+    };
+
+    policy = mkOption {
+      type = types.attrs;
+      default = { };
+      example = lib.literalExpression ''
+        {
+          default = [ { type = "insecureAcceptAnything"; } ];
+          transports = {
+            docker-daemon = {
+              "" = [ { type = "insecureAcceptAnything"; } ];
+            };
+          };
+        }
+      '';
+      description = ''
+        Signature verification policy file.
+        If this option is empty the default policy file from
+        `skopeo` will be used.
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    xdg.configFile."containers/registries.conf".source =
+      toml.generate "registries.conf" {
+        registries = lib.mapAttrs (n: v: { registries = v; }) cfg.registries;
+      };
+
+    xdg.configFile."containers/policy.json".source = if cfg.policy != { } then
+      pkgs.writeText "policy.json" (builtins.toJSON cfg.policy)
+    else
+      "${pkgs.skopeo.src}/default-policy.json";
+  };
+
+}
--- a/modules/virtualisation/oci-containers.nix
+++ b/modules/virtualisation/oci-containers.nix
@@ -1,0 +1,28 @@
+# Equivalent of
+# https://github.com/NixOS/nixpkgs/blob/nixos-unstable/nixos/modules/virtualisation/oci-containers.nix
+{ config, lib, pkgs, ... }:
+
+let
+  cfg = config.virtualisation.oci-containers;
+
+  inherit (lib) mkDefault mkIf mkMerge mkOption types;
+
+  defaultBackend = "podman";
+in {
+  meta.maintainers = [ pkgs.lib.maintainers.michaelCTS ];
+
+  options.virtualisation.oci-containers = {
+    enable = lib.mkEnableOption
+      "a convenience option to enable containers in platform-agnostic manner";
+
+    backend = mkOption {
+      type = types.enum [ "podman" ];
+      default = defaultBackend;
+      description = "Which service to use as a backend for containers.";
+    };
+  };
+
+  config = mkIf (cfg.enable && cfg.backend == "podman") {
+    virtualisation.podman.enable = true;
+  };
+}
--- a/modules/virtualisation/podman.nix
+++ b/modules/virtualisation/podman.nix
@@ -1,0 +1,279 @@
+{ config, lib, pkgs, ... }:
+
+let
+  cfg = config.virtualisation.podman;
+  toml = pkgs.formats.toml { };
+  json = pkgs.formats.json { };
+
+  inherit (lib) mkDefault mkIf mkMerge mkOption types;
+
+  podmanPackage = (pkgs.podman.override { inherit (cfg) extraPackages; });
+
+  # Provides a fake "docker" binary mapping to podman
+  dockerAlias = pkgs.runCommandNoCC
+    "${podmanPackage.pname}-docker-alias-${podmanPackage.version}" {
+      outputs = [ "out" "man" ];
+      inherit (podmanPackage) meta;
+    } ''
+      mkdir -p $out/bin
+      ln -s ${podmanPackage}/bin/podman $out/bin/docker
+
+      mkdir -p $man/share/man/man1
+      for f in ${podmanPackage.man}/share/man/man1/*; do
+        basename=$(basename $f | sed s/podman/docker/g)
+        ln -s $f $man/share/man/man1/$basename
+      done
+    '';
+
+  podmactl = import (pkgs.fetchFromGitHub {
+    owner = "michaelCTS";
+    repo = "podmactl";
+    rev = "v0.0.4";
+    hash = "sha256-7lMOcJZDkNSqBNXxI9iE0yn+hHhEwrWykS+C02NSQCk=";
+  }) { };
+
+  machineOpts = {
+    # Options here are loaded into python. For simplicity, please use
+    # snake_case.
+    options = {
+      active = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          This machine should be started. Only one machine can be active at a time
+        '';
+      };
+
+      qemu_binary = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        example = "''${pkgs.qemu}/bin/qemu-system-x86_64";
+        description = ''
+          Use this to start VM with the qemu appropriate for your architecture.
+        '';
+      };
+
+      # Options passed to Podman machine.
+      # See https://docs.podman.io/en/latest/markdown/podman-machine.1.html
+      cpus = mkOption {
+        type = types.ints.positive;
+        default = 1;
+        description = "The number of CPUs to assign to the VM.";
+      };
+
+      disk_size = mkOption {
+        type = types.ints.positive;
+        default = 100;
+        description = "Size of disk in gigabytes. Can only be increased";
+      };
+
+      image_path = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        example = lib.literalExpression ''
+          builtins.fetchurl "https://builds.coreos.fedoraproject.org/prod/streams/stable/builds/38.20230819.3.0/x86_64/fedora-coreos-38.20230819.3.0-qemu.x86_64.qcow2.xz"'';
+        description = ''
+          Image to be used when starting the VM
+          Can be a local path or a URL to an image.
+          Alternatives can be found at <https://fedoraproject.org/en/coreos/download>.
+        '';
+      };
+
+      memory = mkOption {
+        type = types.ints.positive;
+        default = 2048;
+        description = "RAM in MB to be assigned to the machine";
+      };
+    };
+  };
+
+in {
+  meta.maintainers = [ pkgs.lib.maintainers.michaelCTS ];
+
+  options.virtualisation.podman = {
+    enable = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        This option enables Podman, a daemonless container engine for
+        developing, managing, and running OCI Containers on your Linux System.
+
+        It is a drop-in replacement for the {command}`docker` command.
+      '';
+    };
+
+    enableDockerSocket = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Make the Podman socket available in place of the Docker socket, so
+        Docker tools can find the Podman socket.
+
+        Podman implements the Docker API.
+      '';
+    };
+
+    enableDockerAlias = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Create an alias mapping {command}`docker` to {command}`podman`.
+      '';
+    };
+
+    extraPackages = mkOption {
+      type = with types; listOf package;
+      default = [ ];
+      example = lib.literalExpression "[ pkgs.gvisor ]";
+      description = ''
+        Extra packages to be installed in the Podman wrapper.
+      '';
+    };
+
+    finalPackage = lib.mkOption {
+      type = types.package;
+      internal = true;
+      readOnly = true;
+      default = podmanPackage;
+      description = ''
+        The final Podman package (including extra packages).
+      '';
+    };
+
+    defaultNetwork.extraPlugins = lib.mkOption {
+      type = types.listOf json.type;
+      default = [ ];
+      description = ''
+        Extra CNI plugin configurations to add to Podman's default network.
+      '';
+    };
+
+    machines = lib.mkOption {
+      type = types.attrsOf (types.submodule machineOpts);
+      # One and only one machine may be active at any given time
+      apply = machines:
+        assert ((lib.lists.count (machine: machine.active)
+          (lib.attrsets.attrValues machines)) == 1);
+        machines;
+      default = {
+        podman-machine-default = {
+          active = true;
+          cpus = 2;
+          disk_size = 100;
+          memory = 2048;
+        };
+      };
+      example = lib.literalExpression ''
+        {
+          intel-x86 = {
+            cpus = 2;
+            disk_size = 200;
+            memory = 4096;
+            image_path = "fedora-coreos-38.20230806.3.0-qemu.x86_64.qcow2.xz";
+            qemu_binary = "${pkgs.qemu}/bin/qemu-system-x86_64";
+          };
+        }
+      '';
+      description = ''
+        Virtual machine descriptions when Podman is run in on non-Linux systems.
+      '';
+    };
+
+  };
+
+  config = mkIf cfg.enable (mkMerge [
+    {
+      home.packages = [ cfg.finalPackage ]
+        ++ lib.optional cfg.enableDockerAlias dockerAlias;
+
+      virtualisation.containers = {
+        enable = true; # Enable common /etc/containers configuration
+      };
+    }
+
+    (mkIf pkgs.stdenv.hostPlatform.isLinux (mkMerge [
+      {
+        systemd.user = {
+          services.podman = {
+            Unit = {
+              Description = "Podman API Service";
+              Requires = "podman.socket";
+              After = "podman.socket";
+              Documentation = "man:podman-system-service(1)";
+              StartLimitIntervalSec = 0;
+            };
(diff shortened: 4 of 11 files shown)


MaeIsBad:
I'd be glad to review this once it gets out of the draft stage and help with maintenance.

One thing to note is that for a while the regular nixos `virtualisation.podman.enable` option already installs a user systemd socket you can use

modules/lib/maintainers.nix
```diff
@@ -211,6 +211,12 @@
     githubId = 1472826;
     name = "Max Smolin";
   };
+  michaelCTS = {
+    email = "michael.vogel@cts.co";
+    github = "michaelCTS";
+    githubId = 132582212;
+    name = "Michael Vogel";
+  };
```
rycee, line 219: Since you are already listed in Nixpkgs' maintainer-list.nix you don't need to be listed here as well.

modules/virtualisation/containers.nix
```diff
@@ -0,0 +1,89 @@
+{ config, lib, pkgs, ... }:
+let
+  cfg = config.virtualisation.containers;
+
+  inherit (lib) mkOption types;
+
+  toml = pkgs.formats.toml { };
+in {
+  meta.maintainers = [ lib.hm.maintainers.bad ];
```
rycee, line 9: There is no such maintainer listed anywhere, as far as I can see?

modules/virtualisation/containers.nix
```diff
@@ -0,0 +1,89 @@
...
+  options.virtualisation.containers = {
+
+    enable = lib.mkEnableOption "the common containers configuration module";
+
+    ociSeccompBpfHook.enable = lib.mkEnableOption "the OCI seccomp BPF hook";
+
+    containersConf.settings = mkOption {
+      type = toml.type;
+      default = { };
+      description = "containers.conf configuration";
```
rycee, line 19:
```suggestion
      description = "The {file}`containers.conf` configuration.";
```

modules/virtualisation/containers.nix
```diff
@@ -0,0 +1,89 @@
...
+            docker-daemon = {
+              "" = [ { type = "insecureAcceptAnything"; } ];
+            };
+          };
+        }
+      '';
+      description = ''
+        Signature verification policy file.
+        If this option is empty the default policy file from
+        <literal>skopeo</literal> will be used.
```
rycee, line 64: We now use [Nixpkgs' Markdown](https://nixos.org/manual/nixpkgs/unstable/#sec-contributing-markup) for descriptions so need someting like:

```suggestion
        `skopeo` will be used.
```

modules/virtualisation/podman/podman.nix
```diff
@@ -0,0 +1,165 @@
...
+  options.services.podman = {
+
+    enable = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        This option enables Podman, a daemonless container engine for
+        developing, managing, and running OCI Containers on your Linux System.
+
+        It is a drop-in replacement for the <command>docker</command> command.
```
rycee, line 45: Isn't this true only if `dockerCompat = true`?

modules/virtualisation/podman/podman.nix
```diff
@@ -0,0 +1,165 @@
...
+
+in {
+  meta.maintainers = [ lib.maintainers.michaelCTS ];
+  imports = [
+    (lib.mkRenamedOptionModule [ "virtualisation" "podman" "libpod" ] [
+      "virtualisation"
+      "containers"
+      "containersConf"
+    ])
+  ];
```
rycee, line 35: There has never been such an option before so no need to do anything here:
```suggestion

```

modules/virtualisation/podman/podman.nix
```diff
@@ -0,0 +1,165 @@
...
+    } ''
+      mkdir -p $out/bin
+      ln -s ${podmanPackage}/bin/podman $out/bin/docker
+
+      mkdir -p $man/share/man/man1
+      for f in ${podmanPackage.man}/share/man/man1/*; do
+        basename=$(basename $f | sed s/podman/docker/g)
+        ln -s $f $man/share/man/man1/$basename
+      done
+    '';
```
rycee, line 25: This seems like it should be a feature of the podman package in Nixpkgs, similar to how the yt-dlp package has a `withAlias` argument. For example,

``` nix
pkgs.podman.override { inherit (cfg) extraPackages enableDockerAlias; }
```

Perhaps it would be possible to add this functionality to the podman package in Nixpkgs?

modules/virtualisation/podman/podman.nix
```diff
@@ -0,0 +1,165 @@
...
+      default = false;
+      description = ''
+        Make the Podman socket available in place of the Docker socket, so
+        Docker tools can find the Podman socket.
+
+        Podman implements the Docker API.
+      '';
+    };
+
+    dockerCompat = mkOption {
```
rycee, line 60: Perhaps can have something more descriptive? E.g.,

```suggestion
    enableDockerAlias = mkOption {
```

modules/virtualisation/podman/podman.nix
```diff
@@ -0,0 +1,165 @@
...
+      default = false;
+      description = ''
+        This option enables Podman, a daemonless container engine for
+        developing, managing, and running OCI Containers on your Linux System.
+
+        It is a drop-in replacement for the <command>docker</command> command.
+      '';
+    };
+
+    dockerSocket = mkOption {
```
rycee, line 49: I think prefixing with "enable" will make it a bit clearer:

```suggestion
    enableDockerSocket = mkOption {
```

modules/virtualisation/podman/podman.nix
```diff
@@ -0,0 +1,165 @@
...
+        [
+          pkgs.gvisor
+        ]
+      '';
+      description = ''
+        Extra packages to be installed in the Podman wrapper.
+      '';
+    };
+
+    package = lib.mkOption {
```
rycee, line 89: Typically we call this type of option `finalPackage` (and add `readOnly = true`):
```suggestion
    finalPackage = lib.mkOption {
```

rycee:
Thanks a lot! It would be great to have this functionality. I started adding a few comments. Will continue when I find some more time.

modules/virtualisation/podman/podman.nix
MaeIsBad: Podman is a drop in, docker compat just creates an alias binary for script compatibility.

rycee: Ok, I think it may be my misinterpretation of "drop-in". I was thinking that it would be a drop-in in the sense that if you today have a script that uses the docker command, then after switching you could use the script unchanged with Podman. But it actually is drop-in in the sense that you can use the script with the only change being s/docker/podman/ 🙂

So no need to change, except for the conversion to Markdown
```suggestion
        It is a drop-in replacement for the {command}`docker` command.
```

modules/virtualisation/containers.nix
MaeIsBad: I think this is a leftover from my original PR, where I was the only maintainer. 
I don't feel confident in being the sole maintainer but I would be happy to become a second maintainer for this module.

michaelCTS:
Thanks for the review @rycee . For now, I made only minimal changes after recovering @MaeIsBad's PR. There might be bigger changes coming tomorrow (I get to work on this once a week) as I have a look at how `virtualisation.podman.enable` currently works on nixpkgs. CNI isn't used for networking anymore and there may be more changes to incorporate.

modules/lib/maintainers.nix
michaelCTS: Replace all references with `pkgs.lib.maintainers.michaelCTS`

modules/virtualisation/containers.nix
michaelCTS: I'll let you decide on that once the PR is ready @MaeIsBad . Then you can have an idea of what you're getting yourself into :grin:

MaeIsBad:
With flakes you can point the flake input by using `git+file:///home/user/home-manager`. That's how I test changes to home-manager

modules/virtualisation/podman/podman.nix
michaelCTS: It's a good idea. I'll let somebody else handle that due to a lack of time on my end :innocent:

michaelCTS:
Thanks @MaeIsBad . I don't use flakes, so had to find another way. Turned out it was

```sh
cd $proj
nix-shell -A install
home-manager -I home-manager=$proj switch
```

The PR is ready for review. I'll leave comments where I'm unsure about stuff.

As for the commit messages: only noticed those while checking of the list. What to do? Just squash everything? I tried to add many comments in the code, so a huge commit might be acceptable?
Otherwise I can rebase and reword each commit. Whatever is prefered.

modules/virtualisation/containers.nix
```diff
@@ -0,0 +1,87 @@
...
+
+    ociSeccompBpfHook.enable = lib.mkEnableOption "the OCI seccomp BPF hook";
+
+    # Unimplemented as podman modifies the file and it would have to be merged
+    #  not replaced when activating a new config
+    #    containersConf.settings = mkOption {
+    #      type = toml.type;
+    #      default = { };
+    #      description = "The {file}`containers.conf` configuration.";
+    #    };
```
michaelCTS, line 22: Present in `nixpkgs` https://github.com/NixOS/nixpkgs/blob/230262cda5dc5b704aae91273e2676a9ac65c81e/nixos/modules/virtualisation/containers.nix#L31-L35


Podman modifies the files in the user's home, outside of `home-manager`. Fixing that in this PR is out of scope, so not sure if this comment (and related commented out code) should stay or just be removed.

modules/virtualisation/podman/podmactl/podmactl.py
michaelCTS, line 1: Not sure where to put this. Wanted unit testing so putting it in the `podman.nix` didn't seem right.

modules/virtualisation/podman/podman.nix
```diff
@@ -0,0 +1,276 @@
...
+      '';
+    };
+
+    #    enableNvidia = mkOption {
+    #      type = types.bool;
+    #      default = false;
+    #      description = ''
+    #        Enable use of NVidia GPUs from within podman containers.
+    #      '';
+    #    };
```
michaelCTS, line 116: Requires `virtualisation.containers.containersConf.settings` which is a hassle to implement and out of scope (see comment on the options in `containers.nix`). Remove comment or keep?

modules/virtualisation/podman/podman.nix
```diff
@@ -0,0 +1,276 @@
...
+        enable = true; # Enable common /etc/containers configuration
+        #        containersConf.settings = lib.optionalAttrs cfg.enableNvidia {
+        #          engine = {
+        #            conmon_env_vars =
+        #              [ "PATH=${lib.makeBinPath [ pkgs.nvidia-podman ]}" ];
+        #            runtimes.nvidia =
+        #              [ "${pkgs.nvidia-podman}/bin/nvidia-container-runtime" ];
+        #          };
+        #        };
+      };
```
michaelCTS, line 194: Requires `virtualisation.containers.containersConf.settings` which is a hassle to implement and out of scope (see comment on the options in `containers.nix`).
Delete comment or keep?

modules/virtualisation/podman/podman.nix
```diff
@@ -0,0 +1,276 @@
...
+          image_path = "fedora-coreos-38.20230806.3.0-qemu.x86_64.qcow2.xz";
+          qemu_binary = "$${pkgs.qemu}/bin/qemu-system-x86_64";
+        };
+      };
+      # One and only one machine may be active at any given time
+      apply = machines:
+        assert ((lib.lists.count (machine: machine.active)
+          (lib.attrsets.attrValues machines)) == 1);
+        machines;
+    };
```
michaelCTS, line 175: Unsure how to do this with a proper error message e.g `Only one machine may be active at a time`

modules/virtualisation/podman/podmactl/.gitignore
```diff
@@ -0,0 +1 @@
+result
```
MaeIsBad, line 1: No other modules need to add a .gitignore file, I don't see why this one should. You could maybe put this it into the top level gitignore file

modules/virtualisation/podman/podman.nix
```diff
@@ -0,0 +1,276 @@
...
+    # options here are loaded into python
+    # for simplicity, please use snake_case
+    options = {
+      active = mkOption {
+        type = types.bool;
+        default = false;
+        description =
+          "This machine should be started. Only one machine can be active at a time";
+      };
+      qemu_binary = mkOption {
```
MaeIsBad, line 38: nix options are typically written using camelCase, same for the rest of this file

modules/virtualisation/podman/podman.nix
```diff
@@ -0,0 +1,276 @@
...
+              After = "podman.socket";
+              Documentation = "man:podman-system-service(1)";
+              StartLimitIntervalSec = 0;
+            };
+            Service = {
+              Type = "exec";
+              KillMode = "process";
+              Environment = ''LOGGING=" --log-level=info"'';
+              ExecStart =
+                [ "" "${cfg.finalPackage}/bin/podman $LOGGING system service" ];
```
MaeIsBad, line 214:
```suggestion
                [ "${cfg.finalPackage}/bin/podman" "$LOGGING" "system" "service" ];
```

See my original PR(https://github.com/nix-community/home-manager/pull/2630#discussion_r797006402)

modules/virtualisation/podman/podman.nix
```diff
@@ -0,0 +1,276 @@
...
+            Install.WantedBy = [ "sockets.target" ];
+          };
+
+        };
+      }
+      (mkIf cfg.enableDockerSocket {
+        home.sessionVariables."DOCKER_HOST" =
+          "unix:///run/user/$UID/podman/podman.sock";
+      })
+    ]))
```
MaeIsBad, line 238:
```suggestion
      (mkIf cfg.enableDockerSocket {
        home.sessionVariables."DOCKER_HOST" =
          "unix:///$XDG_RUNTIME_DIR/podman/podman.sock";
      })
    ]))
```

https://github.com/nix-community/home-manager/pull/2630#discussion_r810568878

modules/virtualisation/podman/podmactl/README.md
MaeIsBad, line 1: This feels out of scope for a home-manager module, maybe you should consider moving this to a separate project..? Up to the home-manager devs to decide

modules/virtualisation/containers.nix
michaelCTS: I'll give that a try and see if podman complains :)

modules/virtualisation/oci-containers.nix
```diff
@@ -0,0 +1,31 @@
...
+  inherit (lib) mkDefault mkIf mkMerge mkOption types;
+
+  defaultBackend = "podman";
+in {
+
+  meta.maintainers = [ pkgs.lib.maintainers.michaelCTS ];
+  options.virtualisation.oci-containers = {
+
+    enable = lib.mkEnableOption
+      "a convenience option to enable containers in platform-agnostic manner";
```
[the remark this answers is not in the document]
michaelCTS: This is taken from `nixpkgs` https://github.com/NixOS/nixpkgs/blob/ec68439bdc77ee684a4b2846f03677c5c50920a9/nixos/modules/virtualisation/oci-containers.nix#L335-L349

The problem with `containers` is that it used by `podman`. I'm not versed enough in nix and nixpkgs to not create a recursion problem with `containers.enable` using `containers.backend` to enable `podman` which itself then sets `containers.enable` and configures the settings and other stuff. I gave it a shot and immediately ran into recursion errors.

Since `nixpkgs` doesn't bother with that, I erred on the side of "they probably know better what they're doing that some 1 year nix scrub like me".

modules/virtualisation/podman/podmactl/.gitignore
michaelCTS: Reasonable suggestion.

modules/virtualisation/podman/podman.nix
michaelCTS: This is specifically because it's exported into a JSON which is used by the python script to apply the machine config. There's a comment a few lines up that mentions this.

I'll have a look at the rest of the file and see if I didn't it respect it outside of machine options.

rycee:
Finally had a chance to look at this and overall looks very nice. I made some minor cleanups and squashed the commits but unfortunately was not allowed to push to this branch. Would you mind resetting the PR to https://github.com/nix-community/home-manager/commit/faa4b16358ebd8607499538dbea02bfba5930bd9

The only thing I'm concerned about is the inclusion of the Python project, I would feel more comfortable with it being a separate project and pulled from Nixpkgs. Then I imagine it could be used elsewhere as well, perhaps in nix-darwin or devenv?

modules/virtualisation/podman/podmactl/README.md
rycee: I agree it should be moved into its own repository, if you don't want it under your own account, then perhaps it could be hosted under nix-community, since it may be useful for other Nix related projects?

michaelCTS:
Thanks for the review @rycee and happy new year!

If it's not an issue, I'll move the python code out to a repo I own and later try to get it into nixpkgs or nix-community or something. My experience with nixpkgs hasn't been good as PRs seem to take a considerable amount of time to be merged if they are merged at all, and I don't want that to hold up this PR.

My mac isn't with me at the moment, so the earliest I can test the changes will be this Sunday, 2024-01-07.

Cheers

modules/misc/news.nix
```diff
@@ -1348,14 +1348,24 @@ in
           A new module is available: 'programs.gradle'.
         '';
       }
-
+
```
terlar, line 1351: Noticed some dead white-space here

modules/virtualisation/podman.nix
```diff
@@ -0,0 +1,279 @@
...
+            Service = {
+              Type = "exec";
+              KillMode = "process";
+              Environment = ''LOGGING=" --log-level=info"'';
+              ExecStart = [
+                "${cfg.finalPackage}/bin/podman"
+                "$LOGGING"
+                "system"
+                "service"
+              ];
```
rycee, line 215: This will cause the service file to contain one ExecStart for each entry in the list. I assume you want to run this as a single command?

```suggestion
              ExecStart = lib.concatStringsSep " " [
                "${cfg.finalPackage}/bin/podman"
                "$LOGGING"
                "system"
                "service"
              ];
```

modules/virtualisation/podman.nix
```diff
@@ -0,0 +1,279 @@
...
+              Requires = "podman.socket";
+              After = "podman.socket";
+              Documentation = "man:podman-system-service(1)";
+              StartLimitIntervalSec = 0;
+            };
+
+            Service = {
+              Type = "exec";
+              KillMode = "process";
+              Environment = ''LOGGING=" --log-level=info"'';
```
rycee, line 209: Generally speaking it is best to make the `Environment` attributes have list values to allow for merging:

```suggestion
              Environment = [ ''LOGGING="--log-level=info"'' ];
```

rycee:
Sorry about the delay. I had a chance to try this out. But I don't know how to get it to work. Firstly, the generated systemd service file seems to be invalid due to generating multiple ExecStart fields. Once I turned into a single command I get an error

    Error: command required for rootless mode with multiple IDs:
           exec: "newuidmap": executable file not found in $PATH

I didn't do any special configuration, just added

    virtualisation.podman.enable = true;

in my HM configuration. Do I need some additional configuration to get the service up?

michaelCTS:
I'm sorry, I'll have to close this. The [podman update to 5.0](https://blog.podman.io/2024/03/podman-5-0-has-been-released/) made changes to `podman machine` that I'd have to integrate and I've since moved away from trying to support Mac as my company has another solution.

Anybody is free to try and take this over. I'll leave my branch up.

jlbribeiro:
@michaelCTS Thank you so so much for your work on this PR (and the reviewers for their feedback on it)!

My humble suggestion/request for whoever picks up this wonderful work (unless #4801 gets merged before): I think macOS/Darwin support should _not_ block the merge of this module, as the [original @MaeIsBad PR this one is based off](https://github.com/nix-community/home-manager/pull/2630) and #4336 indicate, this has been a need for a while (at least for me).

Imho, NixOS should be the priority, given it would enable proper rootless `podman`, since https://github.com/NixOS/nixpkgs/issues/207050 and https://github.com/NixOS/nixpkgs/issues/259770 are open, and the discussion in https://github.com/containers/podman/discussions/20573 seems to point in the direction of using `systemd --user` units, and not system `systemd` + `User=` (previously discussed in https://github.com/containers/podman/issues/12778).

After supporting NixOS, generic Linux systems, and only then, macOS/Darwin (which will always be a bit more complex). Linux-only support would already be a great improvement on QoL regarding `podman` on Nix-managed systems.
(With that said, taking the time to design an API that is extension-friendly (so as to allow supporting other platforms) is always good.)

PR #4334, opened 2023-08-12
lf: simplify option validation

musjj:
### Description

Don't try to validate a limited set of hardcoded options, instead just convert them as-is. Now, users can keep all their options in a single attribute set, including arbitrary `user_{option}`s which was impossible to express with a hard-coded submodule. As a plus, there is also less maintainence burden.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@owm111

### DIFF
--- a/modules/programs/lf.nix
+++ b/modules/programs/lf.nix
@@ -2,55 +2,7 @@
 
 with lib;
 
-let
-  cfg = config.programs.lf;
-
-  knownSettings = {
-    anchorfind = types.bool;
-    color256 = types.bool;
-    dircounts = types.bool;
-    dirfirst = types.bool;
-    drawbox = types.bool;
-    globsearch = types.bool;
-    icons = types.bool;
-    hidden = types.bool;
-    ignorecase = types.bool;
-    ignoredia = types.bool;
-    incsearch = types.bool;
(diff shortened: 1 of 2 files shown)


modules/programs/lf.nix
```diff
@@ -68,27 +20,18 @@ in {
       };
 
       settings = mkOption {
-        type = lfSettingsType;
+        type = types.attrs;
```
ncfavier, line 23: `types.attrs` is deprecated. We should use a more precise type that covers everything we need, like `let val = oneOf [ bool int str ]; in attrsOf (either val (listOf val))`.

musjj: Done. Interestingly, `oneOf [ int str (listOf int) (listOf str) ]` doesn't work. You need to do `oneOf [ int str (listOf (either int str)) ]`, I wonder why.

ncfavier: Because the NixOS module system is stupid. `(listOf int).check ["a" "b"]` returns `true` because `check` is non-recursive, so that branch is selected and then `int.check "a"` fails.

ncfavier (approved):

PR #4339, opened 2023-08-13
qcal: add module

antonmosich:
### Description
@teto suggested I write a home-manager module for qcal. qcal is a CLI calendar application, which does not cache anything, unlike most other similar applications.
While I played around in the recent months, I still feel rather new to nix, please review my code carefully.
My module and tests are heavily inspired by the existing khal module. I tried to make the tests such that config files can be semantically compared, not just diffed. The only problem I have there is that for qcal the "order" of calendars doesn't matter and as such the order in the list in the json file doesn't matter, but that is not included in the "semantic" comparison. If you have an idea on how to implement that, please tell me.

qcal has been very recently added to nixpkgs, ~qcal hasn't even made it to nixpkgs-unstable yet,~ so consider that when reviewing this PR.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/qcal.nix
+++ b/modules/programs/qcal.nix
@@ -1,0 +1,66 @@
+{ config, lib, pkgs, ... }:
+
+let
+
+  cfg = config.programs.qcal;
+
+  qcalAccounts = lib.attrValues
+    (lib.filterAttrs (_: a: a.qcal.enable) config.accounts.calendar.accounts);
+
+  rename = oldname:
+    builtins.getAttr oldname {
+      url = "Url";
+      userName = "Username";
+      passwordCommand = "PasswordCmd";
+    };
+
+  filteredAccounts = let
+    mkAccount = account:
+      lib.filterAttrs (_: v: v != null) (with account.remote; {
+        Url = url;
+        Username = if userName == null then null else userName;
+        PasswordCmd =
+          if passwordCommand == null then null else toString passwordCommand;
+      });
+  in map mkAccount qcalAccounts;
+
+in {
+  meta.maintainers = with lib.maintainers; [ antonmosich ];
+
+  options = {
+    programs.qcal = {
+      enable = lib.mkEnableOption "qcal, a CLI calendar application";
+
+      timezone = lib.mkOption {
+        type = lib.types.singleLineStr;
+        default = "Local";
+        example = "Europe/Vienna";
+        description = "Timezone to display calendar entries in";
+      };
+
+      defaultNumDays = lib.mkOption {
+        type = lib.types.ints.positive;
+        default = 30;
+        description = "Default number of days to show calendar entries for";
+      };
+    };
+
+    accounts.calendar.accounts = lib.mkOption {
+      type = with lib.types;
+        attrsOf
+        (submodule { options.qcal.enable = lib.mkEnableOption "qcal access"; });
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    home.packages = [ pkgs.qcal ];
+
+    xdg.configFile."qcal/config.json".source =
+      let jsonFormat = pkgs.formats.json { };
+      in jsonFormat.generate "qcal.json" {
+        DefaultNumDays = cfg.defaultNumDays;
+        Timezone = cfg.timezone;
+        Calendars = filteredAccounts;
+      };
+  };
+}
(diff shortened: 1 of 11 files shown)


modules/accounts/calendar.nix
```diff
@@ -144,6 +144,7 @@ in {
         (import ../programs/vdirsyncer-accounts.nix)
         (import ../programs/khal-accounts.nix)
         (import ../programs/khal-calendar-accounts.nix)
+        (import ../programs/qcal-accounts.nix)
```
rycee, line 147: Should be possible to skip this line (and delete `qcal-accounts.nix`)

```suggestion
```

if you add

```nix
accounts.calendar.accounts = mkOption {
  type = with types;
    attrsOf (submodule {
      options.qcal.enable = mkEnableOption "qcal access";
    });
};
```

to the `options` attribute of `qcal.nix`.

modules/programs/qcal.nix
```diff
@@ -0,0 +1,56 @@
...
+    (lib.mapAttrs' (name: value: lib.nameValuePair (rename name) value))
+    # qcal wants the password command as a string, not as a list of strings
+    (lib.mapAttrs (_: value:
+      if lib.isString value then value else lib.concatStringsSep " " value))
+  ]);
+
+in {
+  options.programs.qcal = {
+    enable = lib.mkEnableOption "qcal, a CLI calendar application";
+    timezone = lib.mkOption {
```
rycee, line 32: Please separate options by empty lines. E.g.,

```suggestion

    timezone = lib.mkOption {
```

tests/modules/programs/qcal/http-calendar.nix
```diff
@@ -0,0 +1,19 @@
...
+  };
+
+  test.stubs = { qcal = { }; };
+
+  nmt.script = let
+    shared = (import ./shared.nix);
+    compare = shared.semanticallyCompare (lib.getExe pkgs.jq);
+  in ''
+    # Semantically compare the expected json output with the generated json
+    ${compare ./http-calendar.json-expected}
```
rycee, line 17: I wouldn't worry about these semantic comparisons, the JSON output should be deterministic so you should be able to just put the expected output as-is in the result files.

modules/programs/qcal.nix
```diff
@@ -0,0 +1,56 @@
...
+    (builtins.getAttr "remote")
+    # Get only the attributes we care about
+    (lib.getAttrs [ "url" "userName" "passwordCommand" ])
+    (lib.filterAttrs (_: attr: !(isNull attr)))
+    # qcal expects these values under other names, so we rename the keys in the attribute set
+    (lib.mapAttrs' (name: value: lib.nameValuePair (rename name) value))
+    # qcal wants the password command as a string, not as a list of strings
+    (lib.mapAttrs (_: value:
+      if lib.isString value then value else lib.concatStringsSep " " value))
+  ]);
```
rycee, line 27: This seems overly complicated, wouldn't something like
```suggestion
  filteredAccounts = let
    mkAccount = account:
      lib.filterAttrs (_: v: v != null) (with account.remote; {
        Url = url;
        Username = userName;
        PasswordCmd =
          if passwordCommand == null then null else toString passwordCommand;
      });
  in map mkAccount qcalAccounts;
```
suffice?

rycee:
Thanks for the contribution! I've added a few comments 🙂

modules/accounts/calendar.nix
antonmosich: Thanks for the suggestion, done

modules/programs/qcal.nix
antonmosich: Yes, that is much better - I had come up with what I initially had and didn't quite know how to best improve it. Username could be null as well though. Added that to the PR

tests/modules/programs/qcal/http-calendar.nix
antonmosich: I see, that's nice. I feel like test cases are much more readable (and easier to write as well) if avoiding the minimized one-line json file though. I can of course take it out, but please also consider that side.

rycee: In that case it may be better to make the module itself generate the pretty JSON file. E.g., by using `pkgs.formats.json { }`. Basically it would be to replace the line

        xdg.configFile."qcal/config.json".text = builtins.toJSON {

by something like

```
xdg.configFile."qcal/config.json".source = let
in jsonFormat.generate "qcal.json" {
```

antonmosich: Implemented it this way. Thanks for the suggestion!

rycee:

antonmosich:
Thank you for your time and reviews!

PR #4340, opened 2023-08-13
Improve the clarity of the comment home.sessionVariables comment

strangemonad:
### Description

Given that the default `home.nix` template is meant to be part of the default out of box experience when you run `home-manager init`, I felt like the comment could be made clearer to new-comers so I took a stab.

### Checklist

  I'm not sure what component this is... help

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

No obvious maintainer

### DIFF
--- a/home-manager/home-manager
+++ b/home-manager/home-manager
@@ -397,8 +397,10 @@ $xdgVars
     # '';
   };
 
-  # You can also manage environment variables but you will have to manually
-  # source
+  # Home Manager can also manage your environment variables through
+  # 'home.sessionVariables'. If you don't want to manage your shell through Home
+  # Manager then you have to manually source 'hm-session-vars.sh' located at
+  # either
   #
   #  ~/.nix-profile/etc/profile.d/hm-session-vars.sh
   #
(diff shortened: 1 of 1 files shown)


home-manager/home-manager
```diff
@@ -392,16 +392,17 @@ $xdgVars
     # '';
   };
 
-  # You can also manage environment variables but you will have to manually
-  # source
+  # Home Manager can also manage your environment variables through
+  # 'home.sessionVariables'. If you don't want to manage your shell through 
+  # Home Manager but still want it to manage env vars you'll have to manually source
```
rycee, line 397: I think this can be simplified a bit to something like:

```suggestion
  # Home Manager then you have to manually source
```

Beside being a bit shorter, this has the benefit of making it a bit more forceful. We'll want to avoid users skipping this sourcing since quite a few modules would not work correctly if they can't set session variables.

rycee:
About the CI error, if you rebase on latest master it should be resolved.

About the commit message, use `home-manager` as component (and shorten the header line). Perhaps something like

```
home-manager: improve configuration template

Specifically, make the comment on `home.sessionVariables` more helpful.
```

PR #4342, opened 2023-08-14
home-manager: allow the tool to be installed in submodule setups

musjj:
### Description

Is there a reason why the standalone tool is not allowed to be installed when you are using home-manager as a submodule? I guess it's not necessary since you can just use `nixos-rebuild` instead. But if the user chooses to enable it anyways, IMO it should be properly installed.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/programs/home-manager.nix
+++ b/modules/programs/home-manager.nix
@@ -27,7 +27,7 @@ in {
     };
   };
 
-  config = mkIf (cfg.enable && !config.submoduleSupport.enable) {
+  config = mkIf cfg.enable {
     home.packages =
       [ (pkgs.callPackage ../../home-manager { inherit (cfg) path; }) ];
   };


rycee:
Yeah, it's been a long-standing wish of mine to install home-manager also when used as a NixOS module. The tool would need to be reduced quite a bit when used like that, though. For example, we cannot support sub-commands like `option`, `build`, `instantiate`, `switch`, etc.

They more or less work right now mostly as an implementation detail; basically that HM maintains its own profile even when used inside a NixOS profile. We'll probably eventually move to place the HM generation inside the NixOS profile and then the standalone and NixOS module versions no longer will be interoperable.

Pretty much the only sub-command that we could support would be things like `packages` and `news`.

musjj:
I see, so you're *not* supposed to run `home-manager switch`, etc. in a submodule setup? I'm currently using it because it's slightly faster since the system configuration isn't processed.

I guess I'll close it for now, since it's not really ready yet.

ncfavier:
I don't see why it couldn't be supported. `nix build "$flakeAttr.config.home-manager.users.$user.home.activationPackage`

PR #4350, opened 2023-08-16
bat: generate cache file in XDG cache home

### LINKED ISSUE #4345: Bat builds cache without xdg dirs being set

### Issue description

When rebuilding with the `bat` package enabled, it automatically runs `bat cache --rebuild` however at the time of running this, XDG dirs don't seem to be set, as it places the cache in .cache rather than where it should go (.local/cache in my case). Bat itself does follow xdg dirs as it places the files in the correct location when I run the command myself

### Maintainer CC

None given but @lovesegfault commited the addition of rebuilding the cache

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.1.41, NixOS, 23.11 (Tapir), 23.11.20230801.9e1960b`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.15.1`
```

rycee:
### Description

Fixes #4345

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/bat.nix
+++ b/modules/programs/bat.nix
@@ -72,8 +72,11 @@ in {
       (name: body: { "bat/themes/${name}.tmTheme" = { text = body; }; }));
 
     home.activation.batCache = hm.dag.entryAfter [ "linkGeneration" ] ''
-      $VERBOSE_ECHO "Rebuilding bat theme cache"
-      $DRY_RUN_CMD ${lib.getExe package} cache --build
+      (
+        export XDG_CACHE_HOME=${escapeShellArg config.xdg.cacheHome}
+        $VERBOSE_ECHO "Rebuilding bat theme cache"
+        $DRY_RUN_CMD ${lib.getExe package} cache --build
+      )
     '';
   };
 }


rycee:
This is a temporary fix. I think we'll need to set all the XDG variables early in the script to make sure they apply everywhere.

KaitlynEthylia:
I haven't had the opportunity to test it, but I see it's the config home that's being set here, not the cache. that seems incorrect to me?

rycee:
Doh, it should of course be `XDG_CACHE_HOME` 🙂

KaitlynEthylia:
This doesn't appear to be working correctly, the config is still generated in .cache, can anyone else verify to be sure I'm not the fault here?

KaitlynEthylia:
> This doesn't appear to be working correctly, the config is still generated in .cache, can anyone else verify to be sure I'm not the fault here?

Oh yes, my bad. I remember now that bat itself prefers BAT_CACHE_PATH over XDG (checking the source code, XDG is still checked but only when built for MacOS? I don't understand why but there we are), perhaps setting both variables should do it

rycee:
Looking at

https://github.com/sharkdp/bat/blob/00b7e3b864a49762d7f666ac970f992386392e33/src/bin/bat/directories.rs#L56-L57

seems to indicate that XDG_CACHE_HOME should be honored by the [`dirs::cache_dir()`](https://docs.rs/dirs/latest/dirs/fn.cache_dir.html) function. I don't think it should be necessary but I updated the PR to use `export XDG_CACHE_HOME=…`.

Could you also verify that you have

    xdg.cacheHome = "${config.home.homeDirectory}/.local/cache"

in your configuration?

KaitlynEthylia:
apologies for my absence, I've been away from my nix system. I can confirm that my directories were set correctly, and I noticed the use of the dirs crate as a fallback but only after that comment.

whatever the reason, it works as expected now!

rycee:
Cool, I merged it now 🙂

PR #4357, opened 2023-08-17
git: replace gitToInit with lib function

### LINKED ISSUE #4354: programs.git: use `lib.generators.toGitINI` instead of `gitToIni`
Is there any good reason that `git.nix` defines an own function [`gitToIni`](https://github.com/nix-community/home-manager/blob/2471d965a3522025157a790fc49c3567fd56e26e/modules/programs/git.nix#L49)  instead of using [`lib.generators.toGitINI`](https://github.com/NixOS/nixpkgs/blob/cde2941c60c5bd68e99682af12745978ca4cd0d3/lib/generators.nix#L197C2-L197C2)?

If not, I'd like to normalize it by using the generator function.

VwCSXg:
### Description
Remove the function gitToInit and underlying functions in favor of using the function lib.generators.toGitINI from nixpkgs. 

closes #4354
depends on https://github.com/NixOS/nixpkgs/pull/250220

### Checklist

  - works when run against https://github.com/VwCSXg/nixpkgs/tree/update-gitINI-generator: `nix develop --override-input nixpkgs 'git+https://github.com/VwCSXg/nixpkgs?ref=update-gitINI-generator' --ignore-environment .#all`

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@rycee

### DIFF
--- a/modules/programs/git.nix
+++ b/modules/programs/git.nix
@@ -6,50 +6,6 @@ let
 
   cfg = config.programs.git;
 
-  # create [section "subsection"] keys from "section.subsection" attrset names
-  mkSectionName = name:
-    let
-      containsQuote = strings.hasInfix ''"'' name;
-      sections = splitString "." name;
-      section = head sections;
-      subsections = tail sections;
-      subsection = concatStringsSep "." subsections;
-    in if containsQuote || subsections == [ ] then
-      name
-    else
-      ''${section} "${subsection}"'';
-
-  mkValueString = v:
-    let
-      escapedV = ''
-        "${
-          replaceStrings [ "\n" "	" ''"'' "\\" ] [ "\\n" "\\t" ''\"'' "\\\\" ] v
-        }"'';
-    in generators.mkValueStringDefault { } (if isString v then escapedV else v);
-
-  # generation for multiple ini values
-  mkKeyValue = k: v:
-    let
-      mkKeyValue =
-        generators.mkKeyValueDefault { inherit mkValueString; } " = " k;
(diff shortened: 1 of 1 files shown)


tests/modules/programs/git/git-expected.conf
```diff
@@ -1,64 +1,64 @@
...
-	path = "~/path/to/conditional.inc"
+	path = ~/path/to/conditional.inc
 
 [includeIf "gitdir:~/src/dir"]
-	path = "@git_include_path@"
+	path = /nix/store/q803rx1jj6pnfpza5dm8d8nkyi0aqwm4-hm_gitconfig
 
 [includeIf "gitdir:~/src/otherproject"]
-	path = "@git_named_include_path@"
+	path = /nix/store/y0c46dscy39gfhbd59b2vjjiysry5vpp-hm_gitconfigwork
```
ncfavier, line 64: !

ncfavier (changes requested):
The quote removal seems like a no-go: I have config values with `#` that definitely should be quoted. Why does the nixpkgs generator remove quotes?

ncfavier:
Proper quoting was added in HM after the extraction: https://github.com/nix-community/home-manager/pull/1209

This will have to be upstreamed as well before this PR can proceed.

VwCSXg:
> Proper quoting was added in HM after the extraction: #1209
Thanks for pointing that out, I missed that.

> This will have to be upstreamed as well before this PR can proceed.
You're right, I'll try that.

rycee:
Thanks! Looks good now :-)

PR #4360, opened 2023-08-18
zsh: allow enabling syntax highlighters

### LINKED ISSUE #4353: Can't enable syntax highlighter in zsh

### Issue description

I want to enable brackets syntax highlighting in zsh, but I don't know how to do that with `programs.zsh.syntaxHighlighting`, since it doesn't have an example. The only options (other than `enable` and `package`) is `styles` that edits the `ZSH_HIGHLIGHT_STYLES` variable.

According to the [documentation](https://github.com/zsh-users/zsh-syntax-highlighting/blob/master/docs/highlighters.md) in order to enable a syntax highlighter, we need to change the `ZSH_HIGHLIGHT_HIGHLIGHTERS` variable.
I think the solution is to add a new string option `syntaxHighlighting.highlighters` that adds this line to the .zshrc: `ZSH_HIGHLIGHT_HIGHLIGHTERS+=('${cfg.syntaxHighlighting.highlighters}')`

### Maintainer CC

@ThinkChaos introduced this feature in https://github.com/nix-community/home-manager/commit/f5f64ac022ed379442286a4e81a2a7cdc552cb28

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.1.45, NixOS, 23.11 (Tapir), 23.11pre514857.caac0eb6bdca`
 - multi-user?: `yes`
 - sandbox: `yes`
 - channels(root): `"nixos, stable-23.05"`
 - channels(gep): `"nixpkgs"`
 - nixpkgs: `/home/gep/.nix-defexpr/channels/nixpkgs`
```

gepbird:
### Description

I added an option to enable individual highlighters (eg. bracket highlighting) for zsh.
Closes #4353.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@ThinkChaos

### DIFF
--- a/modules/programs/zsh.nix
+++ b/modules/programs/zsh.nix
@@ -226,9 +226,20 @@ let
 
       package = mkPackageOption pkgs "zsh-syntax-highlighting" { };
 
+      highlighters = mkOption {
+        type = types.listOf types.str;
+        default = [ ];
+        example = [ "brackets" ];
+        description = ''
+          Highlighters to enable
+          See the list of highlighters: <https://github.com/zsh-users/zsh-syntax-highlighting/blob/master/docs/highlighters.md>
+        '';
+      };
+
       styles = mkOption {
         type = types.attrsOf types.str;
         default = {};
+        example = { comment = "fg=black,bold"; };
         description = ''
           Custom styles for syntax highlighting.
           See each highlighter's options: <https://github.com/zsh-users/zsh-syntax-highlighting/blob/master/docs/highlighters.md>
@@ -624,6 +635,7 @@ in
           # https://github.com/zsh-users/zsh-syntax-highlighting#faq
         ''
           source ${cfg.syntaxHighlighting.package}/share/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh
+          ZSH_HIGHLIGHT_HIGHLIGHTERS+=(${lib.concatStringsSep " " (map lib.escapeShellArg cfg.syntaxHighlighting.highlighters)})
           ${lib.concatStringsSep "\n" (
               lib.mapAttrsToList
                 (name: value: "ZSH_HIGHLIGHT_STYLES+=(${lib.escapeShellArg name} ${lib.escapeShellArg value})")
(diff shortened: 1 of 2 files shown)


modules/programs/zsh.nix
```diff
@@ -622,6 +632,7 @@ in
           # https://github.com/zsh-users/zsh-syntax-highlighting#faq
         ''
           source ${cfg.syntaxHighlighting.package}/share/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh
+          ZSH_HIGHLIGHT_HIGHLIGHTERS+=(${lib.concatStringsSep " " cfg.syntaxHighlighting.highlighters})
```
ThinkChaos, line 635: The values should be escaped:
```suggestion
          ZSH_HIGHLIGHT_HIGHLIGHTERS+=(${lib.concatStringsSep " " (map lib.escapeShellArg cfg.syntaxHighlighting.highlighters)})
```

Tests will probably need to be adjusted, and obviously make sure this still works!

modules/programs/zsh.nix
```diff
@@ -224,6 +224,16 @@ let
...
+        example = [ "brackets" ];
+        description = ''
+          Highlighters to enable
+          See the list of highlighters: <https://github.com/zsh-users/zsh-syntax-highlighting/blob/master/docs/highlighters.md>
+        '';
+      };
+
       styles = mkOption {
         type = types.attrsOf types.str;
         default = {};
```
ThinkChaos, line 239: Let's also add an example as you noted in the issue:

```suggestion
        default = {};
        example = { comment = "fg=black,bold"; };
```

modules/programs/zsh.nix
gepbird: Wow, IIRC I escaped it for the first try and it didn't work. But now I tried it again and no issues, thanks!

gepbird:
@ThinkChaos thanks for the review!

ThinkChaos (approved):
Looks all good to me!

Sorry I couldn't get to this before, was on vacation :)

teto:
Worked well here, could you rebase to trigger CI plz ?

PR #4361, opened 2023-08-19
rbw: enable shell completion

nim65s:
### Description

rbw module autocompletion was not enabled. So I copy/pasted this feature from other modules.

NB: I only checked this on fish.

### Checklist

-->

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@ambroisie

### DIFF
--- a/modules/programs/rbw.nix
+++ b/modules/programs/rbw.nix
@@ -95,11 +95,47 @@ in {
         managed by Home Manager.
       '';
     };
+
+    enableBashIntegration = mkOption {
+      default = true;
+      type = types.bool;
+      description = ''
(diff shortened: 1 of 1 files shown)


ambroisie:
Shell completion works just fine in my configuration, I believe this is an issue with your config rather than the module.

ambroisie:
Shell integrations are, at least IMHO, not supposed to be used for completion, but other integrations that are not sourced when shell completion is otherwise correctly configured.

nim65s:
You're right, thanks. 
On my side, the same config is working on nixos and ubuntu, but not on arch… I guess I'll found out.

PR #4363, opened 2023-08-19
readme: improve overall quality

trueNAHO:
### Description

The following changes were made:

- remove unused reference definition
- sort reference definitions
- internally reference "Words of warning" section
- consistently reference
- fix typo
- consistently wrap text at 72 characters
- format HTML code

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


michaelCTS:
Not a maintainer but why `72` chars? That seems like the biggest source of change in the diff atm.

trueNAHO:
> why `72` chars? *[...]* Why not `80`, `100`, or `110`? `72 seems a bit random and quite small.

Despite a 72-character text width being unconventional for `README`s, it is the current average width in the file. I would have preferred uniform wrapping at 80 characters, but that would have required changes to nearly every line.

> That seems like the biggest source of change in the diff atm. 

907c82eee40d8dd86c6f57d3f07dc0fed81324b5 has only 12 insertions and 8 deletions.

README.md
```diff
@@ -66,67 +69,69 @@ all this out you can go ahead and read the rest of this text.
...
-   NixOS and Darwin, this is the only available choice. It is also
-   recommended for people on NixOS or Darwin that want to manage their
-   home directory independently of the system as a whole. See
-   [Standalone installation][manual standalone install] in the manual
-   for instructions on how to perform this installation.
+   [NixOS][] and Darwin, this is the only available choice. It is also
+   recommended for people on [NixOS][] or Darwin that want to manage
+   their home directory independently of the system as a whole. See
+   [Standalone installation][manual standalone install] in the
+   [manual][] for instructions on how to perform this installation.
```
rycee, line 87: I suspect that these added links will make it unclear to the user which link they actually want to use. Basically, there is only one link in this section that is relevant, the one that goes to the installation instructions and it would probably be best if that was the only thing that draws the user's attention.

rycee: Same goes for 2 and 3 below.

rycee:
Thanks for the contribution! I think it looks very nice in general!

I'm personally not a big fan of putting too many links to the same thing in the document. Basically, the user is probably already familiar with, e.g., Nix and Nixpkgs so I feel it is not necessary to put emphasis on them more than on their first mention.

I'm not having strong feelings about it, though, except in the list of the three installation methods. There I think we should make sure to guide the user to the actual installation description.

README.md
trueNAHO: > I suspect that these added links will make it unclear to the user which link they actually want to use.

Thanks for the feeback! 19c3761958ae134927fe2091a25038b506fae96b should fix this.

> Basically, there is only one link in this section that is relevant, the one that goes to the installation instructions and it would probably be best if that was the only thing that draws the user's attention.

cfcc07502c644f10eb5c23e8596ad867d5e1cb91 addresses this by following the [Home Manager Manual](https://nix-community.github.io/home-manager/index.html#ch-installation) phrasing.

trueNAHO:
https://github.com/nix-community/home-manager/pull/4363#discussion_r1329981053 should resolve your concerns. If a 72-character text width is preferred over 80 characters, we can revert f88c0cdabf83b716fe6ce6d5ed42b128dbce8a1d.

trueNAHO:
LGTM

rycee:
Thanks! I've squashed the commits and merged to master now 🙂

Note, I did remove the section number from the links that point into the manual. Those numbers are generated automatically so may change if a section is added or removed. So I figured it would be safer to avoid the numbering.

Issue #594, opened 2019-02-25
Support shared home-manager configurations between all users


vlaci:
https://github.com/rycee/home-manager/issues/549 is broke my configuration. I have created a [NixOS module](https://github.com/vlaci/blox/blob/master/modules/nixos/home-manager.nix#L64) where I can declare a "shared" home manager configuration which is applied to all users managed by nix. I use this approach to generate homogeneous user environments. 

The above mentioned change introduces infinite recursion because now home-manager writes the `users.users` attribute.

I can work around this problem but I thought that it worth to mention this use case [1]

---
[1] Upon writing the issue description I remembered this xkcd comic https://xkcd.com/1172/

kalbasit:
I use something very similar to you, by assigning one home-manager config to multiple users, and this has not broken my setup. Perhaps it's something specific to you? Can you point out how it's setup on your side?

Here's how it's setup on my side:
- I added a NixOS option called [`mine.home-manager.config`](https://github.com/kalbasit/shabka/blob/c20d47b73784b1a5a5495f42e7c09a40dc7e51b7/modules/nixos/general/home-manager.nix#L21-L26)
- [NixOS configuration](https://github.com/kalbasit/shabka/blob/c20d47b73784b1a5a5495f42e7c09a40dc7e51b7/hosts/hades/configuration.nix#L44) imports [home.nix that sets the above option](https://github.com/kalbasit/shabka/blob/c20d47b73784b1a5a5495f42e7c09a40dc7e51b7/hosts/hades/home.nix)
- [home-manager.users](https://github.com/kalbasit/shabka/blob/c20d47b73784b1a5a5495f42e7c09a40dc7e51b7/modules/nixos/general/users.nix#L83) gets set by [calling the above config option](https://github.com/kalbasit/shabka/blob/c20d47b73784b1a5a5495f42e7c09a40dc7e51b7/modules/nixos/general/users.nix#L35-L38).

> [1] Upon writing the issue description I remembered this xkcd comic https://xkcd.com/1172/

I love it :rofl:!

vlaci:
Your configuration gave me the original inspiration for my version. The difference is that [I use `users.users`](https://github.com/vlaci/blox/blob/master/modules/nixos/home-manager.nix#L64) directly instead of a separate users attribute set hence the recursion.

kalbasit:
@vlaci oh I see, makes sense. @rycee thoughts on how to avoid the infinite loop here?

rycee:
[...]

PR #4368, opened 2023-08-21
nixos module: Fix regression where `users.users` is referred to even if not needed.

JustTNE:
### Description

GitHub issue #594 was previously fixed, but it seems a rewrite of the common.nix file seems to have reintroduced this as a regression. I have fixed the regression in this pull request. Seems like the tests system does not seem to test the nixos module, so I didn't add any regression tests so this doesn't happen again.

### Checklist

-> not applicable, untested component

-> not applicable, untested component

    ```
    {component}: {description}

    {long description}
    ```

-> not applicable, not a new module
#### Maintainer CC

@ncfavier @rycee

### DIFF
--- a/nixos/common.nix
+++ b/nixos/common.nix
@@ -94,20 +94,23 @@ in {
     };
   };
 
-  config = mkIf (cfg.users != { }) {
-    warnings = flatten (flip mapAttrsToList cfg.users (user: config:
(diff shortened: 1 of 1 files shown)


rycee:
Thanks! Rebased to master in ca4126e3c568be23a0981c4d69aed078486c5fce.

Issue #4367, opened 2023-08-20
bug: home-manager requires IFD since #4281


R-VdP:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Since #4281 HM does IFD which breaks nix evaluation on certain CI systems.

The IFD happens [here](https://github.com/nix-community/home-manager/blob/958c06303f43cf0625694326b7f7e5475b1a2d5c/modules/config/home-cursor.nix#L71) due to the call to `builtins.readDir`.

Output of `nixos-rebuild`:

```
➜ nixos-rebuild --verbose --flake '.' build --option allow-import-from-derivation false --option eval-cache false --show-trace
building the system configuration...
Building in flake mode.
$ nix --extra-experimental-features nix-command flakes build .#nixosConfigurations."starbook".config.system.build.toplevel --verbose --option allow-import-from-derivation false --option eval-cache false --show-trace
error:

<snip>
                                                                                                                                                                                                                        
       … from call site
                                                                                                                                                                                                                        
         at /nix/store/7qizhi4dhwji9vprhqvc2g95ibrhcxcs-source/modules/config/home-cursor.nix:72:6:
                                                                                                                                                                                                                        
           71|       (builtins.readDir "${cfg.package}/share/icons");
           72|   in mapAttrs'
             |      ^
           73|   (n: _: nameValuePair ".icons/${n}" "${cfg.package}/share/icons/${n}")
                                                                                                                                                                                                                        
       … while calling 'mapAttrs''
                                                                                                                                                                                                                        
         at /nix/store/s4jqyj35hii03rs7j5n6vn7gpgp6ja81-source/lib/attrsets.nix:522:5:
                                                                                                                                                                                                                        
          521|     # Attribute set to map over.
          522|     set:
             |     ^
          523|     listToAttrs (map (attr: f attr set.${attr}) (attrNames set));
                                                                                                                                                                                                                        
       … from call site
                                                                                                                                                                                                                        
         at /nix/store/7qizhi4dhwji9vprhqvc2g95ibrhcxcs-source/modules/config/home-cursor.nix:70:18:
                                                                                                                                                                                                                        
           69|   iconMappings = let
           70|     knownFiles = filterAttrs (_: type: type != "unknown")
             |                  ^
           71|       (builtins.readDir "${cfg.package}/share/icons");
                                                                                                                                                                                                                        
[...]

PR #4369, opened 2023-08-22
home-cursor: remove IFD when linking icon directories

polykernel:
### Description

An IFD was introduced in https://github.com/nix-community/home-manager/commit/bdb5bcad01ff7332fdcf4b128211e81905113f84, from reading the top-level directories of the
`home.pointerCursor.package` at instantiation time.

This commit removes the IFD introduced when linking icon directories by linking only the icon directory matching `home.pointerCursor.name` in `home.pointerCursor.package`. This should have the same effect as linking all top-level directories of the supplied icon package as the module generates cursor configurations pertaining to the cursor identified by `home.pointerCursor.name`. Since any additional icon directories present are not explicitly referenced by the module, any difference in behavior would be attributed to impurities and should be eliminated.

fixes https://github.com/nix-community/home-manager/issues/4367

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/config/home-cursor.nix
+++ b/modules/config/home-cursor.nix
@@ -66,13 +66,6 @@ let
     '';
   };
 
-  iconMappings = let
-    knownFiles = filterAttrs (_: type: type != "unknown")
-      (builtins.readDir "${cfg.package}/share/icons");
-  in mapAttrs'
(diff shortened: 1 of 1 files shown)


rycee:
Thanks for the quick fix! Merged to master now 🙂

PR #4375, opened 2023-08-24
khal: add options for default and view sections

arjan-s:
### Description

The current khal module doesn't provide options to add settings in the [default] and [view] sections of the configuration file. This PR adds optional settings for both. I chose a generic option instead of separately defining all possible settings, due to the amount of them and the future maintenance required for keeping the list up to date.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/khal.nix
+++ b/modules/programs/khal.nix
@@ -7,6 +7,8 @@ let
 
   cfg = config.programs.khal;
 
+  iniFormat = pkgs.formats.ini { };
+
   khalCalendarAccounts =
     filterAttrs (_: a: a.khal.enable) config.accounts.calendar.accounts;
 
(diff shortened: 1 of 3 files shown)


arjan-s:
I don't know how to fix the macos test, sorry

modules/programs/khal.nix
```diff
@@ -147,13 +147,41 @@ let
...
+
     locale = mkOption {
       type = lib.types.submodule { options = localeOptions; };
       description = ''
         khal locale settings. 
       '';
       default = { };
     };
+
+    defaultSection = mkOption {
```
rycee, line 159: Maybe just add a `settings`-option? Then people can do

```
settings = {
  default = {
    default_calendar = "calendar";
    timedelta = "5d";
  };
  view = {
    agenda_event_format = "{calendar-color}{cancelled}{start-end-time-style} {title}{repeat-symbol}{reset}";
  };
};
```

rycee:
Oh and about the macos test, it is unrelated to this PR and I think it should resolve itself on your next push 🙂

modules/programs/khal.nix
arjan-s: Sounds like a good plan, I didn't really like the two different-but-the-same sections anyway. And this approach also lets people configure other sections like [highlight_days] and [keybindings] too, so it's a double win.

arjan-s:
Thanks! What do you think about it now?

teto:
tested here and worked well

PR #4388, opened 2023-08-29
neomutt: fix STARTTLS

fpletz:
### Description

When smtps is used as a protocol, neomutt expects TLS but will fail if STARTTLS is configured to be used. When using STARTTLS, smtp has to be used as protocol and ssl_force_tls is set.

See https://neomutt.org/guide/optionalfeatures#2-1-%C2%A0starttls.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@ncfavier @sumnerevans @berbiche


fpletz:
Updated to fix the format fail yesterday since I added some more changes without checking formatting previously. Seems the CI is stuck?

fpletz:
Rebased. Ping.

rycee:

PR #4393, opened 2023-08-31
grobi: Fix configuration file to emit YAML as expected by the tool

DxCx:
### Description

When trying to work with 'grobi' i noted the service just fail and die,
when looking into it i've found out that the service expects YAML configuration but home manager converts it into JSON instead.

Looking at the project repo (https://github.com/fd0/grobi) shows there is only YAML support

Note regarding tests:
this is trying to drive by fix a problem found locally,
i looked into it and saw grobi today doesn't have any test case :\
looking at examples i see i can make a simple test, ok, but it doens't really trigger a verification on the tool just that "nix to yaml" conversion works,
is there a way to actually validate the emitted config?
Also, if im adding a new test for grobi, how do i trigger just it and not the whole test suite?

### Checklist

- [X] Code tested through `nix develop --ignore-environment .#all` using Flakes.

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@mbrgm

### DIFF
--- a/modules/services/grobi.nix
+++ b/modules/services/grobi.nix
@@ -94,9 +94,10 @@ in {
       Install = { WantedBy = [ "graphical-session.target" ]; };
     };
 
-    xdg.configFile."grobi.conf".text = builtins.toJSON {
-      execute_after = cfg.executeAfter;
-      rules = cfg.rules;
-    };
+    xdg.configFile."grobi.conf".source =
+      (pkgs.formats.yaml { }).generate "grobi-config" {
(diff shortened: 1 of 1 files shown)


ncfavier:
YAML is supposed to be a superset of JSON.

DxCx:
then i'll try again without this patch, originally i had parsing errors without this patch but im more than willing to retry and find out i can just close it PR =)

DxCx:
Yes, this seem to work, sorry for the confusion :\

PR #4395, opened 2023-08-31
starship: use `use` instead of `source` for nushell with starship v1.16+

davidkna:
### Description

For starship 1.16 it is recommended by upstream to use `use`. Earlier starship verions did not support the use of `use`, so `use` or `source` are used depending on the configure starship version.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/starship.nix
+++ b/modules/programs/starship.nix
@@ -137,7 +137,7 @@ in {
         ${starshipCmd} init nu | save --force ${config.xdg.cacheHome}/starship/init.nu
       '';
       extraConfig = ''
-        source ${config.xdg.cacheHome}/starship/init.nu
+        use ${config.xdg.cacheHome}/starship/init.nu
       '';
     };
   };


modules/programs/starship.nix
```diff
@@ -136,8 +136,13 @@ in {
...
       '';
-      extraConfig = ''
-        source ${config.xdg.cacheHome}/starship/init.nu
+      extraConfig = let
+        sourceCmd = if (versionAtLeast cfg.package.version "1.16.0") then
+          "use"
+        else
+          "source";
+      in ''
+        ${sourceCmd} ${config.xdg.cacheHome}/starship/init.nu
```
rycee, line 145: Generally speaking we avoid trying to do any package version checks. For the master branch it is always assumed that the user is using Nixpkgs unstable. So since Nixpkgs seems to have 1.16.0 I would suggest simply switching over to `use`, without any logic:

```suggestion
      extraConfig = ''
        use ${config.xdg.cacheHome}/starship/init.nu
```

rycee:

Issue #4397, opened 2023-08-31
bug: `xdg.*Home` set outside the home directory breaks


nicoonoclaste:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

`misc/xdg.nix` uses `home.file` to create the files defined in `xdg.{config, data}File.*`.
This implicitly assumes that `xdg.{config, data}Home` are relative to the homedir, which may not be the case.
This breaks everything relying on `xdg.dataFile`, including `misc/tmpfiles.nix` (_i.e._ `config.systemd.user.tmpfiles`).

The same issue also occurs with `xdg.cacheHome`, as `xdg.nix` attempts to create a file at `$XDG_CACHE_HOME/.keep`.

Here is a snippet from my configuration, that exhibits the issue:
```nix
xdg =
	let h = "/var/home/nicoo"; in {
		enable = true;
		cacheHome = "/var/cache/nicoo";
		dataHome = "${h}/xdg/data";
		stateHome = "${h}/xdg/state";
	};
```

And the result is that directory tree being created under `~/` instead:
```
$ tree -a ~/var
/home/nicoo/var
├── cache
│   └── nicoo
│       └── .keep -> /nix/store/kd75z93pgx2djiafv69azrs4vpkp39j9-home-manager-files/var/cache/nicoo/.keep
└── home
    └── nicoo
        └── xdg
            └── data
                └── user-tmpfiles.d
                    └── home-manager.conf -> /nix/store/kd75z93pgx2djiafv69azrs4vpkp39j9-home-manager-files/var/home/nicoo/xdg/data/user-tmpfiles.d/home-manager.conf
```

### Maintainer CC

There doesn't seem to be a documented maintainer for `xdg.nix`, nor a recurring committer to it.  :/
*tentatively pokes @ncfavier, as the last person to have specifically touched it.*

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.1.47, NixOS, 23.11 (Tapir), 23.11pre-git`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.17.0`
 - nixpkgs: `/run/current-system/config/3rd-party/nixpkgs`
```

cjbayliss:
[...]

PR #4398, opened 2023-08-31
tmpfiles: use only `xdg.configFile`

nbraud:
### Description

In `modules/misc/tmpfiles.nix`, replace the single use of `xdg.dataFile` with `xdg.configFile`, which was used anyways.

Not only is this a small (positive) refactor, but it minimises the impact of #4397, as setting `xdg.dataHome` outside the homedir will not break `tmpfiles` anymore.

### Checklist

- [x] Commit messages are formatted as per [CONTRIBUTING](https://github.com/nix-community/home-manager/blob/master/docs/contributing.adoc#sec-commit-style) for more information and [recent commit messages](https://github.com/nix-community/home-manager/commits/master) for examples.

#### Maintainer CC

@dawidsowa

### DIFF
--- a/modules/misc/tmpfiles.nix
+++ b/modules/misc/tmpfiles.nix
@@ -27,8 +27,8 @@ in {
         platforms.linux)
     ];
 
-    xdg = {
-      dataFile."user-tmpfiles.d/home-manager.conf" = {
+    xdg.configFile = {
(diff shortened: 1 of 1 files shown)


rycee:
Thanks! Merged to master now. @dawidsowa I think this is uncontroversial but if you dislike this change then let me know and we can try to resolve it.

PR #4401, opened 2023-09-02
nnn: Add extraSessionVariables and extraArgs

Anomalocaridid:
### Description

I have added two options to the nnn module, `extraSessionVariables` and `extraArgs`. They add environment variables and arguments, respectively, to nnn's wrapper.

My rationale for this change is that it simplifies calling nnn in contexts where environment variables and aliases set in one's shell may not necessarily be available, such as by executing it through keybindings set in window managers such as `hyprland`.

Currently, one would have to run nnn indirectly using `bash -c`, `zsh -c`, etc. in order to give it access to arguments set through aliases and environment variables not already set by Home Manager.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@thiagokokada

### DIFF
--- a/modules/programs/nnn.nix
+++ b/modules/programs/nnn.nix
@@ -5,10 +5,16 @@ with lib;
 let
   cfg = config.programs.nnn;
 
-  renderSetting = key: value: "${key}:${value}";
+  renderAttrs = sep: renderer: attrs:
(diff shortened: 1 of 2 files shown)


Anomalocaridid:
Closing because this has been stale for a while and I have long since stopped using `nnn` anyways.

PR #4404, opened 2023-09-03
vdirsyncer: add urlCommand and userNameCommand options

### LINKED ISSUE #4399: vdirsyncer: secret url
### Description

According to the [manual](https://vdirsyncer.pimutils.org/en/stable/keyring.html) any kind of parameter in a storage section can be combined with `.command`. I'd like to combine it with `url` option to not have plaintext url in my public config repo nor in nix store. I'd like to use it with outlook, but as I know it's a common practice for calendar providers to have a possibility to generate a secret url to access a personal calendar. This will require to add `urlCommand` into [accounts](https://github.com/nix-community/home-manager/blob/f5c15668f9842dd4d5430787d6aa8a28a07f7c10/modules/accounts/calendar.nix#L48). As I see there is a [commented userNameCommand](https://github.com/nix-community/home-manager/blob/f5c15668f9842dd4d5430787d6aa8a28a07f7c10/modules/accounts/calendar.nix#L60) which was disabled in the [original PR](https://github.com/nix-community/home-manager/pull/4078). What was the reason to disable it? cc @teto

kurnevsky:
### Description

Fixes #4399 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@pjones @teto

### DIFF
--- a/modules/accounts/calendar.nix
+++ b/modules/accounts/calendar.nix
@@ -58,15 +58,6 @@ let
         description = "User name for authentication.";
       };
 
-      # userNameCommand = mkOption {
-      #   type = types.nullOr (types.listOf types.str);
-      #   default = null;
-      #   example = [ "~/get-username.sh" ];
-      #   description = ''
-      #     A command that prints the user name to standard output.
-      #   '';
-      # };
-
       passwordCommand = mkOption {
         type = types.nullOr (types.listOf types.str);
         default = null;
--- a/modules/programs/vdirsyncer.nix
+++ b/modules/programs/vdirsyncer.nix
@@ -32,27 +32,25 @@ let
     });
 
   remoteStorage = a:
-    filterAttrs (_: v: v != null) ((getAttrs [
-      "type"
-      "url"
-      "userName"
-      #"userNameCommand"
-      "passwordCommand"
-    ] a.remote) // (if a.vdirsyncer == null then
-      { }
-    else
-      getAttrs [
-        "itemTypes"
-        "verify"
-        "verifyFingerprint"
-        "auth"
-        "authCert"
-        "userAgent"
-        "tokenFile"
-        "clientIdCommand"
-        "clientSecretCommand"
-        "timeRange"
-      ] a.vdirsyncer));
+    filterAttrs (_: v: v != null)
+    ((getAttrs [ "type" "url" "userName" "passwordCommand" ] a.remote)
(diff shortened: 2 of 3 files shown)


modules/accounts/calendar.nix
```diff
@@ -51,20 +51,27 @@ let
...
 
-      # userNameCommand = mkOption {
-      #   type = types.nullOr (types.listOf types.str);
-      #   default = null;
-      #   example = [ "~/get-username.sh" ];
-      #   description = ''
-      #     A command that prints the user name to standard output.
-      #   '';
-      # };
+      userNameCommand = mkOption {
```
rycee, line 67: I guess `userName` and `userNameCommand` are mutually exclusive. I.e., we should have an assertion that causes an error if the user has given both. Same holds for `url` and `urlCommand`.

modules/programs/vdirsyncer.nix
```diff
@@ -24,8 +24,7 @@ let
 
   localStorage = a:
     filterAttrs (_: v: v != null)
-    ((getAttrs [ "type" "fileExt" "encoding" ] a.local) // {
-      path = a.local.path;
+    ((getAttrs [ "type" "fileExt" "encoding" "path" ] a.local) // {
```
rycee, line 27: This seems unrelated to the urlCommand and userNameCommand support? If so then I would suggest putting this change in a separate PR.

rycee:
I think this change is fair enough but I'm concerned that urlCommand and userNameCommand are pretty specific for vdirsyncer. That is, if no other software support reading URL and username from commands then perhaps it would be better to put these options as vdirsyncer specific options?

modules/programs/vdirsyncer.nix
kurnevsky: It is, just noticed that it was unnecessary to set it explicitly. I wouldn't bother to create a separate PR for this, so I guess it will remain as is forever :)

modules/accounts/calendar.nix
kurnevsky: Added.

kurnevsky:
Actually what other software that supports remote accounts? If I don't miss something it's only vdirsyncer :)

kurnevsky:
Still relevant.

kurnevsky:
Why is it stale?

rycee:
Sorry about the delay!

PR #4412, opened 2023-09-07
eza: init

cafkafk:
This copies the exa configuration for eza, the official fork of the
program.

Signed-off-by: Christina Sørensen <christina@cafkafk.com>

Refs: https://github.com/eza-community/eza/issues/207, https://github.com/ogham/exa/issues/1243

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/eza.nix
+++ b/modules/programs/eza.nix
@@ -1,0 +1,67 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+{
+  meta.maintainers = [ maintainers.cafkafk ];
+
+  options.programs.eza = {
+    enable = mkEnableOption "eza, a modern replacement for {command}`ls`";
+
+    enableAliases = mkEnableOption "recommended eza aliases (ls, ll…)";
+
+    extraOptions = mkOption {
+      type = types.listOf types.str;
+      default = [ ];
+      example = [ "--group-directories-first" "--header" ];
+      description = ''
+        Extra command line options passed to eza.
+      '';
+    };
+
+    icons = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Display icons next to file names ({option}`--icons` argument).
+      '';
+    };
+
+    git = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        List each file's Git status if tracked or ignored ({option}`--git` argument).
+      '';
+    };
+
+    package = mkPackageOption pkgs "eza" { };
+  };
+
+  config = let
+    cfg = config.programs.eza;
+
+    args = escapeShellArgs (optional cfg.icons "--icons"
+      ++ optional cfg.git "--git" ++ cfg.extraOptions);
+
+    aliases = {
+      eza = "eza ${args}";
+    } // optionalAttrs cfg.enableAliases {
+      ls = "eza";
+      ll = "eza -l";
+      la = "eza -a";
+      lt = "eza --tree";
+      lla = "eza -la";
+    };
+  in mkIf cfg.enable {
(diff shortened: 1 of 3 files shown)


cafkafk:
I've never contributed code to home manager before, so I'm sorry if there are any obvious mistakes.

modules/programs/eza.nix
```diff
@@ -0,0 +1,67 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+{
+  meta.maintainers = [ maintainers.cafkafk ];
```
ryanccn, line 6: I don't think `maintainers` is in scope here (also doesn't seem like you added yourself to `modules/lib/maintainers.nix`

ryanccn (changes requested):

ryanccn:
Also I believe you have to add the module to `modules/modules.nix` for it to be included

cafkafk:
> Also I believe you have to add the module to `modules/modules.nix` for it to be included

This push addresses this, thanks!

ryanccn:
I ran into this same issue when copying the exa configuration and replacing `exa` with `eza`, lol

modules/programs/eza.nix
cafkafk: > I don't think maintainers is in scope here

Sorry, I think I was too fast, changed it to `lib.pkgs.maintainers`. 

> also doesn't seem like you added yourself to modules/lib/maintainers.nix

The docs said you didn't have to add yourself if you were in the nixpkgs maintainers file. If I can't use nixpkgs I'll add myself to the home manager file.

cafkafk:
> I ran into this same issue (`ezample`) when copying the exa configuration and replacing `exa` with `eza`, lol

:woman_facepalming: I thought I had fixed that before pushing.

modules/programs/eza.nix
```diff
@@ -0,0 +1,67 @@
...
+
+  options.programs.eza = {
+    enable = mkEnableOption "eza, a modern replacement for {command}`ls`";
+
+    enableAliases = mkEnableOption "recommended eza aliases (ls, ll…)";
+
+    extraOptions = mkOption {
+      type = types.listOf types.str;
+      default = [ ];
+      ezample = [ "--group-directories-first" "--header" ];
```
ryanccn, line 16:
```suggestion
      example = [ "--group-directories-first" "--header" ];
```

cafkafk:
Okay fixed it.

modules/programs/eza.nix
```diff
@@ -0,0 +1,67 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+{
+  meta.maintainers = [ pkgs.lib.maintainers.cafkafk ];
```
isabelroses, line 6: as you have put `with lib;` the expression can be shortened to:
```suggestion
  meta.maintainers = [ pkgs.lib.maintainers.cafkafk ];
  meta.maintainers = [ maintainers.cafkafk ];
```

rycee:

rycee:
Btw, I notice that the exa package has been [removed from Nixpkgs](https://github.com/NixOS/nixpkgs/pull/253683). We may therefore want to remove the `programs.exa` module as well?

Doing something like https://github.com/nix-community/home-manager/blob/91341cde4143b10ee66e994a53c35d376ad6cdfb/modules/programs/just.nix should be fine. The error message can recommend using `programs.eza` instead. Would any of you be willing to put together a PR for that?

cafkafk:
> Would any of you be willing to put together a PR for that?

I can give it a try. I'll look at it later today probably.

Issue #4411, opened 2023-09-07
Document the requirement for adding swaylock to PAM


TLATER:
### Description

Super tiny documentation issue, but I just scratched my head over this and intend to add a note about this myself when I have a moment.

[swaylock](https://nix-community.github.io/home-manager/options.html#opt-programs.swaylock.enable) will only be able to unlock sessions if the system PAM is configured to allow it.

On NixOS this is done by default if the sway module is enabled, [but not for other wlroots compositors](https://github.com/NixOS/nixpkgs/issues/143365). On non-NixOS the option will presumably be even less likely to work.

Obviously home-manager can't do anything about the system configuration, but there should be a note in the documentation pointing this out.

TLATER:
Oh, I forgot I actually already did this in #4416.

Thanks stale bot.

PR #4416, opened 2023-09-07
swaylock: Document the need for host PAM configuration

TLATER:
### Description

This adds a note about how swaylock can actually be made to unlock sessions when swaylock-specific PAM configuration is not yet set up.

The note on how to do it on NixOS seems like a useful hint, it's probably the most common use case where this will fail, but shout if NixOS-specific notes are not right here.

### Checklist

  - [x] This is a documentation change; tested with `nix build .#html-docs` instead

  - N/A

    ```
    {component}: {description}

    {long description}
    ```

    - N/A

#### Maintainer CC

@rcerc

### DIFF
--- a/modules/programs/swaylock.nix
+++ b/modules/programs/swaylock.nix
@@ -16,7 +16,21 @@ in {
         false otherwise
       '';
       example = true;
-      description = "Whether to enable swaylock.";
+      description = ''
+        Whether to enable swaylock.
+
+        Note that PAM must be configured to enable swaylock to perform
+        authentication. The package installed through home-manager
+        will *not* be able to unlock the session without this
+        configuration.
+
+        On NixOS, this is by default enabled with the sway module, but
+        for other compositors it can currently be enabled using:
+
+        ```nix
+        security.pam.services.swaylock = {};
+        ```
+      '';
     };
 
     package = mkPackageOption pkgs "swaylock" { };


TLATER:
Resolves #4411, sorry, needed a clear reminder to do this since I didn't want to keep going at 3 am ;)

modules/programs/swaylock.nix
```diff
@@ -16,7 +16,21 @@ in {
...
+        Note that PAM must be configured to enable swaylock to perform
+        authentication. The package installed through home-manager
+        will *not* be able to unlock the session without this
+        configuration.
+
+        On NixOS, this is by default enabled with the sway module, but
+        for other compositors it can currently be enabled using:
+
+        ```
+        security.pam.services.swaylock = {};
```
rcerc, line 31: Thanks! In reply to your comment on #4411 that Home Manager cannot set this option: in Home Manager's [`nixos/default.nix`](https://github.com/nix-community/home-manager/blob/19c6a4081b14443420358262f8416149bd79561a/nixos/default.nix#L21) file, the `sharedModules` option does actually configure the system-wide NixOS options in the case that the Home Manager NixOS module is used. Do you think it would be a good idea to add `security.pam.services.swaylock = {};` there?

rcerc (approved):
Thank you for adding this note! Please disregard my irrelevant comment...

modules/programs/swaylock.nix
TLATER: No worries! home-manager + NixOS integration like this would actually be interesting when/if they are ever merged into nixpkgs. But yeah, for now I don't think it's actually possible.

rcerc: As a NixOS module, HM _could_ do it, but seeing now that it does not tweak the system configuration other than to set up a service for each user, I wonder if that's on purpose to avoid surprising the user...

TLATER: I doubt edge cases like this one were ever actually considered, but yeah, I think home-manager tries to stick to user configuration as much as possible intentionally. The NixOS module is just so you can easily match nixpkgs versions and don't need to run two commands when rebuilding your system.

If home-manager were merged into nixpkgs, it might be interesting to have tighter coupling. It could lead to worse behavior on non-NixOS systems over time though, as these kinds of things would be tested less, so maybe it wouldn't be the best idea anyway. Either way, food for a later discussion ;)

rycee:

leiserfg:
I'm using home-manager in another distro. How can I add this to pam? I tried adding `/etc/pam.d/swaylock` with 
`auth include login` as content and it didn't work and I have no idea how to debug it.

rcerc:
@leiserfg, what are the contents of `/etc/pam.d/login`?

leiserfg:
I set them to `auth include login`. I have tried with other lockers and all have the same problem, so seems like an issue with libpam from not wanting to talk to system's PAM  but I have no idea how to debug why.

rcerc:
I am more curious about what is in `/etc/pam.d/login` (rather than `/etc/pam.d/swaylock`), especially because other lockers do not work.

leiserfg:
I partially found why it does not work, and fixing it is possible but very hard. libpam works by reading the content of /etc/pam.d/<service_name> and "executing" it then:

1. The syntax of pam configs in debian has extra directives they patched to libpam
2. Some of the modules of libpam don't have the same name / are not present in libpam from nix by default.
3. some of the ones that do exist don't work because of been in a foreign environment somehow.

rcerc:
Is it not possible to install Debian's `swaylock` package and then in Home Manager put this?
```nix
programs.swaylock.package = pkgs.runCommandLocal "empty" { } "mkdir $out";
```

leiserfg:
Yes it is, I just wanted to reduce non nix packages but I won't be able to.

PR #4422, opened 2023-09-08
home-manager: fix nix profile detection when use-xdg-base-directories == true

e-nikolov:
If nix.settings.use-xdg-base-directories is true, the ~/.nix-profile directory may not exist, and home.profileDirectory should be checked intead

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/home-manager/home-manager
+++ b/home-manager/home-manager
@@ -29,7 +29,9 @@ function removeByName() {
 }
 
 function setNixProfileCommands() {
-    if [[ -e $HOME/.nix-profile/manifest.json ]] ; then
+    if  [[ -e $HOME/.nix-profile/manifest.json \
+        || -e ${XDG_STATE_HOME:-$HOME/.local/state}/nix/profile/manifest.json ]] ; then
+
         LIST_OUTPATH_CMD="nix profile list"
         REMOVE_CMD="removeByName"
     else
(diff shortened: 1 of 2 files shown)


home-manager/home-manager
```diff
@@ -29,7 +29,10 @@ function removeByName() {
 }
 
 function setNixProfileCommands() {
-    if [[ -e $HOME/.nix-profile/manifest.json ]] ; then
+    if  [[ -e $HOME/.nix-profile/manifest.json 
+        || -e $XDG_STATE_HOME/nix/profile/manifest.json 
+        || -e $HOME/.local/state/nix/profile/manifest.json ]] ; then
```
ambroisie, line 34: This is more correct according to [the spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html):

```suggestion
        || -e ${XDG_STATE_HOME:-$HOME/.local/state}/nix/profile/manifest.json ]] ; then
```

e-nikolov:
@rycee Would you mind reviewing this?

ambroisie (approved):

Arian04:
is there a reason that either this or #4594 haven't been merged? it's not a big deal but it'd be nice if I could remove the empty .nix-profile directory in my home dir

rycee:
Sorry about the delay, I've merged this to master now!

PR #4447, opened 2023-09-12
bat: allow adding custom syntaxes

stuebinm:
adds an option to add custom syntax files to bat's cache, analogous to the existing programs.bat.themes option.

### Checklist

  - [x] Code tested through `nix develop --ignore-environment .#bat`; I also tried the above, but it fails in some emacs test (which also fails on current master for me, i.e. 5171f5ef654425e09d9c2100f856d887da595437, so I've ignored it)
  - (also currently in use in my own home config)

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

maintainer list is empty; @rycee made the last commit touching the module

### DIFF
--- a/modules/programs/bat.nix
+++ b/modules/programs/bat.nix
@@ -60,6 +60,24 @@ in {
       '';
     };
 
+    syntaxes = mkOption {
+      type = types.attrsOf types.lines;
+      default = { };
+      example = literalExpression ''
+        {
+          syntaxes.gleam = builtins.readFile (pkgs.fetchFromGitHub {
+            owner = "molnarmark";
+            repo = "sublime-gleam";
+            rev = "2e761cdb1a87539d827987f997a20a35efd68aa9";
+            hash = "sha256-Zj2DKTcO1t9g18qsNKtpHKElbRSc9nBRE2QBzRn9+qs=";
+          } + "/syntax/gleam.sublime-syntax");
+        }
+      '';
+      description = ''
+        Additional syntaxes to provide.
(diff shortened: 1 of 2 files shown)


modules/programs/bat.nix
```diff
@@ -60,6 +60,24 @@ in {
...
+    syntaxes = mkOption {
+      type = types.attrsOf types.lines;
+      default = { };
+      example = literalExpression ''
+        {
+          syntaxes.gleam = builtins.readFile (pkgs.fetchFromGitHub {
+            owner = "molnarmark";
+            repo = "sublime-gleam";
+            rev = "2e761cdb1a87539d827987f997a20a35efd68aa9";
+            sha256 = "sha256-Zj2DKTcO1t9g18qsNKtpHKElbRSc9nBRE2QBzRn9+qs=";
```
Janik-Haag, line 72:
```nix
            hash = "sha256-Zj2DKTcO1t9g18qsNKtpHKElbRSc9nBRE2QBzRn9+qs=";
```
this is the recommended option in nixpkgs (but just a small nitpick)

Janik-Haag (commented):
this looks good to me except for the small nitpick

stuebinm:
I went for consistency with the example in `programs.bat.themes`, but
could also just change both i guess 🤷

modules/programs/bat.nix
stuebinm: have changed this

rycee:

PR #4456, opened 2023-09-14
home-manager: fix for missing profile directory in nix v2.17.0

### LINKED ISSUE #4403: bug: standalone flake installation fails on new nix installations

### Issue description

I'm not sure if this should be addressed in home-manager or somewhere else, but with a brand new multi-user nix installation, running this home-manager installation command:

```sh
nix run --extra-experimental-features "flakes nix-command" home-manager/master -- init --switch ~/nix-config
```

Results in this error:

```
Could not find suitable profile directory, tried /home/test/.local/state/home-manager/profiles and /nix/var/nix/profiles/per-user/test
```

It looks like `.nix-profile` does not yet exist in brand new nix installations, but is created the first time `nix profile list` or `nix profile install` is executed. I tried both the default nix-installer and https://github.com/DeterminateSystems/nix-installer:

- `sh <(curl -L https://nixos.org/nix/install) --daemon`
- `curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix | sh -s -- install`

### Maintainer CC

@rycee

### System information

```markdown
- system: `"x86_64-linux"`
- host os: `Linux 5.15.90.1-microsoft-standard-WSL2, Ubuntu, 22.04.2 LTS (Jammy Jellyfish), nobuild`
- multi-user?: `yes`
- sandbox: `yes`
- channels(root): `"nixpkgs"`
```

e-nikolov:
### Description

Fixes #4403.

Due to a possible bug in nix v2.15.0 - v2.17.0, `nix run` no longer creates the default profile, which makes the flake installation of home-manager fail on new nix installations. This PR makes sure that `nix-env -q` is executed in order to create the profile.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/home-manager/home-manager
+++ b/home-manager/home-manager
@@ -648,6 +648,8 @@ function doSwitch() {
 }
 
 function doListGens() {
+    setHomeManagerPathVariables
+
     # Whether to colorize the generations output.
     local color="never"
     if [[ ! -v NO_COLOR && -t 1 ]]; then
@@ -665,6 +667,7 @@ function doListGens() {
 # Removes linked generations. Takes as arguments identifiers of
 # generations to remove.
 function doRmGenerations() {
+    setHomeManagerPathVariables
     setVerboseAndDryRun
 
     pushd "$HM_PROFILE_DIR" > /dev/null
@@ -686,6 +689,8 @@ function doRmGenerations() {
 }
 
 function doExpireGenerations() {
+    setHomeManagerPathVariables
+
     local generations
     generations="$( \
         find "$HM_PROFILE_DIR" -name 'home-manager-*-link' -not -newermt "$1" \
@@ -810,6 +815,7 @@ function doShowNews() {
 }
 
 function doUninstall() {
+    setHomeManagerPathVariables
     setVerboseAndDryRun
     setNixProfileCommands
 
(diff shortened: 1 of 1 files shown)


e-nikolov:
I noticed that the error happens for:
- `home-manager switch`
- `home-manager init --switch

inside the [setHomeManagerPathVariables()](https://github.com/nix-community/home-manager/blob/d9b88b43524db1591fb3d9410a21428198d75d49/home-manager/home-manager#L157) function, which is currently called by all home-manager commands, but the variables it sets are only used inside:

- `home-manager generations`
- `home-manager remove-generations`
- `home-manager expire-generations`
- `home-manager uninstall`

Calling `setHomeManagerPathVariables` for `switch` and `init --switch` seems unnecessary, so an alternative solution that might be cleaner is to only set those variables for the commands that use them.

e-nikolov:
@rycee Is there anything I should do to get this merged?

rycee:
Sorry for the delay. Merged now with a slight change, I just moved the function call into the corresponding Bash functions.

Much appreciated!

e-nikolov:
Thanks!

PR #4458, opened 2023-09-14
neovim: expose finalPackage

KFearsoff:
### Description

Currently, there are 5 packages that can be managed with Home Manager that expose `finalPackage` option publicly: `pandoc`, `rofi`, `timidity`, `emacs`, `hyprland`. This commit adds the 6th one: `neovim`. While the final package that is being built after all of the modifications should be considered an implementation detail, in cases of complex software like `neovim` this could hide important details.

One particular case that comes to mind is using `lazy.nvim` package manager. It exposes a number of useful APIs that really help managing configuration, such as conditional keybindings, dependency tree building, merging configurations, and so on. Questions about using it on NixOS come up often, because manually recreating this functionality is hard, and it's a very popular method of managing plugins that has good UI/UX.

As it is now, it is impossible to use `lazy.nvim` with Nix efficiently. `lazy.nvim` monopolizes the plugin management, which means that installing plugins with Nix and managing them with `lazy.nvim` is not supported. Some of the pitfalls are already described: https://github.com/folke/lazy.nvim/issues/402
https://github.com/folke/lazy.nvim/issues/66

In short, to integrate `lazy.nvim` with Nix there are three things that must be done:
1. `lazy.nvim` resets packpath, which should be managed by Nix; it has to be turned off with `performance.reset_packpath = false` option in `lazy.nvim`
2. Same thing applies to rtp, the relevant option is `performance.rtp.reset = false`
3. `dev.path` must be specified as the folder that contains the plugins. This folder is the path to Nix store path with Neovim plugins, and the plugins that are managed by Nix must be marked as `dev = true` 
 
The third condition can not be fulfilled without this PR, as the final package that Home Manager produces is not exposed, therefore it is impossible to extract the Nix store path. This PR makes it possible to be extracted via `${pkgs.vimUtils.packDir
config.home-manager.users.<user>.programs.neovim.finalPackage.passthru.packpathDirs}/pack/myNeovimPackages/start}` expression.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@emilazy

### DIFF
--- a/modules/programs/neovim.nix
+++ b/modules/programs/neovim.nix
@@ -237,7 +237,6 @@ in {
 
       finalPackage = mkOption {
         type = types.package;
-        visible = false;
         readOnly = true;
         description = "Resulting customized neovim package.";
       };


rycee:
The option only has `visible = false`, which just controls whether it is shown in the manual or not. You should already be able to access its value in your configuration.

That said, I don't have anything in particular against making it visible in the documentation.

KFearsoff:
Oh, you are completely right. I just checked it and it works without this PR, lol. I had some issue trying to use `finalPackage` before without this PR, but I guess I might have made some changes that fixed it along the way.

Well, I think it's still better to make it visible, at least for the consistency's sake.

teto:
I dont mind either way.

rycee:
Cool, then let's merge. Thanks all! 🙂

PR #4459, opened 2023-09-14
keychain: improve nushell integration

### LINKED ISSUE #4255: bug: `keychain` `enableNushellIntegration` will break in version 0.82.1 of `nu`

### Issue description

This command fails with version 0.82.1 of nushell as `load-env` will refuse to take anything but a table 
https://github.com/nix-community/home-manager/blob/0841242b94638fcd010f7f64e56b7b1cad50c697/modules/programs/keychain.nix#L119C1-L121C8

### Maintainer CC

@CardboardTurkey 

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.4.2-zen1, NixOS, 23.11 (Tapir), 23.11.20230716.6cee3b5`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix Super) 2.17.0pre20230710_6ff67e4`
```
```

DrRuhe:
### Description
When the pipeline results in an empty list, load-env throws an error. So I added a check.

- If the parse does not match/returns an empty list, `transpose -ird` does not transform it into a record and thus load-env fails. I added an empty check to catch this.
- The parsing assumes keychain to emit bash-style completions, so set SHELL=bash to ensure keychain output is in the expected format.

I think this should also fix #4255 .
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@CardboardTurkey 
@Infinidoge

### DIFF
--- a/modules/programs/keychain.nix
+++ b/modules/programs/keychain.nix
@@ -117,7 +117,10 @@ in {
       eval "$(SHELL=zsh ${shellCommand})"
     '';
     programs.nushell.extraConfig = mkIf cfg.enableNushellIntegration ''
-      ${shellCommand} | parse -r '(\w+)=(.*); export \1' | transpose -ird | load-env
+      let keychain_shell_command = (SHELL=bash ${shellCommand}| parse -r '(\w+)=(.*); export \1' | transpose -ird)
+      if not ($keychain_shell_command|is-empty) {
+        $keychain_shell_command | load-env
+      }
     '';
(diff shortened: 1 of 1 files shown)


Infinidoge:
I don't believe I am a maintainer, but I can take a look later.

DrRuhe:
> I don't believe I am a maintainer, but I can take a look later.

Yeah, the maintainer list is empty and your name came up in git blame :)

DrRuhe:
Thanks :)

PR #4465, opened 2023-09-16
zellij: Add extraConfig option 

VoreckLukas:
### Description

This just implements an extraConfig option for zellij
With the current module its hard to set keybinds (at least i couldnt find a way) and this acts as a bridge until a more idiomatic way has been implemented

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@mainrs

### DIFF
--- a/modules/programs/zellij.nix
+++ b/modules/programs/zellij.nix
@@ -51,6 +51,39 @@ in {
     enableFishIntegration = mkEnableOption "Fish integration" // {
       default = false;
     };
+
+    extraConfig = mkOption {
+      type = lib.types.lines;
+      default = "";
+      example = ''
+        keybinds {
+            // keybinds are divided into modes
+            normal {
+                // bind instructions can include one or more keys (both keys will be bound separately)
+                // bind keys can include one or more actions (all actions will be performed with no sequential guarantees)
+                bind "Ctrl g" { SwitchToMode "locked"; }
+                bind "Ctrl p" { SwitchToMode "pane"; }
+                bind "Alt n" { NewPane; }
+                bind "Alt h" "Alt Left" { MoveFocusOrTab "Left"; }
+            }
+            pane {
+                bind "h" "Left" { MoveFocus "Left"; }
+                bind "l" "Right" { MoveFocus "Right"; }
+                bind "j" "Down" { MoveFocus "Down"; }
+                bind "k" "Up" { MoveFocus "Up"; }
+                bind "p" { SwitchFocus; }
+            }
+            locked {
+                bind "Ctrl g" { SwitchToMode "normal"; }
+            }
+        }
+      '';
+      description = ''
+        Extra configuration lines
+        Versions prior to 0.32.0 use Yaml
+        Versions after 0.32.0 use KDL
+      '';
+    };
   };
 
   config = mkIf cfg.enable {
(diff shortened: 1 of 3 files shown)


rycee:
Unfortunately I don't know much about zellij but perhaps it is sufficient with `settings`? For example, the (admittedly ugly)

```nix
programs.zellij.settings = {
  keybinds = {
    normal = {
      "bind \"Alt n\"" = {
        NewPane = [];
      };
      "bind \"Alt h\" \"Alt Left\"" = {
        MoveFocusOrTab = "Left";
      };
    };
  };
};
```

produces the output

``` kdl
keybinds {
	normal {
		bind "Alt h" "Alt Left" {
			MoveFocusOrTab "Left"
		}
		bind "Alt n" {
			NewPane 
		}
	}
}
```

Perhaps there are issues with this, for example the bind order will be alphabetic, same with the actions inside the bind.

VoreckLukas:
> Unfortunately I don't know much about zellij but perhaps it is sufficient with `settings`? For example, the (admittedly ugly)
> 
> ```nix
> programs.zellij.settings = {
>   keybinds = {
>     normal = {
>       "bind \"Alt n\"" = {
>         NewPane = [];
>       };
>       "bind \"Alt h\" \"Alt Left\"" = {
>         MoveFocusOrTab = "Left";
>       };
>     };
>   };
> };
> ```
> 
> produces the output
> 
> ```
> keybinds {
> 	normal {
> 		bind "Alt h" "Alt Left" {
> 			MoveFocusOrTab "Left"
> 		}
> 		bind "Alt n" {
> 			NewPane 
> 		}
> 	}
> }
> ```
> 
> Perhaps there are issues with this, for example the bind order will be alphabetic, same with the actions inside the bind.

I didn't even think about that. probably the syntax looks really weird. Seems like extraConfig won't be needed then but a better syntax for specifying keybinds would still be nice and is something ill look into

Nyabinary:
Didn't extraConfig get replaced with .settings in nixpkgs in general?

rycee:
@Nyabinary Yes, when possible we prefer a structural "settings" option. It's not always possible, though. For example, if the configuration is made in a general purpose language or something similar.

@VoreckLukas I happened to think that perhaps it works with single quotes? I.e.

``` nix
programs.zellij.settings = {
  keybinds = {
    normal = {
      "bind 'Alt n'" = {
        NewPane = [];
      };
      "bind 'Alt h' 'Alt Left'" = {
        MoveFocusOrTab = "Left";
      };
    };
  };
};
```

Without the escaping it looks a bit better, I think.

cgahr:
@rycee Afaik, single quotes do not work! But it does look nicer..

Also, sorting the keys introduces some problems:
```nix
settings = {
  keybinds = {
    unbind = "Ctrl g";
    shared = {
      "bind \"Ctrl g\"" = { SwitchToMode = "tmux"; };
      "bind \"Esc\" \"q\"" = { "ScrollToBottom; SwitchToMode" = "Normal"; };
    };
  };
}
```

The generated `config.kdl` looks as follows:
```kdl
keybinds {
  shared {
    bind "Ctrl g" {
      SwitchToMode "tmux"
    }
    bind "Esc" "q" {
      ScrollToBottom; SwitchToMode "Normal"
    }
  }
  unbind "Ctrl g"
}
```
As you can see, the `unbind` command is at the bottom and unbinds the just set command. This basically makes using `unbind` impossible. Or did I miss something?

gotcha:
@VoreckLukas 

I have added documentation for the YAML to KDL generator: https://github.com/nix-community/home-manager/pull/4665

Feedback welcome !

JonathanLorimer:
@rycee is there any way we could revisit this PR. It would be super helpful (and it doesn't preclude a `.settings` attr, it just allows a backdoor).

KamWithK:
In the latest Zellij there is a new default mode setting non-colliding which works on all the other OS' by generating a new default config. I tried generating something similar via Nix but in the end it was much simpler to just source that default generated file because of the same syntax problems as mentioned here (once you get a few layer deep into the KDL doesn't seem like it's possible anymore).
If something like alternative defaults can't be generated via Nix, I think it would be very nice to allow extraConfig like this.

exzombie:
Not stale. I've looked at all the open issues and I can't see a good way to import the non-colliding key bindings. Sure, we can use the `home.file` mechanism to install the config file, but then, what's the point of having `programs.zellij.settings` anyway?

jmacdonald:
I understand the desire to use structured settings wherever possible, but mapping to KDL feels incredibly clumsy at the moment, particularly for nested keybinds. Given that this doesn't deprecate the settings option, you can still use both, so simple configurations can leverage the structured format, and more advanced ones can simply inject a string.

PR #4473, opened 2023-09-18
pasystray: add extraOptions

motiejus:
e.g. for `-g`.

### Description

Backport of #4471, let's have the discussion there.

cc @pltanton


rycee:
The backport should be a cherry-pick from master so we cannot have an open PR for it already. Will close for now.

PR #4476, opened 2023-09-19
programs.vscode: add option to disable auto updates for extensions

Zimmi48:
### Description

The current `enableExtensionUpdateCheck` is confusing because it does not prevent VS Code from auto-updating extensions that were installed declaratively. The new `enableExtensionUpdate` option probably matches what most users would like better, by disabling both the update notifications and the auto-updates. Perhaps the `enableExtensionUpdateCheck` option should even be deprecated in favor of this new one.

cc @wamserma who initially contributed this option (#2965)

### Checklist

- [x] ~~Code tested through `nix-shell --pure tests -A run.all` or `nix develop --ignore-environment .#all` using Flakes.~~ Only ran VS Code-specific tests.

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/vscode.nix
+++ b/modules/programs/vscode.nix
@@ -48,6 +48,9 @@ let
     // optionalAttrs (!cfg.enableUpdateCheck) { "update.mode" = "none"; }
     // optionalAttrs (!cfg.enableExtensionUpdateCheck) {
       "extensions.autoCheckUpdates" = false;
+    } // optionalAttrs (!cfg.enableExtensionUpdate) {
+      "extensions.autoUpdate" = false;
+      "extensions.autoCheckUpdates" = false;
     };
 in {
   imports = [
@@ -77,6 +80,15 @@ in {
         default = true;
         description = ''
           Whether to enable update checks/notifications.
+        '';
+      };
+
+      enableExtensionUpdate = mkOption {
+        type = types.bool;
+        default = true;
+        description = ''
+          Whether to enable automatic extension updates.
+          Disabling this also disables update checks.
         '';
       };
 


modules/programs/vscode.nix
```diff
@@ -48,6 +48,9 @@ let
     // optionalAttrs (!cfg.enableUpdateCheck) { "update.mode" = "none"; }
     // optionalAttrs (!cfg.enableExtensionUpdateCheck) {
       "extensions.autoCheckUpdates" = false;
+    } // optionalAttrs (!cfg.enableExtensionUpdate) {
+      "extensions.autoUpdate" = false;
+      "extensions.autoCheckUpdates" = false;
```
wamserma, line 53: It would be nice to emit a warning if settings are conflicting. Evaluation can continue though, with preference for `enableExtensionUpdate` as noted in the description of that option.

wamserma (approved):
Diff LGTM, didn't test.

modules/programs/vscode.nix
Zimmi48: The issue is that AFAIK there is no way to distinguish the user having explicitly set `enableExtensionUpdateCheck = true;` or having just left the option to the default value, and in the latter case, I don't think we should warn.

Maybe deprecating `enableExtensionUpdateCheck` would be the best way to ensure that there are no conflicting settings in the long run...

wamserma: The warning could state that `enableExtensionUpdate` overrides `enableExtensionUpdateCheck` and that enableExtensionUpdateCheck should be set explicitly to false to avoid the warning.

Zimmi48: Yes, but if we are going to ask users to set both options to `false`, we could just as well have two separate options, one that corresponds to `extensions.autoUpdate` and one to `extensions.autoUpdate`.

PR #4478, opened 2023-09-20
Fix fusuma service (grep)

dbaynard:
The version of fusuma in nixpkgs as of two weeks ago requires grep.

https://github.com/iberianpig/fusuma/blob/ea76c7cf547c8ee5222121293bcfc2b23555f989/lib/fusuma/libinput_command.rb#L43

The default service should therefore include a provider of grep — I've selected gnugrep as that is what is on my system.
I did confirm that a service file with gnugrep works.
However, I did _not_ run the default service file — any meaningful file will have something like `xdotool`; I confirmed that my generated service file (with gnugrep and xdotool) works, whereas the version without gnugrep does not.

### DIFF
--- a/modules/services/fusuma.nix
+++ b/modules/services/fusuma.nix
@@ -96,10 +96,10 @@ in {
 
     extraPackages = mkOption {
       type = types.listOf types.package;
-      default = with pkgs; [ coreutils ];
-      defaultText = literalExpression "pkgs.coreutils";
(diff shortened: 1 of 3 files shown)


rycee:
Looks like grep is always needed by fusuma, so I think be unconditionally included in PATH, i.e., not included in extraPackages.

dbaynard:
That's been fixed upstream.

PR #4487, opened 2023-09-21
direnv: fix nushell syntax

siph:
### Description

The current syntax is invalid for the latest release of nushell (`0.85.0`). I think this is related to https://github.com/nushell/nushell/pull/10103.

The latest nushell release is ~~not yet~~ packaged in `nixpkgs`  (https://github.com/NixOS/nixpkgs/pull/256330), however this change is backwards compatible. I verified by testing it with the both the current unstable version `0.84.0` and the latest version `0.85.0`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/direnv.nix
+++ b/modules/programs/direnv.nix
@@ -134,7 +134,7 @@ in {
         $env.config = ($env.config | update hooks.pre_prompt ($env.config.hooks.pre_prompt | append {
           code: "
             let direnv = (${pkgs.direnv}/bin/direnv export json | from json)
-            let direnv = if ($direnv | length) == 1 { $direnv } else { {} }
+            let direnv = if not ($direnv | is-empty) { $direnv } else { {} }
             $direnv | load-env
             "
         }))


aidalgol:
> The latest nushell release is not yet packaged in `nixpkgs` NixOS/nixpkgs#256330

This has now been merged.

siph:
Oops. I also think @happysalada might be a good person to look at the nushell code.

happysalada:
hey, I've got a really weird result testing this.
```
❯ : $direnv | length
1
mono_sweif on  kyb [$!?] via  v20.6.1 via ❄  impure (devshell-env) on ☁  (auto)
❯ : $direnv | is-empty
true
```
(on 0.84, it looks like it's doing the opposite of what is intented)

happysalada:
Should we open an issue upstream to see how this should be fixed ?

siph:
@happysalada Thanks for looking at this. Yeah there is definitely some weird
stuff going on with this after looking at it but I'm not sure what the culprit
is.

I rolled back to `0.84.0` to test this.

First, `$direnv` seems to get consumed after the first call so `($direnv |
is-empty) == true` is actually correct.

You can test this by freshly `cd`ing into a directory with `direnv` and try:
```shell
nixvim-flake on  master [!] via ❄️  impure (nix-shell-env)
❯ : $direnv | is-empty
false

nixvim-flake on  master [!] via ❄️  impure (nix-shell-env)
❯ : $direnv | is-empty
true

nixvim-flake on  master [!] via ❄️  impure (nix-shell-env)
❯ : $direnv
╭──────────────╮
│ empty record │
╰──────────────╯
```

In addition to this, the empty direnv always has a length of `1`, so this isn't
working correctly anyway.

```shell
nixvim-flake on  master [!] via ❄️  impure (nix-shell-env) 
❯ : $direnv
╭──────────────╮
│ empty record │
╰──────────────╯

nixvim-flake on  master [!] via ❄️  impure (nix-shell-env) 
❯ : $direnv | length
1

```

Also, this pr does work. I'm having expected behavior while using it on both
`84` and `85`.

This is while using my branch as the `home-manager` input and on nushell
`0.84.0`:

```shell
~
❯ : statix --version
Error: nu::shell::external_command

  × External command failed
   ╭─[entry #9:1:1]
 1 │ statix --version
   · ───┬──
   ·    ╰── executable was not found
   ╰────
  help: No such file or directory (os error 2)


~
❯ : cd .nix-dotfiles

.nix-dotfiles (013fe12) [!] via ❄️  impure (nix-shell-env) 
❯ : statix --version
statix 0.5.8
```

FWIW this is what the `0.85.0` error looks like.
```shell

.nix-dotfiles on  master via ❄️  impure (nix-shell-env)
❯ : $direnv | length
Error: nu::shell::only_supports_this_input_type

  × Input type not supported.
   ╭─[entry #1:1:1]
 1 │ $direnv | length
   · ───┬───   ───┬──
   ·    │         ╰── only list, and table input data is supported
   ·    ╰── input type: record
   ╰────

```

I'm not really sure how to proceed here because this does fix the upcoming bug
without breaking backwards-compatibility, as far as I can tell anyway. But
there is something wrong in general, I think.

happysalada:
ah, finding the correct code to run and fixing the problem should be two separate things.
I'm going to try to find the "proper" solution for this when I have a moment, but in the meanwhile, let's merge this!

happysalada:
alright the same PR has been made upstream https://github.com/nushell/nu_scripts/pull/615
this is definitely good to merge.

happysalada:
I'm asking there how this script should work actually

rycee:
I can't say that I read through the discussions very closely here but I saw "this is definitely good to merge", so I merged. If that was a mistake let me know 😉

happysalada:
All good ! Thank you for all the work you are doing on home-manager !

siph:
Totally agree. Thanks for the assistance @happysalada.

PR #4492, opened 2023-09-22
fish: also query pname and version for completions

### LINKED ISSUE #2813: Fish completions fail for missing attribute 'name'
After a recent update of my nixpkgs (unstable channel), home-manager will not rebuild anymore if I enable fish. Error message as follows:

```
error: attribute 'name' missing

       at /nix/store/wfgrdl7b24fxwb2rqw7nz1x8wd5fk3ch-source/modules/programs/fish.nix:308:30:

          307|         generateCompletions = package:
          308|           pkgs.runCommand "${package.name}-fish-completions" {
          309|             src = package;
```

My suspicion is that the code at

https://github.com/nix-community/home-manager/blob/835797f3a4a59459a316ae8d4ab91fa59faf61a4/modules/programs/fish.nix#L307-L313

relies too much on the `name` attribute which often is replaced by `pname` and `version`, as seems to be the case in [nixpkgs' definition for fish](https://github.com/NixOS/nixpkgs/blob/19d4f7dc485f74109bd66ef74231285ff797a823/pkgs/shells/fish/default.nix#L135-L137). I have seen that the `stdenv` checker does use a `getName` function to retrieve the package name (see [here](https://github.com/NixOS/nixpkgs/blob/19d4f7dc485f74109bd66ef74231285ff797a823/pkgs/stdenv/generic/check-meta.nix#L10)) but that might be overkill since seemingly only the module here is affected.

I'm not entirely sure (and can't test at the moment) but maybe using `pkgs.runCommand "${package.pname}-${package.version}-fish-completions"` instead of the current `package.name` would already be enough?

Edit: Actually, it seems that not all packages have `pname` either, so I just tried running with the following snippet

```nix
        generateCompletions = let
          getName = attrs: attrs.name or ("${attrs.pname or "«name-missing»"}-${attrs.version or "«version-missing»"}");
        in package:
          pkgs.runCommand "${getName package}-fish-completions" {
            src = package;
            nativeBuildInputs = [ pkgs.python2 ];
            buildInputs = [ cfg.package ];
            preferLocalBuild = true;
          } ''
            mkdir -p $out
            if [ -d $src/share/man ]; then
              find $src/share/man -type f \
                | xargs python ${cfg.package}/share/fish/tools/create_manpage_completions.py --directory $out \
                > /dev/null
            fi
          '';
```

and it at least solved my issue. I don't know whether this is a good place to insert such a `getName` function (taken from the source linked above) or whether it should be imported somehow, hence I won't prepare a PR.

a-kenji:
### Description

Query the `pname` and `version` attributes for completion generation, if the `name` attribute is not available.

Fixes #2813

This is pretty much @stopbystudent's approach from the issue. 
I tried using `pkgs.lib.getName`, but didn't find a good way to integrate it. 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
I couldn't find a `meta.maintainers`, so:
```
$ git log --since='2023-01-01' --follow --format='%an <%ae>' -- modules/programs/fish.nix | sort | uniq
```
@emilazy 
@yu-re-ka 
@marsam

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -320,24 +320,29 @@ in {
             ${postBuild}
           '';
 
-        generateCompletions = package:
-          pkgs.runCommand "${package.name}-fish-completions" {
-            srcs = [ package ] ++ filter (p: p != null)
-              (builtins.map (outName: package.${outName} or null)
-                config.home.extraOutputsToInstall);
-            nativeBuildInputs = [ pkgs.python3 ];
-            buildInputs = [ cfg.package ];
-            preferLocalBuild = true;
-          } ''
-            mkdir -p $out
-            for src in $srcs; do
-              if [ -d $src/share/man ]; then
-                find -L $src/share/man -type f \
-                  | xargs python ${cfg.package}/share/fish/tools/create_manpage_completions.py --directory $out \
-                  > /dev/null
-              fi
-            done
-          '';
+        generateCompletions = let
+          getName = attrs:
+            attrs.name or "${attrs.pname or "«pname-missing»"}-${
+              attrs.version or "«version-missing»"
+            }";
+        in package:
+        pkgs.runCommand "${getName package}-fish-completions" {
+          srcs = [ package ] ++ filter (p: p != null)
+            (builtins.map (outName: package.${outName} or null)
+              config.home.extraOutputsToInstall);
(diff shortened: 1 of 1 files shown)


rycee:
Thanks! I don't know much about fish but this looks reasonable to me so I merged to master now 🙂

PR #4507, opened 2023-09-28
WIP: mpd: add darwin support

Benjamin-L:
### Description

Added darwin support to the `services.mpd` module.

Still need to fix up the tests for this.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

There doesn't seem to be anybody assigned to this module...

### DIFF
--- a/modules/services/mpd.nix
+++ b/modules/services/mpd.nix
@@ -25,6 +25,8 @@ let
     ${cfg.extraConfig}
   '';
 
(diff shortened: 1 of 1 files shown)


Benjamin-L:
I'm still using this patch locally, but have found that it's pretty unreliable when waking up the laptop from sleep. I don't think I'm gonna have energy to work on this any time soon, and am closing the PR. If anybody wants to pick this up, the next steps are:

 - try to reproduce the sleep problems and possibly resolve them. I'm putting this at 50/50 for whether it's a problem with the launchd config or with mpd itself.
 - refactor the existing mpd tests for darwin support

PR #4510, opened 2023-09-29
awscli: add module

anthonyroussel:
### Description

Add `programs.awscli` module to:

* Automatically install awscli2 in your home environment
* Configure your ~/.aws/config and ~/.aws/credentials files

Also added myself as maintainer of the `programs.awscli` module.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/awscli.nix
+++ b/modules/programs/awscli.nix
@@ -1,0 +1,67 @@
+{ config, lib, pkgs, ... }:
+
+let
+  cfg = config.programs.awscli;
+  iniFormat = pkgs.formats.ini { };
+
+in {
+  meta.maintainers = [ lib.maintainers.anthonyroussel ];
+
+  options.programs.awscli = {
+    enable = lib.mkEnableOption "AWS CLI tool";
+
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.awscli2;
+      defaultText = lib.literalExpression "pkgs.awscli2";
+      description = "Package providing {command}`aws`.";
+    };
+
+    settings = lib.mkOption {
+      type = lib.types.submodule { freeformType = iniFormat.type; };
+      default = { };
+      example = lib.literalExpression ''
+        {
+          "default" = {
+            region = "eu-west-3";
+            output = "json";
+          };
+        };
+      '';
+      description = "Configuration written to {file}`$HOME/.aws/config`.";
+    };
+
+    credentials = lib.mkOption {
+      type = lib.types.submodule { freeformType = iniFormat.type; };
+      default = { };
+      example = lib.literalExpression ''
+        {
+          "default" = {
+            "credential_process" = "${pkgs.pass}/bin/pass show aws";
+          };
+        };
+      '';
+      description = ''
+        Configuration written to {file}`$HOME/.aws/credentials`.
+
+        For security reasons, never store cleartext passwords here.
+        We recommend that you use `credential_process` option to retrieve
+        the IAM credentials from your favorite password manager during runtime,
+        or use AWS IAM Identity Center to get short-term credentials.
+
+        See <https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-authentication.html>.
+      '';
+    };
+  };
+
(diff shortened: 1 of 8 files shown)


modules/lib/maintainers.nix
```diff
@@ -13,6 +13,16 @@
...
+  anthonyroussel = {
+    email = "anthony@roussel.dev";
+    github = "anthonyroussel";
+    githubId = 220084;
+    name = "Anthony Roussel";
+    keys = [{
+      longKeyId = "ed25519/0x9DC4987B1A55E75E";
+      fingerprint = "472D 368A F107 F443 F3A5  C712 9DC4 987B 1A55 E75E";
+    }];
+  };
```
rycee, line 25: You already seem to be listed in the Nixpkgs maintainers list so then there is no need to be here as well.
```suggestion
```

modules/programs/awscli.nix
```diff
@@ -0,0 +1,87 @@
...
+      iniFormat.generate "aws-credentials-${config.home.username}"
+      cfg.credentials;
+
+    programs.bash.initExtra = lib.mkIf cfg.enableBashIntegration ''
+      source ${cfg.package}/share/bash-completion/completions/aws
+    '';
+
+    programs.zsh.initExtra = lib.mkIf cfg.enableZshIntegration ''
+      source ${cfg.package}/share/zsh/site-functions/aws_zsh_completer.sh
+    '';
```
rycee, line 85: This should not be necessary, the shells should already look for completions in the expected locations.

Should then also be able to remove these two "enable" options.

modules/programs/awscli.nix
```diff
@@ -0,0 +1,87 @@
...
+      default = { };
+      description = ''
+        Configuration written to {file}`$HOME/.aws/credentials`.
+
+        For security reasons, never store cleartext passwords here.
+        We recommend that you use `credential_process` option to retrieve
+        the IAM credentials from your favorite password manager during runtime,
+        or use AWS IAM Identity Center to get short-term credentials.
+
+        See https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-authentication.html
```
rycee, line 57:
```suggestion
        See <https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-authentication.html>.
```

modules/programs/awscli.nix
```diff
@@ -0,0 +1,87 @@
...
+      type = lib.types.bool;
+      default = true;
+      description = "Whether to enable AWS CLI's Zsh integration.";
+    };
+
+    settings = lib.mkOption {
+      type = lib.types.submodule { freeformType = iniFormat.type; };
+      default = { };
+      description = "Configuration written to {file}`$HOME/.aws/config`.";
+      example = lib.literalExpression ''
```
rycee, line 36: Typically we use the order

1. type
2. default
3. example
4. description

So need to swap the order of example and description for settings and credentials.

tests/modules/programs/awscli/awscli.nix
```diff
@@ -0,0 +1,42 @@
...
+          credential_process = "pass show aws";
+        };
+      };
+      enableBashIntegration = true;
+      enableZshIntegration = true;
+    };
+    bash.enable = true;
+    zsh.enable = true;
+  };
+
```
rycee, line 24: To avoid unnecessary download of awscli2 in the test:
```suggestion

  test.stubs.awscli2 = { };
 
```

tests/modules/programs/awscli/awscli.nix
```diff
@@ -0,0 +1,42 @@
+{ config, lib, pkgs, ... }:
```
rycee, line 1: Can simplify the argument:

```suggestion
{ ... }:
```

rycee:
Thanks for the contribution! I've added some comments but looks good in general :+1:

anthonyroussel:
> Thanks for the contribution! I've added some comments but looks good in general 👍

@rycee Hello. Thanks for the review! I have handled all your feedbacks.

rycee:

Issue #4520, opened 2023-10-02
bug: xplr: `cfg.plugins` does not work for more than one plugin


ChanceHarrison:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

The implementation of `programs.xplr.plugins` does not correctly set the `package.path` in the Lua configuration file that gets written out.

The result is that users cannot enable/load any additional plugins after the first.

---
For the full context, please reference [this discussion](https://github.com/sayanarijit/xplr/discussions/668) on the upstream repository.

### Maintainer CC

@NotAShelf

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.1.49, NixOS, 23.11 (Tapir), 23.11.20230830.a63a64b`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.17.0`
 - nixpkgs: `/nix/store/flsqq53kqi52lz73n543wmxz0mbhpb52-source`
```

NotAShelf:
not stale, PR stil needs merging

PR #4521, opened 2023-10-02
xplr: fix cfg.plugins implementation for more than one plugin

ChanceHarrison:
### Description

Would close https://github.com/nix-community/home-manager/issues/4520. Please see that issue (and the upstream discussion linked within) for more information.

Change the implementation of the `programs.xplr.plugins` option to correctly set the `package.path` in the written Lua config file.

Specifically, each component path of the `package.path` must have appropriate wildcard/template characters (what Lua calls "interrogation marks" or `?` characters).

Furthermore, due to the need (in the current implementation) to use `programs.xplr.extraConfig` to call `require(<plugin-name>).setup`, this implementation ensures that plugins in the Nix store are inside a predictably-named directory.

### Checklist

Unfortunately not. The `cfg.plugins` option currently takes a list for while it is proposed that it takes an attribute set.

While a change could be made that maintains complete backwards compatibility, one could argue that the usability improvements that accompany the change (of an already broken option) outweigh the cost of breaking back-compat.

- [x] Code formatted with ~~`./format`~~ `nixfmt`.

- [x] Commit messages are formatted appropriately

#### Maintainer CC

@NotAShelf

### DIFF
--- a/modules/programs/xplr.nix
+++ b/modules/programs/xplr.nix
@@ -1,8 +1,7 @@
 { config, lib, pkgs, ... }:
 let
   inherit (lib)
-    concatStringsSep types mkIf mkOption mkEnableOption mkPackageOption
-    literalExpression;
+    types mkIf mkOption mkEnableOption mkPackageOption literalExpression;
 
   cfg = config.programs.xplr;
 
@@ -10,21 +9,39 @@ let
     version = '${cfg.package.version}'
   '';
 
-  # We provide a default version line within the configuration file, which is
-  # obtained from the package's attributes. Merge the initial configFile, a
-  # mapped list of plugins and then the user defined configuration to obtain the
-  # final configuration.
-  pluginPath = if cfg.plugins != [ ] then
-    (''
-      package.path=
-    '' + (concatStringsSep " ..\n"
-      (map (p: ''"${p}/init.lua;${p}/?.lua;"'') cfg.plugins)) + ''
-         ..
-        package.path
-      '')
+  # If `value` is a Nix store path, create the symlink `/nix/store/newhash/${name}/*` 
+  # to `/nix/store/oldhash/*` and returns `/nix/store/newhash`.
+  wrapPlugin = name: value:
+    if lib.isStorePath value then
+      pkgs.symlinkJoin {
+        name = name;
+        paths = [ value ];
+        postBuild = ''
+          mkdir '${name}'
+          mv $out/* '${name}/'
+          mv '${name}' $out/
+        '';
+      }
+    else
+      builtins.dirOf value;
+
+  makePluginSearchPath = p: "${p}/?/init.lua;${p}/?.lua";
+
+  pluginPath = if cfg.plugins != { } then
+    let
+      wrappedPlugins = lib.mapAttrsToList wrapPlugin cfg.plugins;
+      searchPaths = map makePluginSearchPath wrappedPlugins;
+      pluginSearchPath = lib.concatStringsSep ";" searchPaths;
+    in (''
+      package.path = "${pluginSearchPath};" .. package.path
+    '')
   else
     "\n";
 
+  # We provide a default version line within the configuration file, which is
+  # obtained from the package's attributes. Merge the initial configFile, a
+  # mapped list of plugins and then the user defined configuration to obtain
+  # the final configuration.
   configFile = initialConfig + pluginPath + cfg.extraConfig;
 in {
   meta.maintainers = [ lib.maintainers.NotAShelf ];
@@ -35,15 +52,25 @@ in {
     package = mkPackageOption pkgs "xplr" { };
 
     plugins = mkOption {
-      type = with types; nullOr (listOf (either package str));
-      default = [ ];
-      defaultText = literalExpression "[]";
+      type = with types; nullOr (attrsOf (either package str));
+      default = { };
+      defaultText = literalExpression "{ }";
       description = ''
-        Plugins to be added to your configuration file.
+        An attribute set of plugin paths to be added to the [package.path]<https://www.lua.org/manual/5.4/manual.html#pdf-package.path> of the {file}`~/config/xplr/init.lua` configuration file.
 
-        Must be a package, an absolute plugin path, or string to be recognized
-        by xplr. Paths will be relative to
-        {file}`$XDG_CONFIG_HOME/xplr/init.lua` unless they are absolute.
+        Must be a package or string representing the plugin directory's path. 
+        If the path string is not absolute, it will be relative to {file}`$XDG_CONFIG_HOME/xplr/init.lua`.
+      '';
+      example = literalExpression ''
+        {
+          wl-clipboard = fetchFromGitHub {
+            owner = "sayanarijit";
+            repo = "wl-clipboard.xplr";
+            rev = "a3ffc87460c5c7f560bffea689487ae14b36d9c3";
+            hash = "sha256-I4rh5Zks9hiXozBiPDuRdHwW5I7ppzEpQNtirY0Lcks=";
+          }
+          local-plugin = "/home/user/.config/plugins/local-plugin";
+        };
       '';
     };
 


modules/programs/xplr.nix
```diff
@@ -35,15 +52,20 @@ in {
...
 
-        Must be a package, an absolute plugin path, or string to be recognized
-        by xplr. Paths will be relative to
-        {file}`$XDG_CONFIG_HOME/xplr/init.lua` unless they are absolute.
+        Must be a package or string representing the plugin directory's path. 
+        If the path is not absolute, it will be relative to {file}`$XDG_CONFIG_HOME/xplr/init.lua`.
+      '';
+      example = literalExpression ''
+        {
+          wl-clipboard = inputs."wl-clipboard.xplr";
```
sayanarijit, line 66: I'd use fetchGit here, for people not using git.

modules/programs/xplr.nix
```diff
@@ -1,30 +1,47 @@
...
+  # to `/nix/store/oldhash/*` and returns `/nix/store/newhash`.
+  wrapPlugin = name: value:
+    if lib.isStorePath value then
+      pkgs.symlinkJoin {
+        name = name;
+        paths = [ value ];
+        postBuild = ''
+          mkdir ${name}
+          mv $out/* ${name}/
+          mv ${name} $out/
```
sayanarijit, line 22:
```suggestion
          mkdir './${name}'
          mv $out/* './${name}/'
          mv './${name}' $out/
```

sayanarijit (changes requested):
Minor suggestions. Other than that, pr looks good. cc @NotAShelf for further review.

NotAShelf:
> Minor suggestions. Other than that, pr looks good. cc @NotAShelf for further review.

I'm in favor of approving this PR as is (with our without the changes you've mentioned) but I *really* would like some second opinion on the plugin wrapper. I'll try communicating on matrix about this.

modules/programs/xplr.nix
NotAShelf: is mkdir even necessary here?

sayanarijit: We need to make a dir named "${name}" so that we can move the contents in there.

NotAShelf (approved):

modules/programs/xplr.nix
```diff
@@ -1,30 +1,47 @@
...
+        '';
+      }
+    else
+      builtins.dirOf value;
+
+  makePluginSearchPath = p: "${p}/?/init.lua;${p}/?.lua";
```
eclairevoyant, line 28: This whole wrapping thing seems strange, what happens if you wrap a plugin in a store path that doesn't have init.lua? I assume it won't match because it'd be `${p}/?/?.lua` not just `${p}/?.lua`

Alternatively, we only allow plugins with `init.lua` (and we can get rid of all this wrapping and wildcard fluff).

eclairevoyant (changes requested):

modules/programs/xplr.nix
sayanarijit: With only ${p}/?.lua, we'd need to import the plugin as `name.init`, assuming the entrypoint is `init.lua`. It's a lua convention to add init.lua in package path to make the lib easy to import.

eclairevoyant: Then I think my original comment applies?

> what happens if you wrap a plugin in a store path that doesn't have init.lua? I assume it won't match because it'd have to be `${p}/?/?.lua` not just `${p}/?.lua

sayanarijit: The wrapping does this:
/nix/store/hash/init.lua ->
/nix/store/hash/name/init.lua

So we can add add `package.path = "/nix/store/hash/?/.init.lua;/nix/store/hash/*.lua"` and `require("name")`.

To understand the issue better, let's reproduce it.

1. Create /fakenixstore
2. Git clone sayanarijit/trash-cli.xplr as /fakenixstore/hashsha123 such that sayanarijit/trash-cli.xplr/init.lua is now /fakenixstore/hashsha123/init.lua
3. Get into lua repl (in any dir) and modify package.path in a way so that you can run `require("trash-cli").setup()`.

Here you'll see that the only way to make it work is to:
1. move or symlink `/fakenixstore/hashsha123/*` as `/fakenixstore/hashsha123/trash-cli/*`.
2. In lua repl, run `package.path = "/fakenixstore/hashsha123/?/.init.lua;/nix/fakenixstorestore/hashsha123/*.lua"`

modules/programs/xplr.nix
rycee: Agreed, although I would suggest `fetchFromGitHub`. If you search for that function you'll find already existing examples.

modules/programs/xplr.nix
```diff
@@ -35,15 +52,20 @@ in {
...
     plugins = mkOption {
-      type = with types; nullOr (listOf (either package str));
-      default = [ ];
-      defaultText = literalExpression "[]";
+      type = with types; nullOr (attrsOf (either package str));
+      default = { };
+      defaultText = literalExpression "{ }";
       description = ''
-        Plugins to be added to your configuration file.
+        An attribute set of plugin paths to be added to the [package.path]<https://www.lua.org/manual/5.4/manual.html#pdf-package.path> of the {file}`~/config/xplr/init.lua` configuration file.
```
rycee, line 59: Perhaps you mean
```suggestion
        An attribute set of plugin paths to be added to the [package.path](https://www.lua.org/manual/5.4/manual.html#pdf-package.path) of the {file}`~/config/xplr/init.lua` configuration file.
```
?

Also, I would suggest wrapping the line at roughly 80 char width.

ChanceHarrison: Thank you. Resolved by latest force push.

modules/programs/xplr.nix
```diff
@@ -35,15 +52,20 @@ in {
...
 
-        Must be a package, an absolute plugin path, or string to be recognized
-        by xplr. Paths will be relative to
-        {file}`$XDG_CONFIG_HOME/xplr/init.lua` unless they are absolute.
+        Must be a package or string representing the plugin directory's path. 
+        If the path is not absolute, it will be relative to {file}`$XDG_CONFIG_HOME/xplr/init.lua`.
```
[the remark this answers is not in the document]
ChanceHarrison: Thank you. Resolved by latest force push.

modules/programs/xplr.nix
ChanceHarrison: Thank you. Resolved by latest force push.

modules/programs/xplr.nix
ChanceHarrison: Thank you. Resolved by latest force push.

modules/programs/xplr.nix
ChanceHarrison: @eclairevoyant Based on the most recent message, is it clear why the wrapping method is being used? Do you see any better way? If I or someone can provide more clarification, just say the word. Otherwise, feel free to resolve the conversation.

ChanceHarrison:
> @NotAShelf  I'm in favor of approving this PR as is (with our without the changes you've mentioned) but I _really_ would like some second opinion on the plugin wrapper. I'll try communicating on matrix about this.

Thanks for the review. Did anything come out of your communications on Matrix regarding the plugin wrapping?

---

To all: Aside from the unresolved conversation regarding the plugin wrapping, I am moving this PR out of draft status based on the feedback received. If there is anything else you would like addressed, just let me know.

NotAShelf:
I haven't received any responses on matrix but rycee has already given a review - I also asked a few other people to provide their own.

I'll go ahead and approve this.

NotAShelf (approved):

fufexan (approved):

sayanarijit (approved):

sayanarijit:
Hi @rycee I think this is good to go, if no-one has any concern.

rycee:
Thanks! Merged to master now. If there is any residual concerns about the wrapper, perhaps those could be resolved in a future PR.

PR #4529, opened 2023-10-04
Modules/babashka

Sohalt:
### Description

Add a `programs.babashka` module for autocompletion for babashka.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/babashka.nix
+++ b/modules/programs/babashka.nix
@@ -1,0 +1,69 @@
+{ config, lib, pkgs, ... }:
+with lib;
+let cfg = config.programs.babashka;
+in {
+  meta.maintainers = [ maintainers.sohalt ];
+
+  options = {
+    programs.babashka = {
+      enable = mkEnableOption "Babashka";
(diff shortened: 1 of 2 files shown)


modules/programs/babashka.nix
```diff
@@ -0,0 +1,69 @@
...
+        function __bb_complete_tasks
+          if not test "$__bb_tasks"
+            set -g __bb_tasks (bb tasks |tail -n +3 |cut -f1 -d ' ')
+          end
+
+          printf "%s\n" $__bb_tasks
+        end
+
+        complete -c bb -a "(__bb_complete_tasks)" -d 'tasks'
+      '';
```
ambroisie, line 66: Completion should be part of the package, not the module.

And if this is the only thing this module does, I would abandon those changes and get the completion files merged in nixpkgs instead.

Sohalt: Thanks! I'll open a PR in nixpkgs then.

PR #4540, opened 2023-10-05
fish: 2 fixes

octplane:
### Description

Wrap babelfish version of `hm-session-vars.sh` into a function to ensure this can be safely evaluated at session startup. The current session vars shell script uses a `return` statement and this breaks the generated script.

Ensure that the fzf key bindings function is not called in non-interactive scenarios when fzf does not declare the function.

- This PR wraps the hm-session-vars into a immediately called function.
- This PR does not evaluate fzf_key_bindings in interactive mode

 ### Checklist

- [x] Commit messages are formatted correctly

#### Maintainer CC

- @yu-re-ka 
- @emilazy

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -221,9 +221,11 @@ let
 
   translatedSessionVariables =
     pkgs.runCommandLocal "hm-session-vars.fish" { } ''
+      (echo "function setup_hm_session_vars;"
       ${pkgs.buildPackages.babelfish}/bin/babelfish \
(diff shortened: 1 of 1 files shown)


octplane:
Hello @rycee, is there something for me to do to make this move forward?

rycee:
Sorry about the delay, I've rebased and merged it now. Thanks for the contribution! 🙂

octplane:
Just finding out about that now. Time flies! Thanks for taking care of the merge 💌

PR #4546, opened 2023-10-08
fish: improve performance by generating alias functions instead of calling `alias`

usagi-flow:
### Description

Defining aliases with the `alias` command causes a performance hit, which may become very noticeable on lower end hardware and/or as the amount of aliases grow.

This is because `alias` is not a builtin command, but rather causes an eval to take place. More background information can be found here: https://github.com/fish-shell/fish-shell/issues/828

As a much more performant alternative, directly generate functions (which is what `alias` would do anyway). Thanks to the `-w` / `--wraps` option, we can keep the inherited completion.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -143,7 +143,7 @@ let
       cfg.shellAbbrs);
 
   aliasesStr = concatStringsSep "\n"
-    (mapAttrsToList (k: v: "alias ${k} ${escapeShellArg v}") cfg.shellAliases);
+    (mapAttrsToList (k: v: "function ${k} -w ${escapeShellArg v}; ${v} $argv; end") cfg.shellAliases);
 
   fishIndent = name: text:
     pkgs.runCommand name {


trueNAHO:
Related: https://github.com/nix-community/home-manager/issues/4427

pvsr:
This would break aliases with the same name as the command they call, since `alias` prevents infinite recursion by prefixing the body with `command` or `builtin`. From https://github.com/nix-community/home-manager/issues/4427#issuecomment-1760581249:
> Alias doesn't do a lot but it does prevent recursion in a way that I don't think we can replicate in HM. For example `fish.shellAliases.vim = "vim ./scratch"` works fine since it gets rewritten to `command vim ./scratch` at runtime, but `fish.functions.vim.body = "vim ./scratch"` fails with "The function “vim” calls itself immediately, which would result in an infinite loop." We can detect when the alias name is the same as the command easily enough, but the problem is that we need to choose between `command` and `builtin` depending on the command and that decision [depends on](https://github.com/fish-shell/fish-shell/blob/c6e905a1b95234a6be88509ea3c02be26c8eda7e/share/functions/alias.fish#L53-L56) the contents of `builtin --names` in the fish instance that runs the function.

pvsr:
> That's a very good point which I didn't cover: We could easily put the `if contains $name (builtin --names)` detection in the function body.
> 
> A slightly more performant reason would be to check this at nix build time, to avoid a lookup each time the alias is invoked. I'm not sure if/how that would be possible, however.
> 
> Maybe, a good tradoff would be to "simply" verify if the alias and the first command token are the same, and if so, generate the `if contains $name (builtin --names)` code in the function body.

Good idea. The performance hit of `if contains $name (builtin --names)` seems to be very low, less than 100 microseconds on my machine, and many aliases wouldn't be affected anyway. And it makes a lot of sense to move the performance cost from startup to alias runtime.

`alias` also [handles a similar recursion issue](https://github.com/fish-shell/fish-shell/blob/c6e905a1b95234a6be88509ea3c02be26c8eda7e/share/functions/alias.fish#L61-L68) with `--wraps` by skipping wrapper generation when the alias name is the first or last word of the command. I haven't tried reproducing the issue against this PR but I'm guessing it's possible and would affect HM users so you'd probably need to copy that check as well.

With those two checks we would be pretty much reimplementing all of `alias` except for the argument parsing. That makes me worry slightly that we may need to keep up with future changes to `alias` to remain fully compatible, but realistically it probably won't change much if at all.

trueNAHO:
> A slightly more performant reason would be to check this at nix build time, to avoid a lookup each time the alias is invoked. I'm not sure if/how that would be possible, however.

The following procedure would determine this at Nix build time:

> > [T]he problem is that we need to choose between `command` and `builtin` depending on the command and that decision [depends on](https://github.com/fish-shell/fish-shell/blob/c6e905a1b95234a6be88509ea3c02be26c8eda7e/share/functions/alias.fish#L53-L56) the contents of `builtin --names` in the fish instance that runs the function.
>
> Either we hard-code the current `builtin --names` output into the fish module and [replicate the official implementation](https://github.com/fish-shell/fish-shell/blob/c6e905a1b95234a6be88509ea3c02be26c8eda7e/share/functions/alias.fish#L53-L56), or we close this issue to reduce the maintenance of keeping the `builtin --names` up-to-date.
>
> However, it might be safe to assume that the language `builtin --names` are stable and will not receive drastic changes in the future. In that case, hard-coding the current `builtin --names` output into the fish module seems like the simplest solution, unless the Home Manager deployment system would update the `builtin --names` output.
>
> (Source: https://github.com/nix-community/home-manager/issues/4427#issuecomment-1764684165)

However, I am in favor of determining this at runtime considering the future-proofness:

> > Maybe, a good tradoff would be to "simply" verify if the alias and the first command token are the same, and if so, generate the `if contains $name (builtin --names)` code in the function body.
> 
> Good idea. The performance hit of `if contains $name (builtin --names)` seems to be very low, less than 100 microseconds on my machine, and many aliases wouldn't be affected anyway. And it makes a lot of sense to move the performance cost from startup to alias runtime.

PR #4547, opened 2023-10-08
password-store-sync: remove module

pacien:
### Description

The module `services.git-sync` (introduced two years after) provides
the same functionality, but in a more general and reliable way.

### Checklist

      __It is not. Release note entry added accordingly.__


ambroisie (commented):
Perhaps there should be something similar to [NixOS' `rename.nix`](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/rename.nix) to give a proper error message to the user (through `mkRemovedOptionModule [ "services" "password-store-sync" "enable" ]`)

ambroisie (approved):

PR #4548, opened 2023-10-09
syncthingtray: add extraOptions

motiejus:
The test currently fails, because the nix path in the generated unit does not get rewritten to `@syncthingtray@`. Pointers would be appreciated: I was not able to figure out how to make it rewrite the resulting paths.

### Description

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

cc @rycee


poelzi (approved):

olmokramer:
There's already multiple ways available to pass extra arguments to the tray service, so I'm not sure if we should create yet another option to do this. Are there other places where the user can set an arbitrary command to be run? How do we handle it in those cases?

```nix
{
  config.services.syncthing.tray = {
    command = "syncthingtray --wait";
  };
}
```

and 

```nix
{
  config.services.syncthing.tray = {
    package = pkgs.writeShellScriptBin "syncthingtray" ''
      ${pkgs.syncthingtray}/bin/syncthingtray --wait
    '';
    command = "syncthingtray";
  };
}
```

rycee:
It may be worth removing the `command` option, I think it could be replaced by the use of `getExe`. Then having an `extraOptions` would make sense, I think. I.e., to do something like

``` nix
[ "${getExe cfg.tray.package}" ] ++ cfg.tray.extraOptions;
```

PR #4549, opened 2023-10-09
aerc: symlink aerc directory on macos

heywoodlh:
### Description

On MacOS, aerc uses the ~/Library/Preferences/aerc directory instead of ~/.config/aerc. This adds a symlink at ~/Library/Preferences/aerc pointing to ~/.config/aerc on MacOS.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@lukasngl

### DIFF
--- a/modules/programs/aerc.nix
+++ b/modules/programs/aerc.nix
@@ -3,6 +3,7 @@
 with lib;
 let
   cfg = config.programs.aerc;
+  homeDir = config.home.homeDirectory;
 
   primitive = with types;
     ((type: either type (listOf type)) (nullOr (oneOf [ str int bool float ])))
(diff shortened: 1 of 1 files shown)


rycee:
Thanks for the contribution! I think it would be better, though, to immediately place the file in the correct location. See, for example, [navi.nix](https://github.com/nix-community/home-manager/blob/3c1d8758ac3f55ab96dcaf4d271c39da4b6e836d/modules/programs/navi.nix) for such a solution.

heywoodlh:
@rycee sounds good, [it looks like this PR is going to do that](https://github.com/nix-community/home-manager/pull/4550), so I'm going to close this one.

PR #4552, opened 2023-10-11
zsh: allow variable expansion for `zsh-history-substring-search` key codes

CardboardTurkey:
Switch from single to double quotes.

Also adds suggestion that I found helpful to descriptions (taken from
the project's [README](https://github.com/zsh-users/zsh-history-substring-search/blob/4abed97b6e67eb5590b39bcd59080aa23192f25d/README.md?plain=1#L71-L74)).

### Description

See above.

There may be an argument for adding `$terminfo[kcuu1]` and `$terminfo[kcud1]` to the defaults although I don't know enough about `terminfo` to know if the variable will be defined for everyone's environments.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/zsh.nix
+++ b/modules/programs/zsh.nix
@@ -204,7 +204,8 @@ let
         default = [ "^[[A" ];
         description = ''
           The key codes to be used when searching up.
-          The default of `^[[A` corresponds to the UP key.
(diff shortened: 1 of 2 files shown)


rycee:

PR #4557, opened 2023-10-12
thefuck: add instant mode option

redxtech:
### Description

This adds the option to add `--enable-experimental-instant-mode` to the initiation of `thefuck`.

I also moved the init command to a variable to be more DRY.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@ilaumjd

### DIFF
--- a/modules/programs/thefuck.nix
+++ b/modules/programs/thefuck.nix
@@ -1,7 +1,8 @@
 { config, lib, pkgs, ... }:
+
 with lib;
-let cfg = config.programs.thefuck;
-in {
+
+{
   meta.maintainers = [ hm.maintainers.ilaumjd ];
 
   options.programs.thefuck = {
@@ -9,6 +10,8 @@ in {
       "thefuck - magnificent app that corrects your previous console command";
 
     package = mkPackageOption pkgs "thefuck" { };
+
+    enableInstantMode = mkEnableOption "thefuck's experimental instant mode";
 
     enableBashIntegration = mkOption {
       default = true;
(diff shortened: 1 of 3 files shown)


redxtech:
if you accept this PR, would you also be able to assign the `HACKTOBERFEST-ACCEPTED` label to it? Gotta get those contributions up!

:slightly_smiling_face:

modules/programs/thefuck.nix
```diff
@@ -27,15 +35,17 @@ in {
     };
   };
 
-  config = mkIf cfg.enable {
+  config = let
+    cmd = ''
+      eval "$(${cfg.package}/bin/thefuck --alias${
+        optionalString cfg.instantMode " --enable-experimental-instant-mode"
+      })"
+    '';
```
ilaumjd, line 43: maybe you can extract the argument into a new variable

ilaumjd (commented):
I was thinking of adding this, but this is experimental and i'm not sure if it is okay to add. Let's wait for the others, it is okay from my side.

modules/programs/thefuck.nix
redxtech: ok, i've moved it to its own variable.

redxtech:
> I was thinking of adding this, but this is experimental and i'm not sure if it is okay to add.

it's been "experimental" for a few years at this point, but i don't see a reason why it shouldn't be added.

modules/programs/thefuck.nix
```diff
@@ -36,9 +36,10 @@ in {
   };
 
   config = let
+    instantFlag = " --enable-experimental-instant-mode";
```
ilaumjd, line 39: here's what i mean
```suggestion
    instantFlag = optionalString cfg.instantMode " --enable-experimental-instant-mode";
```

rycee:
Thanks! I've merged this now, please note that I changed the option name to `enableInstantMode`.

PR #4579, opened 2023-10-18
qt: multiple fixes

thiagokokada:
### Description

Remove remaining Qt 4 support from `qt` module since `nixpkgs` does not support Qt 4 anymore.

Export `QT_PLUGIN_PATH` and `QML2_IMPORT_PATH`, necessary for some plugins to work (e.g. `qt6ct`).

Remove top-level `with lib`.

Add missing style mappings.

Add `lxqt` support.

### Checklist

  + Removal of Qt 4 technically is not backwards compatible, but I don't think nobody cares.

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/qt.nix
+++ b/modules/misc/qt.nix
@@ -3,12 +3,19 @@
 let
   cfg = config.qt;
 
+  # Map platform names to their packages.
   platformPackages = with pkgs; {
     gnome = [ qgnomeplatform qgnomeplatform-qt6 ];
     gtk = [ libsForQt5.qtstyleplugins qt6Packages.qt6gtk2 ];
     kde = [ libsForQt5.plasma-integration libsForQt5.systemsettings ];
     lxqt = [ lxqt.lxqt-qtplugin lxqt.lxqt-config ];
     qtct = [ libsForQt5.qt5ct qt6Packages.qt6ct ];
+  };
+
+  # Maps style names to their QT_QPA_PLATFORMTHEME, if necessary.
+  styleNames = {
+    gtk = "gtk2";
+    qtct = "qt5ct";
   };
 
   # Maps known lowercase style names to style packages. Non-exhaustive.
@@ -152,12 +159,8 @@ in {
       qtVersions = with pkgs; [ qt5 qt6 ];
       makeQtPath = prefix: basePath: qt: "${basePath}/${qt.qtbase.${prefix}}";
     in lib.filterAttrs (n: v: v != null) {
-      QT_QPA_PLATFORMTHEME = if cfg.platformTheme == "gtk" then
-        "gtk2"
-      else if cfg.platformTheme == "qtct" then
-        "qt5ct"
-      else
-        cfg.platformTheme;
+      QT_QPA_PLATFORMTHEME =
+        styleNames.${cfg.platformTheme} or cfg.platformTheme;
       QT_STYLE_OVERRIDE = cfg.style.name;
       QT_PLUGIN_PATH = "$QT_PLUGIN_PATH\${QT_PLUGIN_PATH:+:}"
         + (lib.concatStringsSep ":"
--- a/modules/misc/qt.nix
+++ b/modules/misc/qt.nix
@@ -57,7 +57,7 @@ in {
 
       platformTheme = lib.mkOption {
         type = with lib.types;
-          nullOr (enum [ "gtk" "gnome" "lxqt" "qtct" "kde" ]);
+          nullOr (enum [ "gtk" "gtk3" "gnome" "lxqt" "qtct" "kde" ]);
         default = null;
         example = "gnome";
         relatedPackages = [
@@ -80,6 +80,10 @@ in {
           `gtk`
           : Use GTK theme with
             [`qtstyleplugins`](https://github.com/qt/qtstyleplugins)
+
+          `gtk3`
+          : Use [GTK3 integration](https://github.com/qt/qtbase/tree/dev/src/plugins/platformthemes/gtk3)
+            for file picker dialogs, font and theme configuration
 
           `gnome`
           : Use GNOME theme with
--- a/modules/misc/qt.nix
+++ b/modules/misc/qt.nix
@@ -38,7 +38,7 @@ in {
 
   options = {
     qt = {
-      enable = mkEnableOption "Qt 4, 5 and 6 configuration";
+      enable = mkEnableOption "Qt 5 and 6 configuration";
 
       platformTheme = mkOption {
         type = types.nullOr (types.enum [ "gtk" "gnome" "qtct" "kde" ]);
@@ -170,14 +170,5 @@ in {
 
     xsession.importedVariables = [ "QT_QPA_PLATFORMTHEME" ]
       ++ lib.optionals (cfg.style.name != null) [ "QT_STYLE_OVERRIDE" ];
-
-    # Enable GTK+ style for Qt4 in Gtk/GNOME.
-    # It doesn’t support the platform theme packages.
-    home.activation.useGtkThemeInQt4 =
-      mkIf (cfg.platformTheme == "gtk" || cfg.platformTheme == "gnome")
-      (hm.dag.entryAfter [ "writeBoundary" ] ''
-        $DRY_RUN_CMD ${pkgs.crudini}/bin/crudini $VERBOSE_ARG \
-          --set "${config.xdg.configHome}/Trolltech.conf" Qt style GTK+
-      '');
   };
 }
--- a/modules/misc/qt.nix
+++ b/modules/misc/qt.nix
@@ -4,7 +4,7 @@ let
   cfg = config.qt;
 
   platformPackages = with pkgs; {
-    gnome = [ qgnomeplatform ];
+    gnome = [ qgnomeplatform qgnomeplatform-qt6 ];
     gtk = [ libsForQt5.qtstyleplugins qt6Packages.qt6gtk2 ];
     kde = [ libsForQt5.plasma-integration libsForQt5.systemsettings ];
     lxqt = [ lxqt.lxqt-qtplugin lxqt.lxqt-config ];
@@ -55,6 +55,7 @@ in {
         example = "gnome";
         relatedPackages = [
           "qgnomeplatform"
+          "qgnomeplatform-qt6"
           [ "libsForQt5" "plasma-integration" ]
           [ "libsForQt5" "qt5ct" ]
           [ "libsForQt5" "qtstyleplugins" ]
--- a/modules/misc/qt.nix
+++ b/modules/misc/qt.nix
@@ -127,7 +127,10 @@ in {
   config = let
 
     # Necessary because home.sessionVariables doesn't support mkIf
-    envVars = filterAttrs (n: v: v != null) {
+    envVars = let
+      inherit (config.home) profileDirectory;
+      qtVersions = with pkgs; [ qt5 qt6 ];
+    in filterAttrs (n: v: v != null) {
       QT_QPA_PLATFORMTHEME = if cfg.platformTheme == "gtk" then
         "gtk2"
       else if cfg.platformTheme == "qtct" then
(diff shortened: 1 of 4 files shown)


modules/misc/qt.nix
```diff
@@ -127,14 +127,25 @@ in {
...
         cfg.platformTheme;
       QT_STYLE_OVERRIDE = cfg.style.name;
+      QT_PLUGIN_PATH = "$QT_PLUGIN_PATH\${QT_PLUGIN_PATH:+:}"
+        + (lib.concatStringsSep ":"
+          (map (qt: "${profileDirectory}/${qt.qtbase.qtPluginPrefix}")
+            qtVersions));
+      QML2_IMPORT_PATH = "$QML2_IMPORT_PATH\${QML2_IMPORT_PATH:+:}"
+        + (lib.concatStringsSep ":"
+          (map (qt: "${profileDirectory}/${qt.qtbase.qtQmlPrefix}")
+            qtVersions));
```
thiagokokada, line 148: Tested in my config:

```
$ echo $QT_PLUGIN_PATH
/nix/store/hdvlnzaxdm06xkipg62mj30aa3fzi5fa-fcitx5-with-addons-5.1.1/lib/qt-6/plugins:/etc/profiles/per-user/thiagoko/lib/qt-5.15.10/plugins:/etc/profiles/per-user/thiagoko/lib/qt-6/plugins
$ echo $QML2_IMPORT_PATH
/etc/profiles/per-user/thiagoko/lib/qt-5.15.10/qml:/etc/profiles/per-user/thiagoko/lib/qt-6/qml
```

Seems to be working fine, including properly keeping the current plugins in PATH.

thiagokokada (commented):
CC @ncfavier.

thiagokokada:
I added a workaround since https://github.com/nix-community/home-manager/pull/4468 we also set the `QT_PLUGIN_PATH` when `i18n.inputMethod.enabled = 'fcitx5'.

modules/misc/qt.nix
```diff
@@ -147,37 +182,31 @@ in {
...
+      sessionVariables = envVars;
+      # home.sessionVariables does not support setting the same environment
+      # variable to different values.
+      # Since some other modules may set the QT_PLUGIN_PATH or QML2_IMPORT_PATH
+      # to their own value, e.g.: fcitx5, we avoid conflicts by setting
+      # the values in home.sessionVariablesExtra instead.
+      sessionVariablesExtra = ''
+        export QT_PLUGIN_PATH=${envVarsExtra.QT_PLUGIN_PATH}
+        export QML2_IMPORT_PATH=${envVarsExtra.QML2_IMPORT_PATH}
+      '';
```
thiagokokada, line 198: I know that using `home.sessionVariables` here is not ideal, but this is the only solution I found for now.

We probably need something similar to `environment.profileRelativeEnvVars` or `environment.profileRelativeSessionVariables` in NixOS to fix this issue once for all.

thiagokokada: This is at least better than the previous solution I did: https://github.com/nix-community/home-manager/pull/4579/commits/3452e14ec729e0582a5fb120c14ed5c6d7fb64d3

thiagokokada: > We probably need something similar to environment.profileRelativeEnvVars or environment.profileRelativeSessionVariables in NixOS to fix this issue once for all.

Or even simpler, as long as we have something that allows a similar semantic as `home.sessionPath`, but generic (e.g.: `home.searchPathVariables.QT_PLUGIN_PATH = [ ... ]`), this would already work.

thiagokokada: I decided to fix this issue in a separate PR: https://github.com/nix-community/home-manager/pull/4582.

I think this PR can still be merged before that PR above. In any case once the other PR is accepted I will change this PR to re-implement this in terms of `home.sessionSearchVariables`.

9p4:
Does this supersede https://github.com/nix-community/home-manager/pull/4306

thiagokokada:
> Does this supersede #4306

I would say so.

While technically the approach from #4306 is more flexible, I am not sure it makes sense to have those 2 environment variables configurable, except to workaround the composition issue that should be fixed in #4582.

Also this PR fixes a bunch of other stuff (like removal of Qt 4) and adds some other stuff (the LXQt), so I would say this PR pretty much supersede #4306 in many ways.

CC @rycee.

modules/misc/qt.nix
```diff
@@ -8,15 +8,15 @@ let
...
     plastique = libsForQt5.qtstyleplugins;
 
-    adwaita = adwaita-qt;
-    adwaita-dark = adwaita-qt;
-    adwaita-highcontrast = adwaita-qt;
-    adwaita-highcontrastinverse = adwaita-qt;
+    adwaita = [ adwaita-qt adwaita-qt6 ];
+    adwaita-dark = [ adwaita-qt adwaita-qt6 ];
+    adwaita-highcontrast = [ adwaita-qt adwaita-qt6 ];
+    adwaita-highcontrastinverse = [ adwaita-qt adwaita-qt6 ];
```
9ary, line 34: A bit iffy on this, considering upstream is discontinued (https://github.com/FedoraQt/adwaita-qt#adwaita-qt-project-is-unmaintained-a-no-longer-actively-developed).

I also added it in #4312, but that was a few days before the notice was put up.

modules/misc/qt.nix
```diff
@@ -3,6 +3,14 @@
 let
   cfg = config.qt;
 
+  platformPackages = with pkgs; {
```
9ary, line 7: If you're going to drop `with`, I think you should drop all instances of it.

modules/misc/qt.nix
```diff
@@ -172,16 +189,7 @@ in {
...
-      [ pkgs.qgnomeplatform ]
-    else if cfg.platformTheme == "qtct" then [
-      pkgs.libsForQt5.qt5ct
-      pkgs.qt6Packages.qt6ct
-    ] else if cfg.platformTheme == "kde" then [
-      pkgs.libsForQt5.plasma-integration
-      pkgs.libsForQt5.systemsettings
-    ] else
-      [ pkgs.libsForQt5.qtstyleplugins ])
+    home.packages = (platformPackages.${cfg.platformTheme} or [ ])
```
9ary, line 205: Nice refactor, nicer than my own attempt at adding the gtk3 platform theme.

9ary (commented):
LGTM overall.

modules/misc/qt.nix
thiagokokada: In general `with` is discouraged at the top level but a scoped one like this usage is generally considered fine.

Also I think this is a by maintainer decision and I am the maintainer of this module, so unless there is a good reason I prefer to keep it as is.

modules/misc/qt.nix
thiagokokada: I don't see a good reason to remove this unless we remove the package from nixpkgs itself.

I also recently added this same option in the NixOS module (see https://github.com/NixOS/nixpkgs/pull/262051), and it would be good to keep for feature parity.

If we eventually remove this from nixpkgs, I think we can remove it from here.

9ary: Agreed.

thiagokokada:
@9ary I backported your fixes from #4312, can you re-review?

modules/misc/qt.nix
9ary: Sorry, that was just a suggestion.

>a scoped one like this usage is generally considered fine

It's more that it gets used like this *in spite* of the pitfalls, because it's convenient. Either way, I'm not here to bikeshed, so it's your call.

modules/misc/qt.nix
```diff
@@ -147,37 +187,31 @@ in {
...
-      [ pkgs.libsForQt5.qtstyleplugins ])
+    systemd.user.sessionVariables = envVars // envVarsExtra;
+
+    home.packages = (platformPackages.${cfg.platformTheme} or [ ])
       ++ lib.optionals (cfg.style.package != null)
       (lib.toList cfg.style.package);
 
-    xsession.importedVariables = [ "QT_QPA_PLATFORMTHEME" ]
+    xsession.importedVariables =
+      [ "QT_QPA_PLATFORMTHEME" "QT_PLUGIN_PATH" "QML2_IMPORT_PATH" ]
```
9ary, line 214: Not familiar with the xsession stuff (I use wayland). Do we want `QT_QPA_PLATFORMTHEME` here if we're not setting it?

9ary (commented):
Looks fine.

For the gtk3 platform theme, like I said in #4312, I'm not sure if there are any extra dependencies that aren't automatically pulled in.
Setting the variable is all I need on my setup, but I'm not sure how to test this in isolation.

modules/misc/qt.nix
thiagokokada: I think there is no harm setting it if it is not set, but not sure either.

9ary: Oddly enough that option doesn't show up in the docs, but as far as I can tell it just passes the variable name to `systemctl --user import-environment` in ~/.xsession.

Looking at [the man page](https://www.freedesktop.org/software/systemd/man/latest/systemctl.html#%0A%20%20%20%20%20%20%20%20%20%20%20%20import-environment%0A%20%20%20%20%20%20%20%20%20%20%20%20VARIABLE%E2%80%A6%0A%20%20%20%20%20%20%20%20%20%20):
>Import all, one or more environment variables set on the client into the systemd manager environment block. If a list of environment variable names is passed, client-side values are then imported into the manager's environment block. If any names are not valid environment variable names or have invalid values according to the rules described above, an error is raised.

I'm not entirely sure how to interpret this, what constitutes an invalid variable name?

Empirically:
```
$ systemctl --user import-environment arstarstarstarst
Environment variable $arstarstarstarst not set, ignoring.
```
It just prints a warning and exits with status code 0, so I think it doesn't strictly matter, but I would still gate this to suppress the warning.

thiagokokada: Done.

9ary (approved):

Aleksanaa:
lgtm

thiagokokada:
CC @ncfavier @rycee, can we have this merged?

rycee:
Thanks! I've merged this now 🙂

acristoffers:
Having `qt.enable = true;` in my config is making it fail with 

```
error:
       … while evaluating a branch condition

         at /nix/store/rj6pqzx66pliiwdfk2wjnnlk40rm3mpd-source/lib/lists.nix:57:9:

           56|       fold' = n:
           57|         if n == len
             |         ^
           58|         then nul

       … while calling the 'length' builtin

         at /nix/store/rj6pqzx66pliiwdfk2wjnnlk40rm3mpd-source/lib/lists.nix:55:13:

           54|     let
           55|       len = length list;
             |             ^
           56|       fold' = n:

       error: value is null while a string was expected
```

using a `home.nix` as simple as

```nix
{ config, inputs, pkgs, username, isLinux, ... }:

let
  homeDirectory = if isLinux then "/home/" + username else "/Users/alan";
in
{
  programs.home-manager.enable = true;
  home.stateVersion = "22.11";
  home.username = username;
  home.homeDirectory = homeDirectory;

  qt.enable = true;
}
```

The enterity of the original file and the flake are in my dotfiles repo: https://github.com/acristoffers/dotfiles

acristoffers:
When using `--show-trace` the last error is

```
       … while calling anonymous lambda

         at /nix/store/f4gw9kx4xyjs1i1jxnlr077v7l6lavlh-source/modules/misc/qt.nix:161:35:

          160|     # Necessary because home.sessionVariables doesn't support mkIf
          161|     envVars = lib.filterAttrs (n: v: v != null) {
             |                                   ^
          162|       QT_QPA_PLATFORMTHEME =

       error: value is null while a string was expected
```

thiagokokada:
> When using `--show-trace` the last error is
> 
> ```
>        … while calling anonymous lambda
> 
>          at /nix/store/f4gw9kx4xyjs1i1jxnlr077v7l6lavlh-source/modules/misc/qt.nix:161:35:
> 
>           160|     # Necessary because home.sessionVariables doesn't support mkIf
>           161|     envVars = lib.filterAttrs (n: v: v != null) {
>              |                                   ^
>           162|       QT_QPA_PLATFORMTHEME =
> 
>        error: value is null while a string was expected
> ```

Will take a look later.

Issue #3324, opened 2022-10-12
Prepending directories to $PATH


ralismark:
### Description

Currently, the entries in `home.sessionPath` are only appended to $PATH, and there is no way to add entries before the existing path. My main issue with this is that I'm unable to override binaries that are installed from other places.

Adding packages to `home.packages` with the appropriate contents isn't an option if the contents of the directory I want to add to PATH is allowed to change without requiring a `home-manager switch`.

I see there's the following closed PRs:

- https://github.com/nix-community/home-manager/pull/1752
- https://github.com/nix-community/home-manager/pull/1769

The former was closed as the latter supersedes it, but the latter became stale and was eventually closed, and I'm kinda wondering if that's intentional or not?

ralismark:
I had a go at this and came up with this. It's not the full dag-based solution, and I haven't extensively tested it, but at least it seems decently backwards compatible with the existing definition of `home.sessionPath`

```nix
{ config, lib, pkgs, ... }:
with lib;
let
  cfg = config.home;
in {
  options.home = {
    sessionPath2 = mkOption {
      type = with types; listOf str;
      default = [ ];
      description = ''
        Extra directories to add to <envar>PATH</envar>.

        (rest of description)
      '';
    };
  };

  config = {
    # Include inherited $PATH people can mkBefore on it
    home.sessionPath2 = [ "$PATH" ];

    home.packages = [
      (
        pkgs.writeTextFile {
          name = "path.sh";
          destination = "/etc/profile.d/path.sh";
          text = let
            replaceMatch = match: lib.replaceStrings
              [ ":\$${match}:" ":\$${match}" "\$${match}:" ]
              [ "\${${match}:+:\$${match}:}" "\${${match}:+:\$${match}}" "\${${match}:+\$${match}:}" ];
            pathvar = replaceMatch "PATH" (lib.concatStringsSep ":" cfg.sessionPath2);
          in lib.optionalString (cfg.sessionPath2 != [ "$PATH" ]) ''
            export PATH="${pathvar}"
          '';
        }
      )
    ];
  };
}
```
If this kind of solution seems good I'm happy to clean this up and open a PR!

srid:
On macOS, we need this to be able to put `git` on $PATH ahead of `/usr/bin` (which has Apple's own git binary). Which in turn is necessary for programs like `lazygit` to work:

```
❯ lg
[...]

PR #4582, opened 2023-10-18
home-environment: add home.sessionSearchVariables

thiagokokada:
### Description

This PR introduces `home.sessionSearchVariables` option, that is created to be a "generic" version of `home.sessionPath` for any environment variables that is similar to PATH (e.g.: MANPATH). This allows composition of those variables between multiple modules, avoiding issues like this one:

https://github.com/nix-community/home-manager/pull/4579/files#r1364374048

This PR also reimplements `home.sessionPath` as terms of `home.sessionSearchVariables`, to reduce code duplication and show that the code is correct, and migrates the `im/fcitx5` module to use it for `QT_PLUGIN_PATH`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/home-environment.nix
```diff
@@ -555,11 +577,17 @@ in
...
+      '' + lib.concatStringsSep "\n"
+        (lib.mapAttrsToList
+          (env: values: ''
+            export ${env}="''$${env}''${${env}:+:}${lib.concatStringsSep ":" values}"'')
+          cfg.sessionSearchVariables)
+        + cfg.sessionVariablesExtra;
     };
 
+    home.sessionSearchVariables.PATH =
+      lib.optionals (cfg.sessionPath != [ ]) cfg.sessionPath;
```
thiagokokada, line 589: Tested by this and it is still passing locally: https://github.com/nix-community/home-manager/pull/4582/files#diff-cdf6113d7d7a7dd00dc882684b388b8a0e537bb6488d28c7e736a8eccf6b9186R2

thiagokokada:
CC @ncfavier @rycee.

thiagokokada (commented):
Did a grep in the code for places where we were using the `home.sessionVariables` to set `PATH`-like environment variables and migrate the ones that I found in this PR.

thiagokokada:
> Did a grep in the code for places where we were using the `home.sessionVariables` to set `PATH`-like environment variables and migrate the ones that I found in this PR.

Also, keep in mind that since `home.sessionVariables` and `home.sessionSearchVariables` are set in different parts of the code, you can even set the same environment variable on both and it would work correctly (because `home.sessionSearchVariables` is set later and it preserves the current environment variable value).

ambroisie (changes requested):
It might be nice to issue a warning if the same attribute name is present in both `home.sessionVariables` and `home.sessionSearchVariables`: not sure if there is a use case where that would be justified.

ambroisie:
Ah also: shouldn't this _prepend_ to its variables, rather than _append_ to the variables.

Presumably the home configuration would want to override the system configuration: this means it should have its entries _earlier_ than the system ones.

modules/targets/generic-linux.nix
```diff
@@ -48,12 +48,8 @@ in {
...
     # resolving to the Nix store.
     # https://github.com/nix-community/home-manager/pull/2891#issuecomment-1101064521
-    home.sessionVariables = {
-      XCURSOR_PATH = "$XCURSOR_PATH\${XCURSOR_PATH:+:}" + concatStringsSep ":" [
-        "${config.home.profileDirectory}/share/icons"
-        "/usr/share/icons"
-        "/usr/share/pixmaps"
-      ];
+    home.sessionSearchVariables = {
+      XCURSOR_PATH = [ "/usr/share/icons" "/usr/share/pixmaps" ];
```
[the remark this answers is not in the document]
ambroisie: It's configured in `modules/config/home-cursor.nix`.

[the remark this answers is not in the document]
thiagokokada: Covered by https://github.com/nix-community/home-manager/pull/4582/files#diff-b07d2165982345df3fc8a099e50b9fb079b473aac705be21d41a765a71ba0546R139-R142.

You can see that there we used the `lib.mkDefault`, that I assume is because we would overwrite this anyway.

tests/modules/home-environment/session-search-variables.nix
```diff
@@ -0,0 +1,15 @@
...
+  imports = [
+    ({ ... }: { config.home.sessionSearchVariables.TEST = [ "foo" ]; })
+    ({ ... }: { config.home.sessionSearchVariables.TEST = [ "bar" "baz" ]; })
+  ];
+
+  nmt.script = ''
+    hmSessVars=home-path/etc/profile.d/hm-session-vars.sh
+    assertFileExists $hmSessVars
+    assertFileContains $hmSessVars \
+      'export TEST="$TEST''${TEST:+:}bar:baz:foo"'
```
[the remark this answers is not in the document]
thiagokokada: It is the same basic test as `session-path.nix`, so if there is an issue with this test there is an issue in that one too.

thiagokokada:
> Ah also: shouldn't this _prepend_ to its variables, rather than _append_ to the variables.
> 
> Presumably the home configuration would want to override the system configuration: this means it should have its entries _earlier_ than the system ones.

I found very little usage of _prepend_ in Home Manager, while _append_ there are bunch of usages here in Home Manager. This also matches the `home.sessionPath` behavior right now.

tests/modules/home-environment/session-search-variables.nix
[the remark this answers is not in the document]
thiagokokada: Here is the test: https://github.com/nix-community/home-manager/blob/master/tests/modules/home-environment/session-path.nix

thiagokokada:
> Ah also: shouldn't this prepend to its variables, rather than append to the variables.

BTW, to explain why this PR was created: https://github.com/nix-community/home-manager/pull/4579/files#r1364374048.

In this particular case we just want to add more things to the PATH, so the order really doesn't matter but it seems that "append" is a safer approach them "prepend". We could introduce a `home.sessionSearchVariablesPrepend` later that does the "prepend" if there is a valid use case for it, but from the grep I did in the code we use "prepend" very rarely here in Home Manager.

ambroisie:
> I found very little usage of prepend in Home Manager, while append there are bunch of usages here in Home Manager. This also matches the home.sessionPath behavior right now.

I think that's just because everybody copied what came before them, and the first one got it wrong. I do think that for a user configuration, it makes more sense to prepend, even if that might not be what is done currently in existing usage.

> add more things to the PATH, so the order really doesn't matter

`PATH` is precisely one usage where I think we should prepend, even if home-manager was currently (wrongly) appending `sessionPath`: say I add a `~/bins` folder to my `PATH`, those should take precedence over the system-provided executables if there is a name collision.

As an example of configurations where order matters: I'm looking to migrate my own re-implementation of #4031 in my configuration, but I would need this module to _prepend_ the value to make it work properly.

thiagokokada:
> As an example of configurations where order matters: I'm looking to migrate my own re-implementation of https://github.com/nix-community/home-manager/pull/4031 in my configuration, but I would need this module to prepend the value to make it work properly.

Could we talk about prepend in another PR? Should be trivial to implement, but I think changing the behavior to prepend here can potentially break a bunch of stuff.

ambroisie:
> Could we talk about prepend in another PR? Should be trivial to implement, but I think changing the behavior to prepend here can potentially break a bunch of stuff.

I think that it's important to get the API right, as otherwise we have to use a worse name for what is probably the better behaviour.

thiagokokada:
> > Could we talk about prepend in another PR? Should be trivial to implement, but I think changing the behavior to prepend here can potentially break a bunch of stuff.
> 
> I think that it's important to get the API right, as otherwise we have to use a worse name for what is probably the better behaviour.

I think consistency is better than having "better" names for supposedly "better" behavior (that is also not 100% clear, each one has pros and cons).

Let's say we want to introduce a prepend behavior with the "better" name, so we probably want to do the same for the `home.sessionPath` since we want to have a good API. This is the solution we would have:

```
home.sessionPath <- append, to keep the existing behavior
home.sessionPathPrepend
home.sessionSearchVariables <- prepend, because this is the "better" behavior
home.sessionSearchVariablesAppend
```

I would much prefer to have this consistent instead:

```
home.sessionPath <- append, to keep the existing behavior
home.sessionPathPrepend
home.sessionSearchVariables <- append
home.sessionSearchVariablesPrepend
```

thiagokokada:
> I think `sessionPath` should be fixed to prepend its value as well, not make its name worse.

I am assuming people are using this behavior willingly. One example I can think that would affect me is the `sway` module that I used to enable in both NixOS and Home-Manager, because NixOS now have a `enableRealtime` option that sets `cap_sys_nice` capabilities to the binary (and of course this is impossible to do in Home Manager).

ambroisie:
I can see that people are surprised by the behaviour in #3324.

As for your example, and others that would be using both a NixOS and home-manager module: I think the solution is to add a `package` option (if missing) and set `package = pkgs.emptyDirectory`. home-manager modules don't make use of `sessionPath`.

thiagokokada:
> I can see that people are surprised by the behaviour in #3324.

Yes, sure, but it is also clear that people eventually understand this behavior and add their configuration according to this behavior.

I can update the documentation to be more explicitly about the "appending" behavior, but changing silently is really not the way to go.

> As for your example, and others that would be using both a NixOS and home-manager module: I think the solution is to add a `package` option (if missing) and set `package = pkgs.emptyDirectory`. home-manager modules don't make use of `sessionPath`.

I am not saying that there is no workarounds for this, is just that changing the behavior will lead to breakage (and worse in this case is that it would be a silent breakage, you can't really see the breakage until you look at the `ps -cTa` output and look that `sway` is suddenly running without the RR scheduler). And of course, this is only one example.

alarsyo:
For what it's worth I agree with @ambroisie on getting the behavior right (which I agree should be prepend, for a path) the first time around, rather than merge this quickly and then have to eventually deal with *more* breakage in another PR

thiagokokada:
> For what it's worth I agree with @ambroisie on getting the behavior right (which I agree should be prepend, for a path) the first time around, rather than merge this quickly and then have to eventually deal with _more_ breakage in another PR

There would be no breakage to merge this PR in the current form, while changing the behavior would lead to breakage. Unless I keep the `home.sessionPath` untouched, that is also fine, we can have this one have prepend while the `home.sessionPath` uses append.

I just find this confusing and unpredictable, and think it is better to have a prepend version of both and keep this one as append behavior as it currently is.

Anyway, I would like input from some maintainer. If they think changing the behavior is fine and can't see any issue, I will change the behavior.

modules/misc/qt.nix
```diff
@@ -194,20 +191,9 @@ in {
...
-      # the values in home.sessionVariablesExtra instead.
-      sessionVariablesExtra = ''
-        export QT_PLUGIN_PATH=${envVarsExtra.QT_PLUGIN_PATH}
-        export QML2_IMPORT_PATH=${envVarsExtra.QML2_IMPORT_PATH}
-      '';
+      sessionSearchVariables = envVarsExtra;
     };
 
-    # Apply theming also to apps started by systemd.
-    systemd.user.sessionVariables = envVars // envVarsExtra;
```
thiagokokada, line 209: Removing this since someone pointed that this is already done by `xsession.importedVariables`: https://github.com/nix-community/home-manager/pull/4677#discussion_r1399643300

thiagokokada:
Ok, changed the behavior from append to prepend to see if this get more acceptance.

CC @ambroisie @alarsyo.

ambroisie:
Hey, thanks for listening to feedback :-)

thiagokokada:
> Hey, thanks for listening to feedback :-)

I am still waiting for feedback from the maintainers since they have the final word in this matter.

CC @rycee for review.

modules/misc/qt.nix
```diff
@@ -194,19 +191,14 @@ in {
...
-      '';
+      sessionSearchVariables = envVarsExtra;
     };
 
     # Apply theming also to apps started by systemd.
-    systemd.user.sessionVariables = envVars // envVarsExtra;
+    systemd.user.sessionVariables = envVars // {
+      QT_PLUGIN_PATH = lib.concatStringsSep ":" envVarsExtra.QT_PLUGIN_PATH;
+      QML2_IMPORT_PATH = lib.concatStringsSep ":" envVarsExtra.QML2_IMPORT_PATH;
+    };
```
ambroisie, line 201: This suggests the need for `systemd.user.sessionSearchVariables` to me.

ambroisie (approved):
Small suggestion about `systemd.user.sessionVariables`. Otherwise it looks good.

thiagokokada:
Still an issue.

ambroisie:
Bad bot.

modules/misc/qt.nix
thiagokokada: For a future PR maybe, but not for this one since this already changes too many things.

thiagokokada:
CC @rycee, would be nice to have this merged since this is something that is sorely needed for Home-Manager.

ambroisie:
Still relevant. @rycee friendly ping.

modules/home-environment.nix
```diff
@@ -309,7 +309,7 @@ in
         ".git/safe/../../bin"
       ];
       description = ''
-        Extra directories to add to {env}`PATH`.
+        Extra directories to be prepend to {env}`PATH`.
```
Scrumplex, line 312:
```suggestion
        Extra directories to prepend to {env}`PATH`.
```

Scrumplex (approved):
Changes LGTM

ambroisie:
Still relevant.

thiagokokada:
Replaced by: https://github.com/nix-community/home-manager/pull/6593.

Issue #4580, opened 2023-10-18
bug: Borgmatic yields deprecation warning


randomn4me:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Running the current version of borgmatic (Version 1.8.1) warns the user, that "The repositories option now expects a list of key/value pairs. Lists of strings for this option are deprecated and support will be removed from a future release."

In the current borgmatic module the repositories are just defined as `types.listOf types.str` [[1]](https://github.com/nix-community/home-manager/blob/master/modules/programs/borgmatic.nix#L59). I already created a [commit](https://github.com/randomn4me/home-manager/commit/a1a0ff915c2f0e543992c6a703970b28ca89a89c), which solves this issue, but it is not backwards compatible and yields no warning, as suggested [in the manual](https://nix-community.github.io/home-manager/index.html#sec-guidelines-back-compat). Is there anything else that needs to be done or can I just create a Pull Request.

Cheers,
r4ndom

### Maintainer CC

@DamienCassou

### System information

```markdown
- system: `"x86_64-linux"`
- host os: `Linux 6.5.7, NixOS, 23.11 (Tapir), 23.11.20231011.5e4c2ad`
- multi-user?: `yes`
- sandbox: `yes`
- version: `nix-env (Nix) 2.17.0`
- nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`
```
```

**Comment by rycee** [comment]
I think the software changing its configuration format would be a valid reason to change the options in a backwards-incompatible way.

A possibility would be to change the type from `types.listOf repositoryOption` to `types.listOf (types.either str repositoryOption)` and then issue a warning if a string value is given. Then, after some suitable time, one could remove the `str` alternative.

**Comment by DamienCassou** [comment]
I like @rycee's suggestion to allow both formats make sense because [borgmatic says that labels are optional](https://torsion.org/borgmatic/docs/reference/configuration/):

> A required list of local or remote repositories with paths and optional labels

I would go as far as *not* issuing a warning if a string is given: if labels are optional, a list of paths makes perfect sense to me.

**Comment by rycee** [comment]
Aha, in that case I would suggest not doing a warning and instead make the label have the type `types.nullOr types.str` (with `null` being the default) and include
``` nix
apply = val: if builtins.isString val then { path = val; } else val;
```
in the option.

randomn4me:
[...]

PR #4588, opened 2023-10-20
Prepare borgmatic change

randomn4me:
### Description

This is preparing an upcoming change in borgmatic to make repositories consisting of paths only unusable. This addresses issue #4580 .

This PR:

1. Is backwards compatible. Paths only in the config will be translated to the required `{ path = "path/to/repo"; }`
2. Paths can be given as
```
    {
      path = "path";
      label = "label";
    }
```
3. Paths can be given as just `{ path = "path/to/repo"; }`, however, this resolves to a config file, which contains `null` as label, which is forbidden by `borgmatic`.

### Checklist

```
borgmatic-program-basic-configuration: OK
borgmatic-program-exclude-hm-symlinks: OK
borgmatic-program-exclude-hm-symlinks-nothing-else: OK
borgmatic-program-include-hm-symlinks: OK
borgmatic-service-basic-configuration: OK
```

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/borgmatic.nix
+++ b/modules/programs/borgmatic.nix
@@ -13,6 +13,13 @@ let
       default = null;
     });
 
+  cleanRepositories = repos:
+    map (repo:
+      if builtins.isString repo then {
+        path = repo;
+      } else
+        removeNullValues repo) repos;
+
   mkRetentionOption = frequency:
     mkNullableOption {
       type = types.int;
@@ -25,6 +32,26 @@ let
     type = yamlFormat.type;
     default = { };
     description = "Extra settings.";
+  };
+
+  repositoryOption = types.submodule {
+    options = {
+      path = mkOption {
+        type = types.str;
+        example = "ssh://myuser@myrepo.myserver.com/./repo";
+        description = "Path of the repository.";
+      };
+
+      label = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        example = "remote";
+        description = ''
+          Short text describing the repository. Can be used with the
+          `--repository` flag to select a repository.
+        '';
+      };
+    };
   };
 
   consistencyCheckModule = types.submodule {
@@ -56,10 +83,23 @@ let
         };
 
         repositories = mkOption {
-          type = types.listOf types.str;
-          description = "Paths to repositories.";
-          example =
-            literalExpression ''["ssh://myuser@myrepo.myserver.com/./repo"]'';
+          type = types.listOf (types.either types.str repositoryOption);
+          apply = cleanRepositories;
+          example = literalExpression ''
+            [
+              {
+                "path" = "ssh://myuser@myrepo.myserver.com/./repo";
+                "label" = "server";
+              }
+              {
+                "path" = "/var/lib/backups/local.borg";
+                "label" = "local";
+              }
+            ]
+          '';
+          description = ''
+            List of local or remote repositories with paths and optional labels.
+          '';
         };
 
         excludeHomeManagerSymlinks = mkOption {
--- a/tests/modules/programs/borgmatic/basic-configuration.nix
+++ b/tests/modules/programs/borgmatic/basic-configuration.nix
@@ -12,7 +12,14 @@ in {
       main = {
         location = {
           sourceDirectories = [ "/my-stuff-to-backup" ];
-          repositories = [ "/mnt/disk1" "/mnt/disk2" ];
+          repositories = [
+            "/mnt/disk1"
+            { path = "/mnt/disk2"; }
+            {
+              path = "/mnt/disk3";
+              label = "disk3";
+            }
+          ];
           extraConfig = {
             one_file_system = true;
             exclude_patterns = [ "*.swp" ];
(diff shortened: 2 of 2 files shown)


rycee:
Thanks for the contribution! Looks OK overall, I would suggest adding an [`apply`](https://nixos.org/manual/nixpkgs/unstable/#function-library-lib.options.mkOption) attribute to the option to convert string values into attribute set ones. If you search the HM repo for "apply =" you should find a number of examples.

To complete the checklist you need to squash the commits into a single one having a commit message matching the given format. No need to do immediately, you can do this once the PR is close to being merged. Also you need to add a test. Can have a simple test with three repositories, one a plain string, one an attribute set with only a path, and one an attribute set with both path and label.

Maintainer CC @DamienCassou

modules/programs/borgmatic.nix
```diff
@@ -13,6 +13,9 @@ let
       default = null;
     });
 
+  cleanRepositories = repos:
+    map (repo: if builtins.isString repo then { path = repo; } else repo) repos;
```
DamienCassou, line 17:
```suggestion
    map (repo: if builtins.isString repo then { path = repo; } else removeNullValues repo) repos;
```

modules/programs/borgmatic.nix
```diff
@@ -27,6 +30,23 @@ let
...
+      path = mkOption {
+        type = types.str;
+        description = "Path of the repository";
+        example = "ssh://myuser@myrepo.myserver.com/./repo";
+      };
+
+      label = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        description = "Label of the repository";
```
DamienCassou, line 44: I would give a little more info here. Maybe something like

```suggestion
        description = "Short text describing the repository. Can be used with the `--repository` flag to select a repository";
```

modules/programs/borgmatic.nix
```diff
@@ -56,10 +76,20 @@ let
         };
 
         repositories = mkOption {
-          type = types.listOf types.str;
-          description = "Paths to repositories.";
-          example =
-            literalExpression ''["ssh://myuser@myrepo.myserver.com/./repo"]'';
+          type = types.listOf (types.either types.str repositoryOption);
+          description = "Paths to key/value pairs of repositories.";
```
DamienCassou, line 80: "Paths to key/value pairs" doesn't mean anything to me. Maybe something like:

```suggestion
          description = "List of local or remote repositories with paths and optional labels.";
```

DamienCassou (changes requested):
Great job, thank you!

randomn4me:
@rycee I adjusted the testcases (3 now: 1 str, 1 `{ path="path"}`, 1 `{path="path", label="label"}`) and added @DamienCassou 's commits. If it is now finished I need to find out how to squash and then finish this PR.

modules/programs/borgmatic.nix
```diff
@@ -56,10 +81,21 @@ let
         };
 
         repositories = mkOption {
-          type = types.listOf types.str;
-          description = "Paths to repositories.";
-          example =
-            literalExpression ''["ssh://myuser@myrepo.myserver.com/./repo"]'';
+          type = types.listOf (types.either types.str repositoryOption);
```
DamienCassou, line 84: @rycee suggested to use `apply` here instead of deferring cleaning until much later in the file:

```suggestion
          type = types.listOf (types.either types.str repositoryOption);
          apply = cleanRepositories;
```

modules/programs/borgmatic.nix
```diff
@@ -154,7 +190,7 @@ let
   writeConfig = config:
     generators.toYAML { } (removeNullValues ({
       source_directories = config.location.sourceDirectories;
-      repositories = config.location.repositories;
+      repositories = cleanRepositories config.location.repositories;
```
DamienCassou, line 193: this should no be necessary if using `apply` as suggested

tests/modules/programs/borgmatic/basic-configuration.nix
```diff
@@ -12,7 +12,14 @@ in {
...
           sourceDirectories = [ "/my-stuff-to-backup" ];
-          repositories = [ "/mnt/disk1" "/mnt/disk2" ];
+          repositories = [
+            "/mnt/disk1"
+            { path = "/mnt/disk2"; }
+            {
+              path = "/mnt/disk3";
+              label = "disk3";
+            }
+          ];
```
DamienCassou, line 22: assertions much be changed accordingly. You should replace:

```
    expectations[repositories[0]]="${
      builtins.elemAt backups.main.location.repositories 0
    }"
    expectations[repositories[1]]="${
      builtins.elemAt backups.main.location.repositories 1
    }"
```

with:

```
    expectations[repositories[0].path]="${
      (builtins.elemAt backups.main.location.repositories 0).path
    }"
    expectations[repositories[1].path]="${
      (builtins.elemAt backups.main.location.repositories 1).path
    }"
    expectations[repositories[2].path]="${
      (builtins.elemAt backups.main.location.repositories 2).path
    }"
    expectations[repositories[2].label]="${
      (builtins.elemAt backups.main.location.repositories 2).label
    }"
```

DamienCassou (commented):
Thank you very much. I think there are 2 more little changes to do and we will be good to merge.

modules/programs/borgmatic.nix
randomn4me: I'll remove it

randomn4me:
It should all be done now. If there is more that needs to be done before a PR, just ping me.

DamienCassou:
The result is perfect!

rycee:

Issue #4593, opened 2023-10-21
bug: Home-manager profile failed to activate when use nix with use-xdg-base-directories


ayum:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

I am using nix on Debian with use-xdg-base-directories set to true. In this configuration there is no `$HOME/.nix-profile`.

But activate script of home-manager detect right install command with `if [[ -e $HOME/.nix-profile/manifest.json ]] ;`. Therefore instead of correct `nix profile install` command `nix-env -i` used and activate failed due to the fact that nix-env cannot be used after `nix profile install` once used on profile.

### Maintainer CC

@rycee

### System information

```markdown
this path will be fetched (0.00 MiB download, 0.00 MiB unpacked):
  /nix/store/fyvbj5mvrvdnvhbx6hc977wsgl33pk3i-nix-info
copying path '/nix/store/fyvbj5mvrvdnvhbx6hc977wsgl33pk3i-nix-info' from 'https://cache.nixos.org'...
 - system: `"x86_64-linux"`
 - host os: `Linux 6.1.0-9-amd64, Debian GNU/Linux, 12 (bookworm), nobuild`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.18.1`
 - nixpkgs: `/nix/store/ki98w9ah1k5p62mfv5pjbb58szndc2g5-source`
```

ayum:
I have created PR to illustrate my understanding of the issue.

https://github.com/nix-community/home-manager/pull/4594

PR #4594, opened 2023-10-21
home-manager: fix activate script for use-xdg-base-directories use case

ayum:
### Description

This is illustrative PR for issue https://github.com/nix-community/home-manager/issues/4593

Cannot assure that this is fixing the issue, but for illustrative purpose.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/home-environment.nix
+++ b/modules/home-environment.nix
@@ -207,6 +207,7 @@ in
       type = types.path;
       defaultText = literalExpression ''
         "''${home.homeDirectory}/.nix-profile"  or
+        "''${home.homeDirectory}/.local/state/nix/profile"  or
         "/etc/profiles/per-user/''${home.username}"
       '';
       readOnly = true;
@@ -585,7 +586,7 @@ in
       if config.submoduleSupport.externalPackageInstall
       then
         ''
-          if [[ -e $HOME/.nix-profile/manifest.json ]] ; then
+          if [[ -e $HOME/.nix-profile/manifest.json ]] || [[ -e $HOME/.local/state/nix/profile/manifest.json ]] ; then
             nix profile list \
               | { grep 'home-manager-path$' || test $? = 1; } \
               | cut -d ' ' -f 4 \
@@ -623,7 +624,7 @@ in
             $DRY_RUN_CMD $oldNix profile install $1
           }
 
-          if [[ -e $HOME/.nix-profile/manifest.json ]] ; then
+          if [[ -e $HOME/.nix-profile/manifest.json ]] || [[ -e $HOME/.local/state/nix/profile/manifest.json ]] ; then
             INSTALL_CMD="nix profile install"
             INSTALL_CMD_ACTUAL="nixReplaceProfile"
             LIST_CMD="nix profile list"
--- a/modules/home-environment.nix
+++ b/modules/home-environment.nix
@@ -207,7 +207,7 @@ in
       type = types.path;
       defaultText = literalExpression ''
         "''${home.homeDirectory}/.nix-profile"  or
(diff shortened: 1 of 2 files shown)


teto:
thanks for the PR but I think this should use $XDG_ variables instead of $HOME/.local/state etc otherwise it defeats the whole point of xdg

ayum:
> thanks for the PR but I think this should use $XDG_ variables instead of $HOME/.local/state etc otherwise it defeats the whole point of xdg

While i could agree, but corresponding XDG vars may be not defined (and in fact in my configuration they are not defined). May be checks for right command should be even more elaborated.

modules/home-environment.nix
```diff
@@ -585,7 +586,7 @@ in
       if config.submoduleSupport.externalPackageInstall
       then
         ''
-          if [[ -e $HOME/.nix-profile/manifest.json ]] ; then
+          if [[ -e $HOME/.nix-profile/manifest.json ]] || [[ -e $HOME/.local/state/nix/profile/manifest.json ]] ; then
```
rycee, line 589: Can use `config.xdg.stateHome`?

```suggestion
          if [[ -e $HOME/.nix-profile/manifest.json || -e "${config.xdg.stateHome}/nix/profile/manifest.json" ]] ; then
```

rycee:
Like @teto said, you need to use the corresponding XDG variable. For example, `${XDG_STATE_HOME:-$HOME/.local/state}`.

ayum:
@teto @rycee 

I have updated PR as you say.

But i can't verify that PR fixes issue. When i do `nix run github:ayum/nix-home-manager/1861d4b3cc2cdd04268bfab5818cf04e936e0214 -- build` (latest PR commit) i cannot see PR changes in `result/activate`. If you guide me how can i test it i will verify PR.

ayum:
@teto @rycee 

I have managed to test this PR and can confirm that it fixes the issue. At least in my configuration with  `use-xdg-base-directories` enabled and no `.nix-profile` in home dir. Home manager profile successfully activates.

Please review changes and approve if there are no more issues with it.

Arian04:
unless I'm misunderstanding the purpose of this PR, wouldn't it be better to just merge #4422?

ayum:
> unless I'm misunderstanding the purpose of this PR, wouldn't it be better to just merge #4422?

@Arian04 
Ah, not seeing it. BTW, #4422 not working for me. I suppose due to line https://github.com/nix-community/home-manager/pull/4594/files#diff-5c980e4c769c045b39cdf1cf9597e9cb2d45929f77e106ee8d9abbcdd7b501a0R210 present in my PR and absent in #4422 .

Aside from this little change #4422 is superior as it touch help message also. If @e-nikolov changes a little that PR i would be totally in favour for it over my PR.

ayum:
@Arian04 @e-nikolov 

I believe that `profileDirectory` not correctly detected in both these PRs.

xdg state home based directory will be used only when we have `nix.enable = true`, nix.settings.use-xdg-base-directories = true`, `xdg.enable = true`, `nix.package = ...` all set.

But it is overly excessive to require that. Nix staff surely have not be required as we may want not manage nix configuration with home manager (my case, i have system wide config, as determinate installer for nix set it for me).

Even xdg stuff should not be required as it is configuration decision. But issue that require fix is about correct detection and profile activation al any possible configuration and it must not depends on user configuration choices.

ayum:
While in current state PR is fixing (narrowly speaking) issue, but wider issue persists. All modules that use `home.profileDirectory` will be somewhat broken unless `.nix-profile` exists in home dir.

More correct detection of profile directory needed.

PR #4595, opened 2023-10-21
eww: make `configDir` setting optional

trueNAHO:
### Description

Manually configuring `eww` with `xdg.configFile."eww".text` in order to leverage Nix' capabilities, is inconvenient due to the required `xdg.configFile."eww".source` symlink.

This change makes the symlink optional.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@mainrs

### DIFF
--- a/modules/programs/eww.nix
+++ b/modules/programs/eww.nix
@@ -23,7 +23,8 @@ in {
     };
 
     configDir = mkOption {
-      type = types.path;
+      type = types.nullOr types.path;
+      default = null;
       example = literalExpression "./eww-config-dir";
       description = ''
         The directory that gets symlinked to
@@ -32,8 +33,11 @@ in {
     };
   };
 
-  config = mkIf cfg.enable {
-    home.packages = [ cfg.package ];
-    xdg.configFile."eww".source = cfg.configDir;
-  };
+  config = mkIf cfg.enable (mkMerge [
+    { home.packages = [ cfg.package ]; }
+
+    (mkIf (cfg.configDir != null) {
+      xdg.configFile."eww".source = cfg.configDir;
+    })
+  ]);
 }


rycee:
Sorry about the delay. Honestly I'm not sure what the module adds. Especially if you also can disable the `configDir` option. Then the module is literally the same as `home.packages = [ pkgs.eww ]`, which I think is just silly.

I'm reluctant to merge this because of that. Perhaps there is some way to expand the module so that it provides some more value and also cover your use-case?

trueNAHO:
> Sorry about the delay.

No worries!

> Honestly I'm not sure what the module adds. Especially if you also can disable the `configDir` option. Then the module is literally the same as `home.packages = [ pkgs.eww ]`, which I think is just silly.
>
> I'm reluctant to merge this because of that.

I see your point. Maybe once the module gets expanded on in the future, it would be more reasonable.

> Perhaps there is some way to expand the module so that it provides some more value and also cover your use-case?

Considering that I still did not get around to properly migrate my previous Arch Linux `eww` configuration to Nix, I prototyped using `home.packages = [ pkgs.eww ]` in the meantime. However, for now, I am not using any widget bars (`eww` alternatives).

The Home Manager module could provide options to declare the various files `eww` heavily relies, like `*.sh` (`pkgs.writeShellApplication`) and `*.scss`. Otherwise, we could leave those implementations up to the user.

Once I am migrating my `eww` configuration to Nix, I might be able to better judge what options are beneficial. In the meantime, feel free to close/stale this PR until I pick it up again. For reference, my `eww` migration might only happen in a few months from now.

mainrs:
There is #5645, which adds shell completions and makes a module worthwhile. However, there is no consensus on how to manage the config. There is both this issue and #5585, which is exactly the opposite.

I chose the folder because it allows for an easy migration from an existing eww configuration.

Even then one call specifically install the binaries via `home.packages` that custom eww scripts require. And the config is still deterministic with the folder approach.

I'm not against changing this. I only think the solution we choose should be something that is easy to use without writing too much boilerplate nix code.

Eveeifyeve:
Superseeded by https://github.com/nix-community/home-manager/pull/5979

PR #4597, opened 2023-10-22
neomutt: add IMAP support

### LINKED ISSUE #4376: programs.neomutt: Configure imap accounts
### Description
The `programs.neomutt` module currently configures the accounts as configured in `accounts.email.accounts.<name> `, with the exception of imap. It only configures local maildirs, while neomutt itself does have support for imap. So instead of having to run mbsync to sync my mailbox locally, I would like to request that imap can be enabled directly in the mutt module.

Cybolic:
### Description

This adds support for IMAP accounts in NeoMutt as well as IMAP or maildir overrides for individual mailboxes.
Fixes #4376.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@jvanbruegge @rycee @berbiche @michaeladler @sumnerevans


Cybolic:
Currently, this is not backwards compatible, as it defaults to using IMAP for a mailbox whenever possible.
Keeping in mind that NeoMutt can use a mix of maildir and IMAP mailboxes, I'm planning to add opt-in support for IMAP accounts and mailboxes through the following options:

1. `accounts.email.accounts.<name>.neomutt.mailboxType = types.enum [ "maildir" "imap" ]`
Use either `IMAP` or `maildir` as the default for all mailboxes of this account (it would still default to `maildir` to keep compatibility). If no `extraMailBoxes` are set, this would only hardcode the mailbox type for `inbox`.
  
2. `accounts.email.accounts.<name>.neomutt.extraMailBoxes = listOf { mailbox = types.string; name = types.nullOr types.string; type = types.enum [ "maildir" "imap" ]}` (i.e. add a new `type` option to the `extraMailBoxes` type)
Override the default mailbox type for an individual mailbox.

Feedback on this approach is very welcome.

modules/programs/neomutt.nix
```diff
@@ -145,29 +218,40 @@ let
...
+      mailroot = accountRoot account;
+      hookName = if account.neomutt.mailboxType == "imap" then
+        "account-hook"
+      else
+        "folder-hook";
       extraMailboxes = concatMapStringsSep "\n" (extra:
-        if isString extra then
-          ''mailboxes "${account.maildir.absPath}/${extra}"''
+        let
+          mailboxroot = if isString extra == false && extra.type == "imap" then
```
chayleaf, line 228: see my comment on line 13

modules/programs/neomutt.nix
```diff
@@ -145,29 +218,40 @@ let
...
+            accountRootIMAP account
+          else if isString extra == false && extra.type == "maildir" then
```
chayleaf, line 230: same as above

chayleaf (commented):
other than the small nits, I've checked that it works and the old behavior does get preserved without `neomutt.mailboxType = "imap";`.

modules/programs/neomutt.nix
```diff
@@ -101,6 +154,26 @@ let
...
 
+  accountRootIMAP = account:
+    let
+      userName =
+        if account.userName != null then "${account.userName}@" else "";
+      port = if account.imap.port != null then
+        ":${toString account.imap.port}"
+      else
+        "";
+      protocol = if account.imap.tls.enable || account.imap.port == 993 then
```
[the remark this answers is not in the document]
Cybolic: No, it isn't done anywhere else in home-manager, so you're probably right that this should be left out.

Cybolic:
Thanks @chayleaf for those suggestions! This is the first time I use Nix/HM for anything, so I really do appreciate those tweaks.
I've implemented all your suggestions and I do agree that they simplify things.

I've added an extra change since the initial PR, which is to enable `ssl_force_tls` based on the value of `imap.tls` instead of `smtp.tls` since that's the type of connection NeoMutt actually uses the setting for. The tests still pass after this change.

chayleaf (approved):
I'm not a HM contributor, but LGTM! Please note that in nixpkgs commits are typically squashed, and I recommend doing it here to a reasonable extent too (i.e. adding tests in a separate commit is fine, but any fixes should probably be applied to the original commit rather than added in a new commit).

Cybolic:
I just fixed the failing test. Also, just so Github can pick it up, this PR would fix #4376.

PR #4599, opened 2023-10-22
fish: support simple key bindings

arulagrawal:
In fish shell, support simple keybinds of the form "SEQUENCE" -> COMMAND, as described in the fish docs: https://fishshell.com/docs/current/cmds/bind.html

### Description
Support generating simple keybinds for fish shell. The implementation is the same as that for the aliases which existed already, however I am unsure of the `escapeShellArg` function. It's presence would break the 3rd test case (I included this specifically because it is listed as an example in the [docs](https://fishshell.com/docs/current/cmds/bind.html)), but maybe there is a better way to handle this? 

This is my first time contributing to nix stuff so any pointers are appreciated.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -145,6 +145,9 @@ let
   aliasesStr = concatStringsSep "\n"
     (mapAttrsToList (k: v: "alias ${k} ${escapeShellArg v}") cfg.shellAliases);
 
+  bindsStr = concatStringsSep "\n"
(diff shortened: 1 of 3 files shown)


arulagrawal:
I guess this may run afoul of the "Add only valuable options" guideline, and seeing as no one added this small option already, I don't disagree since these simple binds could easily be placed in `interactiveShellInit`.

@rycee (wasn't sure who else to ping)

arulagrawal:
it seems there is no interest in this.

Issue #4486, opened 2023-09-20
Hyprland: module should set recommended environment variables


gekoke:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

The module should (probably?) set `NIXOS_OZONE_WL = "1";`, like [the Hyprland hm-module flake output does](https://github.com/hyprwm/Hyprland/blob/1925e64c21811ce76e5059d7a063f968c2d3e98c/nix/hm-module.nix#L125).

### Maintainer CC

@fufe

### System information

```markdown
not relevant
```

fufexan:
I should do a follow-up on https://github.com/nix-community/home-manager/pull/3837#discussion_r1190961477 to address this.

nyabinary:
> I should do a follow-up on [#3837 (comment)](https://github.com/nix-community/home-manager/pull/3837#discussion_r1190961477) to address this.

any update?

fufexan:
I have not had time to work on it. If anyone wants to do it, here's a NixOS equivalent to this task https://github.com/NixOS/nixpkgs/pull/240989.

PR #4601, opened 2023-10-24
hyprland: add tray.target

andresilva:
### Description

Duplicates the tray.target from xsession module. Same thing as done for sway (https://github.com/nix-community/home-manager/pull/2713).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@fufexan

### DIFF
--- a/modules/services/window-managers/hyprland.nix
+++ b/modules/services/window-managers/hyprland.nix
@@ -256,5 +256,12 @@ in {
         After = [ "graphical-session-pre.target" ];
       };
     };
+
+    systemd.user.targets.tray = {
+      Unit = {
+        Description = "Home Manager System Tray";
+        Requires = [ "graphical-session-pre.target" ];
+      };
+    };
   };
 }


andresilva:
This is needed for services like https://github.com/nix-community/home-manager/blob/master/modules/services/blueman-applet.nix#L33 and https://github.com/nix-community/home-manager/blob/master/modules/services/syncthing.nix#L124.

fufexan (commented):
We should probably have a shared compositor module which enables the tray, exposes a `recommendedEnvirnoment` option (`NIXOS_OZONE_WL`, see #4486) and other such things.

rycee:

@fufexan I agree, it would be nice to have something like a wayland compositor base module that would do such things.

andresilva:
Locally I also have something mimicking some of the functionality of the xsession module, pretty much this https://github.com/nix-community/home-manager/blob/master/modules/xsession.nix#L174-L221.

PR #4604, opened 2023-10-24
granted: add module

wcarlsen:
### Description

Add granted module https://github.com/common-fate/granted in order to simplify setup, so that others don't encounter the issue referenced in this issue https://github.com/NixOS/nixpkgs/issues/258867.

Granted require an alias set to work correct and they don't take nix into account, hence this module solves the basic setup issue.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/granted.nix
+++ b/modules/programs/granted.nix
@@ -1,0 +1,36 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.programs.granted;
+  package = pkgs.granted;
+
+in {
+  meta.maintainers = [ hm.maintainers.wcarlsen ];
+
+  options.programs.granted = {
+    enable = mkEnableOption "granted";
+
+    enableZshIntegration = mkOption {
+      default = true;
+      type = types.bool;
+      description = ''
+        Whether to enable Zsh integration.
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ package ];
+
+    programs.zsh.initExtra = mkIf cfg.enableZshIntegration ''
+      function assume() {
+        export GRANTED_ALIAS_CONFIGURED="true"
+        source ${package}/bin/.assume-wrapped "$@"
+        unset GRANTED_ALIAS_CONFIGURED
+      }
+    '';
+  };
+}
--- a/modules/lib/maintainers.nix
+++ b/modules/lib/maintainers.nix
@@ -448,4 +448,10 @@
     github = "liyangau";
     githubId = 71299093;
   };
+  wcarlsen = {
+    name = "Willi Carlsen";
+    email = "carlsenwilli+nix@gmail.com";
+    github = "wcarlsen";
+    githubId = 17003032;
+  };
 }
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -1288,6 +1288,13 @@ in
           A new module is available: 'programs.cava'.
         '';
       }
+
+      {
+        time = "2023-11-01T21:18:20+00:00";
+        message = ''
+          A new module is available: 'programs.granted'.
+        '';
+      }
     ];
   };
 }
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -99,6 +99,7 @@ let
     ./programs/gnome-terminal.nix
     ./programs/go.nix
     ./programs/gpg.nix
+    ./programs/granted.nix
     ./programs/havoc.nix
     ./programs/helix.nix
     ./programs/hexchat.nix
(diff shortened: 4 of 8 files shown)


modules/programs/granted.nix
```diff
@@ -0,0 +1,35 @@
...
+      description = ''
+        Whether to enable Zsh integration.
+      '';
+    };
+  };
+
+  config = let
+    manualShellProfile = ''
+      alias assume="source ${package}/bin/.assume-wrapped"
+      export GRANTED_ALIAS_CONFIGURED="true"
```
rycee, line 28: Wouldn't this leave a stray `GRANTED_ALIAS_CONFIGURED` in the environment? Perhaps something like

``` patch
diff --git a/modules/programs/granted.nix b/modules/programs/granted.nix
index 1ead35eb..41de1402 100644
--- a/modules/programs/granted.nix
+++ b/modules/programs/granted.nix
@@ -22,14 +22,15 @@ in {
     };
   };
 
-  config = let
-    manualShellProfile = ''
-      alias assume="source ${package}/bin/.assume-wrapped"
-      export GRANTED_ALIAS_CONFIGURED="true"
-    '';
-  in mkIf cfg.enable {
+  config = mkIf cfg.enable {
     home.packages = [ package ];
 
-    programs.zsh.envExtra = mkIf cfg.enableZshIntegration manualShellProfile;
+    programs.zsh.shellAliases = {
+      granted = concatStringsSep ";" [
+        "GRANTED_ALIAS_CONFIGURED='true'"
+        "source ${package}/bin/.assume-wrapped"
+        "unset GRANTED_ALIAS_CONFIGURED"
+      ];
+    };
   };
 }
```

could work?

modules/programs/granted.nix
```diff
@@ -0,0 +1,35 @@
...
+      type = types.bool;
+      description = ''
+        Whether to enable Zsh integration.
+      '';
+    };
+  };
+
+  config = let
+    manualShellProfile = ''
+      alias assume="source ${package}/bin/.assume-wrapped"
```
rycee, line 27: This honestly seems like a packaging bug to me. If there is no hope to run the assume script as an actual script then it shouldn't be in `/bin` at all, instead something like `/share/bash/assume.sh` would be more suitable. The file could be patched to unconditionally set (and unset) the `GRANTED_ALIAS_CONFIGURED` variable so the user doesn't have to worry about that.

Perhaps could place a script in `/bin/assume` that contains something like

``` sh
cat > /dev/stderr <<EOF
It is not possible to run 'assume' directly, you must create a shell alias.

Bash:  alias assume='{nix profile path}/share/bash/assume.sh'
Zsh:    alias assume='{nix profile path}/share/bash/assume.sh'
Fish:    …
EOF

exit 1
```

When the user has created an alias, then I think all shells would run the alias over the above assume script.

modules/programs/granted.nix
wcarlsen: I agree that this is a better solution than mine. I will test it out to see if it works.

modules/programs/granted.nix
wcarlsen: Packaging is not ideal from granted and I totally agree about the `/bin` part. I think the alias with source comes into play since because otherwise the cannot set environment variables for an AWS session in you current shell. Unfortunately I'm not skilled enough in Nix yet to do advanced stuff in packaging.

modules/programs/granted.nix
wcarlsen: Tested and it works beautifully.

wcarlsen: Yeah found a bug with the above solution this morning. Any parameters passed to to assume won't be propagated correct.

rycee: Yeah, that's a good point. I wonder if not a function would be more appropriate? :-D

``` zsh
programs.zsh.initExtra = ''
  function assume() {
    export GRANTED_ALIAS_CONFIGURED='true'
    source ${package}/bin/.assume-wrapped "$@"
    unset GRANTED_ALIAS_CONFIGURED
  }
'';
```

The [yazi module](https://github.com/nix-community/home-manager/blob/09587fbbc6a669f7725613e044c2577dc5d43ab5/modules/programs/yazi.nix) does something like this already.

wcarlsen: I think you are right. I will tweak and test ✌️

wcarlsen: I got to test it  and bundling the functionality into a function does the trick. Thanks for the valuable input @rycee .

modules/programs/granted.nix
wcarlsen: @rycee I have hard time resolving this one, since it seems to be on the packaging side of things. Anything I can do to progress?

rycee: Yeah, my comment here is only for the package in Nixpkgs. Not relevant for this module (except that the module needs to be updated in case the package in Nixpkgs is updated in the way I suggest above).

rycee:
Great job! Merged to master now 🙂

PR #4606, opened 2023-10-24
hyprland: allow setting finalPackage

andresilva:
### Description

The current implementation assumes that the given package is a proper Hyprland package since it tries to override `enableXWayland`, `enableNvidiaPatches`. This means that you can't use a wrapper for Hyprland as a package. This PR removes the `readOnly` from `finalPackage` so that it can be set by the end-user.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@fufexan

### DIFF
--- a/modules/services/window-managers/hyprland.nix
+++ b/modules/services/window-managers/hyprland.nix
@@ -44,7 +44,6 @@ in {
 
     finalPackage = lib.mkOption {
       type = lib.types.package;
-      readOnly = true;
       default = cfg.package.override {
         enableXWayland = cfg.xwayland.enable;
         enableNvidiaPatches = cfg.enableNvidiaPatches;


fufexan:
Here's a simpler version: remove the `readOnly = true;` key from the `finalPackage` option definition, and you can then set `finalPackage = someHyprlandWrapper;`.

andresilva:
Sure, I'll do that instead.

andresilva:
Updated.

fufexan (approved):

rycee:
I'm not certain what the use-case is here, why not use the `package` option?

Is it just to get around the need for the enableXWayland and enableNvidiaPatches parameters? If so, I would suggest just adding those to your wrapper as well. Something like

``` nix
wayland.windowManager.hyprland.package = pkgs.makeOverridable (
  { enableXWayland, enableNvidiaPatches }:
    # your wrapper derivation goes here
);
```

perhaps.

I'm very reluctant to make the finalPackage option non-read-only since it would break the existing pattern of how these options are used.

andresilva:
@rycee https://github.com/nix-community/home-manager/commit/ce4382235515fbecbc52ee35fdcdfc010ef058a4 this is the original commit I opened the PR with, I then changed it per @fufexan's request. Here's the original PR description as well:

The current implementation assumes that the given package is a proper Hyprland package since it tries to override enableXWayland, enableNvidiaPatches. This means that you can't use a wrapper for Hyprland as a package. I changed the behavior of the module so that the overriding is only done for the default package, this is similar to what is done on the sway module (https://github.com/nix-community/home-manager/blob/master/modules/services/window-managers/i3-sway/sway.nix#L339-L350).

andresilva:
@rycee your suggestion with `makeOverridable` doesn't seem to work due to type mismatch, i.e. it is not a package.

rycee:
@andresilva Ok, I like the original solution with only the default package being configurable slightly better. But I think the ideal is still to keep a finalPackage option and instead make the package compatible. The makeOverridable command should work but the inner package really needs to be a package. If you try to use a path literal then it can be copied over with `runCommand`. Something like

``` nix
pkgs.makeOverridable (
  {enableXWayland ? true, enableNvidiaPatches ? false}:
    pkgs.runCommand "hyprland-wrapper" {} "cp ${./my-wrapper} $out"
) {}
```

may work.

andresilva:
> @andresilva Ok, I like the original solution with only the default package being configurable slightly better. But I think the ideal is still to keep a finalPackage option and instead make the package compatible. The makeOverridable command should work but the inner package really needs to be a package. If you try to use a path literal then it can be copied over with `runCommand`. Something like
> 
> ```nix
> pkgs.makeOverridable (
>   {enableXWayland ? true, enableNvidiaPatches ? false}:
>     pkgs.runCommand "hyprland-wrapper" {} "cp ${./my-wrapper} $out"
> ) {}
> ```
> 
> may work.

This works, I was just missing the `{}` at the end. I also learned about `makeOverridable` so thanks for that! :) So I guess we can close the PR and keep things as they are right now?

rycee:
What we could do to make things a bit more convenient would be to introduce some form of function that could be used. Something like

``` nix
wayland.windowManager.hyprland.package = config.hyprland.mkWrapper ./my-wrapper;
```

but I'm not certain if this is the best form for this function. Will have to think about it a bit.

In any case, glad it worked out for you 🙂

PR #4607, opened 2023-10-24
sxhkd: set scope OOMPolicy to continue

ncfavier:
When a process inside the sxhkd scope is OOM killed, if the OOM policy is set to `stop` then the sxhkd scope itself will exit, terminating every process launched from the keyboard.

This is undesirable, set it to `continue` instead to keep other processes running.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@0qq


rycee:
Looks good to me! Merged now 🙂

PR #4610, opened 2023-10-26
aerc: fix config paths on darwin

kmaasrud:
### Description

Fixes regression after #4550.

When `xdg.enable` is set, aerc uses the `$XDG_CONFIG_HOME` variable instead of the default `$HOME/Library/Preferences`. home-manager needs to check this to make sure aerc can find the proper configuration directory in both cases.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@lukasngl


modules/programs/aerc.nix
```diff
@@ -1,5 +1,4 @@
 { config, lib, pkgs, ... }:
-
```
rycee, line 2: These seem like unrelated changes. I would prefer to have a targeted commit with just the necessary changes. These cosmetic changes will just make `git blame` more difficult. Would you mind amending the commit?

rycee: Although, I do agree that the unnecessary parentheses are a bit annoying 🙂

rycee:
Thanks for the contribution! There is always something about configuration paths on mac 😖

modules/programs/aerc.nix
kmaasrud: Definitely! These changes stemmed from using the Alejandra Nix formatter which is setup in my editor, so I didn't notice before committing :sweat_smile: Rolling back now!

rycee:
Cool, thanks for the fix! Merged to master now.

Issue #4613, opened 2023-10-28
bug: generators: toKDL


cgahr:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

The following nix is parsed incorrectly to KDL:
```
test [
  "unbind \"a\""
  "unbind \"b\""
];
```
is parsed to
```kdl
test unbind "a" unbind "b"
```
which, in my understanding, is not correct.

### Maintainer CC

@h7x4 @antholeole

### System information

```markdown
not relevant
```

cgahr:
Similarly, the following config is not parsed correctly:
```nix
pane = {
  "bind \"t\"" = {
    CloseTab = [];
    CloseFocus = [];
  };
};
It results in
```kdl
pane {
  bind "t" {
    CloseFocus 
    CloseTab 
  }
}
```
since the name are being sorted.

``

```

9p4:
Related: https://github.com/NixOS/nixpkgs/issues/198655

https://github.com/AgathaSorceress/json2kdl

Issue #4485, opened 2023-09-20
feature request: zellij add support for layouts


colemickens:
### Description

I want to adopt `zjstatus`: https://github.com/dj95/zjstatus

It looks like that entails including the plugin in a new custom layout file in `layouts/` . Then launching with that layout or setting `default_layout` in the config.

I can import the KDL converter and use xdgConfigFile to do this by hand, but it would be nice to have module options for it.

colemickens:
example usage: https://github.com/colemickens/nixcfg/commit/90914d2dd8cbecda7f785f2bb833d55c3887347b

but I can't really figure out if there's even a way to express that KDL in Nix...

kacper-uminski:
The proposal by @jaythomas is reasonable, IMO. One modification I would make is to use a `.text` field or some such, so that the
actual layout field is kept as a set. That way we keep extensibility without needing a breaking change once someone makes a decent way to do this in pure nix. That way we could also have a separate field for importing files.

jaythomas:
@kacper-uminski could you provide a modification to my example with your suggestion so I could better understand. I've developed a little more understanding of home manager and the nix language and may as well get a working proposal implemented. I already looked at the KDL utility functions and discovered they're broken implementation, so I'm leaning even more towards the idea of first supporting a pure text value but leaving room in the implementation to pass a native nix attribute set that outputs the appropriate KDL files.

kacper-uminski:
Gladly!

I am thinking something along these lines, though feel free to modify as you see fit:
```nix
{...}:
{
  programs.zellij.layouts = {
    my-layout-A = {
      text = ''some-layout-specs'';
      source = /some/path.kdl;
      layout = {
        # Set made into kdl.
        # Setting this should overwrite the other options.
      };
    };
  };
}

```

This way, we preserve an interface similar to `home.file.<name>`. See docs below:
- `text`: https://nix-community.github.io/home-manager/options.xhtml#opt-home.file._name_.text
- `source`: https://nix-community.github.io/home-manager/options.xhtml#opt-home.file._name_.source

This should make converting a config to nix trivial, since one will be able to just point nix to their old files, or simply paste them in. Setting `layout` should overwrite the other options, IMO. An alternative, and even better option would be this:

```nix
{...}:
{
  programs.zellij.layouts = {
    my-layout-A = ''some-layout-specs'';
    my-layout-B = /some/path.kdl;
    my-layout-C = {
      # Set made into kdl.
    };
  };
}
```

[...]

PR #4614, opened 2023-10-29
generators: new toKDL parser

cgahr:
## Description

My idea for a new toKDL parser. The current one has significant problems since it is not possible to encode all possible KDls in nix (see #4613 #4612 #4485 #4465 #4160 #4054).

I propose this PR as a solution.

### Idea

In its core, KDL is a language encoding not only data but also lists of sequential commands. By its nature, nix has problems encoding this. 

In addition, the current implementation of `toKDL` was inspired by the rules for converting `kdl` to `json`. This, however, is the wrong direction: we want to convert `nix -> json -> kdl`. 

As such, the current implementation cannot encode things like
```
layout {
  pane
  pane
}
```

### Implementation
The idea is the following:
- parse the settings recursively
- lists are converted to lists of strings to be concatenated
- sets are also converted to lists of strings
- every basic datatype is converted to a string and added to the list.
- Lists of lists and lists of sets are automatically flattened

#### List of literals
All lists and sets are "exploded", i.e. each entry gets its own line:
```nix
{
  layout = [ pane pane ];
}
```
before:
```kdl
layout pane pane
```
now:
```kdl
layout {
  pane
  pane
}
```
In this case the previous behavior was intended to make writing the config easier but made the previous example impossible to express.

#### `_args` and `_props`
`_args` and `_props` do not work anymore:
```nix
{
  bind = {
    _args = [ "a" ];
    _props = { maybe = true };
}
```
Before:
```kdl
bind "a" maybe=true
```
Now:
```kdl
bind {
  "a"
  maybe = true
}
```
This behavior can be readded. Afaik, it was only intended to be a workaround to handle nix limited capabilites. With this new parser, they are not needed anymore. We could consider properly depcrecating this behavior.

#### Converting `'` to `"`
The new parser heavily uses strings. To make to syntax easier I propose to convert `'` automatically to '"' if they are in the `name` of a set or in a list:
```nix
{
  "bind 'a'" = { "something" }
  "bind \"b\"" = [ "something" ]
  "bind 'c'" = { "something" = "don't convert here" }
}
```
Now:
```kdl
bind "a" {
  something
}
bind "b" {
  something
}
bind "c" {
  something "don't convert here"
}
```
This is technically incompatible with the kdl specs since the following is a valid config:
```kdl
can't be right "right?"
```
This would be impossible to write as a nix with my proposed parser.

#### much more strings
With this parser much more strings would have to be used. Consider the following nix config for zellij:
```nix
{
  shared = [
    "unbind 'Ctrl p'"
    "unbind 'Ctrl n'"
    "unbind 'Ctrl s'"
    "unbind 'Ctrl o'"
    "unbind 'Ctrl t'"
    "unbind 'Ctrl h'"
    "unbind 'Ctrl b'"
  ];

  locked = {
    "bind \"Alt h\"" = { MoveFocusOrTab = "Left"; };
    "bind \"Alt l\"" = { MoveFocusOrTab = "Right"; };
    "bind \"Alt j\"" = { MoveFocus = "Down"; };
    "bind \"Alt k\"" = { MoveFocus = "Up"; };
  };

  pane = {
    "bind \"t\"" = [ "CloseTab" "CloseFocus" ];
  };

  normal = {
    "bind 'p'" = { SwitchToMode = "Pane"; };
    "bind 'n'" = { SwitchToMode = "Resize"; };
    "bind 's'" = { SwitchToMode = "Scroll"; };
    "bind 'o'" = { SwitchToMode = "Session"; };
    "bind 't'" = { SwitchToMode = "Tab"; };
    "bind 'h'" = { SwitchToMode = "Move"; };
    "bind 'b'" = { SwitchToMode = "Tmux"; };
  };
};
```
In my opinion, this is much more readable compared to before (and of course it is now possible to encode things that were not possible before).

### Checklist

For now the tests are failing, the original tests are falling to have a reference to how it used to be.

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@emilazy @foo-dogsquared @erikkrieg @h7x4 @rycee 

What do you think?


h7x4:
Hello @cgahr!

Thank you for looking into this. I strongly agree with you that the KDL parser is flawed and needs a complete rewrite. It probably shouldn't have been merged in the first place, and more work should have gone into ensuring a more complete one to one correspondence between the generator API and KDL.

I'm currently working on a new set of generators for more obscure config languages in nixpkgs. Starting with libconfig at https://github.com/NixOS/nixpkgs/pull/246115, I'm working my way through HOCON and KDL right now, using a very similar `nix -> json -> rust -> config` based pipeline, with plans of doing RON and Dotconf at some point in the future. The reason behind the rust approach is that the previous libconfig PR, which was of comparable size with this, was staled/rejected due to nix evaluation time overhead. There are also a bunch of other comments there you might find interesting or useful.

I think the KDL parser is much more needed in home-manager right now due to the broken state of the zellij module, but completing this also means potentially large duplicated efforts. If you'd like to join me, I'd be happy to work together on this! Feel free to DM me on matrix. If you decide you'd rather work on completing this here for now, I'll be happy to review and test the code.

PR #4616, opened 2023-10-29
home.shellAliases nushell

haras-unicorn:
### Description

Make `home.shellAliases` apply aliases to `nushell`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/home-environment.nix
+++ b/modules/home-environment.nix
@@ -516,6 +516,7 @@ in
     programs.bash.shellAliases = cfg.shellAliases;
     programs.zsh.shellAliases = cfg.shellAliases;
     programs.fish.shellAliases = cfg.shellAliases;
+    programs.nushell.shellAliases = cfg.shellAliases;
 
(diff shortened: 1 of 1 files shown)


sumnerevans (approved):
This seems sane

Nyabinary:
cc: @rycee

rycee:
My understanding is that nu shell cannot support `home.shellAliases` due to various issues discussed in #3529. I'm not following nu development, though so I don't know if this is something that has changed recently.

haras-unicorn:
Yeah I looked into it and it's still like that. I'll close this and wait for bug fixes in nushell.

For anyone stumbling upon this: nushell does not support pipes in aliases because of [the way alias keyword is parsed](https://github.com/nushell/nushell/issues/7753) and because of a [current bug in aliasing with braces](https://github.com/nushell/nushell/issues/10088). The latter of the two would have to be resolved in order to merge this.

PR #4617, opened 2023-10-30
k9s: add hotkey option

liyangau:
### Description

This PR adds an option to customise k9s hotkeys.

The keyword `hotKey` must be in camel case on the user config and it store the file at `k9s/hotkey.yml`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@katexochen

### DIFF
--- a/modules/programs/k9s.nix
+++ b/modules/programs/k9s.nix
@@ -8,7 +8,7 @@ let
   yamlFormat = pkgs.formats.yaml { };
 
 in {
-  meta.maintainers = [ hm.maintainers.katexochen ];
+  meta.maintainers = [ hm.maintainers.katexochen maintainers.liyangau ];
 
   options.programs.k9s = {
     enable =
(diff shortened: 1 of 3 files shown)


katexochen (approved):
Tested this locally, works great, thanks! We could also add the other defined files like plugin, alias,... but does not need to happen in this PR.

modules/programs/k9s.nix
```diff
@@ -8,7 +8,7 @@ let
   yamlFormat = pkgs.formats.yaml { };
 
 in {
-  meta.maintainers = [ hm.maintainers.katexochen ];
+  meta.maintainers = [ hm.maintainers.katexochen maintainers.liyangau ];
```
katexochen, line 11:
```suggestion
  meta.maintainers = with maintainers;  [ katexochen liyangau ];
```
Should work nowadays, too. ;)

teto (approved):

PR #4625, opened 2023-11-02
clipmenu: set Environment to a list

P1n3appl3:
### Description

The clipmenud service uses environment variables for configuration (such as `CM_MAX_CLIPS` to set a history size limit and `CM_SELECTIONS` to pick whether to use clipboard/primary/secondary selection).

With `Environment` as a string there was no way for me to set those additional env vars, but using a list allows me to write:

```nix
systemd.user.services.clipmenu.Service.Environment = ["CM_SELECTIONS=clipboard"];
```

There should probably be an option on the module to allow specifying those vars along with the service, but in the mean time this allows users to append options separately.

I noticed a similar change (#3083) in the git history. I wonder if _all_ services should use lists for `Environment` so that they're more easily modifiable.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@DamienCassou

### DIFF
--- a/modules/services/clipmenu.nix
+++ b/modules/services/clipmenu.nix
@@ -49,10 +49,12 @@ in {
 
       Service = {
         ExecStart = "${cfg.package}/bin/clipmenud";
-        Environment = "PATH=${
+        Environment = [
+          "PATH=${
             makeBinPath
             (with pkgs; [ coreutils findutils gnugrep gnused systemd ])
-          }";
+          }"
+        ];
       };
 
       Install = { WantedBy = [ "graphical-session.target" ]; };


DamienCassou:
> There should probably be an option on the module to allow specifying those vars along with the service

this would be good.

> I noticed a similar change (https://github.com/nix-community/home-manager/pull/3083) in the git history. I wonder if all services should use lists for Environment so that they're more easily modifiable.

I think this makes sense.

DamienCassou (approved):

rycee:
Thanks! I agree, this should probably always be set to a list.

PR #4626, opened 2023-11-02
firefox: add support for specifying policies

DamienCassou:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

/cc @rycee @kira-bruneau

### DIFF
--- a/modules/programs/firefox.nix
+++ b/modules/programs/firefox.nix
@@ -169,7 +169,10 @@ let
     else if isDarwin then
       package
     else if versionAtLeast config.home.stateVersion "19.09" then
-      package.override (old: { cfg = old.cfg or { } // fcfg; })
+      package.override (old: {
+        cfg = old.cfg or { } // fcfg;
+        extraPolicies = cfg.policies;
+      })
     else
       (pkgs.wrapFirefox.override { config = bcfg; }) package { };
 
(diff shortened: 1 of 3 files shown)


Nyabinary:
Why is this marked as draft, what needs to be added?

DamienCassou:
> Why is this marked as draft, what needs to be added?

Thank you for noticing. This PR was waiting for https://github.com/nix-community/home-manager/pull/4633 to be merged. Now that it has, I added a missing test file and the PR is now ready to be reviewed.

Nyabinary:
@rycee Seems good to merge right?

modules/programs/firefox.nix
```diff
@@ -169,7 +169,10 @@ let
     else if isDarwin then
       package
     else if versionAtLeast config.home.stateVersion "19.09" then
-      package.override (old: { cfg = old.cfg or { } // fcfg; })
+      package.override (old: {
+        cfg = old.cfg or { } // fcfg;
+        extraPolicies = cfg.policies;
```
jian-lin, line 174: This is a breaking change if users use `programs.firefox.package = pkgs.firefox.override { extraPolicies = { ... } }` to specify their policies before.  A release note should be added.

DamienCassou: @rycee this change is part of the already released 23.11. Should old release notes be modified?

jian-lin: Changing the implementation to `extraPolicies = old.extraPolicies // cfg.policies;` makes this almost not breaking.

rycee: Adding to the existing extra policies seems like a good idea. Perhaps need to add a guard like in the `cfg` line above, though? I.e.

``` nix
extraPolicies = old.extraPolicies or { } // cfg.policies;
```

Does that seem ok? With that change I guess it wouldn't be necessary to add to the changelog? Regardless, I have nothing against adding to the 23.11 changelog, if it helps clarify…

PR #4633, opened 2023-11-03
firefox: make programs.firefox.package the same as what is in PATH

DamienCassou:
### Description

Without this, when a user references `config.programs.firefox.package` in its configuration, s·he will get a different path as what is in its $PATH variable.

I didn't add any test. I will add one if you believe that is useful in this cases.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

/cc @rycee @kira-bruneau

### DIFF
--- a/modules/programs/firefox.nix
+++ b/modules/programs/firefox.nix
@@ -152,6 +152,27 @@ let
       '' + concatStringsSep "\n" (mapAttrsToList mkMsg duplicates);
     });
 
+  wrapPackage = package:
+    let
+      # The configuration expected by the Firefox wrapper.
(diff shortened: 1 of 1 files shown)


rycee:
Thanks! I think this makes a lot of sense in general and the code looks just fine.

But I would suggest instead introducing a new, read-only, option called `finalPackage` that is set to the wrapped package. Having both a `package` and `finalPackage` is somewhat common (if you search for `finalPackage` in HM you'll find various examples). The benefit being that you can recover the unwrapped package, which is occasionally helpful.

DamienCassou:
@rycee thank you for your review. Is it better now?

rycee:
Thanks! That looks nice. Merged now 🙂

PR #829, opened 2019-09-04
sway: add module

alexarice:
Creates a sway module.
This is mainly copied from the i3 module but has some changes to make it work with sway as well as make the defaults closer the the sway default config file.

~~This is wip because I'd like to make options for `output` and `input` sections of the config~~

PR #4636, opened 2023-11-05
sway: fix workspace 10 missing from default config

emilylange:
### Description

workspace 10 is part of upstream's default config, but was missing in home-manager.

The initial "sway: add module" PR (#829, 02d60400035028d4c8affab24557344ca371b9fd) went through multiple iterations and had workspace 10 included for a brief moment. Until the author removed it in a force-push commenting

> Have removed the last change which added bound ${modifer}+0 to workspace number 10 as this messed up workspace numbering in sway.

https://github.com/nix-community/home-manager/pull/829#issuecomment-575087553

The reason might have been, that sway used to sort the workspaces in the order they appeared in the config.

Attribute sets in nix are sorted, but not "naturally sorted", meaning `bindsym Mod1+0 workspace number 10` comes before `bindsym Mod1+0 workspace number 1`.

It's unclear if that's what really happened. A workaround would have been to use `lib.lists.naturalSort` in `keybindingsStr`.

But I cannot reproduce this anymore in any way.
I assume this has been fixed many years ago by now.

upstream config: https://github.com/swaywm/sway/blob/020a572ed615b8fe272c7566a27ee0abe73a58d7/config.in#L113-L134

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

### DIFF
--- a/modules/services/window-managers/i3-sway/sway.nix
+++ b/modules/services/window-managers/i3-sway/sway.nix
@@ -91,6 +91,7 @@ let
           "${cfg.config.modifier}+7" = "workspace number 7";
           "${cfg.config.modifier}+8" = "workspace number 8";
           "${cfg.config.modifier}+9" = "workspace number 9";
+          "${cfg.config.modifier}+0" = "workspace number 10";
 
           "${cfg.config.modifier}+Shift+1" =
             "move container to workspace number 1";
@@ -110,6 +111,8 @@ let
             "move container to workspace number 8";
           "${cfg.config.modifier}+Shift+9" =
             "move container to workspace number 9";
+          "${cfg.config.modifier}+Shift+0" =
+            "move container to workspace number 10";
 
           "${cfg.config.modifier}+Shift+minus" = "move scratchpad";
           "${cfg.config.modifier}+minus" = "scratchpad show";
--- a/tests/modules/services/window-managers/sway/sway-bar-focused-colors.conf
+++ b/tests/modules/services/window-managers/sway/sway-bar-focused-colors.conf
@@ -16,6 +16,7 @@ client.urgent #2f343a #900000 #ffffff #900000 #900000
 client.placeholder #000000 #0c0c0c #ffffff #000000 #0c0c0c
 client.background #ffffff
 
+bindsym Mod1+0 workspace number 10
 bindsym Mod1+1 workspace number 1
 bindsym Mod1+2 workspace number 2
 bindsym Mod1+3 workspace number 3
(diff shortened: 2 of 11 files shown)


rvolosatovs (approved):

SebTM (approved):
LGTM, another reason it might got left out is that the first defined workspace in keybindings is the workspace active when sway starts, but this can be controlled with defaultWorkspace already.

Toomoch:
Somehow my Sway desktop is booting to workspace 10 always, and I can't seem to know why until I found this commit. Could it be the cause?
Here is my config: https://github.com/Toomoch/nixos-config/blob/testing/home/arnau/sway/default.nix

SebTM:
Yes, but I don't see you using https://nix-community.github.io/home-manager/options.xhtml#opt-wayland.windowManager.sway.config.defaultWorkspace

correct? Then it's handled like explained here: https://github.com/nix-community/home-manager/pull/4636#pullrequestreview-1872565915

Toomoch:
> Yes, but I don't see you using https://nix-community.github.io/home-manager/options.xhtml#opt-wayland.windowManager.sway.config.defaultWorkspace
> 
> correct? Then it's handled like explained here: [#4636 (review)](https://github.com/nix-community/home-manager/pull/4636#pullrequestreview-1872565915)

So sorry, I didn't even read the whole issue.  
That makes sense as number 10 is defined before number 1, see the config file home-manager built:
image
Using defaultWorkspace fixed it, thank you very much.

Toomoch:
Maybe it would be a good idea to set it to 1 if using mkOptionDefault in the keybindings.

philipwilk:
Hi, I've run into this as well - my default workspace being changed to 10. This is manageable on single display environments, where you can change the default workspace to be 1 explicitly, but this does not work with multi monitor setups. When i connect my laptop to a monitor, the laptop screen is workspace 1 but the monitor is workspace 10 instead of workspace 2. How could I workaround this?
If there is an option to set the default workspace for a specific monitor or resolution or something, I cannot use it, because I'm not using just one specific monitor; I'm using different ones across the uni, and I can't practically manually create a config for each one.

philipwilk:
I'm not sure how to manually test it on my local machine without switching my hm input but i'd be up to try it

emilylange:
hmm turns out throwing `lib.lists.naturalSort` at `keybindingsStr` is not as straight forward as I thought :(

alexarice:
is it possible to add `mkAfter` to the workspace 10 bindings?

Scrumplex:
Why don't we just set a default value for `defaultWorkspace`?

emilylange:
> Why don't we just set a default value for `defaultWorkspace`?

Because it would not fix the ordering issue on multi-monitor setups. See https://github.com/nix-community/home-manager/pull/4636#issuecomment-1953468742

michaelfranzl:
> The reason might have been, that sway used to sort the workspaces in the order they appeared in the config.

@emilylange I just tested it, and Sway still behaves like this. `defaultWorkspace` is not a workaround when using more than one monitor, because this will put workspace "10" to the second position.

PR #4640, opened 2023-11-06
sftpman: add module

fugidev:
### Description

Add a module for [sftpman](https://github.com/spantaleev/sftpman), an application that handles sshfs/sftp file systems mounting.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/sftpman.nix
+++ b/modules/programs/sftpman.nix
@@ -1,0 +1,118 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.programs.sftpman;
+
+  jsonFormat = pkgs.formats.json { };
+
+  mountOpts = { config, name, ... }: {
+    options = {
+      host = mkOption {
+        type = types.str;
+        description = "The host to connect to.";
+      };
+
+      port = mkOption {
+        type = types.port;
+        default = 22;
+        description = "The port to connect to.";
+      };
+
+      user = mkOption {
+        type = types.str;
+        description = "The username to authenticate with.";
+      };
+
+      mountOptions = mkOption {
+        type = types.listOf types.str;
+        default = [ ];
+        description = "Options to pass to sshfs.";
+      };
+
+      mountPoint = mkOption {
+        type = types.str;
+        description = "The remote path to mount.";
+      };
+
+      authType = mkOption {
+        type = types.enum [
+          "password"
+          "publickey"
+          "hostbased"
+          "keyboard-interactive"
+          "gssapi-with-mic"
+        ];
+        default = "publickey";
+        description = "The authentication method to use.";
+      };
+
+      sshKey = mkOption {
+        type = types.nullOr types.str;
+        default = cfg.defaultSshKey;
+        defaultText =
+          lib.literalExpression "config.programs.sftpman.defaultSshKey";
+        description = ''
+          Path to the SSH key to use for authentication.
+          Only applies if authMethod is `publickey`.
+        '';
+      };
+
+      beforeMount = mkOption {
+        type = types.str;
+        default = "true";
+        description = "Command to run before mounting.";
+      };
+    };
+  };
+in {
+  meta.maintainers = with maintainers; [ fugi ];
+
+  options.programs.sftpman = {
+    enable = mkEnableOption
+      "sftpman, an application that handles sshfs/sftp file systems mounting";
+
+    package = mkPackageOption pkgs "sftpman" { };
+
+    defaultSshKey = mkOption {
+      type = types.nullOr types.str;
+      default = null;
+      description =
+        "Path to the SSH key to be used by default. Can be overridden per host.";
+    };
+
+    mounts = mkOption {
+      type = types.attrsOf (types.submodule mountOpts);
+      default = { };
+      description = ''
+        The sshfs mount configurations written to
+        {file}`$XDG_CONFIG_HOME/sftpman/mounts/`.
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable {
+    assertions = [
+      (let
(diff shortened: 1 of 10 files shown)


modules/programs/sftpman.nix
```diff
@@ -0,0 +1,101 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.programs.sftpman;
+
+  jsonFormat = pkgs.formats.json { };
+
+  mountOpts = { config, name, ... }: {
```
rycee, line 10: This would be well-suited as a [freeform module](https://nixos.org/manual/nixos/stable/#sec-freeform-modules):

```suggestion
  mountOpts = types.submodule ({ config, name, ... }: {
    freeformType = jsonFormat.type;
```

modules/programs/sftpman.nix
```diff
@@ -0,0 +1,101 @@
...
+    options = {
+      host = mkOption {
+        type = types.str;
+        description = "The host to connect to.";
+        default = null;
```
rycee, line 15: I generally think it is best to place the fields in the order `type`, `default`, `apply`, `example`, `description`:

```suggestion
        type = types.nullOr types.str;
        default = null;
        description = "The host to connect to.";
```

Note, if you specify `types.str` then `null` is not an allowed value. You need to use `nullOr str`.

modules/programs/sftpman.nix
```diff
@@ -0,0 +1,101 @@
...
+      };
+      port = mkOption {
```
rycee, line 17: Please separate options by an empty line:
```suggestion

      port = mkOption {
```

modules/programs/sftpman.nix
```diff
@@ -0,0 +1,101 @@
...
+          "keyboard-interactive"
+          "gssapi-with-mic"
+        ];
+        description = "The authentication method to use.";
+        default = "publickey";
+      };
+      sshKey = mkOption {
+        type = types.nullOr types.str;
+        description = ''
+          Path to the ssh key to use for authentication.
```
rycee, line 51:
```suggestion
          Path to the SSH key to use for authentication.
```

modules/programs/sftpman.nix
```diff
@@ -0,0 +1,101 @@
...
+          Only applies if authMethod is `publickey`.
+        '';
+        default = cfg.defaultSshKey;
+        apply = sshKey:
+          assert assertMsg (config.authType != "publickey" || sshKey != null)
```
rycee, line 56: It is not recommended to place explicit `assert`s in modules. Use the `assertions` option instead. Something like the following might work (untested):

``` patch
diff --git a/modules/programs/sftpman.nix b/modules/programs/sftpman.nix
index 42fdf8f0..8ecfa7f0 100644
--- a/modules/programs/sftpman.nix
+++ b/modules/programs/sftpman.nix
@@ -52,10 +52,6 @@ let
           Only applies if authMethod is `publickey`.
         '';
         default = cfg.defaultSshKey;
-        apply = sshKey:
-          assert assertMsg (config.authType != "publickey" || sshKey != null)
-            "sftpman ${name}: sshKey must be set when using publickey auth.";
-          sshKey;
       };
       beforeMount = mkOption {
         type = types.str;
@@ -90,6 +86,19 @@ in {
   };
 
   config = mkIf cfg.enable {
+    assertions = [(
+      let
+        hasMissingKey = _: mount: mount.authType == "publickey" && mount.sshKey == null;
+        mountsWithMissingKey = attrNames (filterAttrs hasMissingKey cfg.mounts);
+        mountsWithMissingKeyStr = concatStringsSep ", " mountsWithMissingKey;
+      in {
+        assertion = mountsWithMissingKeyStr == [];
+        message = ''
+          sftpman mounts using authentication type "publickey" but missing 'sshKey': ${mountsWithMissingKeyStr}
+        '';
+      }
+    )];
+
 
     xdg.configFile = mapAttrs' (name: value:
```

modules/programs/sftpman.nix
```diff
@@ -0,0 +1,101 @@
...
+        description = "The authentication method to use.";
+        default = "publickey";
+      };
+      sshKey = mkOption {
+        type = types.nullOr types.str;
+        description = ''
+          Path to the ssh key to use for authentication.
+          Only applies if authMethod is `publickey`.
+        '';
+        default = cfg.defaultSshKey;
```
rycee, line 54: When an option default value depends on the configuration then you must provide a `defaultText` field to avoid making the documentation depend on the configuration. Something like this perhaps:

```suggestion
        default = cfg.defaultSshKey;
        defaultText = lib.literalExpression "config.programs.sftpman.defaultSshKey";
```

modules/programs/sftpman.nix
rycee: With this you can then also add a test of the assertion. See https://github.com/nix-community/home-manager/blob/master/tests/modules/programs/nix-index/assert-on-command-not-found.nix for an example.

rycee:

modules/programs/sftpman.nix
fugidev: The JSON config format is unfortunately not documented, users are supposed to configure the mounts via the cli. The JSON keys are similar to the respective cli flags, but not exactly the same. So I thought it would be better to declare all the options explicitly.

fugidev:
Thank you for the review and the suggestions! I will work on incorporating them soon.

fugidev:
I made the requested changes (except for freeform module, see my comment) and added a test for the sshKey assertion.

modules/programs/sftpman.nix
rycee: Fair enough.

rycee:

PR #4642, opened 2023-11-09
emacs: Socket activation fixes

rvl:
### Description

1. Apparently, the bug was fixed where emacs would delete the socket file which belongs to systemd. See [bug 47511](https://debbugs.gnu.org/47511) and [emacs commit 53dfb51](https://github.com/emacs-mirror/emacs/commit/53dfb51f55bf66bb939505c60d12952dd64820a7).
2. With socket activation enabled, after stopping `emacs.service`, it would immediately start again. The systemd socket unit is tweaked to make stop mean stop.
3. In the corner case where the user starts the service while the socket is stopped, it is then not possible to start the socket, because emacs already created it. Add a dependency from the service to the socket to avoid this situation.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### (Maintainer) CC

@tadfisher @jian-lin @DamienCassou @rcerc @rycee


rycee:

PR #4644, opened 2023-11-10
 sway: Check config file validity 

amarshall:
### Description

Helps avoid successful build but Sway failing to start.

To meaningfully test this, I had to actually use `pkgs.sway` in the test
rather than the stub, but left all other tests using the stub and
changed them to skipping the test.

Only ran Sway tests, not all tests.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

### DIFF
--- a/modules/services/window-managers/i3-sway/sway.nix
+++ b/modules/services/window-managers/i3-sway/sway.nix
@@ -258,12 +258,21 @@ let
 
   variables = concatStringsSep " " cfg.systemd.variables;
   extraCommands = concatStringsSep " && " cfg.systemd.extraCommands;
+  swayPackage = if cfg.package == null then pkgs.sway else cfg.package;
   systemdActivation = ''
     exec "${pkgs.dbus}/bin/dbus-update-activation-environment --systemd ${variables}; ${extraCommands}"'';
 
   configFile = pkgs.writeTextFile {
     name = "sway.conf";
-    text = (concatStringsSep "\n"
+
+    # Sway always does some init, see https://github.com/swaywm/sway/issues/4691
+    checkPhase = lib.optionalString cfg.checkConfig ''
+      export DBUS_SESSION_BUS_ADDRESS=/dev/null
+      export XDG_RUNTIME_DIR=$(mktemp -d)
+      ${pkgs.xvfb-run}/bin/xvfb-run ${swayPackage}/bin/sway --config "$target" --validate --unsupported-gpu
+    '';
+
+    text = concatStringsSep "\n"
       ((optional (cfg.extraConfigEarly != "") cfg.extraConfigEarly)
         ++ (if cfg.config != null then
           with cfg.config;
@@ -319,7 +328,7 @@ let
         else
           [ ]) ++ (optional cfg.systemd.enable systemdActivation)
         ++ (optional (!cfg.xwayland) "xwayland disable")
-        ++ [ cfg.extraConfig ]));
+        ++ [ cfg.extraConfig ]);
   };
 
   defaultSwayPackage = pkgs.sway.override {
@@ -469,6 +478,13 @@ in {
       description = "Sway configuration options.";
     };
 
+    checkConfig = mkOption {
+      type = types.bool;
+      default = true;
+      description =
+        "If enabled (the default), validates the generated config file.";
+    };
+
     extraConfig = mkOption {
       type = types.lines;
       default = "";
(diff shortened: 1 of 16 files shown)


modules/services/window-managers/i3-sway/sway.nix
```diff
@@ -253,63 +253,76 @@ let
...
-      ++ (optional (!cfg.xwayland) "xwayland disable") ++ [ cfg.extraConfig ]));
+  swayPackage = if cfg.package == null then pkgs.sway else cfg.package;
+
+  configFile = pkgs.writeTextFile {
+    name = "sway.conf";
+
+    checkPhase = lib.optionalString cfg.checkConfig ''
+      export DBUS_SESSION_BUS_ADDRESS=/dev/null
+      export XDG_RUNTIME_DIR=$(mktemp -d)
+      ${pkgs.xvfb-run}/bin/xvfb-run ${swayPackage}/bin/sway --config "$target" --validate
```
rycee, line 264: A bit weird that xvfb-run would be necessary to just validate the configuration file? Perhaps worth creating an upstream issue in the sway tracker?

modules/services/window-managers/i3-sway/sway.nix
```diff
@@ -253,63 +253,76 @@ let
...
+    '';
+
+    text = (concatStringsSep "\n"
```
rycee, line 267: Can remove the unnecessary parenthesis:

```suggestion
    text = concatStringsSep "\n"
```

modules/services/window-managers/i3-sway/sway.nix
```diff
@@ -428,6 +441,13 @@ in {
       description = "Sway configuration options.";
     };
 
+    checkConfig = mkOption {
```
rycee, line 444: I'm guessing one would always want to check the configuration? Perhaps could add `visible = false;` or even `internal = true;` to avoid having the option show up in the documentation?

tests/modules/services/window-managers/sway/sway-invalid-config.nix
```diff
@@ -0,0 +1,12 @@
+{ ... }:
+
+{
```
rycee, line 3: Since this will pull down the sway package closure we'll want to mark it as a big test:

```suggestion
{ lib, ... }:

lib.mkIf config.test.enableBig {
```

rycee:
Thanks! I added a few comments 🙂

modules/services/window-managers/i3-sway/sway.nix
```diff
@@ -253,63 +253,76 @@ let
...
-          ++ map floatingCriteriaStr floating.criteria # floating
-          ++ map windowCommandsStr window.commands # window commands
-          ++ map startupEntryStr startup # startup
-          ++ map workspaceOutputStr workspaceOutputAssign # custom mapping
-        )
-      else
-        [ ]) ++ (optional cfg.systemd.enable ''
-          exec "${pkgs.dbus}/bin/dbus-update-activation-environment --systemd DISPLAY WAYLAND_DISPLAY SWAYSOCK XDG_CURRENT_DESKTOP XDG_SESSION_TYPE NIXOS_OZONE_WL; systemctl --user start sway-session.target"'')
-      ++ (optional (!cfg.xwayland) "xwayland disable") ++ [ cfg.extraConfig ]));
+  swayPackage = if cfg.package == null then pkgs.sway else cfg.package;
```
oxalica, line 256: Shouldn't this default to `defaultSwayPackage`? Otherwise, if the user enables any overrides, this will pull in both original one and overridden one.

tests/modules/services/window-managers/sway/sway-invalid-config.nix
```diff
@@ -0,0 +1,12 @@
+{ config, lib, ... }:
+
+lib.mkIf config.test.enableBig {
+  wayland.windowManager.sway = {
+    enable = true;
+    checkConfig = true;
+  };
+
+  nmt.script = ''
+    assertFileExists home-files/.config/sway/config
```
oxalica, line 10: Why this test is called "invalid-config" but expect a success? Is it possible to add some fail tests to ensure the check actually do something?

modules/services/window-managers/i3-sway/sway.nix
oxalica: > swaywm/sway#4691

I suggest to link to that issue in code comments.

rycee:
Thanks! Rebased to master in 3a435342e2e5e2bff1d3331c2bd9e70f25693ea2 🙂

PR #4645, opened 2023-11-10
nix-your-shell: new module

nbraud:
### Description

Add a module for `nix-your-shell`, a wrapper for `nix develop` and `nix-shell`
to run the same shell inside the environment (instead of always Bash)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/nix-your-shell.nix
+++ b/modules/programs/nix-your-shell.nix
@@ -1,11 +1,12 @@
 { config, lib, pkgs, ... }:
 let
-  inherit (lib) genAttrs getExe;
   cfg = config.programs.nix-your-shell;
 
-  # In principle `bash` is supported too, but...  😹
-  shells = [ "fish" "ion" "nushell" "zsh" ];
-  programs = [ "nix" "nix-shell" ];
+  exe = lib.getExe pkgs.nix-your-shell;
+  mkShellAliases = shell: {
+    nix = "${exe} ${shell} nix --";
+    nix-shell = "${exe} ${shell} nix-shell --";
+  };
 in {
   meta.maintainers = with lib.maintainers; [ nicoo ];
 
@@ -14,8 +15,10 @@ in {
     to run the same shell inside the new environment.
   '';
 
-  config.programs = lib.mkIf cfg.enable (genAttrs shells (shell: {
-    shellAliases = genAttrs programs
-      (program: "${getExe pkgs.nix-your-shell} ${shell} ${program} --");
-  }));
+  config.programs = lib.mkIf cfg.enable {
+    fish.shellAliases = mkShellAliases "fish";
+    ion.shellAliases = mkShellAliases "ion";
+    nushell.shellAliases = mkShellAliases "nushell";
+    zsh.shellAliases = mkShellAliases "zsh";
+  };
 }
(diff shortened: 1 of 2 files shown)


modules/programs/nix-your-shell.nix
```diff
@@ -0,0 +1,21 @@
...
+  options.programs.nix-your-shell.enable = lib.mkEnableOption ''
+    `nix-your-shell`, a wrapper for `nix develop` or `nix-shell`
+    to run the same shell inside the new environment.
+  '';
+
+  config.programs = lib.mkIf cfg.enable (genAttrs shells (shell: {
+    shellAliases = genAttrs programs
+      (program: "${getExe pkgs.nix-your-shell} ${shell} ${program} --");
+  }));
+}
```
rycee, line 21: I think it may be worth simplifying the code a bit, the `genAttrs` are always a bit tricky to follow:

```suggestion
{ config, lib, pkgs, ... }:
let
  inherit (lib) getExe;
  cfg = config.programs.nix-your-shell;

  mkShellAliases = shell:
    let exe = getExe pkgs.nix-your-shell;
    in {
      nix = "${exe} ${shell} nix --";
      nix-shell = "${exe} ${shell} nix-shell --";
    };
in {
  meta.maintainers = with lib.maintainers; [ nicoo ];

  options.programs.nix-your-shell.enable = lib.mkEnableOption ''
    {command}`nix-your-shell`, a wrapper for {command}`nix develop` or
    {command}`nix-shell` to run the same shell inside the new environment'';

  config = lib.mkIf cfg.enable {
    programs.fish.shellAliases = mkShellAliases "fish";
    programs.ion.shellAliases = mkShellAliases "ion";
    programs.nushell.shellAliases = mkShellAliases "nushell";
    programs.zsh.shellAliases = mkShellAliases "zsh";
  };
}
```

rycee:
Thanks for the contribution! Looks like a nice tool 🙂

modules/programs/nix-your-shell.nix
```diff
@@ -0,0 +1,21 @@
...
+  meta.maintainers = with lib.maintainers; [ nicoo ];
+
+  options.programs.nix-your-shell.enable = lib.mkEnableOption ''
+    `nix-your-shell`, a wrapper for `nix develop` or `nix-shell`
+    to run the same shell inside the new environment.
+  '';
+
+  config.programs = lib.mkIf cfg.enable (genAttrs shells (shell: {
+    shellAliases = genAttrs programs
+      (program: "${getExe pkgs.nix-your-shell} ${shell} ${program} --");
```
rycee, line 19: I'm not certain it is always the best idea to just make a plain alias. Consider, for example, fish. We have

```
$ nix-your-shell fish
# If you see this output, you probably forgot to pipe it into `source`:
# nix-your-shell fish | source

function nix-shell --description "Start an interactive shell based on a Nix expression"
    nix-your-shell fish nix-shell -- $argv
end

function nix --description "Reproducible and declarative configuration management"
    nix-your-shell fish nix -- $argv
end
```

So it will also provide a nice command description, which will be missing in the plain alias definition.

For the nushell alias, I'm not certain it will work at all. My understanding is that nushell aliases cannot take command line parameters…

musjj: > on ZSH it also breaks autocompletion

I usually use `compdef` for this:

```zsh
compdef _actual_command wrapped_command
```

dawidd6: Yea, we should really use `nix-your-shell fish | source`, like it is done in `zoxide`, `starship`, etc.

dawidd6:
@nbraud are you still working on this?

PR #4648, opened 2023-11-11
swayidle: always restart systemd unit on failure

9p4:
### Description

Occasionally, swayidle crashes with a failure to connect to the Wayland session. Ideally, swayidle should automatically restart instead of leaving the system in a vulnerable state.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@c0deaddict

### DIFF
--- a/modules/services/swayidle.nix
+++ b/modules/services/swayidle.nix
@@ -122,6 +122,7 @@ in {
 
       Service = {
         Type = "simple";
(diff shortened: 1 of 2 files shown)


rycee:
Yeah, makes sense. Merged to master now 🙂

I'll also backported it to the 23.05 branch since it may have security implications.

PR #4649, opened 2023-11-12
hishtory: add module programs.hishtory

willemml:
### Description

Adds `programs.hishtory` module to enable use of [hishtory](https://github.com/ddworken/hishtory). Due to the way this program's configuration is stored it's configuration must be updated using shell commands in the activation script. Because of this, configuration of the program by home-manager (beyond basic installation) is optional, and by default disabled. If anyone familiar with the program has suggestions on how this configuration stage could be better implemented, please let me know.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/lib/maintainers.nix
+++ b/modules/lib/maintainers.nix
@@ -454,6 +454,15 @@
     github = "wcarlsen";
     githubId = 17003032;
   };
+  willemml = {
+    github = "willemml";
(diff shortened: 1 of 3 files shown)


willemml:
I no longer use hishtory. I do not know if this module works anymore. Feel free to use this code if you test it and it still works.

PR #4655, opened 2023-11-12
himitsu: add service module

patwid:
### Description

Add himitsu service, starting himitsud.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/services/himitsu.nix
```diff
@@ -0,0 +1,55 @@
...
+      enable = mkEnableOption ''
+        Himitsu, secret storage system for Unix systems
+      '';
+
+      package = mkOption {
+        type = types.package;
+        default = pkgs.himitsu;
+        defaultText = literalExpression "pkgs.himitsu";
+        description = "The himitsu package to use";
+      };
```
ambroisie, line 18: Use `mkPackageOption`.

modules/services/himitsu.nix
```diff
@@ -0,0 +1,55 @@
...
+
+      prompterPackage = mkOption {
+        type = types.package;
+        default = pkgs.hiprompt-gtk-py;
+        defaultText = literalExpression "pkgs.hiprompt-gtk-py";
+        description = "The himitsu prompter package to use";
+      };
```
ambroisie, line 25: Same feedback.

modules/services/himitsu.nix
```diff
@@ -0,0 +1,55 @@
...
+
+      prompter = mkOption {
+        type = types.str;
+        default = "${cfg.prompterPackage}/bin/hiprompt-gtk";
+        defaultText =
+          literalExpression "\${cfg.prompterPackage}/bin/hiprompt-gtk";
+        description = "The himitsu prompter command to use";
+      };
```
ambroisie, line 33: Rather than have `prompterPackage` and `prompter`, just have `prompter = mkPackageOption "himitsu prompter" [ default = [ "hiprompt-gtk-py" ]; };` and use `lib.getExe` to generate the configuration file.

ambroisie (changes requested):

modules/services/himitsu.nix
patwid: we have to use `lib.getExe'` and specify the name of the executable explicitly (unless `meta.mainProgram` is specified accordingly by the corresponding package.

modules/services/himitsu.nix
```diff
@@ -0,0 +1,45 @@
...
+      Service = {
+        ExecStart = "${cfg.package}/bin/himitsud";
+      };
+    };
+
+    xdg.configFile."himitsu/config.ini" = {
+      text = lib.generators.toINI { } {
+        himitsud = {
+          prompter = lib.getExe' cfg.prompter "hiprompt-gtk";
```
ambroisie, line 38: If `lib.getExe` does not work, add `meta.mainProgram` to the package upstream.

```suggestion
          prompter = lib.getExe cfg.prompter;
```

patwid: I already added a suggestion to the open PR: https://github.com/NixOS/nixpkgs/pull/179022

If this suggestion is applied, we can revert this change.

khaneliman:
Looks like that package never got merged.

PR #4660, opened 2023-11-13
bat: support boolean flags in config (fixes #4657)

### LINKED ISSUE #4657: Support flags without associated values for `bat` config.
### Description
Currently, `programs.bat.config` assumes that flags to go in the config file all take a value. However, some flags, like `--show-all` and `--lessopen` (if the `lessopen` feature is enabled) do not take a value.

I would be up for doing a PR for this myself, but I'm not sure if the best way to implement this is to somehow adapt `programs.bat.config`, to just add another option like `programs.bat.extraConfig`, or something else entirely.

pluiedev:
### Description

Previously, users cannot enable boolean flags like `--show-all` in bat's config since all options were expected to be either a string, or a list of strings. With this commit boolean flags are simply appended to the end of the config if they are set to `true`, and discarded otherwise.

For example, the config
```nix
{
  theme = "TwoDark";
  show-all = true;
  lessopen = false;
}
```
would produce a config file that looks like
```
--theme='TwoDark'
--show-all
```

Fixes #4657

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/bat.nix
+++ b/modules/programs/bat.nix
@@ -8,10 +8,20 @@ let
 
   package = pkgs.bat;
 
-  toConfigFile = generators.toKeyValue {
-    mkKeyValue = k: v: "--${k}=${lib.escapeShellArg v}";
-    listsAsDuplicateKeys = true;
-  };
+  toConfigFile = attrs:
+    let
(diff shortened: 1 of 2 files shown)


rycee:

PR #4661, opened 2023-11-13
git: add alternate signing methods

### LINKED ISSUE #4221: Add option for SSH commit signing for Git config

pluiedev:
### Description

The Git module now supports SSH and X.509 signing in addition to OpenPGP/GnuPG, via setting the `programs.git.signing.format` option. It defaults to `openpgp` for now as a backwards compatibility measure, but I feel like we shouldn't enforce GPG as the default on everyone, especially for people who use SSH signing like me.

Accordingly, `programs.git.signing.gpgPath` has been renamed to `programs.git.signing.signer`, as now the signer binary is not restricted to GnuPG. Users should only get a warning and everything should continue to work.

Fixes #4221, supersedes #4235

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -1656,6 +1656,23 @@ in {
           See https://codeberg.org/dnkl/yambar for more.
         '';
       }
+
+      {
+        time = "2024-05-21T20:22:57+00:00";
+        condition = config.programs.git.signing != { };
+        message = ''
+          The Git module now supports signing via SSH and X.509 keys, in addition to OpenPGP/GnuPG,
+          via the `programs.git.signing.format` option.
+
+          The format defaults to `openpgp` for now, due to backwards compatibility reasons — this is
+          not guaranteed to last! GPG users should manually set `programs.git.signing.format` to
+          `openpgp` as soon as possible.
+
+          Accordingly, `programs.git.signing.gpgPath` has been renamed to the more generic option
+          `programs.git.signing.signer` as not everyone uses GPG.
+          Please migrate to the new option to suppress the generated warning.
+        '';
+      }
     ];
   };
 }
--- a/modules/programs/git.nix
+++ b/modules/programs/git.nix
@@ -14,33 +14,6 @@ let
       supersectionType = attrsOf (either multipleType sectionType);
     in attrsOf supersectionType;
 
-  signModule = types.submodule {
-    options = {
-      key = mkOption {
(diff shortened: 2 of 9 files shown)


pluiedev:
Update: since the branch-off to 23.11 occurred, I'm moving the backwards compat version to 24.05. Hope this can land before I have to do this again :sweat_smile:

olmokramer:
I applied this PR to my clone of the home-manager repo, but when I rebuild my NixOS configuration, I'm getting the following error:

```
       error: The option `home.stateVersion' is used but not defined.
```

even though `home.stateVersion` is set for my user:

```
$ nix repl .
> nixosConfigurations.<redacted hostname>.config.home-manager.users.<redacted username>.home.stateVersion
"22.11"
```

pluiedev:
society if PRs to HM actually get merged

pluiedev:
Since during the time this PR has sat here idle 23.11 has almost already gone through its entire lifecycle, I'm going to move the backwards compat cutoff to 24.11. *Hopefully* I don't have to do this again in another 7 months or so?

pluiedev:
> I ran into the same issue that @olmokramer ran into with:
> 
> ```
> error: The option `home.stateVersion' is used but not defined.
> ```

Hm... honestly I'm not sure what's going on anymore. Maybe `home.stateVersion` was referred to too early?

pluiedev:
Closing in favor of #5516 — all new work should happen there

PR #4663, opened 2023-11-14
i3blocks: added configuration module

Noodlez1232:
### Description

Added a configuration module (and tests) for i3blocks.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/i3blocks.nix
+++ b/modules/programs/i3blocks.nix
@@ -1,0 +1,114 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.programs.i3blocks;
+
+  # Re-make the atom type for the INI files.
+  # For some reason, the normal INI type seems to be incompatible with
+  # DAG
+  configAtomType = let
+    # Keep the INI atom type here
+    optType = with types; (nullOr (oneOf [ int bool str float ]));
+  in types.mkOptionType {
+    name = "INI config atom";
+    description = "INI atom (null, int, bool, string, or float)";
+    check = x: optType.check x;
+    merge = (loc: defs: (optType.merge loc defs));
+  };
+
+  # Create the type of the actual config type
+  configType = types.attrsOf configAtomType;
+
+  # The INI generator
+  mkIni = generators.toINI { };
+
+in {
+  meta.maintainers = [ maintainers.noodlez1232 ];
+
+  options.programs.i3blocks = {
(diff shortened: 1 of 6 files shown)


Noodlez1232:
Any way I can get this upstreamed? I'd really like to be able to move on to other projects in home manager :P

Also, I think this PR should supercede and replace PR #3366

modules/programs/i3blocks.nix
```diff
@@ -0,0 +1,109 @@
...
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.i3blocks;
+      defaultText = literalExpression "pkgs.i3blocks";
+      description = "Package providing {command}`i3blocks`.";
+    };
+
+    bars = mkOption {
+      type = with types; attrsOf (hm.types.dagOf configType);
```
rycee, line 41: Sorry about the long delay!

I'm wondering if it wouldn't be sufficient to have
```suggestion
      type = with types; attrsOf (listOf configType);
```
here since I assume you typically full control over the block order.

Noodlez1232: I tried this for a bit, and lemme show you a snippet of how my HM config works, then it might make more sense why I chose a DAG.
```nix
let
# Host specific configuration
extraBottomBlocks = {
   main = {
      network = lib.hm.dag.entryBetween [ "date" ] [ "vol" ] {
         # Block here
      };
      # Other blocks here
   };
   framework = {
      battery = lib.hm.dag.entryBetween [ "date" ] [ "vol" ] {
         # Block here
      };
      # Other blocks here
   };
};
in
{
   programs.i3blocks = {
      bars = {
        top = {
           # ...
        };
        # Generic blocks
        bottom = {
            date = {
               # ...
            };
            vol = { };
        }; // extraBottomBlocks."${osConfig.networking.hostName}";
      };
   };
}
```

Noodlez1232: I had in mind the ability to set up kinda "generic" block structures and you can choose to put other blocks between them later on, making it a bit easier to work with in that way.

Noodlez1232: Also, I'd have to rewrite pretty much the entire module for this to work, which means a lot of redoing debugging and all that.

rycee: Fair enough, merged to master now 🙂

PR #4665, opened 2023-11-15
Document zellij KDL settings

gotcha:
### Description

Add an example for KDL arguments

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


brianmay:
I don't like to be the pessimist, but wondering if this will work. "With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository." https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow

My experience, not even jobs triggered by dependabot created pull requests have access to secrets. I had to update all of my github workflows as a result.

But nothing wrong with trying it. :-)

Also see: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions

brianmay:
Can you please add another keyboard binding to the example? Maybe one that doesn't have an argument to the command?

(apologies for previous comment which I deleted, somehow I got my repos confused and submitted it it the wrong place)

johnae:
You've given an example of one binding. But if I need to make more, how do I do that? I.e this works ofc:

```yaml
        keybinds = {
          shared_except = {
            _args = [ "move" "locked" ];
            bind = {
              _args = [ "Ctrl m" ];
              SwitchToMode = "move";
            };
          };
        };
```

but this?

```yaml
        keybinds = {
          shared_except = {
            _args = [ "move" "locked" ];
            bind = {
              _args = [ "Ctrl m" ];
              SwitchToMode = "move";
            };
            bind = {
              _args = [ "Alt b" ];
              Run = {
                 _args = ["ls" "-la"];
              };
            };
          };
        };
```

Those two "bind" keys in the attrs aren't unique so you'll get an error. I would imagine perhaps a list of binds, is that how it looks?

gotcha:
@johnae You are right, I did found about that problem later. And had not found the time to report it here.

tennox:
For anyone searching for a workaround - [see this](https://github.com/nix-community/home-manager/pull/4465#issuecomment-1732675770)

gabevenberg:
Is there anything blocking this from being moved of draft stage, @gotcha?

gotcha:
@gabevenberg Unfortunately yes, see https://github.com/nix-community/home-manager/pull/4665#issuecomment-1822999684

gabevenberg:
given that KDL cannot be fully represented in, for example, [JSON](https://github.com/kdl-org/kdl?tab=readme-ov-file#compatibility-with-json-and-xml), and at least one [pr](https://github.com/nix-community/home-manager/pull/4614) claims that the current KDL generator makes it impossible to make all possible KDL configs in nix, should  #4465 be prioritized over this, or, perhaps, a `settingsFile.source`/`settingsfile.text` style option?

fightingdreamer:
I would really like plain text in KDL format. Let's make every example from Zellij page `just works`, including but not limiting to:
- color schemes
- keybindings
- layouts

Doing this in YAML by checking unmaintained docs and trying to find a way to generate correct KDL from YAML feels not right.

b-m-f:
Giving a plaintext option seems like the best way for the moment.

PR #4667, opened 2023-11-15
gradle: add module

britter:
### Description

Introduces a new program called gradle for managing files stored in the
home directory by the [Gradle Build Tool](https://gradle.org). Gradle
uses the $HOME/.gradle folder for all it's configuration.

Features of the new program module are:

- Automatically setting programs.java.enable = true to make a Java
  installation available for running Gradle.
- Specifying an alternate Gradle home directory
- Setting of abitrary values for gradle.properties stored inside the
  Gradle home directory.
- Defining init scripts either by pointing to a directory that will be
  linked to init.d inside the Gradle home directory or by specifying
  a list of init script definitions.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/gradle.nix
+++ b/modules/programs/gradle.nix
@@ -1,0 +1,114 @@
+{ config, pkgs, lib, ... }:
+
+with lib;
+
+let
+  cfg = config.programs.gradle;
+  defaultHomeDirectory = ".gradle";
+  settingsFormat = pkgs.formats.javaProperties { };
+
+  initScript = types.submodule ({ name, config, ... }: {
+    options = {
+      text = mkOption {
+        type = types.nullOr types.lines;
+        default = null;
+        description = ''
+          Text of the init script file. if this option is null
+          then `source` must be set.
+        '';
+      };
+
+      source = mkOption {
+        type = types.path;
+        description = ''
+          Path of the init script file. If
+          `text` is non-null then this option will automatically point
+          to a file containing that text.
+        '';
+      };
+    };
+
+    config.source = mkIf (config.text != null) (mkDefault (pkgs.writeTextFile {
+      inherit (config) text;
+      name = hm.strings.storeFileName name;
+    }));
+  });
+in {
+  meta.maintainers = [ maintainers.britter ];
+
+  options.programs.gradle = {
+    enable = mkEnableOption "Gradle Build Tool";
+
+    home = mkOption {
+      type = types.str;
+      default = defaultHomeDirectory;
+      description = ''
+        The Gradle home directory, relative to [](#opt-home.homeDirectory).
+
+        If set, the {env}`GRADLE_USER_HOME` environment variable will be
+        set accordingly. Defaults to {file}`.gradle`.
+      '';
+    };
+
+    package = mkPackageOption pkgs "gradle" { example = "pkgs.gradle_7"; };
+
+    settings = mkOption {
+      type = types.submodule { freeformType = settingsFormat.type; };
+      default = { };
+      example = literalExpression ''
+        {
+          "org.gradle.caching" = true;
+          "org.gradle.parallel" = true;
+          "org.gradle.jvmargs" = "-XX:MaxMetaspaceSize=384m";
+          "org.gradle.home" = pkgs.jdk17;
+        };
+      '';
+      description = ''
+        Key value pairs to write to {file}`gradle.properties` in the Gradle
+        home directory.
+      '';
+    };
+
+    initScripts = mkOption {
+      type = with types; attrsOf initScript;
+      default = { };
+      example = literalExpression ''
+        {
+          "maven-local.gradle".text = '''
+              allProject {
+                repositories {
+                  mavenLocal()
+                }
+              }
+          ''';
+          "another.init.gradle.kts".source = ./another.init.gradle.kts;
+        }
+      '';
+      description = ''
+        Definition of init scripts to link into the Gradle home directory.
+
+        For more information about init scripts, including naming conventions
+        see https://docs.gradle.org/current/userguide/init_scripts.html.
+      '';
+    };
+  };
+
+  config = let gradleHome = "${config.home.homeDirectory}/${cfg.home}";
+  in mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    home.file = mkMerge ([{
+      "${cfg.home}/gradle.properties" = mkIf (cfg.settings != { }) {
+        source = settingsFormat.generate "gradle.properties" cfg.settings;
+      };
+    }]
+      ++ mapAttrsToList (k: v: { "${cfg.home}/init.d/${k}".source = v.source; })
+      cfg.initScripts);
+
+    home.sessionVariables = mkIf (cfg.home != defaultHomeDirectory) {
+      GRADLE_USER_HOME = gradleHome;
+    };
+
+    programs.java.enable = true;
+  };
+}
--- a/tests/modules/programs/gradle/alternate-home-settings.nix
+++ b/tests/modules/programs/gradle/alternate-home-settings.nix
@@ -1,0 +1,26 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+{
+  config = {
+    programs.gradle = {
+      enable = true;
+      home = ".gbt";
+      settings = { "org.gradle.caching" = true; };
+      initScripts = { "some-script.gradle".text = "println 'hello world'"; };
+    };
+
+    programs.java.package =
+      pkgs.runCommandLocal "java" { home = ""; } "mkdir $out";
+
+    test.stubs.gradle = { };
+
+    nmt.script = ''
+      assertFileContains home-path/etc/profile.d/hm-session-vars.sh \
+        'export GRADLE_USER_HOME="/home/hm-user/.gbt"'
+      assertFileExists home-files/.gbt/gradle.properties
+      assertFileExists home-files/.gbt/init.d/some-script.gradle
+    '';
+  };
+}
--- a/tests/modules/programs/gradle/example-settings.nix
+++ b/tests/modules/programs/gradle/example-settings.nix
@@ -1,0 +1,42 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+{
+  config = {
+    programs.gradle = {
+      enable = true;
+      settings = {
+        "org.gradle.caching" = true;
+        "org.gradle.parallel" = true;
+        "org.gradle.java.home" = pkgs.jdk17;
+        "org.gradle.java.installations.paths" = "${pkgs.jdk8},${pkgs.jdk11}";
+      };
+    };
+
+    programs.java.package =
+      pkgs.runCommandLocal "java" { home = ""; } "mkdir $out";
+
+    test.stubs = {
+      gradle = { };
+      jdk = { };
+      jdk8 = { };
+      jdk11 = { };
+      jdk17 = { };
+    };
+
+    nmt.script = ''
+      assertFileExists home-files/.gradle/gradle.properties
+      assertFileContent home-files/.gradle/gradle.properties ${
+        builtins.toFile "gradle.expected" ''
+          # Generated with Nix
+
+          org.gradle.caching = true
+          org.gradle.java.home = @jdk17@
+          org.gradle.java.installations.paths = @jdk8@,@jdk11@
+          org.gradle.parallel = true
+        ''
+      }
+    '';
+  };
+}
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -1341,6 +1341,13 @@ in
           A new module is available: 'programs.sapling'.
         '';
       }
+
+      {
+        time = "2023-12-20T11:41:10+00:00";
+        message = ''
+          A new module is available: 'programs.gradle'.
+        '';
+      }
     ];
   };
 }
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -99,6 +99,7 @@ let
     ./programs/gnome-terminal.nix
     ./programs/go.nix
     ./programs/gpg.nix
+    ./programs/gradle.nix
     ./programs/granted.nix
     ./programs/havoc.nix
     ./programs/helix.nix
--- a/tests/default.nix
+++ b/tests/default.nix
@@ -84,6 +84,7 @@ import nmt {
     ./modules/programs/git
     ./modules/programs/git-cliff
     ./modules/programs/gpg
+    ./modules/programs/gradle
     ./modules/programs/granted
     ./modules/programs/helix
     ./modules/programs/himalaya
--- a/tests/modules/programs/gradle/default.nix
+++ b/tests/modules/programs/gradle/default.nix
@@ -1,0 +1,7 @@
+{
+  gradle-empty-settings = ./empty-settings.nix;
+  gradle-example-settings = ./example-settings.nix;
+  gradle-init-scripts-settings = ./init-scripts-settings.nix;
+  gradle-alternate-home-settings = ./alternate-home-settings.nix;
+}
+
--- a/tests/modules/programs/gradle/empty-settings.nix
+++ b/tests/modules/programs/gradle/empty-settings.nix
@@ -1,0 +1,17 @@
+{ pkgs, ... }:
+
+{
+  config = {
+    programs.gradle.enable = true;
+
+    programs.java.package =
+      pkgs.runCommandLocal "java" { home = ""; } "mkdir $out";
+
+    test.stubs.gradle = { };
+
+    nmt.script = ''
+      assertPathNotExists home-files/.gradle
+      assertFileNotRegex home-path/etc/profile.d/hm-session-vars.sh 'GRADLE_USER_HOME'
+    '';
+  };
+}
--- a/tests/modules/programs/gradle/external-init-script.gradle
+++ b/tests/modules/programs/gradle/external-init-script.gradle
@@ -1,0 +1,1 @@
+println 'external-init-script-1'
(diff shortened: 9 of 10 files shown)


britter (commented):
This is my first contribution to the nix eco system. I'm still learning nix. Hope this makes sense!

liff:
Would it be possible to add another setting for customizing the dot directory? The environment variable `GRADLE_USER_HOME` can be used to change `~/.gradle` to something else. Something like `programs.gradle.home` perhaps, that sets `home.sessionVariables.GRADLE_USER_HOME` accordingly?

modules/programs/gradle.nix
```diff
@@ -0,0 +1,106 @@
...
+      default = [ ];
+      example = "[ pkgs.jdk8 pkgs.jdk11 ]";
+      description = ''
+        Additional Java packages to make available to Gradle's Java toolchain infrastructure.
+        The configured packages will be made available by setting the org.gradle.java.installations.paths
+        property in ~/.gradle/gradle.properties.
+      '';
+    };
+    gradleProperties = mkOption {
+      type = types.attrsOf types.str;
```
liff, line 49: Would it be possible to use [RFC 0042-style](https://github.com/NixOS/rfcs/pull/42) settings here?

There is an example in NixOS’s [Apache Kafka module](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/misc/apache-kafka.nix#L37-L50).

britter:
@liff I'll have a look into both your suggestions. Thank you for the feedback!

modules/programs/gradle.nix
britter: Done!

modules/programs/gradle.nix
```diff
@@ -0,0 +1,89 @@
...
+  options.programs.gradle = {
+    enable = mkOption {
+      type = types.bool;
+      default = false;
+    };
+    home = mkOption {
+      type = types.nullOr types.path;
+      default = null;
+      description = ''
+        The Gradle home directory.
```
liff, line 20: Since this is managed by `home.file`, it’ll have to be a relative path:

```suggestion
        The Gradle home directory, relative to `home.homeDirectory`.
```

modules/programs/gradle.nix
```diff
@@ -0,0 +1,87 @@
...
+      name = "gradle-init-scripts";
+      paths = mapAttrsToList
+        (name: content: pkgs.writeTextDir "init.d/${name}" content)
+        cfg.initScripts;
+    };
+  in mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+    home.file.".gradle/init.d" =
+      mkIf (cfg.initScripts != { }) { source = "${initScripts}/init.d"; };
+    home.sessionVariables =
```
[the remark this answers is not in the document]
liff: You could maybe check the contents of `home-path/etc/profile.d/hm-session-vars.sh`? That’s what the [`sessionVariables` test does](https://github.com/nix-community/home-manager/blob/948703f3e71f1332a0cb535ebaf5cb14946e3724/tests/modules/home-environment/session-variables.nix).

modules/programs/gradle.nix
```diff
@@ -0,0 +1,106 @@
...
+        "org.gradle.caching" = "true";
+        "org.gradle.parallel" = "true";
+        "org.gradle.jvmargs" = "-XX:MaxMetaspaceSize=384m";
+      };
+      description = ''
+        Key value pairs to write to ~/.gradle/gradle.properties.
+      '';
+    };
+    initScripts = mkOption {
+      type = types.attrsOf types.str;
```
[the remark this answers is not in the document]
britter: Done

britter:
@liff this is now ready for the final review / merge. Thank you for your support. I learned a lot along the way. ❤️

modules/programs/gradle.nix
```diff
@@ -0,0 +1,134 @@
...
+      '';
+    };
+    initScripts = mkOption {
+      type = types.either types.path (types.listOf initScript);
+      default = [ ];
+      example = literalExpression ''
+        [
+          {
+            name = "maven-local.gradle"
+            text = ${"''"}
```
liff, line 79: `''` can be [quoted using a third `'`](https://nixos.org/manual/nix/stable/language/values.html?highlight=multi%20line#type-string).

```suggestion
            text = '''
```

modules/programs/gradle.nix
```diff
@@ -0,0 +1,134 @@
...
+          name = "gradle-init-scripts";
+          paths =
+            map (script: pkgs.writeTextDir "init.d/${script.name}" script.text)
+            cfg.initScripts;
+        };
+      in "${initScriptDirectory}/init.d";
+    gradleHome = "${config.home.homeDirectory}/${cfg.home}";
+  in mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+    home.file."${gradleHome}/init.d" =
```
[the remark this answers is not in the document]
britter: You're right and this shows there's not test case covering this code path. I will add one.

modules/programs/gradle.nix
```diff
@@ -0,0 +1,134 @@
...
+        [
+          {
+            name = "maven-local.gradle"
+            text = ${"''"}
+              allProject {
+                repositories {
+                  mavenLocal()
+                }
+              }
+            ${"''"};
```
britter, line 85:
```suggestion
            ''';
```

modules/programs/gradle.nix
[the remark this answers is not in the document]
britter: Actually this should make the init script tests fail. 🤔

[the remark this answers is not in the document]
britter: @liff I added a test for this explicitly, but it was green even with out this change 🤔 Maybe `home.file` detects this case and then removes `config.home.homeDirectory` from the path again? Anyways, I've applied your suggestion because I think it's easier to understand.

liff (approved):

britter:
@liff anything missing in order to get this merged?

liff:
Looks good to me, but I lack the power to merge 😔

britter:
@liff do you know by any chance who to ping to get this merged?

liff:
I guess it would be @rycee by default? Or you could ask on the [IRC channel](https://nix-community.github.io/home-manager/index.xhtml#preface).

It occurred to me that since this is a new module, perhaps it’s worthy of a [news entry](https://nix-community.github.io/home-manager/index.xhtml#sec-news). New modules usually seem to get one.

modules/programs/gradle.nix
```diff
@@ -0,0 +1,134 @@
...
+        "org.gradle.parallel" = true;
+        "org.gradle.jvmargs" = "-XX:MaxMetaspaceSize=384m";
+        "org.gradle.home" = pkgs.jdk17;
+      };
+      description = ''
+        Key value pairs to write to gradle.properties in the Gradle home directory.
+      '';
+    };
+    initScripts = mkOption {
+      type = types.either types.path (types.listOf initScript);
```
rycee, line 73: Perhaps `type = with types; attrsOf initScript`?

Would look something like

``` nix
initScripts = {
  "maven-local.gradle".text = "…";
  "another.init.gradle.kts".source = ./another.init.gradle.kts;
};
```

tests/modules/programs/gradle/alternate-home-settings.nix
```diff
@@ -0,0 +1,25 @@
...
+      initScripts = [{
+        name = "some-script.gradle";
+        text = "println 'hello world'";
+      }];
+    };
+
+    test.stubs.gradle = { };
+
+    nmt.script = ''
+      assertFileRegex home-path/etc/profile.d/hm-session-vars.sh 'export GRADLE_USER_HOME="/home/hm-user/.gbt"'
```
rycee, line 20: Perhaps

```suggestion
      assertFileContains home-path/etc/profile.d/hm-session-vars.sh \
        'export GRADLE_USER_HOME="/home/hm-user/.gbt"'
```

tests/modules/programs/gradle/example-settings.nix
```diff
@@ -0,0 +1,33 @@
...
+with lib;
+
+{
+  config = {
+    programs.gradle = {
+      enable = true;
+      settings = {
+        "org.gradle.caching" = true;
+        "org.gradle.parallel" = true;
+        "org.gradle.java.home" = pkgs.jdk17;
```
rycee, line 12: Unfortunately you'll need to stub these jdk packages (or just use some other mocked packages), otherwise the test downloads will be enormous just to insert a short text into the output file.

modules/programs/gradle.nix
```diff
@@ -0,0 +1,134 @@
...
+      text = mkOption {
+        type = types.lines;
+        default =
+          if config.source != null then builtins.readFile config.source else "";
+        defaultText = literalExpression
+          "if source is defined, the content of source, otherwise empty";
+        description = ''
+          Text of the init script file.
+          If unset then the source option will be preferred.
+        '';
```
rycee, line 30: This is too complicated w.r.t merging and will cause confusion, and likely infinite recursion in some cases. I would suggest using a simpler setup similar to what is in `file-type.nix`:

```suggestion
        type = types.nullOr types.lines;
        default = null;
        description = ''
          Text of the init script file.
          If unset then the source option must be specified.
        '';
```

modules/programs/gradle.nix
```diff
@@ -0,0 +1,134 @@
...
+
+      source = mkOption {
+        type = types.nullOr types.path;
+        default = null;
+        description = ''
+          Path of the init script file to use.
+          If the text option is set, it will be preferred.
+        '';
+      };
+    };
```
rycee, line 41: Add `config` section where source is populated with the generated file, see `file-type.nix`.

modules/programs/gradle.nix
```diff
@@ -0,0 +1,134 @@
...
+        "org.gradle.caching" = true;
+        "org.gradle.parallel" = true;
+        "org.gradle.jvmargs" = "-XX:MaxMetaspaceSize=384m";
+        "org.gradle.home" = pkgs.jdk17;
+      };
+      description = ''
+        Key value pairs to write to gradle.properties in the Gradle home directory.
+      '';
+    };
+    initScripts = mkOption {
```
rycee, line 72:
```suggestion

    initScripts = mkOption {
```

modules/programs/gradle.nix
```diff
@@ -0,0 +1,134 @@
...
+        '';
+      };
+    };
+  });
+in {
+  meta.maintainers = [ maintainers.britter ];
+
+  options.programs.gradle = {
+    enable = mkEnableOption "Gradle Build Tool";
+    home = mkOption {
```
rycee, line 48:
```suggestion

    home = mkOption {
```

modules/programs/gradle.nix
rycee: … when `mkIf (config.text != null)`, that is… 🙂

modules/programs/gradle.nix
```diff
@@ -0,0 +1,134 @@
...
+      type = types.str;
+      default = defaultHomeDirectory;
+      description = ''
+        The Gradle home directory, relative to `home.homeDirectory`.
+
+        If set, the GRADLE_USER_HOME environment variable will be
+        set accordingly. Defaults to .gradle.
```
rycee, line 55: Can add some fancy markup.

```suggestion
        The Gradle home directory, relative to [](#opt-home.homeDirectory).

        If set, the {env}`GRADLE_USER_HOME` environment variable will be
        set accordingly. Defaults to {file}`.gradle`.
```

modules/programs/gradle.nix
```diff
@@ -0,0 +1,134 @@
...
+      '';
+    };
+    package = mkPackageOption pkgs "gradle" { example = "pkgs.gradle_7"; };
+    settings = mkOption {
+      type = types.submodule { freeformType = settingsFormat.type; };
+      default = { };
+      example = {
```
rycee, line 62: I think this will look best as a literal expression:

```suggestion
      example = literalExpression ''
        {
```

modules/programs/gradle.nix
```diff
@@ -0,0 +1,134 @@
...
+          name = "gradle-init-scripts";
+          paths =
+            map (script: pkgs.writeTextDir "init.d/${script.name}" script.text)
+            cfg.initScripts;
+        };
+      in "${initScriptDirectory}/init.d";
+    gradleHome = "${config.home.homeDirectory}/${cfg.home}";
+  in mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+    home.file."${cfg.home}/init.d" =
```
rycee, line 124: Can add an empty line between the bindings to make it less compact:
```suggestion

    home.file."${cfg.home}/init.d" =
```

modules/programs/gradle.nix
```diff
@@ -0,0 +1,134 @@
...
+      mkIf (cfg.initScripts != [ ]) { source = initScripts; };
+    home.sessionVariables = mkIf (cfg.home != defaultHomeDirectory) {
+      GRADLE_USER_HOME = gradleHome;
+    };
+    home.file."${cfg.home}/gradle.properties" = mkIf (cfg.settings != { }) {
+      source = (settingsFormat.generate "gradle.properties" cfg.settings);
```
rycee, line 130:
```suggestion
      source = settingsFormat.generate "gradle.properties" cfg.settings;
```

modules/programs/gradle.nix
rycee: To be clear, with the above I also suggest removing the possibility to specify just a path. Instead always require an attribute set. It will help future problems with merging. Specifically, it is unclear what

``` nix
mkMerge [
  { initScripts = ./path; }
  { initScripts."extra-init.gradle".source = ./path2; }
]
```

should do.

You could use `apply` to coerce `./path` to `{ source = ./path; }` but I generally think it is best to avoid `apply` to maintain predictability w.r.t merging.

tests/modules/programs/gradle/example-settings.nix
rycee: Had a look at this and unfortunately it is a bit tricky to accomplish this in this case. For now, I think the best possible solution is

``` patch
diff --git a/tests/modules/programs/gradle/example-settings.nix b/tests/modules/programs/gradle/example-settings.nix
index 36b9736f..19b8c070 100644
--- a/tests/modules/programs/gradle/example-settings.nix
+++ b/tests/modules/programs/gradle/example-settings.nix
@@ -14,17 +14,26 @@ with lib;
       };
     };
 
-    test.stubs.gradle = { };
+    programs.java.package =
+      pkgs.runCommandLocal "java" { home = ""; } "mkdir $out";
+
+    test.stubs = {
+      gradle = { };
+      jdk = { };
+      jdk8 = { };
+      jdk11 = { };
+      jdk17 = { };
+    };
 
     nmt.script = ''
       assertFileExists home-files/.gradle/gradle.properties
       assertFileContent home-files/.gradle/gradle.properties ${
-        pkgs.writeText "gradle.expected" ''
+        builtins.toFile "gradle.expected" ''
           # Generated with Nix
 
           org.gradle.caching = true
-          org.gradle.java.home = ${pkgs.jdk17}
-          org.gradle.java.installations.paths = ${pkgs.jdk8},${pkgs.jdk11}
+          org.gradle.java.home = @jdk17@
+          org.gradle.java.installations.paths = @jdk8@,@jdk11@
           org.gradle.parallel = true
         ''
       }
```

rycee:
Thanks for the contribution! I've added a bunch of mostly stylistic comments. The most important is to reduce the test closure to avoid GBs of downloads whenever the tests run 🙂

modules/programs/gradle.nix
britter: Hey @rycee 👋🏻

My thinking was that if I had multiple init scripts, I'd like to reference a directory instead of listing them all. But I now understand that this doesn't work with merging configurations. I will change this.

modules/programs/gradle.nix
britter: I adopted this from the [nushell module](https://github.com/nix-community/home-manager/blob/948703f3e71f1332a0cb535ebaf5cb14946e3724/modules/programs/nushell.nix#L64-L82) as suggested [here](https://github.com/nix-community/home-manager/pull/4667#discussion_r1413594398). Maybe that also needs to be fixed?

modules/programs/gradle.nix
```diff
@@ -0,0 +1,134 @@
...
+    package = mkPackageOption pkgs "gradle" { example = "pkgs.gradle_7"; };
+    settings = mkOption {
+      type = types.submodule { freeformType = settingsFormat.type; };
+      default = { };
+      example = {
+        "org.gradle.caching" = true;
+        "org.gradle.parallel" = true;
+        "org.gradle.jvmargs" = "-XX:MaxMetaspaceSize=384m";
+        "org.gradle.home" = pkgs.jdk17;
+      };
```
britter, line 68:
```suggestion
          "org.gradle.caching" = true;
          "org.gradle.parallel" = true;
          "org.gradle.jvmargs" = "-XX:MaxMetaspaceSize=384m";
          "org.gradle.home" = pkgs.jdk17;
        };
      ''
```

tests/modules/programs/gradle/example-settings.nix
britter: Done!

modules/programs/gradle.nix
britter: Done

modules/programs/gradle.nix
britter: Wow, this was way more tricky than I anticipated. But it's done now.

britter:
@rycee I think I addressed all your comments. Previously the code aggregated all init scripts in a store directory using `symlinkJoin`. I was unable to get that working with the new structure of the initScript submodule where the source is set from the text. So the current version links individual files from the store inside `.gradle/init.d`. Unfortunately I didn't find a more elegant solution than `home.file = mkMerge ...` which feels more complex than it should.

I also added a news entry.

rycee:

PR #4670, opened 2023-11-17
tmux: Add `extraConfigBeforePlugins` to `tmux.conf`

gabyx:
NixOS also has this option which make configuration easier.

### Description

Add  `extraConfigBeforePlugins` to add additional configuration before the plugins are loaded.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/programs/tmux.nix
+++ b/modules/programs/tmux.nix
@@ -1,9 +1,6 @@
 { config, lib, pkgs, ... }:
-
 with lib;
-
 let
-
   cfg = config.programs.tmux;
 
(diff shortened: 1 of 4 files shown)


gabyx:
The file `tmux.nix` was not formatted obviously. Not it is...

teto:
the latest conventions for HM is  to expose a `lines` option where the user has total control on the order via `mkOrder`, mkBefore, mkAfter etc. See neovim's initLua or zsh's `initContent`. Feel free to resubmit with that approach,

PR #4671, opened 2023-11-18
signaturepdf: add service

DamienCassou:
### Description

New signaturepdf service. https://github.com/24eme/signaturepdf/

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/signaturepdf.nix
+++ b/modules/services/signaturepdf.nix
@@ -1,0 +1,70 @@
+{ config, lib, pkgs, ... }:
+
+let
+  cfg = config.services.signaturepdf;
+  extraConfigToArgs = extraConfig:
+    lib.flatten
+    (lib.mapAttrsToList (name: value: [ "-d" "${name}=${value}" ]) extraConfig);
+in {
+  meta.maintainers = [ lib.maintainers.DamienCassou ];
+
+  options.services.signaturepdf = with lib; {
+    enable = mkEnableOption
+      "signaturepdf; signing, organizing, editing metadatas or compressing PDFs";
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.signaturepdf;
+      defaultText = "pkgs.signaturepdf";
+      description = "signaturepdf derivation to use.";
+    };
+
+    port = mkOption {
+      type = types.port;
+      default = 8080;
+      example = 8081;
+      description = "The port on which the application runs";
+    };
+
+    extraConfig = mkOption {
+      default = { };
+      type = with types;
+        let primitive = oneOf [ str int bool float ];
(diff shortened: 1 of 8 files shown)


modules/services/signaturepdf.nix
```diff
@@ -0,0 +1,68 @@
...
+let
+  cfg = config.services.signaturepdf;
+  extraConfigToArgs = extraConfig:
+    lib.flatten
+    (lib.mapAttrsToList (name: value: [ "-d" "${name}=${value}" ]) extraConfig);
+in {
+  meta.maintainers = [ lib.maintainers.DamienCassou ];
+
+  options.services.signaturepdf = with lib; {
+    enable = mkEnableOption "signaturepdf, the clipboard management daemon";
```
rycee, line 12: Maybe needs a new description? I think this is not actually a clipboard management daemon?

modules/services/signaturepdf.nix
```diff
@@ -0,0 +1,68 @@
...
+    };
+
+    port = mkOption {
+      type = types.port;
+      default = 8080;
+      example = 8081;
+      description = "The port on which the application runs";
+    };
+
+    extraConfig = mkOption {
```
rycee, line 28: I guess this should be 
```suggestion
    settings = mkOption {
```
and the type probably should be something like
```nix
with types;
  let primitive = oneOf [ str int bool float ];
  in attrsOf primitive;
```

modules/services/signaturepdf.nix
DamienCassou: indeed, sorry!

modules/services/signaturepdf.nix
DamienCassou: good idea, done.

DamienCassou:
The build is failing because nixos-unstable doesn't contain signaturepdf yet (I just merged it this morning). I will let this PR as draft until this is solved.

ambroisie:
@DamienCassou just curious, I see that you [just packaged it](https://github.com/NixOS/nixpkgs/pull/266336). Why didn't you create a NixOS module for it? It seems more appropriate for what is supposed to be a self-hosted service IMO.

DamienCassou:
> Why didn't you create a NixOS module for it?

First, I (unfortunately) don't use NixOS. Second, I think that every service that can be defined in home-manager should because (1) it helps non-NixOS users and (2) it lets users control the service without requiring a root access on the machine.

DamienCassou:
@rycee signaturepdf has been added to nixos-unstable and this PR can now be reviewed. Thank you.

rycee:

PR #4673, opened 2023-11-19
render documentation without deprecated optionsDocBook

### LINKED ISSUE #4273: bug: Getting 'optionsDocBook is deprecated since 23.11 and will be removed in 24.05' on most runs

### Issue description

**Home Manager installation type**: Standalone + Flake (also happens without the flake)

I seem to get

```
trace: warning: optionsDocBook is deprecated since 23.11 and will be removed in 24.05
trace: warning: optionsDocBook is deprecated since 23.11 and will be removed in 24.05
trace: warning: optionsDocBook is deprecated since 23.11 and will be removed in 24.05
```

On almost every run, the error seems to originate from https://sourcegraph.com/github.com/NixOS/nixpkgs/-/blob/nixos/lib/make-options-doc/default.nix?L160

This doesn't seem to have any effect at the moment, but thought I'd post an issue

### Maintainer CC

_No response_

### System information

```markdown
- system: `"aarch64-darwin"`
- host os: `Darwin 22.6.0, macOS 13.5`
- multi-user?: `yes`
- sandbox: `no`
- version: `nix-env (Nix) 2.16.1`
- channels(root): `"nixpkgs"`
```

considerate:
### Description

This PR replaces the DocBook-based manual generation with documentation generation that is not deprecated in nixpkgs.

- [x] Fix broken links in manual
- [x] Reproduce the manpage for `man home-manager`

Closes #4273


quinn-dougherty:
Woot! Thank you.

considerate:
I'm not sure what to do about `man home-manager`. Should I
1. Keep the old docbook format to generate the manpage?
2. Just stage an [mdoc](https://man.openbsd.org/mdoc.7) version of the manpage directly?
3. Try some kind of markdown alternative similar to how the manual is generated?

It looks like `nix` and `nixpkgs` (for NixOS) does alternative 2. Personally, I'm not proficient in mdoc so I would only be able to do simple fixups, but more realistically just copy the generated manpage from the docbook as on `master`.

considerate:
@emilazy I'd really appreciate your review on this to see if this roughly aligns with what you intended to do about the documentation when you posted here https://github.com/nix-community/home-manager/issues/4273#issuecomment-1650687165.

rycee:
Thanks for the contribution! It seems like a lot of work has gone into it. My understanding, though, was that nixos-render-docs would be fixed to accommodate our use-case. I'm very uneager to introduce packages like these in the repository, would it not be possible to generalize nixos-render-docs instead?

emilazy:
Hey, I'm so sorry that I didn't get around to finishing this myself! Life got in the way and I haven't had the time for any Nix-related work recently. Thank you for picking up where I left off. Given 23.11 is imminent I will try to look over this in the next few days and help out with what I can from my own work on this; I had things mostly working locally but there were a few blockers I still needed to resolve. Are you available Matrix to discuss this work?

And just to confirm to @rycee, nixos-render-docs is indeed suitable for purpose here; I had the manual rendering almost identically to how it was before with no nmd or custom packages, and agreement in principle with upstream on the minor changes to nixos-render-docs that would be required to avoid some slight divergences and hacks. Hopefully we can get that sorted or incorporate the necessary post-processing hacks in HM until it's possible to get upstream changes merged.

considerate:
@rycee @emilazy Thank you very much for taking a look.

I'm in total agreement that it'd be better to avoid patching nixos-render-docs here and instead make upstream more flexible to allow the minor changes we need here. My reasoning is that I wanted to avoid getting blocked on getting something merged into nixpkgs before getting the manual in a workable state for home-manager.

@emilazy Thank you, I appreciate you taking time to review this. I have a Matrix account but I haven't really been active there so I'll take a look tonight (JST). I noticed some things are now conflicting on the manual and I'll to to fix that up as well.

considerate:
Is it ok to base this PR on `release-23.11` or should this first be merged into `master` first and then backported?

I had a look at the [README](https://github.com/nix-community/home-manager/blob/8cedd63eede4c22deb192f1721dd67e7460e1ebe/README.md?plain=1#L32) but wasn't sure if this applies as this isn't really a new module.

rycee:
Thanks @emilazy and @considerate, much appeciated.

I would think if these changes are intended for the master branch as well, then it would be best to merge there first and then backport to the release-23.11 branch.

rycee:
… I assume any backport to the 23.11 branch also would require the nixos-render-docs changes to be backported in Nixpkgs? If so, that complicates things, I imagine 😕

considerate:
This is a draft for now since the `--header` and `--footer` flags for `nixos-render-docs` haven't propagated to nixos-unstable yet.

considerate:
@rycee It looks like the changes are on `release-23.11` but not on any `nixos-` branch. Since home-manager depends on `nixos-unstable` I suppose we should keep on waiting until the `nixos-` branches get the changes as well?

wkral:
Looks like the `--header` and `--footer` flags are in `nixos-unstable` now too: https://nixpk.gs/pr-tracker.html?pr=269942

docs/home-manager-manual.nix
```diff
@@ -0,0 +1,57 @@
+{ stdenv, lib, documentation-highlighter, nmd, revision, home-manager-options
+, nixos-render-docs }:
+let outputPath = "share/doc/home-manager";
+in stdenv.mkDerivation {
+  name = "nixpkgs-manual";
```
rycee, line 5: I think
```suggestion
  name = "home-manager-manual";
```

docs/manual/writing-modules/types.md
```diff
@@ -0,0 +1,368 @@
+# Option Types {#sec-option-types}
+
+Overall the basic option types are the same in Home Manager as NixOS. A
+few Home Manager options, however, make use of custom types that are
+worth describing in more detail. These are the option types `dagOf` and
+`gvariant` that are used, for example, by
+[programs.ssh.matchBlocks](#opt-programs.ssh.matchBlocks) and [dconf.settings](#opt-dconf.settings).
+
+[]{#sec-option-types-dag}`hm.types.dagOf`
```
rycee, line 9: This seems to generate invalid HTML (my formatting):

``` html
<span class="term">
  <a id="sec-option-types-dag" />
  <code class="literal">hm.types.dagOf</code>
</span>
```

Firefox complains about this use of `<a>`, saying

> Self-closing /> is invalid syntax on non-void HTML elements.

The old manual contains

``` html
<span class="term">
  <a id="sec-option-types-dag"></a>
  <code class="literal">hm.types.dagOf</code>
</span>
```

and Firefox is happy with that 🙂

Doesn't matter much for this PR but may be worth updating the generator in Nixpkgs?

docs/manual/usage/configuration.md
```diff
@@ -0,0 +1,112 @@
...
+}
+```
+
+You can use this as a base for your further configurations.
+
+::: note
+If you are not very familiar with the Nix language and NixOS modules
+then it is encouraged to start with small and simple changes. As you
+learn you can gradually grow the configuration with confidence.
+:::
```
rycee, line 36: Doesn't seem to render correctly:
swappy-20231129_230226

rycee: The other `::: note` elements are rendered similarly…

docs/manual/contributing/getting-started.md
```diff
@@ -0,0 +1,43 @@
...
+push the branch to GitHub and [create a pull
+request](https://help.github.com/articles/creating-a-pull-request/).
+
+Assuming your clone is at `$HOME/devel/home-manager` then you can make
+the `home-manager` command use it by either
+
+1.  overriding the default path by using the `-I` command line option:
+
+    ``` console
+    $ home-manager -I home-manager=$HOME/devel/home-manager
```
rycee, line 21: The generated code is slightly different so the custom CSS that makes the `$` non-selectable doesn't trigger. The code https://git.sr.ht/~rycee/nmd/tree/master/item/static/style.scss#L297-300 should probably be changed to something like

``` css
.programlisting.language-shell .hljs-meta.prompt_ {
    user-select: none;
}
```

I can have a look at fixing that.

docs/manual/usage.md
```diff
@@ -0,0 +1,63 @@
...
+
+then building it, for example using `home-manager build`, will result in
+an error message saying something like
+
+``` console
+$ home-manager build
+error: A definition for option `programs.emacs.enable' is not of type `boolean'. Definition values:
+- In `/home/jdoe/.config/home-manager/home.nix': "yes"
+(use '--show-trace' to show detailed location information)
+```
```
rycee, line 38: I wonder if the Markdown renderer actually supports the `console` language? For this block the generated classes are `programlisting hljs language-bash` which won't work:
swappy-20231129_231930

In the existing manual this block gets the classes `programlisting console hljs language-shell`.

docs/manual/usage/dotfiles.md
```diff
@@ -0,0 +1,36 @@
...
+  # …
+
+  programs.git = {
+    enable = true;
+    userName = "Jane Doe";
+    userEmail = "jane.doe@example.org";
+  };
+
+  # …
+}
```
rycee, line 24: Hmm, doesn't get rendered as Nix code for me. The classes in Firefox are `programlisting hljs language-bash`.

Had a quick look at the [generator code](https://github.com/NixOS/nixpkgs/blob/dc55b556b31a24f2a3e4c5d1c831b4c52f523746/pkgs/tools/nix/nixos-render-docs/src/nixos_render_docs/html.py#L99-L101) and it seems it ignores the block language?

docs/manual/3rd-party.md
```diff
@@ -0,0 +1,14 @@
+# Third-Party Tools and Extensions {#ch-3rd-party}
```
rycee, line 1: I don't see this chapter show up in the built manual?

rycee:
Thanks, much nicer without having the Python code in this repo! I wish the Python code could be moved out of Nixpkgs as well, into it's own repo. I don't understand why it needs to be bundled.

In any case. Overall the output looks quite good!

I think the only cosmetic blocker is the issue with `<a id="blah"/>` since it causes pretty awful rendering.

The content blockers would be the missing third-party extensions and release notes chapters. I wish the `home-manager` man page could be included as well but it seems difficult to accomplish, and I don't know what to do about the man page in general, DocBook simply seems like the best option for that.

docs/home-configuration-nix-header.5
```diff
@@ -0,0 +1,9 @@
+.TH "HOME-CONFIGURATION\&.NIX" "5" "01/01/1980" "Home Manager"
+.\" disable hyphenation
+.nh
+.\" disable justification (adjust text to left margin only)
+.ad l
+.\" enable line breaks after slashes
+.cflags 4 /
+.SH "NAME"
+\fIhome\-configuration\&.nix\fP \- Home Manager configuration specification
```
rycee, line 9: I think this should also include the `DESCRIPTION` and `OPTIONS` headers. Compare
swappy-20231129_235403
and
swappy-20231129_235424

The description section is not so important, it should probably be reworded anyway, but it would be nice to have the `OPTIONS` header.

docs/manual/writing-modules/types.md
rycee: I don't think I had to do anything special to make DocBook produce anchors of the form `<a id="blah"></a>` so I'm not sure why they would be different in Nixpkgs.

In any case, this messes up the rendering pretty badly.

This PR:
swappy-20231130_000523

Current:
swappy-20231130_000835

docs/manual/3rd-party.md
considerate: It looks like I forgot the include in the top-level manual.md file

https://github.com/nix-community/home-manager/blob/2014be5ea9bc7ef0f934cd88ef45cfe4aa8831ae/docs/manual/manual.md?plain=1

docs/manual/usage.md
considerate: It looks like the above bad rendering is because the language used by the `console` snippets is set to `bash`, as in a bash script, not console output.

docs/manual/writing-modules/types.md
considerate: I don't understand why there's an empty anchor tag in the generated manual. This doesn't show in the rendered output. Is this for automated scraping of links?

considerate: I can verify that the nixpkgs manual contains the same broken HTML output:
```
<a id="ssec-gnome-hooks-glib" /> <code class="literal">glib</code> 
```

considerate: Even after changing the Content-Type to `application/xhtml+xml; charset=utf8` it renders like this:
xhtml

considerate: For some reason `firefox` renders this correctly but not the `home-manager` manual. Moreover, fetching the `nixpkgs` manual via curl doesn't have this same issue (I built `nix build .#htmlDocs.nixpkgsManual` on the most recent `master` when producing the above HTML). The web page has
```html
<a id="ssec-gnome-hooks-glib" shape="rect"></a> <code class="literal">glib</code> 
```

docs/manual/contributing/getting-started.md
rycee: If you update the `nmdSrc` rev to `f18defadcc25e69e95b04493ee02682005472255`, then selection issue should now be resolved.

docs/manual/usage.md
rycee: Yeah, as I discovered elsewhere the new generator does not pass on the language tag. So the generated HTML contains

``` html
<pre class="programlisting">…</pre>
```

while the nmd generator emits

``` html
<pre class="programlisting console">…</pre>
```

which makes highlight-js pick the right syntax without having to attempt autodetection.

rycee: @pennae The `html.py` code in nixos-render-docs contains

``` python
    def fence(self, token: Token, tokens: Sequence[Token], i: int) -> str:
        # TODO use token.info. docbook doesn't so we can't yet.
        return f'<pre class="programlisting">\n{escape(token.content)}</pre>'
```

Do you remember why the info cannot be passed on? I'm thinking something along the lines of

```
info = quoteattr(info) if info is not None else ""
return f'<pre class="programlisting{info}">\n{escape(token.content)}</pre>'
```

considerate: @pennae I tried to use that PR but unfortunately `highlightjs` as specified in `nmd` fails to pick up the syntax highlighting altogether when using those changes. I'll see if I can modify the highlighting in nmd to work with the `<code>` blocks.

considerate: @pennae Would it be possible to backport https://github.com/NixOS/nixpkgs/pull/271495 to nixpkgs-23.11?

considerate: With this style override
```css
pre {
  padding: 0;
}

pre code.hljs {
  border: none;
  margin: 0;
}

it renders ok to me. The selection mask on `$` seems to work as well. However, this requires using the above pull request.

```
shell
nix

docs/manual/contributing/getting-started.md
rycee: Yeah, I think adding a commit with

``` patch
diff --git a/docs/default.nix b/docs/default.nix
index 0c2c5516..ecef00c3 100644
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -9,8 +9,8 @@ let
 
   nmdSrc = fetchTarball {
     url =
-      "https://git.sr.ht/~rycee/nmd/archive/824a380546b5d0d0eb701ff8cd5dbafb360750ff.tar.gz";
-    sha256 = "0vvj40k6bw8ssra8wil9rqbsznmfy1kwy7cihvm13rajwdg9ycgg";
+      "https://git.sr.ht/~rycee/nmd/archive/f18defadcc25e69e95b04493ee02682005472255.tar.gz";
+    sha256 = "14fazjqbc6sfsxycwadfhcymi37xg5wmrlvg3jb8d4mpbcdwfwz4";
   };
 
   nmd = import nmdSrc {
```

is worth doing regardless :-)

rycee:
Thanks a lot everybody! Everything looks just fine now so I've merged it to master. If no major issue comes up in the next few days I think we can backport it to the 23.11 branch 🙂

tshipenchko:
@rycee Thanks for your hard work! Is there PR/issue where we can watch progress of backport?

rycee:
@tshipenchko That should be https://github.com/nix-community/home-manager/pull/4712, created by @considerate

PR #4677, opened 2023-11-19
qt: fix basic usage when just `qt.enable = true` is set

thiagokokada:
### Description

After #4579 the usage of the module when just `qt.enable = true` is set was broken with an evaluation error. This PR fixes it.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/qt.nix
+++ b/modules/misc/qt.nix
@@ -159,8 +159,10 @@ in {
 
     # Necessary because home.sessionVariables doesn't support mkIf
     envVars = lib.filterAttrs (n: v: v != null) {
-      QT_QPA_PLATFORMTHEME =
-        styleNames.${cfg.platformTheme} or cfg.platformTheme;
+      QT_QPA_PLATFORMTHEME = if (cfg.platformTheme != null) then
+        styleNames.${cfg.platformTheme} or cfg.platformTheme
+      else
+        null;
       QT_STYLE_OVERRIDE = cfg.style.name;
     };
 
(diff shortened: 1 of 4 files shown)


thiagokokada:
CC @acristoffers @rycee.

acristoffers:
I just tried your branch and it works.

modules/misc/qt.nix
```diff
@@ -206,9 +208,10 @@ in {
...
 
-    home.packages = (platformPackages.${cfg.platformTheme} or [ ])
-      ++ lib.optionals (cfg.style.package != null)
-      (lib.toList cfg.style.package);
+    home.packages = (lib.optionals (cfg.platformTheme != null)
+      platformPackages.${cfg.platformTheme})
+      ++ (lib.optionals (cfg.style.package != null)
+        (lib.toList cfg.style.package));
 
     xsession.importedVariables = [ "QT_PLUGIN_PATH" "QML2_IMPORT_PATH" ]
```
andresilva, line 216: This wasn't added in this PR but might as well ask here. Why do we need to set this if we're already importing these variables into the systemd user session above? AFAIU all that this does is to manually import the variables into systemd on xsession start (https://github.com/nix-community/home-manager/blob/master/modules/xsession.nix#L186-L188).

modules/misc/qt.nix
```diff
@@ -159,8 +159,10 @@ in {
 
     # Necessary because home.sessionVariables doesn't support mkIf
     envVars = lib.filterAttrs (n: v: v != null) {
-      QT_QPA_PLATFORMTHEME =
-        styleNames.${cfg.platformTheme} or cfg.platformTheme;
+      QT_QPA_PLATFORMTHEME = if (cfg.platformTheme != null) then
+        styleNames.${cfg.platformTheme} or cfg.platformTheme
```
andresilva, line 163: This doesn't really work. Try setting `platformTheme = "gtk3"`, it will complain about `gtk3` attribute missing from `styleNames`.

andresilva:
```
       error: attribute 'gtk3' missing

       at /nix/store/91m8mdr4lxwjm3h3gz6xribrd0pi7rvp-source/modules/misc/qt.nix:212:7:

          211|     home.packages = (lib.optionals (cfg.platformTheme != null)
          212|       platformPackages.${cfg.platformTheme})
             |       ^
          213|       ++ (lib.optionals (cfg.style.package != null)
       Did you mean gtk?
```

Probably nix just trying to be too smart 😅.

thiagokokada: Good catch, I forgot to add the fallback here.

Will also add another test, otherwise this get too easy to miss in future changes.

rycee:
Thanks! Merged to master now, hopefully that will resolve the issues people are having 🙂

PR #4689, opened 2023-11-22
gpg-agent: improve pinentry-flavor

patwid:
### Description

Fix pinentry-flavor gnome3 on non-gnome systems. Analogous to what nixos does already. The nixos module goes even further by infering the default pinentry flavor based on the configured desktop environment. Also the nixos module falls back to pinentry-flavor curses rather than gtk2.

Nixos module: https://github.com/NixOS/nixpkgs/blob/f37dd68fe080c16a4e0a0733ca0006d31e6d5ca4/nixos/modules/programs/gnupg.nix

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

Not sure, @dixslyf

### DIFF
--- a/modules/services/gpg-agent.nix
+++ b/modules/services/gpg-agent.nix
@@ -193,15 +193,8 @@ in {
           Which pinentry interface to use. If not
           `null`, it sets
           {option}`pinentry-program` in
-          {file}`gpg-agent.conf`. Beware that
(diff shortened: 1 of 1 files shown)


dixslyf:
Tthis won't work because `services.dbus` is a NixOS option, hence the CI failure. Home Manager does not have an equivalent option AFAIK.

(I'm not a maintainer BTW)

PR #4690, opened 2023-11-23
hyprland: plugin and source with higher priority

globin:
### Description

`plugin` and `source` are moved to beginning of config file.

These often are necessary for other configuration, e.g. defining variables or enabling other layouts

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

cc @fufexan


fufexan (approved):

rycee:
Thanks! Merged to master now. Since the source setting was already moved in an earlier PR I change this to only move the plugin.

PR #4695, opened 2023-11-24
Release commits

rycee:
### Description

Prepare for next release cycle.

### Checklist

- [ ] Change is backwards compatible:

    - The `.release` file has been removed as per the 23.05 release notes.

    ```
    {component}: {description}

    {long description}
    ```


quinn-dougherty:
Are we still gonna break next week if https://github.com/nix-community/home-manager/pull/4673 isn't merged in?

rycee:
@quinn-dougherty I don't know, I haven't followed the status of the necessary changes in nixos-render-docs (as discussed in #4215). My hope was that we could avoid the deprecation warnings in time for the 23.11 release but it seems they will stay for now.

Issue #4637, opened 2023-11-06
bug: SSH_AUTH_SOCK not set in nushell


TornaxO7:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

I have the following options saved in my hm-config:
```nix
   programs = {
       ssh.enable = true;
       nushell.enable = true;
   };

  services.ssh-agent.enable = true;
```
but I'm getting the following error message if I'm trying to ssh, for example, to push to my repository on github:
```
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Also entering `$env.SSH_AUTH_SOCK` gives me:
```nushell
Error: nu::shell::column_not_found

  × Cannot find column
   ╭─[entry #13:1:1]
 1 │ $env.SSH_AUTH_SOCK
   · ──┬─ ──────┬──────
   ·   │        ╰── cannot find column 'SSH_AUTH_SOCK'
   ·   ╰── value originates here
   ╰────
```

# Possible issue
The code in for the `ssh-agent` service [doesn't look like valid nushell code](https://github.com/nix-community/home-manager/blob/master/modules/services/ssh-agent.nix#L22-L26). In `nushell` you'd normally, if I'm correct (I'm still a newbie), set the value through `$env.<variable-name>`.

### Maintainer CC

@lheckemann 

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.1.60, NixOS, 23.11 (Tapir), 23.11.20231102.fa804ed`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.17.1`
 - channels(root): `"nixos-23.05"`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`
```

TornaxO7:
Ok, after trying stuff out, I can't get the ssh-agent working on `zsh` as well if `gpg-agent.enableSshSupport = enable;` and the gpg-agent is enabled as well.

However, the [ssh-support-string in the gpg-agent module](https://github.com/nix-community/home-manager/blob/master/modules/services/gpg-agent.nix#L13) doesn't look like valid nushell-code as well.

lheckemann:
[...]

PR #4700, opened 2023-11-25
Fix/gpg agent ssh nushell

TornaxO7:
### Description

Fixes https://github.com/nix-community/home-manager/issues/4637

Please provide a brief description of your change.

This makes sure that `GPG_TTY` and `SSH_AUTH_SOCK` are set with nushell enabled.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@lheckemann

### DIFF
--- a/modules/services/gpg-agent.nix
+++ b/modules/services/gpg-agent.nix
@@ -22,6 +22,16 @@ let
     set -gx GPG_TTY (tty)
   '' + optionalString cfg.enableSshSupport gpgSshSupportStr;
 
+  gpgNushellInitStr = ''
+    $env.GPG_TTY = (tty)
+  '' + optionalString cfg.enableSshSupport ''
+    ${gpgPkg}/bin/gpg-connect-agent updatestartuptty /bye | ignore
+
+    if not "SSH_AUTH_SOCK" in $env {
+      $env.SSH_AUTH_SOCK = (${gpgPkg}/bin/gpgconf --list-dirs agent-ssh-socket)
+    }
+  '';
+
   # mimic `gpgconf` output for use in `systemd` unit definitions.
   # we cannot use `gpgconf` directly because it heavily depends on system
   # state, but we need the values at build time. original:
@@ -216,6 +226,10 @@ in {
       enableFishIntegration = mkEnableOption "Fish integration" // {
         default = true;
       };
+
+      enableNushellIntegration = mkEnableOption "Nushell integration" // {
+        default = true;
+      };
     };
   };
 
@@ -247,6 +261,9 @@ in {
       programs.zsh.initExtra = mkIf cfg.enableZshIntegration gpgInitStr;
       programs.fish.interactiveShellInit =
         mkIf cfg.enableFishIntegration gpgFishInitStr;
+
+      programs.nushell.extraEnv =
+        mkIf cfg.enableNushellIntegration gpgNushellInitStr;
     }
 
     (mkIf (cfg.sshKeys != null) {


modules/services/gpg-agent.nix
```diff
@@ -237,23 +257,25 @@ in {
...
           ++ [ cfg.extraConfig ]);
 
-      home.sessionVariablesExtra = optionalString cfg.enableSshSupport ''
-        if [[ -z "$SSH_AUTH_SOCK" ]]; then
-          export SSH_AUTH_SOCK="$(${gpgPkg}/bin/gpgconf --list-dirs agent-ssh-socket)"
-        fi
-      '';
+      home.sessionVariables = optionalAttrs cfg.enableSshSupport {
+        SSH_AUTH_SOCK = "${gpgPkg}/bin/gpgconf --list-dirs agent-ssh-socket";
+      };
```
TornaxO7, line 262: this didn't work for me with `nushell`. May I ask if you have a better idea how to fix this? @lheckemann

modules/services/gpg-agent.nix
```diff
@@ -22,6 +22,16 @@ let
...
   '' + optionalString cfg.enableSshSupport gpgSshSupportStr;
 
+  gpgNushellInitStr = ''
+    $env.GPG_TTY = (tty)
+  '' + optionalString cfg.enableSshSupport ''
+    ${gpgPkg}/bin/gpg-connect-agent updatestartuptty /bye | ignore
+
+    if not "SSH_AUTH_SOCK" in $env {
+      $env.SSH_AUTH_SOCK = (${gpgPkg}/bin/gpgconf --list-dirs agent-ssh-socket)
+    }
```
TornaxO7, line 32: if we can find a way to move this to `home.sessionVariables` then I can remove this block but I can confirm (on my machine) that this works here.

modules/services/gpg-agent.nix
lheckemann: I don't think this will work even with bash, since this will result in the variable being set to the command string rather than the command's output (which is what the `$()` in the previous code was for, like nu's `()`).

modules/services/gpg-agent.nix
```diff
@@ -237,23 +257,25 @@ in {
...
+      home.sessionVariables = optionalAttrs cfg.enableSshSupport {
+        SSH_AUTH_SOCK = "${gpgPkg}/bin/gpgconf --list-dirs agent-ssh-socket";
+      };
 
       programs.bash.initExtra = mkIf cfg.enableBashIntegration gpgInitStr;
       programs.zsh.initExtra = mkIf cfg.enableZshIntegration gpgInitStr;
       programs.fish.interactiveShellInit =
         mkIf cfg.enableFishIntegration gpgFishInitStr;
+
+      programs.nushell.extraEnv = mkIf cfg.enableZshIntegration gpgNushellInitStr;
```
lheckemann, line 269: This condition should probably be `enableNushellIntegration`, not zsh :)

modules/services/gpg-agent.nix
lheckemann:
```suggestion
      home.sessionVariables = optionalAttrs cfg.enableSshSupport {
        SSH_AUTH_SOCK = "$XDG_RUNTIME_DIR/gnupg/S.gpg-agent.ssh";
      };
```
this is what I meant in my comment in the old PR, and that should avoid the need for any nushell-specific code. I'm just not sure if there are circumstances in which this value will be incorrect.

modules/services/gpg-agent.nix
TornaxO7: oh whoops, good point!

flake.lock
```diff
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1700204040,
-        "narHash": "sha256-xSVcS5HBYnD3LTer7Y2K8ZQCDCXMa3QUD1MzRjHzuhI=",
+        "lastModified": 1700612854,
+        "narHash": "sha256-yrQ8osMD+vDLGFX7pcwsY/Qr5PUd6OmDMYJZzZi0+zc=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "c757e9bd77b16ca2e03c89bf8bc9ecb28e0c06ad",
```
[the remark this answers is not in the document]
TornaxO7: I merged the master branch so it should be fine now

modules/services/gpg-agent.nix
TornaxO7: hm... I don't think that this will work because in nu, if you want to access an environment-veriable in nushell, you'll have to write `$env.<env-variable>`.

TornaxO7:
I'm sorry for my late response!

modules/services/gpg-agent.nix
lheckemann: You're right! I thought this used some shell-agnostic mechanism like pam, but it turns out I was wrong :(

So let's get rid of this part of the change completely, sorry for leading you down the wrong path there!

flake.lock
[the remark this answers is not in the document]
lheckemann: Nope, there's still a diff here.

lheckemann:
I was hardly quicker!

There's still a diff on the flake.lock which probably shouldn't be there.

There's also a fair bit of noise in the diff from auto-formatting, which I'm not sure is desired (but I'm not a frequent h-m contributor and don't know about formatting policies here).

modules/services/gpg-agent.nix
TornaxO7: > So let's get rid of this part of the change completely

hm... so you mean that I should remove the `home.sessionVariables` entry completely?

> sorry for leading you down the wrong path there!

no need to apologize :)

TornaxO7:
> There's still a diff on the flake.lock which probably shouldn't be there.

aye, I think that's because I executed `nix flake update` when I tested it out.

modules/services/gpg-agent.nix
lheckemann: Yeah basically just bring back what it had originally.

lheckemann (approved):
Looks good, just not sure about the auto-formatting (but I'll leave the verdict on that to someone who can actually merge this ^^)

flake.lock
```diff
@@ -24,4 +24,4 @@
   },
   "root": "root",
   "version": 7
-}
+}
```
rycee, line 27: Please remove this change.

rycee:
Thanks for the contribution. Please run the format script in the project root (or alteratively, `nix fmt` if you ave Flakes enabled), also squash all commits into a single commit with a commit message as described in the checklist.

flake.lock
TornaxO7: done

rycee:
Merged to master now 🙂

TornaxO7:
peepoAwesome

Issue #2255, opened 2021-08-07
bug: Prezto completion module is much slower than "vanilla" prezto


AxelTLarsson:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Hi.

I have spent the last few days trying to understand why my zsh startup time become so slow when I converted my from my usual prezto config to using home manager. I went from a startup time measured as such:

```sh
# $SHELL = /nix/store/apkv2fv89zd0knfi1l8pfyxl36xw3kzr-zsh-5.8/bin/zsh
for i in $(seq 1 10); do /usr/bin/time $SHELL -i -c exit; done
```

of around 0.10 seconds to 0.41 seconds using the same modules and config! 🐌 

To make sure I was making a fair comparison, I tried to make a minimal reproducible example by comparing the exact same revision of prezto with the default options - one configured with home manager and the other per prezto's installation instructions.

With default modules I get 0.21 s startup with home manager.

I copied the prezto derivation from my local nix store to my `~/.zprezto` (`cp -R /nix/store/shwbywl4c5fj4ivsm628ari4ivayyfwz-zsh-prezto-unstable-2021-06-02/share/zsh-prezto .zprezto
`) and ran prezto's setup in `~/.zprezto`.

```sh
setopt EXTENDED_GLOB
for rcfile in "${ZDOTDIR:-$HOME}"/.zprezto/runcoms/^README.md(.N); do
  ln -s "$rcfile" "${ZDOTDIR:-$HOME}/.${rcfile:t}"
done
```

I still get 0.21 s as expected (it should be the same configuration, more or less), howerver I noticed that runcoms/zshrc is not sourcing `~/.zprezto/init.zsh` but the `init.zsh` from the nix store, so I changed that:

```sh
# Source Prezto.
if [[ -s "${ZDOTDIR:-$HOME}/.zprezto/init.zsh" ]]; then
  source "${ZDOTDIR:-$HOME}/.zprezto/init.zsh"
fi
```

and re-ran the benchmark and then got excellent performance again at 0.10s, what is going on? 🤔 

Of course I realise that the init.zsh in the nix store and the init.zsh in `~/.zprezto` will actually load the modules from different locations, from the nix-store's prezto `/modules` and from `~/.zprezto/modules` respectively. But the contents are identical!

Does anybody know what is going on here and has anyone experienced anything similar?

[...]

PR #4701, opened 2023-11-26
Enable caseSensitive by default in prezto module

michaelpj:
See https://github.com/nix-community/home-manager/issues/2255.

This causes a really noticeable slowdown that is quite hard to track down!

### Description

Implements the fix from https://github.com/nix-community/home-manager/issues/2255

### Checklist

- [ ] Change is backwards compatible. *Unsure: this changes the option default?*

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@NickHu

### DIFF
--- a/modules/programs/zsh/prezto.nix
+++ b/modules/programs/zsh/prezto.nix
@@ -16,7 +16,8 @@ let
 
       caseSensitive = mkOption {
         type = types.nullOr types.bool;
-        default = null;
+        # See <https://github.com/nix-community/home-manager/issues/2255>.
+        default = true;
         example = true;
         description =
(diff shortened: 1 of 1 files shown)


AxelTLarsson (approved):
Very helpful! 👍

michaelpj:
Yes, this fixes the performance problem. My understanding is limited but based on the linked ticket I don't see an obvious way to fix it without turning case insensitivity off. You *can* of course still turn it back on, but the default is currently a big landmine for unsuspecting users.

michaelpj:
I still think this is useful.

rycee:

PR #4704, opened 2023-11-26
zsh: set autosuggestion color

teto:
### Description

syntax for ref https://askubuntu.com/questions/1393204/changing-zsh-autosuggestions-color
haven't played much with it

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/zsh.nix
```diff
@@ -220,6 +220,21 @@ let
...
+  autosuggestionModule = types.submodule {
+
+    options = {
+        enable = mkOption {
+          default = false;
+          description = "Enable zsh autosuggestions";
+        };
+        highlight = mkOption {
+          default = "fg=cyan";
+          description = "Enable zsh autosuggestions";
```
rycee, line 232: I imagine that the description here needs to be updated to match the option name.

modules/programs/zsh.nix
```diff
@@ -368,6 +379,13 @@ in
         description = "Options related to zsh-history-substring-search.";
       };
 
+      autosuggestion = mkOption {
```
rycee, line 382: I think you can skip the submodule. in this case it should be possible to simply do something like

```
autosuggestion = {
  enable = mkOption …;
  highlight = mkOption …;
};

rycee:
Thanks! Added some minor comments 🙂k

modules/programs/zsh.nix
teto: ha that's how I had first implemented it than I saw the highlighting module and decided to copy whatthey did. I will revert that.

rycee: Yeah, we should probably also simplify the highlighting options in the same way. The submodule looks unnecessary.

teto:
I think it's good but as I went back and forth between the different implementations, I would prefer for someone else to make sure it works and merge.

rycee:
@teto I made some minor changes, like adding `type` fields to the options and making the highlight optional (using `nullOr`). Please have a look and if you think it is OK then I think we can merge it.

rycee:
Merged now, hopefully that's OK 🙂

mfroeh:
Maybe I'm overlooking something, but for me `programs.zsh.autosuggestion.highlight` has no effect.
It sets `AUTOSUGGESTION_HIGHLIGHT_COLOR` instead of `ZSH_AUTOSUGGEST_HIGHLIGHT_STYLE` as required by 
[zsh-autosuggestions](https://github.com/zsh-users/zsh-autosuggestions?tab=readme-ov-file#configuration).

PR #4713, opened 2023-11-29
bemenu: add module

omernaveedxyz:
### Description

Add bemenu module https://github.com/Cloudef/bemenu with ability to configure options through `BEMENU_OPTS` environment variable.

Configuration of bemenu globally is done through setting this environment variable as opposed to using a config file as mentioned [here](https://github.com/Cloudef/bemenu/issues/15) and [here](https://github.com/Cloudef/bemenu/issues/158).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/tests/modules/programs/bemenu/basic-configuration.nix
+++ b/tests/modules/programs/bemenu/basic-configuration.nix
@@ -1,0 +1,26 @@
+{
+  programs.bemenu = {
+    enable = true;
+    settings = {
+      line-height = 28;
+      prompt = "open";
+      ignorecase = true;
+      fb = "#1e1e2e";
+      ff = "#cdd6f4";
+      nb = "#1e1e2e";
+      nf = "#cdd6f4";
+      tb = "#1e1e2e";
+      hb = "#1e1e2e";
+      tf = "#f38ba8";
+      hf = "#f9e2af";
+      af = "#cdd6f4";
+      ab = "#1e1e2e";
+    };
+  };
+
+  nmt.script = ''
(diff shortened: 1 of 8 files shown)


tests/modules/programs/bemenu/basic-configuration.nix
```diff
@@ -0,0 +1,25 @@
...
+      hf = "#f9e2af";
+      af = "#cdd6f4";
+      ab = "#1e1e2e";
+    };
+  };
+
+  nmt.script = ''
+    assertFileExists home-path/etc/profile.d/hm-session-vars.sh
+    assertFileContent home-path/etc/profile.d/hm-session-vars.sh \
+      ${./basic-configuration.sh}
```
ambroisie, line 23: This will lead to needless churn if the surrounding content of `hm-session-vars.sh` changes. You should probably use `assertFileContains` instead.

ambroisie (changes requested):

tests/modules/programs/bemenu/basic-configuration.nix
omernaveedxyz: I've updated the tests to use `assertFileContains` and `assertFileNotRegex` instead. Let me know if there is a better way to handle this.

ambroisie (approved):

omernaveedxyz:
I've updated the `settings` options to allow boolean values for settings such as `ignorecase`. I've also updated the test cases and example to reflect this.

ambroisie (approved):

rycee:

PR #4714, opened 2023-11-29
broot: Add nushell integration

takac:
### Description
Add additional config flag `enableNushellIntegration` in broot.nix to setup broot/br for nushell.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@aheaume @dermetfan


modules/programs/broot.nix
```diff
@@ -175,6 +175,14 @@ in {
       '';
     };
 
+    enableNushellIntegration = mkOption {
+      default = true;
+      type = types.bool;
+      description = ''
+        Whether to enable Fish integration.
```
dermetfan, line 182:
```suggestion
        Whether to enable Nushell integration.
```

modules/programs/broot.nix
```diff
@@ -226,5 +234,8 @@ in {
     programs.zsh.initExtra = mkIf cfg.enableZshIntegration (shellInit "zsh");
 
     programs.fish.shellInit = mkIf cfg.enableFishIntegration (shellInit "fish");
+
+    programs.nushell.extraConfig =
+      mkIf cfg.enableFishIntegration (shellInit "nushell");
```
dermetfan, line 239:
```suggestion
    programs.nushell.extraConfig = mkIf cfg.enableNushellIntegration (shellInit "nushell");
```

dermetfan (changes requested):

modules/programs/broot.nix
takac: oops, bit too hasty, apologies :(

dermetfan: No problem! Will hopefully review properly after work today.

dermetfan:
With Nushell 0.87.0 (which is currently in 23.11) I get this error message on startup:

```
Error:   × Deprecated command
    ╭─[/nix/store/gdwnbsz8wxrgg6iy2prraj3d6yill27p-br.nushell:10:1]
 10 │ # See https://dystroy.org/broot/install-br/
 11 │ def-env br [
    · ───┬───
    ·    ╰── `def-env` is deprecated and will be removed in 0.88.
 12 │     --cmd(-c): string               # Semicolon separated commands to execute
    ╰────
  help: Use `def --env` instead
```

The script still works so this is not a hard error but certainly ugly.

However, I believe it is still the right decision to merge for the following reasons:

1. This is an issue with broot. They will likely fix this and we won't have to change any code, so the problem will disappear by itself.
2. `programs.zoxide.enableNushellIntegration` has the same issue and is also enabled by default.
3. While the default is to enable nushell integration, users are still free to disable it.

I will merge in the coming days if no one raises concerns.

dermetfan:
Thanks for your contribution!

Issue #4568, opened 2023-10-15
bug: Directories for firefox bookmarks are not created


LuisIllig:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Using the default example for bookmarks:
```
programs.firefox.profiles.default.bookmarks = [
  {
    name = "wikipedia";
    tags = [ "wiki" ];
    keyword = "wiki";
    url = "https://en.wikipedia.org/wiki/Special:Search?search=%s&go=Go";
  }
  {
    name = "kernel.org";
    url = "https://www.kernel.org";
  }
  {
    name = "Nix sites";
    toolbar = true;
    bookmarks = [
      {
        name = "homepage";
        url = "https://nixos.org/";
      }
      {
        name = "wiki";
        tags = [ "wiki" "nix" ];
        url = "https://nixos.wiki/";
      }
    ];
  }
];
```
doesn't create the directory "Nix sites"

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.1.57, NixOS, 23.11 (Tapir), 23.11.20231011.5e4c2ad`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.17.0`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`
```

D3vil0p3r:
This issue seems to be still present in 23.11

shard77:
Yep.. also tried, same bug :/

iambeingtracked:
Still happens on 24.05

PR #4731, opened 2023-12-03
firefox: fix folders not showing in toolbar (#4568)

0x5a4:
### Description

Previously the generated `bookmarks.html` was malformed which lead to the inability to add folders to the toolbar(#4568)

I've also updated the documentation to reflect that the `toolbar` option doesnt add the folder to the toolbar, but that folder is the toolbar itself.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@rycee @kira-bruneau

### DIFF
--- a/modules/programs/firefox.nix
+++ b/modules/programs/firefox.nix
@@ -97,13 +97,14 @@ let
       directoryToHTML = indentLevel: directory: ''
         ${indent indentLevel}<DT>${
           if directory.toolbar then
-            ''<H3 PERSONAL_TOOLBAR_FOLDER="true">Bookmarks Toolbar''
+            ''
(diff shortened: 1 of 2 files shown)


modules/programs/firefox.nix
```diff
@@ -377,7 +378,8 @@ in {
                     toolbar = mkOption {
                       type = types.bool;
                       default = false;
-                      description = "If directory should be shown in toolbar.";
+                      description =
+                        "Make this the toolbar directory. This does NOT mean that this directory will be added to the toolbar, this directory IS the toolbar.";
```
rycee, line 382:
```suggestion
                      description = ''
                        Make this the toolbar directory. This does _not_ mean
                        that this directory will be added to the toolbar, this
                        directory _is_ the toolbar.
                      '';
```

PR #4738, opened 2023-12-05
cava: nullable package

khaneliman:
### Description
Currently, cava (in nixpkgs) isn't supported in darwin so I need to be able to mark the package as nullable to just configure my config through nix and retrieve the package from brew.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@bddvlpr

### DIFF
--- a/modules/programs/cava.nix
+++ b/modules/programs/cava.nix
@@ -14,7 +14,7 @@ in {
   options.programs.cava = {
     enable = mkEnableOption "Cava audio visualizer";
 
-    package = mkPackageOption pkgs "cava" { };
+    package = mkPackageOption pkgs "cava" { nullable = true; };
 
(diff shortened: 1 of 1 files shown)


rycee:
Generally we avoid nullable package options. Please give

``` nix
programs.cava.package = pkgs.emptyDirectory;
```

a try instead and see if that works OK.

khaneliman:
> Generally we avoid nullable package options. Please give
> 
> ```nix
> programs.cava.package = pkgs.emptyDirectory;
> ```
> 
> a try instead and see if that works OK.

Alright, I'll try that on my macbook quick.

khaneliman:
Looks like that method works, thanks!

PR #4745, opened 2023-12-07
zsh: add zproof option

MayNiklas:
### Description

Adds `programs.zsh.enableZproof`.
zproof is used to benchmark the shell initialisation -> helps to reduce loading time.
Needs to be the first & last line in `~/.zshrc`.
Since there isn't a `initExtraLast` option currently, it's hard to implement that without this option.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

I haven't found a `meta.maintainers` for zsh. Can somebody help me there?

### DIFF
--- a/modules/programs/zsh.nix
+++ b/modules/programs/zsh.nix
@@ -356,6 +356,13 @@ in
         description = "Enable zsh autosuggestions";
       };
 
+      zproof.enable = mkOption {
+        default = false;
+        description = ''
+          Enable zproof in your zshrc.
+        '';
+      };
+
       syntaxHighlighting = mkOption {
         type = syntaxHighlightingModule;
         default = {};
@@ -534,6 +541,12 @@ in
         ++ optional cfg.oh-my-zsh.enable cfg.oh-my-zsh.package;
 
       home.file."${relToDotDir ".zshrc"}".text = concatStringsSep "\n" ([
+        # zproof must be loaded before everything else, since it
+        # benchmarks the shell initialization.
+        (optionalString cfg.zproof.enable ''
+          zmodload zsh/zprof
+        '')
+
         cfg.initExtraFirst
         "typeset -U path cdpath fpath manpath"
 
@@ -656,6 +669,11 @@ in
             (downKey: "bindkey \"${downKey}\" history-substring-search-down")
             (lib.toList cfg.historySubstringSearch.searchDownKey)
           }
+        '')
+
+        (optionalString cfg.zproof.enable
+        ''
+          zprof
         '')
       ]);
     }


teto:
> Since there isn't a initExtraLast option currently, it's hard to implement that without this option.

IIRC, I was able to do that with `lib.mkAfter` commands.

MayNiklas:
> > Since there isn't a initExtraLast option currently, it's hard to implement that without this option.
> 
> 
> 
> IIRC, I was able to do that with `lib.mkAfter` commands. 

I didn't knew about this!
Do you think my PS still makes sense? Or should I close it?

modules/programs/zsh.nix
```diff
@@ -356,6 +356,13 @@ in
         description = "Enable zsh autosuggestions";
       };
 
+      enableZproof = mkOption {
```
rycee, line 359: I would suggest

```suggestion
      zproof.enable = mkOption {
```

modules/programs/zsh.nix
```diff
@@ -534,6 +541,12 @@ in
         ++ optional cfg.oh-my-zsh.enable cfg.oh-my-zsh.package;
 
       home.file."${relToDotDir ".zshrc"}".text = concatStringsSep "\n" ([
+        # zproof must be loaded before everything else, since it
+        # benchmarks the shell initialization.
+        (optionalString (cfg.enableZproof) ''
```
rycee, line 546:
```suggestion
        (optionalString cfg.enableZproof ''
```

modules/programs/zsh.nix
```diff
@@ -657,6 +670,11 @@ in
             (lib.toList cfg.historySubstringSearch.searchDownKey)
           }
         '')
+
+        (optionalString (cfg.enableZproof)
```
rycee, line 674:
```suggestion
        (optionalString cfg.enableZproof
```

rycee:
I don't have any opinion on whether this is useful enough to add, just added a few minor comments 🙂

MayNiklas:
> I don't have any opinion on whether this is useful enough to add, just added a few minor comments 🙂

Thank you!
I implemented those.

In my personal opinion, having zproof easily accessible is helpful.
Especially with home manager, it's easy to have a bloated shell config, that takes long to initialise.
Having an easy way to debug initialisation time might be helpful in some scenarios.

rycee:
Sounds good. I reworded the commit message and rebased to master in 77c698fa4b3081b6019ad77d1bfedf06dbbde0db. Thanks for the contribution!

mrBliss:
Why did you call it zproof instead of zprof?

MayNiklas:
> Why did you call it zproof instead of zprof?

Human error 😬
@rycee What is easier? You fixing it or me opening another PR?

teto:
@rycee does too much for HM already, please open an MR if you can. I dont think we need an alias yet, just fix the spelling

PR #4748, opened 2023-12-08
hyprland: improve config reload

Lillecarl:
### Description

When using the previous approach I've always gotten errors that I can't reload config on the .lock file that exists in /tmp when you run a standard configured hyprland.
This commit improves this by using hyprctl to find instances to reload instead.
We can remove the HYPRLAND_INSTANCE_SIGNATURE bogus assignment once https://github.com/hyprwm/Hyprland/issues/4088 is resolved.

Note: I've set my home-manager url to my local repo and run a home-switch, verified that hyprland reloads the config and that there's no error message anymore.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@fufexan

### DIFF
--- a/modules/services/window-managers/hyprland.nix
+++ b/modules/services/window-managers/hyprland.nix
@@ -237,11 +237,11 @@ in {
         + lib.optionalString (cfg.extraConfig != "") cfg.extraConfig;
 
       onChange = lib.mkIf (cfg.package != null) ''
-        (  # execute in subshell so that `shopt` won't affect other scripts
-          shopt -s nullglob  # so that nothing is done if /tmp/hypr/ does not exist or is empty
-          for instance in /tmp/hypr/*; do
-            HYPRLAND_INSTANCE_SIGNATURE=''${instance##*/} ${cfg.finalPackage}/bin/hyprctl reload config-only \
-              || true  # ignore dead instance(s)
+        ( # Execute in subshell so we don't poision environment with vars
+          # This var must be set for hyprctl to function, but the value doesn't matter.
+          export HYPRLAND_INSTANCE_SIGNATURE="bogus"
+          for i in $(${cfg.finalPackage}/bin/hyprctl instances -j | jq ".[].instance" -r); do
+            HYPRLAND_INSTANCE_SIGNATURE=$i ${cfg.finalPackage}/bin/hyprctl reload config-only
           done
         )
       '';


fufexan (approved):
LGTM.

Lillecarl:
Once https://github.com/hyprwm/Hyprland/commit/288f1863f0b6886e0cdf85d16bb2460bb042acf2 is released we can remove the Bogus assignment for "His"

modules/services/window-managers/hyprland.nix
```diff
@@ -237,11 +237,11 @@ in {
...
       onChange = lib.mkIf (cfg.package != null) ''
-        (  # execute in subshell so that `shopt` won't affect other scripts
-          shopt -s nullglob  # so that nothing is done if /tmp/hypr/ does not exist or is empty
-          for instance in /tmp/hypr/*; do
-            HYPRLAND_INSTANCE_SIGNATURE=''${instance##*/} ${cfg.finalPackage}/bin/hyprctl reload config-only \
-              || true  # ignore dead instance(s)
+        ( # Execute in subshell so we don't poision environment with vars
+          # This var must be set for hyprctl to function, but the value doesn't matter.
+          export HYPRLAND_INSTANCE_SIGNATURE="bogus"
+          for i in $(${cfg.finalPackage}/bin/hyprctl instances -j | ${pkgs.jq}/bin/jq ".[].instance" -r); do
```
rycee, line 243: Actually, jq is already a dependency and included in PATH by default. So this could be simplified to

```suggestion
          for i in $(${cfg.finalPackage}/bin/hyprctl instances -j | jq ".[].instance" -r); do
```

See https://github.com/nix-community/home-manager/blob/efa36e896951bec8d96e38ea40a22c010bd1bd8f/modules/home-environment.nix#L684-L698

Lillecarl: Nice! But even if it is a dependency already, isnt explicity preferred? That's my Nix mental model, there's no downside other than a bit of verbosity :)

rycee: Well it is explicitly included into the path at the top of the activation script:

``` sh
export PATH="…:/nix/store/wgqafnfqzqa6p51bb32wdmsplgz62nnj-jq-1.6-bin/bin:…"
```

And PATH only contains those paths that are specified there (at least for state version ≥ 22.11). So I wouldn't worry about it and instead go for the slight increase in readability with the bare jq command.

Lillecarl: Applied the change, reworded the commit message 😄

rycee:
Thanks! Merged now 🙂

Issue #4729, opened 2023-12-03
Hyprland: Put `wayland.windowManager.hyprland.settings.source` at the start of `hypr.conf`


rdvdev2:
### Description
The alphabetical order used right now by `wayland.windowManager.hyprland.settings` puts sourced files at the end of the configuration. This is pretty annoying if you are trying to import variables from another file (in my case, I'm trying to import the catppuccin theme color definitions). Putting `source` entries at the start of the file would solve this.

#### Alternatives
Maybe someone relies on the current order, and therefore a `wayland.windowManager.hyprland.sourcedFiles` or similar option could be added instead to avoid breaking their configuration.

PR #4755, opened 2023-12-09
hyprland: add option sourceFirst

rdvdev2:
### Description

Add the option sourceFirst to the hyprland module. When this option is enabled source entries will be put near the top of the file, so that the variables declared in other files can be used by the other configuration entries.

Add "source" to the list of important prefixes when the former option is enabled.

Resolves: #4729 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@fufexan

### DIFF
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -1317,6 +1317,23 @@ in
           A new module is available: 'services.signaturepdf'.
         '';
       }
+
+      {
+        time = "2023-12-10T08:43:02+00:00";
+        condition = config.wayland.windowManager.hyprland.settings ? source;
+        message = ''
+          Entries in
+
+            wayland.windowManager.hyprland.settings.source
+
+          are now placed at the start of the configuration file. If you relied
+          on the previous placement of the 'source' entries, please set
+
+             wayland.windowManager.hyprland.sourceFirst = false
+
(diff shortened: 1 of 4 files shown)


fufexan (approved):
Looks like the best thing we can do currently, so LGTM.

rdvdev2:
### Doubts
- [ ] Is the name of the option okay? (I took it from programs.tmux.sensibleOnTop)
- [ ] Should the implementation be modified so that sourced files are placed before variable definitions? It could make sense to need to override some variables from the sourced file, but this could be solved by using another variable name, so I doubt the value of complicating the logic for this.
- [ ] Should this option be enabled by default? While this would break backwards compatibility, I suspect that the most common use case for sourcing files is importing themes, which needs this option enabled. Some people may rely on the current behaviour (sourceOnTop = false) to configure the sourced files with variables, but this group is most probably a minority.

fufexan:
1. We can also name it `sourceFirst`, as `source`s come first in the config file.
2. Probably not, we're getting real complex real fast.
3. Yes, enabling it by default is a good thing imo.
4. Integrating it into the current test is doable, so let's do that.

fufexan (approved):

rycee:

PR #4770, opened 2023-12-12
gtk: fix GTK 4 theme being ignored

ThinkChaos:
### Description

Haven't dug into any code, but my best guess is GTK 4 ignores it on purpose since devs campaigned against theming: https://stopthemingmy.app

According to the following link, it's fine for users to do it as long as they don't expect support from app devs. And the campaign was against distributions applying custom themes by default.
https://github.com/GradienceTeam/Gradience/tree/0.4.1#%EF%B8%8F-gradience-stopthemingmyapp-and-adwaita-developers

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/misc/gtk.nix
+++ b/modules/misc/gtk.nix
@@ -35,6 +35,8 @@ let
           Package providing the theme. This package will be installed
           to your profile. If `null` then the theme
           is assumed to already be available in your profile.
+
+          For the theme to apply to GTK 4, this option is mandatory.
         '';
       };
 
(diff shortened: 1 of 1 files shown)


rycee:

jh-devv:
@ThinkChaos This brakes some stuff unfortunately:

When building GTK4 applications, you maybe wanna see how the default theme looks on the app, you could use `GTK_THEME` but this does not seem to work anymore cause of the `gtk.css` file always imports the theme.

Though, probably fine for now, this fixes bunch of things, so probably this is not so much of an problem..

ThinkChaos:
It's unfortunate the `gtk.css` file prevents `GTK_THEME` from working but I don't have any better ideas for setting a theme globally. The `GTK_THEME` [docs](https://docs.gtk.org/gtk4/running.html#gtk_theme) say "This is intended mainly for easy debugging of theme issues" so I didn't think of it as a viable solution for the HM module, but I also wasn't aware the CSS file overrides the var when I made the PR.  
As a workaround maybe you can do `XDG_CONFIG_HOME=/tmp your-app` so GTK doesn't find the CSS file.

jh-devv:
> It's unfortunate the `gtk.css` file prevents `GTK_THEME` from working but I don't have any better ideas for setting a theme globally. The `GTK_THEME` [docs](https://docs.gtk.org/gtk4/running.html#gtk_theme) say "This is intended mainly for easy debugging of theme issues" so I didn't think of it as a viable solution for the HM module, but I also wasn't aware the CSS file overrides the var when I made the PR.  
> As a workaround maybe you can do `XDG_CONFIG_HOME=/tmp your-app` so GTK doesn't find the CSS file.

Thanks!

PR #4787, opened 2023-12-18
Better direnv nushell integration

JoaquinTrinanes:
### Description

Nushell allows defining a `ENV_CONVERSIONS` environment variable used to transform the variables passed between nu and the system (by default it will transform `PATH` and `Path` into lists). However, those transformations aren't applied when `load-env`ing them, as nu assumes the string passed to it is intended to be a string.

This PR makes the direnv hook read `ENV_CONVERSIONS` and apply the transformations accordingly. Other than being less surprising/annoying it also prevents some bugs like new binaries in the path not being autocompleted.

The nushell hook is written in such a way that it won't break if any expected value is missing.

While at it, I took the liberty of removing all the `/bin` paths in the file in favor of `lib.getExe` (but that can be removed if not desired 😁 ).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/programs/direnv.nix
+++ b/modules/programs/direnv.nix
@@ -134,21 +134,33 @@ in {
       mkAfter ''
         $env.config = ($env.config? | default {})
         $env.config.hooks = ($env.config.hooks? | default {})
-        $env.config.hooks.pre_prompt = ($env.config.hooks.pre_prompt? | default [] | append {||
-            let direnv = (${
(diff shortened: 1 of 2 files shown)


rycee:
Thanks, don't use nushell myself but LGTM! Merged to master now 🙂

PR #4798, opened 2023-12-21
fish: escape abbr expansions once again

### LINKED ISSUE #4710: bug: New fish abbr features break existing config

### Issue description

Previously the `abbr` values were escaped, now they are not. This means something like `shellAbbrs = { foo = "echo \"rm ./*\""}` will now `rm` your current directory every time the `config.fish` file is loaded. 🥲 Once I updated every new shell was a total mess of random commands. I'm not sure it's a good idea to let home-manager handle my default shell's behavior anymore.

### Maintainer CC

_No response_

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.1.52, NixOS, 23.11 (Tapir), 23.11.20230915.ace5093`
 - multi-user?: `yes`
 - sandbox: `yes`
 - channels(root): `"home-manager, nixos"`
 - channels(stel): `""`
```

stelcodes:
Commit 8cedd6 `fish: support flexible abbreviations` removed shell escaping for fish shell abbr values. This was a dangerous breaking change offered little value and made writing abbr expansions more difficult. This commit restores automatic shell escaping of fish abbr values.

### Description
Fixes #4710 

In my opinion these abbr should be escaped like they always have been. Currently this config:
```nix
shellAbbrs = {
  foo = "printf \"great job\"";
  bar = "mkdir $(date %H:%M:%S%Z)";
};
```
will yield:
```
  abbr --add -- foo "printf "great job
  abbr --add -- bar "mkdir $(date %H:%M:%S%Z)"
```
This `foo` abbr is messed up with the quotes and `bar` abbr's `date` command will run once when the shell is started and hardcode the time into the abbr. This is just confusing behavior that is completely solved by going back to escaping the abbrs.

### Checklist

NOTE: This change reverts a dubious backwards incompatible change made previously, so it's technically not backwards compatible but it will maintain backwards compatibility for users using the `23.11` branch.

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -207,7 +207,8 @@ let
       modifiers = if isAttrs def then mods else "";
       expansion = if isAttrs def then def.expansion else def;
     in "abbr --add ${modifiers} -- ${name}"
-    + optionalString (expansion != null) " \"${expansion}\"") cfg.shellAbbrs);
+    + optionalString (expansion != null) " ${escapeShellArg expansion}")
+    cfg.shellAbbrs);
 
   aliasesStr = concatStringsSep "\n"
     (mapAttrsToList (k: v: "alias ${k} ${escapeShellArg v}") cfg.shellAliases);
--- a/tests/modules/programs/fish/abbrs.nix
+++ b/tests/modules/programs/fish/abbrs.nix
@@ -26,8 +26,13 @@
         };
         "4DIRS" = {
           setCursor = "!";
-          expansion =
-            "$(string join \\n -- 'for dir in */' 'cd $dir' '!' 'cd ..' 'end')";
+          expansion = ''
+            for dir in */
+              cd $dir
+              !
+              cd ..
+            end
+          '';
         };
         dotdot = {
           regex = "^\\.\\.+$";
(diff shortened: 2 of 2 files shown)


stelcodes:
Not sure why the MacOS tests aren't passing..

rycee:
I think it makes sense to keep a safe default so I've merged this.

I wonder, though, if `lib.literalExpression` could be adopted in these cases to allow explicit override of the escaping. There could be a new type `literalStr` which would allow only `lib.literalExpression` values (merging disallowed). Then we could have `either str literalStr`, that is, if you have an option `foo.bar` of that type you could one of

``` nix
{
foo.bar = "this string should be escaped";

# or

foo.bar = lib.literalExpression "this string should be not be escaped";
}
```

@infinisil Greetings, could you have a look at the above paragraph and see if you can recall seeing something like that before?

infinisil:
Haven't seen this before, but I think this would make sense. I wouldn't use `lib.literalExpression` for this though, that exists for the purpose of module docs rendering, not for module values. Something like `lib.mkShellRaw` and `lib.mkShellEscaping` sounds better. Then you could have a type like `types.shellString { defaultMode = "raw"|"escaping"; }`. Maybe even a `defaultMode = "none"` to signal that strings not wrapped in a function aren't allowed.

However, alternatively, and perhaps more idiomatic, would be use e.g.

```nix
options.myString = mkOption {
  type = coercedTo str (v: {
    # Escaping strings are default
    escaping = v;
  }) (submodule ({ config, ... }: {
    options.escaping = mkOption {
      type = types.str;
    }
    options.raw = mkOption {
      type = types.str;
      default = escapeShellArg config.escaping;
    }
  }));
}

# Strings get escaped by default
config.myString = "hello, \"there\"!";

# Also works explicitly
config.myString.escaping = "hello, \"there\"!";

# Raw strings aren't escaped
config.myString.raw = "Now is $(date)"
```

rycee:
Thanks @infinisil! I like your alternative solution. I would like to make the escape function a parameter, though, since this could be used in non-shell contexts. Oh, and I suppose `escaping` should be nullable if it should be possible to only specify the raw value. So perhaps

``` nix
types.escapingStr = { escape }:
  coercedTo str (v: {
    # Escaping strings are default
    escaping = v;
  }) (submodule ({ config, ... }: {
    options.escaping = mkOption {
    }
    options.raw = mkOption {
      type = types.str;
    }
    config.raw = mkIf (config.escaping != null) (
      mkDefault (escape config.escaping)
    );
  }));
```

Which could be used as

``` nix
option.foo = mkOption {
  type = types.escapingStr { escape = escapeDconfStr; }
  …
}
config.foo.raw = "@as ['blah']";
```

Could always also add parameters for whether to escape by default, etc.

PR #4799, opened 2023-12-21
direnv: Fix nushell integration ENV_CONVERSIONS

JoaquinTrinanes:
### Description

I had a base case of works on my machine™️ when testing #4787. I excitedly tried it today after cleaning up my config, and... it didn't work. At all. Turns out I had the direnv hook defined twice in my config and things kept (apparently) working fine.

I can confirm this works on a clean config now 🥲

Btw, I use nushell and direnv daily, so I'd be happy to keep maintaining this module in the foreseeable future :)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/direnv.nix
+++ b/modules/programs/direnv.nix
@@ -149,7 +149,7 @@ in {
                     {
                         key: $key
                         value: (do (
-                            $env.env_conversions?
+                            $env.ENV_CONVERSIONS?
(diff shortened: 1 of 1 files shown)


rycee:
Thanks! Merged to master now. And much appreciated that you can look after the module :+1:

PR #4801, opened 2023-12-22
podman: init module

### LINKED ISSUE #4336: Add support for rootless Podman
### Description
There is an Nix module called `virtualisation.oci-containers` allowing creating OCI (Podman, Docker, …) containers on NixOS. Podman does support rootless containers, but `home-manager` has no module for declarative configuring containers like the `virtualisation.oci-containers` allows.

Consider adding such feature as not all software has its nix derivation.

n-hass:
### Description
Fixes #4336. This module will add support on Linux systems to create podman containers and networks. This has been tested on NixOS, Ubuntu 24.04 and Arch (on ~28/08/24), and _should_ work across any home-manager compatible distro.

`podman.containers` aims to use an interface similar to NixOS `oci-containers`, but this module does not write the systemd units. Podman already includes a robust systemd unit generator - their Quadlet configuration interface - which I have opted to use. 

`podman.networks` is implemented using the Quadlet unit generator as well. The module which builds the derivations is common to all quadlets, so volumes and other resources should be easy to implement following networks as a guide.

Using the Quadlets generator vastly simplifies the process of creating the systemd units to run the podman containers because we simply _don’t have to._ The quadlet API, ie the options used here, are stable.

##### Usage example

<details>

```
{ … }:
let
    homeDir = “/home/myuser”;
in {
    services.podman.networks.caddy_routing = {
      driver = "bridge";
      subnet = "172.21.1.0/24";
    };

    services.podman.containers.caddy = {
      image = "ghcr.io/n-hass/caddy-cloudflare:latest";
      description = "Caddy web server";
      environmentFile = [ "${homeDir}/caddy/.env” ];
      network = [ "caddy_routing" ];
      networkAlias = [ “caddy" ];
      ports = [
        “8080:80"
        “8443:443"
      ];
      volumes = [
        "${homeDir}/caddy/Caddyfile:/etc/caddy/Caddyfile:ro"
        "${homeDir}/caddy/config:/etc/caddy/config"
        "${homeDir}/caddy/data:/data"
      ];
      autoUpdate = "registry”;
      extraConfig.Service = {
        TimeoutStopSec = 60;
      };
      addCapabilities = [ "NET_RAW" ];
    };
}
```

</details>

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

(After moving from Draft -> Ready for review, I will not squash until ready to merge to keep track of reviewed changes)


bamhm182:
Sorry it took so long, but I finally had some time to sit down and review this. Great work, but as I was looking through it, the reasoning behind having as few options as is reasonable started to make sense. The reasoning for this is that if some changes were made to the upstream project, it would require an update to home manager in order to fix them. As such, I feel that it makes more sense to try and implement this in the same way that NixOS implements the following systemd options:

* [unitConfig](https://search.nixos.org/options?channel=23.11&show=systemd.services.%3Cname%3E.unitConfig&from=0&size=50&sort=relevance&type=packages&query=systemd.services.%3Cname%3E.unitConfig)
* [serviceConfig](https://search.nixos.org/options?channel=23.11&show=systemd.services.%3Cname%3E.serviceConfig&from=0&size=50&sort=relevance&type=packages&query=systemd.services.%3Cname%3E.serviceConfig)

Doing it this way would mean that if, for example, `PublishPort` became `Port` in a future release of podman, users would be able to make this change on their own without waiting for us to fix it in HM. Additionally, we wouldn't have to worry about all the headache that would come with determining if it should be PublishPort or Port on the user's system. Quadlet does seem pretty stable at this point, so I don't expect that literal change to happen, but something more likely may be the addition of new arguments which we would then have to add to HM.

Looking forward to your feedback on this.

n-hass:
@rycee Do you have any thoughts or reservations on how I've implemented this module so far? I'm planning to just add tests now and fix whatever this `nmt-report-gpg-immutable-keyfiles.drv` issue is.

terlar:
I believe there are some overlap here, both PRs in the final stage:
https://github.com/nix-community/home-manager/pull/4331

bamhm182:
> I believe there are some overlap here, both PRs in the final stage: #4331

That one seemed stagnant when this one was created, but it looks like it has seen some activity since then.

https://github.com/nix-community/home-manager/issues/4336#issuecomment-1867180896

n-hass:
@terlar @rycee I have reached a point ready for review and added tests. Would appreciate your input

n-hass:
> it seems setting `networkMode`duplicates itself in the X-Container section and service exec line, causing `networkMode = "host"` to fail

Thanks @atagen nice catch, that’s left over from me playing around... Deciding what to put in the Network field is a little annoying since there are some rules and different formats (see the [podman man page](https://docs.podman.io/en/latest/markdown/podman-run.1.html#network-mode-net)). 

It’s intended that networks are defined with `podman.services.networks`, and for a container to use it it should be listed on `podman.services.containers.<name>.networks`. Other network options are put under `podman.services.containers.<name>.networkMode`, such as “host” as you have used. `networkMode` is probably a bad name considering that Quadlet just maps this to the `—network` flag on podman run, so it can technically be used for more than just setting the network mode.

n-hass:
I reproduced the `newuidmap not found` issue on a NixOS system, but only on boot for networks/containers set to autostart.  Anyway, that’s fixed.

But much worse, I ran into [this podman issue](https://github.com/containers/podman/discussions/21739) because podman needs setuid on `newuidmap` which means [it cannot be provided by shadow from nixpkgs](https://github.com/NixOS/nixpkgs/issues/138423#issuecomment-947888673). This is actually quite a complex issue which doesn’t have a good solution yet.

In the interest of trying to support more Linux distributions than just NixOS, I’ve added common places that `newuidmap` with the correct setuid can be found into PATH when podman needs it as well as the correct NixOS `/run/wrappers/bin/newuidmap`, and added an explicit wait for the user namespace to be ready before trying to create a network. In my testing this appears to have fixed the non-deterministic use of newuidmap.

bamhm182:
What still needs to be done to get this across the finish line?

terlar:
It is a quite huge one. I didn't have time to test this yet and reviewing it will take some time. But if you have the ability to test or review this as well. That would be great.

bamhm182:
Sounds good. I'm testing and reviewing it now. Will post back shortly.

bamhm182-hacs:
I'll wait for someone with more authority to comment, but I personally don't feel it is necessary. I like how you have everything laid out and the logic looks good. Very similar to the format of the standard systemd service NixOS config. Only concerns are whether or not the items which are explicitly declared are required, and if not, whether they should be I'm there, or just in extraContainerConfig/etc.

Pressed for time today, will make it a priority to finish looking over it tomorrow morning.

I eventually want to add the other quadlet unit files in there, but we should handle that after this gets in.

bamhm182:
So after digging into this quite a bit more, I think there are some changes that should be made to bring it closer to the other modules we already have in home-manager. If we consider that the quadlets are very similar in structure to systemd unit files, I feel it makes the most sense to model if after the structure of the already existing systemd unit files. The systemd unit files are configured a little differently than they are in NixOS proper, in that there are very few top level attributes, and instead, there are direct references to the sections within the unit file.

```
config.services.podman.containers."mysleep" = {
  Unit = {
    Description = "MySleep Container";
  };
  Container = {
    Image = "registry.access.redhat.com/ubi9-minimal:latest";
    ReadOnlyTmpfs = true;
    Network = [ "mysleep" "mynet" ];
  };
};
```

There should still be defaults for some of the options like you have, such as ContainerName defaulting to "mysleep" in the above example, but they should be able to be overridden. (`..."mysleep".Container.ContainerName = "somethingelse";`)

Happy to write the code to make the suggestions, but I'm not a reviewer and haven't had to recommend changes to a PR before. Do I just make a PR on your fork, n-hass?

bamhm182:
As I was reading back through all of the contribution documentation, I realized I had a misunderstanding when it comes to the options. As is described [here[(https://nix-community.github.io/home-manager/index.xhtml#sec-guidelines-valuable-options), I am now of the opinion that you had it right the first time with "extraContainerConfig", etc. as opposed to my "Container" suggestion. 

There's still a couple of small issues I can see, such as "PodmanArgs" always being present in the Container config, even if there are no arguments, and it seeming like `..."mysleep".extraContainerConfig.ContainerName = "testing";` should only result in 1 `ContainerName` line, which would be overriden with "testing".

Not sure how hard I'm bike shedding this, though, because currently, the outcome would be that ContainerName exists twice, but the second one ("ContainerName=testing") would win, resulting in a container named "testing", and it doesn't currently actually cause anything to break, it's just needlessly defined twice.

bamhm182:
Been working on cleaning some things up [here](https://github.com/bamhm182/home-manager/commit/95149811c5c6451f87a63a3b1d144aa688b47ec3).

- I removed any top level configs with the exception of those that either had some assumed defaults (like `ContainerName`) or would likely be lists or sets (like `PublishedPorts`, `Volumes`, etc.)
- Alphabetized everything
- Moved a lot of the `formatX` functions into more generic `formatListSpaces` (`Key=Val1 Val2`), `formatListNewlines` (`Key=Val1\nKey=Val2`), etc. functions so that we didn't have a separate function for formatting each setting.
- Configured it so that `extraXConfig` attributes would override their top level attributes if both are defined, or use it if only `extraXConfig`'s version is defined. For example, `..."mysleep".extraUnitConfig.Description = "testing"` would override `..."mysleep".description = "Standard description"`.
- Alphabetized everything for easier discovery
- Added `extraXConfig` options for each section in the quadlet unit files, so if things get added/removed/people want to modify them/etc, they should be able to be configured without us touching anything.
- Linted everything and tried to keep everything as standard as possible between files.
- Fixed issue with PodmanArgs always being present in the config, even when empty
- Probably a number of other things I'm forgetting.

Been working on this for far too long today and need to sleep, but I think it's just about to home-manager's expectations. Will be working on tests, give everything a few more reviews, and then submit a PR to your repo, @n-hass.

n-hass:
Thanks for your effort testing and reviewing @bamhm182 - I’ve had a quick look over those changes and have some thoughts, can you please open a PR on my fork so we can discuss there and not clutter this thread

bamhm182:
@n-hass Of course! I didn't submit it yet because I wanted to give it another pass, take a better look at the tests, and test that everything is working as I expect it too. I've submitted one now so we can talk there.

n-hass:
A rewrite and cleanup of the module has just been pushed. There are no plans to do any further work on this unless changes are requested.

bamhm182:
As @n-hass said, we'd been working together a lot on it these past couple weeks. We both currently believe it to be in a great spot and ready for merge. I went ahead and added myself as a maintainer due to the gained familiarity with this module and interest in seeing it get added to HM and continue to work.

Due to the aforementioned size of this module, I figured it wouldn't be a bad idea to add some context to the file names to decrease the effort required to review and understand what's going on.
- activation.nix: A script that will be added to home-manager-generation/activate. Responsible for removing resources which have been removed from the user's config.
- containers.nix: Responsible for generating the [Quadlet INI File](https://www.redhat.com/en/blog/quadlet-podman) for Containers. This is basically a .ini file which follows a similar pattern to a systemd unit file, but with additional sections to fill in relevant information for the container. Podman developers have a script to convert these quadlet files into appropriate systemd unit files. We are not writing the systemd unit files directly because it will provide better forward compatibility if we use the tools provided by the Podman team to do so. We have several high level attributes for commonly used options, and have an extraConfig escape hatch to allow users to add to (or override in case of a merge conflict). This file also creates a `.config/podman/containers.manifest` file to ease cleanup of removed containers.
- install-quadlet.nix: Responsible for actually building the Quadlets and corresponding service files.
- networks.nix: Same as containers, but for Podman Network resources.
- options.nix: Options related to the overall functionality of podman within HM.
- podman-lib.nix: Helper functions used within several of the other files.
- services.nix: Additional systemd units to perform common actions for podman (i.e., scheduled image updates for containers configured to do so)

Happy to answer any additional questions to get this merged. Thanks!

rycee:
Thanks for the contribution! Impressive work and I managed to get a small nginx container to run just fine. I've rebased on the latest master branch and added a commit with some proposed changes. Please have a look at the changes in that commit and see what you think.

n-hass:
Thanks for the feedback @rycee, suggestions LGTM. I think its fine to drop the ‘either single value or list’ types for options, it was really only there to have a way to control precedence in merging option values with extraConfig - ie two lists will be merged, a single value and a list won’t. But that’s pretty implicit behaviour, so maybe it's best that it’s scrapped. I’ve added a couple lines to a test on this.

bamhm182:
The suggestions look good to me as well. Thanks rycee!

modules/services/podman-linux/default.nix
```diff
@@ -1,14 +1,13 @@
...
 
-with lib;
-
 {
+  meta.maintainers = with lib.hm.maintainers; [ bamhm182 n-hass ];
+
   imports =
     [ ./containers.nix ./install-quadlet.nix ./networks.nix ./services.nix ];
 
-  config = mkIf pkgs.stdenv.isLinux {
```
n-hass: @rycee removing this has caused eval to fail on Darwin as it hits the platform assertion. If not having this top-level `mkIf`, what is the preferred way to do the platform guard?

rycee: I think the best solution is to add an option `services.podman.enable` that defaults to `false`. This will guard the inner workings of the module and reduce the amount of code that needs to be evaluated.

It also allows a way to fully disable the module with a single option. I've added this option in a new commit, hopefully that resolves the CI error.

rycee:
Thanks for the great work! Merged to master now.

PR #4803, opened 2023-12-23
conky: Init module

kaleocheng:
### Description

Add new module to enable [conky](https://github.com/brndnmtthws/conky) as service on Linux 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/lib/maintainers.nix
+++ b/modules/lib/maintainers.nix
@@ -119,6 +119,12 @@
     github = "JustinLovinger";
     githubId = 7183441;
   };
+  kaleo = {
+    name = "Kaleo Cheng";
+    email = "kaleocheng@gmail.com";
+    github = "kaleocheng";
+    githubId = 7939352;
+  };
   owm111 = {
     email = "7798336+owm111@users.noreply.github.com";
     name = "Owen McGrath";
--- a/modules/services/conky.nix
+++ b/modules/services/conky.nix
@@ -1,0 +1,53 @@
+{ config, lib, pkgs, ... }:
+
+let
+
+  cfg = config.services.conky;
+
+in with lib; {
+  meta.maintainers = [ hm.maintainers.kaleo ];
+
+  options = {
+    services.conky = {
+      enable = mkEnableOption "Conky, a light-weight system monitor";
+
+      package = mkPackageOption pkgs "conky" { };
+
+      extraConfig = lib.mkOption {
+        type = types.lines;
+        default = "";
+        description = ''
+          Configuration used by the Conky daemon. Check
+          <https://github.com/brndnmtthws/conky/wiki/Configurations> for
+          options. If not set, the default configuration, as described by
+          {command}`conky --print-config`, will be used.
+        '';
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+    assertions =
+      [ (hm.assertions.assertPlatform "services.conky" pkgs platforms.linux) ];
+
+    home.packages = [ cfg.package ];
+
+    systemd.user.services.conky = {
+      Unit = {
+        Description = "Conky - Lightweight system monitor";
+        After = [ "graphical-session.target" ];
+      };
+
+      Service = {
+        Restart = "always";
+        RestartSec = "3";
+        ExecStart = toString ([ "${pkgs.conky}/bin/conky" ]
+          ++ optional (cfg.extraConfig != "")
+          "--config ${pkgs.writeText "conky.conf" cfg.extraConfig}");
+      };
+
+      Install.WantedBy = [ "graphical-session.target" ];
+    };
+  };
+}
+
--- a/tests/modules/services/conky/basic-configuration.nix
+++ b/tests/modules/services/conky/basic-configuration.nix
@@ -1,0 +1,30 @@
+{ config, pkgs, ... }:
+
+{
+  services.conky = {
+    enable = true;
+    extraConfig = ''
+      conky.text = [[
+        S Y S T E M    I N F O
+        $hr
+        Host:$alignr $nodename
+        Uptime:$alignr $uptime
+        RAM:$alignr $mem/$memmax
+      ]]
+    '';
+  };
+
+  test.stubs.conky = { };
+
+  nmt.script = ''
+    serviceFile="$TESTED/home-files/.config/systemd/user/conky.service"
+
+    assertFileExists $serviceFile
+    assertFileRegex "$serviceFile" \
+        'ExecStart=@conky@/bin/conky --config .*conky.conf'
+
+    configFile="$(grep -o '/nix.*conky.conf' "$serviceFile")"
+    assertFileContent "$configFile" \
+        ${./basic-configuration.conf}
+  '';
+}
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -1572,6 +1572,18 @@ in {
           your habits being tracked. See https://freetubeapp.io/ for more.
         '';
       }
+
+      {
+        time = "2024-04-30T21:57:23+00:00";
+        condition = hostPlatform.isLinux;
+        message = ''
+          A new module is available: 'services.conky'.
+
+          Conky is a system monitor for X. Conky can display just about
+          anything, either on your root desktop or in its own window. See
+          https://conky.cc/ for more.
+        '';
+      }
     ];
   };
 }
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -279,6 +279,7 @@ let
     ./services/clipman.nix
     ./services/clipmenu.nix
     ./services/comodoro.nix
+    ./services/conky.nix
     ./services/copyq.nix
     ./services/darkman.nix
     ./services/devilspie2.nix
--- a/tests/default.nix
+++ b/tests/default.nix
@@ -225,6 +225,7 @@ in import nmtSrc {
     ./modules/services/cliphist
     ./modules/services/clipman
     ./modules/services/comodoro
+    ./modules/services/conky
     ./modules/services/darkman
     ./modules/services/devilspie2
     ./modules/services/dropbox
--- a/tests/modules/services/conky/basic-configuration.conf
+++ b/tests/modules/services/conky/basic-configuration.conf
@@ -1,0 +1,7 @@
+conky.text = [[
+  S Y S T E M    I N F O
+  $hr
+  Host:$alignr $nodename
+  Uptime:$alignr $uptime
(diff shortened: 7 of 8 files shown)


modules/services/conky.nix
```diff
@@ -0,0 +1,68 @@
+{ config, lib, pkgs, ... }:
+
+let
+  cfg = config.services.conky;
+
+  configArgument = if cfg.configFile != null then
+    "--config ${cfg.configFile}"
+  else
+    "--config ${config.xdg.configFile."conky/conky.conf".source}";
```
ambroisie, line 9: This will just use the `/nix/store/...` path, I think you might want something like `${config.home.homeDirectory}/${config.xdg.configFile."conky/conky.conf".target}` (not sure if `XDG_CONFIG_DIR` is set in a systemd service, otherwise just use that...).

modules/services/conky.nix
```diff
@@ -0,0 +1,68 @@
...
+in with lib; {
+
+  meta.maintainers = [ maintainers.kaleo ];
+
+  options = {
+    services.conky = {
+      enable = mkEnableOption (lib.mdDoc "Light-weight system monitor");
```
ambroisie, line 16:
```suggestion
      enable = mkEnableOption "Light-weight system monitor";
```

modules/services/conky.nix
```diff
@@ -0,0 +1,68 @@
...
+
+      package = mkPackageOption pkgs "conky" { };
+
+      configFile = mkOption {
+        type = types.nullOr types.path;
+        default = null;
+        description = ''
+          Configuration file path to link to use instead of
```
ambroisie, line 24:
```suggestion
          Configuration file path to use instead of
```

ambroisie (changes requested):
I'm not sure I see the point of the `configFile` option, `readFile`/`mkOutOfStoreSymlink` might be all that's needed for this IMO.

modules/services/conky.nix
kaleocheng: using the `/nix/store/...` pattern  here as it seems a common way for systemd service in `modules/services` compare to the `config.home.homeDirectory` one. 
- https://github.com/search?q=repo%3Anix-community%2Fhome-manager++path%3Amodules%2Fservices+config.xdg.configFile&type=code
- https://github.com/search?q=repo%3Anix-community%2Fhome-manager++path%3Amodules%2Fservices+config.home.homeDirectory&type=code 

I'm not sure if it's in purpose or just  the legacy reason. let me know if we indeed what to switch to the `home.homeDirectory` one

kaleocheng:
> I'm not sure I see the point of the `configFile` option, `readFile`/`mkOutOfStoreSymlink` might be all that's needed for this IMO.

TBH I'm not using that option, added it just coz `configFile` is kind of a common option.  but you may be right. for the init version it's better to keep it simple. I'm going to remove it .

modules/services/conky.nix
kaleocheng: update: use `config.xdg.configHome` seems the good way. I will update it to use that 

-  https://github.com/search?q=repo%3Anix-community%2Fhome-manager++path%3Amodules%2Fservices+config.xdg.configHome&type=code

modules/services/conky.nix
```diff
@@ -0,0 +1,60 @@
+{ config, lib, pkgs, ... }:
+
+let
+  cfg = config.services.conky;
+
+  configFile = if cfg.settings != "" then
+    cfg.settings
+  else
+    builtins.readFile (pkgs.runCommand "conky.conf" { }
+      "${pkgs.conky}/bin/conky --print-config > $out");
```
kaleocheng, line 10: @ambroisie I'm not sure if it is a good idea to get the default config  from command output.  maybe  we just hardcode the default config in here?

ambroisie: Sorry, that's not what I meant when I said one can use `readFile`: you'd introduced an option to give a path to a config file, instead of using the `settings` attribute. Instead one could just do `settings = readFile`.

Definitely a :-1: on doing this, as it's an IFD.

modules/lib/maintainers.nix
```diff
@@ -89,6 +89,12 @@
     github = "JustinLovinger";
     githubId = 7183441;
   };
+  kaleo = {
+    name = "Kaleo Cheng";
+    email = "github@kaleo.me";
+    github = "kaleocheng";
+    githubId = 7939352;
+  };
```
ambroisie, line 97: Please add yourself to the maintainers list in a different commit.

modules/services/conky.nix
```diff
@@ -0,0 +1,60 @@
...
+  };
+
+  config = mkIf cfg.enable {
+
+    assertions =
+      [ (hm.assertions.assertPlatform "services.conky" pkgs platforms.linux) ];
+
+    home.packages = [ cfg.package ];
+
+    xdg.configFile."conky/conky.conf".text = configFile;
```
ambroisie, line 41: See first comment:

```suggestion
    xdg.configFile."conky/conky.conf".text = cfg.settings;
```

ambroisie (changes requested):

modules/services/conky.nix
kaleocheng: ok, I misunderstood the comments before.   

> Definitely a 👎 on doing this, as it's an IFD.

let me try to  summary  the config/settings part:
- keep  `settings` option only ( as `types.lines` )
- no default config, it means
   - if `settings` is not `""`, then we generate the xdg confile file and append `--config {filepath}` to `ExecStart`
   - else only run `conky`  in  `ExecStart`. in that case Conky will then use the builtin-configs to start. 

so we don't have to run the external command to generate or hardcode a config in here. does it make sense to you?

modules/services/conky.nix
```diff
@@ -0,0 +1,55 @@
...
+
+let cfg = config.services.conky;
+
+in with lib; {
+
+  meta.maintainers = [ maintainers.kaleo ];
+
+  options = {
+    services.conky = {
+      enable = mkEnableOption "Light-weight system monitor";
```
ambroisie, line 11: The description for `mkEnableOption` is prefixed by `Whether to enable ...`, so something like this would be more appropriate:

```suggestion
      enable = mkEnableOption "conky, a light-weight system monitor";
```

modules/services/conky.nix
```diff
@@ -0,0 +1,55 @@
...
+
+      package = mkPackageOption pkgs "conky" { };
+
+      settings = lib.mkOption {
+        type = types.lines;
+        default = "";
+        description = ''
+          Configuration written to {file}`$XDG_CONFIG_HOME/conky/conky.conf`.
+          Check <https://github.com/brndnmtthws/conky/wiki/Configurations"> for options.
+          If not set, {command}`conky --print-config` will be used by default.
```
ambroisie, line 21:
```suggestion
          If not set, the default configuration, as described by {command}`conky --print-config`, will be used.
```

ambroisie (changes requested):
Minor nitpicks on the phrasing of the documentation, otherwise LGTM.

kaleocheng:
@ambroisie appreciate for your feedback, I made some updates for the documentation. should I leave it to you next or we need to ping someone else?

kaleocheng:
hmm. I'm considering if  it should be   `program`  module instead of  `service` module. 
asking as I saw the tint2 in `program` https://github.com/nix-community/home-manager/blob/master/modules/programs/tint2.nix

kaleocheng:
@rycee not sure if you are the right person to review this. do you have any suggestion for the next of this PR.

modules/services/conky.nix
```diff
@@ -0,0 +1,55 @@
...
+
+  meta.maintainers = [ maintainers.kaleo ];
+
+  options = {
+    services.conky = {
+      enable = mkEnableOption "conky, a light-weight system monitor";
+
+      package = mkPackageOption pkgs "conky" { };
+
+      settings = lib.mkOption {
```
rycee, line 15: I think this is best changed to be `extraConfig` instead of `settings`. Yeah I know the name is a bit odd considering that there are no other options available but that is the convention.

The reason against `settings` is that it is typically referred to _structural settings_ such as a Nix attribute set serialized to JSON. Not raw text lines, is is the case here.

```suggestion
      extraConfig = lib.mkOption {
```

modules/services/conky.nix
```diff
@@ -0,0 +1,55 @@
...
+        Description = "Conky - Lightweight system monitor";
+        After = [ "graphical-session.target" ];
+      };
+
+      Service = {
+        Restart = "always";
+        RestartSec = "3";
+        ExecStart = toString ([ "${pkgs.conky}/bin/conky" ]
+          ++ optional (cfg.settings != "")
+          "--config ${config.xdg.configHome}/conky/conky.conf");
```
rycee, line 48: Is it necessary to place the configuration file in the user's home directory? Perhaps it would be possible to here do something like
```suggestion
          "--config ${pkgs.writeText "conky.conf" cfg.extraConfig}");
```
instead?

Then there is no need for the file in the home directory _and_ it makes it possible to auto-restart the service when the configuration changes (since the file path changes).

tests/modules/services/conky/basic-configuration.nix
```diff
@@ -0,0 +1,29 @@
...
+        conky.text = [[
+          S Y S T E M    I N F O
+          $hr
+          Host:$alignr $nodename
+          Uptime:$alignr $uptime
+          RAM:$alignr $mem/$memmax
+        ]]
+      '';
+    };
+
```
rycee, line 17: Can add

```suggestion

      test.stubs.conky = { };
      
```

Then the conky package is replaced by a stub and will not need to be downloaded.

modules/services/conky.nix
rycee: Using `config.xdg.configFile."conky/conky.conf".source` is good here since it will cause the file content to change if the configuration changes. This allows automatic systemd unit restarts.

rycee:
@kaleocheng Apologies for the long delay! I've added a few comments.

modules/services/conky.nix
kaleocheng: from tech perspective we don't have to put it in user's home directory. I did that as I saw lots of modules are doing that. e.g. https://github.com/nix-community/home-manager/blob/7e91f2a0ba4b62b88591279d54f741a13e36245b/modules/programs/rofi.nix#L219

but sounds those are kind of "legacy" way, I will follow your comment here to use the writeText instead.

rycee:

PR #4805, opened 2023-12-24
gpg-agent: don't set a default for pinentry

SuperSandro2000:
NixOS stopped building gtk2 pinentry by default in https://github.com/NixOS/nixpkgs/pull/270266 and it does not appear to me that there is a reasonable other default.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/gpg-agent.nix
+++ b/modules/services/gpg-agent.nix
@@ -198,7 +198,7 @@ in {
       pinentryFlavor = mkOption {
         type = types.nullOr (types.enum pkgs.pinentry.flavors);
         example = "gnome3";
-        default = "gtk2";
+        default = null;
         description = ''
           Which pinentry interface to use. If not
           `null`, it sets
@@ -210,8 +210,6 @@ in {
           ```nix
           services.dbus.packages = [ pkgs.gcr ];
           ```
-          For this reason, the default is `gtk2` for
-          now.
         '';
       };
 


lheckemann (approved):

cloudlena:
Why not use `"qt"` as the new default?
For me (using Hyprland), that works well. I think, it might be more user-friendly than not having a default of `null`.

SuperSandro2000:
> Couldn't using `"qt"` as the new default also be an option?

I have no idea. I am using kde, so for me this will obviously work but I have no clue about other systems.

rycee:

I wonder if perhaps it wouldn't be a good idea to simply remove the `nullOr` option and leave the option unset so that users are forced to set this explicitly? I can create a PR for that if nobody can think of anything obviously bad about this (beside causing some people's configurations to stop building without adding the explicit option).

bphenriques:
Is there any official guide on how to upgrade the existing home-manager setups? I use both NixOS (X11 Plasma) and MacOS and the original setup was compatible with both (as I do not do anything system wide in MacOS). 

In my case, I was able to sign my commits under both systems but now I cannot. From what I read the gtk2 is very much deprecated.

(This threw me off while I was upgrading some packages - the price to pay for following `nix-unstable`  :stuck_out_tongue_closed_eyes: )

Issue #4763, opened 2023-12-10
Option to enable program configiration without installing package


deej-io:
### Description

I am currently in the process of moving the configuration for a lot of programs to home-manager on Arch Linux but I'd like to continue to use the packages from the Arch repositories.

Currently, we have to `enable` the program to ensure it writes the configuration files, however this also automatically adds the package to the package list.

I was wondering if there would be any interest in a way of decoupling the configuration and installation through a top-level flag, e.g. `options.programs.installPackages`, that is defaults to `true` for backwards compatibility.

This can be used as a default value for e.g. `config.programs.git.installPackage`, which can then be used to conditionally add the package to the packages list like so:

```
config = mkIf cfg.enable (mkMerge [
  {
    mkIf cfg.installPackage {
      home.packages = [ cfg.package ];
    }
   ...
```

This gives the flexibility of a top-level default but also able to opt in/out for individual programs.

I'm happy to have a go at implementing this if there is interest.

rycee:
No installing the package together with the configuration is generally speaking not supported or recommended since it breaks reproducibility. The typical escape hatch if you really really want, is to override the package using `pkgs.emptyDirectory`. For example,

    programs.git.package = pkgs.emptyDirectory;

deej-io:
Thank you for the responses. That `emptyDirectory` approach seems simple enough and my desired behavior can be  trivially implemented with a helper function to conditionally set that option.

Closing this issue, but will update with the helper function code once I've gotten around to it.

Thanks again.

mainrs:
Sorry to necro this, but this resulted from trying the same thing as OP!

I wanted to configure `fish` using home-manager without installing the binary. However, it seems that the config generation for autocompletions relies on the fact that the package is actually _not_ empty:

https://github.com/nix-community/home-manager/blob/e83414058edd339148dc142a8437edb9450574c8/modules/programs/fish.nix#L424-L446

There is no way to make this work right now, am I right? `fish` cannot be used with the above-mentioned trick of using `pkgs.emptyDirectory`.

Lindenk:
This really should never have been closed. This behaviour is still a serious issue on any non-nixos distro (actually it's a pretty big issue on nixos as well if a system wide graphical program like hyprland is overridden by a different, incorrect version with home-manager). The `emptyDirectory` trick is a workaround that doesn't always work, such as with modules that expect the original package to exist or when dependencies and stuffed into the user's path (like python).

PR #4808, opened 2023-12-24
programs: (WIP) add option to not install packages for enabled programs

djrollins:
### Description

This is a work-in-progress implementation for issue #4763. I have raised this now to get some feedback on whether this is something that would be accepted by the project and whether the approach is sound, before I go through the effort of updating all the program submodules to use the new option.

There is a new global `options.programs.installPackages` option that sets the default behaviour for all programs, which itself is defaulted to true to maintain backwards compatibility.

The `program` submodules should expose their own `options.programs.<program>.install` option that can override this value.

I am not 100% sure the approach I have taken is the best. Ideally I'd like to expose a `mkInstallOption` function from somewhere sensible, however I couldn't work out how to expose a function from `config.programs` or how to pass config into `lib.options` so that I could set the default value from `options.programs.installPackages`.

My second concern is regarding the use of the nix store paths in config for packages that are not installed (See [this comment](https://github.com/nix-community/home-manager/pull/4808/commits/6c23bd9fe8346dd1c92a2fe466bf340d60dc87f8#r1435862722)). My current approach is to hope that the package is on the users "$PATH", however in necessary cases we could add another option for the user to specify these paths if that is not the case.

Finally, whilst I'd be happy to go through all of the packages to introduce the new option, this could definitely bloat the PR and require eyes from many maintainers. Therefore I was going to ask if I should just introduce the new option for one of the programs in this PR and then raise separate ones for others in separate PRs?

I'd be grateful for any feedback on this PR before I go much further with this.

Cheers

Daniel

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/lib/default.nix
+++ b/modules/lib/default.nix
@@ -11,6 +11,7 @@ rec {
   maintainers = import ./maintainers.nix;
   strings = import ./strings.nix { inherit lib; };
   types = import ./types.nix { inherit gvariant lib; };
+  options = import ./options.nix { inherit lib; };
 
   shell = import ./shell.nix { inherit lib; };
   zsh = import ./zsh.nix { inherit lib; };
--- a/modules/lib/options.nix
+++ b/modules/lib/options.nix
@@ -1,0 +1,10 @@
+{ lib }:
+{
+  mkInstallOptionWithDefault =
+    default: name: lib.mkOption {
+      inherit default;
+      type = lib.types.bool;
+      example = true;
+      description = "Whether to install package ${name}";
+    };
+}
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -46,6 +46,7 @@ let
     ./misc/xdg-user-dirs.nix
     ./misc/xdg.nix
     ./misc/xfconf.nix
+    ./programs.nix
     ./programs/abook.nix
     ./programs/aerc.nix
     ./programs/afew.nix
--- a/modules/programs.nix
+++ b/modules/programs.nix
@@ -1,0 +1,11 @@
+{ config, lib, ... }:
+
+{
+  options.programs.installPackages = lib.mkOption {
+    type = lib.types.bool;
+    default = true;
+    description = "Whether to install packages for configured programs by default";
+  };
+
+  config = {};
+}
--- a/modules/programs/feh.nix
+++ b/modules/programs/feh.nix
@@ -5,6 +5,7 @@ with lib;
 let
 
   cfg = config.programs.feh;
+  mkInstallOption = hm.options.mkInstallOptionWithDefault config.programs.installPackages;
 
   bindingsOf = t: with types; attrsOf (nullOr (either t (listOf t)));
 
@@ -26,6 +27,8 @@ let
 in {
   options.programs.feh = {
     enable = mkEnableOption "feh - a fast and light image viewer";
+
+    install = mkInstallOption "feh";
 
     package = mkPackageOption pkgs "feh" { };
 
@@ -71,7 +74,7 @@ in {
         "To disable a keybinding, use `null` instead of an empty string.";
     }];
 
-    home.packages = [ cfg.package ];
+    home.packages = optional cfg.install cfg.package;
 
     xdg.configFile."feh/buttons" =
       mkIf (cfg.buttons != { }) { text = renderBindings cfg.buttons + "\n"; };


modules/programs/gh.nix
```diff
@@ -131,7 +134,7 @@ in {
     programs.git.extraConfig.credential = mkIf cfg.gitCredentialHelper.enable
       (builtins.listToAttrs (map (host:
         lib.nameValuePair host {
-          helper = "${cfg.package}/bin/gh auth git-credential";
+          helper = (if cfg.install then "${cfg.package}/bin/gh" else "gh") ++ "auth git-credential"
```
djrollins, line 137: I'm not sure what the best approach for a fallback would be here? Another option for the user to provide a path?

ambroisie:
Isn't the easiest way to not install a package to simply provide something akin to `package = pkgs.emptyDir` to the relevant module's `package` option?

For modules that need an executable, one can for example use `writeShellScriptBin "gh" ''/bin/gh "$@"''`  to call the relevant program from the base system.

djrollins:
> Isn't the easiest way to not install a package to simply provide something akin to `package = pkgs.emptyDir` to the relevant module's `package` option?
> 
> For modules that need an executable, one can for example use `writeShellScriptBin "gh" ''/bin/gh "$@"''` to call the relevant program from the base system.

Thank you for the response!

Regarding your first point, I guess overriding the `package` option is one way of doing it, however I was thinking a specific and more obvious API, with global and local settings, would be a nice QoL improvement. 


I did try your suggestion, but it seems that `pkgs.emptyDir` doesn't seem to exist and the option requires an attribute of type `package`, for which I can't find an empty value.

I suppose we could do something like the following to allow the user to set the package to `null` to prevent installation.

```nix
options = {
  package = mkOption {
        type = types.nullOr types.package;
        default = pkgs.feh;
        defaultText = literalExpression "pkgs.feh";
        description = "set to `null` to not install packages.";
  };
};
```

With that we could perhaps achieve the desired behavior with something like:

```nix
let
  isNix = ...;
  isMacOS = ...;
  installPackages = isNix || isMacOS
in
{
  programs.git = {
    enable = true;
    package = null;
  };
}
```
But I think this would then error

djrollins:
> Isn't the easiest way to not install a package to simply provide something akin to `package = pkgs.emptyDir` to the relevant module's `package` option?
> 
> For modules that need an executable, one can for example use `writeShellScriptBin "gh" ''/bin/gh "$@"''` to call the relevant program from the base system.

Thank you for the response!

Regarding your first point, I guess overriding the package option is one way of doing it, and is fine for a one-off, however I was thinking a specific and more obvious API, with global and local settings, would be a nice QoL improvement to allow easy conditional installation e.g.

```
let
  isNix = ...;
  isMacOS = ...;
in
{
  programs.installPackages = isNix || isMacOS;

  -- this will only be installed on Nix and MacOS machines.
  programs.git = {
    enable = true; 
  };

  -- only install on Nix machines.
  programs.fzf = {
    enable = true;
    install = isNix;
  }
}
```

I did try your suggestion, but it seems that `pkgs.emptyDir` does not exist. The option requires an attribute of type package, for which I can't find an "empty" value.

I suppose if we had that, we could perhaps achieve the desired behavior with something like:

```
let
  isNix = ...;
  isMacOS = ...;
  installPackages = isNix || isMacOS
in
{
  programs.git = {
    enable = true;
    package = lib.mkIf (!installPackages) _empty_package_ -- Not sure what we can pass here to represent the an empty package?
  };

  programs.fzf = {
    enable = true;
    package = lib.mkIf (!isNix) _empty_package_ -- 
  }
}
```

But I think this is a bit more cumbersome than it needs to be.

ambroisie:
> it seems that `pkgs.emptyDir` does not exist

Sorry, it's `pkgs.emptyDirectory` :upside_down_face:.

> The option requires an attribute of type package, for which I can't find an "empty" value.

`mkPackageOption` [accepts a `nullable` argument](https://github.com/NixOS/nixpkgs/blob/5bc0439d70e4fb8da3fc84144216d922b7bc9f41/lib/options.nix#L160-L162) for such occasions.

djrollins:
Aha! That's really useful thanks. `nullable` packages seem to get us most of the way there, so I'll probably gut this PR entirely.

Just need to figure out how to handle uninstalled packages in scripts (with your `writeShellScriptBin` suggestion) and I'll open a PR with the necessary changes to the `git` module.

Thank you for all of the help.

PR #4809, opened 2023-12-24
targets/darwin: allow configuring application linking

moni-dz:
### Description

Not all packages supply a bundled macOS application, and I would like to configure where apps get placed since I use Raycast instead of Spotlight.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@midchildan

### DIFF
--- a/modules/targets/darwin/linkapps.nix
+++ b/modules/targets/darwin/linkapps.nix
@@ -1,9 +1,24 @@
 { config, lib, pkgs, ... }:
 
-{
-  config = lib.mkIf pkgs.stdenv.hostPlatform.isDarwin {
+let cfg = config.targets.darwin;
+in {
+  options.targets.darwin.linkApps = {
+    enable =
+      lib.mkEnableOption "linking macOS applications to the user environment"
+      // {
+        default = true;
+      };
+
+    directory = lib.mkOption {
+      type = lib.types.str;
+      default = "Applications/Home Manager Apps";
+      description = "Path to link apps relative to the home directory.";
+    };
+  };
+
+  config = lib.mkIf (pkgs.stdenv.hostPlatform.isDarwin && cfg.linkApps.enable) {
     # Install MacOS applications to the user environment.
-    home.file."Applications/Home Manager Apps".source = let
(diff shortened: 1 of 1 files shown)


modules/targets/darwin/linkapps.nix
```diff
@@ -1,9 +1,29 @@
...
+let cfg = config.targets.darwin;
+in {
+  options.targets.darwin.linkApps = {
+    enable = mkOption {
+      type = types.bool;
+      default = true;
+      example = false;
+      description =
+        "Whether to enable linking macOS applications to the user environment.";
+    };
```
sumnerevans, line 14:
```suggestion
    enable = mkEnableOption "linking macOS applications to the user environment" // {
      default = true;
    };
```

modules/targets/darwin/linkapps.nix
```diff
@@ -1,9 +1,29 @@
...
+
+    directory = mkOption {
+      type = types.str;
+      default = "Applications/Home Manager Apps";
+      defaultText = "Applications/Home Manager Apps";
```
sumnerevans, line 19: I don't think this is necessary anymore.

```suggestion
```

modules/targets/darwin/linkapps.nix
```diff
@@ -1,9 +1,29 @@
...
+    enable = mkOption {
+      type = types.bool;
+      default = true;
+      example = false;
+      description =
+        "Whether to enable linking macOS applications to the user environment.";
+    };
+
+    directory = mkOption {
+      type = types.str;
```
sumnerevans, line 17:
```suggestion
      type = types.path;
```

sumnerevans (changes requested):

modules/targets/darwin/linkapps.nix
khaneliman: This is a string that gets injected into the `home.file` below, not an actual full path.

khaneliman (approved):
Allows greater flexibility and verified it works locally, thanks

PR #4811, opened 2023-12-26
osmscout-server: add module

Thra11:
### Description

Osmscout-server includes a setting in its UI to create a systemd user service and socket to run the server on demand. This does not function correctly on NixOS, for two reasons:
1. It assumes that the binary path is stable (e.g. /usr/bin/osmscout-server), which is not the case on NixOS.
2. It auto-detects the unwrapped binary path, which doesn't work.

This module allows the user to access the same functionality on NixOS.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/osmscout-server.nix
+++ b/modules/services/osmscout-server.nix
@@ -1,0 +1,76 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let cfg = config.services.osmscout-server;
+in {
+  meta.maintainers = [ maintainers.Thra11 ];
+
+  options = {
+    services.osmscout-server = {
+      enable = mkEnableOption "OSM Scout Server";
+
+      package = mkPackageOption pkgs "osmscout-server" { };
+
+      network = {
+        startWhenNeeded = mkOption {
+          type = types.bool;
+          default = true;
+          description = ''
+            Enable systemd socket activation.
+          '';
+        };
+
+        listenAddress = mkOption {
+          type = types.str;
+          default = "127.0.0.1";
+          description = ''
+            The address for the server to listen on.
+          '';
+        };
+
+        port = mkOption {
+          type = types.port;
+          default = 8553;
+          description = ''
+            The TCP port on which the server will listen.
+          '';
+        };
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+    assertions = [
+      (lib.hm.assertions.assertPlatform "services.osmscout-server" pkgs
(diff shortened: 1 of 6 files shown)


modules/services/osmscout-server.nix
```diff
@@ -0,0 +1,89 @@
...
+      Unit = { Description = "OSM Scout Server Socket"; };
+
+      Socket = {
+        ListenStream =
+          "${cfg.network.listenAddress}:${toString cfg.network.port}";
+        TriggerLimitIntervalSec = "60s";
+        TriggerLimitBurst = 1;
+      };
+
+      Install = { WantedBy = [ "user-session.target" ]; };
```
Thra11, line 84: `user-session.target` is used in the systemd target generated by upstream osmscout-server, but I'm not sure if it's the right target in this context: It doesn't appear to exist by default.

rycee: No, there is no such target inside a user session. You'll want
```suggestion
      Install = { WantedBy = [ "sockets.target" ]; };
```
instead.

modules/services/osmscout-server.nix
```diff
@@ -0,0 +1,89 @@
...
+      };
+
+      package = mkOption {
+        type = types.package;
+        default = pkgs.osmscout-server;
+        defaultText = "pkgs.osmscout-server";
+        description = ''
+          The OSM Scout Server package to run.
+        '';
+      };
```
rycee, line 26: Should be sufficient with

```suggestion
      package = mkPackageOption pkgs "osmscout-server" { };
```

modules/services/osmscout-server.nix
```diff
@@ -0,0 +1,89 @@
...
+
+  options = {
+    services.osmscout-server = {
+      enable = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          Whether to enable OSM Scout Server as a systemd service.
+        '';
+      };
```
rycee, line 17: Should be sufficient with

```suggestion
      enable = mkEnableOption "OSM Scout Server";
```

rycee:
Thanks! Overall looks good. Just need to add a test case. Something similar to the [cachix-agent test case](https://github.com/nix-community/home-manager/blob/80679ea5074ab7190c4cce478c600057cfb5edae/tests/modules/services/cachix-agent/basic-setup.nix). You can copy that and assert the content of the generated service and socket unit files.

rycee:

Issue #3696, opened 2023-02-22
Alacritty moving towards configuration with TOML format


oceanlewis:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

This isn't a "bug" exactly, but it looks like the Alacritty project is looking at migrating away from using YAML as their configuration language to TOML, which will break `programs.alacritty.settings`.

I don't think `nixpkgs` has a `builtin.toTOML`, but it's easy enough for me to get around this by using `nushell`'s `from yaml | to toml` functionality. Probably there are better ways to fix this though.

See https://github.com/alacritty/alacritty/pull/6673 for where this work is happening.

### Maintainer CC

@timhae @rycee @berbiche @ncfavier

### System information

```markdown
N/A
```

ncfavier:
Thanks for the heads-up. It looks like upstream will continue supporting YAML for some time so we can have a graceful transition.

Should be as simple as changing `pkgs.formats.yaml` to `pkgs.formats.toml`, and maybe that will be a good time to revisit [this TODO](https://github.com/nix-community/home-manager/blob/4295fdfa6b0005c32f2e1f0b732faf5810c1bc7f/modules/programs/alacritty.nix#L53-L54).

jperras:
Tangentially related: Alacritty has already introduced a non-standard YAML `import` directive, which allows a YAML file to import another YAML file. One of the drivers of this was for colorscheme themes: https://github.com/alacritty/alacritty-theme#imports

kchibisov:
>Tangentially related: Alacritty has already introduced a non-standard YAML import directive, which allows a YAML file to import another YAML file. One of the drivers of this was for colorscheme themes: https://github.com/alacritty/alacritty-theme#imports

It's a simple array of strings which is called `import`, it could be called whatever and is a plain yaml array with strings. Those strings we use as a source of other configurations which we manually merge.

```yml
foo:
  - "./foo1.yml"
  - "./foo2.yml"
```

tobiasbaehr:
btw. when you enable alacritty via home manager, then it generates json for alacritty.yml.

 I do not believe this is valid. ;-)

rycee:
@tobiasbaehr YAML is a superset of JSON so any valid JSON is also valid YAML so we have historically used Nix's builtin toJSON function also for YAML output.

ncfavier:
I think the easiest would be something like

[...]

PR #4817, opened 2023-12-28
alacritty: make module compatible with alacritty 0.13

r-vdp:
### Description

The config file is in TOML from 0.13 onwards.
This PR makes the alacritty module compatible with 0.13 while retaining compatibility with previous versions.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/alacritty.nix
+++ b/modules/programs/alacritty.nix
@@ -4,7 +4,9 @@ with lib;
 
 let
   cfg = config.programs.alacritty;
-  yamlFormat = pkgs.formats.yaml { };
+  useToml = lib.versionAtLeast cfg.package.version "0.13";
+  tomlFormat = pkgs.formats.toml { };
+  configFileName = "alacritty.${if useToml then "toml" else "yml"}";
 in {
   options = {
     programs.alacritty = {
@@ -18,7 +20,7 @@ in {
       };
 
       settings = mkOption {
-        type = yamlFormat.type;
+        type = tomlFormat.type;
         default = { };
         example = literalExpression ''
           {
@@ -26,7 +28,7 @@ in {
               lines = 3;
               columns = 200;
             };
-            key_bindings = [
+            keyboard.bindings = [
               {
                 key = "K";
                 mods = "Control";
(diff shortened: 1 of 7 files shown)


modules/programs/alacritty.nix
```diff
@@ -37,27 +40,35 @@ in {
         '';
         description = ''
           Configuration written to
-          {file}`$XDG_CONFIG_HOME/alacritty/alacritty.yml`. See
-          <https://github.com/alacritty/alacritty/blob/master/alacritty.yml>
-          for the default configuration.
+          {file}`$XDG_CONFIG_HOME/alacritty/${configFileName}`. See
```
rycee, line 43: Unfortunately cannot do this since manual generation then depends on the user's configuration. Will need to be something like

``` markdown
{file}`$XDG_CONFIG_HOME/alacritty/alacritty.yml` or
{file}`$XDG_CONFIG_HOME/alacritty/alacritty.toml`,
where the latter is used for Alacritty 0.13 and later.
```

modules/programs/alacritty.nix
```diff
@@ -18,15 +21,15 @@ in {
       };
 
       settings = mkOption {
-        type = yamlFormat.type;
+        type = configFormat.type;
```
rycee, line 24: Hmm, unfortunately cannot really do this either due to the problem with manual depending on configuration. I wonder if it may be best to simply switch unconditionally to TOML, since that seems to be the future format. Then add an assertion that fails the build if the version is below 0.13.

Adding something like

``` nix
assertions = [
  {
    assertion = lib.versionAtLeast cfg.package.version "0.13";
    message = "The Alacritty module requires Alacritty version 0.13 or later.";
  }
];
```

to the `config` attribute might do the trick.

r-vdp: Yeah, that would simplify things a lot, but I didn't want to break backwards compat. Also then we have an issue with the PR for the 0.13 update in nixpkgs and this PR needing to be merged at the same time.

r-vdp: The types for toml and yaml are currently the same though, so we can just use the toml type and remove the yaml support in the near future once the PR in nixpkgs was merged.

r-vdp: Changed to use the toml formatter unconditionally for the type, it's the same as the yaml type anyway.

modules/programs/alacritty.nix
r-vdp: Good point, fixed.

modules/programs/alacritty.nix
rycee: That sounds like a good solution. Thanks!

rycee:

modules/programs/alacritty.nix
```diff
@@ -37,27 +39,37 @@ in {
...
+    xdg.configFile."alacritty/${configFileName}" =
+      lib.mkIf (cfg.settings != { }) (lib.mkMerge [
+        (lib.mkIf useToml {
+          source =
+            (tomlFormat.generate configFileName cfg.settings).overrideAttrs
+            (finalAttrs: prevAttrs: {
+              buildCommand = lib.concatStringsSep "\n" [
+                prevAttrs.buildCommand
+                # TODO: why is this needed? Is there a better way to retain escape sequences?
+                "substituteInPlace $out --replace '\\\\' '\\'"
```
ncfavier, line 64: See https://github.com/nix-community/home-manager/issues/3696 and earlier backlog for why this is needed

modules/programs/alacritty.nix
```diff
@@ -37,27 +39,37 @@ in {
...
-    })
-  ];
+    xdg.configFile."alacritty/${configFileName}" =
+      lib.mkIf (cfg.settings != { }) (lib.mkMerge [
+        (lib.mkIf useToml {
+          source =
+            (tomlFormat.generate configFileName cfg.settings).overrideAttrs
+            (finalAttrs: prevAttrs: {
+              buildCommand = lib.concatStringsSep "\n" [
+                prevAttrs.buildCommand
```
ncfavier, line 62: (This could have used `postBuild`)

tests/modules/programs/alacritty/example-settings-expected.toml
```diff
@@ -0,0 +1,8 @@
+[[keyboard.bindings]]
+chars = "\x0c"
```
ncfavier, line 2: TOML [doesn't support](https://toml.io/en/v1.0.0#string) this escape sequence, and instead wants us to use `\u000c`, which YAML doesn't support.

It's really a shame that the user should have to care about such format-specific bureaucracy, given the whole point of managing your configuration with Nix...

But anyway, this should at least be mentioned in the example for lack of a better solution.

ncfavier: ...or maybe it does, since it's a superset of JSON.

Anyway, the following seems to work independently of the format and doesn't require any backslash substitution:

```nix
keyboard.bindings = let
 c = (builtins.fromTOML ''c = "\u000c"'').c;
in [ {
  key = "K";
  mods = "Control";
  chars = c;
} ];
```

This could probably be turned into a library function `lib.hexByte` as @infinisil suggested [here](https://github.com/NixOS/nixpkgs/issues/97310#issuecomment-687902805).

modules/programs/alacritty.nix
r-vdp: This uses `runCommand` underneath which sets `buildCommand` instead of `buildPhase` and so AFAIU there's no pre and post build hooks.
If I add this to `postBuild`, it doesn't do anything.

ncfavier: You're right.

tests/modules/programs/alacritty/example-settings-expected.toml
r-vdp: Good point! https://github.com/nix-community/home-manager/pull/4853

r-vdp: Oh that might be a better approach, but a bit heavy without the library function. I don't like the substitution in the buildCommand at all though.

zzantares:
@rycee is there a way to back-port this to the `release-23.11` branch? not sure if that's standard practice.

PR #4825, opened 2023-12-29
tmux: True Color support

ozkutuk:
### Description

Adds True Color support based on [this gist](https://gist.github.com/andersevenrud/015e61af2fd264371032763d4ed965b6). There is a failing test in some unrelated module so I couldn't run _all_ tests, but I have tested each tmux test individually and fixed them accordingly.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/tmux.nix
+++ b/modules/programs/tmux.nix
@@ -94,6 +94,11 @@ let
       bind-key -N "Kill the current pane" x kill-pane
     ''}
 
+    ${optionalString cfg.trueColor ''
+      set -g default-terminal "tmux-256color"
+      set -ag terminal-overrides ",xterm-256color:RGB"
+    ''}
+
     set  -g mouse             ${boolToStr cfg.mouse}
     setw -g aggressive-resize ${boolToStr cfg.aggressiveResize}
     setw -g clock-mode-style  ${if cfg.clock24 then "24" else "12"}
@@ -292,6 +297,8 @@ in {
       tmuxp.enable = mkEnableOption "tmuxp";
 
       tmuxinator.enable = mkEnableOption "tmuxinator";
+
+      trueColor = mkEnableOption "true color support";
 
       plugins = mkOption {
         type = with types;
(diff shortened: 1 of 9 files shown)


ambroisie (commented):
I think you might need something similar to [this](https://github.com/ambroisie/nix-config/blob/932717b7548ae5f7a7fdec3e7e2d78d0fca6ef68/modules/home/tmux/default.nix#L92-L99) (as recommended [in the official wiki](https://github.com/tmux/tmux/wiki/FAQ#how-do-i-use-rgb-colour)).

At least, I remember having trouble with `terminfo` databases not containing the `RGB` flag even though the terminal supported it...

EDIT: I just saw the you included it for `xterm-256colors` specifically (` set -ag terminal-overrides ",xterm-256color:RGB"`). This means that e.g: `alacritty` won't work.

ozkutuk:
> I think you might need something similar to [this](https://github.com/ambroisie/nix-config/blob/932717b7548ae5f7a7fdec3e7e2d78d0fca6ef68/modules/home/tmux/default.nix#L92-L99) (as recommended [in the official wiki](https://github.com/tmux/tmux/wiki/FAQ#how-do-i-use-rgb-colour)).

I really liked your approach, definitely seems to be more generally-applicable than what I have here. Do you mind if I use the relevant section of your config as part of this PR?

ambroisie:
Knock yourself out :-)

modules/programs/tmux.nix
```diff
@@ -293,6 +298,8 @@ in {
 
       tmuxinator.enable = mkEnableOption "tmuxinator";
 
+      trueColor = mkEnableOption "true color support";
```
rycee, line 301:
```suggestion
      enableTrueColor = mkEnableOption "true color support";
```

rycee:
Are the additional changes still pending?

ozkutuk:
> Are the additional changes still pending?

Yes, I am still planning to integrate @ambroisie's approach to this but couldn't get around to doing it yet. I am converting the PR to a draft state for now.

PR #4835, opened 2023-12-31
lorri: unbreak due to too tight sandboxing

symphorien:
lorri needs to be able to write to /run/user/uid for the socket, to its own cache directory ~/.cache/lorri and to the directory for gc roots.

I wonder how this ever worked. Did systemd change?

Please backport this change to 23.11

### Checklist

tests pass except for a hash mismatch with `trying https://keys.openpgp.org/pks/lookup?op=get&options=mr&search=0x36cacf52d098cc0e78fb0cb13573356c25c424d4`

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

cc @Gerschtli

### DIFF
--- a/modules/services/lorri.nix
+++ b/modules/services/lorri.nix
@@ -52,6 +52,12 @@ in {
           PrivateTmp = true;
           ProtectSystem = "strict";
           ProtectHome = "read-only";
+          ReadWritePaths = [
+            # /run/user/1000 for the socket
+            "%t"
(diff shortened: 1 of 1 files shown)


rycee:
Thanks! I've merged this now.

I'm not sure what happened. I used to use Lorri, but it was quite a while ago. It may be that the `ProtectSystem` and `ProtectHome` didn't actually apply to user services at that time?

rycee:
And now backported to 23.11.

Issue #4531, opened 2023-10-04
bug: Incompatibility between vdirsyncer and khard modules


DamienCassou:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

I have enabled both khard and vdirsyncer programs. The idea is that vdirsyncer fetches&uploads the contacts that are listed&edited with khard.

It seems that it isn't possible to have both vdirsyncer and khard configured together with home-manager because:
1. vdirsyncer creates a subfolder within `accounts.contact.accounts.<name>.local.path` and puts the vcf files in the subfolder.
2. khard reads the vcf files from `accounts.contact.accounts.<name>.local.path` directly and doesn't look recursively inside the subfolders.
As a result, khard finds no vcf files and thus lists no contacts.

Said differently, here is part of my vdirsyncer configuration:

```ini
[storage contacts_nico_local]
fileext = ".vcf"
path = "~/configuration/contacts"
```

and here is part of my *working* hand-written khard configuration:

```ini
[addressbooks]
[[contacts]]
path = ~/configuration/contacts/172ce...
```

As you can see, the configuration file references the subfolder `172ce.../` instead of directly the path configured for vdirsyncer.

Here is an excerpt of what home-manager generates for khard:

```ini
[addressbooks]
[[nico]]
path = ~/configuration/contacts
```

This doesn't work because of the missing subfolder.

Here is the home-manager configuration that I cannot use:

```nix
{ pkgs, lib, ... }:

{
  programs.vdirsyncer = { enable = true; };

  programs.khard = {
    enable = true;
    settings = {
      ...
    };
  };

  accounts.contact.accounts = {
    nico = {
      vdirsyncer = {
        enable = true;
        ...
      };
      khard = { enable = true; };
      local = {
        path = "~/configuration/contacts";
        ...
      };
      remote = { ... };
    };
  };
}
```

### Maintainer CC

@olmokramer @antonmosich @teto @ayyess @loewenheim @jorsn 

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.5.5-200.fc38.x86_64, Fedora Linux, 38 (Workstation Edition), nobuild`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.17.0`
 - channels(root): `"nixpkgs"`
 - channels(temp): `"home-manager, nixpxkgs"`
[...]

PR #4836, opened 2023-12-31
khal: fix contact integration

felipeqq2:
### Description

Currently, birthday calendars added to Khal via `accounts.contact` throw `error: attribute 'type' missing`. This PR:

- Fixes a bug to correctly add `khal.type = "birthdays"` to these calendars;
- Moves options `color`/`priority` to global Khal configuration (these options can also be set for birthday calendars, and Nix currently throws an error if non-defined);
- Adds multiple calendars for each contact collection (fixing an issue similar to #4531):
  - New option `khal.collections`;
  - Also supports null-collections (as seen [here](https://github.com/nix-community/home-manager/issues/4531#issuecomment-1748332523));

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@teto (git blame)

### DIFF
--- a/modules/programs/khal-accounts.nix
+++ b/modules/programs/khal-accounts.nix
@@ -44,7 +44,7 @@ with lib;
       type = types.int;
       default = 10;
       description = ''
-        Priority of a calendar used for coloring.
+        Priority of a calendar used for coloring (calendar with highest priority is preferred).
       '';
     };
   };
--- a/modules/programs/khal-accounts.nix
+++ b/modules/programs/khal-accounts.nix
@@ -13,5 +13,39 @@ with lib;
         Keep khal from making any changes to this account.
       '';
     };
+
+    color = mkOption {
+      type = types.nullOr (types.enum [
+        "black"
+        "white"
+        "brown"
+        "yellow"
+        "dark gray"
+        "dark green"
+        "dark blue"
+        "light gray"
+        "light green"
+        "light blue"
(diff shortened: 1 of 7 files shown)


modules/programs/khal-accounts.nix
```diff
@@ -13,5 +13,39 @@ with lib;
...
+        Color in which events in this calendar are displayed.
+      '';
+      example = "light green";
+    };
+
+    priority = mkOption {
+      type = types.int;
+      default = 10;
+      description = ''
+        Priority of a calendar used for coloring.
```
teto, line 47: could you mention which is highest priority, eg.
`(priority 0 is shown before priority 10)`

teto:
could you rebase to trigger CI please ? I dont use accounts.contact but I am willing to try the PR this week.

modules/programs/khal-accounts.nix
felipeqq2: Fixed in f6c4426.

felipeqq2:
I don't think the failing tests were caused by my changes (same issue with #5005)

teto (approved):
can you confirm this is backwards compatible ?

felipeqq2:
> can you confirm this is backwards compatible ?

Yes (AFAICT). `priority` is now an existing option with a default value (`10`) for contact calendars, but that's what was being used by khal anyway.

PR #4842, opened 2024-01-02
vscode: add extrapackages option

xyangst:
### Description

added a extrapackages option, similar to helix/neovim
based on #4586, only thing changed is the binary path

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/vscode.nix
+++ b/modules/programs/vscode.nix
@@ -72,6 +72,13 @@ in {
         '';
       };
 
+      extraPackages = mkOption {
+        type = with types; listOf package;
+        default = [ ];
+        example = literalExpression "[ pkgs.nil ]";
+        description = "Extra packages available to hx.";
+      };
+
       enableUpdateCheck = mkOption {
         type = types.bool;
         default = true;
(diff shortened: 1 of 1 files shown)


modules/programs/vscode.nix
```diff
@@ -72,6 +72,13 @@ in {
         '';
       };
 
+      extraPackages = mkOption {
+        type = with types; listOf package;
+        default = [ ];
+        example = literalExpression "[ pkgs.nil ]";
+        description = "Extra packages available to hx.";
```
pedorich-n, line 79:
```suggestion
        description = "Extra packages available to vscode.";
```

modules/programs/vscode.nix
```diff
@@ -221,7 +228,22 @@ in {
...
+    home.packages = if cfg.extraPackages != [ ] then
+      [
+        (pkgs.symlinkJoin {
+          name =
+            "${lib.getName cfg.package}-wrapped-${lib.getVersion cfg.package}";
+          paths = [ cfg.package ];
+          preferLocalBuild = true;
+          nativeBuildInputs = [ pkgs.makeWrapper ];
+          postBuild = ''
+            wrapProgram $out/bin/code \
```
pedorich-n, line 240: I believe this won't work if `cfg.package` is `vscodium`, since its main executable is called [`codium`](https://github.com/NixOS/nixpkgs/blob/612f97239e2cc474c13c9dafa0df378058c5ad8d/pkgs/applications/editors/vscode/vscodium.nix#L65)


Maybe something like `$out/bin/${cfg.package.meta.mainProgram or (lib.getName cfg.package)}`? Like in [`lib.getExe`](https://github.com/hsjobeki/nixpkgs/blob/09fb8770db06195580ab47b22cf9722107df8ea1/lib/meta.nix#L252-L256)

losnappas:
Based on my testing with `symlinkJoin`, occasionally, when I run the vscode command `Developer: Reload Window` or use the direnv extension to reload the env, then the window (sometimes) loses the `extraPackages`. GPT-5 told me to change to `paths = [ pkg ] ++ extraPackages;`.

There's some magic going on though, so I'll report back if that change breaks again.

PR #4851, opened 2024-01-03
firefox: add default containers

rhoriguchi:
### Description

`userContextIdInternal.webextStorageLocal` is used by certain extensions without they don't work correctly <https://searchfox.org/mozilla-central/rev/762f24e00a9548d80ebba1b985c871ba6d9b829d/toolkit/components/contextualidentity/ContextualIdentityService.sys.mjs#93-96>. `userContextIdInternal.thumbnail` is some default added by firefox, so I just added it as well.

> This userContextId is used by ExtensionStorageIDB.jsm to create an IndexedDB database
> opened with the extension principal but not directly accessible to the extension code
> (do not change the userContextId assigned here, otherwise the installed extensions will
> not be able to access the data previously stored with the browser.storage.local API).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

- @DamienCassou 
- @rycee

### DIFF
--- a/modules/programs/firefox.nix
+++ b/modules/programs/firefox.nix
@@ -73,7 +73,24 @@ let
         version = 4;
         lastUserContextId =
           elemAt (mapAttrsToList (_: container: container.id) containers) 0;
-        identities = mapAttrsToList containerToIdentity containers;
+        identities = mapAttrsToList containerToIdentity containers ++ [
+          {
+            userContextId = 4294967294; # 2^32 - 2
+            name = "userContextIdInternal.thumbnail";
+            icon = "";
+            color = "";
+            accessKey = "";
+            public = false;
+          }
+          {
+            userContextId = 4294967295; # 2^32 - 1
+            name = "userContextIdInternal.webextStorageLocal";
+            icon = "";
+            color = "";
+            accessKey = "";
+            public = false;
+          }
+        ];
       }}
     '';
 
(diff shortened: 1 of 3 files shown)


modules/programs/firefox.nix
```diff
@@ -663,6 +680,19 @@ in {
...
+      (let
+        getContainers = profiles:
+          flatten (mapAttrsToList (_: value: value.containers) profiles);
+
+        findInvalidContainerIds = profiles:
+          filter (container: container ? "id" && container.id < 4294967294)
+          (getContainers profiles);
+      in {
+        assertion = cfg.profiles == { }
+          || length (findInvalidContainerIds cfg.profiles) != 0;
```
DamienCassou, line 692: I think `!= 0` should be changed to `== 0`.

modules/programs/firefox.nix
```diff
@@ -663,6 +680,19 @@ in {
           (", namely " + concatStringsSep ", " defaults);
       })
 
+      (let
+        getContainers = profiles:
+          flatten (mapAttrsToList (_: value: value.containers) profiles);
```
DamienCassou, line 685: given that `value.containers` is an attribute set, `mapAttrsToList` is going to return a list of attribute sets. There is nothing to flatten here. For example, the function will probably return something like:

```
[
  { "shopping" = ...; } // 1st profile
  { "shopping" = ...; } // 2nd profile
]
```

I don't think this is what the rest of the code expects.

DamienCassou (changes requested):
Thank you for this. I think the code is broken but the idea is worth implementing. Can you please double-check?

modules/programs/firefox.nix
rhoriguchi: You are right but also had to switch the filter function.

modules/programs/firefox.nix
rhoriguchi: Yeah take a look at the new `getContainers` function, I've also added `attrValues` which then requires `flatten`. Please take an other look.

DamienCassou (approved):
This is great work, thank you!

nyabinary:
Should be good to merge right?

rhoriguchi:
> Should be good to merge right?

Yes, is ready to be merged.

rycee:
Thanks! Merged to master now https://github.com/nix-community/home-manager/pull/4851

brckd:
`profile-setting-expected-containers.json` needs to be updated

PR #4852, opened 2024-01-04
thunderbird: configure signature if set

Lykos153:
### Description

The identity's signature  will be configured according to `signature.text` and `signature.showSignature`. Signature commands are not supported by Thunderbird.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@d-dervishi @jkarlson

### DIFF
--- a/modules/programs/thunderbird.nix
+++ b/modules/programs/thunderbird.nix
@@ -49,6 +49,11 @@ let
       "mail.identity.id_${id}.fullName" = account.realName;
       "mail.identity.id_${id}.useremail" = address;
       "mail.identity.id_${id}.valid" = true;
+      "mail.identity.id_${id}.htmlSigText" =
+        if account.signature.showSignature == "none" then
+          ""
+        else
+          account.signature.text;
     } // optionalAttrs (account.gpg != null) {
       "mail.identity.id_${id}.attachPgpKey" = false;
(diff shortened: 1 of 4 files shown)


modules/programs/thunderbird.nix
```diff
@@ -49,6 +49,11 @@ let
       "mail.identity.id_${id}.fullName" = account.realName;
       "mail.identity.id_${id}.useremail" = address;
       "mail.identity.id_${id}.valid" = true;
+      "mail.identity.id_${id}.htmlSigText" =
+        if account.signature.showSignature == "none" then
+          ""
```
teto, line 54: are we sure this is the correct value and not something like `false` ? I couldn't find a schema of the config

Lykos153:
https://github.com/nix-community/home-manager/blob/51e44a13acea71b36245e8bd8c7db53e0a3e61ee/modules/accounts/email.nix#L70 says it's an enumeration:

```
type = types.enum [ "append" "attach" "none" ];
```

modules/programs/thunderbird.nix
teto: I was thinking of the thunderbird config

Lykos153: Or do you mean the Thunderbird config `htmlSigText`? That's a string. It's what is inside the text field in the identity config window. And `""` makes that field empty.

PR #4853, opened 2024-01-04
alacritty: fix escape sequence in example and test

r-vdp:
### Description

See https://github.com/nix-community/home-manager/pull/4817#discussion_r1441710049

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@ncfavier

### DIFF
--- a/modules/programs/alacritty.nix
+++ b/modules/programs/alacritty.nix
@@ -32,7 +32,7 @@ in {
               {
                 key = "K";
                 mods = "Control";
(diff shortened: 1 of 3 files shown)


ncfavier (approved):

ncfavier (commented):
I think we should keep the backslash workaround as long as there isn't an easy way to get an arbitrary byte from the nixpkgs library.

rycee:

PR #4856, opened 2024-01-04
gradle: Re-add maintainer

britter:
Follow up to 0360475ee0fc870bc450874da5b5d7b2a85a091b


rycee:
It is sorted now. We had some issues due to the DDoS of Sourcehut.

Thanks for the fix, merged to master now 🙂

PR #4870, opened 2024-01-08
introduce `accounts.email.order` for aerc accounts

RyanGibb:
### Description

Nix does not preserve attrset order, instead sorting attributes by alphabetical order https://github.com/NixOS/nixpkgs/issues/81986.

However, aerc displays accounts based on their order in `.config/aerc/accounts.conf`.

This PR introduces a `accounts.email.order` list that provides an optional order for accounts in `accounts.email.accounts`.

I implement support for this in the `aerc` module, and I suspect other front-ends might also be able to benefit from this.

 If a partial order is provided those will be ordered after the primary account, and any unordered accounts will be included with the previous alphabetical behavior after.

### Checklist

    - The primary account is still displayed first and if no order list is provided the previous behavior is maintained.

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@lukasngl

### DIFF
--- a/modules/accounts/email.nix
+++ b/modules/accounts/email.nix
@@ -509,6 +509,13 @@ in {
       default = { };
       description = "List of email accounts.";
     };
+
+    order = mkOption {
+      type = types.listOf types.str;
+      default = [ ];
+      description =
+        "Order of email accounts from [](#opt-accounts.email.accounts).";
+    };
   };
 
   config = mkIf (cfg.accounts != { }) {
--- a/modules/programs/aerc.nix
+++ b/modules/programs/aerc.nix
@@ -23,6 +23,9 @@ let
 
   aerc-accounts =
     attrsets.filterAttrs (_: v: v.aerc.enable) config.accounts.email.accounts;
+  aerc-accounts-order =
+    builtins.filter (n: config.accounts.email.accounts.${n}.aerc.enable)
+    config.accounts.email.order;
 
   configDir = if (pkgs.stdenv.isDarwin && !config.xdg.enable) then
     "Library/Preferences/aerc"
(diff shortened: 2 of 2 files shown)


modules/accounts/email.nix
```diff
@@ -508,6 +508,13 @@ in {
       default = { };
       description = "List of email accounts.";
     };
+
+    order = mkOption {
+      type = types.listOf types.str;
+      default = [ ];
+      description =
+        "Order of email accounts from [](#opt-accounts.email.accounts).";
+    };
```
olmokramer, line 517: I think a nicer approach would be to add an `order` field to each account, and sort the accounts based on that field. Because with the current implementation I see a couple of issues:

* If you add an account to `accounts.email.accounts` but forget to add it to `accounts.email.order`, that account will be ignored by aerc.
* It is possible to make a typo in the account name, or have an entry in `accounts.email.order` that does not exist in `accounts.email.accounts`. Both of which will cause an error.

Of course there's also a downside: multiple accounts can have the same `order` value. But that's much easier to debug, and we could even add an assertion to prevent that from happening.

RyanGibb: Thanks for the review @olmokramer! To address your issues:

> If you add an account to accounts.email.accounts but forget to add it to accounts.email.order, that account will be ignored by aerc.

I think you misunderstand, any accounts not included in `order` will be included in the aerc account config by virtue of the `unorderedAccounts` variable.

If only some account names are in the `order` list then they will be included in that order after the primary account, and any additional accounts will be included with the previous behavior.

> It is possible to make a typo in the account name, or have an entry in accounts.email.order that does not exist in accounts.email.accounts. Both of which will cause an error.

This is true. Perhaps some assertions could make this easier to debug?

> a nicer approach would be to add an order field to each account

I'm not sure I agree. For me this would be less ergonomic, in that if one were to re-order an account you would need to change this index in multiple places, as opposed to moving a list entry.

olmokramer: Hey @RyanGibb thanks for the reply. You're right, I missed the `unorderedAccounts` so forget about that.

And I hadn't really thought about the need to reorder accounts, I can see that being a pain with my proposed method. I don't use Home Manager's email infrastructure, so I don't mind which direction this takes in the end.

Finally, I think some assertions would go a long way to prevent the errors I had in mind in my previous comment, or at least make them much easier to debug.

RyanGibb: Cool! I've added an assertion that fails with:
```
       Failed assertions:
       - ryan profile: 'example@example.com' is present in config.accounts.email.orders but not in config.accounts.email.accounts.
```

RyanGibb:
Hi @olmokramer, just putting a gentle ping here.

olmokramer:
Oh I'm sorry for just disappearing into the void! I think the changes look good! But I'm not in a position to merge PRs, sorry, I should have been more clear about that.

olmokramer (approved):

RyanGibb:
No problem at all! Thanks for the approval :-)

RyanGibb:
Can I ask what you think of this @rycee ?

RyanGibb:
Rebased on master

RyanGibb:
This is finished, but I'm no longer interested in trying to get it merged.

PR #4882, opened 2024-01-11
listenbrainz-mpd: use sdnotify when possible

DeeUnderscore:
### Description
The ability for listenbrainz-mpd to notify systemd when it's ready was added in 2.3.2: <https://codeberg.org/elomatreb/listenbrainz-mpd/releases/tag/v2.3.2>

listenbrainz-mpd in Nixpkgs is built with the `systemd` feature enabled, so this just works.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@Scrumplex

### DIFF
--- a/modules/services/listenbrainz-mpd.nix
+++ b/modules/services/listenbrainz-mpd.nix
@@ -40,6 +40,10 @@ in {
         ExecStart = "${cfg.package}/bin/listenbrainz-mpd";
         Restart = "always";
         RestartSec = 5;
+        Type = if lib.versionAtLeast cfg.package.version "2.3.2" then
+          "notify"
+        else
+          "simple";
       };
       Install.WantedBy = [ "default.target" ];
     };


Scrumplex (approved):

modules/services/listenbrainz-mpd.nix
```diff
@@ -40,6 +40,10 @@ in {
         ExecStart = "${cfg.package}/bin/listenbrainz-mpd";
         Restart = "always";
         RestartSec = 5;
+        Type = if (lib.versionAtLeast cfg.package.version "2.3.2") then
```
rycee, line 43: Can remove the unnecessary parentheses
```suggestion
        Type = if lib.versionAtLeast cfg.package.version "2.3.2" then
```

rycee:
Thanks! Merged to master now.

PR #4888, opened 2024-01-12
home-manager: Check VISUAL before EDITOR for editor

srstevenson:
Check VISUAL for a visual editor before EDITOR, as per Unix convention
and as followed by Git, crontab, mutt, and other tools.

### DIFF
--- a/home-manager/home-manager
+++ b/home-manager/home-manager
@@ -571,10 +571,14 @@ function presentNews() {
 }
 
 function doEdit() {
-    if [[ ! -v EDITOR || -z $EDITOR ]]; then
-        # shellcheck disable=2016
-        _i 'Please set the $EDITOR environment variable' >&2
-        return 1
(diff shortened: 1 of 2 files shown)


home-manager/home-manager
```diff
@@ -920,7 +924,7 @@ function doHelp() {
     echo
     echo "  help         Print this help"
     echo
-    echo "  edit         Open the home configuration in \$EDITOR"
+    echo "  edit         Open the home configuration in \$VISUAL or \$EDITOR"
```
rycee, line 927: This would need a similar update in `docs/home-manager.1`.

rycee:
Thanks! Makes sense. I think the only thing missing is to also update the `home-manager` man page.

home-manager/home-manager
srstevenson: Done 🙂

somasis (approved):

PR #4890, opened 2024-01-12
zig: add module programs.zig

luisnquin:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/zig.nix
+++ b/modules/programs/zig.nix
@@ -1,0 +1,47 @@
+{ config, pkgs, lib, ... }:
+
+with lib;
+
+let
+
+  cfg = config.programs.zig;
+
+in {
+  meta.maintainers = [ maintainers.luisnquin ];
+
+  options = {
+    programs.zig = {
+      enable = mkEnableOption "zig";
+
+      package = mkOption pkgs {
+        type = types.package;
+        default = pkgs.zig;
+        defaultText = literalExpression "pkgs.zig";
+        description = "The Zig package to use.";
+      };
+
+      enableZshIntegration = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          Whether to enable Zig’s Zsh integration.
+        '';
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    programs.zsh.plugins = mkIf cfg.enableZshIntegration [{
+      name = "zig-zsh-completions-plugin";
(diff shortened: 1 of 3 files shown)


modules/programs/zig.nix
```diff
@@ -0,0 +1,47 @@
...
+        description = "The Zig package to use.";
+      };
+
+      enableZshIntegration = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          Whether to enable Zig’s Zsh integration.
+        '';
+      };
```
sumnerevans, line 29:
```suggestion
      enableZshIntegration = mkEnableOption "Zig's Zsh integration";
```

modules/programs/zig.nix
```diff
@@ -0,0 +1,47 @@
...
+    programs.zsh.plugins = mkIf cfg.enableZshIntegration [{
+      name = "zig-zsh-completions-plugin";
+      file = "zig-shell-completions.plugin.zsh";
+      src = pkgs.fetchFromGitHub {
+        owner = "ziglang";
+        repo = "shell-completions";
+        rev = "31d3ad12890371bf467ef7143f5c2f31cfa7b7c1";
+        sha256 = "1fzl1x56b4m11wajk1az4p24312z7wlj2cqa3b519v30yz9clgr0";
+      };
+    }];
```
sumnerevans, line 45: definitely don't love this... Maintaining a package in home-manager seems like the wrong approach.

Any way that we can get this plugin into nixpkgs first?

sumnerevans (changes requested):

modules/programs/zig.nix
luisnquin: I'm barely new creating options, I just aligned myself to the pattern used in other options, [example here](https://github.com/nix-community/home-manager/blob/b84191db127c16a92cbdf7f7b9969d58bb456699/modules/programs/direnv.nix#L59).I'm barely new to creating options, I just aligned myself with the pattern used in other options, example here.

luisnquin: I'm barely new to creating options, I just aligned myself with the pattern used in [other options](https://github.com/nix-community/home-manager/blob/b84191db127c16a92cbdf7f7b9969d58bb456699/modules/programs/direnv.nix#L59).

luisnquin:
I just created a new issue in https://github.com/NixOS/nixpkgs/issues/281900. Thanks for your feedback @sumnerevans.

PR #4895, opened 2024-01-14
rbw/gpg-agent: improve `pinentry` options

ambroisie:
### Description

This is mostly a follow-up to #4805.

### Checklist

     - Technically the change of `default` isn't backwards compatible, but breaks on new nixpkgs versions anyway...

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee


modules/programs/rbw.nix
```diff
@@ -52,15 +51,15 @@ let
...
           '';
           # we want the program in the config
           apply = val:
             if builtins.isString val then
               "${pkgs.pinentry.${val}}/bin/pinentry"
+            else if val ? binaryPath then
+              "${val}/${val.binaryPath}"
             else
-              "${val}/${val.binaryPath or "bin/pinentry"}";
+              lib.getExe val;
```
ambroisie, line 62: I'm not sure if we could migrate `pinentry_mac` to define `meta.mainProgram`, that could remove the `binaryPath` part and rely on `lib.getExe` for every package.

modules/services/gpg-agent.nix
```diff
@@ -246,8 +256,7 @@ in {
           ++ optional (cfg.maxCacheTtlSsh != null)
           "max-cache-ttl-ssh ${toString cfg.maxCacheTtlSsh}"
           ++ optional (cfg.pinentryFlavor != null)
-          "pinentry-program ${pkgs.pinentry.${cfg.pinentryFlavor}}/bin/pinentry"
-          ++ [ cfg.extraConfig ]);
+          "pinentry-program ${pinentryFlavor}" ++ [ cfg.extraConfig ]);
```
ambroisie, line 259: Kind of annoyed at the list being inlined here, but the format script wouldn't be forced into putting a newline here...

ambroisie:
I tried running all tests, but got an FOD error in `immutable-keyfiles.nix`...

<details>

```none
building '/nix/store/0pka8pqyqk6k8i43112423y8hxsm0bd5-activation-script.drv'...
building '/nix/store/35wnsrj3m2npjdwvxf6dga0li3f916bc-home-manager-generation.drv'...
building '/nix/store/dpwjima15fr4k9kanlkn8sl82xa82gmd-nmt-report-gpg-agent-override-homedir.drv'...
building '/nix/store/y67jnqdl22qf4vji9yjw0r6q3i1f127c-home-manager-path.drv'...
building '/nix/store/kiddsys4690v680mhnvlx4zpgz50bb02-lookup?op=get-options=mr-search=0x36cacf52d098cc0e78fb0cb13573356c25c424d4.drv'...
building '/nix/store/28v6fwhzdaa4kc5nv5g92bh3nxk1fscm-pubkey.txt.drv'...
created 299 symlinks in user environment

trying https://keys.openpgp.org/pks/lookup?op=get&options=mr&search=0x36cacf52d098cc0e78fb0cb13573356c25c424d4
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

trying https://www.rsync.net/resources/pubkey.txt
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1855  100  1855    0     0   1440      0  0:00:01  0:00:01 --:--:--  1441
100 13209  100 13209    0     0   8088      0  0:00:01  0:00:01 --:--:--  8083
error: hash mismatch in fixed-output derivation '/nix/store/kiddsys4690v680mhnvlx4zpgz50bb02-lookup?op=get-options=mr-search=0x36cacf52d
098cc0e78fb0cb13573356c25c424d4.drv':
         specified: sha256-9Zjsb/TtOyiPzMO/Jg3CtJwSxuw7QmX0pcfZT2/1w5E=
            got:    sha256-Pvm0CpH13pvnD/c+IJAkRElR5/IhZt8pg7RjT/dZdfM=
error: 1 dependencies of derivation '/nix/store/07qfnkm1l1zz8519x3y824a5jibkjgbg-gpg-pubring.drv' failed to build
error: 1 dependencies of derivation '/nix/store/n7ghchmzg9k0hx6db9dfvgs1lbqj9qva-activation-script.drv' failed to build
error: 1 dependencies of derivation '/nix/store/4px1ar29q9dghisziqcncc1xw6as3nm4-home-manager-generation.drv' failed to build
error:

       … while evaluating derivation 'nmt-run-all-tests'
         whose name attribute is located at /nix/store/lrmxkfadhlcrakza60w0mc8sc327a5qg-source/pkgs/stdenv/generic/make-derivation.nix:348:7

       … while evaluating attribute 'shellHook' of derivation 'nmt-run-all-tests'

         at /nix/store/3vax7vf054q4vvb1bj8sd9lvr8dcsfjb-source/default.nix:38:40:

           37|   runShellOnlyCommand = name: shellHook:
           38|     pkgs.runCommandLocal name { inherit shellHook; } ''
             |                                        ^
           39|       echo This derivation is only useful when run through nix-shell.

       error: 1 dependencies of derivation '/nix/store/28rshsv810bbm5lr3nzjlvi9c7alcr92-nmt-report-gpg-immutable-keyfiles.drv' failed to build
```

<details/>

modules/programs/rbw.nix
ambroisie: See [this PR]() which adds it to `pinentry_mac` (and others).

See [that PR]() which adds it to the various `pinentry.<flavor>` packages.

So we could further improve this once both PRs are merged, though it shouldn't block this one.

ambroisie (commented):
The fixup commit should be rebases before merging.

fpletz:
This PR makes the pinentry mess even more uglier. IMHO we should only support flavor strings as `pinentryFlavor` and pick the correct package via `pkgs.pinentry-${pinentryFlavor}` like in https://github.com/NixOS/nixpkgs/pull/277221/files#diff-9aace735eba8386316153e3fd2ba7c726a1f6ccfc37a981c0efcc173ed2481f3R90.

If we wanted to support pinentry packages a new option like `pinentryPackage` would make more sense instead of making this one option more complicated to use than it should be. We should rather move away from multiple outputs anyway.

ambroisie:
I think it's essential to support packages, not just flavors -- for one, I would rather use `pinentry-rofi` than any of the specific flavors made available by the package. It also sounds like a more modular way of specifying the option.

Rather than have two competing `pinentry*` options, I think we could deprecate `pinentryFlavor` (in `gpg-agent`) and introduce a `pinentry` option which is a `mkPackageOption { null = true; }`.
For `rbw`, we could add a warning if the user's option is using a string "flavor" to introduce the deprecation.

ambroisie:
Switching this to a draft until https://github.com/NixOS/nixpkgs/pull/133542 is merged.

ambroisie:
@fpletz now that the upstream changes have been merged, I updated this PR.

modules/services/gpg-agent.nix
```diff
@@ -192,10 +197,9 @@ in {
           configuration file.
         '';
       };
-
-      pinentryFlavor = mkOption {
-        type = types.nullOr (types.enum pkgs.pinentry.flavors);
-        example = "gnome3";
+      agent.pinentryPackage = mkOption {
```
ambroisie, line 200:
```suggestion
      pinentryPackage = mkOption {
```

modules/services/gpg-agent.nix
```diff
@@ -81,6 +81,11 @@ let
 in {
   meta.maintainers = [ maintainers.rycee ];
 
+  imports = [
+    (mkRemovedOptionModule [ "programs" "gnupg" "agent" "pinentryFlavor" ]
+      "Use programs.gnupg.agent.pinentryPackage instead")
```
ambroisie, line 86:
```suggestion
    (mkRemovedOptionModule [ "services" "gpg-agent" "pinentryFlavor" ]
      "Use services.gpg-agent.pinentryPackage instead")
```

modules/services/gpg-agent.nix
rycee: The standard trick for forcing newlines used elsewhere is to put `# \` at the end of the line. For example,

``` nix
          ++ optional (cfg.pinentryFlavor != null)
          "pinentry-program ${pinentryFlavor}" # \
          ++ [ cfg.extraConfig ]);
```

It's not the most pretty but I think it's not too bad.

modules/misc/news.nix
```diff
@@ -1436,6 +1436,22 @@ in {
           A new module is available: 'programs.ranger'.
         '';
       }
+
+      {
+        time = "2024-03-10T10:27:11+00:00";
+        message = ''
+          'services.gpg-agent.pinentryFlavor' has been removed and replaced by.
```
rycee, line 1443:
```suggestion
          'services.gpg-agent.pinentryFlavor' has been removed and replaced by
```

modules/programs/rbw.nix
```diff
@@ -41,9 +41,9 @@ let
         };
 
         pinentry = mkOption {
```
rycee, line 43: Maybe worth renaming this as well to `pinentryPackage`. Then could also add a `mkRemovedOptionModule`?

ambroisie: It's used directly by the config file, which expects `pinentry`.

modules/programs/rbw.nix
```diff
@@ -87,7 +81,7 @@ in {
         {
           email = "name@example.com";
           lock_timeout = 300;
-          pinentry = "gnome3";
+          pinentry = "pkgs.pinentry-gnome";
```
rycee, line 84: I guess this should be
```suggestion
          pinentry = pkgs.pinentry-gnome;
```

modules/programs/rbw.nix
rycee: Ah, right. Fair enough!

modules/misc/news.nix
```diff
@@ -1436,6 +1436,22 @@ in {
           A new module is available: 'programs.ranger'.
         '';
       }
+
+      {
+        time = "2024-03-10T10:27:11+00:00";
+        message = ''
```
rycee, line 1442: Could limit this so that it's only shown to people that have the module enabled:
```suggestion
        condition = config.services.gpg-agent.enable;
        message = ''
```

modules/misc/news.nix
```diff
@@ -1436,6 +1436,22 @@ in {
...
+          'services.gpg-agent.pinentryFlavor' has been removed and replaced by
+          'services.gpg-agent.pinentryPackage'.
+        '';
+      }
+
+      {
+        time = "2024-03-10T10:27:11+00:00";
+        message = ''
```
rycee, line 1450: Same here:
```suggestion
        condition = config.programs.rbw.enable;
        message = ''
```

rycee:
Thanks! I just added a few minor final comments. You will also have to run the `./format` script in the project root to make the CI happy.

Perhaps this PR could be marked as ready to review?

modules/misc/news.nix
ambroisie: Oh I forgot that existed.

ambroisie:
Urgh, so much back and forth, sorry about that @rycee.

Should be good now though.

modules/programs/rbw.nix
```diff
@@ -87,7 +81,7 @@ in {
         {
           email = "name@example.com";
           lock_timeout = 300;
-          pinentry = "gnome3";
+          pinentry = pkgs.pinentry-gnome;
```
Stzx, line 84: Upstream seems to be only `pinentry-gnome3` ?

ambroisie: Wait no, my branch might be too old.

Stzx: So this might not work on `unstable`? `unstable` has been renamed to `pinentry-gnome3` `23.11` is still `pinentry-gnome`

ambroisie: This is just an example...

rycee:
Thanks! Merged to master now 🙂 Note, I switched back to pinentry-gnome3 since it seems to be what is in Nixpkgs right now. I'm a bit confused, though, since I'm pretty sure I saw pinentry-gnome for a while.

modules/services/gpg-agent.nix
```diff
@@ -243,8 +247,8 @@ in {
           "max-cache-ttl ${toString cfg.maxCacheTtl}"
           ++ optional (cfg.maxCacheTtlSsh != null)
           "max-cache-ttl-ssh ${toString cfg.maxCacheTtlSsh}"
-          ++ optional (cfg.pinentryFlavor != null)
-          "pinentry-program ${pkgs.pinentry.${cfg.pinentryFlavor}}/bin/pinentry"
+          ++ optional (cfg.pinentryPackage != null)
+          "pinentry-program ${lib.getExe pinentryPackage}"
```
yshui, line 251: shouldn't this be `cfg.pinentryPackage`?

musjj, line 251: I'm late, but this should be:
```diff
- "pinentry-program ${lib.getExe pinentryPackage}"
+ - "pinentry-program ${lib.getExe cfg.pinentryPackage}"
```

Getting this error:
```bash
┃        error: undefined variable 'pinentryPackage'
┃
┃        at /nix/store/kww2vp5mhnvb0mczv2xll0m5qsrjrw9q-source/modules/services/gpg-agent.nix:251:42:
┃
┃           250|           ++ optional (cfg.pinentryPackage != null)
┃           251|           "pinentry-program ${lib.getExe pinentryPackage}"
┃              |                                          ^
┃           252|           ++ [ cfg.extraConfig ]);
```

musjj: Woops, review sniped :)

colemickens:
I don't understand the choice to specify package here rather than binary. My pinentry of choice does not provide it's binary as the default exe, nor is it named pinentry.

danjl1100:
I didn't see the removal of `pinentryFlavor` mentioned in the [24.05 release notes](https://nix-community.github.io/home-manager/release-notes.xhtml#sec-release-24.05). This is mentioned in `news.nix` for the PR... is there some easy way to find that when upgrading home-manager? (e.g. 23.11 -> 24.05)

---
Is the required change in home-manager configs as simple as this?
```patch
 # example for `pinentry-curses`
-services.gpg-agent.pinentryFlavor = "curses";
+services.gpg-agent.pinentryPackage = pkgs.pinentry-curses;
```

PR #4907, opened 2024-01-15
made activationPackage working when crossSystem is set

dguibert:
### Description

In NixOS setting CrossSystem allow to compile for a system that can be different of the host system but the tools that need to be run on the build host must come from pkgs.buildPackages.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


rycee:
Thanks for the contribution! This seems quite valuable.

I'm not super fond of the `activationPackageSet` naming but can't think of anything better at the moment except perhaps `activationPkgs`, which somewhat mirrors the naming in Nixpkgs (e.g., `crossPkgs`).

Also, I'm wondering if not `_module.args.activationPkgs` would be better than `home.activationPackageSet` so that it could be taken from the module argument similar to `pkgs`.

Finally, I think that it would be good if we could come up with a way to test this because I imagine the `pkgs != activationPkgs` case will not be very widely used. We will definitely miss using `activationPkgs` unless the CI catches it.

dguibert:
> Also, I'm wondering if not `_module.args.activationPkgs` would be better than `home.activationPackageSet` so that it could be taken from the module argument similar to `pkgs`.

I've applied your suggestion.

Still need to implement some testing...

Issue #4909, opened 2024-01-16
bug: rtx has been renamed to mise


AntonFriberg:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

rtx has been renamed to mise to avoid confusion with Nvidia's brand of graphics cards. See https://mise.jdx.dev/rtx.html for more info. Further, https://github.com/jdx/rtx now points to https://github.com/jdx/mise

nixpkgs have merged a fix https://github.com/NixOS/nixpkgs/pull/279158

Not sure how to handle a breaking change like this but programs.rtx needs some changes due to this 
https://github.com/nix-community/home-manager/blob/37d6eeceee464adc03585404eebd68765b3c8615/modules/programs/rtx.nix 

I think simply changing the links and configuration file paths should be enough but the question is how to notify users of the breaking change so that they change their project specific config files. Probably enough to reference rtx official documentation.

### Maintainer CC

@pedorich-n 

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 5.10.0-27-amd64, Debian GNU/Linux, 11 (bullseye), nobuild`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.17.0`
 - channels(root): `"nixpkgs"`
 - channels(afriberg): `"home-manager"`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixpkgs`
```

pedorich-n:
Thanks for the issue!
I saw the program being renamed but missed a moment when the change got merged into `nixpkgs`.
I searched for similar cases in `home-manager`, but the only PR I found is this: https://github.com/nix-community/home-manager/pull/3265. They just moved `programs` to `services`.

So, I think I can just rename the program. I will add a news entry about breaking change as well.

PR #4910, opened 2024-01-16
rtx: rename to mise

pedorich-n:
### Description

Fixes: https://github.com/nix-community/home-manager/issues/4909
`rtx` was renamed to `mise`: https://mise.jdx.dev/rtx.html
`nixpkgs`: https://github.com/NixOS/nixpkgs/pull/279158

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@pedorich-n


AntonFriberg (approved):
Looks good from my point of view!

AntonFriberg:
There is one issue and that is that the [settings] section of the generated example is no longer supported and will result in a warning.

The example in the description
```nix
tools = {
  node = "lts";
  python = ["3.10" "3.11"];
};

settings = {
  verbose = false;
  experimental = false;
};
```
Will result in the following `.config/mise/config.toml`
```
[settings]
experimental = false
verbose = false

[tools]
node = "lts
python = ["3.10", "3.11"]
```

This will give you a warning
```
mise [settings] inside of ~/.config/mise/config.toml is deprecated. A separate ~/.config/mise/settings.toml file should be used instead.
mise Removing the [settings] section from ~/.config/mise/config.toml will remove this warning
```

pedorich-n:
Huh, I missed when the file got split into two. 
Let me change the module.

AntonFriberg:
I think globalConfig makes sense. I can not think of anything either! Thanks for the quick response and making it work! I really like this addition to home-manager and use it daily.

modules/programs/mise.nix
```diff
@@ -3,16 +3,16 @@
 with lib;
 
 let
-  cfg = config.programs.rtx;
+  cfg = config.programs.mise;
   tomlFormat = pkgs.formats.toml { };
 in {
   meta.maintainers = [ hm.maintainers.pedorich-n ];
```
rycee, line 10: Should add a `mkRemovedOptionModule` import to notify users about the move. It may be sufficient to notify about the `rtx.enable` option. Something like

```suggestion

  imports = [
    (mkRemovedOptionModule [ "programs" "rtx" "enable" ] ''
      The rtx package has been replaced by mise, please switch over to
      using the options under

          programs.mise.*

      instead.
    '')
  ];

```

perhaps.

modules/misc/news.nix
```diff
@@ -1383,6 +1383,13 @@ in {
           A new module is available: 'xdg.portal'.
         '';
       }
+
+      {
+        time = "2024-01-16T15:43:51+00:00";
+        message = ''
+          'programs.rtx' was renamed to 'programs.mise'
```
rycee, line 1390:
```suggestion
          A new module is available: 'programs.mise'.
          
          This module replaces 'programs.rtx', which has been removed.
```

modules/programs/mise.nix
rycee: I don't know how compatible the configuration is between rtx and mise. If all rtx confiurations are valid mise configurations, then you could even use `mkRenamedOptionModule` for the different rtx options.

pedorich-n: Thanks! That's exactly what I need.
Given the split of the configuration file now they're not exactly the same anymore. So I think it's better to use `mkRemovedOptionModule`

pedorich-n:
@rycee please have a look. I've added the `mkRemovedOptionModule` and changed the `news.nix` entry.

modules/programs/mise.nix
```diff
@@ -0,0 +1,111 @@
...
+        The `rtx` package has been replaced by `mise`, please switch over to
+        using the options under `programs.mise.*` instead.
+      '');
+
+  in map mkRemovedWarning [
+    "enable"
+    "package"
+    "enableBashIntegration"
+    "enableZshIntegration"
+    "enableFishIntegration"
```
sumnerevans, line 23: These all seem like they can be `mkRenamedOptionModule`, and just make `settings` be a `mkRemovedOptionModule`

rycee: That would be cool. Perhaps best not to rename `package`, though since I guess it is likely to point to an rtx package in existing configurations?

pedorich-n: Given that the configuration from `rtx` doesn't translate fully to `mise` and `settings` now means a different thing I would argue that it's better to fail fast and let the user know they must take action. 

I don't think we should allow configuration like this to be valid

```nix
    rtx = {
      enable = true;
      enableFishIntegration = false;
    };

    mise = {
      settings = {
        experimental = true;
      };
      globalConfig = {
        tools = {
          python = [ "3.11" ];
        };
      };
    };
```

rycee:

pedorich-n:
Thanks!

PR #4915, opened 2024-01-17
hyprland: do not override existing plugins settings in config

SamueleFacenda:
The plugin setting in the hyprland config is used both for defining plugin paths and configuring the plugins. This fix removes the silent override of the plugins settings converting them to the `plugin:<name> { ...settings }` syntax.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@fufexan

### DIFF
--- a/modules/services/window-managers/hyprland.nix
+++ b/modules/services/window-managers/hyprland.nix
@@ -207,7 +207,8 @@ in {
             else
               entry;
         in map mkEntry cfg.plugins;
-      };
+      } // lib.mapAttrs' (n: v: lib.nameValuePair "plugin:${n}" v)
+        (lib.attrByPath [ "plugin" ] { } cfg.settings);
 
       shouldGenerate = cfg.systemd.enable || cfg.extraConfig != ""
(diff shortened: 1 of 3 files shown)


fufexan:
Defining plugin paths should not be done through `settings.plugin`, but through the `plugins` option. I haven't had issues with that setup.

However, if you've tested this and it works, then it looks good to me.

SamueleFacenda:
Yes sorry. I mean that plugin settings are defined via `settings.plugin.<pluginname>` and this option is overwitted by the `plugin` option generated from `hyprland.plugins`.

rycee:
I sort of think I understand what is going on here, and agree with @fufexan that it seems reasonable. But I would like to see `tests/modules/services/window-managers/hyprland/simple-config.nix` updated to show it more concretely (and to avoid future regressions).

SamueleFacenda:
Ok. I updated the test.

rycee:

SamueleFacenda:
Great, thanks!

PR #4920, opened 2024-01-17
gradle: Don't enable programs.java

britter:
The gradle package from nixpkgs will install a default Java version and
pass it to the gradle executable without polluting the global env.
Users can customize this by overriding the gradle package like so:

```nix
  programs.gradle = {
    package = (pkgs.gradle.override {
      java = pkgs.jdk21;
    });
  }
```

### DIFF
--- a/modules/programs/gradle.nix
+++ b/modules/programs/gradle.nix
@@ -110,7 +110,5 @@ in {
     home.sessionVariables = mkIf (cfg.home != defaultHomeDirectory) {
       GRADLE_USER_HOME = gradleHome;
     };
-
-    programs.java.enable = true;
   };
 }


britter:
Actually I'm not sure this is a good idea, because the java package configured by `nixpkgs.gradle` does not work with the gradlew wrapper script that almost everybody uses. So maybe we should keep enabling Java for now.

britter:
On second thought I think we should actually remove this. The fact that most people decide to add a wrapper script to their repository has nothing to do with the fact that nixpkgs.gradle will work without enabling `programs.java`. So this module should not enable it unnecessarily.

rycee:
Agreed 🙂 Merged to master now.

PR #4921, opened 2024-01-18
gnome-keyring: add gpg component and optionally use security wrapper

Aleksanaa:
### Description

1. gnome-keyring supports managing GPG keys by default, provided with `gpg` component. I've used this myself for a long time without issues.
2. If the binary doesn't have `CAP_IPC_LOCK`, it will complain `no process capabilities, insecure memory might get used` (`journalctl --user -xeu gnome-keyring.service`).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee


modules/services/gnome-keyring.nix
```diff
@@ -13,8 +13,16 @@ in {
     services.gnome-keyring = {
       enable = mkEnableOption "GNOME Keyring";
 
+      useSecurityWrapper = mkEnableOption ''
+        using gnome-keyring-daemon wrapped by NixOS security wrapper
+        (i.e. {file}`/run/wrappers/bin/gnome-keyring-daemon`) with
+        `CAP_IPC_LOCK` to enhance memory security. This option will
+        only work on NixOS with system-wide option
+        {option}`services.gnome.gnome-keyring.enable` is enabled.
```
sumnerevans, line 21:
```suggestion
        only work on NixOS with the system-wide
        {option}`services.gnome.gnome-keyring.enable` option enabled
```

sumnerevans (changes requested):

Aleksanaa:
Thanks for marking my PR as stale!

Aleksanaa:
🙄

PR #4925, opened 2024-01-18
Deprecate `DRY_RUN_CMD` and `DRY_RUN_NULL`

rycee:
### Description

This adds the `dryRun` helper function. It deprecates the `DRY_RUN_CMD` and `DRY_RUN_NULL` variables.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/lib/bash/home-manager.sh
+++ b/lib/bash/home-manager.sh
@@ -83,3 +83,27 @@ function _iNote() {
     _i "$@"
     echo -n "${normalColor}"
 }
+
+# Runs the given command on live run, otherwise prints the command to standard
+# output.
+#
+# If given the command line option `--silence`, then the command's standard and
+# error output is sent to `/dev/null` on a live run.
+#
+# Note, the run function is exported. I.e., it is available also to called Bash
+# script.
+function run() {
+    if [[ $1 == '--silence' ]]; then
+        local silence=1
+        shift
+    fi
+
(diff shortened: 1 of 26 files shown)


lib/bash/home-manager.sh
```diff
@@ -83,3 +83,19 @@ function _iNote() {
     _i "$@"
     echo -n "${normalColor}"
 }
+
+# Runs the given command on live run, otherwise prints the command to standard
+# output.
+#
+# If given the command line option `--silence`, then the command's standard and
+# error output is sent to `/dev/null` on a live run.
+function dryRun() {
```
teto, line 92: more than `dryRun`, its a `runCommand` that might be a dryrun

teto: ie. not a fan of the name, bikeshedding a bit.

rycee: Yes, I agree about the naming being poor, I was having difficulties finding a good naming scheme, though. An alternative I considered but figured maybe was too long was something like `dryEchoLiveRun`.

FWIW I'm thinking to add a `verboseEcho` as well (as a replacement of the `VERBOSE_ECHO` variable).

rycee: Had a think and ended up with simply calling the function `run`. It seems more sensible to me.

PR #4926, opened 2024-01-19
mullvad-browser: initial support

heywoodlh:
### Description

Initial support for configuring Mullvad Browser.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Additional Info

I tested my fork and the Mullvad Browser seems to be configured properly on both Linux and MacOS! :smile:

I'm not sure if the tests I added will work. Since Mullvad Browser is just a copy of Firefox, I tried to copy Firefox and modify the tests in a way that made sense. I could use some help on that. I did run the tests on Linux and I didn't seem to have issues -- but, again, I could use help on the tests.


brckd:
i tried to use the module and it seems like you forgot to import it in `modules/modules.nix` and add the package to `home.packages`

heywoodlh:
@brckd I believe I added the necessary changes, could you test again and let me know?

brckd:
for some reason i get [this error](https://l.termbin.com/mfss) when trying to `config.programs.mullvad-browser.enable = true`

heywoodlh:
With https://github.com/nix-community/home-manager/pull/5128 merged I'm going to close this as it's likely unnecessary at this point since Mullvad Browser is a Firefox derivative.

PR #4927, opened 2024-01-19
mcfly: Add mcfly-fzf integration

NovaViper:
### Description
This PR adds integration for [mc-fly](https://github.com/bnprks/mcfly-fzf), what it does it just adds `eval "$(${pkgs.mcfly-fzf}/bin/mcfly-fzf init $shell)"` to each of the shells (bash, zsh, and fish) right after mcfly's initialization script. This is ensures mcfly-fzf is loaded in order (it has to be loaded after mcfly); also checks if any of the shell integrations are disabled (since if those are disabled, then there's no way to add the initialization script for it anyway). 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@marsam 
@aisamu

### DIFF
--- a/modules/programs/mcfly.nix
+++ b/modules/programs/mcfly.nix
@@ -5,6 +5,24 @@ let
 
   cfg = config.programs.mcfly;
 
+  bashIntegration = ''
+    eval "$(${getExe pkgs.mcfly} init bash)"
+  '' + optionalString cfg.fzf.enable ''
+    eval "$(${getExe pkgs.mcfly-fzf} init bash)"
+  '';
+
+  fishIntegration = ''
+    ${getExe pkgs.mcfly} init fish | source
+  '' + optionalString cfg.fzf.enable ''
+    ${getExe pkgs.mcfly-fzf} init fish | source
+  '';
+
+  zshIntegration = ''
+    eval "$(${getExe pkgs.mcfly} init zsh)"
+  '' + optionalString cfg.fzf.enable ''
+    eval "$(${getExe pkgs.mcfly-fzf} init zsh)"
+  '';
+
 in {
   meta.maintainers = [ ];
 
@@ -29,6 +47,8 @@ in {
         Key scheme to use.
       '';
     };
+
+    fzf.enable = mkEnableOption "McFly fzf integration";
 
     enableLightTheme = mkOption {
       default = false;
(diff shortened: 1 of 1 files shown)


marsam:
Cool, TIL about mcfly-fzf.
What do you think of writing its integration like the following?
```diff
diff --git c/modules/programs/mcfly.nix i/modules/programs/mcfly.nix
index 75f6f2b3..b56a058f 100644
--- c/modules/programs/mcfly.nix
+++ i/modules/programs/mcfly.nix
@@ -5,6 +5,24 @@ let
 
   cfg = config.programs.mcfly;
 
+  bashIntegration = ''
+    eval "$(${getExe pkgs.mcfly} init bash)"
+  '' + optionalString cfg.mcfly-fzf.enable ''
+    eval "$(${getExe pkgs.mcfly-fzf} init bash)"
+  '';
+
+  fishIntegration = ''
+    ${getExe pkgs.mcfly} init fish | source
+  '' + optionalString cfg.mcfly-fzf.enable ''
+    ${getExe pkgs.mcfly-fzf} init fish | source
+  '';
+
+  zshIntegration = ''
+    eval "$(${getExe pkgs.mcfly} init zsh)"
+  '' + optionalString cfg.mcfly-fzf.enable ''
+    eval "$(${getExe pkgs.mcfly-fzf} init zsh)"
+  '';
+
 in {
   meta.maintainers = [ ];
 
@@ -30,6 +48,8 @@ in {
       '';
     };
 
+    mcfly-fzf = { enable = mkEnableOption "McFly fzf integration"; };
+
     enableLightTheme = mkOption {
       default = false;
       type = types.bool;
@@ -77,17 +97,11 @@ in {
     {
       home.packages = [ pkgs.mcfly ];
 
-      programs.bash.initExtra = mkIf cfg.enableBashIntegration ''
-        eval "$(${pkgs.mcfly}/bin/mcfly init bash)"
-      '';
+      programs.bash.initExtra = mkIf cfg.enableBashIntegration bashIntegration;
 
-      programs.zsh.initExtra = mkIf cfg.enableZshIntegration ''
-        eval "$(${pkgs.mcfly}/bin/mcfly init zsh)"
-      '';
+      programs.zsh.initExtra = mkIf cfg.enableZshIntegration zshIntegration;
 
-      programs.fish.shellInit = mkIf cfg.enableFishIntegration ''
-        ${pkgs.mcfly}/bin/mcfly init fish | source
-      '';
+      programs.fish.shellInit = mkIf cfg.enableFishIntegration fishIntegration;
 
       home.sessionVariables.MCFLY_KEY_SCHEME = cfg.keyScheme;
     }
```

NovaViper:
> Cool, TIL about mcfly-fzf. What do you think of writing its integration like the following?

Done! Got it all cleaned up and gave it a test on my computer and it works perfectly. Should be ready to be merged now

modules/programs/mcfly.nix
```diff
@@ -30,6 +48,8 @@ in {
       '';
     };
 
+    mcfly-fzf.enable = mkEnableOption "McFly fzf integration";
```
rycee, line 51: I think you could omit the `mcfly-` from the option name, it seems a bit redundant.
```suggestion
    fzf.enable = mkEnableOption "McFly fzf integration";
```

modules/programs/mcfly.nix
```diff
@@ -74,20 +94,34 @@ in {
   };
 
   config = mkIf cfg.enable (mkMerge [
+    (mkIf cfg.mcfly-fzf.enable {
```
rycee, line 97: I don't think the assertions are necessary, and indeed may cause false positives. For example, if you have `enableFishIntegration == false` but `programs.fish.enable = false`

modules/programs/mcfly.nix
NovaViper: Yeah it does, just went ahead and added the change

rycee:

PR #4933, opened 2024-01-20
hyprland: change plugins settings generation

SamueleFacenda:
Sorry for being back so quickly, but digging in the hyprland source code (and debugging my config) I found out that the syntax that I used in #4915 will lead to a [silent](https://github.com/hyprwm/Hyprland/blob/f40e382fc6208d4fe2e53581ea27510cb62417dd/src/config/ConfigManager.cpp#L455C13-L455C40) wrong config parsing. The result is that all the sections after a `plugin:name` section are ignored (just the sections, not the fields).
There is no documentation, just the [source code](https://github.com/hyprwm/Hyprland/blob/main/src/config/ConfigManager.cpp#L1574). 
With this update the plugins paths are generate right before the full config and the plugins settings are generated like the other fields.

- [x] Code tested through `nix develop --ignore-environment .#hyprland-*`

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@fufexan

### DIFF
--- a/modules/services/window-managers/hyprland.nix
+++ b/modules/services/window-managers/hyprland.nix
@@ -199,19 +199,8 @@ in {
     home.packages = lib.optional (cfg.package != null) cfg.finalPackage;
 
     xdg.configFile."hypr/hyprland.conf" = let
-      combinedSettings = cfg.settings // {
-        plugin = let
-          mkEntry = entry:
-            if lib.types.package.check entry then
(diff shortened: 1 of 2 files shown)


fufexan (approved):
LGTM.

(Actually, the syntax you used in your previous PR would be valid if Hyprland was using hyprlang, but it currently isn't.)

rycee:
Thanks! I merged this to master yesterday :-)

PR #4939, opened 2024-01-22
tealdeer: add cache update activation script

pedorich-n:
### Description

Since by default `tldr`'s cache [auto-update is disabled](https://dbrgn.github.io/tealdeer/config_updates.html#auto_update) this activation script will call `tldr --update` on home-manager switch. 

Build  similar to [`programs.bat`](https://github.com/nix-community/home-manager/blob/2d47379ad591bcb14ca95a90b6964b8305f6c913/modules/programs/bat.nix#L162-L168)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

No maintainer

### DIFF
--- a/tests/modules/programs/tealdeer/default-settings.nix
+++ b/tests/modules/programs/tealdeer/default-settings.nix
@@ -1,0 +1,12 @@
+{ config, ... }: {
+  config = {
+    programs.tealdeer = {
+      package = config.lib.test.mkStubPackage { name = "tldr"; };
+      enable = true;
+    };
+
+    nmt.script = ''
+      assertFileRegex activate '/nix/store/.*tealdeer.*/bin/tldr --update'
+    '';
+  };
+}
--- a/modules/programs/tealdeer.nix
+++ b/modules/programs/tealdeer.nix
@@ -49,5 +49,10 @@ in {
     home.file."${configDir}/tealdeer/config.toml" = mkIf (cfg.settings != { }) {
       source = tomlFormat.generate "tealdeer-config" cfg.settings;
     };
+
+    home.activation.tealdeerCache = hm.dag.entryAfter [ "linkGeneration" ] ''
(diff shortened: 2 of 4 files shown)


tests/modules/programs/tealdeer/default-settings.nix
```diff
@@ -0,0 +1,19 @@
...
+      enable = true;
+    };
+
+    home.file."tealdeer-test.txt".text =
+      hmActivationDagResult config.home.activation;
+
+    nmt.script = ''
+      assertFileExists home-files/tealdeer-test.txt
+
+      assertFileRegex "home-files/tealdeer-test.txt" '/nix/store/.*tealdeer.*/bin/tldr --update'
```
pedorich-n, line 16: Not sure if there's a better way to test that activation script is there.

rycee: I think it should be sufficient to do

```suggestion
      assertFileRegex "activate" '/nix/store/.*tealdeer.*/bin/tldr --update'
```

pedorich-n: Thanks! That's much simpler. 
But what magic is this? Where this `activate` is coming from? 😲

pedorich-n: Is this the [main activation script](https://github.com/pedorich-n/home-manager/blob/8a39882f926e68840bf1131f4b6740d1b0ab4a89/modules/home-environment.nix#L713), but in tests `$out` is `/`?

rycee: Yes, paths in the tests are relative the built activation package. So, basically the output of `home-manager build`, You'll notice that there is an `activate` file in there, as well as a link `home-files`, that typically is followed in tests.

rycee:
Looks good. Merged to master now 🙂

Issue #4916, opened 2024-01-17
bug: Nushell v0.89.0 with zoxide integration give a warning


Ligthiago:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Nushell v.0.89.0 with zoxide integration gives deprecation warning every time `z` command is used with some argument.

![warning](https://github.com/nix-community/home-manager/assets/142721811/b0a6d52c-0af0-4c58-85bc-18eeba943818)

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.7.0, NixOS, 24.05 (Uakari), 24.05.20240115.c3e128f`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.19.2`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`
```

PR #4940, opened 2024-01-22
zoxide: fix nushell 0.89 deprectation

jpteb:
Hi, this is my first pull request to home-manager, so I would appreciate if someone could double-check that my fix does not break backwards compatibility (or anything else). I don't think it does, as the current version of home-manager has the broken versions. Please let me know if I'm wrong here.

### Description

Since nushell 0.89, automatically spreading lists is deprecated. This commit introduces a string replace for the zoxide init script to replace the deprecated code.
See: https://github.com/ajeetdsouza/zoxide/issues/662

Fixes: #4916

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


rycee:
LGTM. Merged to master now 🙂

PR #4946, opened 2024-01-24
core: add extendModules attribute to homeManagerConfiguration result

### LINKED ISSUE #4943: ExtendModules functionality matching NixOS
### Description
Using private nixos modules together with public nixos configurations is quite easy ([see this post](https://determinate.systems/posts/extending-nixos-configurations)):
```nix
nixosConfigurations.teapot =
  original.nixosConfigurations.teapot.extendModules {
    modules = [
      private-extension.nixosModules.default
    ];
  };
```

However, the result of `hm.lib.homeManagerConfiguration` does not provide the function `extendModules`.
I use home-manager in standalone mode where I have separate modules that are not shared on my public github.

Is there a current workaround to achieve this "injecting" of private modules into public home manager configurations?
### Implementation
After looking at the way nixpkgs does it, I suspect that the collection of warnings and assertions in `modules/default.nix` would need to be pushed down into the top-level module that defines the `activationPackage`. A home manager configuration would then return the configuration it returns currently *with* an additional attribute `extendModules` similar to [nixpkgs here](https://github.com/NixOS/nixpkgs/blob/5b91a0cca2543f429f1b8cb3ff704228f95ad929/nixos/lib/eval-config.nix#L117).

I would certainly be willing to help implement this feature if necessary!

GrizzlT:
### Description
home-manager.lib.homeManagerConfiguration now has an additional attribute that can be used to extend a home manager configuration with additional modules outside the project tree.

It works similar to the result of lib.nixosSystem from nixpkgs. This PR closes #4943 .

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@rycee

### DIFF
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -28,28 +28,34 @@ let
     specialArgs = { modulesPath = builtins.toString ./.; } // extraSpecialArgs;
   };
 
-  module = showWarnings (let
-    failed = collectFailed rawModule.config;
-    failedStr = concatStringsSep "\n" (map (x: "- ${x}") failed);
-  in if failed == [ ] then
-    rawModule
-  else
-    throw ''
+  moduleChecks = raw:
+    showWarnings (let
+      failed = collectFailed raw.config;
+      failedStr = concatStringsSep "\n" (map (x: "- ${x}") failed);
+    in if failed == [ ] then
+      raw
+    else
+      throw ''
 
-      Failed assertions:
-      ${failedStr}'');
+        Failed assertions:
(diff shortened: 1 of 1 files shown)


rycee:
Thanks! Looks quite useful. Merged to master now 🙂

GrizzlT:
@rycee Thanks for the quick merge! Is there a way this could be backported into the current stable release (23.11)?
Or should I use my fork until nixos 24.05 comes out instead?

PR #4962, opened 2024-01-27
arrpc: add module

NotAShelf:
### Description

Adds a module for the [arRPC](https://arrpc.openasar.dev/) project. Adaptation of my previously standalone [arrpc-flake](https://github.com/NotAShelf/arrpc-flake). Adds a user service bound to the graphical service (as arRPC is useless without it) that can be used for monitoring custom Discord RPC events.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/arrpc.nix
+++ b/modules/services/arrpc.nix
@@ -1,0 +1,39 @@
+{ config, pkgs, lib, ... }:
+let
+  inherit (lib) mkIf mkOption mkPackageOption mkEnableOption types;
+
+  cfg = config.services.arrpc;
+in {
+  meta.maintainers = [ lib.maintainers.NotAShelf ];
+
+  options.services.arrpc = {
+    enable = mkEnableOption "arrpc";
+    package = mkPackageOption pkgs "arrpc" { };
+
+    systemdTarget = mkOption {
+      type = types.str;
+      default = "graphical-session.target";
+      example = "sway-session.target";
+      description = ''
+        Systemd target to bind to.
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable {
+    systemd.user.services.arRPC = {
+      Unit = {
+        Description =
+          "Discord Rich Presence for browsers, and some custom clients";
+        PartOf = [ "graphical-session.target" ];
+      };
+
+      Service = {
+        ExecStart = lib.getExe cfg.package;
+        Restart = "always";
+      };
+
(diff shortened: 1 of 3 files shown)


fufexan (approved):

modules/services/arrpc.nix
```diff
@@ -0,0 +1,31 @@
+{ config, pkgs, lib, ... }:
+let
+  inherit (lib) mkIf mkPackageOption mkEnableOption;
```
Lillecarl, line 3: I'd stick to "with lib;" many HM modules use this

modules/services/arrpc.nix
```diff
@@ -0,0 +1,31 @@
...
+    enable = mkEnableOption "arrpc";
+    package = mkPackageOption pkgs "arrpc" { };
+  };
+
+  config = mkIf cfg.enable {
+    systemd.user.services.arRPC = {
+      Unit = {
+        Description =
+          "Discord Rich Presence for browsers, and some custom clients";
+        After = [ "graphical-session.target" ];
```
Lillecarl, line 19: It's always nice to make the systemd target configurable (see swayidle & and friends for inspiration).

Lillecarl (changes requested):
Generally looking good :)

modules/services/arrpc.nix
NotAShelf: I believe `with lib; is discouraged practice - and there's no downside to inheriting functions from the lib.

modules/services/arrpc.nix
NotAShelf: Good idea, I'll take a look at that.

modules/services/arrpc.nix
```diff
@@ -0,0 +1,31 @@
...
+        PartOf = [ "graphical-session.target" ];
+      };
+
+      Install.WantedBy = [ "graphical-session.target" ];
+
+      Service = {
+        ExecStart = "${lib.getExe cfg.package}";
```
itslychee, line 26: couldn't you write this without the string interpolation?

itslychee (approved):

modules/services/arrpc.nix
Lillecarl: All good, just comparing with other modules I've seen :)

rycee:
Thanks! Rebased to master in 88aca6dc9934b242608e3967043302e097cbd5cf.

PR #4968, opened 2024-01-28
tealdeer: add option to toggle update on activation

Lillecarl:
### Description

Add option to disable updating tealdeer cache on activation, since it's a network operation it can be quite annoying on shitty connections like trains and such.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/tealdeer.nix
+++ b/modules/programs/tealdeer.nix
@@ -17,6 +17,14 @@ in {
   options.programs.tealdeer = {
     enable = mkEnableOption "Tealdeer";
 
+    updateOnActivation = mkOption {
+      type = with types; bool;
+      default = true;
+      description = ''
+        Whether to update tealdeer's cache on activation.
+      '';
+    };
+
     settings = mkOption {
       type = tomlFormat.type;
       default = { };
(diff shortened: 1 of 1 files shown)


pedorich-n:
Ah, sorry you had to experience that! I added the activation feature 😓 
Initially, I also wrote code to check the `config.updates.auto_update` flag to disable the activation script if it's set but I thought it looked too complex and removed it. 
I should've added some kind of toggle from the beginning.

pedorich-n (approved):

Lillecarl:
@pedorich-n No worries, I just set the activation script to ```echo Disabled``` and kept on working 😄

I figured I still want this option though as it seems the module system doesn't give an easy option to unset an option, and not every HM user knows how to mkForce things :)

Also, would you like to be in the maintainers list for this module?

rycee:
Thanks! I merged it now.

I didn't actually realize this did network stuff. I would then recommend not including it at all in the activation script. The activation script should be fast and idempotent, especially since we may move towards running the activation script on login.

Instead perhaps this update functionality could be done in a systemd unit?

pedorich-n:
I don't think it's worth doing as systemd, since `tealdeer` already has the periodic updates functionality built-in: https://dbrgn.github.io/tealdeer/config_updates.html#automatic-updates. It's just not enabled by default.

Should we remove the activation part? Basically rollback two last commits for this module.

rycee:
@pedorich-n Yeah, perhaps that is a good idea. An alternative to make that option more prominent would be to make the settings option be a [freeform module](https://nixos.org/manual/nixos/stable/#sec-freeform-modules). See e.g. [`gh.nix`](https://github.com/nix-community/home-manager/blob/afcedcf2c8e424d0465e823cf833eb3adebe1db7/modules/programs/gh.nix#L11-L45).

PR #4970, opened 2024-01-28
borgmatic: add option for pattern matching

Liassica:
### Description

Add option option for pattern matching in Borgmatic backups. It is mutually exclusive with the existing `sourceDirectories` option, so assertions have been added to make sure that both aren't set at once (but also that at least one of them is).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@DamienCassou

### DIFF
--- a/modules/programs/borgmatic.nix
+++ b/modules/programs/borgmatic.nix
@@ -76,10 +76,37 @@ let
       (mkAfter [ (toString hmExcludeFile) ]);
     options = {
       location = {
-        sourceDirectories = mkOption {
+        sourceDirectories = mkNullableOption {
           type = types.listOf types.str;
-          description = "Directories to backup.";
+          default = null;
+          description = ''
+            Directories to backup.
+
+            Mutually exclusive with [](#opt-programs.borgmatic.backups._name_.location.patterns).
+          '';
           example = literalExpression "[config.home.homeDirectory]";
+        };
+
+        patterns = mkNullableOption {
+          type = types.listOf types.str;
+          default = null;
+          description = ''
+            Patterns to include/exclude.
+
+            See the output of `borg help patterns` for the syntax. Pattern paths
+            are relative to `/` even when a different recursion root is set.
+
+            Mutually exclusive with [](#opt-programs.borgmatic.backups._name_.location.sourceDirectories).
+          '';
+          example = literalExpression ''
+            [
+              "R /home/user"
+              "- home/user/.cache"
+              "- home/user/Downloads"
+              "+ home/user/Videos/Important Video"
+              "- home/user/Videos"
+            ]
+          '';
         };
 
         repositories = mkOption {
@@ -194,6 +221,7 @@ let
   writeConfig = config:
     generators.toYAML { } (removeNullValues ({
       source_directories = config.location.sourceDirectories;
+      patterns = config.location.patterns;
       repositories = config.location.repositories;
       encryption_passcommand = config.storage.encryptionPasscommand;
       keep_within = config.retention.keepWithin;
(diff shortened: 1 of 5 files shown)


modules/programs/borgmatic.nix
```diff
@@ -193,7 +218,11 @@ let
 
   writeConfig = config:
     generators.toYAML { } (removeNullValues ({
-      source_directories = config.location.sourceDirectories;
+      source_directories = if (config.location.patterns == null) then
```
DamienCassou, line 221: should we add an assertion to make sure that the 2 options are not used together?

DamienCassou (changes requested):
Thank you for the great work. Should we add a few test cases?

- one test case to make sure that specifying both `sourceDirectories` and `patterns` triggers an error
- one test case to check that patterns are written to the generated configuration file.

modules/programs/borgmatic.nix
Liassica: I spent some time earlier today attempting to do this, but I'm afraid I'm still rather new to Nix, so trying to define an assertion related to submodule options was giving me some trouble. I'd appreciate any pointers or suggestions you could give regarding that. In the meantime I've converted this pull request to a draft.

Liassica: Never mind, I was able to figure it out by using [autorandr](https://github.com/nix-community/home-manager/blob/master/modules/programs/autorandr.nix#L346) as an example.

Liassica:
I added the tests and assertion you suggested, so I think this is ready for review now.

DamienCassou:
I wrote the assertion and a test in: https://github.com/DamienCassou/home-manager/commit/f7b779d32992835a0ff55a87b9a139f1e9f800d2. What is missing now:

- an assertion that both sourceDirectories and patterns are not null at the same time
- a test to check that patterns are correctly exported to the configuration file

Can you try to implement that?

Liassica:
I believe that latest commit should do it. I already added the pattern configuration export test earlier unless you'd like something to be added to it.

Liassica:
I've also just noticed a slight error in the example for this option that I'll fix tomorrow. In the meantime, I've converted this pull back into a draft. I'll squash the commits before it gets pulled.

DamienCassou:
Please tell me when you are ready. Thanks for your work.

Liassica:
I just squashed the commits, so I think it's ready now.

tests/modules/programs/borgmatic/neither-sourcedirectories-nor-patterns.nix
```diff
@@ -0,0 +1,26 @@
...
+  backups = config.programs.borgmatic.backups;
+
+in {
+  programs.borgmatic = {
+    enable = true;
+    backups = {
+      main = {
+        location = {
+          sourceDirectories = null;
+          patterns = null;
```
DamienCassou, line 14: nit: From a user point of view, I think it is more likely that none of these 2 options are specified rather both being explicitly set to `null`. As a result, I would delete these 2 lines.

tests/modules/programs/borgmatic/patterns-configuration.nix
```diff
@@ -0,0 +1,147 @@
...
+      main = {
+        location = {
+          patterns = [
+            "R my-stuff-to-backup"
+            "- *.swp"
+            "+ my-stuff-to-backup/junk/important_file"
+            "- my-stuff-to-backup/junk"
+          ];
+          repositories = [
+            "/mnt/disk1"
```
DamienCassou, line 21: Most of these configuration options are already tested by basic-configuration.nix. To reduce maintenance effort, I would only test location.patterns here and remove all other options and checks.

DamienCassou (approved):
Excellent job, thank you. I left 2 small comments but I'm approving the PR anyway.

Liassica:
Good points on both of those, made the suggested changes. With any luck these'll be the last things!

DamienCassou (approved):
Excellent job, thank you very much.

Liassica:
I noticed a typo in one of the assertions and I'm working on fixing it, however internet issues are preventing me from running the tests right now. I'll have it fixed as soon as I can.

rycee:

Issue #4566, opened 2023-10-15
bug: file.old would be clobbered by backing up file


viperML:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

The automatic backup setting doesn't handle having a backup already in place.

Steps to reproduce:

1. Have some `file` be backed up by HM into `file.old`
2. Replace `file` again with something else.
3. HM won't be able to backup `file` into `file.old`.

This behaviour is located here: https://github.com/nix-community/home-manager/blob/d4a5076ea8c2c063c45e0165f9f75f69ef583e20/modules/files.nix#L116-L119

My suggestion: add a new option, that will force the backup, removing the old backup.

### System information

```markdown
$ nix-shell -p nix-info --run "nix-info -m"
 - system: `"x86_64-linux"`
 - host os: `Linux 6.5.5, NixOS, 23.11 (Tapir), 23.11.20231006.87828a0`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.17.0`
 - nixpkgs: `/nix/store/g68f5abh3xhcz8xsdlfw7wkgkkcx3nwy-source`
```
```

**Comment by viperML** [comment]
CC the git blame: @nyarly

**Comment by nyarly** [comment]
The backup flag takes an option for what to backup to. I generally just change it to e.g. `old2`

**Comment by viperML** [comment]
I am using it through the NixOS module and the `backupFileExtension` option. So changing it every now and then is not an option.

**Comment by viperML** [comment]
Still an issue, everytime I boot NixOS the HM systemd service fails with this.

**Comment by snoopdouglas** [comment]
Shame this issue is stale, I've just been tripped up by it:

```
warning: the following units failed: home-manager-doug.service
× home-manager-doug.service - Home Manager environment for doug
     Loaded: loaded (/etc/systemd/system/home-manager-doug.service; enabled; preset: ignored)
     Active: failed (Result: exit-code) since Fri 2025-08-15 18:50:02 BST; 261ms ago
 Invocation: cfe19ca418c54d91b96095ad655f314d
    Process: 35292 ExecStart=/nix/store/xrz52c09p9ba8hdlcbrl9fnfg87f0w6z-hm-setup-env /nix/store/x2g3nagidbc9kbmv3yfby4yg56zdwvjd-home-manager-generation (code=exited, status=1/FAILURE)
[...]

Issue #4199, opened 2023-07-04
Add option to replace existing files


justinlovinger:
Having Home Manager fail if an existing file exists is a big foot-gun. As a veteran user, I know what it means when the Home Manager service fails. I know to delete the offending file and reactive the service.

However, I have seen new users struggle with broken configurations, unsure why their Home Manager changes are not realized. Worse, this allows NixOS to update with Home Manager, potentially breaking systems. Worse still, if an existing file exists, users cannot cleanly roll back to previous NixOS generations because Home Manager will fail to realize those previous generations. In the worst case, this can result in an unusable machine with no way to roll back, requiring recovery from a bootable USB.

P.S. This is not a one-off issue. Some applications will override files managed by Home Manager, resulting in this error cropping up repeatedly, even if `backupFileExtension` is set.

ncfavier:
You can set `home.file."foo/bar".force = true;` to replace existing files. Maybe this option should be made visible in the manual.

JustinLovinger:
`home.file."foo/bar".force = true;` solves the issue for individual files, but not for Home Manager as a whole. An option on the Home Manager module, alongside where `backupFileExtension` is defined, could solve it more generally.

I also suggest mentioning this option in error messages about existing files. Once a user is informed Home Manager can replace their files, they can back up whatever they want, enable the option, and never have to deal with it again.

ToasterUwU:
I ran into this issue as well now. Specifically with the .gtkrc-2.0 file, so "force" wasn't even an option. I used the home manager gtk option to set a theme and then kept getting that error over and over telling me to enable something I already have. 
All I had to do was delete the .gtkrc-2.0.backup file that already existed and was in the way, but this should not be necessary in the first place.

ToasterUwU:
I global option for either not making a backup of a file entirely or even better, an option to always overwrite a backup file with the new one if needed would solve this issue nicely, but currently im stuck in an endless loop of having to manually delete a single file, which really defeats the purpose of this whole system i made myself.

`home-manager.overridebackupFiles` might be a good name for this.

ToasterUwU:
Seems like there was already an issue and a PR about this. But the person who made the fix for this couldn't get the approval of the maintainer.

Issue: #4566 
PR: #4971

OsiPog:
I'm having the problem decribed above with stylix. I have to change `home-manager.backupFileExtension` every time I reboot otherwise starting the home manager service fails.

Does anyone have a fix for this?

[...]

PR #4971, opened 2024-01-29
home-manager: add backup overwrite option

viperML:
### Description

When using --backup or backupFileExtension, if the backup exists, the activation process fails with:

https://github.com/nix-community/home-manager/blob/4d54c29bce71f8c261513e0662cc573d30f3e33e/modules/files.nix#L116-L119

This new options lets you remove the old backup automatically.

TODO:

- [x] Check if it works properly on NixOS
- [ ] Check if it works properly on home-manager standalone
- [ ] Check if it works properly on nix-darwin
- [ ] Add CLI flag

Closes: #4566

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/files.nix
+++ b/modules/files.nix
@@ -113,9 +113,11 @@ in
               elif [[ ! -L "$targetPath" && -n "$HOME_MANAGER_BACKUP_EXT" ]] ; then
                 # Next, try to move the file to a backup location if configured and possible
                 backup="$targetPath.$HOME_MANAGER_BACKUP_EXT"
-                if [[ -e "$backup" ]]; then
-                  errorEcho "Existing file '$backup' would be clobbered by backing up '$targetPath'"
+                if [[ -e "$backup" && -z "$HOME_MANAGER_BACKUP_OVERWRITE" ]]; then
+                  errorEcho "Existing file '$backup' would be clobbered by backing up '$targetPath', use HOME_MANAGER_BACKUP_OVERWRITE=1 if you want to overwrite the backup"
                   collision=1
+                elif [[ -e "$backup" && -n "$HOME_MANAGER_BACKUP_OVERWRITE" ]]; then
+                  warnEcho "Existing file '$targetPath' is in the way of '$sourcePath', and existing backup '$backup' is on the way, old backup will be overwritten"
                 else
                   warnEcho "Existing file '$targetPath' is in the way of '$sourcePath', will be moved to '$backup'"
                 fi
@@ -179,6 +181,9 @@ in
             if [[ -e "$targetPath" && ! -L "$targetPath" && -n "$HOME_MANAGER_BACKUP_EXT" ]] ; then
               # The target exists, back it up
               backup="$targetPath.$HOME_MANAGER_BACKUP_EXT"
+              if [[ -e "$backup" && "$HOME_MANAGER_BACKUP_OVERWRITE" ]]; then
+                run rm $VERBOSE_ARG "$backup"
+              fi
               run mv $VERBOSE_ARG "$targetPath" "$backup" || errorEcho "Moving '$targetPath' failed!"
             fi
 
--- a/nix-darwin/default.nix
+++ b/nix-darwin/default.nix
@@ -21,6 +21,8 @@ in {
               "export HOME_MANAGER_BACKUP_EXT=${
                 lib.escapeShellArg cfg.backupFileExtension
               }"}
+              ${lib.optionalString (cfg.backupOverwrite)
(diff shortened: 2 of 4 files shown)


viperML:
This would be an opt-in option, so the user assumes the risk of losing information.

teto:
it's easy to make mistakes, especially if the flag is toggled via the environment (HOME_MANAGER_BACKUP_OVERWRITE) ? best to not make it possible IMO

viperML:
What do you propose instead of using an env variable for the activation script instead?

teto:
My suggestion is the same as in my previous post, dont make this a setting and dont remove any data (by overwriting backups). 
What I would suggest instead is to change the default behavior such that backups created by home-manager are suffixed with FILENAME.hm-backup-TIMESTAMP instead (you get the idea). You can use `date -u +%s` to get the timestamp.

viperML:
For my usecase, that would involve creating a new file for each activation, which doesn't sound that good. (Until I figure out which application is modifying the file)

But that gave me a different idea:

Maybe instead of HOME_MANAGER_BACKUP_OVERWRITE, use HOME_MANAGER_BACKUP_LIMIT. A limit = 1 would have the same effect as the original solution, but it could be configured to let it create backups without limit (maybe setting the env var to 0, but that could imply to wipe backups? ...)

teto:
> Maybe instead of HOME_MANAGER_BACKUP_OVERWRITE, use HOME_MANAGER_BACKUP_LIMIT. A limit = 1 would have the same effect as the original solution, but it could be configured to let it create backups without limit (maybe setting the env var to 0, but that could imply to wipe backups? ...)

It is a bit better but still I dont see the appeal for home-manager to remove data. It's a low impact and high-risk change. IMO it's best to just give a command to the user that they can run by themselves when they want to discard their backups.

> For my usecase, that would involve creating a new file for each activation, which doesn't sound that good. (Until I figure out which application is modifying the file)

Also one thing I dont understand is why you would have several backups for a same file ? usually it's done once when home-manager takes over ?!  One usecase I may have encountered is a backup being created while the home-manager switch failed, and the next run would fail because of an existing file. But this should be an exceptional case and just creating a 2nd backup with tiemstamp would solve that.

viperML:
> But this should be an exceptional case and just creating a 2nd backup with tiemstamp would solve that.

If an application modifies the file inbetween each activation. Too bad I am under this "exceptional" case

viperML:
Also: I don't have the willpower to get through your approval

teto:
> If an application modifies the file inbetween each activation.

a file handled by nix would usually reference the /nix/store, in principle unreadable.
If the file is generated and overwritten, it's an unusual hybrid mode so you need to explain yourself to convince maintainers to add code that potentially removes data.

viperML:
I think my alternatives are good enough, and I don't see the "maintainers" in pluar. Anyways, feel free to have the last word.

jezcope:
> a file handled by nix would usually reference the /nix/store, in principle unreadable.

See the case described at https://github.com/nix-community/home-manager/issues/4199#issuecomment-2156617636

justinlovinger:
> I dont see the appeal for home-manager to remove data.

I understand what you are trying to say, but consider, you may be in the minority here. This feature touches on <https://github.com/nix-community/home-manager/issues/4199>, which has a number of thumbs-ups and comments in support, and nobody has expressed concerns.

Personally, I would still rather have an option to disable backups entirely, but this will at least fix the most egregious issues.

teto:
I dont see how removing the file is necessary to fix the issue you linked. I suggested renaming the file which fixes it the same with the advantage of not taking the risk of losing all my data because once you introduce `rm` there will come a time where it deletes something it shouldn't. home-manager is understaffed, and even big projects like steam make the mistake.

A complimentary path to help those thumb up users is to improve the home-manager reporting: it displays only the last lines which usually dont contain the error. Maybe we could reprint the errors there or hide "info" lines instead, add desktop notifications etc.

nyabinary:
I still think this is a good option for some users tbh

Doc-Steve:
### alternative proposal for hookupScript in case of backup failures

I understand the concerns, but there are modules like disko and nixos-anywhere which wipe entire hard disks without any backup or "are-you-sure" toggles. Users are fine with that type of risk. Compare this to the backup-file case ...  Overwriting a previous backup file (and still have a current backup) is very, very low risk. If you don't want to implement the "rm" function yourself, how about a hookup-script for backup-failure, which can be user defined. There you can remove, move individually - no maintainer is responsible.

PR #4973, opened 2024-01-29
nix-gc: add automatic garbage collector for nix

shivaraj-bh:
### Description

Add `nix-gc` module that configures systemd unit for linux and launchd agent for macOS to automatically run the `nix-collect-garbage` executable at a specified `frequency`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/nix-gc.nix
+++ b/modules/services/nix-gc.nix
@@ -1,0 +1,120 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.nix.gc;
+
+  mkCalendarInterval = frequency:
+    let
+      freq = {
+        "hourly" = [{ Minute = 0; }];
+        "weekly" = [{
+          Weekday = 1;
+          Hour = 0;
+          Minute = 0;
+        }];
+        "monthly" = [{
+          Day = 1;
+          Hour = 0;
+          Minute = 0;
+        }];
+        "semiannually" = [
+          {
+            Month = 1;
+            Day = 1;
+            Hour = 0;
+            Minute = 0;
+          }
+          {
+            Month = 7;
+            Day = 1;
+            Hour = 0;
+            Minute = 0;
+          }
+        ];
+        "annually" = [{
+          Month = 1;
+          Day = 1;
+          Hour = 0;
+          Minute = 0;
+        }];
+      };
+    in freq.${frequency};
+
+  nixPackage = if config.nix.enable && config.nix.package != null then
+    config.nix.package
+  else
+    pkgs.nix;
+in {
+  meta.maintainers = [ maintainers.shivaraj-bh ];
+
+  options = {
+    nix.gc = {
+      automatic = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          Automatically run the garbage collector at a specific time.
+
+          Note: This will only garbage collect the current user's profiles.
+        '';
+      };
+
+      frequency = mkOption {
+        type =
+          types.enum [ "hourly" "weekly" "monthly" "semiannually" "annually" ];
+        default = "weekly";
+        example = "monthly";
+        description = ''
+          The frequency at which to run the garbage collector.
+
+          These enums are based on special expressions from the
+          {manpage}`systemd.time(7)`
+        '';
+      };
+
+      options = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        example = "--max-freed $((64 * 1024**3))";
+        description = ''
+          Options given to {file}`nix-collect-garbage` when the
+          garbage collector is run automatically.
+        '';
+      };
+    };
+  };
+
+  config = lib.mkIf cfg.automatic (mkMerge [
+    (mkIf pkgs.stdenv.isLinux {
+      systemd.user.services.nix-gc = {
+        Unit = { Description = "Nix Garbage Collector"; };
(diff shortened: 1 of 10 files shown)


modules/services/nix-gc.nix
```diff
@@ -0,0 +1,115 @@
...
+    pkgs.nix;
+in {
+  meta.maintainers = [ maintainers.shivaraj-bh ];
+
+  options = {
+    nix.gc = {
+      automatic = mkOption {
+        default = false;
+        type = types.bool;
+        description = lib.mdDoc
```
rycee, line 56: The `lib.mdDoc` function is unnecessary. The description is interpreted as Markdown by default.

modules/services/nix-gc.nix
```diff
@@ -0,0 +1,114 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.nix.gc;
+
+  mkCalendarInterval = frequency:
```
shivaraj-bh, line 8: @rycee I have taken the liberty to create a function that provides a mapping from systemd time-frequency enums to that of launchd’s. is it a good idea to also expose an option to let override the frequency, if they don’t want to use the default mapping? How would you think that should be? separate option for both macOS and linux frequency?

rycee: My spontaneous opinion would be to have a single duration type that can be mapped into a systemd or launchd interval. I've made such years ago. I've looked at this a few times and eventually I think it would be good to introduce. See e.g. https://github.com/NixOS/nixpkgs/pull/43852 for my, perhaps, most elaborate take on this.

I think I have some half-baked simplified version of that as well that probably could be resurrected.

In any case, perhaps it would be sufficient to have what you have now and revisit in the future?

modules/services/nix-gc.nix
```diff
@@ -0,0 +1,114 @@
...
+        ];
+        "annually" = [{
+          Month = 1;
+          Day = 1;
+          Hour = 0;
+          Minute = 0;
+        }];
+      };
+    in freq.${frequency};
+  nix-package = if config.nix.enable && config.nix.package != null then
```
rycee, line 44: Variables should be camel cased. Also add an empty line to make it clear where one variable ends and the next begins.

```suggestion

  nixPackage = if config.nix.enable && config.nix.package != null then
```

modules/services/nix-gc.nix
```diff
@@ -0,0 +1,114 @@
...
+        type =
+          types.enum [ "hourly" "weekly" "monthly" "semiannually" "annually" ];
+        description = ''
+          The frequency at which to run the garbage collector.
+
+          These enums are based on special expressions from the
+          {manpage}`systemd.time(7)`
+        '';
+        example = "monthly";
+        default = "weekly";
```
rycee, line 69: We typically use the field order

1. type
2. default
3. example
4. description

```suggestion
        default = "weekly";
        example = "monthly";
        description = ''
          The frequency at which to run the garbage collector.

          These enums are based on special expressions from the
          {manpage}`systemd.time(7)`
        '';
```

modules/services/nix-gc.nix
```diff
@@ -0,0 +1,114 @@
...
+      };
+
+      options = mkOption {
+        default = null;
+        example = "--max-freed $((64 * 1024**3))";
+        type = types.nullOr types.str;
```
rycee, line 75:
```suggestion
        type = types.nullOr types.str;
        default = null;
        example = "--max-freed $((64 * 1024**3))";
```

modules/services/nix-gc.nix
```diff
@@ -0,0 +1,114 @@
...
+        description = ''
+          Options given to {file}`nix-collect-garbage` when the
+          garbage collector is run automatically.
+        '';
+      };
+    };
+  };
+
+  config = lib.mkIf (config.nix.enable && cfg.automatic) (mkMerge [
```
rycee, line 84: Since `nixPackage` is regardless of the value of `config.nix.enable`:

```suggestion
  config = lib.mkIf cfg.automatic (mkMerge [
```

rycee:
Thanks for the contribution! I've added a few comments. I think it would be good to also add a test case each for Linux and Darwin. See, e.g., https://github.com/nix-community/home-manager/tree/master/tests/modules/services/git-sync and https://github.com/nix-community/home-manager/tree/master/tests/modules/services/git-sync-darwin.

modules/services/nix-gc.nix
```diff
@@ -0,0 +1,114 @@
...
+    pkgs.nix;
+in {
+  meta.maintainers = [ maintainers.shivaraj-bh ];
+
+  options = {
+    nix.gc = {
+      automatic = mkOption {
+        default = false;
+        type = types.bool;
+        description = "Automatically run the garbage collector at a specific time.";
```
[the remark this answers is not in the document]
shivaraj-bh: Yes, I can provide an example to add extra options to gc home-manager generations.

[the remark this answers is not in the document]
shivaraj-bh: Ignore the part about example to add extra options. I will update the description to specify that it only gc’s user specific profiles.

shivaraj-bh:
@rycee Thanks for the inputs! I have addressed all the points and also added the tests.

rycee:

PR #4976, opened 2024-01-29
WIP home-manager: avoid profile management during activation

rycee:
### Description

This deprecates profile management from the activation script. The profile management is instead the responsibility of the driving software, for example, the `home-manager` tool in the case of standalone installs.

The legacy behavior is still available for backwards compatibility but may be removed in the future.

The new behavior resolves (or moves us closer to resolving) a number of long standing open issues:

- `home-manager switch --rollback`, which performs a rollback to the previous Home Manager generation before activating. While it was previously possible to accomplish this by activating an old generation, it did always create a new profile generation.

  This option has been implemented as part of this commit.

- `home-manager switch --test`, which activates the configuration but does not create a new profile generation.

  This option has _not_ been implemented here since it relies on the current configuration being activated on login, which we do not currently do.

- When using the "Home Manager as a NixOS module" installation method we previously created an odd `home-manager` per-user "shadow profile" for the user. This is no longer necessary.

  This has been implemented as part of this commit.

### Checklist

- [ ] Change is backwards compatible: Some care has been put into making this backwards compatible but more testing would be good. Also, with this the NixOS and nix-darwin modules will no longer produce a shadow profile. If you rely on it existing, then this would break your use-case.

    ```
    {component}: {description}

    {long description}
    ```


oxalica:
Is there still some unsolved issues, or is it waiting for more stability testing? I'm waiting for this to land for a long time. Currently we h-m NixOS module users need to manually `rm ~/.local/state/nix/profiles/home-manager-*` periodically or it can easily stack to thousands of profiles.

nyabinary:
What's left for this to be completed?

rycee:
Nothing really blocking, I've been using this without any issue for some time now and was mostly waiting for the 25.05 release to settle. I've removed the WIP marker.

It would be great to have some additional users of this before merging and some extra eyes on the code.

rycee:
BTW. The one concern I have is that I've seen that some people rely on the current "compatibility" between running as a NixOS module and standalone. This would no longer work and those users would have to explicitly switch to the legacy mode.

Rosuavio:
> ...the current "compatibility" between running as a NixOS module and standalone.

I dont understand what you mean, can you explain? Who needs legacy mode?

isabelroses:
> It would be great to have some additional users of this before merging and some extra eyes on the code.

Have been running this for about ~2 weeks, and intermittently for ~2 months and I have yet to face any issues. Across nix-darwin and nixos. However, I have no experience with standalone home-manager.

rycee:
@Rosuavio Basically, if you currently use the NixOS module you can switch to the standalone installation and vice versa without losing track of your Home Manager generations since both ways of operation will keep these inside `~/.local/state`. Indeed, there are probably people that use both simultaneously. I.e., apply a HM configuration via the NixOS module but also have the standalone `home-manager` tool installed to work on their home configuration "in-between system updates".

With the new mode of operation these will not work very well. The above has never been an explicitly supported way to use HM and indeed I have discouraged such use whenever possible but I expect some complaints once this gets merged.

r-vdp:
I'll give this a try then, I've been eyeing this PR for a while now, thanks a lot for all the work that went in to this!

tests/integration/nixos/basics.nix
```diff
@@ -82,17 +82,13 @@
...
-      assert gcrootTarget == profile1Target, \
-        f"expected GC root and profile to point to same, but pointed to {gcrootTarget} and {profile1Target}"
+    with subtest("no GC root and profile"):
+      # There should be no GC root and Home Manager profile since we are not
+      # using legacy profile management.
+      hmState = "/home/alice/.local/state/home-manager"
+      machine.succeed(f"test ! -e {hmState}")
+
+      hmProfile = "/home/alice/.local/state/nix/profiles/home-manager"
+      machine.succeed(f"test ! -e {hmProfile}")
```
khaneliman, line 92: Is there something we should test the presence of? Just thinking from a testing validity standpoint, just seeing that nothing is there doesn't mean it successfully deployed.

khaneliman (commented):
LGTM overall and just did a basic test with my main system and everything seemed fine. Forgot to test with my macbook but can tomorrow if i remember.

r-vdp:
I just included this PR in my HM, rebuild my config, and rebooted. Then I cleaned up everything hm-related in `~/.local/state` and removed the profiles `~/.local/state/nix/profiles`.
After reactivating my nixos config, I then noticed that indeed no profile is being created, but there is a link generated in `~/.local/state/home-manager/gcroots` called `current-home`.
Is this expected? I would expect that we don't need a GC root anymore sinces the nixos config will keep the integrated HM profile alive. Also, if I read the tests correctly, they assert that no such link exists.

r-vdp:
Hmm, upon removing those directories again and reactivating again, now they didn't come back. So maybe there was still some old HM code involved somehow.

tests/integration/nixos/basics.nix
rycee: I think the subtests above this one would ensure that the activation happened successfully. These just verify that the activation didn't produce the profile stuff.

rycee:
I figure it is time to merge this before it languishes even more. I'll try to be as available as possible for any issues that crop up. This is a pretty fundamental change and it is difficult to cover all possible scenarios in advance, so I hope people are understanding if some flaw is discovered!

nyabinary:
How does one switch to the new activation script because as mentioned in the doc in the internal it still defaults to the old script for backwards-comp.

SEIAROTg:
It looks like config file deletion is no longer happening after this if using NixOS module.

If I change `xdg.configFile.foo.text = "foo"` into `xdg.configFile.bar.text = "bar"` and trigger a switch at NixOS level, `~/.config/foo` still hangs around while `~/.config/bar` is created, which feels a bug.

Using home-manager at the preceding commit or setting `home-manager.enableLegacyProfileManagement = true` seems to work fine.

<details>
<summary>NixOS test that reproduces the problem</summary>

```nix
{
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";

    # bad
    home-manager.url = "github:nix-community/home-manager/de448dcb577570f2a11f243299b6536537e05bbe";
    # good
    # home-manager.url = "github:nix-community/home-manager/e4bf85da687027cfc4a8853ca11b6b86ce41d732";
    home-manager.inputs.nixpkgs.follows = "nixpkgs";
  };

  outputs =
    { nixpkgs, home-manager, ... }: let
      system = "x86_64-linux";
    in {
      checks."${system}" = let
        pkgs = import nixpkgs { inherit system; };
      in {
        foo = pkgs.testers.runNixOSTest ({ ... }: {
          name = "foo";
          testScript = { nodes, ... }: ''
            machine.wait_for_unit("default.target")
            machine.wait_for_unit("default.target", user="alice")

            machine.succeed("stat /home/alice/.config")
            machine.succeed("stat /home/alice/.config/foo")

            machine.succeed("${nodes.machine.system.build.toplevel}/specialisation/bar/bin/switch-to-configuration switch")

            machine.succeed("stat /home/alice/.config/bar")
            machine.fail("stat /home/alice/.config/foo")
          '';

          nodes.machine = { lib, ... }: {
            imports = [
              home-manager.nixosModules.home-manager
            ];

            users.users.alice = {
              group = "alice";
              isNormalUser = true;
              linger = true;
            };
            users.groups.alice = {};

            home-manager.users.alice = lib.mkDefault ({ config, ... }: {
              home.stateVersion = config.home.version.release;
              xdg.configFile.foo.text = "content of foo";
            });

            specialisation.bar = {
              configuration = {
                home-manager.users.alice = ({ config, ... }: {
                  home.stateVersion = config.home.version.release;
                  xdg.configFile.bar.text = "content of bar";
                });
              };
            };
          };
        });
      };
    };
}
```

</details>

netadr:
There is an issue I discovered with this change where `sd-switch` is not passed an `old_dir` argument which results in all systemd services being stopped then started unconditionally on activation. `sd-switch` is invoked properly when `home-manager.enableLegacyProfileManagement = true;`, however. Let me know if I should break this out into a separate issue!

Stinjul:
I can confirm the same sd-switch behavior and workaround to fix it, it's especially annoying if you're using something like [impermanence](https://github.com/nix-community/impermanence) as it tries to remount everything 😬

rycee:
Sorry about the troubles! Could you try https://github.com/nix-community/home-manager/pull/7588 and see if it resolves the problems?

SEIAROTg:
> Sorry about the troubles! Could you try #7588 and see if it resolves the problems?

Works for me, and the above test now passes. Thanks!

netadr:
> Sorry about the troubles! Could you try #7588 and see if it resolves the problems?

Sorry for the late reply -- I can confirm that PR fixed my issue :)

thattemperature:
I use NixOS and I used to run "nix-collect-garbage --delete-old" as different users to delete their own old home-manager profiles. After this PR I found that I only need to run the command as root and it can deal with every user's home-manager, which is a nice improvement. 

But it seems that it will leave the newest profile as gcroot which is useless, and I do not know how to remove it elegantly. So I delete it from the disk manually and everything looks okay. Is my operation right?

(In fact, I did not find this problem before I found some old packages in `/nix/store` had not be removed, and I think many users have not find this problem yet)

r-vdp:
@thattemperature isn't that what the PR referenced in the comment just above yours introduced? You can check the discussion above to see why this is needed (basically, IIUC, when switching HM needs to have a reference to the previously active profile).

PR #4981, opened 2024-01-31
hyprland: fix hyprctl crash

jur0:
### Description

`/tmp/hypr` must be present so `hyprctl` doesn't crash.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/window-managers/hyprland.nix
+++ b/modules/services/window-managers/hyprland.nix
@@ -246,11 +246,13 @@ in {
 
       onChange = lib.mkIf (cfg.package != null) ''
         ( # Execute in subshell so we don't poision environment with vars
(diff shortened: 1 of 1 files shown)


Lillecarl:
LGTM 😄 

How do you end up in a state where it doesn't exist? When enabling Hyprland for the first time only I'm guessing, since Hyprland creates the directory when it's ran.

rycee:
Thanks! Merged to master now.

jur0:
> LGTM 😄
> 
> How do you end up in a state where it doesn't exist? When enabling Hyprland for the first time only I'm guessing, since Hyprland creates the directory when it's ran.

It also crashes with every boot when `/tmp` is wiped. The code is called as part of `hm-activate-${user}`.

PR #4982, opened 2024-01-31
Added vifm module

willbou1:
### Description

I added a module for vifm with support for displaying images in the terminal. There was another PR about vifm, but I felt like mine had way more functionalities.

### Checklist

- [x] Commit messages are formatted well


rycee:
Thanks for the contribution! I see that this PR adds quite useful features over the #3214 one. I'm a bit concerned that it introduces custom tooling, though, in particular the vifmimg program. Is this a standard script to have available for vifm? If so, perhaps it is distributed somewhere so it could be included as a Nixpkgs package? Then it would also be useful for people that don't use Home Manager.

For now I merged #3214 as a minimal base for future development so if you would like to introduce vifmimg tooling then it could be done as a separate feature PR. Will therefore close this PR. Apologies for the confusion.

Issue #4972, opened 2024-01-29
Control user services on non-NixOS


khlopkov:
### Description

I use Home Manager on a non-NixOS (Debian in my case). Home Manager installed few systemd services with a packages into `~/.nix-profile/share/systemd/user/` from nixpkgs. Those services can not be found, for example this command:

```bash
systemctl --user restart xdg-desktop-portal-gtk
```
gives the following output:
```
Failed to restart xdg-desktop-portal-gtk.service: Unit xdg-desktop-portal-gtk.service not found.
```

I was able to fix this by adding a symlink from the ``~/.nix-profile/share/systemd`  to the `~/.config/systemd`:
```bash
ln -s ~/.nix-profile/share/systemd/ ~/.config/systemd
```

But there are few issues with this workaround:
1. I can not enable those services from my Home Manager configuration and have to enable them manually.
2. It conflicts with the [xdg.portal.enable](https://nix-community.github.io/home-manager/options.xhtml#opt-xdg.portal.enable) (and I suppose with other services that can be enabled through the Home Manager). It prints following when trying to set this option to true:
```
Existing file '/home/etsilence/.config/systemd/user/xdg-desktop-portal.service' is in the way of '/nix/store/d3yw2n212zbz7473pbwwc41dj1ipkk8x-home-manager-files/.config/systemd/user/xdg-desktop-portal.service'
```

It can be fixed by linking specific services into :
```bash
ln -s ~/.nix-profile/share/systemd/user/xdg-desktop-portal-gtk.service ~/.config/systemd/user/xdg-desktop-portal-gtk.service
```
But in this case I have to symlink each required service

There is a similar issue: https://github.com/nix-community/home-manager/issues/4922, but I started a new one, because this thread is about more wide problem rather just a problem with a `xdg.portal.enable` option. Will it be possible to control over systemd.user services installed from nixpkgs on non NixOS environments?

khlopkov:
@baloo from the linked issue's thread suggested adding this activation script into home-manager's configuration:

```nix
  home.activation.linkSystemd = let
    inherit (lib) hm;
  in hm.dag.entryBefore [ "reloadSystemd" ] (''
    find $HOME/.config/systemd/user/ \
      -type l \
      -exec bash -c "readlink {} | grep -q $HOME/.nix-profile/share/systemd/user/" \; \
      -delete

    find $HOME/.nix-profile/share/systemd/user/ \
      -type f -o -type l \
[...]

Issue #4922, opened 2024-01-18
bug: xdg-desktop-portal-*.service files not found


johanneshorner:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Using the [xdg-portal](https://github.com/Misterio77/home-manager/blob/fcbc70a7ee064f2b65dc1fac1717ca2a9813bbe6/modules/misc/xdg-portal.nix) module with the config
```
xdg.portal = {
  enable = true;
  extraPortals = with pkgs; [
    xdg-desktop-portal-wlr
    <hyprland from flake>.xdg-desktop-portal-hyprland
    xdg-desktop-portal-gtk
  ];
  configPackages = [
    <hyprland from flake>.default
  ];
}
```
results in only the `xdg-desktop-portal` service running with warnings that it can't find the other portal services:
```
● xdg-desktop-portal.service - Portal service
     Loaded: loaded (/home/johannes/.config/systemd/user/xdg-desktop-portal.service; linked; preset: enabled)
     Active: active (running) since Thu 2024-01-18 17:56:31 CET; 9min ago
   Main PID: 314794 (.xdg-desktop-po)
      Tasks: 6 (limit: 38049)
     Memory: 17.3M
        CPU: 49ms
     CGroup: /user.slice/user-1000.slice/user@1000.service/session.slice/xdg-desktop-portal.service
             └─314794 /nix/store/0n9s802xa305dppw1k9rz9w5nf29zy1k-xdg-desktop-portal-1.18.2/libexec/xdg-desktop-portal

Jan 18 17:56:31 wslinuxhoo .xdg-desktop-po[314794]: No skeleton to export
Jan 18 17:56:31 wslinuxhoo .xdg-desktop-po[314794]: Failed to create email proxy: Error calling StartServiceByName for org.freedesktop.impl.portal.desktop.gtk: Unit xdg-desktop-portal-gtk.service not found.
Jan 18 17:56:31 wslinuxhoo .xdg-desktop-po[314794]: No skeleton to export
Jan 18 17:56:31 wslinuxhoo .xdg-desktop-po[314794]: Failed to create global_shortcuts proxy: Error calling StartServiceByName for org.freedesktop.impl.portal.desktop.hyprland: Unit xdg-desktop-portal-hyprland.service not found.
[...]

PR #4990, opened 2024-02-02
systemd: import services from packages

khlopkov:
### Description

Packages installed through `home.packages` can contain systemd services. Those services are stored in `~/.nix-profile/share/systemd/user/` but they do not exist in `~/.config/systemd/user/`. That's why currently it is not possible to enable those services without linking them manually (or copying/creating new ones). I described in the following issue: https://github.com/nix-community/home-manager/issues/4972

This PR intents to add a new option `systemd.user.servicesOfPackages` that supposed to enable services declared in packages provided into the option.

Also that causing a bug (https://github.com/nix-community/home-manager/issues/4922) on non-NixOS environments with `xdg.portal.extraPortals` options, because services from `extraPortals` packages can not be found by `xdg-desktop-portal.service`. This PR fixes this bug as well by using the new `systemd.user.servicesOfPackages` option. It seems that there are other similar issues listed in the thread with other options, this PR does not fix them, but I hope this PR will help to fix them easier.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


Nemo157:
AFAIK linking these services in isn't necessary, the combined `share/systemd/user` directory is in the search-path already (check `systemd-path systemd-search-user-unit`). The issue is more that there needs to be some way to _enable_ units defined in the packages (and not just services, sockets etc. too) by adding them to the generated `.wants` directories.

chaorace:
I just so happen to be a non-NixOS Linux user who is also affected by the issue with the extraPortals option, so I checked my own paths via `systemd-path` and confirmed that the nix-profile paths do already exist there.

With that being said, I think it would be better to properly process the [Install] section of the service unit file rather than do the linking ourselves. It looks like HM has avoided using `systemctl enable` in the past, but perhaps this is a situation where that would be warranted?

In any case, one change I'd like to see with the currently proposed configuration setup is to provide more granularity in service selection.

I seem to recall that some packages provide multiple versions of the same service where only one is intended to be used at a given time. To better handle cases like these, there should be some optional mechanism to only install specifically named services from the specified package.

AntonFriberg:
Hi! I am also running non-nixos (Ubuntu 24) and was very confused about why I was not getting xdg-desktop-portal to start or even be listed. It would be very nice if we could get some more attention on this and hopefully have a solution for non-nixOS users like myself.

DerRockWolf:
> AFAIK linking these services in isn't necessary, the combined `share/systemd/user` directory is in the search-path already (check `systemd-path systemd-search-user-unit`). 

Can someone verify this on a different system (I'm on Ubuntu 24.04).

There `systemd-path systemd-search-user-unit` does indeed contain `$HOME/.nix-profile/share/systemd/user` but `systemctl --user show -p UnitPath` does not and units that are not linked to `~/.config/systemd/user` cannot be found by systemd!

> The issue is more that there needs to be some way to _enable_ units defined in the packages (and not just services, sockets etc. too) by adding them to the generated `.wants` directories.

For this I would suggest to replicate how NixOS handles it: `systemd.*.<name>.wantedBy` ([implementation](https://github.com/NixOS/nixpkgs/blob/d9bc5c7dceb30d8d6fafa10aeb6aa8a48c218454/nixos/lib/systemd-lib.nix#L501-L511))

Until this is properly implemented one can just do:
```nix
xdg.configFile = {
  "systemd/user/default.target.wants/<name>.service".source = "${pkgs.<name>}/share/systemd/user/<name>.service";
};
```

modules/systemd.nix
```diff
@@ -29,6 +29,23 @@ let
       in "${key}=${value'}";
   };
 
+  importServicesFromPkg = pkg:
+    let
+      isService = (path: type: (builtins.match ".service$" path) == null);
```
DerRockWolf, line 34: This currently only includes files that aren't matching (`== null`) 😅 

I would do it like this (and also include `.target` files):


```suggestion
      isService = (path: type: lib.hasSuffix ".service" path || lib.hasSuffix ".target" path);
```

modules/systemd.nix
```diff
@@ -123,6 +140,12 @@ in {
         example = unitExample "Service";
       };
 
+      servicesOfPackages = mkOption {
```
DerRockWolf, line 143: As a similar option already exists for NixOS I would suggest to call it similarly.
```suggestion
      packages = mkOption {
```

modules/systemd.nix
```diff
@@ -287,7 +310,7 @@ in {
   # available, in particular we assume that systemctl is in PATH.
   # Do not install any user services if username is root.
   config = mkIf (pkgs.stdenv.isLinux && config.home.username != "root") {
-    xdg.configFile = mkMerge [
+    xdg.configFile = mkMerge ([
```
DerRockWolf, line 313: As mentioned in https://github.com/nix-community/home-manager/issues/4972#issuecomment-3036799318 it would be "more" correct to use `xdg.dataFile`

DerRockWolf (changes requested):
Some few things I found 🙂

Would be great if we could get this merged soon.

Nemo157:
> There `systemd-path systemd-search-user-unit` does indeed contain `$HOME/.nix-profile/share/systemd/user` but `systemctl --user show -p UnitPath` does not and units that are not linked to `~/.config/systemd/user` cannot be found by systemd!

Mine does

```console
> systemctl --user show -p UnitPath | rg --only-matching "$HOME/.nix-profile/share/systemd/user"
/home/nemo157/.nix-profile/share/systemd/user
```

But that's not where the services seem to get installed, instead they're in `/etc/profiles/per-user/$USER/share/systemd/user` (maybe because I use home-manager via the nixos module).

For the [`u2f-touch-detector` module](https://github.com/Nemo157/u2f-touch-detector/blob/main/module.nix) I don't explicitly install the service anywhere, but I have the package in `home.packages` and that causes the service to end up in `/etc/profiles/per-user/nemo157/share/systemd/user/u2f-touch-detector.service`, then the module just adds the correct `.wants` symlink to make systemd activate it.

DerRockWolf:
@Nemo157 yes, because you are on NixOS. There everything works as expected because `~/.nix-profile/share/systemd/user` is already included in the unit search path.
On non NixOS systems where you just use home-manager `~/.nix-profile/share/systemd/user` is not included in the unit search path, because the distribution obviously doesn't know that there might be units located there.

Nemo157:
Maybe on non-NixOS it should automatically replicate the same kind of setup: symlink all units from `home.packages` into `xdg.dataHome`.

teto:
can you address and rebase comments please ?

modules/systemd.nix
Invizory: Moreover, we already have a similar option `dbus.packages`, so this would be better for consistency.

Invizory:
I have created an alternative implementation in #8540, focused solely on systemd user units without messing with `xdg.portal`, which I believe could be discussed and resolved in #8541.

khaneliman:
Superseded by #8540; remaining portal wiring continues in #8541.

PR #4999, opened 2024-02-05
neomutt: move crypt_use_gpgme from account to global config

arjan-s:
### Description
Move `crypt_use_gpgme` configuration option from the account config to the global config.
Fixes config load error since newer versions of neomutt don't allow `crypt_use_gpgme` to be loaded dynamically (from folder hooks).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@fpletz @ncfavier @sumnerevans

### DIFF
--- a/modules/programs/neomutt.nix
+++ b/modules/programs/neomutt.nix
@@ -224,7 +224,6 @@ let
       set certificate_file=${toString config.accounts.email.certificatesFile}
 
       # GPG section
-      set crypt_use_gpgme = yes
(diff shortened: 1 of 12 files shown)


arjan-s:
Notes:

- I'm not sure about backwards compatibility, but given the sort of change, I expect there to be no problems.
- Tests for neomutt are successful, but some unrelated firefox test is failing.

rycee:

PR #5003, opened 2024-02-06
programs.fish: Configurable generation of completions.

jcollie:
### Description

Add a config option to optionally disable the generation of command completions based upon man pages.

### Checklist

- [?] Code tested through `nix-shell --pure tests -A run.all` or `nix develop --ignore-environment .#all` using Flakes.

  The tests fail on the master branch for me my changes made no difference.

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -253,6 +253,12 @@ in {
         '';
       };
 
+      generateCompletions = mkEnableOption
+        "the automatic generation of completions based upon installed man pages"
+        // {
+          default = true;
+        };
+
       shellAliases = mkOption {
         type = with types; attrsOf str;
         default = { };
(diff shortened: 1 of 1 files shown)


modules/programs/fish.nix
```diff
@@ -253,6 +253,15 @@ in {
...
       };
 
+      generateCompletions = mkOption {
+        type = types.bool;
+        default = true;
+        defaultText = literalExpression "true";
+        description = ''
+          Enable or disable the automatic generation of completions based upon installed man pages.
+        '';
+      };
```
sumnerevans, line 263:
```suggestion
      generateCompletions = mkEnableOption "the automatic generation of completions based upon installed man pages" // {
        default = true;
      };
```

sumnerevans (changes requested):

modules/programs/fish.nix
jcollie: Thanks for the feedback! I've updated the patch.

sumnerevans (changes requested):
We should probably have a test for this. (No idea why the macos tests are failing, but that doesn't look at all related)

dawidd6 (approved):

rycee:

PR #5005, opened 2024-02-06
tealdeer: module improvements

pedorich-n:
### Description

As discussed in https://github.com/nix-community/home-manager/pull/4968#issuecomment-1927480035:

- Cache update on HM activation removed;
- `freeformType` settings;
- More tests added;
- Platform-dependent tests;
- Maintainer added;

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

N/A

### DIFF
--- a/modules/programs/tealdeer.nix
+++ b/modules/programs/tealdeer.nix
@@ -4,31 +4,64 @@ with lib;
 let
   cfg = config.programs.tealdeer;
 
-  tomlFormat = pkgs.formats.toml { };
-
   configDir = if pkgs.stdenv.isDarwin then
     "Library/Application Support"
   else
     config.xdg.configHome;
 
+  tomlFormat = pkgs.formats.toml { };
+
+  settingsFormat = let
+    updatesSection = types.submodule {
+      options = {
+        auto_update = mkEnableOption "auto-update";
+
+        auto_update_interval_hours = mkOption {
+          type = types.ints.positive;
+          default = 720;
+          example = literalExpression "24";
+          description = ''
+            Duration, since the last cache update, after which the cache will be refreshed.
+            This parameter is ignored if {var}`auto_update` is set to `false`.
+          '';
+        };
+      };
+    };
+  in types.submodule {
+    freeformType = tomlFormat.type;
+    options = {
+      updates = mkOption {
+        type = updatesSection;
+        default = { };
+        description = ''
+          Tealdeer can refresh the cache automatically when it is outdated.
+          This behavior can be configured in the updates section.
+        '';
+      };
+    };
+  };
+
 in {
-  meta.maintainers = [ ];
+  meta.maintainers = [ hm.maintainers.pedorich-n ];
+
+  imports = [
+    (mkRemovedOptionModule [ "programs" "tealdeer" "updateOnActivation" ] ''
+      Updating tealdeer's cache requires network access.
+      The activation script should be fast and idempotent, so the option was removed.
+      Please use
+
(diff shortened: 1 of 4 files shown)


modules/programs/tealdeer.nix
```diff
@@ -4,29 +4,59 @@ with lib;
...
   else
     config.xdg.configHome;
 
+  tomlFormat = pkgs.formats.toml { };
+
+  settingsFormat = let
+    updatesSection = types.submodule {
+      options = {
+        auto_update = mkOption {
+          type = types.bool;
```
pedorich-n, line 18: I'm not sure about this. Should I make this `nullOr bool`?
Currently, this mimics the default [settings](https://dbrgn.github.io/tealdeer/config_updates.html#auto_update), but in the future, it might change. Maybe it's better to not produce a config file at all?

modules/programs/tealdeer.nix
```diff
@@ -4,29 +4,59 @@ with lib;
...
+          default = false;
+          defaultText = literalExpression "false";
+          example = literalExpression "true";
+          description = ''
+            Specifies whether the auto-update feature should be enabled.
+          '';
+        };
+        auto_update_interval_hours = mkOption {
+          type = types.ints.positive;
+          default = 720;
```
pedorich-n, line 28: Same here. 720 is the [default](https://dbrgn.github.io/tealdeer/config_updates.html#auto_update_interval_hours)

pedorich-n:
@rycee would you mind having a look, please?

modules/programs/tealdeer.nix
```diff
@@ -4,29 +4,59 @@ with lib;
...
+          defaultText = literalExpression "720";
```
sumnerevans, line 29: `defaultText` isn't required for simple expressions anymore.

```suggestion
```

modules/programs/tealdeer.nix
sumnerevans: Can't remember if this is possible, but maybe we can make the `options` submodule actually default to `null` and just not include it if none of the options are specified?

modules/programs/tealdeer.nix
```diff
@@ -4,29 +4,59 @@ with lib;
...
     "Library/Application Support"
   else
     config.xdg.configHome;
 
+  tomlFormat = pkgs.formats.toml { };
+
+  settingsFormat = let
+    updatesSection = types.submodule {
+      options = {
+        auto_update = mkOption {
```
sumnerevans, line 17: this should just be a `mkEnableOption`

modules/programs/tealdeer.nix
```diff
@@ -4,29 +4,59 @@ with lib;
...
+  };
+
 in {
-  meta.maintainers = [ ];
+  meta.maintainers = [ hm.maintainers.pedorich-n ];
 
   options.programs.tealdeer = {
     enable = mkEnableOption "Tealdeer";
 
-    updateOnActivation = mkOption {
```
sumnerevans, line 20: this needs a `mkRenamedOptionModule` or `mkRemovedOptionModule`

sumnerevans (changes requested):

modules/programs/tealdeer.nix
pedorich-n: @sumnerevans please have another look. I made the submodule of type `types.nullOr settingsFormat;` with the default value of `null`

sumnerevans (approved):

rycee:
Great, thanks! Merged to master now 🙂

PR #5007, opened 2024-02-07
zellij: use exact cmd instead of relying on $PATH being set

poliorcetics:
### Description

Use exact command when calling `zellij` instead of relying on it being in the `PATH` already.

### Checklist

- [ ] **Change is backwards compatible**: it **may** not be if someone was overriding the path with a different `zellij`. I'm unsure if this is important or not for home-manager ?

- [ ] ~~Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).~~ irrelevant

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/zellij.nix
+++ b/modules/programs/zellij.nix
@@ -6,6 +6,7 @@ let
 
   cfg = config.programs.zellij;
   yamlFormat = pkgs.formats.yaml { };
+  zellijCmd = getExe cfg.package;
 
 in {
   meta.maintainers = [ hm.maintainers.mainrs ];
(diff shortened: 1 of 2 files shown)


modules/programs/zellij.nix
```diff
@@ -6,6 +6,7 @@ let
 
   cfg = config.programs.zellij;
   yamlFormat = pkgs.formats.yaml { };
+  zellijCmd = "${config.home.profileDirectory}/bin/zellij";
```
pedorich-n, line 9: I believe it's better to use `${lib.getExe cfg.package}`, instead of relying on a hard-coded path.

rycee: Yes, agreed 🙂

rycee:
Thanks for the contribution! I think this makes a lot of sense. I think it is worth using `getExe` as per the earlier comment, after that I think it it is ready to merge.

Issue #3091, opened 2022-07-14
bug: Unable to install home-manager (NixOS, standalone)


NielsGx:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Hello,

on a VPS, built with NixOS 21.11, I updated the channel to 22.05, rebuilt, then changed my configuration.nix to have a user
then on this user, I add the channel for home-manager 22.05 and update it.

but when I do "nix-shell '<home-manager>' -A install" I get : 
error: file 'home-manager' was not found in the Nix search path (add it using $NIX_PATH or -I)

I tried to reboot the vps, nothing. Also ssh into root and then do su myuser, but this shouldn't cause issue.

I tried to mess around with $NIX_PATH but then it would return a permission error

It does work if I add the channel to the root and install with it, but afaik the manual says to use a standard user, at least for the standalone install

(no, I don't want to update home-manager through nixos-rebuild)

thanks a lot for this awesome tool

### System information

```markdown
- system: `"x86_64-linux"`                      - host os: `Linux 5.15.53, NixOS, 22.05 (Quokka), 22.05.1627.4aceab3cadf`                       - multi-user?: `no`                             - sandbox: `yes`
 - version: `nix-env (Nix) 2.8.1`
 - channels(niels): `"home-manager-22.05.tar.gz"`
 - channels(root): `"nixos-22.05"`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`

(idk why it says no to multi-user, bc I have root and a user "niels")
```

zeratax:
i have the same issue on a brand new install of 22.11

mupinnn:
Hello, I'm experiencing similar issue but in my case when installing [home-manager](https://nix-community.github.io/home-manager).

```sh
$ nix-shell '<home-manager>' -A install
The file /home/sandbox/.config/home-manager/home.nix already exists, leaving it unchanged...

Creating initial Home Manager generation...

mktemp: failed to create directory via template ‘/run/user/1000/home-manager-build.XXXXXXXXXX’: No such file or directory
error: getting status of '/run/user/1000': No such file or directory
Uh oh, the installation failed! Please create an issue at

    https://github.com/nix-community/home-manager/issues

if the error seems to be the fault of Home Manager.
rm: cannot remove '': No such file or directory
```

Some information I can provide about my Nix setup:
```sh
$ echo $NIX_PATH
[...]

PR #5008, opened 2024-02-08
Documentation: Standalone Installation Note

skyethepinkcat:
Added a note to the standalone installation instructions, which recommends adding -I ~/.nix-defexpr/channels to the nix-shell command if home-manager cannot be found.

### Description
This is just a documentation update -- I've previously run into 
issues where after adding the home-manager channel as a user,
it would not show up in the $NIX_PATH, I thought it'd be a good idea to mention it.

### DIFF
--- a/docs/manual/installation/standalone.md
+++ b/docs/manual/installation/standalone.md
@@ -35,6 +35,14 @@
 
     Once finished, Home Manager should be active and available in your
     user environment.
+    
+::: {.note}
+If you are having issues here related to home-manager not being found and added the channel as a user earlier, try running
+``` shell
+$ nix-shell '<home-manager>' -A install -I ~/.nix-defexpr/channels
+```
+instead. This tells nix-shell to search the correct path.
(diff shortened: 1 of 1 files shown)


Zocker1999NET:
I assume your solution does (attempt to) fix #3091. In that context: is your proposal a long-term solution? Because, while I also tried to fix #3091 for myself, I discovered this issue has something to do with https://github.com/NixOS/nix/issues/2033#issuecomment-1366974053. So I tried:

- Using a different command for nix-env at the installation. This worked for the first generation but afterwards I still was not able to execute any `home-manager` command because it did not found the channel again.
- Attempted to change NIX_PATH persistently (e.g. https://github.com/NixOS/nix/issues/2033#issuecomment-744059453). This worked persistently, but I wasn’t quite happy with that.

In the end, I decided to add those channels using `sudo` to have all nix-channels I use together. And it also fixes #3091 persistently for me.

PR #5015, opened 2024-02-10
emacs: use `overrideScope` instead of `overrideScope'`

zendo:
fix annoying warning:
```
trace: warning: `overrideScope'` (from `lib.makeScope`) has been renamed to `overrideScope`.
```

https://github.com/NixOS/nixpkgs/blob/master/lib/customisation.nix#L328


rycee:

PR #5016, opened 2024-02-10
jujutsu: fix completions commands

cdmistman:
### Description

starting in jujutsu [0.14.0](https://github.com/martinvonz/jj/releases/tag/v0.14.0), jujutsu prefers `jj util completion <shell>` instead of `jj util completion --<shell>`. When the flag variation is used, jujutsu prints the following on shell startup:
```
Warning: `jj util completion --zsh` will be removed in a future version, and this will be a hard error
Hint: Use `jj util completion zsh` instead
```

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@shikanime

### DIFF
--- a/modules/programs/jujutsu.nix
+++ b/modules/programs/jujutsu.nix
@@ -65,12 +65,12 @@ in {
     '';
 
     programs.zsh.initExtra = mkIf cfg.enableZshIntegration ''
-      source <(${pkgs.jujutsu}/bin/jj util completion --zsh)
+      source <(${pkgs.jujutsu}/bin/jj util completion zsh)
       compdef _jj ${pkgs.jujutsu}/bin/jj
     '';
 
     programs.fish.interactiveShellInit = mkIf cfg.enableFishIntegration ''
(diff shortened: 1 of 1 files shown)


teto:
this looks ok but can it be safely enabled before 0.14 ? the HM flake still points at 0.13

ambroisie:
I don't quite understand how the module came to explicitly call this, the package already took care of installing the completion files which should be picked up by home-manager when it manages the shell conifguration.

And this was fixed when the update was merged upstream ;-) (https://github.com/NixOS/nixpkgs/commit/59b0ff3b92f980d16d749930fa0d84c056d19a5b).

rycee:
Nice, yeah if the package provides the completion files then there should be no need to also have them here. So in principle it should be OK to remove the shell init configuration and remove the "enable integration" options (using `mkRemovedOptionModule`).

shikanime:
I wasn't even aware that Nixpkgs already handle shell completions, but does it work with non-NixOS installations ?

cdmistman:
ok, closing and another PR can be opened to remove the completions later

PR #5020, opened 2024-02-12
joplin-desktop: add basic options

zorrobert:
### Description

This module adds basic options to configure [joplin-desktop](https://joplinapp.org/) via home-manager.
Joplin uses a JSON file for its configuration. To avoid overriding an existing configuration, settings already in that file will be kept and only overridden if they are set via home-manager options.

The errors listed below still need to be solved but I don't know how to do that. I'm new to writing home-manager modules (or nix modules), so feedback about the code or tips how to simplify things are very welcome. 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/joplin-desktop.nix
+++ b/modules/programs/joplin-desktop.nix
@@ -1,0 +1,129 @@
+{ config, lib, pkgs, ... }:
+let
+  cfg = config.programs.joplin-desktop;
+  # config path is the same for linux and mac
+  configPath = (config.home.homeDirectory + "/.config/joplin-desktop");
+in {
+  options.programs.joplin-desktop = {
+    enable = lib.mkEnableOption "joplin-desktop";
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.joplin-desktop;
+      defaultText = lib.literalExpression "pkgs.joplin-desktop";
+      description = "Package containing the joplin-desktop program.";
+    };
+    extraConfig = lib.mkOption {
+      type = lib.types.attrs;
+      default = { };
+      description = ''
+        Use this to add options to the Joplin config file. Settings are written in JSON, so "sync.interval":600 is written as "sync.interval" = 600'';
+      example = {
+        "newNoteFocus" = "title";
+        "markdown.plugin.mark" = true;
+      };
+    };
+    ### General
+    general = {
+      editor = lib.mkOption {
+        type = lib.types.nullOr lib.types.str;
+        default = null;
+        example = "kate";
+        description =
+          "The editor command (may include arguments) that will be used to open a note. If none is provided Joplin will try to auto-detect the default editor.";
+      };
+    };
+    ### Sync
+    sync = {
+      target = lib.mkOption {
+        type = lib.types.enum [
+          null
+          "none"
+          "file-system"
+          "onedrive"
+          "nextcloud"
+          "webdav"
+          "dropbox"
+          "s3"
+          "joplin-server"
+          "joplin-cloud"
+        ];
+        default = null;
+        description = "What is the type of sync target.";
+        example = "dropbox";
+      };
+      interval = lib.mkOption {
+        type = lib.types.nullOr lib.types.int;
+        default = null;
+        description = ''
+          Sync Interval in seconds. Only values shown as options in the settings are valid.
+          Disabled: 0
+          5 Min: 300
+          10 Min: 600
+          30 Min: 1800
+          1 hour: 3600
+          12 Hours: 43200
+          24 Hours: 86400
+        '';
+        example = 600;
+      };
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    # write new config to temporary read-only file
+    home.file."home-manager-settings.json" = {
+      target = (configPath + "/home-manager-settings.json");
+      text = builtins.toJSON ((
+        # check if config file exists
+        if builtins.pathExists (configPath + "/settings.json") then
+          let
+            currentConfig = builtins.readFile (configPath + "/settings.json");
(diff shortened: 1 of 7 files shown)


zorrobert:
I tried running `nix-shell --pure tests -A run.all`, but the test fails while building the derivation for gpg-pubring. I think that the problem is something on my system, so it would be helpful if somebody could run the tests again just to be sure. Still, because the error is not with the joplin-desktop module, I think that this is ready for review / merge.

modules/programs/joplin-desktop.nix
```diff
@@ -0,0 +1,129 @@
...
+  configPath = (config.home.homeDirectory + "/.config/joplin-desktop");
+in {
+  options.programs.joplin-desktop = {
+    enable = lib.mkEnableOption "joplin-desktop";
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.joplin-desktop;
+      defaultText = lib.literalExpression "pkgs.joplin-desktop";
+      description = "Package containing the joplin-desktop program.";
+    };
```
rycee, line 14: Can use the `mkPackageOption` function. Also please separate options by empty lines:

```suggestion

    package = lib.mkPackageOption pkgs "joplin-desktop" { };

```

modules/programs/joplin-desktop.nix
```diff
@@ -0,0 +1,129 @@
...
+        default = null;
+        description = ''
+          Sync Interval in seconds. Only values shown as options in the settings are valid.
+          Disabled: 0
+          5 Min: 300
+          10 Min: 600
+          30 Min: 1800
+          1 hour: 3600
+          12 Hours: 43200
+          24 Hours: 86400
```
rycee, line 65: Not sure I understand the description, only 300, 600, etc. are valid values for the interval? If so, then I would suggest using an enum:

``` nix
type = types.enum [ 300 600 1800 3600 43200 86400 ];
```

or even

``` nix
type = types.enum [ "5m" "10m" "30m" "1h" "12h" "1d" ];
```

zorrobert: These are the values that are available via the GUI, from my testing it does not seem like other values can be used. When using `type = types.enum [ "5m" "10m" "30m" "1h" "12h" "1d" ]`, we need some way to convert that into a number of seconds to write to the config file and I didn't find an elegant way to do so.

modules/programs/joplin-desktop.nix
zorrobert: Thx, I'll do that now

modules/programs/joplin-desktop.nix
```diff
@@ -0,0 +1,129 @@
...
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    # write new config to temporary read-only file
+    home.file."home-manager-settings.json" = {
+      target = (configPath + "/home-manager-settings.json");
+      text = builtins.toJSON ((
```
rycee, line 78: I think this is best done using existing Home Manager functionality:

``` nix
xdg.configFile."joplin-desktop/settings.json".source = jsonFormat.generate "joplin-settings.json" ((
```

where

``` nix
```

is defined at the top of the file, after `cfg = config.programs.joplin-desktop;`.

If you really do need to merge the HM generated configuration into the one generated by Joplin Desktop, then you don't need to put anything in the user's home directory. You will instead need use `jq` in the activation block and reference the HM generated file directly in the Nix store.

modules/programs/joplin-desktop.nix
zorrobert: I changes it from 
`type = lib.types.nullOr lib.types.int;`
to 
`type = lib.types.nullOr type = types.enum [ 300 600 1800 3600 43200 86400 ];`

modules/programs/joplin-desktop.nix
zorrobert: Yes it has to be merged because Joplin writes some other things in that file. I'll rewrite it to use `jq` and the file from the nix store.

zorrobert: I've changed the way the config file is written to use `jq` like you suggested, see latest commit.

modules/programs/joplin-desktop.nix
rycee: The typical way to do such a mapping would be something like

``` nix
intervalToSeconds = i:
  {
    "5m"  =   300;
    "10m" =   600;
    "30m" =  1800;
    "1h"  =  3600;
    "12h" = 43200;
    "1d"  = 86400;
  }.${i} or null;
```

Example:

```
nix-repl> intervalToSeconds 4
error:
       … while evaluating an attribute name

         at «none»:0: (source not available)

       error: value is an integer while a string was expected

nix-repl> intervalToSeconds "sdf"
null

nix-repl> intervalToSeconds "5m"  
300

nix-repl> intervalToSeconds "1d" 
86400
```

zorrobert: Thanks a lot for that, I've used that now for the sync target as well.

zorrobert:
@rycee is there anything that has to be done before this is merged?

rycee:
Thanks! I made some additional fixes, squashed, and rebased to master in 383296ffa45b539c28bf79ec2a272f652838ddd1 🙂

zorrobert:
@rycee I used meta.maintainers wrong in joplin-desktop.nix, which causes the automated test nix-build --show-trace -A docs.jsonModuleMaintainers to fail for other modules. I fixed it in the latest commit, sorry
https://github.com/zorrobert/home-manager/commit/220a42e38d31f9605a0c5faf762bd63a857eb975

rycee:
No worries, it is fixed in master now 🙂

PR #5023, opened 2024-02-12
scrutiny-collector: add module

dudeofawesome:
### Description

Adding a module to configure & run the [Scrutiny](https://github.com/AnalogJ/scrutiny) collector.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/services/scrutiny-collector.nix
```diff
@@ -0,0 +1,127 @@
+{ config, osConfig, lib, pkgs, ... }:
+
+let
+  inherit (lib) mkEnableOption mkIf mkOption mkPackageOption types mdDoc;
+
+  cfg = config.services.scrutiny-collector;
+in {
+  meta.maintainers = [ lib.maintainers.dudeofawesome or "dudeofawesome" ];
```
rycee, line 8: If you're not in the Nixpkgs maintainer list then you'll need to add yourself to `maintainers.nix` in this repository and change this line to
```suggestion
  meta.maintainers = [ lib.hm.maintainers.dudeofawesome ];
```

modules/services/scrutiny-collector.nix
```diff
@@ -0,0 +1,127 @@
...
+
+  options.services.scrutiny-collector = {
+    enable = mkEnableOption (mdDoc "scrutiny-collector");
```
rycee, line 11:
```suggestion
    enable = mkEnableOption "scrutiny-collector";
```

modules/services/scrutiny-collector.nix
```diff
@@ -0,0 +1,127 @@
...
+
+    package = mkPackageOption pkgs "scrutiny-collector" { };
+
+    config-path = mkOption {
```
rycee, line 15: Options should be camel cased

```suggestion
    configPath = mkOption {
```

modules/services/scrutiny-collector.nix
```diff
@@ -0,0 +1,127 @@
...
+    # TODO: this needs to be a list of calendar attrs
+    calendar = mkOption {
+      type = types.str;
+      default = "*-*-* 00:00:00";
+      description = mdDoc ''
+        Configured when to run the service systemd unit (DayOfWeek Year-Month-Day Hour:Minute:Second).
+      '';
+    };
+
+    # TODO: is this relevant?
```
rycee, line 67: If it is a user service it would typically run as the user, if you want to run as some other user then I think a system service is best. I.e., I think the user and group options can be removed.

modules/services/scrutiny-collector.nix
```diff
@@ -0,0 +1,127 @@
...
+
+    config-path = mkOption {
+      type = types.nullOr types.path;
+      default = null;
+      description = mdDoc ''
+        Specify the path to the devices file
+      '';
+    };
+
+    configuration = mkOption {
```
rycee, line 23: The standard name for freeform options is `settings`, so should do
```suggestion
    settings = mkOption {
```

modules/services/scrutiny-collector.nix
```diff
@@ -0,0 +1,127 @@
...
+  options.services.scrutiny-collector = {
+    enable = mkEnableOption (mdDoc "scrutiny-collector");
+
+    package = mkPackageOption pkgs "scrutiny-collector" { };
+
+    config-path = mkOption {
+      type = types.nullOr types.path;
+      default = null;
+      description = mdDoc ''
+        Specify the path to the devices file
```
rycee, line 19:
```suggestion
        Specify the path to the devices file.
```

modules/services/scrutiny-collector.nix
```diff
@@ -0,0 +1,127 @@
...
+      description = lib.mdDoc ''
+        Specify the configuration for the Scrutiny collector in Nix.
+      '';
+    };
+
+    api-endpoint = mkOption {
+      type = types.str;
+      default = "http://localhost:8080";
+      description = mdDoc ''
+        The api server endpoint
```
rycee, line 35:
```suggestion
        The API server endpoint.
```

modules/services/scrutiny-collector.nix
```diff
@@ -0,0 +1,127 @@
...
+      '';
+    };
+
+    log-file = mkOption {
+      type = types.nullOr types.path;
+      default = null;
+      description = mdDoc ''
+        Path to file for logging. Leave empty to use STDOUT
```
rycee, line 43:
```suggestion
        Path to file for logging. Leave empty to use `STDOUT`.
```

modules/services/scrutiny-collector.nix
```diff
@@ -0,0 +1,127 @@
...
+    };
+
+    debug = mkEnableOption "debug logging";
+
+    host-id = mkOption {
+      type = types.str;
+      default = osConfig.networking.hostName;
+      defaultText = lib.literalExpression "config.networking.hostName";
+      description = mdDoc ''
+        Host identifier/label, used for grouping devices
```
rycee, line 54:
```suggestion
        Host identifier/label, used for grouping devices.
```

modules/services/scrutiny-collector.nix
```diff
@@ -0,0 +1,127 @@
...
+    configFile = if cfg.configuration != null then
+      (pkgs.writeTextFile {
+        name = "collector.yaml";
+        text = (lib.generators.toYAML { } cfg.configuration);
+      })
+    else
+      null;
+  in lib.mkMerge [
+    { home.packages = [ cfg.package ]; }
+    (mkIf pkgs.stdenv.targetPlatform.isDarwin {
```
rycee, line 96: If the intent is to have it be Darwin only then will need to add an assert similar to https://github.com/nix-community/home-manager/blob/17431970b4ebc75a92657101ccffcfc9e1f9d8f0/modules/services/mpd.nix#L158-L160

rycee:
Thanks! Added some comments.

PR #5035, opened 2024-02-18
wlsunset: update options

Daru-san:
### Description

Adds the `output`, `sunrise` and `sunset` config options to wlsunset.nix

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@matrss

### DIFF
--- a/modules/services/wlsunset.nix
+++ b/modules/services/wlsunset.nix
@@ -11,7 +11,7 @@ in {
     enable = mkEnableOption "wlsunset";
 
     package = mkOption {
-      type = types.package;
+      type = with types; package;
       default = pkgs.wlsunset;
       defaultText = "pkgs.wlsunset";
       description = ''
@@ -20,7 +20,9 @@ in {
     };
 
     latitude = mkOption {
-      type = types.str;
+      type = with types; nullOr (either str (either float int));
+      default = null;
+      example = -74.3;
       description = ''
         Your current latitude, between `-90.0` and
         `90.0`.
@@ -28,7 +30,9 @@ in {
     };
 
     longitude = mkOption {
-      type = types.str;
+      type = with types; nullOr (either str (either float int));
+      default = null;
+      example = 12.5;
       description = ''
         Your current longitude, between `-180.0` and
         `180.0`.
@@ -37,7 +41,7 @@ in {
 
     temperature = {
       day = mkOption {
-        type = types.int;
+        type = with types; int;
         default = 6500;
         description = ''
           Colour temperature to use during the day, in Kelvin (K).
@@ -46,7 +50,7 @@ in {
       };
 
       night = mkOption {
-        type = types.int;
+        type = with types; int;
         default = 4000;
         description = ''
           Colour temperature to use during the night, in Kelvin (K).
@@ -56,15 +60,42 @@ in {
     };
 
     gamma = mkOption {
-      type = types.str;
-      default = "1.0";
+      type = with types; (either str (either float int));
+      default = 1.0;
+      example = 0.6;
       description = ''
         Gamma value to use.
       '';
     };
 
+    output = mkOption {
+      type = with types; nullOr str;
+      default = null;
+      description = ''
+        Name of output to use, by default all outputs are used.
+      '';
+    };
+
+    sunrise = mkOption {
+      type = with types; nullOr str;
+      default = null;
+      example = "06:30";
+      description = ''
+        The time when the sun rises (in 24 hour format).
+      '';
+    };
+
+    sunset = mkOption {
+      type = with types; nullOr str;
+      default = null;
+      example = "18:00";
+      description = ''
+        The time when the sun sets (in 24 hour format).
+      '';
+    };
+
     systemdTarget = mkOption {
-      type = types.str;
+      type = with types; str;
       default = "graphical-session.target";
       description = ''
         Systemd target to bind to.
@@ -76,6 +107,22 @@ in {
     assertions = [
       (lib.hm.assertions.assertPlatform "services.wlsunset" pkgs
         lib.platforms.linux)
+      {
+        assertion = (cfg.sunrise != null || cfg.sunset != null)
+          != (cfg.latitude != null || cfg.longitude != null);
+        message =
+          "Either `sunrise` and `sunset` together or `longitude` and `latitude` together must be set for wlsunset";
+      }
+      {
+        assertion = (cfg.sunrise != null) == (cfg.sunset != null);
+        message =
+          "Both `sunset` and `sunrise` together must be set for wlsunset";
+      }
+      {
+        assertion = (cfg.latitude != null) == (cfg.longitude != null);
+        message =
+          "Both `latitude and `longitude` together must be set for wlsunset";
+      }
     ];
 
     systemd.user.services.wlsunset = {
@@ -86,14 +133,17 @@ in {
 
       Service = {
         ExecStart = let
-          args = [
-            "-l ${cfg.latitude}"
-            "-L ${cfg.longitude}"
-            "-t ${toString cfg.temperature.night}"
-            "-T ${toString cfg.temperature.day}"
-            "-g ${cfg.gamma}"
-          ];
-        in "${cfg.package}/bin/wlsunset ${concatStringsSep " " args}";
+          args = cli.toGNUCommandLineShell { } {
+            t = cfg.temperature.night;
+            T = cfg.temperature.day;
+            g = cfg.gamma;
+            l = cfg.latitude;
+            L = cfg.longitude;
+            S = cfg.sunrise;
+            s = cfg.sunset;
+            o = cfg.output;
+          };
+        in "${cfg.package}/bin/wlsunset ${args}";
       };
 
       Install = { WantedBy = [ cfg.systemdTarget ]; };
--- a/tests/modules/services/wlsunset/wlsunset-service-expected.service
+++ b/tests/modules/services/wlsunset/wlsunset-service-expected.service
@@ -2,7 +2,7 @@
 WantedBy=test.target
 
 [Service]
-ExecStart=@wlsunset@/bin/wlsunset -l 12.3 -L 128.8 -t 3500 -T 6000 -g 0.6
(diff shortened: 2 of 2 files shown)


modules/services/wlsunset.nix
```diff
@@ -91,7 +124,11 @@ in {
             "-L ${cfg.longitude}"
             "-t ${toString cfg.temperature.night}"
             "-T ${toString cfg.temperature.day}"
+            "-S ${cfg.time.sunrise}"
+            "-s ${cfg.time.sunset}"
```
ambroisie, line 128: Isn't this redundant with `latitute`/`longitude`?

I understand wanting to use fixed times rather than location-based schedules, but then I think all the options should be `type = nullOr str` and only given to `wlsunset` _if_ they're defined.

ambroisie (changes requested):

modules/services/wlsunset.nix
```diff
@@ -63,6 +63,39 @@ in {
...
+        Name of output to use, by default all outputs are used.
+      '';
+    };
+
+    time = {
+      duration = mkOption {
+        type = types.int;
+        description = ''
+          The duration in seconds.
+          e.g 1800
```
ambroisie, line 78: Use `example = 1800` rather than adding an example in the description.

modules/services/wlsunset.nix
Daru-san: I like that idea, it would be way better to do that instead actually.
Something like this: 
```nix
sunset = mkOption {
  description = ''
     The time when the sun sets (in 24 hour format).
  '';
};
```

modules/services/wlsunset.nix
Daru-san: I'll definitely do that, I forgot that existed. Thank you for pointing that out

modules/services/wlsunset.nix
Daru-san: Sorry for the late reply. Yeah I'm trying to work out how to do that without making it very complicated, I've been working on doing it like this.
Although I am stuck in terms of making sure either `latitude and longitude` or `sunrise and sunset` are present, as I believe that that will add a lot of extra code to the module.
```nix
{
  systemd.user.services.wlsunset.Service = {
     ExecStart = let
        args = [
           "-t ${toString cfg.temperature.night}"
           "-T ${toString cfg.temperature.day}"
           "-g ${cfg.gamma}"
           ((mkIf cfg.latitude != null) "-l ${cfg.latitude}")
           ((mkIf cfg.longitude != null) "-L ${cfg.longitude}")
           ((mkIf cfg.time.sunrise != null) "-S ${cfg.time.sunrise}")
           ((mkIf cfg.time.sunset != null) "-s ${cfg.time.sunset}")
           ((mkIf cfg.time.duration != null) "-d ${toString cfg.time.duration}")
           ((mkIf cfg.output != null) "-o ${cfg.output}")
        ];
       in "${cfg.package}/bin/wlsunset ${concatStringsSep " " args}";
   };
}
```

rycee: I think you may be able to use `cli.toGNUCommandLine`. Something like this may work (untested):
``` nix
# The definition for the systemd service
{
  systemd.user.services.wlsunset.Service = {
     ExecStart = let
        args = cli.toGNUCommandLine {
           t = cfg.temperature.night;
           T = cfg.temperature.day;
           g = cfg.gamma;
           l = cfg.latitude;
           L = cfg.longitude;
           S = cfg.time.sunrise;
           s = cfg.time.sunset;
           d = cfg.time.duration;
           o = cfg.output;
       };
       in "${cfg.package}/bin/wlsunset ${args}";
   };
}
```

Daru-san: Thank you for the recommendation. With a small tweak to it, it works perfectly
```nix
# The definition for the systemd service
{
  systemd.user.services.wlsunset.Service = {
     ExecStart = let
        args = cli.toGNUCommandLineShell {} {
           t = cfg.temperature.night;
           T = cfg.temperature.day;
           g = cfg.gamma;
           l = cfg.latitude;
           L = cfg.longitude;
           S = cfg.sunrise;
           s = cfg.sunset;
           o = cfg.output;
       };
       in "${cfg.package}/bin/wlsunset ${args}";
   };
}
```

Now it's just finding a good way to make sure that either `latitude and longitude` or `sunset and sunrise` have values.

Daru-san: Now it's just finding a good way to make sure that either `latitude and longitude` or `sunset and sunrise` have values, otherwise the service won't work. Something like 
```nix
mkIf (cfg.latitude & cfg.longitude) or (cfg.sunset & cfg.sunrise) # of course this is a terrible example
```

matrss: > Now it's just finding a good way to make sure that either `latitude and longitude` or `sunset and sunrise` have values, otherwise the service won't work. Something like
> 
> ```nix
> mkIf (cfg.latitude & cfg.longitude) or (cfg.sunset & cfg.sunrise) # for example
> ```

Something like this should work, I think: https://github.com/NixOS/nixpkgs/blob/7bb2ccd8cdc44c91edba16c48d2c8f331fb3d856/nixos/modules/services/networking/xray.nix#L72-L77

I.e. an assertion that checks that either one or the other is set, but not both.

(sorry for the late response)

rycee: @Daru-san Will you apply the `cli.toGNUCommandLineShell` change?

Daru-san: I will do that actually, I seemed to have forgotten to do that

Daru-san: @matrss Looks like that works perfectly, I spent a very long time trying to find a way to that that but it seems like this simple additions does the trick
```nix
{
  assertion = (cfg.sunrise != null && cfg.sunset != null) != (cfg.latitude != null && cfg.longitude != null);
  message = "Either `sunset` and `sunrise` or `latitude` and `longitude` have to be set for wlsunset";
}

```

Daru-san: I also wonder if having the options `latitude` and `longitude` and `gamma` free to be either strings, floats or integers since cli.toGNUCommandLineShell converts them all
```nix
      type = with types; nullOr (either float (either str int));
```

matrss: Just FYI, to find that example I searched the NixOS options for the word "either" with the expectation that there must be some module with the same constraint and that word must be somewhere in the options description. So that approach can be a good starting point. Also, just grep'ing through nixpkgs can be very informative.

That said, I think your assertion still has an issue. Namely, it wouldn't catch the case that 3 of the 4 options are set, I think.

Daru-san: I see, I tried experimenting with it to see how far I could go with it and it seems like there is a way, but I feel it's a bit complicated. There might be an easier way, I can try looking through nixpkgs to see what I can find
```nix
 {
   assertion = (cfg.sunrise != null && cfg.sunset != null) != (cfg.latitude != null && cfg.longitude != null)
     && (cfg.sunrise != null || cfg.sunrise != null) != (cfg.latitude != null && cfg.longitude != null);
   message = "Either `sunset` and `sunrise` or `latitude` and `longitude` have to be set for wlsunset";
 }
```

ambroisie: I think you're conflating checking that `sunrise` and `sunset` have to be both `null` or set (same for `latitude`/`longitude`) with the assertion that `sunrise`/`sunset` is incompatible with `latitude`/`longitude`.

In my mind, those are three different assertions, and make the code that much easier to write.

```
 {
   assertion = (cfg.sunrise != null || cfg.sunset != null) != (cfg.latitude != null || cfg.longitude != null);
 }
 {
   assertion = (cfg.sunrise != null) == (cfg.sunset != null);
 }
 {
   assertion = (cfg.latitude != null) == (cfg.longitude != null);
 }
 ```

Comment (Daru-san) [reply modules/services/wlsunset.nix]: That would make it way more readable, I wanted to try to make it one assertion at best but that approach is a lot easier to understand

Comment (matrss) [reply modules/services/wlsunset.nix]: FWIW, I would have probably written something like
```
(cfg.sunrise != null && cfg.sunset != null && cfg.latitude == null && cfg.longitude == null) || (cfg.sunrise == null && cfg.sunset == null && cfg.latitude != null && cfg.longitude != null)
```
to make the two cases that are correct super obvious.

But splitting it into 3 assertions is even better, since you can also set meaningful messages for each of the three error cases.

Comment (Daru-san) [reply modules/services/wlsunset.nix]: I agree, it works the best out of the other options. I will use that approach here

File: modules/services/wlsunset.nix [hunk 3]
```diff
@@ -1,17 +1,19 @@
...
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+with lib; let
+  cfg = config.services.wlsunset;
 in {
-  meta.maintainers = [ hm.maintainers.matrss ];
+  meta.maintainers = [hm.maintainers.matrss];
```
matrss, line 10: The formatting is off. The way it was previously should satisfy the code style of home-manager: https://nix-community.github.io/home-manager/index.xhtml#sec-code-style.

matrss (changes requested):
Other than the formatting this seems good to me now. As noted further up I am not currently able to actually test this though, I just read the code.

matrss:
Oh, also, I think this should be squashed into a single commit, since logically it is a single self-contained change to the module and the other commits are amendments to the first. I am not sure what the exact customs are regarding this in home-manager though.

modules/services/wlsunset.nix
```diff
@@ -20,15 +22,19 @@ in {
     };
 
     latitude = mkOption {
-      type = types.str;
+      type = with types; nullOr (either float int);
```
matrss, line 25: Just noticed: changing this type of the pre-existing options is backwards-incompatible, i.e. this would break existing configurations. As much as float and int make more sense than str, I think str should continue to be allowed for these.

modules/services/wlsunset.nix
Daru-san: I always use the Alejandra formatter so I assumed that you could use any formatter as long as it nix could evaluate it, I will change that

modules/services/wlsunset.nix
Daru-san: I didn't consider backward compatibility originally but I can make that change

rycee:

PR #5036, opened 2024-02-19
zellij: add fish completions

tennox:
### Description

Added completions script as shown in docs
https://zellij.dev/documentation/controlling-zellij-through-cli

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/zellij.nix
+++ b/modules/programs/zellij.nix
@@ -80,6 +80,7 @@ in {
     programs.fish.interactiveShellInit = mkIf cfg.enableFishIntegration
       (mkOrder 200 ''
         eval (${zellijCmd} setup --generate-auto-start fish | string collect)
+        eval (${zellijCmd} setup --generate-completion fish | string collect)
       '');
   };
 }


modules/programs/zellij.nix
```diff
@@ -80,6 +80,7 @@ in {
     programs.fish.interactiveShellInit = mkIf cfg.enableFishIntegration
       (mkOrder 200 ''
         eval (${zellijCmd} setup --generate-auto-start fish | string collect)
+        eval (${zellijCmd} setup --generate-completion fish | string collect)
```
pedorich-n, line 83: This is already handled in `nixpkgs` :thinking: 
https://github.com/NixOS/nixpkgs/blob/7234dbc8f86738c6925d47a705526f69364b9eca/pkgs/tools/misc/zellij/default.nix#L55

Are you missing the completions?

pedorich-n: Hmm... I don't think aliases are installed with `installShellCompletion`.
`installShellCompletion` works by saving shell completions under `share/completion/...` folders and then it gets merged by Home-Manager and added to the ENV. But aliases have to be installed to the shell "directly". Either via `home.shellAliases` or by modifying the `rc` files. So maybe it makes sense to have the eval in HM, to have aliases. 🤔

PR #5037, opened 2024-02-19
jujutsu: Remove superfluous shell completion

jcf:
This issue has come up before, and the recommendation was to remove the shell completion logic from Home Manager. This pull request now does just that.

- https://github.com/martinvonz/jj/pull/2945
- https://github.com/nix-community/home-manager/pull/5016#issuecomment-1947449541
- https://github.com/nix-community/home-manager/pull/5037#pullrequestreview-1888843990

### Description

``` sh
$ for sh (--bash --fish --zsh); jj util completion "$sh" >/dev/null
Warning: `jj util completion --bash` will be removed in a future version, and this will be a hard error
Hint: Use `jj util completion bash` instead
Warning: `jj util completion --fish` will be removed in a future version, and this will be a hard error
Hint: Use `jj util completion fish` instead
Warning: `jj util completion --zsh` will be removed in a future version, and this will be a hard error
Hint: Use `jj util completion zsh` instead
```

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@shikanime

### DIFF
--- a/modules/programs/jujutsu.nix
+++ b/modules/programs/jujutsu.nix
@@ -10,6 +10,12 @@ let
 in {
   meta.maintainers = [ maintainers.shikanime ];
 
+  imports = let
+    mkRemovedShellIntegration = name:
+      mkRemovedOptionModule [ "programs" "jujutsu" "enable${name}Integration" ]
+      "This option is no longer necessary.";
+  in map mkRemovedShellIntegration [ "Bash" "Fish" "Zsh" ];
+
   options.programs.jujutsu = {
     enable =
       mkEnableOption "a Git-compatible DVCS that is both simple and powerful";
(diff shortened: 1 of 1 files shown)


jcf:
I've tested this locally by overriding the Jujutsu program like so:

``` nix
programs.jujutsu.enableZshIntegration = false;
programs.zsh.initExtra = ''
  source <(${pkgs.jujutsu}/bin/jj util completion zsh)
    compdef _jj ${pkgs.jujutsu}/bin/jj
'';
```

ambroisie (changes requested):
Duplicate of #5016, and same issue: this shouldn't even be in Home-Manager.

jcf:
It sounds like the resolution was to [remove this completion](https://github.com/nix-community/home-manager/pull/5016#issuecomment-1947449541) from Home-Manager. Would such a pull request be welcome?

jcf:
I've pushed another commit to remove the shell completion entirely.

modules/programs/jujutsu.nix
```diff
@@ -33,24 +33,6 @@ in {
...
-      type = types.bool;
-      default = true;
-      description = "Whether to enable Zsh integration.";
-    };
-
-    enableFishIntegration = mkOption {
-      type = types.bool;
-      default = true;
-      description = "Whether to enable Fish integration.";
-    };
```
ambroisie, line 53: These should have a `mkRemoveModuleOption` to explain why they were removed.

ambroisie: Also, please squash your commits.

rycee: I've applied these suggestions. Thanks!

rycee:

PR #5038, opened 2024-02-19
hyprland: add device option for per-input device configs

matt1432:
### Description

Since https://github.com/hyprwm/Hyprland/commit/13f6f0b923ff3ec94a3bec886c28b90402ceef91 the syntax for per-input config changed from this:

```
device:razer-naga-pro {
    ...
}
```

to this:

```
device {
    name = razer-naga-pro
    ...
}
```

You can see it more clearly here: https://github.com/hyprwm/Hyprland/commit/5fc0b772c72ce23f3d31fec31fa3688f7ed1afff

This causes issues because in `wayland.windowManager.hyprland.settings` the attributes don't allow a list of attribute sets which is needed now that Hyprland needs multiple blocks with the same name.

This is why I made a separate option `wayland.windowManager.hyprland.device` which does take a list of attribute sets.
The device blocks also requires the `name` key to be first, which is why I did it the way I did.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@fufexan


fufexan:
Imo we shouldn't break backwards-compatibility. I managed to fix this in a different way.
```diff
diff --git a/modules/services/window-managers/hyprland.nix b/modules/services/window-managers/hyprland.nix
index 168bab65..6bd56fd0 100644
--- a/modules/services/window-managers/hyprland.nix
+++ b/modules/services/window-managers/hyprland.nix
@@ -211,20 +211,32 @@ in {
             ${indent}${n} {
             ${toHyprconf attrs (indentLevel + 1)}${indent}}
           '';
-          sections = filterAttrs (n: v: isAttrs v) attrs;
+
+          sections = filterAttrs (n: v: (isAttrs v) && !(hasPrefix "device" n)) attrs;
+
+          specialCategories = filterAttrs (n: v: (isAttrs v) && (hasPrefix "device" n)) attrs;
+
+          mkSpecialCategory = n: attrs: ''
+            ${indent}${n} {
+              name=${attrs.name}
+            ${toHyprconf (filterAttrs (n: _: "name" != n) attrs) (indentLevel + 1)}${indent}}
+          '';
 
           mkFields = generators.toKeyValue {
             listsAsDuplicateKeys = true;
             inherit indent;
           };
           allFields = filterAttrs (n: v: !(isAttrs v)) attrs;
+
           importantFields = filterAttrs (n: _:
             (hasPrefix "$" n) || (hasPrefix "bezier" n)
             || (cfg.sourceFirst && (hasPrefix "source" n))) allFields;
+
           fields = builtins.removeAttrs allFields
             (mapAttrsToList (n: _: n) importantFields);
         in mkFields importantFields
         + concatStringsSep "\n" (mapAttrsToList mkSection sections)
+        + concatStringsSep "\n" (mapAttrsToList mkSpecialCategory specialCategories)
         + mkFields fields;
 
       pluginsToHyprconf = plugins:
diff --git a/tests/modules/services/window-managers/hyprland/simple-config.conf b/tests/modules/services/window-managers/hyprland/simple-config.conf
index 0db599a3..1e91587d 100644
--- a/tests/modules/services/window-managers/hyprland/simple-config.conf
+++ b/tests/modules/services/window-managers/hyprland/simple-config.conf
@@ -35,6 +35,10 @@ plugin {
     dummy=plugin setting
   }
 }
+device {
+  name=some:device
+  enable=true
+}
 bindm=$mod, mouse:272, movewindow
 bindm=$mod, mouse:273, resizewindow
 bindm=$mod ALT, mouse:272, resizewindow
diff --git a/tests/modules/services/window-managers/hyprland/simple-config.nix b/tests/modules/services/window-managers/hyprland/simple-config.nix
index f45bf6b3..e7b11ffe 100644
--- a/tests/modules/services/window-managers/hyprland/simple-config.nix
+++ b/tests/modules/services/window-managers/hyprland/simple-config.nix
@@ -46,6 +46,11 @@
         "$mod ALT, mouse:272, resizewindow"
       ];
 
+      device = {
+        name = "some:device";
+        enable = true;
+      };
+
       plugin = {
         plugin1 = {
           dummy = "plugin setting";
```

matt1432:
Ahh I see how you make it not break backwards compatibility, yeah that does seem like a better approach.

andresilva:
Came here looking for this, I agree with @fufexan about keeping backwards-compatible. Could you open a PR with your approach from the diff above? I think creating a separate option for devices would be more inline with the HM module from https://github.com/hyprland-community/hyprland-nix.

fufexan:
I left the patch here in case @matt1432 wanted to update the PR to use some ideas from it (or use it entirely).

matt1432:
I'm not sure I understand how to specify multiple device blocks using your solution @fufexan

fufexan:
```nix
devices = [
  {
    name = "mouse1234";
    option1 = "value1";
  }
  {
    name = "keyboard_something";
    kb_layout = "colemak";
  }
];
```

Just like any other duplicate key.

matt1432:
this doesn't work for me when I do this:

```nix
  nagaProNames = [
    # Wireless
    "razer-razer-naga-pro"

    # Wired (it always changes)
    "razer-razer-naga-pro-1"
    "razer-naga-pro"
    "razer-naga-pro-1"
    "razer-naga-pro-2"
    "razer-naga-pro-3"
  ];
  nagaConf = name: {
    inherit name;
    sensitivity = 0;
    accel_profile = "flat";
  };
in {
  wayland.windowManager.hyprland = {
    settings = {
      device = map (d: (nagaConf d)) nagaProNames;
```

```console
error: evaluation aborted with the following error message: 'generators.mkValueStringDefault: attrsets not supported: {
         accel_profile = "flat";
         name = "razer-razer-naga-pro";
         sensitivity = 0;
       }'
```

fufexan:
Really? Didn't know that. @rycee would you know how to work around this limitation, or have I messed up somewhere?

matt1432:
I just pushed something that works well on my fork if you wanna take a look at it @fufexan

fufexan:
Does that work in case `device` isn't a list?

matt1432:
Oops it doesn't, let me fix that real quick

rycee:

PR #5040, opened 2024-02-20
lorri service cachedir

nyarly:
- **lorri: systemd allow access to cache directories**

### Description

Lorri fails builds with fetchers if it can't write to .cache/nix;
This PR adds that to ReadWrite

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@gerschtli

Am also adding myself as a maintainer, as I've been doing my best to maintain Lorri itself.

### DIFF
--- a/modules/services/lorri.nix
+++ b/modules/services/lorri.nix
@@ -7,7 +7,7 @@ let
   cfg = config.services.lorri;
 
 in {
-  meta.maintainers = [ maintainers.gerschtli ];
+  meta.maintainers = [ maintainers.gerschtli maintainers.nyarly ];
 
   options.services.lorri = {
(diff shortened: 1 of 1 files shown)


Gerschtli (approved):

modules/lib/maintainers.nix
```diff
@@ -370,6 +370,12 @@
     github = "mainrs";
     githubId = 5113257;
   };
+  nyarly = {
+    name = "Judson";
+    email = "nyarly@gmail.com";
+    github = "nyarly";
+    githubId = 127548;
+  };
```
rycee, line 378: No need to add yourself here since you are already in the Nixpkgs maintainers list.

PR #5041, opened 2024-02-20
zsh: fix broken ZDOTDIR when path contains spaces

jtt9340:
### Description

Zsh configuration breaks when `programs.zsh.dotDir` is set to a path containing spaces due to the spaces not being escaped. This PR fixes that by surrounding the path in quotes in the resulting Zsh startup scripts.

The motivation for this is, on Darwin, setting `xdg.configHome` to be `~/Library/Application Support` to try to follow file system conventions on macOS.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@uvNikita (based on `git blame`)

### DIFF
--- a/modules/programs/zsh.nix
+++ b/modules/programs/zsh.nix
@@ -22,7 +22,7 @@ let
     mapAttrsToList (k: v: ''hash -d ${k}="${v}"'') cfg.dirHashes
   );
 
-  zdotdir = "$HOME/" + cfg.dotDir;
+  zdotdir = "$HOME/" + lib.escapeShellArg cfg.dotDir;
 
   bindkeyCommands = {
     emacs = "bindkey -e";


modules/programs/zsh.nix
```diff
@@ -511,15 +511,15 @@ in
 
     (mkIf (cfg.dotDir != null) {
       home.file."${relToDotDir ".zshenv"}".text = ''
-        export ZDOTDIR=${zdotdir}
+        export ZDOTDIR="${zdotdir}"
```
ambroisie, line 514: `escapeShellArg` rather than quotes? Because if the path contains double quote (`"`) this breaks (unlikely, but might as well try to be _more_ correct).

jtt9340: Ah good point. I had to call `escapeShellArg` on `cfg.dotDir` directly though instead of on the entire `zdotdir` since we don't want to escape `$HOME`.

rycee:

Issue #4822, opened 2023-12-28
bug:  yazi nushell integration is broken (arg passing: "Cannot convert nothing to string")


tukanoidd:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

When running 'ya', I get this error:

```
❯ ya
Error: nu::shell::external_command

  × External command failed
     ╭─[/home/<user>/.config/nushell/config.nu:659:1]
 659 │   let tmp = $"($env.TEMP)(char path_sep)yazi-cwd." + (random chars -l 5)
 660 │   yazi $args --cwd-file $tmp
     ·        ──┬──
     ·          ╰── Cannot convert nothing to a string
 661 │   let cwd = (open $tmp)
     ╰────
  help: All arguments to an external command need to be string-compatible
```

### Maintainer CC

@XYenon 

### System information

```markdown
- system: `"x86_64-linux"`
- host os: `Linux 6.6.7-zen1, NixOS, 24.05 (Uakari), 24.05.20231224.5f64a12`
- multi-user?: `yes`
- sandbox: `yes`
- version: `nix-env (Nix) 2.18.1`
- channels(root): `"nixos"`
- nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`
```

tukanoidd:
Sorry, for some reason github was very buggy when putting maintainers name, i thought it was fine, but just realized it cut off part of it when posting

tukanoidd:
I temporarily decided to fix this locally with a simple if-else statement:
```
def --env ya [args?] {
      	let tmp = $"($env.TEMP)(char path_sep)yazi-cwd." + (random chars -l 5)

        if ($args == null) {
          yazi --cwd-file $tmp
        } else {
          yazi $args --cwd-file $tmp
        }

      	let cwd = (open $tmp)
      	if $cwd != "" and $cwd != $env.PWD {
      		cd $cwd
      	}
      	rm -f $tmp
      }
```
Should I make a PR myself or let @XYenon do it? Sorry, I'm not sure what the rules here are if there is a module maintainer already. 

Also, dk if it's just my setup being borked, but I had to also manually do 
```nix
   nushell.extraEnv = ''
      $env.TEMP = /tmp
    '';
```

[...]

PR #5045, opened 2024-02-22
[Backport 23.11] yazi: shell integration

XYenon:
### Description

Backport https://github.com/nix-community/home-manager/pull/4699 https://github.com/nix-community/home-manager/pull/4746 https://github.com/nix-community/home-manager/pull/4846 to 23.11

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/tests/modules/programs/yazi/bash-integration-enabled.nix
+++ b/tests/modules/programs/yazi/bash-integration-enabled.nix
@@ -3,7 +3,7 @@
 let
   shellIntegration = ''
     function ya() {
-      tmp="$(mktemp -t "yazi-cwd.XXXXX")"
+      local tmp="$(mktemp -t "yazi-cwd.XXXXX")"
       yazi "$@" --cwd-file="$tmp"
       if cwd="$(cat -- "$tmp")" && [ -n "$cwd" ] && [ "$cwd" != "$PWD" ]; then
         cd -- "$cwd"
--- a/tests/modules/programs/yazi/bash-integration-enabled.nix
+++ b/tests/modules/programs/yazi/bash-integration-enabled.nix
@@ -4,7 +4,7 @@ let
   shellIntegration = ''
     function ya() {
       tmp="$(mktemp -t "yazi-cwd.XXXXX")"
-      yazi --cwd-file="$tmp"
+      yazi "$@" --cwd-file="$tmp"
       if cwd="$(cat -- "$tmp")" && [ -n "$cwd" ] && [ "$cwd" != "$PWD" ]; then
         cd -- "$cwd"
       fi
--- a/tests/modules/programs/yazi/zsh-integration-enabled.nix
+++ b/tests/modules/programs/yazi/zsh-integration-enabled.nix
@@ -3,7 +3,7 @@
 let
   shellIntegration = ''
     function ya() {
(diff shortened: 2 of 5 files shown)


tests/modules/programs/yazi/bash-integration-enabled.nix
```diff
@@ -4,7 +4,7 @@ let
   shellIntegration = ''
     function ya() {
       tmp="$(mktemp -t "yazi-cwd.XXXXX")"
```
sxyazi, line 6:
```suggestion
      local tmp="$(mktemp -t "yazi-cwd.XXXXX")"
```

tests/modules/programs/yazi/zsh-integration-enabled.nix
```diff
@@ -4,7 +4,7 @@ let
   shellIntegration = ''
     function ya() {
       tmp="$(mktemp -t "yazi-cwd.XXXXX")"
-      yazi --cwd-file="$tmp"
+      yazi "$@" --cwd-file="$tmp"
```
sxyazi, line 7:
```suggestion
      local yazi "$@" --cwd-file="$tmp"
```

sxyazi:
It's better for the Nushell wrapper to keep consistent with [the documentation](https://yazi-rs.github.io/docs/quick-start) - use `...` to simplify if-else blocks, add the `rm -p` parameter to permanently delete temporary files instead of sending them to the recycle bin:

```nu
def --env ya [...args] {
	let tmp = (mktemp -t "yazi-cwd.XXXXX")
	yazi ...$args --cwd-file $tmp
	let cwd = (open $tmp)
	if $cwd != "" and $cwd != $env.PWD {
		cd $cwd
	}
	rm -fp $tmp
}
```

tests/modules/programs/yazi/zsh-integration-enabled.nix
```diff
@@ -4,7 +4,7 @@ let
   shellIntegration = ''
     function ya() {
       tmp="$(mktemp -t "yazi-cwd.XXXXX")"
```
sxyazi, line 6:
```suggestion
      local tmp="$(mktemp -t "yazi-cwd.XXXXX")"
```

XYenon:
@sxyazi I think we should make these changes on master and then backport them together?

sxyazi:
Ah yep, that makes sense.

musjj:
I think it would be better if you could specify a custom alias name. Maybe something like:
```nix
shellAlias = mkOption {
  type = types.str;
  default = "ya";
  description = "The alias name used when shell integration is enabled.";
};
```
Not sure if there's a more established pattern/convention though.

musjj (commented):
Not super critical, but it would be nice if these functions can trigger shell completion. On my `zsh` config, this is how I did it:
```zsh
compdef _yazi ya
```
But I'm not really sure how you'd achieve it in other shells.

XYenon:
@musjj I think we should backport these commits as-is first, to fix some bugs like https://github.com/nix-community/home-manager/issues/4822. And then add new features to the master branch.

rycee:
Thanks! Merged to master now 🙂 I definitely agree that additional improvements should go through master first.

PR #5046, opened 2024-02-22
session-variables: escape quotes in shell env vars

pcasaretto:
Add support for environment variables with values containing quotes.

### Description

When trying to set EDITOR and VISUAL to the recommended value for emacsclient (`emacsclient -t -a ""`) 
I kept getting truncated values (`emacsclient -t -a`) in the resulting shell environment.
This PR adds support for environment variables with values containing quotes.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/lib/shell.nix
+++ b/modules/lib/shell.nix
@@ -1,8 +1,10 @@
 { lib }:
-
-rec {
+let
+  escapeQuotes =
+    lib.strings.stringAsChars (x: if x == ''"'' then ''\"'' else x);
+in rec {
   # Produces a Bourne shell like variable export statement.
-  export = n: v: ''export ${n}="${toString v}"'';
+  export = n: v: ''export ${n}="${escapeQuotes (toString v)}"'';
 
   # Given an attribute set containing shell variable names and their
   # assignment, this function produces a string containing an export
--- a/modules/lib/zsh.nix
+++ b/modules/lib/zsh.nix
@@ -1,12 +1,14 @@
 { lib }:
-
-rec {
+let
+  escapeQuotes =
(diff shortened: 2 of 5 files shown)


pcasaretto:
What is [tests/modules/home-environment/session-variables-expected.txt](https://github.com/nix-community/home-manager/blob/0e0e9669547e45ea6cca2de4044c1a384fd0fe55/tests/modules/home-environment/session-variables-expected.txt) used for ?

pcasaretto:
Working on the test failure. @stasjok you are most recent committer for the tests, can you give a hand?
I think this has been working because the double quotes are inside that variable assignment default expression.
On my local testing, it seems we could ditch the double quotes and the result would be the same.

```
λ unset XDG_RUNTIME_DIR                                                                                                                                                                                                                                                      
λ echo "${XDG_RUNTIME_DIR:-\"/run/user/$(id -u)\"}"                                                                                                         
"/run/user/501"                                                                                                                        
λ echo "${XDG_RUNTIME_DIR:-/run/user/$(id -u)}"                                                                                                             
/run/user/501
```

stasjok:
> When trying to set EDITOR and VISUAL to the recommended value for emacsclient (`emacsclient -t -a ""`)
> I kept getting truncated values (`emacsclient -t -a`) in the resulting shell environment.

I think you should better use (escape quotes manually):

```
EDITOR = ''emacsclient -t -a \"\"'';
```

I don't think replacing `"` with `\"` is the right way. What if I want quotes as is?

pcasaretto:
I swear I had tried that. Nevermind then.
Sorry for wasting your time.

PR #5053, opened 2024-02-24
poetry: init module

mirkolenz:
### Description

Create new module for poetry. I added the corresponding tests based on those used by `tealdeer`, but strangely the test `poetry-custom-config` is failing even though it works if I include the new module in my own nixos configuration. That is also why this PR is marked as draft at the moment. I would be more than happy for suggestions as to why the test is failing.

Edit: I identified the issue: I forgot to include the new module in the file `modules/modules.nix`, so it was not picked up during the tests. I marked the PR as ready to review.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/poetry.nix
+++ b/modules/programs/poetry.nix
@@ -1,0 +1,55 @@
+{ pkgs, config, lib, ... }:
+
+let
+
+  inherit (lib) mkEnableOption mkPackageOption mkOption literalExpression;
+
+  tomlFormat = pkgs.formats.toml { };
+
+  configDir = if pkgs.stdenv.isDarwin then
+    "Library/Application Support"
+  else
+    config.xdg.configHome;
+
+  cfg = config.programs.poetry;
+
+in {
+  meta.maintainers = with lib.maintainers; [ mirkolenz ];
+
+  options.programs.poetry = {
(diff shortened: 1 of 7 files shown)


modules/programs/poetry.nix
```diff
@@ -0,0 +1,48 @@
...
+  options.programs.poetry = with lib; {
+    enable = mkEnableOption "poetry";
+    package = mkOption {
+      type = types.package;
+      default = pkgs.poetry;
+      defaultText = literalExpression "pkgs.poetry";
+      description = "The poetry package to use (e.g., with custom plugins).";
+      example = literalExpression
+        "pkgs.poetry.withPlugins (ps: with ps; [ poetry-plugin-up ])";
+    };
```
ambroisie, line 20: Please use `mkPackageOption`.

mirkolenz: Thanks for the suggestion, I was not aware this function exists. I updated the option accordingly.

mirkolenz:
Is there anything left to do in order to proceed with the PR? Thanks in advance!

rycee:

mirkolenz:
Great, thank you!

PR #5055, opened 2024-02-25
river: fix systemd activation

johnae:
### Description
exec does not do any shell parsing and does not understand the && which is how the extraCommands are added after dbus activation. There seems to be no clear need for exec here anyway so just remove it and allow shell parsing.

The error you'll get when there are extraCommands (and there are by default) looks like this:

```sh
bash: /nix/store/3n8s5m00h7qqvbmahhr4fxskkbz85z03-dbus-1.14.10/bin/dbus-update-activation-environment --systemd DISPLAY WAYLAND_DISPLAY XDG_CURRENT_DESKTOP NIXOS_OZONE_WL XCURSOR_THEME XCURSOR_SIZE && systemctl --user stop river-session.target && systemctl --user start river-session.target: No such file or directory
```

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@GaetanLepage

### DIFF
--- a/modules/services/window-managers/river.nix
+++ b/modules/services/window-managers/river.nix
@@ -9,7 +9,7 @@ let
   extraCommands = builtins.concatStringsSep " "
     (map (f: "&& ${f}") cfg.systemd.extraCommands);
   systemdActivation = ''
-    exec "${pkgs.dbus}/bin/dbus-update-activation-environment --systemd ${variables} ${extraCommands}"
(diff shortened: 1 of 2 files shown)


GaetanLepage (commented):
Diff LGM.
Thanks for the fix !

GaetanLepage (approved):

GaetanLepage:
cc @rycee for review

rycee:
Thanks! Merged to master now.

Issue #4989, opened 2024-02-02
bug: firefox creates a new `containers.json` on quite that is not overwritten


slotThe:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

When configuring firefox to use container tabs, e.g.,

``` nix
programs.firfeox = {
  …
  profiles.persistent = {
    name = "persistent";
    id = 1;
    containers = {
      "foo" = {
        id    = 1;
        color = "purple";
        icon  = "chill";
      };
      …
    };
  };
}
```

Firefox does respect the containers, but it upon exit it *sometimes* seems to create the default `containers.json` file in its place.

``` console
$ cat /home/slot/.mozilla/firefox/persistent/containers.json
{"version":5,"lastUserContextId":5,"identities":[{"icon":"fingerprint","color":"blue","l10nId":"user-context-personal","public":true,"userContextId":1},{"icon":"briefcase","color":"orange","l10nId":"user-context-work","public":true,"userContextId":2},{"icon":"dollar","color":"green","l10nId":"user-context-banking","public":true,"userContextId":3},{"icon":"cart","color":"pink","l10nId":"user-context-shopping","public":true,"userContextId":4},{"public":false,"icon":"","color":"","name":"userContextIdInternal.thumbnail","accessKey":"","userContextId":5},{"userContextId":4294967295,"public":false,"icon":"","color":"","name":"userContextIdInternal.webextStorageLocal","accessKey":""}]}
```

Then on switching to a new profile I'm greeted with things like

    Existing file '/home/slot/.mozilla/firefox/persistent/containers.json' is in the way of '/nix/store/pzk1y5rz7l7r8qwzkvdwswlffsnsclp0-home-manager-files/.mozilla/firefox/persistent/containers.json'

I will try to bisect this on the weekend, as this was working until very recently

### Maintainer CC

@rycee @kira-bruneau 

### System information

```markdown
console
$ nix-shell -p nix-info --run "nix-info -m"
 - system: `"x86_64-linux"`
 - host os: `Linux 6.1.75, NixOS, 24.05 (Uakari), 24.05.20240130.222c194`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.18.1`
 - nixpkgs: `/nix/store/v9ndcychjpdnnxqdhkwj60brx6d0xaz9-source`
```
```

[...]

PR #5057, opened 2024-02-25
firefox: Add `containersForce`

slotThe:
### Description

Firefox, upon exit, creates the default containers.json file in place of the one that home-manager created. This leads to errors when switching to a new profile, as home-manager is careful with overwriting existing files. The added option toggles that behaviour.

Closes: https://github.com/nix-community/home-manager/issues/4989

---

I guess it would be best if this was inlined into the existing `containers` field, in a backwards compatible way; e.g.,

``` nix
containers = {
  force = true;

  "name" = {
    id    = 1;
    color = "purple";
  };
```

but I haven't managed to make that work.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee @kira-bruneau

### DIFF
--- a/modules/programs/firefox.nix
+++ b/modules/programs/firefox.nix
@@ -573,6 +573,17 @@ in {
               };
             };
 
+            containersForce = mkOption {
+              type = types.bool;
+              default = false;
+              description = ''
+                Whether to force replace the existing containers
+                configuration. This is recommended since Firefox will
+                replace the symlink on every launch, but note that you'll
+                lose any existing configuration by enabling this.
+              '';
+            };
+
             containers = mkOption {
               type = types.attrsOf (types.submodule ({ name, ... }: {
                 options = {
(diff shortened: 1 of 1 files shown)


DamienCassou (approved):
Looks good to me.

DamienCassou:
> I guess it would be best if this was inlined into the existing containers field, in a backwards compatible way

I agree a `profile.<name>.containers.force` would be better but only if container definitions where not directly within `profile.<name>.containers`. Mixing both options and containers in the same attribute set is just confusing I think.

teto:
since I've enabled this, firefox crashes on nixos-rebuild (at least the one in focus). Does anyone else see that ?

kanopeld:
working great for me, thanks! when to expect it in the master?

DamienCassou:
Still using it and didn't notice anything.

rycee:

PR #5059, opened 2024-02-26
hyprpaper: add module

### LINKED ISSUE #4632: Hyprpaper support
It would be nice to have hyprpaper support.

Links:
- [GitHub](https://github.com/hyprwm/hyprpaper)
- [Nixpkgs](https://search.nixos.org/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=hyprpaper)

silmarp:
### Description

Closes #4632

Adds the module ``programs.hyprpaper``
### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/hyprpaper.nix
```diff
@@ -0,0 +1,108 @@
...
+      enable = mkEnableOption "hyprpaper";
+
+      package = mkOption {
+        type = types.package;
+        default = pkgs.hyprpaper;
+        defaultText = literalExpression "pkgs.hyprpaper";
+        description = "The package to use for the hyprpaper binary.";
+      };
+
+      systemd = {
```
rycee, line 20: I think this is best put in a separate module `services.hyprpaper`. We have a few such modules, e.g., emacs and have made such a separation there.

modules/programs/hyprpaper.nix
```diff
@@ -0,0 +1,108 @@
...
+  config = {
+    home.packages = mkIf cfg.enable [ cfg.package ];
+
+    systemd.user.services.hyprpaper = mkIf cfg.systemd.enable {
+      Unit = {
+        Description = "autostart service for Hyprpaper";
+        Documentation = "https://github.com/hyprwm/hyprpaper";
+        BindsTo = [ "graphical-session.target" ];
+        After = [ "graphical-session-pre.target" ];
+      };
```
rycee, line 75: I guess this should restart when the configuration changes, if so you need something like

```suggestion
        X-Restart-Triggers = [ config."hypr/hyprpaper.conf".source ];
      };
```

modules/programs/hyprpaper.nix
```diff
@@ -0,0 +1,108 @@
...
+
+      Service = {
+        ExecStart = "${pkgs.hyprpaper}/bin/hyprpaper";
```
rycee, line 78: Perhaps something like
```suggestion
        ExecStart = "${getExe cfg.package}";
```
?

modules/programs/hyprpaper.nix
```diff
@@ -0,0 +1,108 @@
...
+        default = pkgs.hyprpaper;
+        defaultText = literalExpression "pkgs.hyprpaper";
+        description = "The package to use for the hyprpaper binary.";
+      };
+
+      systemd = {
+        enable =
+          mkEnableOption "Whether to enable hyprpaper systemd integration.";
+
+        target = mkOption {
```
rycee, line 24: This seems unnecessary. Any session that could run this service should already bind to `graphical-session.target`.

modules/programs/hyprpaper.nix
```diff
@@ -0,0 +1,108 @@
...
+          type = types.str;
+          default = "graphical-session.target";
+          example = "hyprland-session.target";
+          description = ''
+            The systemd target that will automatically start the Hyprpaper service.
+          '';
+        };
+      };
+
+      settings = with types; {
```
rycee, line 34: This combination of `settings` and `extraConfig` seems a bit unconventional.

Perhaps make settings a [freeform module](https://nixos.org/manual/nixos/stable/#sec-freeform-modules) with `preload` and `wallpapers` being explicitly defined. Then there would be no need for an `extraConfig` option.

PR #5068, opened 2024-02-29
neomutt: adding unmailboxes option

randomn4me:
### Description

Adding unmailboxes option to neomutt, which adds it to every account.email with neomutt enabled.

Docs: https://www.mutt.org/doc/manual/#mailboxes

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/neomutt.nix
```diff
@@ -397,6 +402,12 @@ in {
           default = true;
         };
 
+      unmailboxes = mkOption {
+        type = types.bool;
+        default = false;
+        description = "Set 'unmailboxes *' at the start of account configs";
```
rycee, line 408:
```suggestion
        description = "Set `unmailboxes *` at the start of account configs.";
```

modules/programs/neomutt.nix
```diff
@@ -397,6 +402,12 @@ in {
           default = true;
         };
 
+      unmailboxes = mkOption {
```
rycee, line 405: Would this name be easily understood by a neomutt user? I wonder if perhaps a more descriptive option name could be used?

Or at least expand the option description with a sentence on what the actual effect the option has.

rycee:
Oh, and about the test error. It should be resolved if you rebase your branch on the latest master.

modules/programs/neomutt.nix
randomn4me: I will add more information to this option as I learned about this option only after scouring the mutt manual :+1:

randomn4me:
Updaten everything and it should now work :+1:

rycee:

PR #5072, opened 2024-03-02
fcitx5: fix reference to fcitx5-with-addons

rycee:
### Description

Should hopefully fix the tests. I guess there may be more work to fully support qt6 but I'll leave that for somebody else.

CC recent committers @thiagokokada @sefidel

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


thiagokokada (approved):

sefidel (approved):
LGTM

Issue #4675, opened 2023-11-19
bug: programs.khal fails to build because of null value


gvolpe:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

The simplest configuration enabling `khal` results in the following build failure:

```console
       … while calling 'genCalendarStr'

         at /nix/store/6zmjz22khxi92dh43w1dp2mjadysjjy8-source/modules/programs/khal.nix:27:26:

           26|
           27|   genCalendarStr = name: value:
             |                          ^
           28|     concatStringsSep "\n" ([

       error: value is null while a set was expected
```

Here's a minimal config that builds to fail with the latest Home Manager and nixpkgs from nixos-unstable:

```nix
{ ... }:

{
  accounts.calendar.accounts.gvolpe = {
    primary = true;
    khal = {
      enable = true;
      color = "light green";
      glob = "*";
      priority = 10;
      readOnly = false;
      type = "discover";
    };
  };

  programs.khal = {
    enable = true;
    locale = { }; 
  };
}
```

Here's also a PR I made to my personal config, resulting in the same build failure on CI: https://github.com/gvolpe/nix-config/pull/246

Not sure if I'm missing something, but I feel we could do better than failing on a `null` value (e.g. using assertions).

### Maintainer CC

@Janik-Haag @DanielSiepmann @antonmosich @teto @arjan-s 

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.6.1, NixOS, 23.11 (Tapir)`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.18.1`
 - channels(root): `"nixos-22.05pre343321.78cd22c1b86"`
 - channels(gvolpe): `""`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`
```

Janik-Haag:
I can try to debug this some time next week. Just looking over the code there seems to be some filter issue with an attribute that has empty values: https://github.com/nix-community/home-manager/blob/master/modules/programs/khal.nix#L185

gvolpe:
Thansk @Janik-Haag ! Yes, and that seems to resolve to the following values, which are set in my config (the calendar bit):

```nix
  khalCalendarAccounts =
    filterAttrs (_: a: a.khal.enable) config.accounts.calendar.accounts;

[...]

PR #5073, opened 2024-03-02
Simplify calendar setup

7FM:
### Description

This PR addresses all problems (mentioned in https://github.com/nix-community/home-manager/issues/4675) that multiple users had while trying to setup the calendar submodule with khal (and vdirsyncer).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/accounts/calendar.nix
+++ b/modules/accounts/calendar.nix
@@ -18,12 +18,13 @@ let
 
         type = mkOption {
           type = types.enum [ "filesystem" "singlefile" ];
+          default = "filesystem";
           description = "The type of the storage.";
         };
 
         fileExt = mkOption {
           type = types.nullOr types.str;
-          default = null;
+          default = ".ics";
           description = "The file extension to use.";
         };
 
@@ -108,7 +109,7 @@ let
 
       local = mkOption {
         type = localModule name;
-        default = {};
+        default = { };
         description = ''
           Local configuration for the calendar.
         '';
@@ -130,6 +131,7 @@ in {
   options.accounts.calendar = {
     basePath = mkOption {
       type = types.str;
+      example = ".calendar";
       apply = p:
         if hasPrefix "/" p then p else "${config.home.homeDirectory}/${p}";
       description = ''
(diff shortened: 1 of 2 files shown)


modules/programs/khal-calendar-accounts.nix
```diff
@@ -5,8 +5,8 @@ with lib;
 {
   options.khal = {
     type = mkOption {
-      type = types.nullOr (types.enum [ "calendar" "discover" ]);
-      default = null;
+      type = types.enum [ "calendar" "discover" ];
+      default = "calendar";
```
teto, line 9: isn't discover more robust ? aka supports more calendars/includes the "calendar" scenario

modules/accounts/calendar.nix
```diff
@@ -129,6 +131,7 @@ in {
   options.accounts.calendar = {
     basePath = mkOption {
       type = types.str;
+      default = ".calendar";
```
teto, line 134: Not sure I like this, I would rather have the user know where the files are. Maybe move it to "example"

modules/programs/khal-calendar-accounts.nix
7FM: Not really with `discover` it searches for subfolder structure in `local.path` (as in this [example](https://lostpackets.de/khal/configure.html#exemplary-discover-usage)). Without a subfolder structure no calendar is found (I haven't tested if the calendar entries in the root folder `local.path` would even be registered).
One user had it set to `discover` at first and had problems because of that: https://github.com/nix-community/home-manager/issues/4675#issuecomment-1968611460

So I think "calendar" is the more common scenario, at least that's what I am using for two completely different CalDav calendar.

modules/accounts/calendar.nix
7FM: Agreed, the error message without an default value should be good enough as a hint to lookup configuration options: `error: The option 'accounts.calendar.basePath' is used but not defined.`

7FM:
I think the tests fail due to: https://github.com/nix-community/home-manager/pull/5072
I will rebase to the latest changes

PR #5074, opened 2024-03-02
waybar: accept null as package

thrix:
### Description

I use `home-manager` in Fedora Silverblue toolbox to configure my home.
I do not need waybar package installed, it is installed by my OS.

Seem this could be nicely achieved by:

    package = null;

Docs currently say:

    Waybar package to use. Set to `null` to install the default package.

What is a bit misleading, you cannot set it to really to a `null` value, you can omit it.

This MR provides a way to provide a null value to skip the package installation.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/waybar.nix
+++ b/modules/programs/waybar.nix
@@ -148,11 +148,11 @@ in {
     enable = mkEnableOption "Waybar";
 
     package = mkOption {
-      type = package;
+      type = nullOr package;
       default = pkgs.waybar;
       defaultText = literalExpression "pkgs.waybar";
       description = ''
-        Waybar package to use. Set to `null` to use the default package.
+        Waybar package to use. Set to `null` to skip the package installation.
       '';
     };
 
@@ -282,7 +282,7 @@ in {
         })
       ];
 
-      home.packages = [ cfg.package ];
+      home.packages = lib.optionals (cfg.package != null) [ cfg.package ];
 
       xdg.configFile."waybar/config" = mkIf (settings != [ ]) {
         source = configSource;
--- a/tests/modules/programs/waybar/package-null.nix
+++ b/tests/modules/programs/waybar/package-null.nix
@@ -1,0 +1,25 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+{
+  config = {
+    home.stateVersion = "23.11";
+
+    programs.waybar = {
+      package = null;
+      enable = true;
+    };
+
+    test.stubs.waybar = {
+      buildScript = "mkdir -p $out/bin; touch $out/bin/waybar";
+      outPath = null;
+    };
+
+    nmt.script = ''
+      assertPathNotExists home-files/.config/waybar/style.css
+      assertPathNotExists home-files/.config/waybar/config
+      assertPathNotExists home-path/bin/waybar
+    '';
+  };
+}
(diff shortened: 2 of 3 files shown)


modules/programs/waybar.nix
```diff
@@ -282,7 +282,7 @@ in {
         })
       ];
 
-      home.packages = [ cfg.package ];
+      home.packages = if cfg.package != null then [ cfg.package ] else [ ];
```
teto, line 285: more elegant with `lib.optionals`

thrix: @teto thank you, I hope I got it right, new to nix :)

tests/modules/programs/waybar/package-null.nix
```diff
@@ -0,0 +1,19 @@
...
+  config = {
+    home.stateVersion = "23.11";
+
+    programs.waybar = {
+      package = null;
+      enable = true;
+    };
+
+    nmt.script = ''
+      assertPathNotExists home-files/.config/waybar/style.css
```
teto, line 15: hum shouldn't it check that the executable "wayland" is not available instead ?

thrix: @teto I agree, but I need some help :) I was looking at other tests and did not find if there is already a test does this for other programs. This test would anyway fail on the old code, as `null` was not accepted at all.

Any hints please how to test that waybar package would not be installed?

thrix: @rycee hi, maybe some clues from you, I am a bit new to home-manager and nix :(

teto: just checking `which waybar` return code should be ok ?

thrix: @teto I also had to define the package stub to actually do something :)

It should be resolved, please check!

thrix:
kindly poking for merge, I hope I addressed all the concerns. Now I need to run against my fork :(

rycee:
Sorry, I'm somewhat worried about these `nullOr` package options. The idea is to have configurations be reproducible and bringing in the package from elsewhere goes against that. The traditional way to get around this (if you absolutely insist) is to set the package option to `pkgs.emptyDirectory` and I sort of like that since it is somewhat cumbersome and ugly, which hopefully discourages people from using it 😉

So, basically have you tried `programs.waybar.package = pkgs.emptyDirectory`, if that works well for you then I would prefer that over merging this.

thrix:
> Sorry, I'm somewhat worried about these `nullOr` package options. The idea is to have configurations be reproducible and bringing in the package from elsewhere goes against that. The traditional way to get around this (if you absolutely insist) is to set the package option to `pkgs.emptyDirectory` and I sort of like that since it is somewhat cumbersome and ugly, which hopefully discourages people from using it 😉
> 
> So, basically have you tried `programs.waybar.package = pkgs.emptyDirectory`, if that works well for you then I would prefer that over merging this.

Thank you, that works for me also, so I can drop part of this MR.

Still, I would like to still fix the help if you do not mind:

```
        Waybar package to use. Set to `null` to use the default package.
```

As it is misleading, as a new user I am unable to recognize I should use `pkgs.emptyDirectory` and rather document it explicitely.

And keep the tests to verify.

WDYT @rycee ?

thrix:
@rycee so after more testing, it does not work as well as with my patch, namely I get issues like:

{code}
/tmp/home-manager-build.mEFT4G3Gie/generation/activate: line 391: /nix/store/4d0ix5djms3n2jnjdc58l916cwack1rp-empty-directory/bin/swaymsg: No such file or directory
{code}

I did not see that with my patch :(

thrix:
I decided the proposed solution is enough for me, I can live with swaymsg unavailable for the empty package.

Thanks for you feedback.

PR #5076, opened 2024-03-03
vim,neovim: set EDITOR to full binary path when `defaultEditor` set for vim and nvim

zzantares:
### Description

Prior to this change the `EDITOR` variable was set only to the name of the
program that should be looked up in `PATH` to use as the editor, but this will
only work for user-level stuff, for stuff like `sudo -e` it won't work as the
`PATH` will not contain the location of such programs. This makes it so that
the when the `defaultEditor` variable is set to `true` then `EDITOR` will
point to the full binary location of the `vim` and `nvim` packages accordingly.

This is similar to how we set `defaultEditor` in the Emacs module.

Related #3496.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@ncfavier @teto

### DIFF
--- a/modules/programs/neovim.nix
+++ b/modules/programs/neovim.nix
@@ -390,7 +390,10 @@ in {
 
     home.packages = [ cfg.finalPackage ];
 
-    home.sessionVariables = mkIf cfg.defaultEditor { EDITOR = "nvim"; };
+    home.sessionVariables = mkIf cfg.defaultEditor {
+      EDITOR = getBin
+        (pkgs.writeShellScript "editor" "exec ${getBin cfg.package}/bin/nvim");
+    };
 
     xdg.configFile =
       let hasLuaConfig = hasAttr "lua" config.programs.neovim.generatedConfigs;
--- a/modules/programs/vim.nix
+++ b/modules/programs/vim.nix
@@ -174,7 +174,10 @@ in {
 
     home.packages = [ cfg.package ];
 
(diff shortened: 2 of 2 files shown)


musjj:
Is there a reason not to just set it as:
```nix
EDITOR = lib.getExe cfg.package;
```

zzantares:
I'm not familiar with all the library functions available, I followed example of what it's done in the [Emacs module](https://github.com/nix-community/home-manager/blob/d579633ff9915a8f4058d5c439281097e92380a8/modules/services/emacs.nix#L187-L190), but if what you suggest is equivalent and preferred I can do the change.

modules/programs/neovim.nix
```diff
@@ -390,7 +390,10 @@ in {
 
     home.packages = [ cfg.finalPackage ];
 
-    home.sessionVariables = mkIf cfg.defaultEditor { EDITOR = "nvim"; };
+    home.sessionVariables = mkIf cfg.defaultEditor {
+      EDITOR = getBin
+        (pkgs.writeShellScript "editor" "exec ${getBin cfg.package}/bin/nvim");
```
KokaKiwi, line 395: Shouldn't the target package be `cfg.finalPackage`?

ambroisie:
Another issue is this will break old terminal sessions after a switch + GC.

teto:
passing the full path seems overkill. Can't you use `sudo -E` instead ?

teto:
status quo looks better for now

PR #5077, opened 2024-03-03
thunderbird: change settings type to json

MatthewCash:
### Description

This sets Thunderbird's `settings` option to use a JSON type, just like the Firefox module, so that nested sets can be used.

Tests failed due to something gpg-related before this change.

It might also be worthwhile to extract out Firefox and Thunderbird's modules to some generic Mozilla module so that these common options can be shared 🤷.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@d-dervishi 
@jkarlson

### DIFF
--- a/modules/programs/thunderbird.nix
+++ b/modules/programs/thunderbird.nix
@@ -7,6 +7,11 @@ let
 
   cfg = config.programs.thunderbird;
 
+  thunderbirdJson = types.attrsOf (pkgs.formats.json { }).type // {
+    description =
+      "Thunderbird preference (int, bool, string, and also attrs, list, float as a JSON string)";
+  };
+
   enabledAccounts = attrValues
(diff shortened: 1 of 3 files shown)


modules/programs/thunderbird.nix
```diff
@@ -7,6 +7,8 @@ let
 
   cfg = config.programs.thunderbird;
 
+  jsonFormat = pkgs.formats.json { };
```
teto, line 10: this could have been `thunderbirdJson` including the description to avoid the repetition.

teto:
could you give an example of configuration that requires nested attributes and even better a test with it ?

modules/programs/thunderbird.nix
MatthewCash: Done

MatthewCash:
> could you give an example of configuration that requires nested attributes and even better a test with it ?

Done.

rycee:

Issue #3516, opened 2022-12-23
bug: Activation fails on Darwin due to missing `readlink` `-m` option


ereslibre:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Trying to activate my Home Manager, I'm seeing the following error:

```
...
copying path '/nix/store/59xjikhmv1vzcki7143a6ir5cpkzgm91-home-manager-generation' from 'https://workstations.cachix.org/'...
readlink: illegal option -- m
usage: readlink [-fn] [file ...]
usage: dirname string [...]
Starting Home Manager activation
/nix/store/59xjikhmv1vzcki7143a6ir5cpkzgm91-home-manager-generation/bin/home-manager-generation: line 68: nix-build: command not found
...
```

The corresponding CI run [can be found here](https://github.com/ereslibre/dotfiles/actions/runs/3763326552/jobs/6396768328#step:5:865).

I believe the issue boils down to:

https://github.com/nix-community/home-manager/blob/886675991b643b701a33f533443db165c70692d1/modules/home-environment.nix#L691

Given that on MacOS `readlink` has no `-m` option.

### System information

```markdown
- system: `"x86_64-darwin"`
- host os: `Darwin 22.2.0, macOS 10.16`
- multi-user?: `yes`
- sandbox: `no`
- version: `nix-env (Nix) 2.12.0`
- channels(root): `"nixpkgs"`
- nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixpkgs`
```

rycee:
Thanks for the report, should be fixed in master now.

ereslibre:
Thanks to you @rycee!

bestlem:
This seems not to be complete.

My generated activate includes bare readlinks

```
setupLaunchAgents() {
  local oldDir newDir dstDir domain err
  oldDir=""
  err=0
  if [[ -n "${oldGenPath:-}" ]]; then
    oldDir="$(readlink -m "$oldGenPath/LaunchAgents")" || err=$?
    if (( err )); then
      oldDir=""
    fi
  fi
...
```
and
```
[...]

PR #11, opened 2017-05-15
Mac issues with coreutils

rycee:
Hey, I'm trying to get some Nix basics working on MacOS, including home-manager. Turns out there are a few assumptions built-in to home-manager that break on Mac, but they should be fixable.

  - The major issue is that we can't assume the GNU versions of commands like `readlink`, `ln`, `mkdir`, etc. are the default ones. The GNU coreutils as packaged by Nix does compile fine on MacOS, so it seems like substituting in `${pkgs.coreutils}/bin/readlink` etc. will fix that just fine. I'll attach a patch to illustrate. You might know of a better way to ensure that a coreutils environment is in the PATH before executing all this stuff. (But importantly, we can't assume that coreutils is in the user's path... he/she may prefer to access the BSD user-space commands by default.)

  - Next, I hit a problem linking into the gcroot, because `/nix/var/nix/gcroots/per-user/$USER` did not exist. On a Mac installation, all of `/nix` is owned by the user, not by `root`, so it didn't install a `per-user` subdir. I did a `mkdir -p` to create it anyway, and then it was fine, so we could add that to the script, assuming we're careful to ensure the same command works fine on NixOS.

  - Finally, we're almost there, but it dies `Activating reloadSystemD` because `systemctl` is not found. I'm pretty sure I'm not enabling any systemd services in `home.nix` on the Mac, but it does a `daemon-reload` anyway. We'd need to be able to skip that.

PR #5091, opened 2024-03-05
home-manager: use coreutils for readlink

andreykaipov:
### Description

I'm using HM on macOS and ran into some issues:

```
Activating setupLaunchAgents
readlink: illegal option -- m
usage: readlink [-fn] [file ...]
readlink: illegal option -- m
usage: readlink [-fn] [file ...]
```

Similar issue to https://github.com/nix-community/home-manager/issues/3516 which was fixed in https://github.com/nix-community/home-manager/pull/3517, but there were a few more readlink commands in the existing Nix code. 

Also related: https://github.com/nix-community/home-manager/pull/11

There might be more tools conflict on macOS that need to be replaced with GNU equivalents from coreutils, but readlink was what I ran into.

### Checklist

  - I seem to get the same build failure when running the tests against my branch and against master, so I'd say they pass. 😅 

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/files.nix
+++ b/modules/files.nix
@@ -84,7 +84,7 @@ in
 
           # A symbolic link whose target path matches this pattern will be
           # considered part of a Home Manager generation.
-          homeFilePattern="$(readlink -e ${escapeShellArg builtins.storeDir})/*-home-manager-files/*"
+          homeFilePattern="$(${pkgs.coreutils}/bin/readlink -e ${escapeShellArg builtins.storeDir})/*-home-manager-files/*"
 
           forcedPaths=(${forcedPaths})
 
@@ -105,7 +105,7 @@ in
             if [[ -n $forced ]]; then
               verboseEcho "Skipping collision check for $targetPath"
             elif [[ -e "$targetPath" \
-                && ! "$(readlink "$targetPath")" == $homeFilePattern ]] ; then
+                && ! "$(${pkgs.coreutils}/bin/readlink "$targetPath")" == $homeFilePattern ]] ; then
               # The target file already exists and it isn't a symlink owned by Home Manager.
               if cmp -s "$sourcePath" "$targetPath"; then
                 # First compare the files' content. If they're equal, we're fine.
@@ -136,7 +136,7 @@ in
       ''
         function checkNewGenCollision() {
           local newGenFiles
-          newGenFiles="$(readlink -e "$newGenPath/home-files")"
+          newGenFiles="$(${pkgs.coreutils}/bin/readlink -e "$newGenPath/home-files")"
           find "$newGenFiles" \( -type f -or -type l \) \
               -exec bash ${check} "$newGenFiles" {} +
         }
@@ -199,7 +199,7 @@ in
 
           # A symbolic link whose target path matches this pattern will be
           # considered part of a Home Manager generation.
-          homeFilePattern="$(readlink -e ${escapeShellArg builtins.storeDir})/*-home-manager-files/*"
+          homeFilePattern="$(${pkgs.coreutils}/bin/readlink -e ${escapeShellArg builtins.storeDir})/*-home-manager-files/*"
 
           newGenFiles="$1"
           shift 1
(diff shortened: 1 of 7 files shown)


rycee:
Thanks for the report. Would you mind trying to add

    home.emptyActivationPath = true;

to your configuration and then do a HM switch without this patch?

Also, what is the value of your `home.stateVersion` option?

andreykaipov:
Hello! Sorry for the slow response!

So with `home.emptyActivationPath = true;` explicitly set I still got the same `readlink: illegal option -- m` error.

My `home.stateVersion` is 22.11 and I saw here that means it should be true for me: https://github.com/nix-community/home-manager/blob/601c22f8af61b9a2e47001474c595c78771aa3b2/modules/home-environment.nix#L354-L360

So I tried setting it to false, but I strangely got the same issue.

I then added `echo "$PATH"` to one one of activation scripts to compare the outputs of each. In both cases I saw `/bin:/usr/bin:~/.nix-profile/bin` prepended to my path and was puzzled for a while until I remembered that 8 months ago I tried to get fancy with my home manager config and abstracted my activation script entrypoint:

https://github.com/andreykaipov/self/commit/a9eccd78db4dac6f3899643ce0a524ad08711d5c#diff-70bea2658bcd8c1952d4a1c3590bcc07fde1755308e680317c2559432bd7f4d8R36-R46

Where I literally **prepend** exactly those paths. And I distinctly remember the reason I did that at the time was because most of the binaries in my activation scripts weren't recognized otherwise! So in hindsight all I really needed to do was set `home.emptyActivationPath = false;`! Or just append my paths instead of prepend... And I also see now there's a `extraActivationPath` variable that can do the same thing.

In any case, this all happened on a Linux box, so I guess I never noticed any quirks there, but only when I tried using the same configs on this macOS machine.

---

So TLDR is I'm an idiot. However maybe there is still merit in making these readlink calls use coreutils so that it can still work regardless of how much a user might mangle their path? 😅

rycee:
Ok, glad you found the source of these calls to readline. I'm quite reluctant to change the existing behavior, though. Especially for coreutils, which are used _everywhere_.

One possibility would be to mark the `PATH` variable readonly, which would prevent people from messing around with it, instead they would get an error message `PATH: readonly variable`. I'll have to think a bit about it.

In any case, I think I'll close this PR, thanks for contributing, though. And thanks for the clear explanation about what happened!

MXfive:
Hitting this issue also, but without doing anything myself to the PATH in the activation script.  What's strange is `PATH` looks correct when I add it into the activation script.

Modified activate script
```bash
  echo $PATH
  echo "$(which readlink)"
  newDir="$(readlink -m "$newGenPath/LaunchAgents")"
```

```
Activating setupLaunchAgents
/nix/store/d72z8b7fvvcin2p84x7h5s0abkggplzc-coreutils-9.7/bin:OTHER_NIX_PATHS:/opt/homebrew/bin:/opt/homebrew/sbin:/Users/martyn/.local/bin:/Users/martyn/.krew/bin:/Users/martyn/go/bin:/Users/martyn/.nix-profile/bin:/run/current-system/sw/bin:/nix/var/nix/profiles/default/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin
/usr/bin/readlink
readlink: illegal option -- m
usage: readlink [-fn] [file ...]
readlink: illegal option -- m
usage: readlink [-fn] [file ...]
find: : No such file or directory
```

Replacing with /nix/store/d72z8b7fvvcin2p84x7h5s0abkggplzc-coreutils-9.7/bin/readlink works without issue.


@rycee I think it would be more robust to explicity call the correct `readlink` binary explicitly rather than implicitly rely on $PATH being exactly how it needs to be.

MXfive:
Update, found the culprit in my case, was part of my redacted PATH, Krew does this:

```
315:export PATH="$HOME/.krew/bin:/nix/store/9pa6rygnkrdp03j4ba0w6yb5qid5x3qb-git-2.49.0/bin:/usr/bin:$PATH";
```

So really do think it's worth having this be explicity in home-manager to not be fragile to badly written neighbour dependencies users may pull in.

Edit: in this case it was actually being pulled in by my config, Krew was part of another nix dotfiles that I'd taken some program configs from.

Still would be good to have home-manager be more robust and explicit IMO though.

PR #5093, opened 2024-03-05
ci: remove cachix action

rycee:
All the extra binary cache lookups and uploads to Cachix make the build quite a bit slower than necessary.


terlar:
The uploads could be sped up with `useDaemon: true` as they then will be done async instead of at the end. But I guess if the cache lookup and cache download doesn't help much there is no need for it at all..

rycee:
@terlar Thanks for the feedback. Yeah, I think our use-case (creating an enormous number of small files without pinning Nixpkgs) is not good for Cachix so it doesn't really make a difference in runtime, just adds unnecessary load on the Cachix servers.

I think we eventually would benefit from moving the CI out of GitHub actions into a system with direct and persistent access to a Nix store.

We also run a lot of unnecessary tests, ideally we would be able to tell which tests are affected by a given PR. For example, it is pretty unnecessary to run git module tests when the PR only changed something in the bat module.

PR #5096, opened 2024-03-06
vdirsyncer: fix verify option type

BagMan9:
### Description

Vdirsyncer options for calendar/contacts accounts specify the "verify" option as types.bool. However, this is not how vdirsyncer treats the option - it expects a path to certificate to verify against.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@teto

### DIFF
--- a/modules/programs/vdirsyncer-accounts.nix
+++ b/modules/programs/vdirsyncer-accounts.nix
@@ -93,9 +93,9 @@ in {
     };
 
     verify = mkOption {
-      type = types.nullOr types.bool;
+      type = types.nullOr types.path;
       default = null;
-      description = "Verify SSL certificate.";
+      description = "Null or path to certificate to verify SSL against";
     };
 
     verifyFingerprint = mkOption {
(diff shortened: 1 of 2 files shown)


modules/programs/vdirsyncer-accounts.nix
```diff
@@ -93,9 +93,9 @@ in {
     };
 
     verify = mkOption {
-      type = types.nullOr types.bool;
+      type = types.nullOr types.path;
       default = null;
-      description = "Verify SSL certificate.";
```
teto, line 98: Could you add the example from http://vdirsyncer.pimutils.org/en/stable/ssl-tutorial.html#custom-root-cas ?

teto:
LGTM, see http://vdirsyncer.pimutils.org/en/stable/ssl-tutorial.html#custom-root-cas

modules/programs/vdirsyncer-accounts.nix
BagMan9: Added.

modules/programs/vdirsyncer.nix
```diff
@@ -92,7 +92,7 @@ let
     else if (n == "passwordPrompt") then
       ''password.fetch = ["prompt", "${v}"]''
     else if (n == "verify") then
-      "verify = ${if v then "true" else "false"}"
+      ''verify = "${v}"''
```
teto, line 95: out of scope but this function is weird and unmaintainable :/ we should have a more transparent mapping !

PR #5101, opened 2024-03-07
xdg-user-dirs: check for existing symlink

9ary:
### Description

When a non-directory, such as a file or a dead symlink, already exists, mkdir -p fails with "cannot create directory ‘...’: File exists".

This is a problem when, for example, a symlink points to a directory on a filesystem that isn't mounted yet.

This PR simply checks for an existing symlink before calling mkdir.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@pacien


bb010g (approved):

rycee:

PR #5103, opened 2024-03-07
eza: replaced enableAliases with shell integration options

michaelBelsanti:
### Description
This commit replaces the enableAliases option with enable$SHELLIntegration options for each shell. This lets works more similarly to other programs in home-manager and lets users choose which shell's should use these aliases. 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@cafkafk

### DIFF
--- a/modules/programs/eza.nix
+++ b/modules/programs/eza.nix
@@ -4,16 +4,40 @@ with lib;
 
 {
   imports = let
+    msg = ''
+      'programs.eza.enableAliases' has been deprecated and replaced with integration
+      options per shell, for example, 'programs.eza.enableBashIntegration'.
+
+      Note, the default for these options is 'true' so if you want to enable the
+      aliases you can simply remove 'rograms.eza.enableAliases' from your
+      configuration.'';
     mkRenamed = opt:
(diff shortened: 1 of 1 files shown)


modules/programs/eza.nix
```diff
@@ -4,16 +4,29 @@ with lib;
...
+    (mkRemovedOptionModule [ "programs" "eza" "enableAliases" ] msg)
+  ];
 
   meta.maintainers = [ maintainers.cafkafk ];
 
   options.programs.eza = {
     enable = mkEnableOption "eza, a modern replacement for {command}`ls`";
 
-    enableAliases = mkEnableOption "recommended eza aliases (ls, ll…)";
+    enableBashIntegration = mkEnableOption "Bash integration";
```
rycee, line 21: Perhaps these new options should default to true?

I imagine that people typically want these aliases if they enable the eza module and with this change they would have to explicitly enable it for each shell.

rycee: Not to mention that `enable*Integration` options typically are enabled by default and it would be confusing to have too many exceptions to this rule…

rycee:

PR #5107, opened 2024-03-08
offlineimap: disable starttls if tls is disabled

### LINKED ISSUE #5097: bug: `accounts.email.accounts.<name>.imap.tls.useStartTls = false` not respected in OfflineIMAP

### Issue description

Attempting to configure OfflineIMAP to not use TLS/SSL or STARTTLS does not work properly. The generated config only sets `ssl = no` but does not set `starttls = no` despite `useStartTls = false` being explicitly stated. OfflineIMAP will fallback to STARTTLS when SSL/TLS is disabled which results in this inconsistent behavior.

### Sample Nix  config
```nix
accounts.email.accounts = {
  "[REDACTED]" = {
    [...]

    offlineimap.enable = true;

    imap = {
      host = "[REDACTED]";
      tls = {
        enable = false;
        useStartTls = false;
      };
    };

    [...]
  };
};
```

### Resulting OfflineIMAP config
```config
# Generated by Home Manager.
# See https://github.com/OfflineIMAP/offlineimap/blob/master/offlineimap.conf
# for an exhaustive list of options.

[...]

[Repository [REDACTED]-remote]
remotehost = [REDACTED]
remoteuser = [REDACTED]
ssl = no
type = IMAP
```

### Offending log line
```log
Account sync [READACTED]:
 [imap]: Using STARTTLS connection
```

### Maintainer CC

_No response_

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.5.0-25-generic, Ubuntu, 23.10 (Mantic Minotaur), nobuild`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.19.3`
 - nixpkgs: `/nix/store/snasjsbyrkp48jb63wnpgkn4flkhnlvr-source`
```

rycee:
### Description

Fixes #5097

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@teto Do you know if this makes sense? Should it be `starttls = imap.tls.useStartTls` instead, although that seems a bit odd to me. Thanks!

### DIFF
--- a/modules/programs/offlineimap.nix
+++ b/modules/programs/offlineimap.nix
@@ -42,6 +42,7 @@ let
         starttls = imap.tls.useStartTls;
       } else {
         ssl = false;
+        starttls = false;
       };
 
       remotePassEval =


rycee:
Merged this. Can always improve it in the future.

PR #5112, opened 2024-03-09
programs.bat: handle existing cache in activation script

### LINKED ISSUE #4826: bug: `programs.bat` causes activation to fail

### Issue description

With `programs.bat.enable = true`, `nixos-rebuild switch --flake .` fails during activation with the following error:

```
Dec 29 16:43:20 xxxx hm-activate-matt[587149]: Activating batCache
Dec 29 16:43:20 xxxx hm-activate-matt[587870]: error: unexpected argument '--build' found
Dec 29 16:43:20 xxxx hm-activate-matt[587870]:   tip: to pass '--build' as a value, use '-- --build'
Dec 29 16:43:20 xxxx hm-activate-matt[587870]: Usage: bat <FILE>...
Dec 29 16:43:20 xxxx hm-activate-matt[587870]: For more information, try '--help'.
```

### Maintainer CC

@lovesegfault 

### System information

```markdown
- system: `"x86_64-linux"`
- host os: `Linux 6.1.69, NixOS, 24.05 (Uakari), 24.05.20231224.5f64a12`
- multi-user?: `yes`
- sandbox: `yes`
- channels(root): `"nixos"`
```

mcwitt:
### Description

Fixes #4826

When `programs.bat.enable = true` and `~/cache` is present, activation currently fails due to `cache` in `bat cache --build` being interpreted as a path instead of a subcommand.

This PR modifies the activation script to run `bat cache --build` in an empty directory.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@lovesegfault

### DIFF
--- a/modules/programs/bat.nix
+++ b/modules/programs/bat.nix
@@ -159,10 +159,14 @@ in {
         };
       })));
 
+      # NOTE: run `bat cache --build` in an empty directory to work
+      # around failure when ~/cache exists
+      # https://github.com/sharkdp/bat/issues/1726
       home.activation.batCache = hm.dag.entryAfter [ "linkGeneration" ] ''
         (
           export XDG_CACHE_HOME=${escapeShellArg config.xdg.cacheHome}
           verboseEcho "Rebuilding bat theme cache"
+          cd "${pkgs.emptyDirectory}"
           run ${lib.getExe package} cache --build
         )
       '';


mcwitt:
I tried updating the module test to fail (e.g. with `home.file."cache".text = ""`) in https://github.com/nix-community/home-manager/blob/master/tests/modules/programs/bat/bat.nix, but I think I'd misunderstood that the test would run the activation script. Curious if there's a simple way to test the activation script?

modules/programs/bat.nix
```diff
@@ -159,10 +159,15 @@ in {
...
 
+      # NOTE: run `bat cache --build` in a fresh temporary directory to work around failure when ~/cache exists
+      # https://github.com/sharkdp/bat/issues/1726
       home.activation.batCache = hm.dag.entryAfter [ "linkGeneration" ] ''
         (
           export XDG_CACHE_HOME=${escapeShellArg config.xdg.cacheHome}
           verboseEcho "Rebuilding bat theme cache"
+          tempdir=$(mktemp -d)
+          trap "rm -r '$tempdir'" EXIT
+          cd "$tempdir"
```
rycee, line 170: I think you could simplify this to
```suggestion
          cd "${pkgs.emptyDirectory}"
```

rycee:
Yeah, the tests don't are simply asserting on the built configuration. If you want to check activation behavior then you need to do an integration test (see under `tests/integration`) but it is unfortunately more involved. I'm not certain it is worth it in this case, but if you want to then feel free to add such a test.

It should be noted that the integration tests are quite new and are not currently run automatically anywhere.

modules/programs/bat.nix
mcwitt: That's a nice solution, thanks!

rycee:

PR #5114, opened 2024-03-10
pqiv: add extraConfig option

iynaix:
### Description

Adds a `extraConfig` text option to `programs.pqiv`, so that non-INI config such as `actions` or `keybindings` can be added.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@donovanglover

### DIFF
--- a/modules/programs/pqiv.nix
+++ b/modules/programs/pqiv.nix
@@ -3,12 +3,10 @@
 with lib;
 
 let
-
   cfg = config.programs.pqiv;
   iniFormat = pkgs.formats.ini { };
-
 in {
-  meta.maintainers = with lib.maintainers; [ donovanglover ];
+  meta.maintainers = with lib.maintainers; [ donovanglover iynaix ];
 
   options.programs.pqiv = {
     enable = mkEnableOption "pqiv image viewer";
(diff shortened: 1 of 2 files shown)


modules/programs/pqiv.nix
```diff
@@ -41,6 +39,26 @@ in {
...
       '';
     };
+
+    extraConfig = mkOption {
+      type = types.lines;
+      default = "";
+      description = ''
+        Extra lines to be added to <filename>$XDG_CONFIG_HOME/pqivrc</filename> See <link
+        xlink:href="https://github.com/phillipberndt/pqiv/blob/master/pqiv.1"/>
+        for a list of available options.
```
rycee, line 49: Should be Markdown
```suggestion
        Extra lines to be added to {file}`$XDG_CONFIG_HOME/pqivrc`. See
        {manpage}`pqiv(1)` for a list of available options.
```

rycee:
Thanks! Just need to fix the description.

modules/programs/pqiv.nix
iynaix: Ok, I've also edited the description for the other `settings` option as well.

rycee:

PR #5125, opened 2024-03-12
make zsh session vars safer

postsolar:
### Description

If the shell is launched with `-u` flag, references to unset variables result in errors. Since in this case the variable is expected to be unset, there is no reason for this code to result in error. This commit prevents this potential error by providing a fallback value (the empty string) for the variable.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/zsh.nix
+++ b/modules/programs/zsh.nix
@@ -529,7 +529,7 @@ in
         . "${config.home.profileDirectory}/etc/profile.d/hm-session-vars.sh"
 
         # Only source this once
-        if [[ -z "$__HM_ZSH_SESS_VARS_SOURCED" ]]; then
+        if [[ -z "''${__HM_ZSH_SESS_VARS_SOURCED-}" ]]; then
           export __HM_ZSH_SESS_VARS_SOURCED=1
           ${envVarsStr}
         fi


ambroisie:
I'm just curious, what's your usecase for launching an interactive shell with `set -u` enabled?

postsolar:
Nothing specific, I just like to be informed of that kind of things.

modules/programs/zsh.nix
```diff
@@ -529,7 +529,7 @@ in
         . "${config.home.profileDirectory}/etc/profile.d/hm-session-vars.sh"
 
         # Only source this once
-        if [[ -z "$__HM_ZSH_SESS_VARS_SOURCED" ]]; then
+        if [[ -z "''${__HM_ZSH_SESS_VARS_SOURCED-}" ]]; then
```
teto, line 532: I usually see a column before the dash https://stackoverflow.com/questions/2013547/assigning-default-values-to-shell-variables-with-a-single-command-in-bash . Is that syntax official ? or zsh specific maybe

postsolar: This syntax is universal to POSIX-compliant shells, and in this situation makes slightly more sense semantically. See [the table here](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02) for a comparison between them (scroll a bit down).

PR #5128, opened 2024-03-13
firefox: add support for firefox derivatives

### LINKED ISSUE #2803: Add librewolf support to the firefox config module

### LINKED ISSUE #4179: `programs.librewolf`: allow installing extensions like `programs.firefox`
### Description

Feel free to close if you feel this is a duplicate of #2803, but I think it's slightly different as far as I understand?

Firefox has `programs.firefox.profiles.myprofile.extensions = [ foo bar ];`.

I'd like to be able to do the same for Librewolf like: `programs.librewolf.profiles.myprofile.extensions = [ foo bar ];`

### LINKED ISSUE #4912: Can you add support for floorp browser?
### Description
Floorp

Currently If I use `package = pkgs.floorp` in my firefox settings configured using home-manager, it does not update the settings for floorp browser. I checked and there is no module for it atm. If I am not wrong a seperate module is needed for it to work as floorp browser does not use `~/.mozilla` folder for configs instead it uses `~/.floorp`. I can't use it due to this reason but want to, please help. Please correct me if I am doing anything wrong.

brckd:
Adds support for Firefox derivatives.

### Description

Adds support for Firefox forks by introducing methods that create generic configs and options. Additional configs and options can be added in separate modules.

Most changes are just formatting. Compare them [here](https://github.com/nix-community/home-manager/pull/5128/files?w=1) or with `git diff -w`.

### Browsers that can benefit from this (Reference implementations)

| Browser | Changes | Resolves | Definition | Consumption |
| --- | --- | --- | --- | --- |
| Firefox | Functionality remains unchanged. | | [View](https://github.com/brckd/home-manager/blob/firefox-forks/modules/programs/firefox.nix) | [View](https://github.com/brckd/nixdots/blob/demo/home-manager/firefox/forks/modules/home/firefox/default.nix) |
| Librewolf | Extends the existing module. | Resolves #4179, resolves #2803, closes #5004 and closes #3339. | [View](https://github.com/brckd/home-manager/blob/firefox/librewolf/modules/programs/librewolf.nix) | [View](https://github.com/brckd/nixdots/blob/demo/home-manager/firefox/librewolf/modules/home/firefox/default.nix)
| Floorp | New Module. | Resolves #4912. | [View](https://github.com/brckd/home-manager/blob/firefox/floorp/modules/programs/floorp.nix) | [View](https://github.com/brckd/nixdots/blob/demo/home-manager/firefox/floorp/modules/home/firefox/default.nix)
| Mullvad Browser | New Module WIP. Package not yet compatible. | #4926 would be resolved if fixed. |

`firefox-profile-settings` already fails for me on the master branch, but other tests pass.

Functionality remains the same.
Generic tests added.

    ```
    {component}: {description}

    {long description}
    ```

Wondering if I should add myself to `mkFirefoxModule` despite it not being a pure module.

#### Maintainer CC

@rycee @kira-bruneau


tomodachi94:
In theory this could work for Zotero as well, being based on some sort of Mozilla technology and having the same profiles system as Firefox and derivatives. For example, extensions are stored there, which could make declarative setup easier.

brckd:
Edited the PR to contain fewer changes. Additional configs and options for other browsers are probably not scope of this PR and can be added once this is merged.

brckd:
Derived branches [Librewolf](https://github.com/brckd/home-manager/tree/ff-forks/librewolf) and [Floorp](https://github.com/brckd/home-manager/tree/ff-forks/floorp) seem to work. Will add tests for derivatives soon.

brckd:
It would probably more elegant to keep the old module, move it to `modules/programs/firefox/mkFirefoxModule.nix` and wrap it into a method that accepts config parameters.
Then just import it in `modules/programs/firefox/default.nix` as
```nix
let mkFirefoxModule = import ./mkFirefoxModule.nix; in {
  imports = [
    mkFirefoxModule { /** configs **/ }
    # ...
  ]
}
```
I'll work on the test first to finish the draft though.

brckd:
Created the mkFirefoxModule method and added generic tests. This means the PR ticks all boxes and is ready for review! :tada:

brckd:
Some things that could still be worked on are
- Changing all configuration options that are passed to `mkFirefoxModule` to `internal` options
- Moving generic module factories to `modules/lib`
What do you think?

tomodachi94:
Any hard blockers that are preventing this from being merged?

brckd:
A maintainer needs to review it.

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -0,0 +1,984 @@
+modulePath:
+{ name, description ? null, wrappedPackageName ? null
```
rycee, line 2: I think these arguments could be merged?

```suggestion
{ modulePath, name, description ? null, wrappedPackageName ? null
```

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -0,0 +1,984 @@
...
+      (pkgs.wrapFirefox.override { config = bcfg; }) package { };
+
+in {
+  options = setAttrByPath modulePath {
+    enable = mkOption {
+      type = types.bool;
+      default = false;
+      example = true;
+      description = ''
+        Whether to enable ${cfg.name}.${
```
rycee, line 235: I'm afraid that `${cfg.name}` cannot be used in descriptions. This is because the description is included in the manual and including `cfg.name` then causes the manual to have a dependency on the configuration. This, in turn, makes it more or less impossible to get the manual from a binary cache 🙁

The most convenient solution would probably be to grab the name from the first `mkFirefoxModule` argument.

rycee:
Thanks a lot! I'll had a quick look and overall everything looks great. I just added a few minor comments, and you'll also need to add yourself to `maintainers.nix` (and change `maintainers.bricked` to `hm.maintainers.bricked`).

brckd:
Thanks for the review! My NixOS installation is currently broken, so it'a difficult for me to test the suggested changes, but I will make sure to try them soon.

brckd:
Uh yea so I had like a personality crisis but I might reinstall NixOS next week and see how far I get. Sorry for letting you waiting.

modules/programs/firefox/mkFirefoxModule.nix
brckd: I tested the manual when I commited this and it worked just fine using the default name value. Wouldn't bet on this however, so I changed it nonetheless. The internal attribute is kept so that tests can access it.

modules/programs/firefox/mkFirefoxModule.nix
brckd: I took inspiration from another mk method for this, which I cannot find currently. Not sure why the arguments were separated there, probably because it's not a property of the config, but rather of the module itself. The same attribute layout is used in the test factories. Changed it nonetheless.

brckd:
I implemented the suggested changes and waiting for the tests to finish. I will be updating the linked test branches for the derivatives to confirm that the changes were sufficient. As mentioned in https://github.com/nix-community/home-manager/pull/5128#issuecomment-2002081452 there is still some reformatting that could be done, like minimizing the diff.

brckd:
I updated the reference implementations of derivatives how they can be used in working configs. Librewolf and Floorp work flawlessly, but Mullvad Browser's package doesn't seem to be compatible.

kkoniuszy:
There's a small issue with your LibreWolf module, but that's not exactly the scope of this PR. The `appName` used in the disclaimer salt for the default search engine name is supposed to be `LibreWolf`, but it's set to `Librewolf`. This produces different hashes than the browser does and results in the default search engine setting not working for any custom engines. It works for the stock ones because Firefox skips checking the hash for those.

https://github.com/kkoniuszy/home-manager/commit/7752389b13efe4cb789c6481d157da8a53f58c56

brckd:
Thanks for pointing it out! I will make sure your suggestion will be part of a future PR after this one is merged.

kkoniuszy:
The search engine stuff broke again due to some changes in librewolf 128. Looks like the name that it uses in the disclaimer is now `Firefox`. It can be seen in `about:support` under 'Application Basics -> Name'.

image

With Floorp 11.15 it matches.

image

This is obviously some bug in librewolf, but to work around I added a separate `internalName` option that is used specifically for the search engines: https://github.com/kkoniuszy/home-manager/commit/d98a83067c28bfbffd219a5509b2573982c2bdc2. I don't think we should bother with such a workaround though, librewolf will likely fix their branding stuff anyways and then it would have to be reverted.

kkoniuszy:
Hm, it looks like the app name is available in an .ini file. Maybe we could read that in an activation script and substitute it before hashing? 

```
> cat /nix/store/j3p1hplzqr6kjh7vi0wnq81ild6lmd7c-librewolf-128.0-2/lib/librewolf/application.ini
; This file is not used. If you modify it and want the application to use
; your modifications, move it under the browser/ subdirectory and start with
; the "-app /path/to/browser/application.ini" argument.
[App]
Vendor=LibreWolf
Name=Firefox
RemotingName=librewolf
CodeName=LibreWolf
```

kkoniuszy:
I modified the script that builds `search.json.mozlz4` to read the proper appName from `lib/*/application.ini` in the firefox package. It should finally solve this problem. https://github.com/kkoniuszy/home-manager/commit/84a1ac06a5d7f4b8646e19f81efdd5ce9e340bbb

Mikilio:
Just a question here, but why is the vendorPath and configPath separated?

brckd:
@Mikilio
This distinction was already made before this PR, so I can only guess why the original author decided to do so.
`configPath` like `.mozilla/firefox` stores browser configs.
`vendorPath` like `.mozilla` on the other side isn't used by most derivatives. If it exists, it's currently currently only used for native messaging. So in terms of functionality it's not related to `configPath`. `configPath` is just conventionally a subpath of `vendorPath`.

Mikilio:
> @Mikilio This distinction was already made before this PR, so I can only guess why the original author decided to do so. `configPath` like `.mozilla/firefox` stores browser configs. `vendorPath` like `.mozilla` on the other side isn't used by most derivatives. If it exists, it's currently currently only used for native messaging. So in terms of functionality it's not related to `configPath`. `configPath` is just conventionally a subpath of `vendorPath`.

Got it, in that case the Floorp module definitely needs to set this (it's currently not). It's not included in this branch, though, do you intend to merge your other branches separately?

rycee:
I've now rebased on latest master (taking the language packs commit into account), squashed, and merged to master. I think it is good to get this in as soon as possible to avoid further merge conflicts. So please give latest master a try and see if it all works OK.

brckd:
> > @Mikilio This distinction was already made before this PR, so I can only guess why the original author decided to do so. `configPath` like `.mozilla/firefox` stores browser configs. `vendorPath` like `.mozilla` on the other side isn't used by most derivatives. If it exists, it's currently currently only used for native messaging. So in terms of functionality it's not related to `configPath`. `configPath` is just conventionally a subpath of `vendorPath`.
> 
> Got it, in that case the Floorp module definitely needs to set this (it's currently not). It's not included in this branch, though, do you intend to merge your other branches separately?
> 

I am currently on vacation, so do feel free to create your own PR! Once I am back I will create PRs for the remaining browsers.

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -0,0 +1,1017 @@
...
+    });
+
+    finalPackage = mkOption {
+      inherit visible;
+      type = with types; nullOr package;
+      readOnly = true;
+      description = "Resulting ${cfg.name} package.";
+    };
+
+    policies = optionalAttrs (unwrappedPackageName != null) (mkOption {
```
timon-schelling, line 332: @brckd shouldn't this be dependet on `wrappedPackageName`?

brckd: Yes, you are absolutely right! This should use `isWrapped`, since that is also used by `wrapPackage` to decide whether the policies can be overridden.

I probably didn't notice this during testing, as the configs I tested it with had unwrapped packages.

brckd: I will work on a fix once I'm back from vacation, but anyone else feel free to attempt it in the meantime!

brckd: Created #5841.

PR #5134, opened 2024-03-13
eza: use `mkDefault` for aliases

NobbZ:
### Description

Using `mkDefault` for the individual aliases makes it easier to override or
replace individual entries by the user, without having to use `mkForce` which
is often confusing for new users.

### Checklist

- [X] Change is backwards compatible. (more than the previous PR!)

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

* @cafkafk

### DIFF
--- a/modules/programs/eza.nix
+++ b/modules/programs/eza.nix
@@ -75,7 +75,7 @@ with lib;
 
     optionsAlias = { eza = "eza ${args}"; };
 
-    aliases = {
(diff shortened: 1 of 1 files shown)


NobbZ:
I ran the tests now, and get some firefox related failure which I ignore, as it happens on `master` the same.

rycee:

Issue #4265, opened 2023-07-22
service request: eww


musjj:
### Description

[eww](https://github.com/elkowar/eww) is a traybar just like [polybar](https://github.com/polybar/polybar), and it is expected to be run on startup. So, it should also be made into a startup service, just like polybar.

Right now eww is just a program: 

https://github.com/nix-community/home-manager/blob/master/modules/programs/eww.nix

Eveeifyeve:
I might look into a pr that will enable it like waybar via `systemd.enable = true;`

PR #5136, opened 2024-03-13
services.eww: init

jkopanski:
### Description

There is no point in making eww a program (as it is now) as it does nothing more that install eww package.  Eww uses it's own configuration language, therefore it's not likely going to be configured via nix settings etc.  Current `programs.eww.config` option seems to be broken as per: #4595 #4886 .  Personally I think we should remove eww from programs.

There was a packaging request: #4265

Perhaps `programs.eww` should be removed?

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

new module but there is a `programs.eww`: @mainrs

### DIFF
--- a/modules/lib/maintainers.nix
+++ b/modules/lib/maintainers.nix
@@ -171,6 +171,12 @@
     github = "LucasWagler";
     githubId = 32136449;
   };
+  madnat = {
+    name = "Jakub Kopański";
+    email = "nix@jakub.famisoft.pl";
+    github = "jkopanski";
+    githubId = 611423;
+  };
   matrss = {
     name = "Matthias Riße";
     email = "matrss@users.noreply.github.com";
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -1443,6 +1443,13 @@ in {
           A new module is available: 'programs.joplin-desktop'.
         '';
       }
+
+      {
+        time = "2024-03-13T17:04:13+00:00";
+        message = ''
+          A new module is available: 'services.eww'.
+        '';
+      }
     ];
   };
 }
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -280,6 +280,7 @@ let
     ./services/emacs.nix
     ./services/etesync-dav.nix
     ./services/espanso.nix
+    ./services/eww.nix
     ./services/flameshot.nix
     ./services/fluidsynth.nix
     ./services/fnott.nix
--- a/modules/services/eww.nix
+++ b/modules/services/eww.nix
@@ -1,0 +1,41 @@
+{ config, lib, pkgs, ... }:
+
+let cfg = config.services.eww;
+
+in {
+  meta.maintainers = [ lib.hm.maintainers.madnat ];
+
+  options = {
+    services.eww = {
+      enable = lib.mkEnableOption "ElKowars wacky widgets daemon";
+
+      package = lib.mkPackageOption pkgs "eww" { };
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    assertions = [
+      (lib.hm.assertions.assertPlatform "services.eww" pkgs lib.platforms.linux)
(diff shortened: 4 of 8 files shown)


trueNAHO:
> Eww uses it's own configuration language, therefore it's not likely going to be configured via nix settings etc.

Unfortunately, I have not used or looked at EWW since opening https://github.com/nix-community/home-manager/pull/4595 as I currently do not have the time for it. However, IIRC its DSL is fairly early in development and its syntax is still suspect to extensions and breaking changes. Consequently, it might be easier to avoid implementing a Nix to EWW converter for now.

> There is no point in making eww a program (as it is now) as it does nothing more that install eww package.

> Personally I think we should remove eww from programs.

Unless functionality similar to https://github.com/nix-community/home-manager/pull/5585 gets merged, I agree:

> > Honestly I'm not sure what the module adds. Especially if you also can disable the `configDir` option. Then the module is literally the same as `home.packages = [ pkgs.eww ]`, which I think is just silly.
> > I'm reluctant to merge this because of that.
>
> I see your point. Maybe once the module gets expanded on in the future, it would be more reasonable.
>
> -- https://github.com/nix-community/home-manager/pull/4595#issuecomment-1908931031

jkopanski:
> However, IIRC its DSL is fairly early in development and its syntax is still suspect to extensions and breaking changes. Consequently, it might be easier to avoid implementing a Nix to EWW converter for now.

Even if it stabilizes, I think the value from this converter would be limited. I think there is enough nix functionality to move files verbatim into appropriate places.

I think this pr add one thing that is missing which is service management via systemd, but as first time contributor I'm not sure what/how to get the ball rolling on the pr.

mainrs:
> Personally I think we should remove eww from programs.

There is another PR open that adds shell completion support for `eww`: #5645, which would validate the existence of its own package.

I cannot remember right now how I programmed the configuration back when I created the module. But I think either linking the content of whole folder or specifically linking the `yuck` and `scss` file should be the way to go. If someone splits their config into multiple fragments, they can also rely on home-manager's file linking capabilities.

Jeansidharta:
Having `eww` as a service does sound more appropriate, but at least in my configuration, it'd not be enough. My `eww` config has `deflisten` calls to external programs that I use to monitor some system events. These programs have to either be on the service's `$PATH`, or their full path have to be patched in `eww`s configuration somehow.

Having a dedicated converter from nix to `eww`'s DSL would solve this problem, as we could directly reference a package's full path inside the config, but that does sound like more trouble than it's worth.

This is more of an advanced use-case though. I think having `eww` as a systemd service is better than as a standalone program. The only thing I'd improve here is maybe having `/bin` and `${pkgs.eww}/bin` in the service path. Something like `Environment = "PATH=/bin:${pkgs.eww}/bin"` in the systemd service. This would allow the user to call `eww` actions (such as `eww open window_name`) and basic shell scripts from within the config.

mainrs:
> Having `eww` as a service does sound more appropriate, but at least in my configuration, it'd not be enough. My `eww` config has `deflisten` calls to external programs that I use to monitor some system events. These programs have to either be on the service's `$PATH`, or their full path have to be patched in `eww`s configuration somehow.
> 
> Having a dedicated converter from nix to `eww`'s DSL would solve this problem, as we could directly reference a package's full path inside the config, but that does sound like more trouble than it's worth.
> 
> This is more of an advanced use-case though. I think having `eww` as a systemd service is better than as a standalone program. The only thing I'd improve here is maybe having `/bin` and `${pkgs.eww}/bin` in the service path. Something like `Environment = "PATH=/bin:${pkgs.eww}/bin"` in the systemd service. This would allow the user to call `eww` actions (such as `eww open window_name`) and basic shell scripts from within the config.

I don't think that's an issue to be honest. You can always use a string to write your configuration file in nix. And then you can also interpolate package paths.

mainrs:
Superseeded by #5979

PR #5139, opened 2024-03-14
fix(gpg-agent): fix broken variable reference

cpcloud:
### Description

Fix a reference to a non-existent variable.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/tests/modules/services/gpg-agent/default-homedir.nix
+++ b/tests/modules/services/gpg-agent/default-homedir.nix
@@ -5,11 +5,14 @@ with lib;
 {
   config = {
     services.gpg-agent.enable = true;
-    services.gpg-agent.pinentryPackage = null; # Don't build pinentry package.
+    services.gpg-agent.pinentryPackage = pkgs.pinentry-gnome3;
     programs.gpg.enable = true;
 
-    test.stubs.gnupg = { };
-    test.stubs.systemd = { }; # depends on gnupg.override
+    test.stubs = {
+      gnupg = { };
+      systemd = { }; # depends on gnupg.override
+      pinentry-gnome3 = { };
+    };
 
     nmt.script = ''
       in="${config.systemd.user.sockets.gpg-agent.Socket.ListenStream}"
@@ -18,6 +21,9 @@ with lib;
         echo $in
         fail "gpg-agent socket directory not set to default value"
       fi
+
+      configFile=home-files/.gnupg/gpg-agent.conf
+      assertFileRegex $configFile "pinentry-program @pinentry-gnome3@/bin/dummy"
     '';
   };
 }
--- a/modules/services/gpg-agent.nix
+++ b/modules/services/gpg-agent.nix
@@ -248,7 +248,7 @@ in {
           ++ optional (cfg.maxCacheTtlSsh != null)
           "max-cache-ttl-ssh ${toString cfg.maxCacheTtlSsh}"
           ++ optional (cfg.pinentryPackage != null)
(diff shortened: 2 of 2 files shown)


SebTM (approved):

andresilva (approved):

SebTM:
Can we add the test from #5141 here or afterwards?

cpcloud:
> Can we add the test from #5141 here or afterwards?

Yep, I'll add the test

cpcloud:
Done.

tests/modules/services/gpg-agent/default-homedir.nix
```diff
@@ -5,7 +5,7 @@ with lib;
 {
   config = {
     services.gpg-agent.enable = true;
-    services.gpg-agent.pinentryPackage = null; # Don't build pinentry package.
+    services.gpg-agent.pinentryPackage = pkgs.cowsay;
```
attilaolah, line 8: In hindsight, pkgs.pinentry-curses or something like that is a better candidate for tests, because cowsay doesn't set the main program so this line produces a warning.

ambroisie: You should use a stub package in either case, which _does_ set `meta.mainProgram`.

But for illustration purposes, using an actual pinentry package name would be better.

cpcloud: Fixed

cpcloud:
I get this error when running the test:

```
❯ nix develop --ignore-environment '.#test-gpg-agent-default-homedir'
trace: warning: getExe: Package "pinentry-1.2.1" does not have the meta.mainProgram attribute. We'll assume that the main program has the same name for now, but this behavior is deprecated, because it leads to surprising errors when the assumption does not hold. If the package has a main program, please set `meta.mainProgram` in its definition to make this warning go away. Otherwise, if the package does not have a main program, or if you don't control its definition, use getExe' to specify the name to the program, such as lib.getExe' foo "bar".
error:

       … while evaluating derivation 'nmt-test-gpg-agent-default-homedir'
         whose name attribute is located at /nix/store/nra828scc8qs92b9pxra5csqzffb6hpl-source/pkgs/stdenv/generic/make-derivation.nix:354:7

       … while evaluating attribute 'buildCommand' of derivation 'nmt-test-gpg-agent-default-homedir'

         at /nix/store/nra828scc8qs92b9pxra5csqzffb6hpl-source/pkgs/build-support/trivial-builders/default.nix:98:16:

           97|         enableParallelBuilding = true;
           98|         inherit buildCommand name;
             |                ^
           99|         passAsFile = [ "buildCommand" ]

       error: attribute 'override' missing

       at /nix/store/nra828scc8qs92b9pxra5csqzffb6hpl-source/pkgs/top-level/all-packages.nix:28399:20:

        28398|   };
        28399|   systemdMinimal = systemd.override {
             |                    ^
        28400|     pname = "systemd-minimal";
```

cpcloud:
Is the warning because home-manager is tied to a nixpkgs that doesn't have `meta.mainProgram` set for `pinentry-curses`?

rycee:
@cpcloud I've force pushed a change which uses a stub instead. Hopefully that'll work OK and avoid the unnecessary downloads.

cpcloud:
Tests pass for me locally.

rycee:
Yep, looks good. Thanks for the fix!

PR #5153, opened 2024-03-17
direnv: Make stdlib mergeable/sortable

JoaquinTrinanes:
### Description

This PR allows using the module system to create the `direnvrc` file.

#### Motivation
I wanted to globally run `nix_direnv_manual_reload` in `direnvrc`, but wasn't able to because sourcing `nix-direnv` was hardcoded at the end of the file. The only possible solution was to disable `nix-direnv` and manually add it to the config (doable, but not ideal).

#### Changes

Based on review comments, sourcing the `nix-direnv` file was removed altogether in favour of adding it to direnv's lib. In practice, this makes it load _before_ the `direnvrc`, instead of at the end of it. I'm open to change it back with the premise of avoiding creating unneeded files in the user's home (which I'm all for 🙂). 

~~Now, the contents are `mkMerge`d, allowing `mkAfter`, `mkBefore`, and family to be used to enforce the order of the lines.~~

~~I also made it so `nix-direnv` is loaded first by default, to more closely follow [the NixOS module](https://github.com/NixOS/nixpkgs/blob/d691274a972b3165335d261cc4671335f5c67de9/nixos/modules/programs/direnv.nix#L113).~~

~~I'm not incredibly familiar with the module system yet, so I don't know if the internal `finalStdlib` pattern to get the result of a merge is the best possible option for this. A trivial alternative would be moving the `nix-direnv` sourcing up and leaving everything as-is. I'm open to feedback!~~

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/programs/direnv.nix
+++ b/modules/programs/direnv.nix
@@ -1,8 +1,8 @@
 { config, lib, pkgs, ... }:
-
-with lib;
-
 let
+  inherit (lib)
+    mkOption mkRenamedOptionModule mkRemovedOptionModule mkEnableOption types
+    mkPackageOption mkIf mkAfter getExe;
 
   cfg = config.programs.direnv;
 
@@ -19,7 +19,7 @@ in {
       "Flake support is now always enabled.")
   ];
 
-  meta.maintainers = [ maintainers.rycee ];
+  meta.maintainers = [ lib.maintainers.rycee ];
 
   options.programs.direnv = {
     enable = mkEnableOption "direnv, the environment switcher";
@@ -104,11 +104,12 @@ in {
       source = tomlFormat.generate "direnv-config" cfg.config;
     };
 
-    xdg.configFile."direnv/direnvrc" = let
-      text = concatStringsSep "\n" (optional (cfg.stdlib != "") cfg.stdlib
-        ++ optional cfg.nix-direnv.enable
-        "source ${cfg.nix-direnv.package}/share/nix-direnv/direnvrc");
-    in mkIf (text != "") { inherit text; };
+    xdg.configFile."direnv/lib/hm-nix-direnv.sh" = mkIf cfg.nix-direnv.enable {
+      source = "${cfg.nix-direnv.package}/share/nix-direnv/direnvrc";
+    };
+
+    xdg.configFile."direnv/direnvrc" =
+      lib.mkIf (cfg.stdlib != "") { text = cfg.stdlib; };
 
     programs.bash.initExtra = mkIf cfg.enableBashIntegration (
       # Using mkAfter to make it more likely to appear after other
--- a/tests/modules/programs/direnv/nix-direnv.nix
+++ b/tests/modules/programs/direnv/nix-direnv.nix
@@ -10,9 +10,7 @@ with lib;
 
     nmt.script = ''
       assertFileExists home-files/.bashrc
-      assertFileRegex \
-        home-files/.config/direnv/direnvrc \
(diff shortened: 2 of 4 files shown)


ambroisie:
Why do we source the file, rather than link it in `XDG_CONFIG_HOME/direnv/lib/`? To quote the documentation:

> It’s also possible to create your own extensions by creating a bash file at ~/.config/direnv/direnvrc or ~/.config/direnv/lib/*.sh. This file is loaded before your .envrc and thus allows you to make your own extensions to direnv.

JoaquinTrinanes:
@ambroisie I didn't even think about it, but it looks like a cleaner solution. The only potential downside I see is maybe conflicting with another file, but in that case they probably shouldn't even coexist in the first place. Just pushed those changes :)

ambroisie:
> The only potential downside I see is maybe conflicting with another file

You could prefix it to avoid conflicts (e.g: `hm-nix-direnv.sh`).

JoaquinTrinanes:
Just changed the filename, but the more I think about it the more I like the previous way of just sourcing `nix-direnv` at the top of the file, it's a 100% guarantee that the correct code is loaded at the correct place, without polluting the user's home more than necessary 😅

At the end of the day it's just a matter of taste tho, so I'm open to both options 🤔

JoaquinTrinanes:
@rycee I forgot to mention you in the original PR description and edited afterwards, but I'm not sure if you got notified 🙂 

How do you feel about possible approaches? i.e.:
- Move sourcing to the top of the config (less flexible, but do we really need that much flexibility?)
- Move nix-direnv script into direnv's lib folder (follows direnv's standards, pollutes user home)
- `mkMerge` the config, `mkBefore` the sourcing of the script (more complex)

rycee:
I think having a separate lib file is OK. The other alternative I could imagine is to have.

``` nix
programs.direnv = mkIf cfg.nix-direnv.enable {
  stdlib = "source ${cfg.nix-direnv.package}/share/nix-direnv/direnvrc";
}
```

Then you should be able to say, for example,

``` nix
programs.direnv.stdlib = mkAfter "whatever";
```

to put `whatever` after the nix-direnv sourcing.

JoaquinTrinanes:
Thanks @rycee! I think the PR is fine to merge as is. I'll open up a followup if I find any friction point or think about a better way to manage it 🙂

modules/programs/direnv.nix
```diff
@@ -104,11 +104,13 @@ in {
...
     };
 
-    xdg.configFile."direnv/direnvrc" = let
-      text = concatStringsSep "\n" (optional (cfg.stdlib != "") cfg.stdlib
-        ++ optional cfg.nix-direnv.enable
-        "source ${cfg.nix-direnv.package}/share/nix-direnv/direnvrc");
-    in mkIf (text != "") { inherit text; };
+    xdg.configFile."direnv/lib/hm-nix-direnv.sh" = mkIf cfg.nix-direnv.enable {
+      source = "${cfg.nix-direnv.package}/share/nix-direnv/direnvrc";
+      executable = true;
```
ambroisie, line 109: Nit: I don't think this is necessary.

JoaquinTrinanes: I thought lib files had to end in `.sh` _and_ be executable, but only the first condition is actually needed. I'll remove it and squash everything once I get back on my PC

rycee:

Issue #4702, opened 2023-11-26
bug/feat: qt: replace gnome platformTheme from qgnomeplatform to qadwaitadecorations


nyabinary:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Now that https://github.com/NixOS/nixpkgs/pull/264774 has been finally merged, the qt theme option should follow suite and change from qgnomeplatform to qadwaitadecorations, since qgnomeplatform has been deprecated and the that worked on qgnomeplatform made qadwaitadecorations as a replacement. qadwaitadecorations don't depend on gtk or adwaita-qt libraries.

### Maintainer CC

@rycee
@thiagokokada 

### System information

```markdown
N/A
```

nyabinary:
nop

PR #5156, opened 2024-03-18
qt: add adwaita platform theme

tschai-yim:
### Description

This PR adds `adwaita` as an option to `qt.platformTheme` fixing #4702 and also https://github.com/NixOS/nixpkgs/issues/295345 (the `gnome` option breaks some Qt6 tray applications). It also extends the configuraton interface to match `qt.style` which has a `name` and `package` sub-option for edge cases.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee @thiagokokada

### DIFF
--- a/modules/misc/qt.nix
+++ b/modules/misc/qt.nix
@@ -6,6 +6,7 @@ let
   # Map platform names to their packages.
   platformPackages = with pkgs; {
     gnome = [ qgnomeplatform qgnomeplatform-qt6 ];
+    adwaita = [ qadwaitadecorations qadwaitadecorations-qt6 ];
     gtk = [ libsForQt5.qtstyleplugins qt6Packages.qt6gtk2 ];
     kde = [ libsForQt5.plasma-integration libsForQt5.systemsettings ];
     lxqt = [ lxqt.lxqt-qtplugin lxqt.lxqt-config ];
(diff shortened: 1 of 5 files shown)


thiagokokada:
> It also extends the configuraton interface to match qt.style which has a name and package sub-option for edge cases.

One reason I did not do this before is because those are all breaking changes, and there is no way to do this in a way that we could inform the users (e.g.: `mkRenamedOption` wouldn't work here).

So I am not sure about those changes. This will probably generate difficult to debug errors for the users.

tschai-yim:
Added backwards compatibility with the following warning:
```shell
$ nix develop --ignore-environment .#qt-platform-theme-gnome
trace: warning: The option `qt.platformTheme` has been renamed to `qt.platformTheme.name`.
trace: warning: The value `gnome` for option `qt.platformTheme` is deprecated. Use `adwaita` instead.
bash: warning: setlocale: LC_NUMERIC: cannot change locale (de_CH.UTF-8): No such file or directory
bash: warning: setlocale: LC_TIME: cannot change locale (de_CH.UTF-8): No such file or directory
qt-platform-theme-gnome: OK
```
I think that should be enough for users to know what to do.

modules/misc/qt.nix
```diff
@@ -55,56 +56,85 @@ in {
...
+            '';
+          };
+        };
+      in lib.mkOption {
         type = with lib.types;
-          nullOr (enum [ "gtk" "gtk3" "gnome" "lxqt" "qtct" "kde" ]);
+          nullOr
+          (either (enum [ "gtk" "gtk3" "gnome" "adwaita" "lxqt" "qtct" "kde" ])
+            (lib.types.submodule { options = newOption; }));
         default = null;
```
thiagokokada, line 133: Maybe we should set this option to `internal` so it doesn't show in docs?

thiagokokada (approved):
LGTM, thanks for the backwards compatibility changes.

modules/misc/qt.nix
tschai-yim: No, I think it's better if users of the old option can still find it and then see the deprecation message.

thiagokokada: IMO, no because we don't want users to use those options. We want new users to see the new option only, and old users will receive the deprecation message.

tschai-yim: True, I'll set it to internal then. Though now `nix build .#docs-html` throws an error because [release-notes/rl-2105.md](https://github.com/nix-community/home-manager/blob/master/docs/release-notes/rl-2105.md?plain=1#L145) references the option. Because it is an old release note, we can probably remove the reference.

Another thing that I noticed is that the `.name` and `.package` submodule options don't get listed (even if `qt.platformTheme` isn't internal) in the reference documentation. The same happens with `services.syncthing.tray` which also has hidden submodules. Do you know of an option where this doesn't happen to see how we could solve this?

tschai-yim: Or we just accept it like that and keep `qt.platformTheme` not internal. I have a feeling the doc generator isn't capable of resolving submodules, especially because they're like normal [modules](https://discourse.nixos.org/t/override-submodule-options/12165). 

At least the correct option is listed in its description, and it says `or (submodule)` as type:
image
Not the most user-friendly, though.

thiagokokada: If this hides the submodule documentation themselves then I don't think setting it as internal is worth it. Sorry about the extra work.

tschai-yim: The submodules are hidden no matter the `internal` setting. I'll resolve this thread then.

modules/misc/qt.nix
```diff
@@ -149,18 +179,37 @@ in {
...
+        (warnGnomeDeprecation "qt.platformTheme" cfg.platformTheme);
+      package = null;
+    } else if cfg.platformTheme == null then {
+      name = null;
+      package = null;
+    } else {
+      name =
+        warnGnomeDeprecation "cfg.platformTheme.name" cfg.platformTheme.name;
+      package = cfg.platformTheme.package;
+    };
```
rycee, line 207: Could the warnings here be expressed through the `warnings` option instead? Then it would be displayed consistently with other warnings and also be testable. See, e.g., https://github.com/nix-community/home-manager/blob/1c2acec99933f9835cc7ad47e35303de92d923a4/modules/programs/aerc.nix#L172-L182 and https://github.com/nix-community/home-manager/blob/1c2acec99933f9835cc7ad47e35303de92d923a4/tests/modules/programs/aerc/assertion.nix#L12-L20

rycee:

oxalica:
It seems my QT theme breaks after the update, and it's mixing the light theme and dark theme together. I guess this PR may be relevant. Either using the original `qt.platformTheme = "gnome"` or the new `qt.platformTheme.name = "adwaita";` has the same visual effect as screenshot below. Both dolphin from QT5 and QT6 suffer the visual issue. I'm not sure if my setting is the correct way to use Breeze-dark for QT apps.

The items are interleaving dark and light theme. (The grey box is just for redacted texts.)

image

My previously working configuration is [here](https://github.com/oxalica/nixos-config/blob/ab3c83836759087451e2068e960546634c00693f/home/modules/sway/default.nix#L62-L68).

Birdy2014:
The README of QAdwaitaDecorations says that it is a "Qt decoration plugin", not a platform theme and that it should be applied by setting the QT_WAYLAND_DECORATION environment variable.

I don't know much about Qt, but it seems to me like it doesn't make much sense to try to set QAdwaitaDecorations as a platform theme.

Ramblurr:
@lleheny0 did you ever figure this out?

lukeleheny:
> @lleheny0 did you ever figure this out?

I did not unfortunately. Still using the same config that throws the warning, and it continues to work fine.

PR #5158, opened 2024-03-19
xdg-portal: update env vars, match nixos module

Misterio77:
### Description

Nixpkgs has recently(ish) made a few major changes to its xdg-desktop-portal package (https://github.com/NixOS/nixpkgs/pull/268197), which silently breaks our module here:
- The NIXOS_XDG_DESKTOP_PORTAL_CONFIG_DIR variable patch has been removed (in favor of putting portal configurations in /etc or XDG_CONFIG_HOME).
- A new variable, NIX_XDG_DESKTOP_PORTAL_DIR, was introduced in a patch to avoid setting XDG_DESKTOP_PORTAL_DIR (which also affected portal configuration reading, not only portal definitions)

I updated our module to match the changes, but this breakage also made me revisit this module and look into some improvements.

Long story short, I think it's worth it to make it more similar to the NixOS one, as it will make behavior more predictable and consistent.

The main change is relying on the upstream linked systemd unit (instead of using systemd.user.services), and setting the environment variables globally instead of scoping it to the unit, as it's a very global thing anyway.

This might not be backwards compatible for people already facing #4922/#4972. Maybe I should also provide the unit to ~/.config?

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/xdg-portal.nix
+++ b/modules/misc/xdg-portal.nix
@@ -82,28 +82,9 @@ in {
 
   config = let
     cfg = config.xdg.portal;
-
-    joinedPortals = pkgs.buildEnv {
-      name = "xdg-portals";
-      paths = cfg.extraPortals;
-      pathsToLink =
-        [ "/share/xdg-desktop-portal/portals" "/share/applications" ];
-    };
-
-    portalConfigPath = n:
-      "share/xdg-desktop-portal/${
-        optionalString (n != "common") "${n}-"
-      }portals.conf";
-    mkPortalConfig = desktop: conf:
-      pkgs.writeTextDir (portalConfigPath desktop)
-      (lib.generators.toINI { } { preferred = conf; });
-
-    joinedPortalConfigs = pkgs.buildEnv {
-      name = "xdg-portal-configs";
-      ignoreCollisions = true; # Let config override configPackages cfgs
-      paths = (mapAttrsToList mkPortalConfig cfg.config) ++ cfg.configPackages;
-      pathsToLink = [ "/share/xdg-desktop-portal" ];
-    };
+    packages = [ pkgs.xdg-desktop-portal ] ++ cfg.extraPortals;
+    portalsDir =
+      "${config.home.profileDirectory}/share/xdg-desktop-portal/portals";
   in mkIf cfg.enable {
     warnings = optional (cfg.configPackages == [ ] && cfg.config == { }) ''
       xdg-desktop-portal 1.17 reworked how portal implementations are loaded, you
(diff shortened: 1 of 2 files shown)


rycee:

andresilva:
This broke portals for me. 

```
❯ la $NIX_XDG_DESKTOP_PORTAL_DIR
"/etc/profiles/per-user/andre/share/xdg-desktop-portal/portals": No such file or directory (os error 2)
```

Is this expected?

Misterio77:
Hey @andresilva,

I can reproduce your issue when using the NixOS module and `useUserPackages` set to true, as in this case only specific stuff will get linked (as opposed to the entire `home-manager-path` being linked). For this, users will have to set `environment.pathsToLink`, as you noted.

I will open a follow up PR adding this info to the manual, similarly to `programs.bash.enableCompletion`.

Misterio77:
Alternatively, I wonder if the module could perhaps provide a sane `pathsToLink` default? Maybe conditionally?

Also, https://github.com/nix-community/home-manager/pull/2548 possibly fixes this, too.

Misterio77:
Note added in https://github.com/nix-community/home-manager/pull/5184

bobvanderlinden:
I ran into issues with this solution. Somehow `/etc/profiles/per-user/bob.vanderlinden/share/xdg-desktop-portal/` didn't exist. Adding `extraPortals` to `home.packages` also did not work to get `share/xdg-desktop-portal`.

A solution that did work was to explicitly set `NIX_XDG_DESKTOP_PORTAL_DIR` to a `symlinkJoin` path of the `extraPortals` packages:

```
          home.sessionVariables.NIX_XDG_DESKTOP_PORTAL_DIR = lib.mkForce
            (let
              root = pkgs.symlinkJoin {
                name = "root";
                paths = config.xdg.portal.extraPortals;
              };
            in
              "${root}/share/xdg-desktop-portal/portals"
            );
          systemd.user.sessionVariables.NIX_XDG_DESKTOP_PORTAL_DIR = lib.mkForce config.home.sessionVariables.NIX_XDG_DESKTOP_PORTAL_DIR;
```

Would this be an viable solution to adopt in home-manager?

Misterio77:
Hey @bobvanderlinden,

I went with the profile approach as it is more consistent with the NixOS module, and relying on these nixpkgs-specific variables is a little risky, in my opinion (as this PR was written to fix breakage caused by `NIXOS_XDG_DESKTOP_PORTAL_CONFIG_DIR` being removed).

The issue you're facing, I believe, only happens if you're running NixOS and have `useUserPackages = true`, which links _a lot_ less stuff. In that case, you should set `pathsToLink` like so:
```
environment.pathsToLink = [ "/share/xdg-desktop-portal" "/share/applications" ];
```

See https://github.com/nix-community/home-manager/pull/5184/files

PR #5160, opened 2024-03-19
xonsh: add module and some shell integrations

inmaldrerah:
### Description

Add a module for shell Xonsh and some shell integrations for it.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@XYenon for yazi integration
@jpteb for zoxide integration, as the latest editor, since it seems the original maintainer had left the community (sadly)

Do I also need to CC the maintainer of the xonsh package in nixpkgs?

### DIFF
--- a/modules/programs/yazi.nix
+++ b/modules/programs/yazi.nix
@@ -39,6 +39,21 @@ let
       rm -fp $tmp
     }
   '';
+
+  xonshIntegration = ''
+    def __yazi_init():
+      def ya(args):
+        tmp = $(mktemp -t "yazi-cwd.XXXXX")
+        $[yazi @(args) @(f"--cwd-file={tmp}")]
+        cwd = fp"{tmp}".read_text()
+        if cwd != "" and cwd != $PWD:
+          xonsh.dirstack.cd(cwd)
+        $[rm -f -- @(tmp)]
+
+      aliases['ya'] = ya
+    __yazi_init()
+    del __yazi_init
+  '';
 in {
   meta.maintainers = [ maintainers.xyenon ];
 
@@ -59,6 +74,8 @@ in {
     enableFishIntegration = mkEnableOption "Fish integration";
 
     enableNushellIntegration = mkEnableOption "Nushell integration";
+
+    enableXonshIntegration = mkEnableOption "Xonsh integration";
 
     keymap = mkOption {
       type = tomlFormat.type;
@@ -149,6 +166,9 @@ in {
 
     programs.nushell.extraConfig =
       mkIf cfg.enableNushellIntegration nushellIntegration;
+
+    programs.xonsh.rcFiles."yazi.xsh".text =
+      mkIf cfg.enableXonshIntegration xonshIntegration;
 
     xdg.configFile = {
       "yazi/keymap.toml" = mkIf (cfg.keymap != { }) {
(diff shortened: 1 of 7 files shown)


jpteb:
Hi, I've finally had a look and the integration with zoxide looks good to me.

inmaldrerah:
Thanks. By the way, how can I initiate the tests? Should I ping someone or just wait for maintainer approval?

inmaldrerah:
Sorry, I forgot to use `hm.maintainers`. Should be fixed now.

inmaldrerah:
Can any maintainer help with the approval please?

XYenon:
I'm not familiar with xonsh. But I have taken a look of [the document](https://xon.sh/xonshrc.html#real-world-sample-xonshrc), and I think it uses an alias instead of a `init()` function.

inmaldrerah:
> I'm not familiar with xonsh. But I have taken a look of [the document](https://xon.sh/xonshrc.html#real-world-sample-xonshrc), and I think it uses an alias instead of a `init()` function.

Ah yes, Xonsh does use aliases to make a Pythonic function available as a Shell-like command, and that's why I set the field of `aliases` [here](https://github.com/nix-community/home-manager/blob/ddecf9681e819e996150ef2a7e896709853679bb/modules/programs/yazi.nix#L53).

In Xonsh, using `def` will make the function available directly as that name in a Pythonic manner. If `ya(arg)` is defined directly (without being wrapped by `init()`) and `"ya"` is set as an alias of the `ya` function (`aliases["ya"] = ya`), it will be available as both a shell-like command and a Pythonic function, and Xonsh will prefer the latter when only `ya` is typed and will print `str(ya)` rather than call `ya([])`.

The `init()` function is neither recognized nor directly called by Xonsh. Instead, it is called immediately after it is created, and then `del`ed. In this way, the `ya(args)` function will stay inside the body of the `init()` function, and `alias["ya"]` will be a function inside a closure, making it impossible to directly call `ya` with `ya(args)`, only possible through the shell alias, i.e. `ya`, or `ya path/to/folder/or/file`.

But I have to admit `init` is not a good name, as it might collide with other functions defined with that name. I'm going to change it to `__yazi_init`.

inmaldrerah:
Could anyone help with workflow approval?

modules/programs/yazi.nix
```diff
@@ -39,6 +39,21 @@ let
       rm -fp $tmp
     }
   '';
+
+  xonshIntegration = ''
+    def __yazi_init():
+      def ya(args):
```
sxyazi, line 45: It's better to use `yy` instead of `ya` here, since `ya` has been used as the CLI program name in the new version of Yazi. See https://github.com/sxyazi/yazi/issues/914 for details.

inmaldrerah: Fixed for xonsh integration. Though I wonder whether it should be done in this PR, as it does not cover other yazi integrations.

sxyazi: It should be done in this PR - the new integrations should avoid using `ya`, and the upcoming Yazi v0.3 has marked this as a breaking change https://github.com/sxyazi/yazi/issues/1046

inmaldrerah:
Continue in https://github.com/nix-community/home-manager/pull/5494 instead.

PR #5174, opened 2024-03-24
home-manger: fix runtime closure

NobbZ:
### Description

Without this fix a `nix-buid . -A home-manager` would successfully
create a `home-manager`. And one could use the included binary until
the next garbage collection, after that it would fail with an
error that it can not find the `home-manager` sources.

Similarily a `nix-copy-closure`d `home-manager` would fail with the
same error on the copies target machine.

This problem existed on both, the flake as well as the non-flake build
of `home-manager`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,6 +1,12 @@
 { pkgs ? import <nixpkgs> { } }:
 
-rec {
+let
+  path = builtins.path {
+    path = ./.;
+    name = "home-manager-source";
+  };
+
+in rec {
   docs = let releaseInfo = pkgs.lib.importJSON ./release.json;
   in with import ./docs {
     inherit pkgs;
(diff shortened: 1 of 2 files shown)


pdg137:
Thanks for submitting this fix! I have been using this workaround to get the source dependency properly registered:

```nix
  home-manager-src = pkgs.fetchFromGitHub {
    owner = "nix-community";
    repo = "home-manager";
    # 23.11 release
    rev = "652fda4ca6dafeb090943422c34ae9145787af37";
    hash = "sha256-cLbLPTL1CDmETVh4p0nQtvoF+FSEjsnJTFpTxhXywhQ=";
  };

  src-ref-pkg = pkgs.stdenv.mkDerivation {
    name = "source-ref-pkg";
    dontUnpack = true;
    buildPhase = ''
      mkdir $out
      ln -s "${home-manager-src}" $out/src
    '';
  };

  home-manager = pkgs.buildEnv {
    name = "home-manager";
    paths = [
      (import home-manager-src {inherit pkgs;}).home-manager
      src-ref-pkg
    ];
  };
```

I verified that `nix-build https://github.com/nix-community/home-manager/archive/996a74808639cc4084c42e224ae5fcc58c3f0823.zip -A home-manager` builds a home-manager with the correct dependencies that can run successfully after a garbage collection.

shroomist:
is there a way we can help getting this to an upstream?
standalone flake users are facing issues with garbage collection.

khaneliman (approved):

PR #5178, opened 2024-03-24
spotify-player: add module

diniamo:
### Description

Adds a module for `spotify-player` with the fields `enable`, `package`, `settings`, `themes`, `keymaps`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/spotify-player.nix
+++ b/modules/programs/spotify-player.nix
@@ -1,0 +1,154 @@
+{ config, lib, pkgs, ... }:
+
+let
+  inherit (lib)
+    mkEnableOption mkPackageOption mkOption types literalExpression mkIf;
+
+  cfg = config.programs.spotify-player;
+  tomlFormat = pkgs.formats.toml { };
+
+in {
+  meta.maintainers = with lib.hm.maintainers; [ diniamo ];
+
+  options.programs.spotify-player = {
+    enable = mkEnableOption "spotify-player";
+
+    package = mkPackageOption pkgs "spotify-player" { };
+
+    settings = mkOption {
+      type = tomlFormat.type;
+      default = { };
+      example = literalExpression ''
+        {
+          theme = "default";
+          playback_window_position = "Top";
+          copy_command = {
+            command = "wl-copy";
+            args = [];
+          };
+          device = {
+            audio_cache = false;
+            normalization = false;
+          };
+        }
+      '';
+      description = ''
+        Configuration written to
+        {file}`$XDG_CONFIG_HOME/spotify-player/app.toml`.
+
+        See
+        <https://github.com/aome510/spotify-player/blob/master/docs/config.md#general>
+        for the full list of options.
+      '';
+    };
+
+    themes = mkOption {
+      type = types.listOf tomlFormat.type;
+      default = [ ];
+      example = literalExpression ''
+        [
+          {
+            name = "default2";
+            palette = {
+              black = "black";
+              red = "red";
+              green = "green";
+              yellow = "yellow";
+              blue = "blue";
+              magenta = "magenta";
+              cyan = "cyan";
+              white = "white";
+              bright_black = "bright_black";
+              bright_red = "bright_red";
+              bright_green = "bright_green";
+              bright_yellow = "bright_yellow";
+              bright_blue = "bright_blue";
+              bright_magenta = "bright_magenta";
+              bright_cyan = "bright_cyan";
+              bright_white = "bright_white";
+            };
+            component_style = {
+              block_title = { fg = "Magenta"; };
+              border = {};
+              playback_track = { fg = "Cyan"; modifiers = ["Bold"]; };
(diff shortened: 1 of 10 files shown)


diniamo:
Do the commits really have to be renamed? Can't the pull request be merged while squashing?

If so, what to and how do I rename them?

diniamo:
Should be good now, I think

modules/programs/spotify-player.nix
```diff
@@ -0,0 +1,154 @@
...
+
+  options.programs.spotify-player = {
+    enable = mkEnableOption "spotify-player";
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.spotify-player;
+      defaultText = literalExpression "pkgs.spotify-player";
+      description = "The package to use for the spotify-player binary.";
+    };
```
sumnerevans, line 18:
```suggestion
    package = mkPackageOption pkgs "spotify-player" { };
```

modules/programs/spotify-player.nix
```diff
@@ -0,0 +1,154 @@
...
+        {file}`$XDG_CONFIG_HOME/spotify-player/app.toml`.
+
+        See
+        <https://github.com/aome510/spotify-player/blob/master/docs/config.md#general>
+        for the full list of options.
+      '';
+    };
+
+    themes = mkOption {
+      type = types.listOf types.attrs;
```
sumnerevans, line 48: not sure if this works, but I think it's better than just `types.attrs` if it works

```suggestion
      type = types.listOf tomlFormat.type;
```

modules/programs/spotify-player.nix
```diff
@@ -0,0 +1,154 @@
...
+        {file}`$XDG_CONFIG_HOME/spotify-player/theme.toml`.
+
+        See
+        <https://github.com/aome510/spotify-player/blob/master/docs/config.md#themes>
+        for the full list of options.
+      '';
+    };
+
+    keymaps = mkOption {
+      type = types.listOf types.attrs;
```
sumnerevans, line 99:
```suggestion
      type = types.listOf tomlFormat.type;
```

tests/modules/programs/spotify-player/settings.nix
```diff
@@ -0,0 +1,103 @@
...
+        key_sequence = "M-enter";
+      }
+      {
+        command = "None";
+        key_sequence = "q";
+      }
+    ];
+  };
+
+  test.stubs.yazi = { };
```
sumnerevans, line 92: I assume this was a copy-paste error.

```suggestion
  test.stubs.spotify-player = { };
```

sumnerevans (changes requested):

diniamo:
Suggestions, implemented, tests still run.

That last one is something else though.

sumnerevans (changes requested):
Try rebasing on `master`? The tests are passing there.

Also, please squash this down to a single commit.

diniamo:
If you check the error, then it's saying it can't find me in the maintainers list. I have no idea why though.

modules/programs/spotify-player.nix
```diff
@@ -0,0 +1,150 @@
+{ config, lib, pkgs, ... }:
+let
+  inherit (lib)
+    mkEnableOption mkPackageOption mkOption types literalExpression mkIf;
+
+  cfg = config.programs.spotify-player;
+  tomlFormat = pkgs.formats.toml { };
+in {
+  meta.maintainers = with lib.maintainers; [ diniamo ];
```
sumnerevans, line 9:
```suggestion
  meta.maintainers = with lib.hm.maintainers; [ diniamo ];
```

sumnerevans (changes requested):
Ah, the issue is that `lib.maintainers` is the list of Nixpkgs maintainers, not the Home Manager maintainers.

diniamo:
All done!

sumnerevans (approved):

diniamo:
:sob:

rycee:

repparw:
Hi! i'm pretty new to nix and hm so I don't know what the proper way to raise this question is so I'm dropping it here, sorry!
Is there a proposal for adding actions to the module? It's a special type of keymap added to spotify-player in 0.19.1

`[[keymaps]]
command = "LyricPage"
key_sequence = "g l"

[[actions]]
action = "CopyLink"
key_sequence = "y"
target = "PlayingTrack"
`

it probably would be very similar to keymap adding an optional target (and it would have to be merged to the same keymap.toml file)

diniamo:
How do actions differ from keymaps? I don't get it exactly.

diniamo:
Although it seems like we will be adding an `actions` option to avoid breaking changes.

repparw:
apart from having a different heading in the file I think not much. I mean,
the commands are different, but I don't think the module checks for
validity on those?

PR #5182, opened 2024-03-25
antidote: use builtins.storeDir rather than /nix/store

graham33:
### Description

The antidote module hardcodes `/nix/store`, which breaks in the (rare!) case where a custom store directory is being used.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


sumnerevans (approved):
Seems sane

graham33:
Thanks!

Issue #5148, opened 2024-03-16
bug: cant set khal.color to an RGB value. Can't set khal.addresses too


teto:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

It's currently not possible to set an rgb value for a khal calendar https://github.com/nix-community/home-manager/blob/206f457fffdb9a73596a4cb2211a471bd305243d/modules/programs/khal-accounts.nix#L17 as the type constrains to values in enum.
I wonder if we should just allow freeform strings ? (be careful when setting an RGB value, it has to be quoted else it is ignored, got bitten by it with a manual config https://khal.readthedocs.io/en/latest/configure.html#the-calendars-section )

NB: It's also not possible to set addresses https://khal.readthedocs.io/en/latest/configure.html#the-calendars-section

### System information

```markdown
unstable
```

teto:
fixed in linked PR

PR #5192, opened 2024-03-26
programs.khal: ability to set RGB color

teto:
The current module constrains to values in enum but khal supports RGB colors as well
khal.readthedocs.io/en/latest/configure.html#the-calendars-section ! (be careful when setting an RGB value, it has to be quoted else it is ignored, got bitten by it with a manual config )

NB: It's also not possible to set addresses khal.readthedocs.io/en/latest/configure.html#the-calendars-section

fix the rgb part of https://github.com/nix-community/home-manager/issues/5148

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/khal.nix
+++ b/modules/programs/khal.nix
@@ -45,10 +45,11 @@ let
         + (optionalString
           (value.khal.type == "birthdays" && value.khal ? thisCollection)
           value.khal.thisCollection)
-      }"
-    ] ++ optional (value.khal.readOnly) "readonly = True" ++ [
-      (toKeyValueIfDefined (getAttrs [ "type" "color" "priority" ] value.khal))
-    ] ++ [ "\n" ]);
+      }\n        "
+    ] ++ optional (value.khal.readOnly) "readonly = True"
+      ++ optional (value.khal.color != null) "color = '${value.khal.color}'"
+      ++ [ (toKeyValueIfDefined (getAttrs [ "type" "priority" ] value.khal)) ]
+      ++ [ "\n" ]);
 
   localeFormatOptions = let
     T = lib.types;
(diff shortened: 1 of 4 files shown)


teto:
@felipeqq2 are you fine with me removing the check ? I am a bit annoyed to remove the check since it's there but it prevents usage of RGB and I am not excited to reimplement program-specifc checks in nix (a never-ending process).

felipeqq2:
> are you fine with me removing the check ?

@teto, check as in the test cases?

teto:
sry I was not clear: I meant about the change of `type` for color from the enum to a string .

felipeqq2:
Oh, that's fine by me. I agree 100%

modules/programs/khal.nix
```diff
@@ -45,10 +45,11 @@ let
         + (optionalString
           (value.khal.type == "birthdays" && value.khal ? thisCollection)
           value.khal.thisCollection)
-      }"
-    ] ++ optional (value.khal.readOnly) "readonly = True" ++ [
-      (toKeyValueIfDefined (getAttrs [ "type" "color" "priority" ] value.khal))
-    ] ++ [ "\n" ]);
+      }\n        "
```
octvs, line 48: Minor question: why the whitespace?

octvs, line 48: Minor question: why the empty line?

teto: mistake

octvs:
> mistake
I thought it was intentional since the test files have been also updated
accordingly. Which might be explained if these "expected" configs are generated
by running code above.

teto:
I intended to remove them on the next PR for `addresses` support but it wont be anytime soon I think so feel free to contribute if you are interested

octvs:
> `addresses` support
I can't say that I'm informed about it.

> feel free to contribute if you are interested
Just did, thanks for the response!

PR #5195, opened 2024-03-27
firefox: add languagePacks option

### LINKED ISSUE #5186: Add `programs.firefox.languagePacks` option
### Description
Add an [option from NixOS](https://search.nixos.org/options?channel=23.11&show=programs.firefox.languagePacks&from=0&size=50&sort=relevance&type=packages&query=languagePacks) for declaratively installing language packs

nawordar:
### Description

Port the `programs.firefox.languagePacks` option from NixOS.

**Edit** Forgot to add earlier: closes #5186 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee @kira-bruneau

### DIFF
--- a/modules/programs/firefox.nix
+++ b/modules/programs/firefox.nix
@@ -262,6 +262,18 @@ in {
         '';
       };
 
+      languagePacks = mkOption {
+        type = types.listOf types.str;
+        default = [ ];
+        description = ''
+          The language packs to install. Available language codes can be found
+          on the releases page:
+          `https://releases.mozilla.org/pub/firefox/releases/''${version}/linux-x86_64/xpi/`,
+          replacing `''${version}` with the version of Firefox you have.
+        '';
+        example = [ "en-GB" "de" ];
+      };
+
       nativeMessagingHosts = mkOption {
         type = types.listOf types.package;
         default = [ ];
@@ -736,6 +748,14 @@ in {
         message = "Container id must be smaller than 4294967294 (2^32 - 2)";
       })
 
+      {
+        assertion = cfg.languagePacks == [ ] || cfg.package != null;
+        message = ''
+          'programs.firefox.languagePacks' requires 'programs.firefox.package'
+          to be set to a non-null value.
+        '';
+      }
+
       (mkNoDuplicateAssertion cfg.profiles "profile")
     ] ++ (mapAttrsToList
       (_: profile: mkNoDuplicateAssertion profile.containers "container")
@@ -749,6 +769,15 @@ in {
     '';
 
     programs.firefox.finalPackage = wrapPackage cfg.package;
+
+    programs.firefox.policies = {
+      ExtensionSettings = listToAttrs (map (lang:
+        nameValuePair "langpack-${lang}@firefox.mozilla.org" {
+          installation_mode = "normal_installed";
+          install_url =
+            "https://releases.mozilla.org/pub/firefox/releases/${cfg.package.version}/linux-x86_64/xpi/${lang}.xpi";
+        }) cfg.languagePacks);
+    };
 
     home.packages = lib.optional (cfg.finalPackage != null) cfg.finalPackage;
 


nawordar:
The tests now fail with:
```
firefox-profile-settings: FAILED
Expected home-files/.mozilla/firefox/containers/containers.json to be same as /nix/store/22gkd3v4f6y5yz182nx9gi5w1czvjgb6-profile-settings-expected-containers.json but were different:
--- actual
+++ expected
@@ -1 +1 @@
-{"identities":[{"color":"yellow","icon":"circle","name":"shopping","public":true,"userContextId":6},{"accessKey":"","color":"","icon":"","name":"userContextIdInternal.thumbnail","public":false,"userContextId":4294967294},{"accessKey":"","color":"","icon":"","name":"userContextIdInternal.webextStorageLocal","public":false,"userContextId":4294967295}],"lastUserContextId":6,"version":4}
+{"identities":[{"color":"yellow","icon":"circle","name":"shopping","public":true,"userContextId":6}],"lastUserContextId":6,"version":4}
```
Can somebody help me with fixing tests? `userContextId` seems to be randomly generated and I don't see what this output has to do with changes introduced by me.

Also, I found out that system profiles take a precedence over the user ones. I lost some time debugging why my changes didn't get applied. Is there any way to detect that there are is already a system-wide profile at `/etc/firefox/policies/policies.json`, so we can warn user about that? This is my first time doing anything with Firefox profiles and this code is mostly adapted from the equivalent in the nixpkgs repo, so maybe the reason why the language packs weren't getting installed before removing system-wide Firefox was completely different.

modules/programs/firefox.nix
```diff
@@ -206,7 +206,14 @@ let
...
         cfg = old.cfg or { } // fcfg;
-        extraPolicies = (old.extraPolicies or { }) // cfg.policies;
+        extraPolicies = (old.extraPolicies or { }) // {
+          ExtensionSettings = listToAttrs (map (lang:
+            nameValuePair "langpack-${lang}@firefox.mozilla.org" {
+              installation_mode = "normal_installed";
+              install_url =
+                "https://releases.mozilla.org/pub/firefox/releases/${cfg.package.version}/linux-x86_64/xpi/${lang}.xpi";
+            }) cfg.languagePacks);
+        } // cfg.policies;
```
alois31, line 216: Any reason why this merging code (which should probably be recursive, too) is used instead of just setting the `policies` option?

alois31, line 216: Why not simply set the `policies` option (which also takes care of properly merging things recursively)?

cezdro: I don't remember why I did it that way. I will check if it still works if I set `policies` instead

modules/programs/firefox.nix
```diff
@@ -262,6 +262,111 @@ in {
...
+          "tr"
+          "trs"
+          "uk"
+          "ur"
+          "uz"
+          "vi"
+          "xh"
+          "zh-CN"
+          "zh-TW"
+        ]);
```
alois31, line 365: `fur`, `sat`, `sc` and `tg` seem to be missing on a cursory look over what 127 supports.

cezdro: I just copied it from the `nixpkgs` repository. Maybe it's just better to leave it at `listOf str` and write in the description how to find the value?

alois31: Given that the list of languages is not static, and making a mistake is easily noticed, I think this is a reasonable option.

modules/programs/firefox.nix
```diff
@@ -745,6 +757,15 @@ in {
 
     programs.firefox.finalPackage = wrapPackage cfg.package;
 
+    programs.firefox.policies = {
+      ExtensionSettings = listToAttrs (map (lang:
+        nameValuePair "langpack-${lang}@firefox.mozilla.org" {
+          installation_mode = "normal_installed";
+          install_url =
+            "https://releases.mozilla.org/pub/firefox/releases/${cfg.package.version}/linux-x86_64/xpi/${lang}.xpi";
```
cezdro, line 765: I just noticed, that `cfg.package` has a nullable type. Is that a problem?

alois31: Unfortunately yes. Maybe an assertion should be added to inform the user that the `languagePacks` option does not work when `package` is set to null.

modules/programs/firefox.nix
```diff
@@ -743,6 +743,14 @@ in {
         message = "Container id must be smaller than 4294967294 (2^32 - 2)";
       })
 
+      {
+        assertion = length cfg.languagePacks != 0 && cfg.package != null;
```
alois31, line 747: This is logically wrong, you probably want `cfg.languagePacks == [ ] || cfg.package != null` here.

alois31 (changes requested):

alois31 (approved):

rycee:

PR #5196, opened 2024-03-27
readline: optionally place config file in XDG directory

svrana:
Based on PR #4189

### Description

Adds a home-manager option `preferXdgDirectories`. This is used in the 
readline module to configure the location of of the generated inputrc file.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


rycee:
Thanks! Merged to master now 🙂 I made some minor cleanups and added a test.

modules/home-environment.nix
```diff
@@ -473,6 +473,12 @@ in
         will be printed when the user configuration is being built.
       '';
     };
+
+    home.preferXdgDirectories = mkEnableOption "" // {
+    description = ''
+      Whether to make programs use XDG directories whenever supported.
+    '';
+};
```
ambroisie, line 481: Formatting looks weird :/

rycee: You're right. Fixed now!

PR #5197, opened 2024-03-28
amberol: add module 

surfaceflinger:
### Description

Amberol is a small and simple music player.

Added a module for it that can allow us to customize (the only) 2 settings. It also enables the gapplication-service for background playback.

Please check if everything's fine as haven't worked on home-manager before, mostly copypasted from existing easyeffects module.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/amberol.nix
+++ b/modules/services/amberol.nix
@@ -1,0 +1,70 @@
+{ config, lib, pkgs, ... }:
+
+let
+
+  cfg = config.services.amberol;
+
+in {
+  meta.maintainers = with lib.maintainers; [ surfaceflinger ];
+
+  options.services.amberol = {
+    enable = lib.mkEnableOption "" // {
+      description = ''
+        Whether to enable Amberol music player as a daemon.
+
+        Note, it is necessary to add
+        ```nix
+        programs.dconf.enable = true;
+        ```
+        to your system configuration for the daemon to work correctly.
+      '';
+    };
+
+    package = lib.mkPackageOption pkgs "amberol" { };
+
+    enableRecoloring = lib.mkOption {
+      type = lib.types.bool;
+      default = true;
+      description = "UI recoloring using the album art.";
+    };
+
+    replaygain = lib.mkOption {
+      type = lib.types.enum [ "album" "track" "off" ];
+      default = "track";
+      description = "ReplayGain mode.";
+    };
+  };
+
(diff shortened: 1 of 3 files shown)


surfaceflinger:
@rycee hi, not sure how review/merging process in home-manager looks, could you take a look at this? 🙂

modules/lib/maintainers.nix
```diff
@@ -467,6 +467,12 @@
     keys =
       [{ fingerprint = "75F0 AB7C FE01 D077 AEE6  CAFD 353E 4A18 EE0F AB72"; }];
   };
+  surfaceflinger = {
+    email = "nat@nekopon.pl";
+    github = "surfaceflinger";
+    githubId = 44725111;
+    name = "nat";
+  };
```
rycee, line 475: No need to add yourself since you're already in Nixpkgs mantainers-list
```suggestion
```

modules/services/amberol.nix
```diff
@@ -0,0 +1,73 @@
...
+      programs.dconf.enable = true;
+      ```
+      to your system configuration for the daemon to work correctly'';
+
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.amberol;
+      defaultText = lib.literalExpression "pkgs.amberol";
+      description = "The `amberol` package to use.";
+    };
```
rycee, line 24:
```suggestion
    package = lib.mkPackageOption pkgs "amberol" { };
```

modules/services/amberol.nix
```diff
@@ -0,0 +1,73 @@
...
+        description = "UI recoloring using the album art";
+      };
+
+      replaygain = lib.mkOption {
+        type = lib.types.enum [ "album" "track" "off" ];
+        default = "track";
+        description = "ReplayGain mode";
+      };
+    };
+  };
```
rycee, line 39: Only use `settings` for RFC-42 style settings. Can simply do something like

```suggestion
    enableRecoloring = lib.mkOption {
      type = lib.types.bool;
      default = true;
      description = "UI recoloring using the album art";
    };

    replaygain = lib.mkOption {
      type = lib.types.enum [ "album" "track" "off" ];
      default = "track";
      description = "ReplayGain mode";
    };
```

rycee:

PR #5199, opened 2024-03-28
fish: source vendor completions

### LINKED ISSUE #5119: fish vendor completions
### Description

Right now, we generate fish completions from man-pages, which works well. However, some programs include their own completion definitions that are better than the generated ones, and these are not used by fish.

For example, in `zsh.nix`, we add various locations to `fpath`: https://github.com/nix-community/home-manager/blob/36f873dfc8e2b6b89936ff3e2b74803d50447e0a/modules/programs/zsh.nix#L559-L561

If I look in `/etc/profiles/per-user/paho/share/zsh/site-functions`, I see definitions for various packages I have installed, and zsh is able to use them for completions.

However, we don't seem to do anything similar for fish.

I suspect this is an easy fix; I _do_ have the directory `/nix/store/8js1fw8jwj9f1l2r7jlfbcx2k3vzi42k-home-manager-path/share/fish/vendor_completions.d` populated with the desired completions, and if I run

```fish
set fish_complete_path /nix/store/8js1fw8jwj9f1l2r7jlfbcx2k3vzi42k-home-manager-path/share/fish/vendor_completions.d/ $fish_complete_path
```

then they are used. However, I am not sure the correct way to specify this in `fish.nix`. I think part of my confusion is that we also have `functions` and `completions` symlinks there (which work in fish), but none of those end up in my config directory or config file.

leana8959:
### Description

This would make `config.fish` add `${config.home-path}/share/fish/vendor_completions.d` to the completion script path.
This would fix #5119.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@rycee


paholg:
I have tested this, and it works. Thanks!

Vonfry:
This pr only add `vendor_completions.d` to `fish_complete_path` and `vendor_functions.d` isn't added here. Should we also make plugin system to work with `vendor_conf.d` and other `vendor_...` as well?

leana8959:
@rycee Could you review this PR? Thanks!
Have a nice day :)

leana8959:
Update for branch 24.05: this patch is still needed for fish to source vendor completions !

Here's an example:
Without sourcing vendor completions (shows path)
image


After sourcing vendor completions (in this case, shows password entries)
image

It would be nice for this to be merged :)

Have a nice day :)

modules/programs/fish.nix
```diff
@@ -439,11 +439,12 @@ in {
         # add completions generated by Home Manager to $fish_complete_path
         begin
           set -l joined (string join " " $fish_complete_path)
-          set -l prev_joined (string replace --regex "[^\s]*generated_completions.*" "" $joined)
+          set -l prev_joined (string replace --regex "[^\s]*home-manager-path/share/fish/vendor_completions\.d" "" $joined)
```
budimanjojo, line 442: Is this line actually needed? I don't quite understand what this `prev_joined` is for but I think it's just to make sure that path with `*generated_completions.*` is in front?

I think this PR should do is to handle the vendor completions generated by NixOS system configuration which I believe is `/etc/profiles/per-user/<username>/share/fish/vendor_completions.d`?

leana8959: Hello :)

This begin...end section constructs the fish shell variable `fish_complete_path`. Path variables in fish are handled like lists; however, there isn't a friendly API to 
- move an element to the front if it's not in the right position
- do nothing if the element exists
- remove an element matching a regex (not as I know without using a loop)
This is most likely done to ensure idempotency.

The solution to this (done by contributor who wrote this section before me) joined the list into a string, removes the generated completion paths using regex, and then update the `fish_complete_path`. The use of regex here is notable, it is here to remove all generated completion, because they are prefixed with nix store's cryptographic hash.

For the above reasons, I don't think we can remove this line. This might result in nix sourcing multiple times the same path. 

I didn't source what's in `/etc/profiles/per-user/<username>/share/fish/vendor_completions.d`, because this would be an invalid path if someone were to use `home-manager` as a standalone module (or with `home-manager.useUserPackages = false;`)
`"${config.home.path}/share/fish/vendor_completions.d"` refers to the vendor_completions.d aggregated in home-manager's linked path.

Have a nice day :)

budimanjojo: Thank you for the explanation, what I meant by handling the vendor completion by system module is when user has `programs.fish.vendor.completions.enable = true;` in the configuration.nix file. I believe this will cause duplicate vendor completions? It's not deal breaker but sort of innefficient I think.

leana8959: > when user has programs.fish.vendor.completions.enable = true; in the configuration.nix file. I believe this will cause duplicate vendor completions?

Interesting, I didn't think of this.
What is proposed in this PR will indeed created duplicated vendor completions. Is there a way to detect whether a configuration is being used as a NixOS module or not, so that we can decide which path to source programmatically ?
`/etc/profiles/per-user/<username>/share/fish/vendor_completions.d` won't exist if the user doesn't enable NixOS managed `fish`, or if home-manager is was installed as standalone.
Maybe in `config.fish` script we can detect whether `/share/fish/vendor_completions.d` in `/etc` exists and then source that conditionally ?

budimanjojo: I don't know for sure the best way to do this, but maybe using `fish` `string replace --regex` to filter them out? Regex something like `*/per-user/\.+/share/fish/vendor_completions.d`? But I think this should be a module option that user can switch on/off in case this break anything? Maybe @rycee or other maintainers can provide better idea on this.

budimanjojo:
@leana8959 I found out something that to me is a better idea. If you override the fish package there's a switch that will make `fish` source vendor completions properly. Look at my code here for better understanding: https://github.com/budimanjojo/nix-config/blob/00f515a1d0016b5f493e9b3ccc6c0456005d95b1/home/_modules/shell/fish/default.nix#L29-L43

Basically, if you install Nix using the official or the one from Determinate System, the installer will put something like this in `/etc/fish/conf.d/nix.fish` (official installer do this only when the directory `/etc/fish/conf.d` exists so it's hit and miss) or `/usr/share/fish/conf.d/nix.fish` (DetSys installer do this and it's also hit or miss). This will set `$NIX_PROFILES` env var and add them to your `$PATH`, but it's flawed in some situations because `$NIX_PROFILES` needed by fish in nixpkgs is sourced too late so the package can not do these: https://github.com/NixOS/nixpkgs/blob/797f7dc49e0bc7fab4b57c021cdf68f595e47841/pkgs/shells/fish/default.nix#L92-L117

By patching the fish package to source `/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.fish` first, this will make the internal script that will setup fish to work properly in Nix to be working again.

leana8959:
Hey, thanks for the suggestion @budimanjojo ! I think yours is arguably better than my implementation.

Does `/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.fish` only exist for non-NixOS systems ? I don't have it on my machine, where can I find one copy of the file, so I can understand what it does and how it does it ?

budimanjojo:
The presence of `/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.fish` depends on where your `$NIX_PROFILE` is set by the installer. For NixOS it's in `/run/current-system/sw` so the file should be in `/run/current-system/sw/etc/profile.d/nix-daemon.fish`.

leana8959:
I see, thanks :)

@budimanjojo Do you want to make another PR ? I can continue this one with your idea, but that would be me taking the credit of your patch.

Also, if I understand correctly, this is only a problem on non NixOS machine ? I admit that I didn't know of the option `environment.pathsToLink`, and now it works correctly for me after linking adding this configuration.

```nix
environment.pathsToLink = [
  "/share/fish/vendor_conf.d"
  "/share/fish/vendor_completions.d"
  "/share/fish/vendor_functions.d"
];
```

budimanjojo:
My suggestion is only useful in non NixOS machines because on NixOS the system module will handle the `$NIX_PROFILES` correctly. So I don't think it should be in `home-manager` module and most definitely shouldn't be defaulting to patching the package like this. If you want to do this in HM, I think it should be done in `config.targets.genericLinux` module instead?

As for taking the credit from my suggestion, don't worry and just go ahead. I'm very happy to be able to help :)

Vonfry:
It is still important for me.
On NixOS, the fish module can handle this as expect. 
But for home-manager only, they should be processed this as well, because plugins in Nixpkgs are stored under vendors.

leana8959:
Hello,

Sorry, after all I don't know I am aware of the best way to do this, so I'm closing the PR.

PR #5203, opened 2024-03-30
xmonad: fix cp failure if libFiles with subdirectories

Vonfry:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/window-managers/xmonad.nix
+++ b/modules/services/window-managers/xmonad.nix
@@ -124,6 +124,7 @@ in {
               cfg.libFiles)
           })
           for key in "''${!libFiles[@]}"; do
+            mkdir -p "xmonad-config/lib/$(dirname "$key")"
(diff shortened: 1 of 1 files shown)


modules/services/window-managers/xmonad.nix
```diff
@@ -124,6 +124,7 @@ in {
               cfg.libFiles)
           })
           for key in "''${!libFiles[@]}"; do
+            mkdir -p xmonad-config/lib/$(dirname $key)
```
rycee, line 127: Probably good to be defensive here and handle, e.g., whitespace in `$key`. Something like this, perhaps:
```suggestion
            mkdir -p "xmonad-config/lib/$(dirname "$key")"
```

rycee:
Thanks! Just added a minor comment.

rycee:
Thanks! Merged to master now.

PR #5204, opened 2024-03-30
kitty: allow redefining `kitty-themes` package

PerchunPak:
### Checklist

  - One test in firefox failed, but I am sure it is not my fault.

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

There is no one maintainer as I can see.

### DIFF
--- a/modules/programs/kitty.nix
+++ b/modules/programs/kitty.nix
@@ -69,6 +69,15 @@ in {
       '';
     };
 
+    themePackage = mkOption {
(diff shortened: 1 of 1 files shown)


teto:
isn't it enough to redefine it in an overlay ?

PerchunPak:
Yeah, after reading https://github.com/NixOS/nixpkgs/issues/50476 I agree.

Issue #4824, opened 2023-12-29
bug: Home manager not found at ... after nix-collect-garbage


Luxvao:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Home manager can't be found in the nix store after I run nix-collect-garbage. Is this normal? I'm running the standalone installation with flakes and am on nixos (I'm on nixos-unstable).
![image](https://github.com/nix-community/home-manager/assets/95533447/72e55fd3-454a-4fd4-8a8a-0771feb66dca)

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.1.69, NixOS, 24.05 (Uakari), 24.05.20231224.5f64a12`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.18.1`
 - channels(bor): `"nixos"`
 - channels(root): `"nixos"`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`
```

Alanazane:
I just ran into this issue as well. Garbage collection also deleted the link to my wallpaper made by home manager. Home manager doesn't seem to see anything wrong and won't re-link it after switching.
Since I can still use home manager though I assume this is relatively harmless?

evils:
same here, after running
```
sudo rm /nix/var/nix/gcroots/auto/*
sudo nix-collect-garbage -d
```

i get
```
The program 'home-manager' is not in your PATH. You can make it available in an
ephemeral shell by typing:
  nix-shell -p home-manager
```
and in that shell i get
```
Starting Home Manager activation
The previous generation number and path are in conflict! These
must be either both empty or both set but are now set to

    '7' and ''

If you don't mind losing previous profile generations then
the easiest solution is probably to run

   rm /home/evils/.local/state/nix/profiles/home-manager*
   rm /home/evils/.local/state/home-manager/gcroots/current-home

and trying home-manager switch again. Good luck!
```

and after running those `rm` commands things work as expected again

evils:
additionally, after `nix-collect-garbage` i get
`error: file 'home-manager/home-manager/home-manager.nix' was not found in the Nix search path (add it using $NIX_PATH or -I)`

until i do `nix-channel --update`
after which a GC wil break it again

Luxvao:
@evils I've never experienced that. Are you using flakes?

[...]

PR #5205, opened 2024-03-31
home-manager: fix missing string context

Kareem-Medhat:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee 
I added myself as a maintainer of the `home-manager` module as my changes directly affect it, however this might not the followed convention (I'm not sure) so I am OK on removing myself from it.

Fixes:
#4824

References:
https://discourse.nixos.org/t/home-manager-not-found/31337
https://www.reddit.com/r/Nix/comments/15mhb50/recurring_home_manager_not_found_error_after/

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -78,8 +78,7 @@
           in throwForRemovedArgs (import ./modules {
             inherit pkgs lib check extraSpecialArgs;
             configuration = { ... }: {
-              imports = modules
-                ++ [{ programs.home-manager.path = toString ./.; }];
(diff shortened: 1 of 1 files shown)


rycee:
Thanks! I think for now I'll leave myself listed as the sole maintainer of the core HM module. I do think it would be good with additional "official" maintainers but I feel a bit reluctant to add somebody doing their first commit, no offense meant. Especially with the recent xz events 🙂

I do encourage you to continue contributing and, if so, then I would be happy to consider it.

PR #5206, opened 2024-03-31
gtk: GTK4 theming should be opt-in

### LINKED ISSUE #5133: bug: Impossible to disable theming for GTK4

### Issue description

The commit https://github.com/nix-community/home-manager/commit/e9b9ecef4295a835ab073814f100498716b05a96#diff-7d7e3718822ace30b2dc9dd86ef97329bfc89c9262ccad07f5777d0ff3cc8924 introduced a bug for me, as I was using adw-gtk3 to theme legacy applications.

I don't want to theme gtk4 applications with that theme.
The manpage for nix also states that `gtk.enable` enables theming for gtk2/3. `gtk.theme` says the gtk2/3 theme to use.

I don't think setting the theme using this option should automatically apply to all gtk versions.
Please note that I'm new to nixos so if one could inform me of any way to overwrite the `gtk4Css` variable this would work for me.

However I think a better approach would be to let the user choose themes for each different gtk version. Something like this:

```nix
gtk = {
    enable = true;

    gtk3 = {
      enable = true;
      theme = {
        package = pkgs.adw-gtk3;
        name = "adw-gtk3";
      };
    };
    
    gtk4 = {
      enable = false;
    };
  };
```

### Maintainer CC

@rycee 

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.6.21, NixOS, 24.05 (Uakari), 24.05.20240309.3030f18`
 - multi-user?: `yes`
 - sandbox: `yes`
 - nixpkgs: `/nix/store/nra828scc8qs92b9pxra5csqzffb6hpl-source`
```

myuujiku:
### Description

Add an option to enable the GTK4 theming added in #4770 and disable it by default. Fixes #5133.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/misc/gtk.nix
+++ b/modules/misc/gtk.nix
@@ -237,8 +237,8 @@ in {
         gtk-cursor-theme-size = cfg.cursorTheme.size;
       };
 
-    gtk4Css =
-      lib.optionalString (cfg4.enableTheme && cfg.theme != null && cfg.theme.package != null) ''
+    gtk4Css = lib.optionalString
+      (cfg4.enableTheme && cfg.theme != null && cfg.theme.package != null) ''
         /**
          * GTK 4 reads the theme configured by gtk-theme-name, but ignores it.
          * It does however respect user CSS, so import the theme from here.
--- a/modules/misc/gtk.nix
+++ b/modules/misc/gtk.nix
@@ -215,6 +215,8 @@ in {
             {file}`$XDG_CONFIG_HOME/gtk-4.0/gtk.css`.
           '';
         };
+
+        enableTheme = mkEnableOption "Add gtkk theme to user CSS";
       };
     };
   };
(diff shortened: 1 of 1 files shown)


jkopanski (commented):
I think this should also prevent setting `gtk-theme-name` in the `gtkIni` for gtk4 variant. Otherwise it still tries to apply theme, and breaks if the selected one has poor gtk4 compatibility, see:

image

Perhaps even we should move `theme` attribute into `gtk4` one for separate control over gtk4 theming?

myuujiku:
@jkopanski 

Yeah, moving separating gtk4 theming might be a good idea. I will take a look at that soon (probably this week, but I can't promise anything because I'm pretty busy atm).
Oh, and I just noticed a typo in my commit [here](https://github.com/myuujiku/home-manager/blob/ed6c662ed0014186f39f72ed727244a74e969dce/modules/misc/gtk.nix#L219) (gtkk instead of gtk4).

diamondburned:
> I think this should also prevent setting `gtk-theme-name` in the `gtkIni` for gtk4 variant. Otherwise it still tries to apply theme, and breaks if the selected one has poor gtk4 compatibility

Does this mean that the comment in the `user.css` file is wrong? It seems to suggest that the `gtk-theme-name` option is ignored for GTK4.

KaiserCalm:
Not stale, still very nuch an issue.

ThinkChaos:
This was fixed by https://github.com/nix-community/home-manager/pull/7817

Issue #4750, opened 2023-12-08
Helix extraPackages packages should be able to be overwritten by a flake.nix/direnv 


nyabinary:
### Description

Currently, if I have rust-analyzer specified in extraPackages but then I CD into a project that has a direnv flake.nix the extraPackages currently overrides the flake.nix, in my opinion the flake.nix should override the wrapper. For example, I have a Rust nightly project with a direnv (to install rust-analyzer nightly and rust nightly in general). But the extraPackages currently overrides the rust-analyzer with the stable version of rust-analyzer instead of the nightly version, like I want it to. This is pretty annoying since I have rust-analyzer in my extraPackages to support typical Rust or standalone Rust files that don't require the nightly version of rust-analyzer. Is there any way to make it, so the package is beneath the package specified in the flake.nix/direnv environment? 

@Philipp-M

khaneliman:
Fixed by #5208.

PR #5208, opened 2024-03-31
helix: change wrapping when adding extraPackages to suffix

mrnossiom:
### Description

This makes extraPackages the default, but they do not shadow the env.

So you can still have packages (e.g. LSPs) with a different version than the global one in you local env like nix's shells.

### Checklist

  I guess this is not really backwards compatible but to me this is the expected behaviour. This PR is there to discuss my point.

  I checked and it seems there is no test to update. I'm not sure adding a test is necessary.

    ```
    {component}: {description}

    {long description}
    ```

@Philipp-M

### DIFF
--- a/modules/programs/helix.nix
+++ b/modules/programs/helix.nix
@@ -188,7 +188,7 @@ in {
           nativeBuildInputs = [ pkgs.makeWrapper ];
           postBuild = ''
             wrapProgram $out/bin/hx \
-              --prefix PATH : ${lib.makeBinPath cfg.extraPackages}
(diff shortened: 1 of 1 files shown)


ambroisie (approved):

Philipp-M (approved):
LGTM

nyabinary:
Does this fix #4750 ?

mrnossiom:
I would say that yes, It does

Philipp-M:
I don't have merge rights (and approved already), so maybe @rycee ?

mrnossiom:
I just rebased. This is still ready to merge.

rycee:

PR #5217, opened 2024-04-02
khard: add option to contact module for khard dir

### LINKED ISSUE #4531: bug: Incompatibility between vdirsyncer and khard modules

octvs:
- Add an extra option, `accounts.contact.accounts.<name>.khard.defaultCollection` to hardcode a subdir for khard to function as intended.
  - Khard expects to be given a collection directory directly, from which there can be multiple of in a vdir. This contrasts khal or vdirsyncer which support recursive search.

Fixes #4531

### Description

Details can be found on the commit msg. This is my first time trying to contribute, thus I appreciate any feedback & shown patience.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

   - ~~[ ] Added myself as module maintainer. See [example](https://github.com/nix-community/home-manager/blob/068ff76a10e95820f886ac46957edcff4e44621d/modules/programs/lesspipe.nix#L6).~~

#### Maintainer CC

@olmokramer @antonmosich

### DIFF
--- a/home-manager/po/nb_NO.po
+++ b/home-manager/po/nb_NO.po
@@ -8,8 +8,8 @@ msgstr ""
 "Project-Id-Version: Home Manager\n"
 "Report-Msgid-Bugs-To: https://github.com/nix-community/home-manager/issues\n"
 "POT-Creation-Date: 2024-02-15 16:04+0100\n"
-"PO-Revision-Date: 2024-02-16 22:01+0000\n"
-"Last-Translator: Robert Helgesson <robert@rycee.net>\n"
+"PO-Revision-Date: 2024-04-02 21:38+0000\n"
+"Last-Translator: LilleAila <olai.solsvik@gmail.com>\n"
 "Language-Team: Norwegian Bokmål <https://hosted.weblate.org/projects/"
 "home-manager/cli/nb_NO/>\n"
 "Language: nb_NO\n"
@@ -17,7 +17,7 @@ msgstr ""
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=n != 1;\n"
-"X-Generator: Weblate 5.4\n"
+"X-Generator: Weblate 5.5-dev\n"
 
 #: home-manager/home-manager:58
 msgid "No configuration file found at %s"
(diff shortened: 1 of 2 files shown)


octvs:
> Test cases updated/added. 
I wasn't sure if this is applicable for this minor tweak, but I can look into it.

modules/programs/khard.nix
```diff
@@ -79,7 +84,12 @@ in {
       [addressbooks]
       ${lib.concatMapStringsSep "\n" (acc: ''
         [[${acc.name}]]
-        path = ${acc.local.path}
+        path = ${
+          lib.strings.concatStringsSep "/" [
```
sumnerevans, line 88:
```suggestion
          lib.path.subpath.join [
```

modules/programs/khard.nix
```diff
@@ -68,6 +68,11 @@ in {
       type = with lib.types;
         attrsOf (submodule {
           options.khard.enable = lib.mkEnableOption "khard access";
+          options.khard.defaultCollection = lib.mkOption {
+            type = types.str;
+            default = "";
+            description = "Sub-vdir to hardcode for khard";
```
sumnerevans, line 74: I don't quite understand this, but I think this is more descriptive of what it is.

```suggestion
            description = "Sub-vdir where the khard address book resides.";
```

sumnerevans (changes requested):

modules/programs/khard.nix
octvs: Thank you very much for the review!
I accidentally destroyed the src branch for the PR, but I've already took these into account, could we continue on #5220?

PR #5220, opened 2024-04-03
khard: add option to contact module for khard dir

### LINKED ISSUE #4531: bug: Incompatibility between vdirsyncer and khard modules

octvs:
- Add an extra option, `accounts.contact.accounts.<name>.khard.defaultCollection` to hardcode a subdir for khard to function as intended.
  - Khard expects to be given a collection directory directly, from which there can be multiple of in a vdir. This contrasts khal or vdirsyncer which support recursive search.

Fixes #4531

### Description

Details can be found on the commit msg. This is my first time trying to contribute, thus I appreciate any feedback & shown patience.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

   - ~~[ ] Added myself as module maintainer. See [example](https://github.com/nix-community/home-manager/blob/068ff76a10e95820f886ac46957edcff4e44621d/modules/programs/lesspipe.nix#L6).~~

#### Maintainer CC

@olmokramer @antonmosich


octvs:
Referring to [this change](https://github.com/nix-community/home-manager/pull/5217#discussion_r1548748932) @sumnerevans 

My config fails to build with: 
```sh
      error: lib.path.subpath.join: Element at index 0 is not a valid subpath string:
           The given string "/home/<redacted>" starts with a `/`, representing an absolute path
```

Considering the `accounts.contact.accounts.<name>.local.path` is also a [string](https://github.com/nix-community/home-manager/blob/80546b220e95a575c66c213af1b09fe255299438/modules/accounts/contacts.nix#L12-L13), wouldn't it make sense to keep it the way it was before with `lib.strings.concatStringsSep`?

octvs:
I've used `builtins.toPath` on the resulting path instead, now it passes the tests without further changes.

sumnerevans:
Ah, I think that maybe https://noogle.dev/f/lib/path/append is better then.

octvs:
I can incorporate `lib.path.append` to the logic with the following block

```nix
xdg.configFile."khard/khard.conf".text = let
  getPath = acc:
    if (acc.khard.defaultCollection != "") then
      (lib.path.append (/. + acc.local.path) acc.khard.defaultCollection)
    else
      acc.local.path;
in ''
  [addressbooks]
  ${lib.concatMapStringsSep "\n" (acc: ''
    [[${acc.name}]]
    path = ${getPath acc}
  '') (lib.attrValues accounts)}

  ${renderSettings cfg.settings}
'';
```

which passes the tests, but the resulting path will be in the nix store.
Testing on my config:

```sh
$ cat .config/khard/khard.conf | head -n 3 | tail -n 1
path = /nix/store/0mh5j53f4s8k5bnl1z555pa5rd6xyn9x-default
```

Don't we want it to point to a user specified path?

sumnerevans:
does 

```
 (lib.path.append acc.local.path acc.khard.defaultCollection)
```

not work?

Basically, I just want to avoid having to implement the check for trailing `/` in `acc.local.path`

octvs:
>  (lib.path.append acc.local.path acc.khard.defaultCollection)
As stated, it results in a path in the store.

> Basically, I just want to avoid having to implement the check for trailing
> `/` in `acc.local.path`
I think `toPath` sanitizes it already.

I made some quick tests about this, I don't know the policy about the tests so
I'm just attaching it, if desired I can commit them as well.

octvs:
[test.txt](https://github.com/nix-community/home-manager/files/14858646/test.txt)

sumnerevans:
Ah, if `toPath` is already sanitizing, then that's good enough. Sorry for the running around in circles on this.

It would be good to add a test of the new option here: https://github.com/nix-community/home-manager/tree/master/tests/modules/programs/khard

octvs:
> Ah, if `toPath` is already sanitizing, then that's good enough. 
I can't find more info to back this, since `toPath` got deprecated[1]. I have
replaced it with `toString (/. + stringVariable)` which shows the same
behaviour, inspired by [similar changes on nixpkgs][2].

> It would be good to add a test of the new option here:
> https://github.com/nix-community/home-manager/tree/master/tests/modules/programs/khard
That I'm aware already, as seen on the patch above. I wasn't sure whether it is
okay to copy an existing test and bend it around as I did.  Anyhow, I will
include the tests I wrote.

I would rather squash all of them to a single proper commit before it gets
merged so I would appreciate a heads up before, so I can squash once at the
end. Otherwise I'm spamming the issue I've linked on the commit message.

> Sorry for the running around in circles on this.
Not at all! The feedback is much appreciated. Even further comments from you or
others are welcome, it helps my learning process.

[1]: https://github.com/NixOS/nix/commit/0bea4a50e035c06805ac65360b1b50d3681e4f87
[1]: https://github.com/NixOS/nixpkgs/commit/6da6accd303be776bb4e958da52046da86f9cb5c

olmokramer:
I've never gotten vdirsyncer collections to work, but it looks like this will keep working when you are not using collections and don't set the `defaultCollection` option. So I'm fine with merging this as is :+1: (although I cannot merge this myself).

olmokramer (approved):

rycee:
Thanks! Merged to master now.

Issue #5224, opened 2024-04-04
neovim: allow external package managers like mason.nvim


misumisumi:
### Description
Add environment variables to the wrapper to enable building and running the binary.

 [mason.nvim](https://github.com/williamboman/mason.nvim) and [lazy.nvim](https://github.com/folke/lazy.nvim) are package managers often used with neovim.
When sharing settings between different OS, it is difficult to use these tools in NIX environments due to the unique file structure.
Add environment variables to enable these tools to build and run binaries.

Although settings can be reused between OS, I believe that this has the side effect of impairing the reproducibility of the environment.
I suspect this will be a concern for many NIX users.

PR #5225, opened 2024-04-04
neovim: enable use of external package manager

misumisumi:
### Description

#5224
Maybe it's better to leave room for custom wrapArgs (like `extraArgs`)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@emilazy @teto

### DIFF
--- a/modules/programs/neovim.nix
+++ b/modules/programs/neovim.nix
@@ -198,6 +198,28 @@ in {
         '';
       };
 
+      extraWrapperArgs = mkOption {
+        type = with types; listOf str;
+        default = [ ];
+        example = literalExpression ''
+          [
+            "--suffix"
+            "LIBRARY_PATH"
+            ":"
+            "''${lib.makeLibraryPath [ pkgs.stdenv.cc.cc pkgs.zlib ]}"
+            "--suffix"
+            "PKG_CONFIG_PATH"
+            ":"
+            "''${lib.makeSearchPathOutput "dev" "lib/pkgconfig" [ pkgs.stdenv.cc.cc pkgs.zlib ]}"
+          ]
+        '';
+        description = ''
+          Extra arguments to be passed to the neovim wrapper.
+          This option sets environment variables required for building and running binaries
+          with external package managers like mason.nvim.
+        '';
+      };
+
       generatedConfigViml = mkOption {
         type = types.lines;
         visible = true;
@@ -415,7 +437,8 @@ in {
 
     programs.neovim.finalPackage = pkgs.wrapNeovimUnstable cfg.package
       (neovimConfig // {
-        wrapperArgs = (lib.escapeShellArgs neovimConfig.wrapperArgs) + " "
+        wrapperArgs = (lib.escapeShellArgs
+          (neovimConfig.wrapperArgs ++ cfg.extraWrapperArgs)) + " "
           + extraMakeWrapperArgs + " " + extraMakeWrapperLuaCArgs + " "
           + extraMakeWrapperLuaArgs;
         wrapRc = false;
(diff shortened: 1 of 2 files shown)


teto:
This is the kind of workaround we dont want to support in home-manager IMO, if you have to manually add dependencies then better to use the nix package directly. 
The good thing is that you just need to pass arguments to the wrapper so instead you could expose wrapperArguments as a neovim module option and maintain the argument generator code in your dotfiles as a home-manager out of tree module.

misumisumi:
Yeah, this suggestion goes against nix thinking, so it seems better to provide some leeway to set args than direct support.
to correct.

misumisumi:
I add `extraWrapperArgs` option.
Could you please confirm?

modules/programs/neovim.nix
```diff
@@ -198,6 +198,16 @@ in {
         '';
       };
 
+      extraWrapperArgs = mkOption {
+        type = with types; listOf str;
+        default = [ ];
+        example =
+          literalExpression ''[ "--suffix" "PATH" ":" "/path/to/bin" ]'';
```
teto, line 205: maybe put a working example instead, like `${golsp}/bin/golsp` (make sure to escape `${golsp}` suc htat it doesn't get interpolated.

teto: also this example might not be the best because one should use `extraPackages` instead. Take the example from what you want to achieve with this PR, aka pass additionnal variables for the linker and make mason work.
Bonus points if you can test it but that's not mandatory, the implementation is simple and is already tested in nixpkgs.

misumisumi: Is it okay to just add items to `tests/modules/programs/neovim/no-init.nix` or `tests/modules/programs/neovim/runtime.nix` for the test?

misumisumi: add test to [runtime.nix](https://github.com/nix-community/home-manager/blob/master/tests/modules/programs/neovim/runtime.nix).

modules/programs/neovim.nix
```diff
@@ -417,7 +434,8 @@ in {
       (neovimConfig // {
         wrapperArgs = (lib.escapeShellArgs neovimConfig.wrapperArgs) + " "
           + extraMakeWrapperArgs + " " + extraMakeWrapperLuaCArgs + " "
-          + extraMakeWrapperLuaArgs;
+          + extraMakeWrapperLuaArgs + " "
+          + concatStringsSep " " cfg.extraWrapperArgs;
```
teto, line 438: maybe put those into the escapeShellArgs 2 line 435 that has better  behavior

tests/modules/programs/neovim/runtime.nix
```diff
@@ -20,6 +20,13 @@ with lib;
             };
           }
         ];
+        extraWrapperArgs = let buildDeps = with pkgs; [ stdenv.cc.cc zlib ];
+        in [
+          "--suffix LIBRARY_PATH : ${lib.makeLibraryPath buildDeps}"
+          "--suffix PKG_CONFIG_PATH : ${
+            lib.makeSearchPathOutput "dev" "lib/pkgconfig" buildDeps
+          }"
+        ];
```
teto, line 29: you dont check this was taken into account: 
in /home/teto/hm/tests/modules/programs/neovim/./runtime.nix 's nmt.script you could check if the nvim wrapper contains `PKG_CONFIG_PATH`, aka `grep PKG_CONFIG_PATH $(which nvim)` or something like that

modules/programs/neovim.nix
misumisumi: I changed to add `cfg.extraWrapperArgs` to `neovimConfig.wrapperArgs`.
The example has also been modified accordingly.

tests/modules/programs/neovim/runtime.nix
misumisumi: I added `assertFileRegex` to nmt.script, is this OK?

teto:
thank you !

PR #5226, opened 2024-04-04
joplin-desktop: fix rebuild failure for undefined options

### LINKED ISSUE #5222: bug: multiple errors setting up joplin

### Issue description

1. Joplin derivation is not building due to this error `error: value is null while a string was expected` when I don't set the options  `sync.target` and `sync.interval`. If i set those I get error number two.
2. `touch: cannot touch '/home/xxxx/.config/joplin-desktop/settings.json': No such file or directory`. As I'm doing a fresh install the folder does not yet exist so it needs to be created before this line https://github.com/nix-community/home-manager/blob/80546b220e95a575c66c213af1b09fe255299438/modules/programs/joplin-desktop.nix#L116

### Maintainer CC

@zorrobert

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.6.22, NixOS, 24.05 (Uakari), 24.05.20240316.c75037b`
 - multi-user?: `yes`
 - sandbox: `yes`
 - nixpkgs: `/nix/store/7ay2dd0w41iqvrnkx19v4vf4f5kkxvc2-source`
```

zorrobert:
### Description

This PR is a fix for two issues that cause rebuild to fail. 

The first was caused when `sync.target` and `sync.interval` were not set, this was fixed by changing the default values from `null` to `"undefined"` and filtering these out later.

The second error occurred when the `.config/joplin-desktop` directory didn't exist (e.g. when installing Joplin for the first time) which caused the `touch` command to fail. This was fixed using `mkdir` to ensure that `.config/joplin-desktop` exists.

Closes #5222 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee @sumnerevans @berbiche

### DIFF
--- a/modules/programs/joplin-desktop.nix
+++ b/modules/programs/joplin-desktop.nix
@@ -49,7 +49,7 @@ in {
     sync = {
       target = lib.mkOption {
         type = lib.types.enum [
-          null
+          "undefined"
           "none"
           "file-system"
           "onedrive"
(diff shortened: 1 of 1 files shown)


rycee:
Thanks! Would you mind just changing the commit message to be a bit more descriptive? With the current message it is necessary to find and read the issue to understand what is going on.

zorrobert:
@rycee I updated the PR description, is this fine? Or should I change the actual commit message?

zorrobert:
@rycee No problem, I wrote a short explanation. Is this fine?

zorrobert:
@rycee I've added more info to the commit and fixed a typo, it should be ready for merge now.

rycee:

PR #5232, opened 2024-04-05
home-manager: Extract inline shell script to a file

Smaug123:
### Description

Extract a single inline-string shell script into a file, so that it is amenable to shellcheck and autoformatting. (To make this diff easier to review, I have not fixed the two existing shellcheck errors, nor have I formatted the file.)

Note that module/files.nix is currently excluded from autoformatting, but I have tried to structure this such that it doesn't look too awful when `nixfmt` is applied to it.

For reference, here's the diff displayed in Beyond Compare:
Screenshot 2024-04-05 at 14 19 22

This should be a no-op, so of course feel free to reject. If this is accepted, I may start doing this to more of the chonky inline shell scripts.

There's a solid argument for passing these things in as env vars instead, rather than doing string replacements; shout if you want that and I can try and work out how.

### Checklist

- [ ] Code formatted with `./format`. No, this file is excluded from formatting.

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/files.nix
+++ b/modules/files.nix
@@ -79,59 +79,14 @@ in
             (mapAttrsToList (n: v: v.target)
             (filterAttrs (n: v: v.force) cfg));
 
-        check = pkgs.writeText "check" ''
-          ${config.lib.bash.initHomeManagerLib}
+        storeDir = escapeShellArg builtins.storeDir;
 
-          # A symbolic link whose target path matches this pattern will be
-          # considered part of a Home Manager generation.
-          homeFilePattern="$(readlink -e ${escapeShellArg builtins.storeDir})/*-home-manager-files/*"
+        check = pkgs.substituteAll {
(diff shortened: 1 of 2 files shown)


rycee:
Thanks! Very nice change. I'll have a look as soon as possible. I should mention that I've been meaning to replace this code with a more robust tool, tentatively developed at <https://git.sr.ht/~rycee/putter>, so I would suggest not to put too much effort into the current solution. That said, any improvements would be welcome.

rycee:
Thanks! Merged to master now. Note, I made some minor changes, mainly to switch to use `pkgs.substituteAll`, which I think it more idiomatic.

Smaug123:
Thanks very much - I was using a pretty clunky idiom and was only dimly aware there are better ways to do it!

PR #5237, opened 2024-04-07
fish: let plugin read vendor_* dirs which is used in nixpkgs fishPlugins

Vonfry:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```


Vonfry:
this pr is still important for me.

modules/programs/fish.nix
```diff
@@ -523,15 +523,29 @@ in {
...
+            if test -d $plugin_dir/share/fish/vendor_completions.d
+              set fish_complete_path $fish_complete_path[1] $plugin_dir/share/fish/vendor_completions.d $fish_complete_path[2..-1]
+            end
+
             if test -d $plugin_dir/completions
               set fish_complete_path $fish_complete_path[1] $plugin_dir/completions $fish_complete_path[2..-1]
             end
 
             # Source initialization code if it exists.
+            if test -d $plugin_dir/share/fish/vendor_conf.d
```
teto, line 543: is the test needed or the for loop could just be empty if folder is missing

teto:
I wonder if the test for the existing folder is necessary. The folders could be added nevertheless ? ping the fish maintainers if you want feedback so we can merge

VergeDX:
not stale

Vonfry:
I don't use fish any more and this pr won't be maintained as well. Let me close this issue.
If any one want to add this into hm, please open a new one.

By the way, if using nixos, nixos's fish module support load these scripts automatically with user profiles where hm's packages installed if enabling useUserPackages. However, I think hm support this will be better because  someone use nixpkgs without nixos.

PR #5238, opened 2024-04-07
alacritty: add `theme` option

nicoonoclaste:
### Description

Add `program.alacritty.theme` option, to import a theme from the [`alacritty-theme`] repository.

I've been using something similar in my config repo for a while, it was high time I upstreamed it.  ;3

[`alacritty-theme`]: https://github.com/alacritty/alacritty-theme

### Checklist

- [x] Code tested through `nix-shell --pure tests -A run.$t` for the following tests:
  - alacritty-empty-settings
  - alacritty-example-settings
  - alacritty-merging-settings
  - alacritty-toml-config

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

No listed maintainer or recurrent committer.  @r-vdp was the last person to edit the module.

### DIFF
--- a/modules/programs/alacritty.nix
+++ b/modules/programs/alacritty.nix
@@ -9,6 +9,22 @@ in {
 
       package = lib.mkPackageOption pkgs "alacritty" { };
 
+      theme = lib.mkOption {
+        type = with lib.types; nullOr str;
+        default = null;
+        example = "solarized_dark";
+        description = ''
+          A theme from the
+          [alacritty-theme](https://github.com/alacritty/alacritty-theme)
+          repository to import in the configuration.
+          See <https://github.com/alacritty/alacritty-theme/tree/master/themes>
+          for a list of available themes.
+          If you would like to import your own theme, use
+          {option}`programs.alacritty.settings.general.import` or
+          {option}`programs.alacritty.settings.colors` directly.
+        '';
+      };
+
       settings = lib.mkOption {
         type = tomlFormat.type;
         default = { };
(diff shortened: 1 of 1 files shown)


r-vdp:
Cool! I'm afk most of the day today, but I'll have a look tomorrow!

r-vdp (approved):
Tried it and it works fine, thanks!

r-vdp:
@rycee any chance of getting this merged?

r-vdp:
@nbraud upstream renamed the `import` setting to `general.import`, would you mind updating that here?

After that, we can try once more to get this merged.

r-vdp:
Still relevant. This has been sitting idle for a long time waiting to be merged.

Probably needs a rebase to fix the failing tests.

modules/programs/alacritty.nix
```diff
@@ -9,6 +9,31 @@ in {
 
       package = lib.mkPackageOption pkgs "alacritty" { };
 
+      theme = lib.mkOption {
+        type = let
+          themes = lib.pipe pkgs.alacritty-theme [
+            builtins.readDir
```
sumnerevans, line 15: Are we confident in this? It sure seems like this is enumerating the wrong directory. Do we need to specifically enumerate the `/share/alacritty-theme` directory within the package?

sumnerevans: Actually, this just feels wrong to do anyway: in general we should not be importing the package into the closure for config. Let me see what I can do to resolve this...

sumnerevans: I'm not convinced that we should restrict this to just a theme from the `alacritty-themes` repository. I think that we should just allow the user to specify a file, and have as an example using the alacritty-theme package.

sumnerevans: Actually, there is an argument for having this, and I think that we should keep it but also note how to configure a theme that isn't in the alacritty-theme package.

sumnerevans:
image

OK, now the docs render fine.

modules/programs/alacritty.nix
```diff
@@ -40,8 +56,31 @@ in {
...
+      # alacritty-theme package.
+      assertion = let
+        available = lib.pipe "${pkgs.alacritty-theme}/share/alacritty-theme" [
+          builtins.readDir
+          (lib.filterAttrs
+            (name: type: type == "regular" && lib.hasSuffix ".toml" name))
+          lib.attrNames
+          (lib.map (lib.removeSuffix ".toml"))
+        ];
+      in cfg.theme == null || (builtins.elem cfg.theme available);
```
ambroisie, line 70: This will fail to eval for users that have disabled IFD.

ambroisie: IFD essentially happens whenever you have Nix code which depends on the content of a package (it's not about actually `import`ing, any logic which depends on the contents of a derivation to keep evaluating is called IFD).

sumnerevans: Well, I guess we can just remove the assertion, then. It kinda sucks that we can't assert basic things like this, but if it's breaking things, then I guess there's no way.

nbraud:
@r-vdp and everyone, thanks for finally getting this merged in  <3

PR #5247, opened 2024-04-10
Jplana/k9s darwin xdg

jplana:
### Description

module k9s: Adds support for darwin configuration files without xdg (using the `~/Library/Application Support/k9s/` folder instead of `~/.config/k9s` in this platform)
module k9s: fix some typos in some of the configuration file names (I checked [latest published release](https://github.com/derailed/k9s/blob/v0.32.4/internal/config/files.go#L144) for this)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@katexochen
@liyangau 
@hm.maintainers.LucasWagler

### DIFF
--- a/modules/programs/k9s.nix
+++ b/modules/programs/k9s.nix
@@ -80,7 +80,7 @@ in {
       type = yamlFormat.type;
       default = { };
       description = ''
-        Hotkeys written to {file}`$XDG_CONFIG_HOME/k9s/hotkey.yaml`. See
+        Hotkeys written to {file}`$XDG_CONFIG_HOME/k9s/hotkeys.yaml`. See
         <https://k9scli.io/topics/hotkeys/> for supported values.
       '';
       example = literalExpression ''
@@ -101,7 +101,7 @@ in {
       type = yamlFormat.type;
       default = { };
       description = ''
-        Plugins written to {file}`$XDG_CONFIG_HOME/k9s/plugin.yaml`. See
+        Plugins written to {file}`$XDG_CONFIG_HOME/k9s/plugins.yaml`. See
         <https://k9scli.io/topics/plugins/> for supported values.
       '';
       example = literalExpression ''
@@ -178,11 +178,11 @@ in {
         source = yamlFormat.generate "k9s-aliases" cfg.aliases;
       };
 
-      "k9s/hotkey.yaml" = mkIf (cfg.hotkey != { }) {
+      "k9s/hotkeys.yaml" = mkIf (cfg.hotkey != { }) {
         source = yamlFormat.generate "k9s-hotkey" cfg.hotkey;
       };
 
-      "k9s/plugin.yaml" = mkIf (cfg.plugin != { }) {
+      "k9s/plugins.yaml" = mkIf (cfg.plugin != { }) {
         source = yamlFormat.generate "k9s-plugin" cfg.plugin;
       };
 
--- a/modules/programs/k9s.nix
+++ b/modules/programs/k9s.nix
@@ -6,6 +6,7 @@ let
 
   cfg = config.programs.k9s;
   yamlFormat = pkgs.formats.yaml { };
+  inherit (pkgs.stdenv.hostPlatform) isDarwin;
 
 in {
   meta.maintainers = with maintainers; [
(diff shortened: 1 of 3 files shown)


rycee:
Thanks for the contribution! Looks OK to me but I'll wait for a maintainer to have a look.

CC: @LucasWagler

modules/programs/k9s.nix
```diff
@@ -162,33 +170,70 @@ in {
...
+        "k9s/skins/${name}.yaml"
+      else
+        "Library/Application Support/k9s/skins/${name}.yaml") {
+          source = yamlFormat.generate "k9s-skin-${name}.yaml" value;
+        }) cfg.skins;
+
   in mkIf cfg.enable {
     home.packages = [ cfg.package ];
 
     xdg.configFile = {
```
rycee, line 183: Seems unnecessary to duplicate the condition for each file:
```suggestion
    xdg.configFile = mkIf (!(isDarwin && !config.xdg.enable)) {
```

rycee: Or perhaps better, introduce a let binding

``` nix
enableXdgConfig = !isDarwin || config.xdg.enable
```

and have

```suggestion
    xdg.configFile = mkIf enableXdgConfig {
```

modules/programs/k9s.nix
```diff
@@ -162,33 +170,70 @@ in {
...
+          source = yamlFormat.generate "k9s-plugin" cfg.plugin;
+        };
+
+      "k9s/views.yaml" =
+        mkIf (cfg.views != { } && !(isDarwin && !config.xdg.enable)) {
+          source = yamlFormat.generate "k9s-views" cfg.views;
+        };
+    } // (if !(isDarwin && !config.xdg.enable) then skinFiles else { });
+
+    home.file = {
```
rycee, line 211:
``` nix
    home.file = mkIf (!enableXdgConfig) ({
      # …
    } // skinFiles)
```

modules/programs/k9s.nix
jplana: Hey! Thanks for taking the time to review the PR, and provide these suggestions :-) .  I added them as a new commit .

LucasWagler:
I don't see tests for the new file paths. Looks good otherwise!

I'll sanity check it on my (non-Darwin) machine when I get the chance.

Out of the scope of this PR: Would it be worthwhile to create new Home Manager options like `darwin.configFile` and `darwin.configHome`, akin to the current `xdg.config*` options?

jplana:
@LucasWagler regarding the tests for paths in darwin, I looked into it and I'm not sure If I'll be able to actually test the interesting logic. I assume that while running the tests I can do something along these lines:

```nix
 nmt.script = let
    configDir = if !pkgs.isDarwin then
      ".config/k9s"
        else
      "Library/Application Support/k9s";
    in
    ''
    assertFileExists "home-files/${configDir}/config.yaml"
...
...
```
however, I realized that the module running the tests is forcing `xdg.enable = true` (see [here](https://github.com/jplana/home-manager/blob/jplana%2Fk9s-darwin-xdg/tests/default.nix#L22)). I've even tried to override it with

```nix

let

  lib = import ../modules/lib/stdlib-extended.nix pkgs.lib;

in {
  programs.k9s.enable = true;
  xdg.enable = lib.mkForce false;
...
...
```

but when running the tests I keep getting

```
error: getting status of '/nix/store/1yky8nb7s3ia5i5qgasjfgpy0awy1r22-source/tests/modules/programs/modules': No such file or directory
```

LucasWagler:
`lib` should be readily available without needing a manual `import` in the test file. Something like this:
```nix
{ config, lib, ... }:
{
  xdg.enable = lib.mkForce false;
  programs.k9s.enable = true;
...
```

jplana:
Thanks @LucasWagler ! 

I added the tests for darwin paths without xdg enabled.

LucasWagler:
Looks good on my machine! You may need to run the formatter one last time.

Thanks for your contribution!

LucasWagler (approved):

jplana:
Formatting done! Thanks for your patience :-)

Finally, about creating the equivalent for `xdg.config*`, I looked into it. Probably it wouldn't make any sense for `xdg.configFile` as Apple specs is very clear to only allow "subdirectories" in `~/Library/Application Support` (see Apple's [Filesystem Programming Guide](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html#//apple_ref/doc/uid/TP40010672-CH2-SW1) for reference). As for `xdg.configHome`, `darwin.configHome`? `darwin.userAppSupport`? it might make sense if used as a constant so all references scattered through the home-manager codebase (and programs) use the same option. But I wouldn't expect it to change in the future, as even iOS has this very same FS organization and this folder seems to be backed up by the system there.

I you think this is worthwhile I can open a new PR and try my luck there.

rycee:

Issue #4207, opened 2023-07-07
nushell: Add read-only `programs.nushell.configDir` option


lilyball:
### Description

The nushell module puts configuration into a different location on macOS vs Linux, but it doesn't expose this to the user. This should be exposed as a read-only option `programs.nushell.configDir` that I can use to place other files into the same location without having to be aware of the per-platform location differences. This would be useful for e.g. putting files into the `scripts` subdirectory that the default config uses, or just keeping auxiliary files in the same location.

rycee:
Seems reasonable, I think. Would you be able to submit a PR that adds this option?

khaneliman:
Implemented by #6763.

PR #5250, opened 2024-04-10
nushell: fix nushell config path on darwin

### LINKED ISSUE #5248: bug: Nushell fails to load config file since version 0.92.0 on darwin

### Issue description

Since the version 0.92.0 of nushell and the full support of XDG_CONFIG_HOME, it seems on startup it fails to load its configuration file set from home-manager:

```
    programs.nushell = {
      enable = true;

      configFile = {
        text = ''
          $env.config = {
            show_banner: false

            ls: {
                use_ls_colors: true
                clickable_links: true
            }

            rm: {
                always_trash: true
            }
          }
        '';
      };
    };
```

I always got the following warnings:

```
No environment config file found at /Users/jaudiger/.config/nushell/env.nu
Would you like to create one with defaults (Y/n):
n
Continuing without config file
No config file found at /Users/jaudiger/.config/nushell/config.nu
Would you like to create one with defaults (Y/n):
n
Continuing without config file
```

I think the problem comes from these lines in home-manager: https://github.com/nix-community/home-manager/blob/40a99619da804a78a0b166e5c6911108c059c3a8/modules/programs/nushell.nix#L9-L12

### Maintainer CC

@Philipp-M 
@oceanlewis 

### System information

```markdown
- system: `"aarch64-darwin"`
 - host os: `Darwin 23.4.0, macOS 14.4.1`
 - multi-user?: `yes`
 - sandbox: `no`
 - version: `nix-env (Nix) 2.21.0`
 - channels(root): `"nixpkgs"`
```

Philipp-M:
### Description

Fixes #5248 

I'm not sure about the exact situation on darwin, whether this should be set by default (i.e. remove the if check completely), since nushell only loads from that path when `XDG_CONFIG_HOME` is set. As this requires `config.xdg.enable = true`.

(when this is the right solution, I'll fix the other programs/modules that depend on nushell such as yazi etc.) 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/nushell.nix
+++ b/modules/programs/nushell.nix
@@ -6,7 +6,7 @@ let
 
   cfg = config.programs.nushell;
 
-  configDir = if pkgs.stdenv.isDarwin then
+  configDir = if pkgs.stdenv.isDarwin && !config.xdg.enable then
     "Library/Application Support/nushell"
   else
     "${config.xdg.configHome}/nushell";
--- a/tests/modules/programs/carapace/nushell.nix
+++ b/tests/modules/programs/carapace/nushell.nix
@@ -1,4 +1,4 @@
-{ pkgs, ... }:
+{ pkgs, config, ... }:
 
 {
   programs = {
@@ -7,7 +7,7 @@
   };
 
   nmt.script = let
-    configDir = if pkgs.stdenv.isDarwin then
+    configDir = if pkgs.stdenv.isDarwin && !config.xdg.enable then
       "home-files/Library/Application Support/nushell"
     else
       "home-files/.config/nushell";
--- a/tests/modules/programs/direnv/nushell.nix
+++ b/tests/modules/programs/direnv/nushell.nix
@@ -1,4 +1,4 @@
-{ pkgs, ... }:
+{ pkgs, config, ... }:
 
 {
   programs.nushell.enable = true;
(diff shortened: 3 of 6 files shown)


Philipp-M:
Based on the confirmation on #5248 I've fixed the other dependent modules, so this is ready for review

rycee:
Thanks! It becomes a bit complicated, though. Perhaps the nushell module could define an internal option `configDir` or similar that the other modules could use directly? Then they would not have to be aware of this level of detail. More concretely add an option

``` nix
options.programs.nushell.configDir = mkOption {
  type = types.str;
  internal = true;
  readOnly = true;
  description = "Path to directory holding nushell configuration files.";
};
```

and bind it

``` nix
config.programs.nushell.configDir =
    if pkgs.stdenv.isDarwin && !config.xdg.enable then
      "Library/Application Support/nushell"
    else
      "${config.xdg.configHome}/nushell";
```

Then in the other modules use

``` nix
let configPath = config.programs.nushell.configDir;
```

Philipp-M:
You're right, thanks for the suggestions.

It's only used in the tests though other than in the nushell module itself, and all of those have a different `$HOME` dir as base (`home-files`). Do you think it makes sense then anyway to do something like that (probably needs to be a function then though)?

rycee:
Fair enough, maybe not worth the bother when it is just in the tests. Merged now 🙂

lilyball:
Darn it, I've got a long-standing issue asking for that `configDir` var as it's useful for third-party config that wants to put stuff in the same location https://github.com/nix-community/home-manager/issues/4207

Philipp-M:
Sorry haven't seen that issue. But I guess it should be simple enough to implement that in a PR?

PR #5252, opened 2024-04-10
todoman: add todoman module

Mikilio:
Adds Mikilio as maintainer for new module for todoman a standards-based task manager based on iCalendar

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/programs/todoman.nix
+++ b/modules/programs/todoman.nix
@@ -56,7 +56,7 @@ in {
 
     xdg.configFile."todoman/config.py".text = lib.concatLines [
       ''path = "${config.accounts.calendar.basePath}/${cfg.glob}"''
-      cfg.config
+      cfg.extraConfig
     ];
   };
 }
--- a/modules/programs/todoman.nix
+++ b/modules/programs/todoman.nix
@@ -1,0 +1,62 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.programs.todoman;
+
+  format = pkgs.formats.keyValue { };
+
+in {
+
+  meta.maintainers = [ hm.maintainers.mikilio ];
+
+  options.programs.todoman = {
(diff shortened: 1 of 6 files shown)


modules/programs/todoman.nix
```diff
@@ -0,0 +1,62 @@
...
+    glob = mkOption {
+      type = types.str;
+      default = "*";
+      description = ''
+        The glob expansion which matches all directories relevant.
+      '';
+      example = "*/*";
+    };
+
+    config = mkOption {
```
rycee, line 27: The convention is to call this option `extraConfig`:
```suggestion
    extraConfig = mkOption {
```

modules/programs/todoman.nix
```diff
@@ -0,0 +1,62 @@
...
+      type = types.lines;
+      default = "";
+      description = ''
+        Text for configuration of todoman.
+        The syntaxt is python.
```
rycee, line 32:
```suggestion
        The syntax is Python.
```

rycee:
Thanks for the contribution! I've added a few comments. Also, it would be good to have an simple test to verify that the generated configuration file is as expected.

Mikilio:
Thanks for the review! By test do you mean something like [this](https://github.com/nix-community/home-manager/blob/master/tests/modules/programs/khal/config.nix)?

Mikilio:
I'm sorry for the delay life has been getting kind of hectic and I figured this was low priority. Everything should be ready to merge now.

Mikilio:
@rycee reminder

Mikilio:
I cleaned up the commits in case this was blocking @rycee

teto:
since rycee was ok with the idea of merging such a module and the code looks ok, I merge. I haven't tested it yet and just found out todoman could sync with caldav. nice

PR #5253, opened 2024-04-10
Added a `programs.jujutsu.configFile` file option to configure output…

aftix:
### Description

Added a `programs.jujutsu.configFile` option to configure output location of `programs.jujutsu.settings`

I'm not sure if there's a convention on how to name options that relocate the config files.

### Checklist

The option defaults to the exact string that was previously used.

`./format` slightly changed the formatting in parts of the file I didn't touch, I assume the format rules have changed since this file was last edited.

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@shikanime

### DIFF
--- a/modules/programs/jujutsu.nix
+++ b/modules/programs/jujutsu.nix
@@ -1,12 +1,8 @@
 { config, lib, pkgs, ... }:
-
 with lib;
-
 let
-
   cfg = config.programs.jujutsu;
   tomlFormat = pkgs.formats.toml { };
-
 in {
   meta.maintainers = [ maintainers.shikanime ];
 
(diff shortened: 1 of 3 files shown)


ambroisie:
Why not just follow `XDG_CONFIG_FILE` always?

shikanime:
I don't think it's a good idea. Unfortunately there's some configuration files that doesn't follow the XDG spec, but I don't think jj fall into it.

aftix:
What do you think isn't a good idea, following the XDG spec directly or having the location configurable? At this point I think the module should retain backwards compatibility and keep the file in ~/.jjconfig toml by default since I don't think jujutsu has any plans to stop supporting that location. My use case is I have XDG_CONFIG_HOME as a tmpfs and I'd like to keep configuration there

ambroisie:
> should retain backwards compatibility and keep the file in ~/.jjconfig toml by default since I don't think jujutsu has any plans to stop supporting that location.

The [`programs.git` module changed the location](https://github.com/nix-community/home-manager/commit/356c0bf751a55878c448461d8ef1d17f43911acd), and there is precedent for other modules switching to `XDG_CONFIG_HOME` unconditionnaly (e.g: [`programs.less`](https://github.com/nix-community/home-manager/commit/829e89a16f4f96428d1b94e68d4c06107b5491c0)).

rycee:
Thanks for the contribution! I chose to merge #5207, however. I think if the program natively supports the XDG location then we should always use that.

PR #5257, opened 2024-04-12
zsh: support associative arrays as local variables

muggenhor:
### Description

This adds support for `attrset` members in `programs.zsh.localVariables` and ensures they're reconstructed as ZSH associative arrays. This is particularly useful for the `ZSH_HIGHLIGHT_STYLES` associative array used by the, already included, `syntaxHighlighting` plugin.

One such example is added to the documentation.

New test case added to cover all previous types handled for this option. With the new type added on top of that of course.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/lib/zsh.nix
+++ b/modules/lib/zsh.nix
@@ -13,11 +13,15 @@ rec {
         shell = import ./shell.nix { inherit lib; };
       in
       "(${shell.formatShellArrayContent (map toString v)})"
+    else if builtins.isAttrs v then
+      "(${
+        lib.concatStringsSep " " (lib.mapAttrsToList (n: v: "[${lib.escapeShellArg n}]=${toZshValue v}") v)
+      })"
     else
       ''"${toString v}"'';
 
   # Produces a Zsh shell like definition statement
-  define = n: v: "${n}=${toZshValue v}";
+  define = n: v: "${lib.optionalString (builtins.isAttrs v) "typeset -A "}${n}=${toZshValue v}";
 
   # Given an attribute set containing shell variable names and their
   # assignments, this function produces a string containing a definition
--- a/tests/modules/programs/zsh/local-variables.nix
+++ b/tests/modules/programs/zsh/local-variables.nix
@@ -10,12 +10,13 @@
         V2 = false;
         V3 = "some-string";
         V4 = 42;
-        V5 = [
-          V1
-          V2
-          V3
-          V4
-        ];
+        V5 = builtins.attrValues V6;
+        V6 = {
+          a = V1;
+          b = V2;
+          c = V3;
+          d = V4;
+        };
       };
     };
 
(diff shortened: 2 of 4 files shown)


tests/modules/programs/zsh/local-variables.nix
```diff
@@ -0,0 +1,40 @@
...
+          d = V4;
+        };
+      };
+    };
+
+    test.stubs.zsh = { };
+
+    nmt.script = ''
+      assertFileExists home-files/.zshrc
+      assertFileContains home-files/.zshrc 'V1=true'
```
sumnerevans, line 29: Will this pass if `typeset -A V1=true` is in the file?

muggenhor: Ah, yes it probably will. I'll switch these to something like `^V1=true$`.

But I won't be able to do that for `V5` and `V6` because those contain `(` and `[` characters that I may or may not need to escape depending on which regex dialect is used by grep. And I'm unable to control that as that's handled by nmt...

muggenhor: Hmm, looks I could address this for `V5` at least by wrapping the braces in square brackets as an escaping mechanism independent of dialect (`[(]` and `[)]`) (works in POSIX basic, POSIX extended and Perl compatible regular expressions apparently).

modules/lib/zsh.nix
```diff
@@ -9,11 +9,20 @@ rec {
...
+      })"
     else
       ''"${toString v}"'';
 
   # Produces a Zsh shell like definition statement
-  define = n: v: "${n}=${toZshValue v}";
+  define = n: v:
+    "${lib.optionalString (builtins.isAttrs v) "typeset -A "}${n}=${
+      toZshValue v
+    }";
```
ambroisie, line 25: Does it work with `exportAll`? If not, can we make it work with it?

muggenhor: I don't see how as `export` is for environment variables and those are exclusively strings. And this deals with associative arrays, which are not strings.

Note, even though `typeset` has the `-x` option to "mark for automatic export to the environment of subsequently executed commands", it doesn't actually do anything for associative arrays. Likely because, without defining some serialization format, that doesn't make sense.

Example that demonstrates the difference between using `-x` for exporting to the environment:

```shell
$ typeset -x STR=regular-string-content
$ typeset -p STR # makes ZSH dump a command string that would duplicate this variable
export STR=regular-string-content
$ printenv STR
regular-string-content
$ echo $?
0
$ typeset -x -A ASSOC=([a]=mooh)
typeset -Ax ASSOC=( [a]=mooh )
$ printenv ASSOC
$ echo $?
1
```

ambroisie: > those are exclusively strings

... Yeah, I had a brain fart there.

Thanks for the long answer though, TIL about `typeset -x`.

PR #5259, opened 2024-04-12
fd: add module

uncenter:
### Description

`fd` is a simple, fast and user-friendly alternative to `find`.

This module includes support for the [`.fdignore`/`$XDG_CONFIG_HOME/fd/ignore`](https://github.com/sharkdp/fd#excluding-specific-files-or-directories) global ignore file (`ignores`), the search hidden files and directories `--hidden` option (`hidden`), and `extraOptions` for additional CLI flags.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

N/A

### Additional comments

I've refrained from adding myself to `modules/lib/maintainers.nix` as I have an upstream PR that adds myself as a maintainer in nixpkgs: https://github.com/NixOS/nixpkgs/pull/303399/files#diff-c1fcec3afa1da96245de1d7954ca721c979a5f25aee0e8b59d59d3d21e31f8fa.

### DIFF
--- a/modules/programs/fd.nix
+++ b/modules/programs/fd.nix
@@ -1,0 +1,59 @@
+{ config, lib, pkgs, ... }:
+with lib; {
+  meta.maintainers = [ maintainers.uncenter ];
+
+  options.programs.fd = {
+    enable = mkEnableOption
+      "fd, a simple, fast and user-friendly alternative to {command}`find`";
+
+    ignores = mkOption {
+      type = types.listOf types.str;
+      default = [ ];
+      example = [ ".git/" "*.bak" ];
+      description = "List of paths that should be globally ignored.";
+    };
+
+    hidden = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Search hidden files and directories ({option}`--hidden` argument).
+      '';
+    };
+
+    extraOptions = mkOption {
+      type = types.listOf types.str;
+      default = [ ];
+      example = [ "--no-ignore" "--absolute-path" ];
+      description = ''
+        Extra command line options passed to fd.
+      '';
+    };
+
+    package = mkPackageOption pkgs "fd" { };
+  };
+
+  config = let
+    cfg = config.programs.fd;
+
+    args = escapeShellArgs (optional cfg.hidden "--hidden" ++ cfg.extraOptions);
+
+    optionsAlias = { fd = "fd ${args}"; };
+  in mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    programs.bash.shellAliases = optionsAlias;
+
+    programs.zsh.shellAliases = optionsAlias;
+
+    programs.fish.shellAliases = optionsAlias;
+
+    programs.ion.shellAliases = optionsAlias;
+
+    programs.nushell.shellAliases = optionsAlias;
+
+    xdg.configFile."fd/ignore" = mkIf (cfg.ignores != [ ]) {
+      text = concatStringsSep "\n" cfg.ignores + "\n";
+    };
+  };
+}
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -1476,6 +1476,13 @@ in {
           A new module is available: 'programs.bun'.
         '';
       }
+
+      {
+        time = "2024-04-18T22:30:49+00:00";
+        message = ''
+          A new module is available: 'programs.fd'.
+        '';
+      }
     ];
   };
 }
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -86,6 +86,7 @@ let
     ./programs/emacs.nix
     ./programs/eww.nix
     ./programs/eza.nix
+    ./programs/fd.nix
     ./programs/feh.nix
     ./programs/firefox.nix
     ./programs/fish.nix


rycee:
Looks good. Thanks for the contribution! I've merged the cargo-shear PR and this can be merged once that PR moves to NixOS unstable (https://nixpk.gs/pr-tracker.html?pr=303399).

modules/programs/fd.nix
```diff
@@ -0,0 +1,77 @@
...
+    programs.fish.shellAliases = optionsAlias;
+
+    programs.ion.shellAliases = optionsAlias;
+
+    programs.nushell.shellAliases = optionsAlias;
+
+    xdg.configFile."fd/ignore" = mkIf (cfg.ignores != [ ]) {
+      text = concatStringsSep "\n" cfg.ignores + "\n";
+    };
+  };
```
loicreynier, line 76:
```suggestion
  config = let
    cfg = config.programs.fd;

    args = escapeShellArgs (optional cfg.hidden "--hidden" ++ cfg.extraOptions);

    ignores = mkIf (cfg.ignores != [ ]) {
      text = concatStringsSep "\n" cfg.ignores + "\n";
    };

    optionsAlias = { fd = "fd ${args}"; };
  in mkIf cfg.enable (mkMerge [
    {
      home.packages = [ cfg.package ];

      programs.bash.shellAliases = optionsAlias;

      programs.zsh.shellAliases = optionsAlias;

      programs.fish.shellAliases = optionsAlias;

      programs.ion.shellAliases = optionsAlias;

      programs.nushell.shellAliases = optionsAlias;
    }
    (mkIf (!config.home.preferXdgDirectories) {
      home.file.".fdignore".text = ignores;
    })
    (mkIf (!config.home.preferXdgDirectories) {
      xdg.configFile."fd/ignore" = ignores;
    })
  ]);
```

loicreynier: We could add something like this to choose between `~/.fdignore`  and `$XDG_CONFIG_HOME/fd/ignore` using `config.home.preferXdgDirectories`.

uncenter: From `fd`'s man page:

> ### `XDG_CONFIG_HOME`, `HOME`
> Used  to  locate the global ignore file. If `XDG_CONFIG_HOME` is set, use `$XDG_CONFIG_HOME/fd/ignore`.  Otherwise, use `$HOME/.config/fd/ignore`.

I'm fine with that I guess but we should write to `home.file.".config/fd/ignore"` instead?

rycee: I guess you could have
``` nix
let configPath = if config.home.preferXdgDirectories then "${config.xdg.configHome}/fd/ignore" else ".fdignore";
```

and then do

``` nix
home.file.${configPath}.text = ignores;
```

loicreynier:
```suggestion
  config = let
    cfg = config.programs.fd;

    args = escapeShellArgs (optional cfg.hidden "--hidden" ++ cfg.extraOptions);

    optionsAlias = { fd = "fd ${args}"; };

    ignores = mkIf (cfg.ignores != [ ]) {
      text = concatStringsSep "\n" cfg.ignores + "\n";
    };

    configPath = if config.home.preferXdgDirectories then
      "${config.xdg.configHome}/fd/ignore"
    else
      ".fdignore";
  in mkIf cfg.enable {
    home.packages = [ cfg.package ];

    programs.bash.shellAliases = optionsAlias;

    programs.zsh.shellAliases = optionsAlias;

    programs.fish.shellAliases = optionsAlias;

    programs.ion.shellAliases = optionsAlias;

    programs.nushell.shellAliases = optionsAlias;

    home.file.${configPath}.text = ignores;
  };
}
```

This indeed looks cleaner.

But if `fd` requires the `XDG_CONFIG_HOME` variable to look for `~/.config/fd/ignores` I guess we should also check for `config.xdg.enable`? Would not it be better to always enable `config.xdg.enable` when setting `home.preferXdgDirectories = true`?

modules/programs/fd.nix
```diff
@@ -0,0 +1,77 @@
...
+
+    enableFishIntegration = mkEnableOption "Fish integration" // {
+      default = true;
+    };
+
+    enableIonIntegration = mkEnableOption "Ion integration" // {
+      default = true;
+    };
+
+    enableNushellIntegration = mkEnableOption "Nushell integration";
```
ambroisie, line 25: Not enabled by default?

modules/programs/fd.nix
ambroisie: > Used to locate the global ignore file. If XDG_CONFIG_HOME is set, use $XDG_CONFIG_HOME/fd/ignore. Otherwise, use $HOME/.config/fd/ignore.

That's just following the XDG spec no? So always storing it using `xdg.configFile` should just work.

IMO, `preferXdgDirectories` should be avoided in most cases, and used for programs that have no understanding of XDG at all and need e.g: an alias to change the config location.

Anyways, I don't even see `~/.fdignore` mentionned in the manual, just `~/.config/fd/ignore` (the default `XDG_CONFIG_HOME` location).

uncenter: > Anyways, I don't even see `~/.fdignore` mentionned in the manual, just `~/.config/fd/ignore` (the default `XDG_CONFIG_HOME` location).

That's what I've been trying to say! 😅

isabelroses: The [documentation](https://github.com/sharkdp/fd#excluding-specific-files-or-directories) only mentions the xdg config home location. So I think this comment can be safely ignored.

loicreynier: Oh yes indeed! I misunderstood your comment @uncenter, ignore my comments.

isabelroses (approved):

uncenter:
> Looks good. Thanks for the contribution! I've merged the cargo-shear PR and this can be merged once that PR moves to NixOS unstable ([nixpk.gs/pr-tracker.html?pr=303399](https://nixpk.gs/pr-tracker.html?pr=303399)).

Based on that tracker link it looks like that cargo-shear PR made it to unstable! Anything else I need to do before merging?

modules/programs/fd.nix
```diff
@@ -0,0 +1,79 @@
...
+    };
+
+    extraOptions = mkOption {
+      type = types.listOf types.str;
+      default = [ ];
+      example = [ "--no-ignore" "--absolute-path" ];
+      description = ''
+        Extra command line options passed to fd.
+      '';
+    };
```
isabelroses, line 51: Might be better to do something more like ripgrep
https://github.com/nix-community/home-manager/blob/fa8c16e2452bf092ac76f09ee1fb1e9f7d0796e7/modules/programs/ripgrep.nix#L17-L28

uncenter: I'm going with what the `eza` module did, by having one of the rather important options separate.

modules/programs/fd.nix
```diff
@@ -0,0 +1,79 @@
...
+      default = true;
+    };
+
+    enableIonIntegration = mkEnableOption "Ion integration" // {
+      default = true;
+    };
+
+    enableNushellIntegration = mkEnableOption "Nushell integration" // {
+      default = true;
+    };
```
rycee, line 27: Just noticed that none of these options seem to actually do anything. Perhaps the intent is to use `mkIf` for the various `shellAliases` options?

uncenter: Oops! Thanks for pointing that out - can't believe I didn't notice. I originally based this off of the `eza` module, but now looking at that one it seems that `eza ${args}` is always aliased no matter if `enable___Integration` is enabled, and the `enable___Integration` options are used to alias things like `ls` and `ll` and so on. I think we can just remove all of those options in that case?

rycee: Yeah, can remove them.

uncenter:
Ah sorry I didn't know I was supposed to force push in this repository too!

rycee:
No worries. Merged to master now, thanks! 🙂

PR #5267, opened 2024-04-14
neomutt: allow binds to override vimKeys

gbirke:
### Description
In the generated neomutt configuration, source the optional `vim-keys.rc` before applying the `binds` configuration, to allow the user to override keybindings from `vim-keys.rc`.

### Checklist

- [X] Test cases updated/added. 

    ```
    {component}: {description}

    {long description}
    ```


rycee:

PR #5272, opened 2024-04-15
generators.toKDL: support repeated nodes, break JiK compat

jrobsonchase:
Replaces the list attr -> KDL conversion logic with a more flexible approach, allowing for multiple nodes with the same name in a scope. This unfortunately also breaks the existing JSON-in-KDL semantics in favor of ergonomics. As far as I can tell though, zellij is the only program using it, and it doesn't accept JiK anyway.

For example, the following KDL was previously impossible to generate, since nix attrs were mapped 1:1 to KDL nodes:
```kdl
resize {
	bind "k" "Up" { Resize "Increase Up"; }
	bind "j" "Down" { Resize "Increase Down"; }
}
```

Now, this can be achieved with the nix expression:

```nix
resize.bind = [
	{ _args = ["k" "Up"]; Resize = "Increase Up"; }
	{ _args = ["j" "Down"]; Resize = "Increase Down"; }
];
```

which would previously have generated the not-very-useful:

```kdl
resize {
	bind {
		- "k" "Up" { Resize "Increase Up"; }
		- "j" "Down" { Resize "Increase Down"; }
	}
}
```

which, in turn, can now be generated via:

```nix
resize.bind."-" = [
	{ _args = ["k" "Up"]; Resize = "Increase Up"; }
	{ _args = ["j" "Down"]; Resize = "Increase Down"; }
];
```

### Checklist

- [ ] ~Change is backwards compatible.~
  It's explicitly not. I'm not sure that the previous behavior that's being broken was actually usable, so it doesn't seem worth keeping intact at the cost of ergonomics.

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@h7x4 
@mainrs

### DIFF
--- a/modules/lib/generators.nix
+++ b/modules/lib/generators.nix
@@ -68,10 +68,8 @@
         in "${name} ${concatStringsSep " " flatElements}";
 
       # String -> ListOf Anything -> String
-      convertListOfNonFlatAttrsToKDL = name: list: ''
-        ${name} {
-        ${indentStrings (map (x: convertAttributeToKDL "-" x) list)}
-        }'';
+      convertListOfNonFlatAttrsToKDL = name: list:
+        "${concatStringsSep "\n" (map (x: convertAttributeToKDL name x) list)}";
 
       # String -> ListOf Anything  -> String
       convertListToKDL = name: list:
--- a/tests/lib/generators/tokdl-result.txt
+++ b/tests/lib/generators/tokdl-result.txt
@@ -27,15 +27,11 @@ listInAttrsInList {
 		}
 	}
 	list2 {
-		- {
-			a 8
-		}
+		a 8
 	}
 }
-nested {
-	- 1 2
-	- true false
-	- 
-	- null
-}
+repeated 1 2
+repeated true false
+repeated 
+repeated null
 unsafeString " \" \n 	 "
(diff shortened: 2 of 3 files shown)


jrobsonchase:
Whoops, should've searched a little harder. Looks like KDL improvements were in progress over in #4614 at some point, but the efforts may have stalled?

mainrs:
I'll take a look next weekend!

mdrssv:
Is there any help needed to further the effort?

mainrs:
This seems to be a sensible change. I am not too sure on how to handle the merge. Since this technically breaks existing configurations, doesn't it? As you mentioned though, previous _valid_ configurations could only have been ones that do not use the above KDL syntax (I think). It's better safe to be sorry, though.

mainrs (changes requested):
Could you add a test that specifically tests for repeated nodes as in the examples you posted with the bind syntax?

One that generates something like this:

```kdl
resize {
	bind "k" "Up" { Resize "Increase Up"; }
	bind "j" "Down" { Resize "Increase Down"; }
}
```

khaneliman:
https://github.com/nix-community/home-manager/pull/7429 addresses already? Resolve conflicts and reopen if i'm wrong.

Issue #5274, opened 2024-04-15
Remove `systemd` from `home-manager`?


drupol:
Hello,

Today I was looking for improving and reducing the size of Home Manager profile's closure.
During my investigation, I noticed that `systemd` is included by default, even in the smallest Home Manager profile.

With the smallest profile I could make, I can't go under 200M.

`systemd` on its own is 165M:

```
/nix/store/3np3qw5y5xarl4hxbhk9vj2d5kmgqsir-systemd-255.2               34.7M  165.3M
```

I use Home-Manager in a Docker instance where I only need to trigger the installation of the profile once and never use Home Manager later on. 

Therefore, I open this issue to discuss of the possibility to enable or disable `systemd`.

If it is already possible, please let me know.

drupol:
To disable `systemd` with configuration:

```
  home.activation.reloadSystemd = lib.mkForce "";
```

Samasaur1:
Out of curiosity, I tried to make a minimal home-manager config on Darwin, and I was able to get it down to 50.3M. And home-manager on Darwin definitely doesn't use systemd and doesn't seem to use launchd to activate configurations, so presumably this would be possible on Linux as well?

drupol:
I opened a PR for that @ https://github.com/nix-community/home-manager/pull/5276

Feedback is very welcome!

sumnerevans:
Closed by # #5276. Next time, please include "Closes #ISSUE_NUM" in the description text of the PR so that it automatically closes the corresponding issue.

drupol:
You're right I'll double pay attention, thanks for doing it.

PR #5276, opened 2024-04-15
systemd: add `enable` option

drupol:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/systemd.nix
+++ b/modules/systemd.nix
@@ -5,8 +5,8 @@ let
   cfg = config.systemd.user;
 
   inherit (lib)
-    any attrValues getAttr hm isBool literalExpression mkIf mkMerge mkOption
-    types;
+    any attrValues getAttr hm isBool literalExpression mkIf mkMerge
+    mkEnableOption mkOption types;
 
   settingsFormat = pkgs.formats.ini { listsAsDuplicateKeys = true; };
 
@@ -105,6 +105,11 @@ in {
 
   options = {
     systemd.user = {
+      enable = mkEnableOption "the user systemd service manager" // {
+        default = pkgs.stdenv.isLinux;
+        defaultText = literalExpression "pkgs.stdenv.isLinux";
+      };
+
       systemctlPath = mkOption {
         default = "${pkgs.systemd}/bin/systemctl";
         defaultText = literalExpression ''"''${pkgs.systemd}/bin/systemctl"'';
@@ -286,7 +291,12 @@ in {
   # If we run under a Linux system we assume that systemd is
   # available, in particular we assume that systemctl is in PATH.
   # Do not install any user services if username is root.
-  config = mkIf (pkgs.stdenv.isLinux && config.home.username != "root") {
+  config = mkIf (cfg.enable && config.home.username != "root") {
+    assertions = [{
+      assertion = pkgs.stdenv.isLinux;
+      message = "This module is only available on Linux.";
+    }];
+
     xdg.configFile = mkMerge [
       (lib.listToAttrs ((buildServices "service" cfg.services)
         ++ (buildServices "slice" cfg.slices)


modules/systemd.nix
```diff
@@ -105,6 +105,14 @@ in {
 
   options = {
     systemd.user = {
+      enable = mkOption {
+        default = true;
```
bandithedoge, line 109: Should be `pkgs.stdenv.isLinux`

modules/systemd.nix
```diff
@@ -286,71 +294,73 @@ in {
...
-          else
-            "";
-        in ''
-          ${pkgs.sd-switch}/bin/sd-switch \
-            ''${DRY_RUN:+--dry-run} $VERBOSE_ARG ${timeoutArg} \
-            ''${oldGenPath:+--old-units $oldGenPath/home-files/.config/systemd/user} \
-            --new-units $newGenPath/home-files/.config/systemd/user
-        '';
-      };
+  config = mkIf (cfg.enable == true && pkgs.stdenv.isLinux
```
bandithedoge, line 297: `pkgs.stdenv.isLinux` should probably be moved to an assertion

drupol: What do you mean by that, do you have an example?

drupol: I did something, let me know if it's OK

modules/systemd.nix
drupol: Done.

modules/systemd.nix
bandithedoge: Yes, that's good

modules/systemd.nix
```diff
@@ -105,6 +105,14 @@ in {
 
   options = {
     systemd.user = {
+      enable = mkOption {
+        default = pkgs.stdenv.isLinux;
+        type = types.bool;
+        description = ''
+          Whether to enable the user systemd service manager.
+        '';
+      };
```
sumnerevans, line 114:
```suggestion
      enable = mkEnableOption "the user systemd service mangaer" // {
        default = pkgs.stdenv.isLinux;
      };
```

sumnerevans (changes requested):
What is the use-case for this? What situations would you not want systemd to be configured?

drupol:
> What is the use-case for this? What situations would you not want systemd to be configured?

I explained it here: https://github.com/nix-community/home-manager/issues/5274

modules/systemd.nix
drupol: Nice, I fixed it.

modules/systemd.nix
```diff
@@ -105,6 +105,10 @@ in {
 
   options = {
     systemd.user = {
+      enable = mkEnableOption "The user systemd service manager" // {
```
rycee, line 108:
```suggestion
      enable = mkEnableOption "the user systemd service manager" // {
```

modules/systemd.nix
```diff
@@ -286,7 +290,12 @@ in {
   # If we run under a Linux system we assume that systemd is
   # available, in particular we assume that systemctl is in PATH.
   # Do not install any user services if username is root.
-  config = mkIf (pkgs.stdenv.isLinux && config.home.username != "root") {
+  config = mkIf (cfg.enable == true && config.home.username != "root") {
```
rycee, line 293:
```suggestion
  config = mkIf (cfg.enable && config.home.username != "root") {
```

modules/systemd.nix
```diff
@@ -105,6 +105,10 @@ in {
 
   options = {
     systemd.user = {
+      enable = mkEnableOption "The user systemd service manager" // {
+        default = pkgs.stdenv.isLinux;
```
rycee, line 109: To avoid documentation dependency on pkgs:

```suggestion
        default = pkgs.stdenv.isLinux;
        defaultText = literalExpression "pkgs.stdenv.isLinux";
```

rycee:

drupol:
Thank you!

PR #5282, opened 2024-04-16
experimental(generators.toKDL): brand new ugly (but correct) kdl structure

jrobsonchase:
### Description

Represent KDL more accurately in nix, complete with ordering and optional typing.

I don't expect any sane person to actually write their KDL documents in this structure, but it makes a decent target for higher-level expressions to generate.

### Checklist

- [ ] ~Change is backwards compatible.~
  Nope.

    ```
    {component}: {description}

    {long description}
    ```

#### Prior Art

My previous attempt: #5272
An earlier attempt: #4614

### DIFF
--- a/modules/lib/generators.nix
+++ b/modules/lib/generators.nix
@@ -1,10 +1,21 @@
 { lib }:
 
 {
+
   toKDL = { }:
     let
-      inherit (lib) concatStringsSep splitString mapAttrsToList any;
+      inherit (lib) concatStringsSep;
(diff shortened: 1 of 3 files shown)


mainrs:
IS this supposed to superseed the other PR you have open?

jrobsonchase:
It was originally intended to, yes. I still think I like it better, despite it being excessively verbose. But that's kinda what you get when effectively trying to represent an HTML/XML-like language in nix :upside_down_face:

PR #5304, opened 2024-04-19
nh: add module

JohnRTitor:
### Description
nh is a Nix CLI helper designed to provide beautiful outputs when rebuilding system or garbage collecting. Module already exists for NixOS, and so should for home manager too.

This PR
- adds `programs.nh` option
- Add nh to news
- Add johnrtitor as maintainer

### Checklist

#### Maintainer CC

initial PR, I'll be maintaining this HM module

### DIFF
--- a/modules/programs/nh.nix
+++ b/modules/programs/nh.nix
@@ -1,0 +1,94 @@
+{ config, osConfig, lib, pkgs, ... }:
+
+let
+
+  cfg = config.programs.nh;
+
+in {
+  meta.maintainers = with lib.maintainers; [ johnrtitor ];
+
+  options.programs.nh = {
+    enable = lib.mkEnableOption "nh, yet another Nix CLI helper";
+
+    package = lib.mkPackageOption pkgs "nh" { };
+
+    flake = lib.mkOption {
+      type = lib.types.nullOr lib.types.path;
+      default = null;
+      description = ''
+        The path that will be used for the {env}`FLAKE` environment variable.
+
+        {env}`FLAKE` is used by nh as the default flake for performing actions,
+        like {command}`nh os switch`.
+      '';
+    };
+
+    clean = {
+      enable = lib.mkEnableOption ''
+        periodic garbage collection for user profile and nix store with nh clean
+        user'';
+
+      dates = lib.mkOption {
+        type = lib.types.singleLineStr;
+        default = "weekly";
+        description = ''
+          How often cleanup is performed.
+
+          The format is described in {manpage}`systemd.time(7)`.
+        '';
+      };
+
+      extraArgs = lib.mkOption {
+        type = lib.types.singleLineStr;
+        default = "";
+        example = "--keep 5 --keep-since 3d";
+        description = ''
+          Options given to nh clean when the service is run automatically.
+
+          See `nh clean all --help` for more information.
+        '';
+      };
+    };
+  };
+
+  config = {
+    warnings = lib.optionals (!(cfg.clean.enable -> !osConfig.nix.gc.automatic))
+      [
+        "programs.nh.clean.enable and nix.gc.automatic (system-wide in configuration.nix) are both enabled. Please use one or the other to avoid conflict."
+      ];
+
+    assertions = [{
+      assertion = (cfg.flake != null) -> !(lib.hasSuffix ".nix" cfg.flake);
+      message = "nh.flake must be a directory, not a nix file";
+    }];
+
+    home = lib.mkIf cfg.enable {
+      packages = [ cfg.package ];
+      sessionVariables = lib.mkIf (cfg.flake != null) { FLAKE = cfg.flake; };
+    };
+
+    systemd.user = lib.mkIf cfg.clean.enable {
+      services.nh-clean = {
+        Unit.Description = "Nh clean (user)";
+
+        Service = {
+          Type = "oneshot";
+          ExecStart =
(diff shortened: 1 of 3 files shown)


modules/programs/nh.nix
```diff
@@ -0,0 +1,87 @@
...
+        description = "Nh clean (user)";
+        script =
+          "exec ${lib.getExe cfg.package} clean user ${cfg.clean.extraArgs}";
+        startAt = cfg.clean.dates;
+        path = [ config.nix.package ];
+        serviceConfig.Type = "oneshot";
+      };
+
+      timers.nh-clean = { timerConfig = { Persistent = true; }; };
+    };
```
rycee, line 85: I assume this is coming from some NixOS module. It will not work on HM since it does not use the same option naming for systemd units.

I would suggest creating some test case that enables the nh-clean service and then tests the content of the generated unit files.

modules/lib/maintainers.nix
```diff
@@ -101,6 +101,12 @@
     githubId = 1204734;
     name = "Emil Karlson";
   };
+  johnrtitor = {
+    email = "masumrezarock100@gmail.com";
+    github = "johnrtitor";
+    githubId = 50095635;
+    name = "Masum Reza";
+  };
```
rycee, line 109: You already seem to be present in the Nixpkgs maintainers list, in which case you don't need to add yourself here as well.

```suggestion
```

modules/programs/nh.nix
```diff
@@ -0,0 +1,87 @@
...
+        "programs.nh.clean.enable and nix.gc.automatic (system-wide in configuration.nix) are both enabled. Please use one or the other to avoid conflict."
+      ];
+
+    assertions = [
+      # Not strictly required but probably a good assertion to have
+      {
+        assertion = cfg.clean.enable -> cfg.enable;
+        message = "programs.nh.clean.enable requires programs.nh.enable";
+      }
+
```
rycee, line 62: I think it is best to remove this assertion, it is a common use case to do all setup of a module in a common Nix file and then enable the module in a user specific configuration. This assertion would prevent such usage.

```suggestion
      # Not strictly required but probably a good assertion to have
      {
        assertion = cfg.clean.enable -> cfg.enable;
        message = "programs.nh.clean.enable requires programs.nh.enable";
      }

```

rycee:

modules/programs/nh.nix
JohnRTitor: Done, Thanks. Removed the assertion.

modules/lib/maintainers.nix
JohnRTitor: Done, thanks :)

modules/programs/nh.nix
JohnRTitor: Thanks for the suggestions. Indeed this is coming from a NixOS module.
I am trying to recreate the following in Home Manager syntax. How do I set these? LOCALE_ARCHIVE, variable, for example. I am new to NixOS and Nix syntax.

The following is the content of nh-clean.service.

```
[Service]
Environment="LOCALE_ARCHIVE=/nix/store/zbn9jsvqg4gp6dsa01gmfwald132fk49-glibc-locales-2.39-31/lib/locale/locale-archive"
Environment="PATH=/nix/store/5dyaxm5ksj6956xbnssyr2pwm4bq9jkf-nix-2.18.2/bin:/nix/store/hq8765g3p1i7qbargnqli5mn0jpsdbfl-coreutils-9.5/bin:/nix/store/d1v47ybpl5cv9ycffgbfagfhvbvj8xdx-findutils-4.9.0/bin:/nix/store/wy37jk2hirzqzx0666w1849kjdgzdam6-gnugrep-3.11/bin:/nix/store/4cps736z7in3d37qc801lwv9z0ib67ps-gnused-4.9/bin:/nix/store/7kq9hvrhcy8g1v0jd07nr54279asg9vc-systemd-255.4/bin:/nix/store/5dyaxm5ksj6956xbnssyr2pwm4bq9jkf-nix-2.18.2/sbin:/nix/store/hq8765g3p1i7qbargnqli5mn0jpsdbfl-coreutils-9.5/sbin:/nix/store/d1v47ybpl5cv9ycffgbfagfhvbvj8xdx-findutils-4.9.0/sbin:/nix/store/wy37jk2hirzqzx0666w1849kjdgzdam6-gnugrep-3.11/sbin:/nix/store/4cps736z7in3d37qc801lwv9z0ib67ps-gnused-4.9/sbin:/nix/store/7kq9hvrhcy8g1v0jd07nr54279asg9vc-systemd-255.4/sbin"
Environment="TZDIR=/nix/store/7lzfzr3xsdgscccnfl9rykiwncwwvhbi-tzdata-2024a/share/zoneinfo"
ExecStart=/nix/store/8454db7gv30ls496k5cam45j1amykrnn-unit-script-nh-clean-start/bin/nh-clean-start
Type=oneshot
```
I was able to convert the timers config :)

```nix
      timers.nh-clean = {
        Timer = {
          OnCalendar = cfg.clean.dates;
          Persistent = true;
        };
      };
```

Or we can simply remove the service part and get it going.

JohnRTitor: @rycee any suggestions?

JohnRTitor:
CC @teto 
the tests failure seems unrelated to this module

modules/programs/nh.nix
JohnRTitor: Done.

Issue #5307, opened 2024-04-20
bug: sway is broken with custom packages (like swayfx) coming from the nixos configuration


ElysaSrc:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

With the changes inside this commit: https://github.com/nix-community/home-manager/commit/3a435342e2e5e2bff1d3331c2bd9e70f25693ea2

If you have a programs.sway with custom configurations at nixos level, and you put a `package = null` to use the os level package, it cannot validates the configuration since it will call the wrong pkgs.sway.

To reproduce, you can have this in your nixos configuration.nix file:

```
    programs = {
      sway = {
        enable = true;
        package = pkgs.swayfx;
        extraSessionCommands = ''
          export SDL_VIDEODRIVER=wayland
          export MOZ_ENABLE_WAYLAND=1
          export NIXOS_OZONE_WL=1
          export QT_QPA_PLATFORM=wayland-egl
          export QT_WAYLAND_DISABLE_WINDOWDECORATION="1"
          export _JAVA_AWT_WM_NONREPARENTING=1
          export SSH_ASKPASS_REQUIRE=prefer
          export SSH_AUTH_SOCK=/run/user/1000/keyring/ssh
        '';
        extraPackages = [];
      };
    };
```

and if on your home you put this:

```
wayland.windowManager.sway = {
   enable = true;
   package = null;
   # ...
};
```

The configuration will fail if you use SwayFX settings, like "blur" and stuff.

Workaround : set `checkConfig = false;`.
Fix : Use correctly the sway coming from the nixos configuration itself.

### Maintainer CC

@amarshall @rycee 

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.8.6, NixOS, 24.05 (Uakari), 24.05.20240419.5c24cf2`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.18.2`
 - nixpkgs: `/nix/store/v4pcs3nzx54m5bmxd39win0rgl2d2hbx-source`
```

PR #5309, opened 2024-04-20
sway: do not use pkgs.sway when cfg.package = null

amarshall:
### Description

- Implicitly disable checkConfig when `cfg.package = null` as we don’t
  have any exe to use for the check
- Implicitly disable `swaymsg reload` on activation, since we have no
  exe to use for running it

See https://github.com/nix-community/home-manager/issues/5307

### Checklist

Need to think about if there is a good test to write for this…I did run existing Sway tests and all OK.

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

### DIFF
--- a/modules/services/window-managers/i3-sway/sway.nix
+++ b/modules/services/window-managers/i3-sway/sway.nix
@@ -331,13 +331,6 @@ let
         ++ [ cfg.extraConfig ]);
   };
 
-  defaultSwayPackage = pkgs.sway.override {
-    extraSessionCommands = cfg.extraSessionCommands;
-    extraOptions = cfg.extraOptions;
-    withBaseWrapper = cfg.wrapperFeatures.base;
-    withGtkWrapper = cfg.wrapperFeatures.gtk;
-  };
-
 in {
   meta.maintainers = with maintainers; [
     Scrumplex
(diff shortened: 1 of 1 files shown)


ElysaSrc:
Why couldn't we use the `osConfig` to get the value ?
Locally, to keep both features I've done this:

ElysaSrc:
> IMO, that should be something the user decides to do themselves, not something implicit in the default config. There’s also no precedent for consuming `osConfig` in HM.

I was suggesting it because the documentation says:

> Sway package to use. Will override the options ‘wrapperFeatures’, ‘extraSessionCommands’, and ‘extraOptions’. Set to null to not add any Sway package to your path. This should be done if you want to use the NixOS Sway module to install Sway.

But I do understand why we shouldn't consume it then.

I haven't had time to check the PR, I'll check it later. For now, I just use osConfig.

Scrumplex (approved):
LGTM

rycee:
Sorry about the delay. I don't use Sway so cannot do any major reviewing. It looks innocent enough, though so I've now rebased to master in a9c9cc6e50f7cbd2d58ccb1cd46a1e06e9e445ff.

About using `osConfig` in modules, we have historically been very restrictive about that because the `osConfig` argument not being present in all cases. That has recently changed but I still think it is inadvisable to make use of inside the modules to avoid making the behavior too different between standalone and embedded usage.

PR #5315, opened 2024-04-21
sway: add preCheckConfig to customize check environment

amarshall:
### Description

See commit messages.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

### DIFF
--- a/modules/services/window-managers/i3-sway/sway.nix
+++ b/modules/services/window-managers/i3-sway/sway.nix
@@ -269,8 +269,9 @@ let
     checkPhase = lib.optionalString cfg.checkConfig ''
       export DBUS_SESSION_BUS_ADDRESS=/dev/null
       export XDG_RUNTIME_DIR=$(mktemp -d)
+      cp "$target" sway.conf
       ${cfg.preCheckConfig}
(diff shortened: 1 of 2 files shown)


teto (changes requested):
not worth the complexity IMO. Either users make it work in hte sandbox or they disable the check. One thing we could do if there are too many complaints is to toggle the check off by default

teto:
> Where exactly is the complexity? It’s one new option and two lines of code to use it (plus tests).

the diff is 40 lines to modify the check such that it doesn't check the final config, turning it into a potentially useless check. If you dont use it I dont see why we should carry the burden upstream.

> One argument is that, unlike pkgs, it’s not possible to use overlays for modules, which causes frustration for folks when they want extensibility (so the only way to extend is to patch, which can be cumbersome for some).

You can perfectly add options under the same namespace via imports. You can see here https://github.com/teto/home/blob/main/hm/modules/zsh.nix how I extend the zsh module for instance.

buhl:
I think this is a simple and elegant solution giving user the freedom to workaround their limitations of an otherwise great sway hm module. Of cause i am not the maintainer so it's easy for me to see this as no real added complexity is this feature just some extra legroom for the users.

PR #5318, opened 2024-04-21
swayosd: add custom style option

SergioRibera:
### Description
Add custom style path file option

> The new option is part of this new version https://github.com/NixOS/nixpkgs/pull/305672 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@pltanton

### DIFF
--- a/modules/services/swayosd.nix
+++ b/modules/services/swayosd.nix
@@ -26,6 +26,15 @@ in {
       description = "OSD margin from top edge (0.5 would be screen center).";
     };
 
+    stylePath = mkOption {
+      type = types.nullOr types.path;
+      default = null;
+      example = "/etc/xdg/swayosd/style.css";
+      description = ''
+        Use a custom Stylesheet file instead of looking for one.
+      '';
+    };
+
     display = mkOption {
       type = types.nullOr types.str;
       default = null;
@@ -57,6 +66,8 @@ in {
           Type = "simple";
           ExecStart = "${cfg.package}/bin/swayosd-server"
             + (optionalString (cfg.display != null) " --display ${cfg.display}")
+            + (optionalString (cfg.stylePath != null)
+              " --style ${escapeShellArg cfg.stylePath}")
             + (optionalString (cfg.topMargin != null)
               " --top-margin ${toString cfg.topMargin}");
           Restart = "always";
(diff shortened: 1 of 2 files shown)


modules/services/swayosd.nix
```diff
@@ -26,6 +26,15 @@ in {
       description = "OSD margin from top edge (0.5 would be screen center).";
     };
 
+    style = mkOption {
+      type = types.nullOr types.str;
```
rycee, line 30: Perhaps
```suggestion
      type = types.nullOr types.path;
```

modules/services/swayosd.nix
```diff
@@ -57,6 +66,7 @@ in {
           Type = "simple";
           ExecStart = "${cfg.package}/bin/swayosd-server"
             + (optionalString (cfg.display != null) " --display ${cfg.display}")
+            + (optionalString (cfg.style != null) " --style \"${cfg.style}\"")
```
rycee, line 69: Should be more robust to use `escapeShellArg`:
```suggestion
            + (optionalString (cfg.style != null) " --style ${escapeShellArg cfg.style}")
```

modules/services/swayosd.nix
```diff
@@ -26,6 +26,15 @@ in {
       description = "OSD margin from top edge (0.5 would be screen center).";
     };
 
+    style = mkOption {
```
rycee, line 29: Perhaps more clear to include "path" in the name:

```suggestion
    stylePath = mkOption {
```

rycee: Alternatively, can leave it as `style` and use the type `types.lines`. The user could then either provide the style inline, or provide it through a file through `builtins.readFile`.

SergioRibera: > Alternatively, can leave it as `style` and use the type `types.lines`. The user could then either provide the style inline, or provide it through a file through `builtins.readFile`.

I was just thinking about keeping both, but the `style` inline would have to store it in a file and provide that path, I think this would complicate things a bit because it raises the question of `Where do we store the content?`, perhaps where we store it by default won't be appreciated by the users, for now I'll resolve the other suggestions.

rycee: In the context of Nix, the answer to the "Where do we store the content?" is always "the Nix store" 🙂 With pretty much the only exception being sensitive information such as passwords.

In any case, having `stylePath` is fine. If somebody wants to merge multiple style definitions, then perhaps can consider switching to lines in the future.

SergioRibera:
Sorry, I already adjusted the formatting

rycee:

PR #5319, opened 2024-04-22
foot: unset PATH in server's systemd unit file

### LINKED ISSUE #5302: bug: footclient has wrong path.

### Issue description

Hello,

when using `foot.service.enable`. All `footclient` have the wrong PATH. Especially they do not have `/run/current-system/bin` etc. 
(Which makes it impossible for me to find out, what the path actually is, since not even `echo` is working correctly)

Since `foot` and `footclient` are the only(?) command working I think the culprit is #5254  
However I was not able to check this. 

### Maintainer CC

@brynedwards @plabadens 

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.6.27, NixOS, 24.05 (Uakari), 24.05.20240416.66adc1e`
 - multi-user?: `yes`
 - sandbox: `yes`
 - nixpkgs: `/nix/store/6bf943g0mlhymyffs9dflzgc78r93s60-source`
```
```

michaelBelsanti:
This reverts commit 40ab43ae98cb3e6f07eaeaa3f3ed56d589da21b0 #5254 

Fixes #5302 

### Description

This change caused footclient to launch with the set path on some systems, making the shell unusable. There may be a way to keep it working while fixing the initial problem #5254 was solving, but I think this should be done in the meantime. 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@plabadens

### DIFF
--- a/modules/programs/foot.nix
+++ b/modules/programs/foot.nix
@@ -68,7 +68,6 @@ in {
         };
 
         Service = {
-          Environment = "PATH=${makeBinPath [ cfg.package ]}";
           ExecStart = "${cfg.package}/bin/foot --server";
           Restart = "on-failure";
           OOMPolicy = "continue";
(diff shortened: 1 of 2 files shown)


rycee:

PR #5324, opened 2024-04-23
Hypridle: add module

khaneliman:
### Description

Adding hypridle module so i can remove another input from my flake and just use the config in home-manager. Meant to do a long time ago but forgot to create PR.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@fufexan

### DIFF
--- a/modules/services/hypridle.nix
+++ b/modules/services/hypridle.nix
@@ -1,0 +1,95 @@
+{ config, lib, pkgs, ... }:
+with lib;
+let
+
+  cfg = config.services.hypridle;
+in {
+  meta.maintainers = [ maintainers.khaneliman maintainers.fufexan ];
+
+  options.services.hypridle = {
+    enable = mkEnableOption "Hypridle, Hyprland's idle daemon";
+
+    package = mkPackageOption pkgs "hypridle" { };
+
+    settings = lib.mkOption {
+      type = with lib.types;
+        let
+          valueType = nullOr (oneOf [
+            bool
+            int
+            float
+            str
+            path
+            (attrsOf valueType)
+            (listOf valueType)
+          ]) // {
+            description = "Hypridle configuration value";
+          };
+        in valueType;
+      default = { };
+      description = ''
+        Hypridle configuration written in Nix. Entries with the same key
+        should be written as lists. Variables' and colors' names should be
+        quoted. See <https://wiki.hyprland.org/Hypr-Ecosystem/hypridle/> for more examples.
+      '';
+      example = lib.literalExpression ''
+        {
+          general = {
+            after_sleep_cmd = "hyprctl dispatch dpms on";
+            ignore_dbus_inhibit = false;
+            lock_cmd = "hyprlock";
+          };
+
+          listener = [
+            {
+              timeout = 900;
+              on-timeout = "hyprlock";
+            }
+            {
+              timeout = 1200;
+              on-timeout = "hyprctl dispatch dpms off";
+              on-resume = "hyprctl dispatch dpms on";
+            }
+          ];
+        }
+      '';
+    };
+
+    importantPrefixes = lib.mkOption {
+      type = with lib.types; listOf str;
+      default = [ "$" ];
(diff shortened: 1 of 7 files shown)


khaneliman:
@rycee Does the home-file config need to be separated from the systemd service config?

fufexan:
I believe it's best to reuse the nix to hyprlang translator that the hyprland module uses. Perhaps externalize it to a separate file/lib function.

That way, we won't have to constantly change the module options each time an option is added/renamed/removed.

khaneliman:
> I believe it's best to reuse the nix to hyprlang translator that the hyprland module uses. Perhaps externalize it to a separate file/lib function.
> 
> That way, we won't have to constantly change the module options each time an option is added/renamed/removed.

Makes sense, I'll take a look at that.

modules/services/hypridle.nix
```diff
@@ -0,0 +1,82 @@
...
+  meta.maintainers = [ maintainers.khaneliman maintainers.fufexan ];
+
+  options.services.hypridle = {
+    enable = mkEnableOption "Hypridle, Hyprland's idle daemon";
+
+    package = mkOption {
+      description = "The hypridle package";
+      type = types.package;
+      default = pkgs.hypridle;
+    };
```
rycee, line 16:
```suggestion
    package = mkPackageOption pkgs "hypridle" { };
```

modules/services/hypridle.nix
```diff
@@ -0,0 +1,82 @@
...
+          ];
+        }
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable {
+    xdg.configFile."hypr/hypridle.conf".text =
+      lib.optionalString (cfg.settings != { })
+      (config.lib.hyprlang.toHyprconf cfg.settings 0 true);
```
rycee, line 66: Do you still want to create the file if the settings are empty? If not, then perhaps you want
```suggestion
    xdg.configFile."hypr/hypridle.conf" = mkIf (cfg.settings != { }) {
      text = config.lib.hyprlang.toHyprconf cfg.settings 0 true;
    }
```

rycee:

modules/services/hypridle.nix
```diff
@@ -0,0 +1,78 @@
...
+in {
+  meta.maintainers = [ maintainers.khaneliman maintainers.fufexan ];
+
+  options.services.hypridle = {
+    enable = mkEnableOption "Hypridle, Hyprland's idle daemon";
+
+    package = mkPackageOption pkgs "hypridle" { };
+
+    settings = lib.mkOption {
+      type = with lib.types;
```
fufexan, line 15: Perhaps the type could be extracted into the hyprlang lib module and used from there throughout all hypr* modules.

khaneliman: I do have a larger refactor of the hyprlang module in the hyprlock PR... https://github.com/nix-community/home-manager/pull/5325/files#diff-d125ebbed352e5f662545c5f6f0ef5d1f4b301290af30f43cc091069159b7f04 wondering if i should just consolidate all of these into a single PR

khaneliman: Implemented this in all 3 prs

modules/services/hypridle.nix
```diff
@@ -0,0 +1,95 @@
...
+            (listOf valueType)
+          ]) // {
+            description = "Hypridle configuration value";
+          };
+        in valueType;
+      default = { };
+      description = ''
+        Hypridle configuration written in Nix. Entries with the same key
+        should be written as lists. Variables' and colors' names should be
+        quoted. See <https://wiki.hyprland.org> for more examples.
```
fufexan, line 33:
```suggestion
        quoted. See <https://wiki.hyprland.org/Hypr-Ecosystem/hypridle> for more examples.
```

khaneliman:
@rycee anything else we need to change in these modules or do they look good now?

rycee:

PR #5325, opened 2024-04-23
Hyprlock: add module

khaneliman:
### Description
Adding the hyprlock module, similar to https://github.com/nix-community/home-manager/pull/5324

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@fufexan

### DIFF
--- a/modules/programs/hyprlock.nix
+++ b/modules/programs/hyprlock.nix
@@ -1,0 +1,127 @@
+{ config, pkgs, lib, ... }:
+
+with lib;
+
+let
+
+  cfg = config.programs.hyprlock;
+
+in {
+  meta.maintainers = [ maintainers.khaneliman maintainers.fufexan ];
+
+  options.programs.hyprlock = {
+    enable = mkEnableOption "" // {
+      description = ''
+        Whether to enable Hyprlock, Hyprland's GPU-accelerated lock screen
+        utility.
+
+        Note that PAM must be configured to enable hyprlock to perform
+        authentication. The package installed through home-manager will *not* be
+        able to unlock the session without this configuration.
+
+        On NixOS, it can be enabled using:
+
+        ```nix
+        security.pam.services.hyprlock = {};
+        ```
+      '';
+    };
+
+    package = mkPackageOption pkgs "hyprlock" { };
+
+    settings = lib.mkOption {
+      type = with lib.types;
+        let
+          valueType = nullOr (oneOf [
+            bool
+            int
+            float
+            str
+            path
+            (attrsOf valueType)
+            (listOf valueType)
+          ]) // {
+            description = "Hyprlock configuration value";
+          };
+        in valueType;
+      default = { };
+      example = lib.literalExpression ''
+        {
+          general = {
+            disable_loading_bar = true;
+            grace = 300;
+            hide_cursor = true;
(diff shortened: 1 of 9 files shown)


khaneliman:
@rycee rebased and resolved merge conflicts

modules/programs/hyprlock.nix
```diff
@@ -0,0 +1,130 @@
...
+            }
+          ];
+        }
+      '';
+    };
+
+    extraConfig = lib.mkOption {
+      type = lib.types.lines;
+      default = "";
+      example = ''
```
andresilva, line 91: This example is unrelated to hyprlock

modules/programs/hyprlock.nix
```diff
@@ -0,0 +1,130 @@
...
+    sourceFirst = lib.mkEnableOption ''
+      putting source entries at the top of the configuration
+    '' // {
+      default = true;
+    };
+
+    importantPrefixes = lib.mkOption {
+      type = with lib.types; listOf str;
+      default = [ "$" "monitor" "size" ]
+        ++ lib.optionals cfg.sourceFirst [ "source" ];
```
andresilva, line 109: Can you actually source files in hyprlock config? (not sure if this only applies to hyprland)

andresilva (changes requested):

modules/programs/hyprlock.nix
khaneliman: https://github.com/hyprwm/hyprlock/blob/c87af3aa1f6e6bd06cffaabcc400bd45e26d565a/nix/hm-module.nix#L74

andresilva: Yep, I checked the hyprlock code, it's supported.

modules/programs/hyprlock.nix
andresilva: `extraConfig` is needed for Hyprland since you can't define submaps in the generic settings attribute. For hyprlock not sure if it's needed, but might be useful as an escape hatch to just define stuff plainly.

modules/programs/hyprlock.nix
```diff
@@ -0,0 +1,126 @@
...
+        security.pam.services.hyprlock = {};
+        ```
+      '';
+    };
+
+    package = mkOption {
+      description = "The hyprlock package";
+      type = types.package;
+      default = pkgs.hyprlock;
+    };
```
sumnerevans, line 29:
```suggestion
    package = mkPackageOption pkgs "hyprlock" { };
```

andresilva (approved):

rycee:

nyabinary:
https://github.com/hyprwm/hyprlock/blob/main/nix/hm-module.nix
Does this reach feature parity with this?

khaneliman:
> https://github.com/hyprwm/hyprlock/blob/main/nix/hm-module.nix Does this reach feature parity with this?

It is done slightly differently, that had nix named attributes and validation unique to each setting. This just validates you're passing in valid nix that can be converted into hyprlang. You can see a linked commit above where I migrated from the flake hm-module to using the home-manager module. Main difference being everything is nested under a settings attribute, the names for attributes need to match the actual hyprlock config names, and there are no default values injected.

PR #5329, opened 2024-04-23
ibus: add module

nyadiia:
### Description

Adds ibus as an IME

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/i18n/input-method/default.nix
+++ b/modules/i18n/input-method/default.nix
@@ -22,7 +22,8 @@ let
   '';
 
 in {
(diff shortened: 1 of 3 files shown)


musjj:
Been looking to try out ibus, so thanks for the PR. Is there any reason that this in a draft state?

nyadiia:
sorry for taking so long, i just don't have tests to use so that's why i stopped working on this ( and i no longer use ibus personally ). since this is mostly taken from the nixos module, it should just already work, i'll look into this more soon.

PR #5332, opened 2024-04-24
init: workarounds.nixgl module

### LINKED ISSUE #3968: feature request: nixGL integration
### Description

https://github.com/NixOS/nixpkgs/issues/9415 which tracks the problem of libGL in nix on non NixOS systems has been open for about 8 years now. Luckily, [nixGL](https://github.com/guibou/nixGL) exists which provides a wrapper around programs that require openGL (games, gpu accelerated terminals, browsers, etc.).

It would be great if there were an option to wrap all GUI applications or user-specified applications with `nixGL`. I'm assuming there's a generic way and bepoke method per application. Care should be taken for the desktop entries to call with `nixGL`

**`workarounds.opengl.packages`**

Is given a list of package names (or maybe packages themselves) that it has to wrap

**`lib.wrapWithLibGL`**

It could be a wrapper that takes a package and spits out another which is called by libGL.

michaelCTS:
### Description

The option allows packages to be wrapped by nixgl in order to make GUI applications work when installed with home-manager. The goal of this PR is just to make it easy to use graphically accelerated programs installed with home-manager in a desktop environment. 

No need to install and run `nixGL $program`. Just add the package to `workaround.nixgl.packages` like so `[ { pkg = pkgs.some_package; }]`.

Anything more advanced can come in further PRs by other people.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

Closes #3968

### DIFF
--- a/modules/misc/nixgl.nix
+++ b/modules/misc/nixgl.nix
@@ -1,0 +1,115 @@
+{ config, pkgs, lib, ... }:
+
+let
+  cfg = config.targets.genericLinux.nixgl;
+
+  # Expects a packageConfiguration hashset
+  # Adapted from https://nixos.wiki/wiki/Nix_Cookbook
+  wrapWithNixGL = pkgConfig:
+    (let
+      binary = if (builtins.hasAttr "binary" pkgConfig) then
+        pkgConfig.binary
+      else
+        null;
+      package = pkgConfig.package;
+
+      bin = if binary == null then package.pname else binary;
+      nixGL = import cfg.src { inherit pkgs; };
+      wrapperPkg = lib.attrsets.getAttrFromPath cfg.wrapperPkgPath nixGL;
+      wrapperBinPath = "${wrapperPkg}/bin/${cfg.wrapperBinName}";
+    in pkgs.runCommand "nixgl-${package.pname}" {
+      buildInputs = [ pkgs.makeWrapper ];
+    } ''
+      mkdir $out
+      # Link every top-level folder from pkgs.hello to our new target
+      ln -s ${package}/* $out
+      # Except the bin folder
+      rm $out/bin
+      mkdir $out/bin
+      # We create the bin folder ourselves and link every binary in it
+      ln -s ${package}/bin/* $out/bin
+      # Except the target binary
+      rm $out/bin/${bin}
+      # Because we create this ourself, by creating a wrapper
+      makeWrapper ${wrapperBinPath} $out/bin/${bin} \
+        --add-flags ${package}/bin/${bin}
+    '');
+  packageConfiguration = with lib;
+    types.submodule {
+      options = {
+        binary = mkOption {
+          type = types.nullOr types.str;
+          example = "alacritty";
+          default = null;
+          description = ''
+            Name of the binary in bin/ of the package.
+            By default this uses `package.pname`.
+          '';
+        };
+        package = mkOption {
+          type = types.package;
+          example = "pkgs.alacritty";
+          description = ''
+            The package to be wrapped.
+            It's expected to have a `pname` attribute.
+          '';
+        };
+      };
+    };
+in {
+  meta.maintainers = with lib.maintainers; [ michaelCTS ];
+
+  options.targets.genericLinux.nixgl = {
+
+    src = lib.mkOption {
+      type = lib.types.package;
+      example = ''
+        pkgs.fetchFromGithub {
+          owner = "nix-community";
+          repo = "nixGL";
+          rev = "v1.3";
+          hash = "SOMEHASH-HERE";
+        }
+      '';
+      description = ''
+        Path to a downloaded source of nixGL.
+        You can download the nixGL version of your choice and point to it here.
+      '';
+    };
+
+    wrapperBinName = lib.mkOption {
+      type = lib.types.str;
+      example = "nixGL";
+      default = "nixGL";
+      description =
+        "Name of the nixGL binary to be called from within the wrapper package";
+    };
+
+    wrapperPkgPath = lib.mkOption {
+      type = lib.types.listOf lib.types.str;
+      example = [ "auto" "nixGLNvidia" ];
+      default = [ "auto" "nixGLDefault" ];
+      description = "Attribute path within `src` attrset to the nixGL package";
+    };
+
+    packages = lib.mkOption {
+      type = lib.types.listOf packageConfiguration;
+      default = [ ];
+      example = "[{ package = pkgs.alacritty; }]";
+      description = ''
+        List of packages that will be wrapped with nixGL in order to be able to use nixgl.
+        Without the wrapper, graphical applications cannot access nixgl and thus have no
+        graphical acceleration.
+      '';
+    };
+
+  };
+
+  config = {
+    lib.nixGl.wrap = wrapWithNixGL;
+    home.packages = builtins.map config.lib.nixGl.wrap
+      config.targets.genericLinux.nixgl.packages;
+  };
+
+}
+
--- a/tests/default.nix
+++ b/tests/default.nix
@@ -49,6 +49,7 @@ in import nmtSrc {
     ./modules/misc/fontconfig
     ./modules/misc/manual
     ./modules/misc/nix
+    ./modules/misc/nixgl
     ./modules/misc/specialisation
     ./modules/programs/aerc
     ./modules/programs/alacritty


modules/misc/opengl.nix
```diff
@@ -0,0 +1,62 @@
+{ config, pkgs, lib, ... }:
+
+let
+  nixGL = (import (builtins.fetchTarball
+    "https://github.com/nix-community/nixGL/archive/main.tar.gz") {
+      inherit pkgs;
+    });
```
michaelCTS, line 7: This is used in the default option below, but I'm unsure how to create an option that depends on an external package and only pull the external package when a user needs it. In its current state, it always pulls it in which means `home-manager` becomes dependent on nixgl, even for users who use home-manager in a headless environment.

modules/misc/opengl.nix
```diff
@@ -0,0 +1,62 @@
...
+      example = "$${nixGL.auto.nixGLDefault}/bin/nixGL";
+      default = "${nixGL.auto.nixGLDefault}/bin/nixGL";
+    };
+    packages = lib.mkOption {
+      type = lib.types.listOf lib.types.anything;
+      default = [ ];
+      example = ''[{
+        binPathRel = "bin/alacritty";
+        pkg = pkgs.alacritty;
+      }]'';
```
michaelCTS, line 47: Dunno if a custom type is needed here of it is sufficient to use `lib.types.anything`.

michaelCTS:
Well, I tried adding tests, but for some reason this fails

```nix
{ lib, pkgs, ... }:                                                                           

with lib;

{
 config = {
  workarounds.nixgl.packages = with pkgs; [{ pkg = alacritty; }];                           
  
  nmt.script = ''
   assertFileRegex .nix-profile/bin/alacritty 'exec "/nix/store/.*/bin/nixGL" .*/alacritty'
   assertLinkExists .nix-profile/share/applications/Alacritty.desktop                      
  '';
 };
}
```

It doesn't find the files even though when I run `home-manage switch` I do see them.

modules/misc/nixgl.nix
```diff
@@ -0,0 +1,103 @@
...
+          description = ''
+            The package to be wrapped.
+            It's expected to have a `pname` attribute.
+          '';
+        };
+      };
+    };
+in {
+  meta.maintainers = with lib.maintainers; [ michaelCTS ];
+  options.workarounds.nixgl = {
```
rycee, line 55: I assume this only applies to non-NixOS systems? In which case I think the options should go under `targets.genericLinux.nixgl`.

modules/misc/nixgl.nix
```diff
@@ -0,0 +1,103 @@
...
+          rev = "v1.3";
+          hash = "SOMEHASH-HERE";
+        }
+      '';
+      description = ''
+        Path to a downloaded source of nixGL.
+        You can download the nixGL version of your choice and point to it here.
+      '';
+    };
+    wrapperBinName = lib.mkOption {
```
rycee, line 73: Please separate options by an empty line.
```suggestion

    wrapperBinName = lib.mkOption {
```

modules/misc/nixgl.nix
```diff
@@ -0,0 +1,103 @@
...
+      type = lib.types.str;
+      example = "nixGL";
+      default = "nixGL";
+      description =
+        "Name of the nixGL binary to be called from within the wrapper package";
+    };
+    wrapperPkgPath = lib.mkOption {
+      type = lib.types.listOf lib.types.str;
+      example = ''["auto" "nixGLNvidia"]'';
```
rycee, line 82:
```suggestion
      example = ["auto" "nixGLNvidia"];
```

modules/misc/nixgl.nix
```diff
@@ -0,0 +1,103 @@
...
+        };
+      };
+    };
+in {
+  meta.maintainers = with lib.maintainers; [ michaelCTS ];
+  options.workarounds.nixgl = {
+    src = lib.mkOption {
+      type = lib.types.package;
+      default = builtins.fetchTarball
+        "https://github.com/nix-community/nixGL/archive/main.tar.gz";
```
rycee, line 59: The user will have to provide this themselves. We cannot have the situation that HM downloads a random revision of some repository. Further, this will add a documentation dependency on this repository, i.e., pretty much all HM users would need to download this repository, whether or not they use nixGL.

```suggestion
```

modules/misc/nixgl.nix
```diff
@@ -0,0 +1,103 @@
...
+    };
+    wrapperBinName = lib.mkOption {
+      type = lib.types.str;
+      example = "nixGL";
+      default = "nixGL";
+      description =
+        "Name of the nixGL binary to be called from within the wrapper package";
+    };
+    wrapperPkgPath = lib.mkOption {
+      type = lib.types.listOf lib.types.str;
```
rycee, line 81: If the paths are stable in nixGL then perhaps this could be `enum [ "nixGLDefault" "nixGLNvidia" "nixGLNvidiaBumblebee" "nixGLIntel" ]`? The module can know the corresponding attribute path.

rycee: If so then can rename the option to something like `wrapperFlavor` or something.

modules/misc/nixgl.nix
rycee, line 73: Is the binary ever called something other than `nixGL`?

modules/misc/nixgl.nix
```diff
@@ -0,0 +1,103 @@
...
+      mkdir $out/bin
+      # We create the bin folder ourselves and link every binary in it
+      ln -s ${pkg}/bin/* $out/bin
+      # Except the target binary
+      rm $out/bin/${bin}
+      # Because we create this ourself, by creating a wrapper
+      makeWrapper ${wrapperBinPath} $out/bin/${bin} \
+        --add-flags ${pkg}/bin/${bin}
+    '');
+  PackageConfiguration = with lib;
```
rycee, line 31:
```suggestion
  packageConfiguration = with lib;
```

modules/misc/nixgl.nix
```diff
@@ -0,0 +1,103 @@
...
+        List of packages that will be wrapped with nixGL in order to be able to use nixgl.
+        Without the wrapper, graphical applications cannot access nixgl and thus have no
+        graphical acceleration.
+      '';
+    };
+  };
+
+  config = {
+    home.packages =
+      builtins.map wrapWithNixGL config.workarounds.nixgl.packages;
```
rycee, line 100: Instead of wrapping each package individually, perhaps it would be less work to simply wrap the binaries in the final profile package. Not sure if it is a good idea. But I guess it would look something like this:

``` nix
home.packages = cfg.packages;
home.extraProfileCommands = ''
  for pkg in ${escapeShellArgs cfg.packages}; do
    for exe in "$out"/bin/*; do
      ${wrapWithNixGL} "$out/bin/$(basename "$exe")"
    done
  done
'';
```

rycee: Ah, right you also want to wrap only a specific path in bin, then perhaps it is not worth using `extraProfileCommands`.

modules/misc/nixgl.nix
```diff
@@ -0,0 +1,103 @@
...
+        binary = mkOption {
+          type = types.nullOr types.str;
+          example = "alacritty";
+          default = null;
+          description = ''
+            Name of the binary in bin/ of the package.
+            By default this uses `package.pname`.
+          '';
+        };
+        pkg = mkOption {
```
rycee, line 43:
```suggestion

        package = mkOption {
```

modules/misc/nixgl.nix
```diff
@@ -0,0 +1,103 @@
...
+          type = types.package;
+          example = "pkgs.alacritty";
+          description = ''
+            The package to be wrapped.
+            It's expected to have a `pname` attribute.
+          '';
+        };
+      };
```
rycee, line 51: Can just set the binary option directly in the submodule:
```suggestion
      };
      
      config = {
        binary = mkOptionDefault (config.package.meta.mainProgram or config.package.pname);
      };
```

Can then remove allowing `null` values.

modules/lib/maintainers.nix
```diff
@@ -201,6 +201,12 @@
     github = "MForster";
     githubId = 4067975;
   };
+  michaelCTS = {
+    name = "michaelCTS";
+    email = "michael.vogel@cts.co";
+    github = "michaelCTS";
+    githubId = 132582212;
+  };
```
rycee, line 209: No need to add yourself here if you already is in the Nixpkgs `maintainer-list.nix`.
```suggestion
```

rycee:
Thanks for the contribution! I've added a bunch of, hopefully, useful comments. Note, to actually include the module you need to add it to the `modules/modules.nix` file.

exzombie:
I'm really happy that you found the time and will to make a pull request! Many of us have built our own hacks and workarounds, and will appreciate having a solution as part of HM.

I haven't had the time to try this yet, but just looking at the interface, I'm a bit surprised that you chose to implement a list of packages to enable nixgl for. Having followed the various discussions, I got the impression that most of us went with wrapper functions. Is there a particular reason why you chose to go with a list of packages instead?

One reason why a wrapper is more convenient is that you can wrap a package regardless of where it's used, be it in `home.packages` or in `programs.whatever.package`. As noted by @joefiorini in #3968, not being able to use nixgl in `programs.kitty.package` is a real problem.

One more question: does this module require building the home configuration with `--impure`?

modules/misc/nixgl.nix
michaelCTS: I was hoping lazy eval wouldn't evaluate it until a user put stuff into `.packages`, but the example should give enough guidance :+1:

modules/misc/nixgl.nix
michaelCTS: I'm not sure, honestly. There seems to also be `nixGLNvidiaBumblebee` and `nixVulkanNividia` according to the [README](https://github.com/nix-community/nixGL). If we introduced an `enum` that would make users unable to target them.

modules/misc/nixgl.nix
michaelCTS: Yeah, seems like `nixGLNvidia`, `nixGLNvidiaBumblebee`, and others exist.

modules/misc/nixgl.nix
michaelCTS: I'm not able to get this to work. It says `attribute "package" missing`. I tried `binary = mkOptionDefault (config: (config.package.meta.mainProgram or getName config.package));` hoping the function would be evaluated, but it doesn't.

modules/misc/nixgl.nix
michaelCTS: Thanks, renamed

modules/misc/nixgl.nix
michaelCTS: Applied and force-pushed.

modules/misc/nixgl.nix
michaelCTS: Applied and force-pushed.

modules/misc/nixgl.nix
michaelCTS: Renamed and force-pushed

modules/misc/nixgl.nix
michaelCTS: Renamed and force pushed

modules/lib/maintainers.nix
michaelCTS: Removed and force-pushed

michaelCTS:
@rycee I addressed your comments and applied the changes. 

Additionally, I added the wrapper to `config.lib.nixGl.wrap` like @Smona. That should resolves your problem @joefiorini 

```nix
programs.kitty = {                                          
 enable = true;                                            
 package = config.lib.nixGl.wrap { package = pkgs.kitty; };
};                                                          
```

> One more question: does this module require building the home configuration with `--impure`?

@exzombie I have no idea what that does. It works without that for me.

michaelCTS:
Closed in favor of  #5355

PR #5333, opened 2024-04-24
Zellij: add fish completions & autostart options

tennox:
### Description
First, I just wanted to add completions to the fish integration, then I figured I often want completions but not autostart, then I added the autostart options listed in [zellij docs](https://zellij.dev/documentation/integration.html#fish), and finally adapted the tests.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@mainrs

### DIFF
--- a/modules/programs/zellij.nix
+++ b/modules/programs/zellij.nix
@@ -9,7 +9,7 @@ let
   zellijCmd = getExe cfg.package;
 
 in {
-  meta.maintainers = [ hm.maintainers.mainrs ];
+  meta.maintainers = with hm.maintainers; [ mainrs tennox ];
 
   options.programs.zellij = {
     enable = mkEnableOption "zellij";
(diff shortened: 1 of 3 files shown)


modules/programs/zellij.nix
```diff
@@ -49,9 +49,28 @@ in {
...
     };
 
-    enableFishIntegration = mkEnableOption "Fish integration" // {
+    enableFishIntegration = mkEnableOption
+      "Fish integration (enables both enableFishAutoStart and enableFishCompletions)"
+      // {
+        default = false;
+      };
+    enableFishCompletions = mkEnableOption "load zellij completions" // {
       default = false;
```
mainrs, line 58: These completion options should be true by default. This is true for all shells afaik.

modules/programs/zellij.nix
```diff
@@ -49,9 +49,28 @@ in {
...
+      mkEnableOption "autostart zellij in interactive sessions" // {
+        default = false;
+      };
+    autoStartAttachIfSessionExists = mkEnableOption
+      "If the zellij session already exists, attach to the default session. (requires enableFishAutoStart)"
+      // {
+        default = false;
+      };
+    autoStartExitShellOnZellijExit = mkEnableOption
+      "When zellij exits, exit the shell as well. (requires enableFishAutoStart)"
```
mainrs, line 70: mkEnableOption uses a default description text template of the form `Whether to <your string you pass>`. It would be great if you can adapt your text so that the whole sentence makes sense. 

Or you can simply rely on `mkOption` with a bool type instead. And fully customize the description text.

mainrs (changes requested):
@rycee Is the change request correct regarding shell completions? Always enable them, since they are only added if the shell is used anyway.

modules/programs/zellij.nix
tennox: Good call. changed :)

modules/programs/zellij.nix
tennox: Also good call - updated all descriptions :)

tennox:
I've rebased and adapted - @mainrs wanna review again?

khaneliman:
Should have been addressed already in https://github.com/nix-community/home-manager/pull/6695

PR #5338, opened 2024-04-25
firefox: update extensions option description

IldenH:
### Description

The "extensions.autoDisableScopes" option defaults to 15, which disables nearly all extensions, see https://kb.mozillazine.org/About:config_entries#Extensions.

What I think would be preferable would be to have "extensions.autoDisableScopes" = 0; set as a default in settings. I don't know the best way to do that, [and it may cause problems for some](https://bugzilla.mozilla.org/show_bug.cgi?id=666437)?

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/programs/firefox.nix
+++ b/modules/programs/firefox.nix
@@ -683,6 +683,11 @@ in {
 
                 Note that it is necessary to manually enable these extensions
                 inside Firefox after the first installation.
+
+                To automatically enable extensions add
+                `"extensions.autoDisableScopes" = 0;`
+                to
+                [{option}`programs.firefox.profiles.<profile>.settings`](#opt-programs.firefox.profiles._name_.settings)
               '';
             };
 


modules/programs/firefox.nix
```diff
@@ -683,6 +683,11 @@ in {
 
                 Note that it is necessary to manually enable these extensions
                 inside Firefox after the first installation.
+
+                To automatically enable extensions add
+                "extensions.autoDisableScopes" = 0;
+                to
+                programs.firefox.profiles.<profile>.settings
```
rycee, line 690: Can use some markup to improve formatting:
```suggestion
                `"extensions.autoDisableScopes" = 0;`
                to
                [{option}`programs.firefox.profiles.<profile>.settings`](#opt-programs.firefox.profiles._name_.settings)
                
```

rycee:

PR #5340, opened 2024-04-26
git: sendemail: add realName to From field

d4ilyrun:
### Description

We currently have no way of specifying the sender's name inside the `From` field, making a patch sent through `git send-email` appear as coming from "xxx@domain.com".

In this commit we make this field follow the standard `realName <email>` format whenever applicable.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/programs/git.nix
+++ b/modules/programs/git.nix
@@ -392,7 +392,7 @@ in {
           nameValuePair "sendemail.${name}" (if account.msmtp.enable then {
             smtpServer = "${pkgs.msmtp}/bin/msmtp";
             envelopeSender = "auto";
-            from = address;
+            from = "${realName} <${address}>";
           } else
             {
               smtpEncryption = if smtp.tls.enable then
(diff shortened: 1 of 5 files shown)


modules/programs/git.nix
```diff
@@ -407,7 +407,10 @@ in {
                 mkIf smtp.tls.enable (toString smtp.tls.certificatesFile);
               smtpServer = smtp.host;
               smtpUser = userName;
-              from = address;
+              from = if (builtins.stringLength realName == 0) then
```
rycee, line 410: I think it should be safe to simply assume that `realName` is set, it does not have a default value so people would have to manually set `realName = ""` in their configuration for it to be an empty string.

But if you want to keep this if-expression, then you can simplify the condition
```suggestion
              from = if realName == "" then
```

rycee:
Thanks for the contribution, just added a minor comment.

modules/programs/git.nix
d4ilyrun: Oh yes I think misread the module's doc 😆 

Would you be ok with:
```suggestion
              from = if account ? realName then
```

rycee: I think the best would be to simply do without the if-expression:

``` nix
from = "${realName} <${address}>";
```

Since the `realName` option is non-nullable it should be always set by the user.

rycee:

PR #5341, opened 2024-04-27
lib/generators.nix: init toHyprconf generator

khaneliman:
### Description
Creating a library module to be used amongst the various hypr ecosystem modules. In response to https://github.com/nix-community/home-manager/pull/5324#issuecomment-2080203195. I had to refactor the function when moving to a separate module so I created a test to verify that it didn't break anything. 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@fufexan

### DIFF
--- a/modules/lib/generators.nix
+++ b/modules/lib/generators.nix
@@ -1,6 +1,49 @@
 { lib }:
 
 {
+  toHyprconf = { attrs, indentLevel ? 0, importantPrefixes ? [ "$" ], }:
+    let
+      inherit (lib)
+        all concatMapStringsSep concatStrings concatStringsSep filterAttrs foldl
+        generators hasPrefix isAttrs isList mapAttrsToList replicate;
+
+      initialIndent = concatStrings (replicate indentLevel "  ");
+
+      toHyprconf' = indent: attrs:
+        let
+          sections =
+            filterAttrs (n: v: isAttrs v || (isList v && all isAttrs v)) attrs;
+
+          mkSection = n: attrs:
+            if lib.isList attrs then
+              (concatMapStringsSep "\n" (a: mkSection n a) attrs)
+            else ''
+              ${indent}${n} {
+              ${toHyprconf' "  ${indent}" attrs}${indent}}
+            '';
+
+          mkFields = generators.toKeyValue {
+            listsAsDuplicateKeys = true;
+            inherit indent;
+          };
+
+          allFields =
+            filterAttrs (n: v: !(isAttrs v || (isList v && all isAttrs v)))
+            attrs;
+
+          isImportantField = n: _:
+            foldl (acc: prev: if hasPrefix prev n then true else acc) false
+            importantPrefixes;
+
+          importantFields = filterAttrs isImportantField allFields;
+
+          fields = builtins.removeAttrs allFields
+            (mapAttrsToList (n: _: n) importantFields);
+        in mkFields importantFields
+        + concatStringsSep "\n" (mapAttrsToList mkSection sections)
+        + mkFields fields;
+    in toHyprconf' initialIndent attrs;
+
   toKDL = { }:
     let
       inherit (lib) concatStringsSep splitString mapAttrsToList any;
--- a/modules/services/window-managers/hyprland.nix
+++ b/modules/services/window-managers/hyprland.nix
@@ -9,7 +9,6 @@ let
   systemdActivation = ''
     exec-once = ${pkgs.dbus}/bin/dbus-update-activation-environment --systemd ${variables} ${extraCommands}
   '';
-
 in {
   meta.maintainers = [ lib.maintainers.fufexan ];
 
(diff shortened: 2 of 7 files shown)


modules/lib/hyprlang.nix
```diff
@@ -0,0 +1,44 @@
...
+
+      mkFields = generators.toKeyValue {
+        listsAsDuplicateKeys = true;
+        inherit indent;
+      };
+      allFields = filterAttrs (n: v: !(isAttrs v) && n != "device") attrs;
+
+      importantFields = filterAttrs (n: _:
+        (hasPrefix "$" n) || (hasPrefix "bezier" n)
+        || (sourceFirst && (hasPrefix "source" n))) allFields;
```
rycee, line 37: Maybe `importantFields` should be extracted into a function argument?

khaneliman: @fufexan  should these be hardcoded in each module as an argument or would it be another config option from the module to pass to the library?

khaneliman: Decided to make it an option... I could see a benefit to a user being able to organize their config themself through it.

modules/lib/hyprlang.nix
```diff
@@ -0,0 +1,56 @@
+{ lib }: rec {
+  valueType = with lib.types;
+    nullOr
+    (oneOf [ bool int float str path (attrsOf valueType) (listOf valueType) ])
+    // {
+      description = "Hyprlang configuration value";
+    };
+
+  toHyprconf = with lib;
```
fufexan, line 9: Maybe having an argset instead of multiple args would be better, given their number. Also allows us to set defaults.

modules/lib/hyprlang.nix
```diff
@@ -0,0 +1,62 @@
...
```
rycee, line 9: It seems this contains quite a few things that are very specific to Hyprland. Consider, for example, a program other than Hyprland that uses "hyprlang" and a configuration entry "device"…

I would suggest, for example, to instead allow `toHyprconf` to either take an attribute set or a list of attribute sets as input. If the list method is chosen then the output is in that order. The Hyprland module would then be responsible for doing the device and important prefixes stuff and constructing the list accordingly. This function shouldn't have to know anything Hyprland specific.

Also, the function should be placed in `generators.nix` next to `toKDL` 🙂

khaneliman: Hmm.. alright, I'll look into that... If we go this route of moving the generator out of the library module, does the valueType need to go back to being declared in each module.

fufexan: Apparently so. I think a Nixpkgs `pkgs/pkgs-lib/formats.nix`-style file would fare better in this context.

To be honest, I still don't understand the separation in Nixpkgs between `formatters` and `generators`. But that is off-topic.

khaneliman: In implementing this... is there any actual reason for the ordering of specific sections? ie: the device configuration in the hyprland example being at the top or was it simply preference of the name attribute at the top of the attribute set?

khaneliman: Update includes this change where it can function handling the list of attribute sets properly without needing to manually specify them. Also added another test so we can see it properly handling a singular device attribute set and a list of device attribute sets.

fufexan: > In implementing this... is there any actual reason for the ordering of specific sections? ie: the device configuration in the hyprland example being at the top or was it simply preference of the name attribute at the top of the attribute set? Because I can just set up an important prefix in modules for the attributes we want generated at the top of their sections, if that was the reason.

The device section can be anywhere. But the `name` field has to be the first in the section, that's how vaxry implemented the parsing.

rycee:
Thanks, nice improvements! It all looks OK to me but I'll let @fufexan have a change to look. I'll merge tomorrow unless any complaints arise.

khaneliman:
> Thanks, nice improvements! It all looks OK to me but I'll let @fufexan have a change to look. I'll merge tomorrow unless any complaints arise.

Sounds good, I have these others that will follow using this generator.

- https://github.com/nix-community/home-manager/pull/5324
- https://github.com/nix-community/home-manager/pull/5325
- https://github.com/nix-community/home-manager/pull/5344

modules/lib/generators.nix
```diff
@@ -1,6 +1,49 @@
 { lib }:
 
-{
+rec {
+  toHyprconf = with lib;
```
fufexan, line 4: Imo lib functions should be inherited, as it's being done in the `toKDL` function, knowing `with x;` is an anti-pattern in the Nix language.

fufexan (approved):

khaneliman:
> Thanks, good job! Merged to master now 🙂
> 
> Note, I made a slight change in the function definition to avoid the frequent list generation and concatenation.

Thanks, I can rebase the others when I get home.

khaneliman:
@rycee @fufexan the other module PRs should be ready for review now that the generator is set up. Do you want me to keep them as separate PRs or merge them into one to handle the conflicts that will arise from adding multiple news items and the module imports?

rycee:
Keep them separate. I can resolve the merge conflicts as they come.

PR #5344, opened 2024-04-27
Hyprpaper: add module

### LINKED ISSUE #4632: Hyprpaper support

khaneliman:
### Description
Adding the hyprpaper module, similar to https://github.com/nix-community/home-manager/pull/5324

Looked like https://github.com/nix-community/home-manager/pull/5059 was abandoned so I created this along the same lines as the other hypr modules i'm setting up. 

Closes #5059
Closes #4632

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@fufexan

### DIFF
--- a/modules/services/hyprpaper.nix
+++ b/modules/services/hyprpaper.nix
@@ -1,0 +1,89 @@
+{ config, lib, pkgs, ... }:
+with lib;
+let
+
+  cfg = config.services.hyprpaper;
+in {
+  meta.maintainers = [ maintainers.khaneliman maintainers.fufexan ];
+
+  options.services.hyprpaper = {
+    enable = mkEnableOption "Hyprpaper, Hyprland's wallpaper daemon";
+
+    package = mkPackageOption pkgs "hyprpaper" { };
+
+    settings = lib.mkOption {
+      type = with lib.types;
+        let
+          valueType = nullOr (oneOf [
+            bool
+            int
+            float
+            str
+            path
+            (attrsOf valueType)
+            (listOf valueType)
+          ]) // {
+            description = "Hyprpaper configuration value";
+          };
+        in valueType;
+      default = { };
+      description = ''
+        hyprpaper configuration written in Nix. Entries with the same key
+        should be written as lists. Variables' and colors' names should be
+        quoted. See <https://wiki.hyprland.org/Hypr-Ecosystem/hyprpaper/> for more examples.
+      '';
+      example = lib.literalExpression ''
+        {
+          ipc = "on";
+          splash = false;
+          splash_offset = 2.0;
+
+          preload =
+            [ "/share/wallpapers/buttons.png" "/share/wallpapers/cat_pacman.png" ];
+
+          wallpaper = [
+            "DP-3,/share/wallpapers/buttons.png"
+            "DP-1,/share/wallpapers/cat_pacman.png"
+          ];
+        }
+      '';
+    };
+
+    importantPrefixes = lib.mkOption {
+      type = with lib.types; listOf str;
+      default = [ "$" ];
+      example = [ "$" ];
+      description = ''
+        List of prefix of attributes to source at the top of the config.
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable {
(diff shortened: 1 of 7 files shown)


modules/services/hyprpaper.nix
```diff
@@ -0,0 +1,90 @@
...
+        attrs = cfg.settings;
+        inherit (cfg) importantPrefixes;
+      };
+    };
+
+    systemd.user.services.hyprpaper = {
+      Unit = {
+        Description = "hyprpaper";
+        After = [ "graphical-session.target" ];
+        WantedBy = [ "default.target" ];
```
rycee, line 80: Shouldn't this be the more conventional

```suggestion
        After = [ "graphical-session-pre.target" ];
        PartOf = [ "graphical-session.target" ];
```

?

Also, need to add

```
X-Restart-Triggers = [ "${config.xdg.configFile."hypr/hyprpaper.conf".source}" ];
```

to enable automatic restart when the configuration changes.

rycee: Also, you'll probably need

``` nix
Install = { WantedBy = [ "graphical-session.target" ]; };
```

to get it started.


And

``nix
ConditionEnvironment = [ "WAYLAND_DISPLAY" ];
```

to only start for Wayland sessions…

Comment (rycee) [reply modules/services/hyprpaper.nix:80]: Btw, an alternative to `X-Restart-Triggers` would be to provide the path to the configuration file directly in the command. Something like

```nix
ExecStart = let
  configFile = pkgs.writeText "hyprpaper.conf"
    (lib.hm.generators.toHyprconf {
      attrs = cfg.settings;
      inherit (cfg) importantPrefixes;
    });
in "${getExe cfg.package} -c '${configFile}'";
```

perhaps.

khaneliman: I'll implement with the X-restart-triggers since it seems more flexible with supporting people who just want the service and do the config file outside the module, or even wanting to test a different binary or test config file not through the service since the program will read from the expected file location.

khaneliman: Did the same update to hypridle's PR.

khaneliman:
@rycee rebased and resolved merge conflicts

modules/services/hyprpaper.nix
```diff
@@ -0,0 +1,95 @@
...
+            description = "Hyprpaper configuration value";
+          };
+        in valueType;
+      default = { };
+      description = ''
+        hyprpaper configuration written in Nix. Entries with the same key
+        should be written as lists. Variables' and colors' names should be
+        quoted. See <https://wiki.hyprland.org/Hypr-Ecosystem/hyprpaper/> for more examples.
+      '';
+      example = lib.literalExpression ''
```
andresilva, line 35: This example is for hypridle

andresilva (changes requested):

modules/services/hyprpaper.nix
khaneliman: Thanks, I guess I forgo to copy the config over after setting up the unit test.

andresilva (approved):

fufexan (approved):

modules/services/hyprpaper.nix
```diff
@@ -0,0 +1,90 @@
...
+          splash_offset = 2.0;
+
+          preload =
+            [ "/share/wallpapers/buttons.png" "/share/wallpapers/cat_pacman.png" ];
+
+          wallpaper = [
+            "DP-3,/share/wallpapers/buttons.png"
+            "DP-1,/share/wallpapers/cat_pacman.png"
+          ];
+
```
sumnerevans, line 48: nit

```suggestion
```

sumnerevans (approved):

modules/services/hyprpaper.nix
khaneliman: Cleaned up

PR #5346, opened 2024-04-27
cliphist: option to include images in clipboard history

zvolin:
### Description

Adds an option to also store images in cliphist's store. This requires creating another systemd service which watches wl-paste with dedicated change type.

I didn't find in any explicit requirement about this in cliphist, however it did solve the issue for me. Also this is the way presented in hyprland docs and accepted solution in this issue https://github.com/sentriz/cliphist/issues/28

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@Janik-Haag

### DIFF
--- a/modules/services/cliphist.nix
+++ b/modules/services/cliphist.nix
@@ -9,6 +9,14 @@ in {
 
     package = lib.mkPackageOption pkgs "cliphist" { };
 
+    allowImages = lib.mkOption {
+      type = lib.types.bool;
+      default = true;
+      description = ''
+        Store images in clipboard history.
+      '';
+    };
+
     systemdTarget = lib.mkOption {
       type = lib.types.str;
       default = "graphical-session.target";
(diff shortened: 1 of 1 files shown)


rycee:

modules/services/cliphist.nix
```diff
@@ -46,5 +54,21 @@ in {
...
+
+      Service = {
+        Type = "simple";
+        ExecStart =
+          "${pkgs.wl-clipboard}/bin/wl-paste --type image --watch ${cfg.package}/bin/cliphist store";
+        Restart = "on-failure";
+      };
+
+      Install = { WantedBy = [ cfg.systemdTarget ]; };
+    };
```
Janik-Haag, line 72: This should just add the flags conditionally Instead of copying the entire unit.
e.g.:
```nix
"${pkgs.wl-clipboard}/bin/wl-paste}/bin/wl-paste ${lib.optionals cfg.allowImages "--type image"} ...
```

Janik-Haag:
Sorry for my late response, I'm quiet strapped on time the last few days/weeks.
Btw I didn't read up on the linked issue yet but fwiw pictures work for me without that option and are just shown as binary data and their data type and size.

modules/services/cliphist.nix
zvolin: without it, if I copy something from firefox, I just get the html entry in cliphist
```
nixos on  main [!] 
❯ cliphist list
905	<meta http-equiv="content-type" content="text/html; charset=utf-8"><img src="https://avatars.githubu
```

PR #5352, opened 2024-04-29
mpv: add extraInput option

### LINKED ISSUE #5212: programs.mpv.bindings: Multiple items with the same key
Not sure if this should be classified as a feature request, documentation request, or simply brain dump for the future intrepid explorers stumbling through the same problem.

Use case: uosc allows defining [custom menus](https://github.com/tomasklaen/uosc?tab=readme-ov-file#adding-items-to-menu), for example:

```
#    #! ---
#    #! Section > ---
```

The problem is that nix cannot have a dictionary with the same value. I figured I could just go funky and define it like this:

```nix
"#    #! ---" = "";
"#    #! Section > ---" = "";
```

It appears to work, but it's undocumented. I think this either should be added as an example [here](https://mynixos.com/home-manager/option/programs.mpv.bindings), or binding values should be allowed to be an array:

```nix
"#": [
  "#! ---"
  "#! Section > ---"
]
```

Another option is to address the source directly and add the ability to define the menu in `programs.mpv.scriptOpts.uosc`.

If no solution is feasible, I hope my workaround would help someone in the future.

Lichthagel:
### Description

Adds an option `programs.mpv.extraInput` for appending additional lines to `$XDG_CONFIG_DIR/mpv/input.conf`.

One use case for this is adding menu entries in uosc (an alternative osc). Especially when not assigned to a key. See https://github.com/tomasklaen/uosc?tab=readme-ov-file#menu-1.

Closes #5212.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@tadeokondrak @thiagokokada @chuangzhu

### DIFF
--- a/modules/programs/mpv.nix
+++ b/modules/programs/mpv.nix
@@ -176,6 +176,20 @@ in {
           }
         '';
       };
+
+      extraInput = mkOption {
+        description = ''
+          Additional lines that are appended to {file}`$XDG_CONFIG_HOME/mpv/input.conf`.
+           See {manpage}`mpv(1)` for the full list of options.
+        '';
+        type = with types; lines;
+        default = "";
+        example = ''
+          esc         quit                        #! Quit
+          #           script-binding uosc/video   #! Video tracks
+          # additional comments
+        '';
+      };
     };
   };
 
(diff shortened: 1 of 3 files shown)


thiagokokada (approved):

rycee:

PR #5362, opened 2024-05-01
cliphist: add extraOptions option

ErrorNoInternet:
Add an extraOptions option that would be appended to the cliphist command.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@Janik-Haag

### DIFF
--- a/modules/services/cliphist.nix
+++ b/modules/services/cliphist.nix
@@ -17,6 +17,14 @@ in {
       '';
     };
 
+    extraOptions = lib.mkOption {
+      type = with lib.types; listOf str;
+      default = [ "-max-dedupe-search" "10" "-max-items" "500" ];
+      description = ''
+        Flags to append to the cliphist command.
+      '';
+    };
+
     systemdTarget = lib.mkOption {
       type = lib.types.str;
       default = "graphical-session.target";
@@ -31,7 +39,8 @@ in {
     };
   };
 
-  config = lib.mkIf cfg.enable {
+  config = let extraOptionsStr = lib.escapeShellArgs cfg.extraOptions;
+  in lib.mkIf cfg.enable {
     assertions = [
       (lib.hm.assertions.assertPlatform "services.cliphist" pkgs
         lib.platforms.linux)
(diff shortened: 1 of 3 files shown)


modules/services/cliphist.nix
```diff
@@ -17,6 +17,22 @@ in {
...
+      '';
+    };
+
+    maxItems = lib.mkOption {
+      type = lib.types.int;
+      default = 750;
+      description = ''
+        Maximum number of items to store.
+      '';
+    };
```
Janik-Haag, line 34: I think instead of adding options for every comment line flag we should do something like [`programs.atuin.flags`](https://github.com/nix-community/home-manager/blob/master/modules/programs/atuin.nix#L54)

modules/services/cliphist.nix
```diff
@@ -17,6 +17,14 @@ in {
       '';
     };
 
+    flags = lib.mkOption {
```
rycee, line 20: This type of option is typically called `extraOptions`. If you search, you'll find multiple examples so I would suggest to follow that convention:
```suggestion
    extraOptions = lib.mkOption {
```

rycee:
Thanks for the contribution! Just added a minor comment.

modules/services/cliphist.nix
Janik-Haag: Huh, I didn't know that yet, and remembered adding the flag option to atuin some time ago. Thanks for sharing.

modules/services/cliphist.nix
```diff
@@ -17,6 +17,14 @@ in {
       '';
     };
 
+    extraOptions = lib.mkOption {
+      type = with lib.types; listOf str;
+      default = [ ];
+      description = ''
```
Janik-Haag, line 23: can you add your usecase as example?

modules/services/cliphist.nix
rycee: No worries, it's not a hard rule or anything.

modules/services/cliphist.nix
rycee: I added the example from the tests.

rycee:

PR #5367, opened 2024-05-03
mopidy: make scan service depend on `mopidy-local`

nekowinston:
### Description

The systemd service `mopidy-scan` provided by this module depends on `mopidy local` which is only available via an optional extension package. When the extension isn't installed, the service fails to start.
This PR checks if `mopidy-local` is part of `cfg.extensionPackages` before enabling the service.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@foo-dogsquared


rycee:
Thanks! This seems sensible to me so I merged it. Hopefully @foo-dogsquared doesn't mind 🙂

PR #5378, opened 2024-05-05
autorandr: configModule.extraConfig

zackattackz:
### Description

Add an option to programs.autorandr's configModule to allow arbitrary extra config lines.

No option exists for adding arbitrary key/values to generated autorandr profile config, as is common in other nix modules. This commit adds one.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@uvNikita


zackattackz:
@uvNikita Hello no worries on this but just wanted to ping you if you have the time for a review. I need this option for some device specific config that is not included in the regular config options. I'm sure having extraConfig can be useful to others who need to add something to their config quick and dirty.

Thank you!

uvNikita:
Makes sense to have possibility to add extra options.

If the options are always in the format of `key` `value`, maybe it makes sense to add `settings` option instead, in line with [rfcs42](https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md)?

uvNikita:
You can see implementation example in other modules, e.g. https://github.com/nix-community/home-manager/blob/d7830d05421d0ced83a0f007900898bdcaf2a2ca/modules/programs/i3status.nix#L30

zackattackz:
@uvNikita Good point. I am not sure of the config format exactly for autorandr, but I can investigate! If it is strictly keys/values I can try to make settings. But if it turns out more complicated, or if it also allows flag type options where it's just a key and no value, then would extraConfig be fine then?

uvNikita:
@zackattackz I think so. Thank you for looking into this and sorry for a late review :)

modules/programs/autorandr.nix
```diff
@@ -177,6 +177,18 @@ let
         default = null;
         example = "nearest";
       };
+
+      extraConfig = mkOption {
+        type = types.nullOr types.lines;
```
uvNikita, line 182: This can be just `types.lines` with default value empty string `""`.
Then can just append it to the generated config file unconditionally.

zackattackz:
Hi @uvNikita 

Sorry for the lack of response here. I appreciate your review.

I looked into the config file format and it just seems to be a bespoke parser: https://github.com/phillipberndt/autorandr/blob/770dc86bec9d20bbe22b596aad9aa1520c966720/autorandr.py#L505

It can be either keys/values pairs or also just independent keys ("off" for example).

I think this can be represented as a settings type of `attrsOf str (either str bool)`

so for example:
```nix
config.autorandr.profiles.${profile}.config.output1.settings = {
  "someBoolConfig" = true;
};
config.autorandr.profiles.${profile}.config.output2.settings = {
  "somekey" = "somevalue";
};
```

Which would generate config like:
```
output output1
someBoolConfig

output output2
somekey someval
```

I think this wouldnt be too hard to implement?

But honestly, because I couldn't find a solid definition of the config format, and because the autorandr code is kind of a lot to take in, I could be missing some considerations here.

I feel like extraConfig would be easier to keep and maintain in the long run, though I do understand settings is preferred generally.

If you'd like, I can implement settings as described above (or some other way you see fit).

Or I can amend with the suggestion you made for extraConfig to merge that instead?

uvNikita:
Since it's a custom format, I don't mind if we keep your original suggestion of adding `extraConfig` option. Another one can also be added later if need arises.

zackattackz:
OK! Just lmk if anything else is needed, no rush of course.

modules/programs/autorandr.nix
uvNikita, line 182:
```suggestion
        type = types.lines;
```

modules/programs/autorandr.nix
```diff
@@ -177,6 +177,18 @@ let
...
+        description = "Extra lines to append to this profile's config.";
+        default = null;
```
uvNikita, line 184:
```suggestion
        default = "";
```

modules/programs/autorandr.nix
```diff
@@ -267,7 +279,8 @@ let
           + concatMapStringsSep "," toString (flatten config.transform))
         ++ optional (config.scale != null)
         ((if config.scale.method == "factor" then "scale" else "scale-from")
-          + " ${toString config.scale.x}x${toString config.scale.y}"))
+          + " ${toString config.scale.x}x${toString config.scale.y}")
+        ++ optional (config.extraConfig != null) config.extraConfig)
```
uvNikita, line 283:
```suggestion
        + config.extraConfig)
```

uvNikita (changes requested):

zackattackz:
Sorry forgot about those suggestions. Added them in, just needed a slight difference from your last suggestion.

modules/programs/autorandr.nix
zackattackz: Well I guess my new amend could maybe cause an issue with an extra \n at the end?
I thought I ran the tests right but wasn't sure. I'll double check

zackattackz: OK so not using `optional` here would make the autorandr scale test fail bc it would be introducting a newline for empty `extraConfig`. So I think using `optional` with the condition being `config.extraConfig != ""` is most correct.

zackattackz:
Amended with the requested changes 🫡

uvNikita (approved):

uvNikita:
@rycee Do you mind merging this one? I can't merge until test succeed, but doesn't look like this is related?

zackattackz:
Thanks for the review :)

Issue #5383, opened 2024-05-07
bug: `programs.gpg.settings` should never have default algo SHA512


Cryolitia:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

when enabling `programs.gpg.enable`, this module will have a set of **even hide** settings:
https://github.com/nix-community/home-manager/blob/6e277d9566de9976f47228dd8c580b97488734d4/modules/programs/gpg.nix#L243-L263

Even these settings is not documented https://github.com/nix-community/home-manager/blob/6e277d9566de9976f47228dd8c580b97488734d4/modules/programs/gpg.nix#L153-L171

The algo SHA512 in the hide settings, is hardly supported by almost all hardware TPM, including my AMD Ryzen 7840U. This will immediately cause gpg sign fault and hard to find out.

We should never provide default settings, but leave it to gpg itself.

Related: https://lists.gnupg.org/pipermail/gnupg-devel/2021-December/035018.html

### Maintainer CC

No maintainer, CC blame: @offlinehacker

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.8.6-zen1, NixOS, 24.05 (Uakari), 24.05.20240505.25865a4`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.18.2`
 - channels(root): `"nixos, nixpkgs"`
 - nixpkgs: `/home/cryolitia/.nix-defexpr/channels/nixpkgs`
```

rycee:
Thanks for opening the issue. I'm inclined to agree. I assume that the GnuPG upstream defaults are well chosen nowadays so well worth to revisit.

Cryolitia:
I have opened a PR for it and more ^_^ https://github.com/nix-community/home-manager/issues/5383

Cryolitia:
It's still a issue, and PR is here: https://github.com/nix-community/home-manager/pull/5386

Cryolitia:
still

Cryolitia:
still

Cryolitia:
still

Cryolitia:
still

Cryolitia:
still

PR #5386, opened 2024-05-08
gpg: delete default settings, add dirmngrSettings, add maintainer

Cryolitia:
### Description

fix https://github.com/nix-community/home-manager/issues/5383

I add myself as gpg module's maintainer, due to no maintainer now.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/gpg.nix
```diff
@@ -8,15 +8,11 @@ let
...
-
-  scdaemonCfgText = generators.toKeyValue {
-    inherit mkKeyValue;
-    listsAsDuplicateKeys = true;
-  } cfg.scdaemonSettings;
+  generateCfgText = attr:
+    generators.toKeyValue {
+      inherit mkKeyValue;
+      listsAsDuplicateKeys = true;
+    } attr;
```
ambroisie, line 15:
```suggestion
  generateCfgText = generators.toKeyValue {
    inherit mkKeyValue;
    listsAsDuplicateKeys = true;
  };
```

Cryolitia:
It's still active

modules/programs/gpg.nix
```diff
@@ -240,44 +258,21 @@ in {
...
-      no-comments = mkDefault true;
-      no-emit-version = mkDefault true;
-      keyid-format = mkDefault "0xlong";
-      list-options = mkDefault "show-uid-validity";
-      verify-options = mkDefault "show-uid-validity";
-      with-fingerprint = mkDefault true;
-      require-cross-certification = mkDefault true;
-      no-symkey-cache = mkDefault true;
-      use-agent = mkDefault true;
-    };
```
sumnerevans, line 263: This change doesn't look backwards compatible... Are these the defaults that GPG uses anyway?

sumnerevans (changes requested):
We should probably also add tests for this module.

modules/programs/gpg.nix
Cryolitia: It's absulutely not the defaults that GPG uses, it will cause https://github.com/nix-community/home-manager/issues/5383 , and GPG's defaults not. So it should be deleted whatever the backwards compatible.

Cryolitia:
I think tests could be another PR's target

modules/programs/gpg.nix
sumnerevans: ok, I unchecked the backwards compatible checkbox in the PR description. @rycee do we need to gate this behind a `stateVersion`?

Cryolitia:
it's still active

PR #5387, opened 2024-05-09
silicon: init module

afresquet:
### Description

Add new module to manage [silicon](https://github.com/Aloxaf/silicon) configuration.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/silicon.nix
```diff
@@ -0,0 +1,131 @@
...
+          if isNull val.file then "${val.src}" else "${val.src}/${val.file}";
+      };
+    })));
+
+    # NOTE: we are ensuring `themes` and `syntaxes` directories exist
+    # because silicon assumes they do when running `--build-cache`
+    # https://github.com/Aloxaf/silicon/issues/242
+    home.activation.siliconCache = hm.dag.entryAfter [ "linkGeneration" ] ''
+      (
+        export XDG_CACHE_HOME=${escapeShellArg config.xdg.cacheHome}
```
uncenter, line 121: Why is this needed?

modules/programs/silicon.nix
```diff
@@ -0,0 +1,131 @@
...
+      example = literalExpression ''
+        --shadow-color '#555'
+        --background '#fff'
+        --shadow-blur-radius 30
+        --no-window-controls
+      '';
+      description = ''
+        Silicon configuration.
+      '';
+    };
```
uncenter, line 27: When I worked on my own module for Silicon, I went with a more true-to-Nix approach and I'd love to hear your thoughts. Settings would be set with Nix values (https://github.com/uncenter/flake/blob/9908d5d4f3e345459dfc638f6434d9d28ec5446b/modules/extra/home-manager/silicon.nix#L11-L23) and this attrset would then be converted into the configuration file with https://github.com/uncenter/flake/blob/9908d5d4f3e345459dfc638f6434d9d28ec5446b/modules/extra/home-manager/silicon.nix#L43-L57. You can see it in action at https://github.com/uncenter/flake/blob/9908d5d4f3e345459dfc638f6434d9d28ec5446b/home/programs/silicon.nix#L6-L10.

uncenter (commented):
It also doesn't look like I can reuse https://github.com/sharkdp/bat themes and syntaxes. A configuration option for the directory in which to build the cache would be nice!

PR #5397, opened 2024-05-12
bash: Add missing 'ignoreboth' to historyControl allowed values

RoadRoller01:
reference: https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html#index-HISTCONTROL

### Checklist

- [ofc(:] Code tested through `nix-shell --pure tests -A run.all` or `nix develop --ignore-environment .#all` using Flakes.

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/bash.nix
+++ b/modules/programs/bash.nix
@@ -67,8 +67,8 @@ in {
       };
 
       historyControl = mkOption {
-        type =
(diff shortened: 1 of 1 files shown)


teto (approved):

rycee:

PR #5406, opened 2024-05-15
submodule-support: add default values for top-level configs

name-snrl:
This way the end user can easily check whether the home-manager configuration is part of the module or not. Example of use:

```nix
{ lib, nixosConfig, ... }:
let
  mkIfNixos = lib.mkIf nixosConfig != null;
in
{
  programs.foot.enable = mkIfNixos true;
}
```

related: https://github.com/NixOS/nixpkgs/issues/311709
many thanks to @roberth for the explanation

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/submodule-support.nix
+++ b/modules/misc/submodule-support.nix
@@ -29,4 +29,19 @@ with lib;
       '';
     };
   };
+
+  config = {
+    # To make it easier for the end user to override the values in the
+    # configuration depending on the installation method, we set default values
+    # for the arguments that are defined in the NixOS/nix-darwin modules.
+    #
+    # Without these defaults, these attributes would simply not exist, and the
+    # module system can not inform modules about their non-existence; see
+    # https://github.com/NixOS/nixpkgs/issues/311709#issuecomment-2110861842
+    _module.args = {
+      osConfig = mkDefault null;
+      nixosConfig = mkDefault null;
+      darwinConfig = mkDefault null;
+    };
+  };
 }


name-snrl:
@rycee

name-snrl:
@emilazy @rycee @SuperSandro2000 @roberth I apologize for mentioning again those who have already been mentioned before, but perhaps during the weekend I have a better chance to draw your attention to this simple but useful PR

modules/misc/submodule-support.nix
```diff
@@ -29,4 +29,15 @@ with lib;
       '';
     };
   };
+
+  config = {
+    # To make it easier for the end user to override the values in the
+    # configuration depending on the installation method, we set default values
+    # for the arguments that are defined in the NixOS/nix-darwin modules.
```
roberth, line 36:
```suggestion
    # for the arguments that are defined in the NixOS/nix-darwin modules.
    #
    # Without these defaults, these attributes would simply not exist, and the
    # module system can not inform modules about their non-existence; see
    # https://github.com/NixOS/nixpkgs/issues/311709#issuecomment-2110861842
```

roberth (approved):

name-snrl:
@rycee could you merge this please?

rycee:
Thanks. Seems fair enough, so I'll merge once the CI job finishes. But wouldn't something like

``` nix
{ config, lib, ... }:

{
  config.lib.mkIfNixos = lib.mkIf (lib.hasAttr "nixosConfig" config._module.args.nixosConfig);
}
```

work similarly? I haven't tested it, though, so it may fail spectacularly :-)

name-snrl:
> But wouldn't something like

How did I not realize this before.....

> OK they are passed as special args so that would make the above not work very well

Yes, but `_module` also contains `specialArgs`. I tried to use `lib.hasAttr "nixosConfig/osConfig" config._module.specialArgs` and it works.

___

We can close this without merging, but I still think this change will make the use of `osConfig`, `nixosConfig` and `darwinConfig` more obvious

PR #5409, opened 2024-05-15
radicle: init

lorenzleutgeb:
### Description

[Radicle](https://radicle.xyz) is a distributed code forge based on Git. I recently [overhauled the package in Nixpkgs](https://github.com/NixOS/nixpkgs/pull/309050) and have been using [`programs.radicle`](https://github.com/lorenzleutgeb/nur/blob/main/hm/module/programs/radicle.nix) (for basic configuration of Radicle) and [`services.radicle`](https://github.com/lorenzleutgeb/nur/blob/main/hm/module/services/radicle.nix) (for managing `radicle-node` and `radicle-http` as systemd user services) for a few weeks without problems (links to my personal repo).

I could need some help with the tests. `home-files/.radicle` seems to be missing and I don't understand why:
```
$ ls -la /nix/store/afg8b60b5a79ziw03yfiwdcmmnis58l3-nmt-report-radicle-basic-configuration/tested/home-files
dr-xr-xr-x - root 1970-01-01 02:00 .cache
dr-xr-xr-x - root 1970-01-01 02:00 .config
dr-xr-xr-x - root 1970-01-01 02:00 asserts
```

### Checklist

- [x] Change is backwards compatible. (Trivially, since there is no earlier version.)
- [x] Commit messages are formatted.


ju1m:
Thanks a lot for the nixpkgs PR and this draft!
Here's the changes I had to do to make progress https://github.com/ju1m/home-manager/commits/radicle/
Alas, with `nixos-23.11`'s `rustc` building `radicle-node` after cherry-picking the nixpkgs PR fails:
```
radicle-node> Found a `cargo::key=value` build directive which is reserved for future use.                                                                                                                                                                     
radicle-node> Either change the directive to `cargo:key=value` syntax (note the single `:`) or upgrade your version of Rust.                                                                                                                                   
radicle-node> See https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script for more information about build script outputs.
```
so I'll have to update to the upcoming `nixos-24.05` before trying this further more.

lorenzleutgeb:
Hey! Thanks for your help.

If you like, you can also try the Flake at https://github.com/lorenzleutgeb/heartwood in the meantime.

I also have things coming at https://github.com/lorenzleutgeb/radicle-nix, including more changes to the modules in this PR...

What I don't understand is why you reverted to http://github.com/lorenzleutgeb/nur ? The formatting in these files does not conform to home-manager's requirements.

ju1m:
> If you like, you can also try the Flake at https://github.com/lorenzleutgeb/heartwood in the meantime.

Thanks @lorenzleutgeb

> I also have things coming at https://github.com/lorenzleutgeb/radicle-nix, including more changes to the modules in this PR...

Nice to also have a NixOS service, do you plan to open a PR to upstream it?
Just in passing, something that should be checked is whether `LoadCredential=` can load a path from `LoadCredentialEncrypted=`, otherwise we may have to add some logic to support `LoadCredentialEncrypted=`.

> What I don't understand is why you reverted to [http://github.com/lorenzleutgeb/nur](https://github.com/lorenzleutgeb/nur) ? The formatting in these files does not conform to home-manager's requirements.

Well it's only after seeing that there were differences between this PR and the NUR version, and that the NUR contained two fixes (`services.radicle.httpd.environnement.default`'s  `cfg.node.enable` => `{}` and {`RestartDelayMaxSec` => `RestartMaxDelaySec`}) while looking more advanced on socket activation, that I rebased my changes on the NUR version. The formatting can always be fixed later on.

Nice to see you're actively working on this, happy to review your changes on this PR whenever you're ready to push them.

matthiasbeyer:
@lorenzleutgeb please have a look at https://github.com/matthiasbeyer/home-manager/tree/radicle (`git fetch https://github.com/matthiasbeyer/home-manager radicle && git checkout FETCH_HEAD`) which is a rebase + fixup commits of this PR.

All patches that **I** think are uncontroversial are committed as `--fixup` commits. The others have proper commit messages.

I did not touch your initial commit, modulo the rebase changes that were necessary.

I committed my changes in a way where I think it is best if you review them commit-by-commit.

lorenzleutgeb:
@matthiasbeyer thanks a lot! I squashed your changes in. Next I will see what CI has to say, and then integrate changes I made to my personal configuration in the meantime.

khaneliman:
Looks like it needs `nix fmt` and test error: 
> nmt-test-radicle-basic-configuration> 
error: builder for '/nix/store/n8xvqw23ysn3r8n4ilaq9y8vrrsq3sl7-nmt-test-radicle-basic-configuration.drv' failed with exit code 1;
       last 3 log lines:
       > diff: /nix/store/jcvkswcb12lvgsrrdnb97rqjwnfjhd18-home-manager-generation/home-files/.radicle/config.json: No such file or directory
       > Expected home-files/.radicle/config.json to be same as /nix/store/39gc85n30s5qsij1dn620qabzpwl29jb-basic-configuration.json but were different:

> building '/nix/store/kk86pr5r64snk8gf7lgsb1n713rrc9i9-home-manager-files.drv'...
options.json> error: option programs.radicle.environment has no description
options.json> error: option programs.radicle.httpd.args has no description
options.json> error: option programs.radicle.node.args has no description
options.json> error: option programs.radicle.settings has no description
options.json> error: option programs.radicle.settings.cli has no description
options.json> error: option programs.radicle.settings.cli.hints has no description
options.json> error: option programs.radicle.settings.node has no description
options.json> error: option programs.radicle.settings.node.alias has no description
options.json> error: option programs.radicle.settings.node.network has no description
options.json> error: option programs.radicle.settings.node.peers has no description
options.json> error: option programs.radicle.settings.node.peers.target has no description
options.json> error: option programs.radicle.settings.node.peers.type has no description
options.json> error: option programs.radicle.settings.node.policy has no description
options.json> error: option programs.radicle.settings.node.relay has no description
options.json> error: option programs.radicle.settings.node.scope has no description
options.json> error: option programs.radicle.settings.node.workers has no description
options.json> error: option programs.radicle.settings.preferredSeeds has no description
options.json> error: option programs.radicle.settings.publicExplorer has no description
options.json> error: option programs.radicle.settings.web has no description
options.json> error: option programs.radicle.settings.web.pinned has no description
options.json> error: option programs.radicle.settings.web.pinned.repositories has no description
options.json> error: option programs.radicle.uri.rad.browser.preferredNode has no description
options.json> error: option programs.radicle.uri.rad.vscode.extension has no description
options.json> error: option services.radicle.httpd.args has no description
options.json> error: option services.radicle.httpd.environment has no description
options.json> error: option services.radicle.node.args has no description
options.json> error: option services.radicle.node.environment has no description

matthiasbeyer:
It also seems that nixpkgs needs to be refreshed:

> error: radicle-cli cannot be found in pkgs

I don't know how this mechanic works in home-manager, so I cannot comment on that. But without `radicle-cli`, this PR is not possible :laughing:

lorenzleutgeb:
@khaneliman, @matthiasbeyer thanks for your attention, but please don't waste your time reviewing this just yet. I know how to make my way through CI failures, and I will change the status of the PR from "Draft" to "Ready for review" when ... well, when it is *ready for review*.

In the meantime, feel free to push actual fixes like @matthiasbeyer did in https://github.com/nix-community/home-manager/pull/5409#issuecomment-3196993658, but posting errors that CI can catch is just noise to me, TBH.

Thanks for your understanding.

lorenzleutgeb:
Opening this up for review. The simple test added passes. Happy to have your comments @matthiasbeyer, @khaneliman.

modules/programs/radicle.nix
```diff
@@ -0,0 +1,272 @@
...
+                (hasSuffix publicExplorerSuffix cfg.settings.publicExplorer) && pkgs.stdenv.hostPlatform.isLinux;
+              defaultText = "`true` if a suitable public explorer is detected.";
+              example = false;
+            };
+            preferredNode = mkOption {
+              type = str;
+              description = "The hostname of an instance of `radicle-node`, reachable via HTTPS.";
+              default = "iris.radicle.xyz";
+              example = "radicle-node.example.com";
+            };
```
matthiasbeyer, line 74: Can this be a list?

modules/programs/radicle.nix
```diff
@@ -0,0 +1,272 @@
...
+        message = "${opt.uri.rad.browser.enable} is only compatible with a public explorer URL ending in '${publicExplorerSuffix}' but '${cfg.settings.publicExplorer}' does not end with '${publicExplorerSuffix}'.";
+      }
+    ];
+
+    home = {
+      packages = [ cfg.cli.package ];
+      file.".radicle/${configFile.name}".source = configFile.path;
+    };
+
+    programs.git.enable = mkDefault true;
```
matthiasbeyer: I know that having radicle without git is useless, but I am not sure whether it is okay to have this, the home-manager team must tell us here. @nix-community/home-manager-maintainers

modules/programs/radicle.nix
```diff
@@ -0,0 +1,272 @@
...
+
+    services.ssh-agent.enable = mkDefault pkgs.stdenv.hostPlatform.isLinux;
```
matthiasbeyer: Same as above @nix-community/home-manager-maintainers

matthiasbeyer (changes requested):
In general this seems good to me, modulo one possible improvement and a question to the home-manager maintainers how this should be handled.

modules/programs/radicle.nix
lorenzleutgeb: Does not really make sense to me. We want to generate *one* HTTPS URL to open. If we allow multiple nodes, then we would have to think about how select the best one. This of course would be nice, and people could implement their own selection mechanism, but I think it's overkill for the home-manager module.

modules/programs/radicle.nix
lorenzleutgeb: Agreed, happy to remove this.

khaneliman: We have been preferring warnings/assertions > automatically enabling other modules.

khaneliman: Also, that ping was to a large group (~250 people) instead of the smaller @nix-community/home-manager-committers if that is what you were trying to ping.

matthiasbeyer: @khaneliman and all people I pinged: Sorry! I now know what the appropriate group is, won't happen again!

@lorenzleutgeb lets change the code to what @khaneliman suggested: Lets emit a warning rather than enabling stuff for users, okay?

modules/services/radicle.nix
```diff
@@ -0,0 +1,239 @@
...
+            ];
+            After = [
+              "radicle-node-proxy.socket"
+              "radicle-node.service"
+            ];
+            Documentation = [ "man:systemd-socket-proxyd(8)" ];
+          };
+          Service = {
+            ExecSearchPath = "${pkgs.systemd}/lib/systemd";
+            ExecStart = "systemd-socket-proxyd --exit-idle-time=30min %t/radicle-node/proxy.sock";
```
ju1m: I guess this should use `cfg.node.lazy.exitIdleTime` instead of the hardcoded `30min` :-)

ju1m: @lorenzleutgeb IIUC, you [manage systemd socket activation upstream](https://app.radicle.xyz/nodes/seed.radicle.garden/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/commits/2776e003dc2f7e585c205cfa966ffb90f2c43ff5), and the [use of `systemd-socket-proxyd` is only to be able to use `--exit-idle-time=30min`](https://unix.stackexchange.com/a/701241/392771).
If I'm correct, maybe a comment could be added to explain it because it was not obvious to me why `systemd-socket-proxyd` is used when I first read the code.

Thinking about this, maybe that exit-idle-time feature would benefit more users and be better implemented (eg. more graciously or cleanly) upstream?
Aside from that, **thank you very much** for this PR which works fine for me both in non-lazy and lazy mode.
Though I had to unencrypt my radicle key
(I've not tried to use `systemd-creds --user` + `LoadCredentialsEncrypted=` yet).

lorenzleutgeb: Thank you very much for taking a look, @ju1m! Indeed I think we could upstream the idle timer. But maybe we don't wait for that to merge this PR...

However, @fintanh and I am running into issues when announcing (`rad sync -a`) with this PR. I am looking into this currently. Did you manage to announce just fine?

You might have been able to work.around decrypting by populating `$RAD_PASSPHRASE`. Anyway, we are working on improving this upstream as well.

modules/programs/radicle.nix
lorenzleutgeb: Removed in 4a0213a852c4c8100b706cae04ea54dbe1e8f956.

modules/programs/radicle.nix
lorenzleutgeb: Removed in 4a0213a852c4c8100b706cae04ea54dbe1e8f956.

modules/services/radicle.nix
lorenzleutgeb: > I guess this should use `cfg.node.lazy.exitIdleTime` instead of the hardcoded `30min` :-)

Fixed in 4a0213a852c4c8100b706cae04ea54dbe1e8f956.

lorenzleutgeb:
I believe I have resolved most issues pointed out by the reviewers, and the hardening was relaxed to not deny `@timer`, which according to [`man:systemd.exec(5)`](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html) is about system calls for scheduling operations by time ([`man:alarm(2)`](https://man7.org/linux/man-pages/man2/alarm.2.html), [`man:timer_create(2)`](https://man7.org/linux/man-pages/man2/timer_create.2.html), …), and I think that is acceptable.

modules/services/radicle.nix
ju1m: FYI, I've tried to use `LoadCredentialEncrypted=`, but AFAIU [systemd v257.7 is not yet able to make it work, at least on some systems like mine](https://github.com/systemd/systemd/issues/37598#issuecomment-2907667208):
```console
$ ./testLoadCredentialEncrypted.sh 
+ unit=testLoadCredentialEncrypted
+ systemctl --version
systemd 257 (257.7)
+PAM +AUDIT -SELINUX +APPARMOR +IMA +IPE +SMACK +SECCOMP +GCRYPT -GNUTLS +OPENSSL +ACL +BLKID +CURL +ELFUTILS +FIDO2 +IDN2 -IDN +IPTC +KMOD +LIBCRYPTSETUP +LIBCRYPTSETUP_PLUGINS +LIBFDISK +PCRE2 +PWQUALITY +P11KIT +QRENCODE +TPM2 +BZIP2 +LZ4 +XZ +ZLIB +ZSTD +BPF_FRAMEWORK -BTF -XKBCOMMON +UTMP -SYSVINIT +LIBARCHIVE
+ echo CLEARTEXT
+ systemd-creds encrypt --user --with-key=host --name test test.txt test.cred
+ systemd-creds decrypt --user --name test test.cred
CLEARTEXT
+ systemctl --user reset-failed testLoadCredentialEncrypted
+ systemd-run --user --unit testLoadCredentialEncrypted --pipe -p LoadCredentialEncrypted=test:/home/julm/test-systemd-creds-user/test.cred systemd-creds cat test
Running as unit: testLoadCredentialEncrypted.service; invocation ID: 6e8a62bcd7d44c91a0f89f4b2ed747f3
ERROR:tcti:src/tss2-tcti/tcti-device.c:455:Tss2_Tcti_Device_Init() Failed to open specified TCTI device file /dev/tpmrm0: Permission denied 
+ journalctl --user --unit testLoadCredentialEncrypted --since -5s
sept. 11 01:16:24 pumpkin systemd[2529]: Starting [systemd-run] /etc/profiles/per-user/julm/bin/systemd-creds cat test...
sept. 11 01:16:24 pumpkin (md-creds)[4098619]: testLoadCredentialEncrypted.service: Failed to set up credentials: State not recoverable
sept. 11 01:16:24 pumpkin (md-creds)[4098619]: testLoadCredentialEncrypted.service: Failed at step CREDENTIALS spawning /etc/profiles/per-user/julm/bin/systemd-creds: State not recoverable
sept. 11 01:16:24 pumpkin systemd[2529]: testLoadCredentialEncrypted.service: Main process exited, code=exited, status=243/CREDENTIALS
sept. 11 01:16:24 pumpkin systemd[2529]: testLoadCredentialEncrypted.service: Failed with result 'exit-code'.
sept. 11 01:16:24 pumpkin systemd[2529]: Failed to start [systemd-run] /etc/profiles/per-user/julm/bin/systemd-creds cat test.
+ ls -l /dev/tpmrm0
crw------- 1 root root 243, 65536 30 août  19:31 /dev/tpmrm0
```

So instead I used `LoadCrendential=` to test, and that's what I had to do:

```nix
systemd.user.services.radicle-node = {
  Service = {
    LoadCredential = ["radicle:${pumpkin/radicle/radicle.clear}"];
    BindReadOnlyPaths = [
      "/run/user/1000/credentials/radicle-node.service/radicle:/home/julm/.radicle/keys/radicle"
      "${pumpkin/radicle/radicle.pub}:/home/julm/.radicle/keys/radicle.pub"
    ];
  };
  Unit = {
    # Remove:
    #   ConditionPathExists=/home/julm/.radicle/keys/radicle
    #   ConditionPathExists=/home/julm/.radicle/keys/radicle.pub
    # because they now come from BindReadOnlyPaths=
    ConditionPathExists = lib.mkForce [];
  };
};
```

ju1m:
> I believe I have resolved most issues pointed out by the reviewers, and the hardening was relaxed to not deny `@timer`, which according to [`man:systemd.exec(5)`](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html) is about system calls for scheduling operations by time ([`man:alarm(2)`](https://man7.org/linux/man-pages/man2/alarm.2.html), [`man:timer_create(2)`](https://man7.org/linux/man-pages/man2/timer_create.2.html), …), and I think that is acceptable.

Nice catch, we allowed `@timer` in NixOS too, because it is required by `radicle-node` during `git upload-pack`:

https://github.com/NixOS/nixpkgs/blob/7c05be024d9aaf7647204b8a456c74b03bd7d808/nixos/modules/services/misc/radicle.nix#L293-L296

modules/services/radicle.nix
ju1m: > Thank you very much for taking a look, @ju1m! Indeed I think we could upstream the idle timer. But maybe we don't wait for that to merge this PR...

sure!

> However, @FintanH and I am running into issues when announcing (`rad sync -a`) with this PR. I am looking into this currently. Did you manage to announce just fine?

Ah, indeed, it does not work due to `git upload-pack` requiring some `@timer` syscalls:
```
[…]
sept. 11 03:19:51 pumpkin .radicle-node-wrapped[24921]: Read channel timed out for upload-pack rad:z2364hmzZUAGy1nKdSFa1gLSoUE2M
[…]
```
But it works well with the latest revision of the PR allowing `@timer` syscalls :-)

> You might have been able to work.around decrypting by populating `$RAD_PASSPHRASE`. Anyway, we are working on improving this upstream as well.

AFAIU using `RAD_PASSPHRASE` would force me to put the passphrase in clear into my Nix config, which is why I was rather looking towards `LoadCredentialEncrypted=` which enables me to put an encrypted version of the key (with `systemd-creds`, not `ssh`'s internal encryption this time) into my Nix configs (alas, it requires systemd >= v258 for user services).

modules/programs/radicle.nix
```diff
@@ -0,0 +1,268 @@
...
+  };
+
+  publicExplorerSuffix = "$rid$path";
+in
+{
+  options = {
+    programs.radicle = {
+      enable = mkEnableOption "Radicle";
+
+      cli.package = mkPackageOption pkgs "radicle-node" { };
```
khaneliman, line 57:
```suggestion
      package = mkPackageOption pkgs "radicle-node" { nullable = true; };
```
Why is this nested under cli ? It's the package for the `program` and there are no other package options in this module. 

If we can untie the config file generation from the module's package, we can also make it nullable. But, if we can't just leave it required.

modules/programs/radicle.nix
```diff
@@ -0,0 +1,268 @@
...
+          options.node.listen = mkOption {
+            type = listOf str;
+            description = "Addresses to bind to and listen for inbound connections.";
+            default = [ ];
+            example = [ "127.0.0.1:58776" ];
+          };
+        };
+
+      };
+
```
khaneliman:
```suggestion
           options = {
              publicExplorer = mkOption {
                type = str;
                description = "HTTPS URL pattern used to generate links to view content on Radicle via the browser.";
                default = "https://app.radicle.xyz/nodes/$host/$rid$path";
                example = "https://radicle.example.com/nodes/seed.example.com/$rid$path";
              };

              node = {
                alias = mkOption {
                  type = str;
                  description = "Human readable alias for your node.";
                  default = config.home.username;
                  defaultText = lib.literalExpression "config.home.username";
                };
                listen = mkOption {
                  type = listOf str;
                  description = "Addresses to bind to and listen for inbound connections.";
                  default = [ ];
                  example = [ "127.0.0.1:58776" ];
                };
              };
            };
```

modules/programs/radicle.nix
```diff
@@ -0,0 +1,268 @@
...
+      }
+      {
+        assertion =
+          cfg.uri.rad.browser.enable -> hasSuffix publicExplorerSuffix cfg.settings.publicExplorer;
+        message = "${opt.uri.rad.browser.enable} is only compatible with a public explorer URL ending in '${publicExplorerSuffix}' but '${cfg.settings.publicExplorer}' does not end with '${publicExplorerSuffix}'.";
+      }
+    ];
+
+    home = {
+      packages = [ cfg.cli.package ];
```
khaneliman, line 186:
```suggestion
      packages = lib.mkIf (cfg.package != null) [ cfg.package ];
```

modules/programs/radicle.nix
```diff
@@ -0,0 +1,268 @@
...
+    name = "config.json";
+    path = pkgs.runCommand name { nativeBuildInputs = [ cfg.cli.package ]; } ''
+      mkdir keys
+      echo "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAID/this/is/not/a/real/key/only/a/placeholder" \
+        > keys/radicle.pub
+      cp ${format.generate name cfg.settings} ${name}
+      RAD_HOME=$PWD rad config
+      cp ${name} $out
+    '';
+  };
```
khaneliman, line 48: This feels weird, what does `rad` do here with the config that we can't just generate the json normally?

modules/services/radicle.nix
```diff
@@ -0,0 +1,245 @@
...
+    package
+    path
+    str
+    ;
+
+  cfg = config.services.radicle;
+
+  radicleHome = config.home.homeDirectory + "/.radicle";
+
+  gitPath = [ "PATH=${getBin pkgs.gitMinimal}/bin" ];
```
khaneliman, line 37: If git is a requirement of the module, I'd rather this be a package option to set the git package to use. Then a user can choose what we're adding to PATH.

tests/modules/programs/radicle/basic-configuration.nix
```diff
@@ -0,0 +1,13 @@
...
+{
+  config = {
+    programs.radicle.enable = true;
+
+    nmt.script = ''
+      assertFileContent \
+        home-files/.radicle/config.json \
+        ${./basic-configuration.json}
+    '';
+  };
```
khaneliman, line 12: For the size of these 2 modules, it would be nice to have some more helpful tests to verify all of the configuration and assertions.

modules/programs/radicle.nix
lorenzleutgeb: Validates it. Since it is an antipattern to model all configuration options in `settings`, I thought it would be good to validate that whatever the user decided to put in `settings` actually gives a sane configuration. `rad config` will exit non-zero if this is not the case.

modules/services/radicle.nix
lorenzleutgeb: That's just for the `$PATH` of the systemd user service, because `radicle-node` wants to exec `git` and friends. That is to say, it won't end up in the users' path. And `radicle-node` really only requires vanilla `git`. Do you insist on making this configurable?

modules/programs/radicle.nix
lorenzleutgeb: For future compatibility. I created the two attrsets `cli` and `uri` to split configuration of the CLI and URI handling. Note that `settings` writes the overall configuration, it affects CLI, URI handling, and also the `radicle-node` service. For CLI there just happens to be this one option so far, but there might be more in the future, and there might be a `package` option for other components in the future.

Note that there are many more components floating around, most notably Radicle Desktop and Radicle TUI, which might get `programs.radicle.{desktop,tui}.package`. See https://search.nixos.org/packages?channel=unstable&query=radicle

modules/programs/radicle.nix
lorenzleutgeb: Blocked on https://github.com/nix-community/home-manager/pull/5409#discussion_r2342314367

tests/modules/programs/radicle/basic-configuration.nix
lorenzleutgeb: Waiting for your reply to other open discussions before investing time in writing tests.

modules/services/radicle.nix
lorenzleutgeb: > AFAIU using `RAD_PASSPHRASE` would force me to put the passphrase in clear into my Nix config, [...]

Well, it depends how you would populate it. There are solutions with sops-nix that do not require this.

Anyway, I think `radicle-node` will soon gain a command line option to change the location of the secret key, and then this will have to be rethought. See https://app.radicle.xyz/nodes/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/patches/ed450c98018617be362fcb51c50a2a0f4c66ef69

modules/programs/radicle.nix
khaneliman: Actually sounds like a good use for a `checkConfig` option to support nullable package configuration. Some users intend to use home-manager for managing the files themselves and installing from other means. We could also leverage this script in `home.checks` and the check defaults to enable when the user is installing `cli.package` through home-manager. There is a similar concept of `checkConfig` in `sway`, but they do a `checkPhase` because we didn't have `home.checks` then. But, similar idea

I don't think it needs to be a blocker, though. Since a change after the module is out there wouldn't be a breaking change to add nullable package support and move the config validation to `opt-out` instead of forced.

modules/services/radicle.nix
khaneliman: It's still better to be configurable for overriding to help people keep their closures small and overcome any issues that might occur with the default package we use.

modules/programs/radicle.nix
khaneliman: *Not a blocker:* Ah alright, so the module is going to be for a suite of solutions related to `radicle` not just a single `program`. Might recommend breaking it out into a folder structure early and modularize it a bit more so that each entity has its own module that you can define/config within `programs.radicle`

modules/programs/radicle.nix
khaneliman: This can be handled whenever the config validation has an `opt-out` option.

khaneliman:
In the end, I think the only thing that would be nice before a merge is the `gitPackage` package option and some tests to verify the module behaviors. The other things can be refactorings/features whenever someone is interested in tackling the support. 

After typing that out, though... it does make me wonder if it would be worth introducing the concept of `dependencies` like we did in Nixvim so that each module doesn't need to declare it's own ancillary dependency package options...

ju1m:
I've been using this PR for a month, is there any hard blocker to get this merged?
Or is it only problems that can be solved later?

khaneliman:
> I've been using this PR for a month, is there any hard blocker to get this merged? Or is it only problems that can be solved later?

Looks like contributor stopped responding to proposed changes. We can merge, as-is, and add the desired changes in a follow up PR.

lorenzleutgeb:
Hey @khaneliman! Yes, I did stop responding here, but only because I was busy with other things. No bad intentions or anything. I could've pinged here that it'd take me a while to follow up. Sorry that I did not. Thanks for the merge, and I hope that the remaining issues can be resolved soon.

MordragT:
Hey sorry if this is obvious but I just tried to use this and I cannot really seem to get this to work. So when using programs.radicle it seems like the settings are not sufficient to directly use the cli, now I need to use `rad auth` which conflicts with the config.json symlink (read-only). Then I tried out services.radicle.node and set programs.radicle.enable to false, using `rad auth` to setup my profile. Now the systemd user service is telling me that the keystore is encrypted and a passphrase is required. Now I didn't test if setting no password for my profile works then with the radicle node service, but I feel like I might be missing something.

lorenzleutgeb:
@MordragT if you:
 1. want to share the same secret key for your user identity and your node ID (which is the standard way of doing it currently), and
 2. you have set a passphrase for the secret key, and
 3. you are using this module to start `radicle-node`

then, you will have to communicate your passphrase to `radicle-node`.

With Radicle 1.5, the current release, you have three options:
 1. Set the environment variable `RAD_PASSPHRASE` of the `radicle-node` process to your passphrase. You can do this by using home-manager and setting the environment for the `radicle-node` service.
 2. Remove the passphrase from the secret key (using OpenSSH tooling; your key is probably in `~/.radicle/keys/radicle`). Just keep in mind that this means someone copying this file is copying your Radicle identity. (This removes precondition 2. above.)
 3. Do not use the home-manager module for now and use `rad node start` instead, which will prompt you for the passphrase and then hand it to the `radicle-node` process. (This removes precondition 3. above.)

With Radicle 1.6 (which will likely be released within the next two weeks), you will have more options to resolve:
 1. It will be possible to [use systemd credentials](https://app.radicle.xyz/nodes/seed.radicle.xyz/rad:z3gqcJUoA1n9HaHKufZs5FCSGazv5/patches/cfbc8ab4edd0eddb24e56a7ce56a1a0f6d2d3d39). This is similar to suggestion 1. above for Radicle 1.5, but more secure.
 2. It will be possible to use a different secret key for your user identity and for the node ID. I would not recommend doing this just yet though, as `rad sync status` does not perfectly handle this case yet. This is also why you won't find it announced in the release notes. (This removes precondition 1. above.)

MordragT:
@lorenzleutgeb Ah thank you for the information. Then I will just use the cli for now and wait for 1.6 :+1:

ju1m:
@MordragT beware that [as far as I was able to test above](https://github.com/nix-community/home-manager/pull/5409#discussion_r2338228043), as of systemd v257.7, `LoadCredentialEncrypted=` is not supported in **user** services, it should be in systemd v258. You need it to keep your SSH key encrypted when `radicle-node.service` is not running.

To use it, you'll have to reencrypt your SSH key **for each one of your hosts and users** with:
```
cp ~/.radicle/keys/radicle{,.encrypted-backup} # In case you need to migrate to another computer
ssh-keygen -p -f ~/.radicle/keys/radicle
systemd-creds encrypt --user --uid=$USER --name xyz.radicle.node.secret ~/.radicle/keys/radicle ~/.radicle/keys/radicle.cred
```
Beware that by default `systemd-creds --user` uses `/var/lib/systemd/credential.secret`, `/etc/machine-id`, any TPM chip on the encrypting host, and the user's numeric UID and name.

I guess it would be more user friendly if `radicle` could just use something like `ssh-agent` (or `gpg-connect-agent` in SSH compatibility mode) to ask for the passphrase of the SSH key. With the activation socket enabled it would ask the passphrase at the right time.

PR #5411, opened 2024-05-16
espanso: Don't run as `background` processtype on darwin

n8henrie:
This leads to a considerably slower expansion time for snippets,
which severely degrades the user experience.

Also, the launchd plist from the source omits
this key:
https://github.com/espanso/espanso/blob/8daadcc949c35a7b7aa20b7f544fdcff83e2c5f7/espanso/src/res/macos/com.federicoterzi.espanso.plist

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/espanso.nix
+++ b/modules/services/espanso.nix
@@ -137,7 +137,6 @@ in {
           Crashed = true;
           SuccessfulExit = false;
         };
-        ProcessType = "Background";
         RunAtLoad = true;
       };
     };
--- a/tests/modules/services/espanso-darwin/launchd.plist
+++ b/tests/modules/services/espanso-darwin/launchd.plist
@@ -16,8 +16,6 @@
 	</dict>
 	<key>Label</key>
 	<string>org.nix-community.home.espanso</string>
-	<key>ProcessType</key>
-	<string>Background</string>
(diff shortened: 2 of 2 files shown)


n8henrie:
That looks like an unrelated failure to me...

n8henrie:
I'd also like to add myself to maintainers (especially for darwin issues) if okay with @lucasew @bobvanderlinden @liyangau -- if so, should that be in a separate PR I would assume?

lucasew:
> I'd also like to add myself to maintainers (especially for darwin issues) if okay with @lucasew @bobvanderlinden @liyangau -- if so, should that be in a separate PR I would assume?

LGTM. I don't have anything Darwin to test stuff.

A separate PR is a little faster to merge. If you have it from nixpkgs it also shows up there. I think it's better.

n8henrie:
Thanks for the input. https://github.com/nix-community/home-manager/pull/5413

n8henrie:
Is there a user-facing way to rerun tests? It looks like the darwin failure is while building `nmt-report-mise-zsh-integration`.

rycee:
Thanks! Rebased to master in e3ad5108f54177e6520535768ddbf1e6af54b59d 🙂

PR #5423, opened 2024-05-20
Qutebrowser: Way to define settings for each domain

willbou1:
### Description

This PR adds a way to define per-domain settings in config.py. The option is called domainSettings and contains an attribute set of attribute sets with the base domain as the name.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
- Justin Lovinger
- ivann


teto:
this doesn't seem to bring much value/typing compared to an `extraConfig` so not sure that's worth carrying in hm

somasis:
This PR has been superseded by https://github.com/nix-community/home-manager/pull/7078.

PR #5424, opened 2024-05-20
Allow path customization for awscli files

obvionaoe:
### Description

This PR changes the `programs.awscli` module to allow setting custom paths for the config files.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@anthonyroussel

### DIFF
--- a/modules/programs/awscli.nix
+++ b/modules/programs/awscli.nix
@@ -3,7 +3,14 @@
 let
   cfg = config.programs.awscli;
   iniFormat = pkgs.formats.ini { };
-
+  settingsPath =
+    if config.programs.awscli.settings.path != ""
+    then config.programs.awscli.settings.path
+    else "${config.home.homeDirectory}/.aws/config";
+  credentialsPath =
+    if config.programs.awscli.credentials.path != ""
+    then config.programs.awscli.credentials.path
+    else "${config.home.homeDirectory}/.aws/credentials";
 in {
   meta.maintainers = [ lib.maintainers.anthonyroussel ];
 
@@ -57,13 +64,21 @@ in {
   config = lib.mkIf cfg.enable {
     home.packages = [ cfg.package ];
 
-    home.file."${config.home.homeDirectory}/.aws/config" =
+    home.sessionVariables =
+      (lib.mkIf (cfg.settings.path != "") {
+        AWS_CONFIG_FILE = cfg.settings.path;
+      }) //
+      (lib.mkIf (cfg.credentials.path != "") {
+        AWS_SHARED_CREDENTIALS_FILE = cfg.credentials.path;
+      });
+
+    home.file.(settingsPath) =
       lib.mkIf (cfg.settings != { }) {
-        source =
-          iniFormat.generate "aws-config-${config.home.username}" cfg.settings;
+        source = iniFormat.generate "aws-config-${config.home.username}"
+            cfg.settings;
       };
 
-    home.file."${config.home.homeDirectory}/.aws/credentials" =
+    home.file.(credentialsPath) =
       lib.mkIf (cfg.credentials != { }) {
         source = iniFormat.generate "aws-credentials-${config.home.username}"
           cfg.credentials;


modules/programs/awscli.nix
```diff
@@ -57,13 +64,21 @@ in {
...
     home.packages = [ cfg.package ];
 
-    home.file."${config.home.homeDirectory}/.aws/config" =
+    home.sessionVariables =
+      (lib.mkIf (cfg.settings.path != "") {
+        AWS_CONFIG_FILE = cfg.settings.path;
+      }) //
+      (lib.mkIf (cfg.credentials.path != "") {
+        AWS_SHARED_CREDENTIALS_FILE = cfg.credentials.path;
+      });
```
rycee, line 73: Use `mkMerge` when combining, otherwise it mostly works by accident 🙂

```suggestion
    home.sessionVariables = mkMerge [
      (lib.mkIf (cfg.settings.path != "") {
        AWS_CONFIG_FILE = cfg.settings.path;
      })
      (lib.mkIf (cfg.credentials.path != "") {
        AWS_SHARED_CREDENTIALS_FILE = cfg.credentials.path;
      })
    ];
```

modules/programs/awscli.nix
```diff
@@ -57,13 +64,21 @@ in {
...
+    home.file.(settingsPath) =
       lib.mkIf (cfg.settings != { }) {
-        source =
-          iniFormat.generate "aws-config-${config.home.username}" cfg.settings;
+        source = iniFormat.generate "aws-config-${config.home.username}"
+            cfg.settings;
       };
 
-    home.file."${config.home.homeDirectory}/.aws/credentials" =
+    home.file.(credentialsPath) =
```
rycee, line 81:
```suggestion
    home.file.${credentialsPath} =
```

modules/programs/awscli.nix
```diff
@@ -57,13 +64,21 @@ in {
...
-    home.file."${config.home.homeDirectory}/.aws/config" =
+    home.sessionVariables =
+      (lib.mkIf (cfg.settings.path != "") {
+        AWS_CONFIG_FILE = cfg.settings.path;
+      }) //
+      (lib.mkIf (cfg.credentials.path != "") {
+        AWS_SHARED_CREDENTIALS_FILE = cfg.credentials.path;
+      });
+
+    home.file.(settingsPath) =
```
rycee, line 75:
```suggestion
    home.file.${settingsPath} =
```

modules/programs/awscli.nix
```diff
@@ -3,7 +3,14 @@
 let
   cfg = config.programs.awscli;
   iniFormat = pkgs.formats.ini { };
-
+  settingsPath =
+    if config.programs.awscli.settings.path != ""
```
rycee, line 7: Hmm, I had a quick look and there doesn't seem like there is a `path` field in `~/.aws/config` configuration file? Similarly, in the `~/.aws/credentials` file. The entries inside `programs.awscli.settings` and `programs.awscli.credentials` should match actual options honored by the aws-cli tool.

rycee: I.e., if you want to control the path of these files, then it should be set outside the `settings` and `credentials` options.

obvionaoe: Yeah, I also realized after submitting the PR that it'd make it backwards incompatible, so instead of altering the `settings` and `credentials` options it'd be better to create an extra option to specify the path of these files.

I also thought about creating an option to `xdg` called `honor` or something, that would apply the required variables to make the tools all comply with XDG Base Directory Specification. It's outside the scope of this PR though.

obvionaoe:
I addressed these changes and made the change backwards compatible. I was wondering if there's a more idiomatic way to construct the if statements in lines 8-15, though. @rycee

modules/programs/awscli.nix
```diff
@@ -2,8 +2,17 @@
...
+
   iniFormat = pkgs.formats.ini { };
 
+  settingsPath =
+    if config.programs.awscli.settingsPath != ""
+    then config.programs.awscli.settingsPath
+    else "${config.home.homeDirectory}/.aws/config";
+  credentialsPath =
+    if config.programs.awscli.credentialsPath != ""
+    then config.programs.awscli.credentialsPath
```
rycee, line 14:
```suggestion
    if cfg.settingsPath != ""
    then cfg.settingsPath
    else "${config.home.homeDirectory}/.aws/config";
  credentialsPath =
    if cfg.credentialsPath != ""
    then cfg.credentialsPath
```

modules/programs/awscli.nix
rycee: Yeah, having separate path options work much better.

Btw, there is already a [`home.preferXdgDirectories`](https://nix-community.github.io/home-manager/options.xhtml#opt-home.preferXdgDirectories) option for the purpose of forcing programs to use XDG directories.

obvionaoe:
@rycee, can you take a look again?

obvionaoe:
Still needs a review @rycee @anthonyroussel

obvionaoe:
Any chance someone could review this? @rycee @anthonyroussel

teto:
to me it's a bit out of scope. Lot of boilerplate/complexity for a niche scenario.
Could you try:
```
home.file."${config.home.homeDirectory}/toto".source = home.file."${config.home.homeDirectory}/.aws/credentials".source;
home.file."${config.home.homeDirectory}/.aws/credentials".disable = true;
```

obvionaoe:
not stale

modules/programs/awscli.nix
```diff
@@ -2,8 +2,17 @@
...
   iniFormat = pkgs.formats.ini { };
 
+  settingsPath =
+    if cfg.settingsPath != ""
+    then cfg.settingsPath
+    else "${config.home.homeDirectory}/.aws/config";
+  credentialsPath =
+    if cfg.credentialsPath != ""
+    then cfg.credentialsPath
+    else "${config.home.homeDirectory}/.aws/credentials";
```
khaneliman, line 15: Move to a let block above config or change to using a `cfg.settingsPath = lib.mkOptionDefault "${config.home.homeDirectory}/.aws/config"` to simplify logic and just make it default, as intended.

PR #5429, opened 2024-05-21
home-manager: add lib support for non-flake users

Janik-Haag:
### Description

This patch moves the lib functionality from the flake into a file and imports in both the `flake.nix` and `default.nix` which makes it accessible for non-flake users.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -24,6 +24,7 @@ in rec {
     pkgs.callPackage ./home-manager/install.nix { inherit home-manager; };
 
   nixos = import ./nixos;
+  lib = import ./lib { inherit (pkgs) lib; };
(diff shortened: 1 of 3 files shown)


Janik-Haag:
@rycee would you mind taking a look at this since I don't think there really is a maintainer for the `default.nix`/`flake.nix`

Janik-Haag:
This is still relevant.

Janik-Haag:
This is still relevant. cc @rycee

Janik-Haag:
Rebased because of #5174 cc @khaneliman

Issue #5264, opened 2024-04-13
Fcitx5 sessionVariables on Wayland


mirror-kt:
### Description
Currently if `i18n.inputMethod.enabled = "fcitx"`, the following sessionVariables are set:
https://github.com/nix-community/home-manager/blob/8fdf329526f06886b53b94ddf433848a0d142984/modules/i18n/input-method/fcitx5.nix#L27-L34

However, when using fcitx5 on wayland, the sessionVariables we want to set are different value, as wrote in the following wiki:
https://fcitx-im.org/wiki/Using_Fcitx_5_on_Wayland

I suggest adding `i18n.inputMethod.fcitx5.wayland` bool option to allow the user to choose which of the above values to set.

ailelix:
Agree, it would be better to set like in global nix
`As I spent a whole night finding out who set the ENV, much pain right there`

For now, a temporary fix can be done with:
`configuration.nix`
```nix
i18n.inputMethod = {
  enabled = "fcitx5";
  fcitx5.waylandFrontend = true;
};
```
`home/fcitx5.nix`
```nix
{ config, pkgs, ... }:
{
  i18n.inputMethod = {
    fcitx5.addons = 
    # The let-in part import rime config with packageOverride
    # As so far rime has not (and almost cannot) been merged into nix
    let
      config.packageOverride = pkgs: {
        fcitx5-rime = pkgs.fcitx5-rime.override {
          rimeDataPkgs = [
            ./rime
          ];
        };
      };
    in
    with pkgs; [
      fcitx5-gtk
      fcitx5-rime
      fcitx5-chinese-addons
    ];
  };
}
```

utzuro:
The same issue was discussed on nixpkgs (https://github.com/NixOS/nixpkgs/issues/270432) and was resolved by adding [i18n.inputMethod.fcitx5.waylandFrontend](https://mynixos.com/nixpkgs/option/i18n.inputMethod.fcitx5.waylandFrontend) option which allows user to prevent setting "GTK_IM_MODULE" and "QT_IM_MODULE" environment variables. 

I'm working on the same patch for the home-manager at the moment: https://github.com/nix-community/home-manager/pull/5431.

PR #5431, opened 2024-05-22
fcitx5: add waylandFrontend option to not set env vars

utzuro:
### Description

Fixes: https://github.com/nix-community/home-manager/issues/5264
This patch adds new option to the fcitx5, which allows to specify that user wishes to use it with wayland, so environment variables which are necessary for the X11, should not be automatically set.

The same change was already made for the nixos repo: https://github.com/NixOS/nixpkgs/pull/278765

I've confirmed on my machine, that with this change, fctix5 no longer sets  "GTK_IM_MODULE" and "QT_IM_MODULE" environment variables.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/i18n/input-method/fcitx5.nix
+++ b/modules/i18n/input-method/fcitx5.nix
@@ -25,6 +25,15 @@ in {
           Enabled Fcitx5 addons.
         '';
       };
+
+      waylandFrontend = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          Use the Wayland input method frontend.
+          See [Using Fcitx 5 on Wayland](https://fcitx-im.org/wiki/Using_Fcitx_5_on_Wayland).
+        '';
+      };
     };
   };
 
@@ -33,11 +42,12 @@ in {
 
     home.sessionVariables = {
       GLFW_IM_MODULE = "ibus"; # IME support in kitty
-      GTK_IM_MODULE = "fcitx";
-      QT_IM_MODULE = "fcitx";
       XMODIFIERS = "@im=fcitx";
       QT_PLUGIN_PATH =
         "$QT_PLUGIN_PATH\${QT_PLUGIN_PATH:+:}${fcitx5Package}/${pkgs.qt6.qtbase.qtPluginPrefix}";
+    } // lib.optionalAttrs (!cfg.waylandFrontend) {
+      GTK_IM_MODULE = "fcitx";
+      QT_IM_MODULE = "fcitx";
     };
 
     systemd.user.services.fcitx5-daemon = {
--- a/tests/modules/i18n/input-method/fcitx5-configuration.nix
+++ b/tests/modules/i18n/input-method/fcitx5-configuration.nix
@@ -5,10 +5,12 @@
 
   i18n.inputMethod = {
     enabled = "fcitx5";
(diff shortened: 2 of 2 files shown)


tests/modules/i18n/input-method/fcitx5-configuration.nix
```diff
@@ -6,6 +6,7 @@
   i18n.inputMethod = {
     enabled = "fcitx5";
     fcitx5.addons = with pkgs; [ fcitx5-chinese-addons ];
+    fcitx5.waylandFrontend = true;
```
utzuro, line 9: I'm not sure how to test that environment variables are not set when by fcitx5 when waylandFrontend is set to true, so I just added that option to the existed test. Please guide me, if there is a good way to approach this test.

modules/i18n/input-method/fcitx5.nix
```diff
@@ -18,6 +18,14 @@ in {
           Enabled Fcitx5 addons.
         '';
       };
+      waylandFrontend = mkOption {
+        type = types.bool;
+        default = false;
+        description = lib.mdDoc ''
```
rycee, line 24: No need for `mdDoc`, it is the default anyway

```suggestion
        description = ''
```

tests/modules/i18n/input-method/fcitx5-configuration.nix
rycee: The most straight forward way would probably be to use `assertFileNotRegex`. An example: https://github.com/nix-community/home-manager/blob/5d151429e1e79107acf6d06dcc5ace4e642ec239/tests/modules/programs/gradle/empty-settings.nix#L12-L15

modules/i18n/input-method/fcitx5.nix
```diff
@@ -18,6 +18,14 @@ in {
           Enabled Fcitx5 addons.
         '';
       };
+      waylandFrontend = mkOption {
```
rycee, line 21: I suggest always putting a space between options, otherwise it is a bit tricky to tell them apart.

```suggestion

      waylandFrontend = mkOption {
```

rycee (commented):

utzuro:
@rycee 
> Thanks for the contribution! I've added a few comments.
Thank you! I've fixed the things you mentioned. PR is ready for another review.

utzuro:
@rycee 
Small reminder here! The PR is ready for another review.

utzuro:
@rycee Sorry for another reminder! The PR became stale already, but I'd really want to merge it if everything is OK.

RAVENz46:
@rycee 
I apologize for being so busy, but could you please merge this PR soon?

Dreamail:
waiting for merge too.

utzuro:
This PR https://github.com/nix-community/home-manager/pull/5770 was merged 2 weeks ago, while this one is still being reviewed, so I had to resolve some conflicts.

@rycee Could you give this PR a look? Maybe if there are other people who I can communicate with, could you kindly mention them? This PR is already 8 months old, I think, and I don't know what can I do on my side to make it merged.

utzuro:
@teto I've noticed that you merged this PR https://github.com/nix-community/home-manager/pull/5770, so maybe you could take a look on this one too?

PR #5432, opened 2024-05-22
thunderbird: add message filters option

javaes:
### Description

This adds an option to the Thunderbird module which declares message filters to easily replicate filter settings on other machines.

To make it more easily usable I could either extend its documentation or I could create attrsets as enums for the `type` and `action` fields of a filter. What do you think?

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@d-dervishi @ethorsoe

### DIFF
--- a/modules/programs/thunderbird.nix
+++ b/modules/programs/thunderbird.nix
@@ -124,6 +124,28 @@ let
     '') prefs)}
     ${extraPrefs}
   '';
+
+  filterIniHeader = ''
+    version="9"
+    logging="no"
+  '';
+
+  mkFilterToIniStr = f: ''
+    name="${f.name}"
+    enabled="${if f.enabled then "yes" else "no"}"
+    type="${f.type}"
+    action="${f.action}"
+    actionValue="${f.actionValue}"
+    condition="${f.condition}"
+  '';
+
+  mkFilterListToIni = filters:
+    filterIniHeader + concatStrings (map (f: mkFilterToIniStr f) filters);
+
+  getAccountsForProfile = profileName: accounts:
+    (filter (a:
+      a.thunderbird.profiles == [ ]
+      || any (p: p == profileName) a.thunderbird.profiles) accounts);
 in {
   meta.maintainers = with hm.maintainers; [ d-dervishi jkarlson ];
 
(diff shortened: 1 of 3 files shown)


modules/programs/thunderbird.nix
```diff
@@ -295,6 +317,28 @@ in {
                 argument is an automatically generated identifier.
               '';
             };
+
+            msgFilters = mkOption {
+              type = (types.listOf (types.attrs));
```
ethorsoe, line 322: shouldn't this be listOf $module, I guess I also like attrsOf since it can be overridden more easily, but I guess this one is ordered?

ethorsoe:
> Thunderbird module maintainers are no longer active. It looks like there are no assigned reviewers.

Yeah sorry have been busy with family and dayjob, in general it's reasonable to bypass awol maintainers.

modules/programs/thunderbird.nix
hakan-demirli: Thunderbird requires filters to be ordered as they are applied sequentially, so listOf is necessary to preserve the correct order. For example if you are deleting certain emails no need to move them around first since they will be deleted regardless. So, filters are generally setup up like a pipeline.

347Online:
Where were we at with this? I went to file an issue about exactly this and stumbled upon this PR :D

modules/programs/thunderbird.nix
```diff
@@ -124,6 +124,28 @@ let
     '') prefs)}
     ${extraPrefs}
   '';
+
+  filterIniHeader = ''
+    version="9"
+    logging="no"
```
347Online, line 130: Should logging have an option associated with it?

347Online (approved):

modules/programs/thunderbird.nix
```diff
@@ -124,6 +124,28 @@ let
...
+  '';
+
+  mkFilterToIniStr = f: ''
+    name="${f.name}"
```
ethorsoe, line 134: I guess, if you are explicitly listing these, maybe you should also document them separately, alternatively I am moderately sure nixpkgs has some generators for things like this, maybe ini?

khaneliman:
https://github.com/nix-community/home-manager/pull/6652

PR #5437, opened 2024-05-23
hyprland: onChange: check XDG_RUNTIME_DIR as well

Misterio77:
### Description

Fixes Hyprland's onChange check for versions >= 0.40.0, as the socket has now moved to $XDG_RUNTIME_DIR.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@fufexan

### DIFF
--- a/modules/services/window-managers/hyprland.nix
+++ b/modules/services/window-managers/hyprland.nix
@@ -237,7 +237,8 @@ in {
 
       onChange = lib.mkIf (cfg.package != null) ''
         ( # Execute in subshell so we don't poision environment with vars
-          if [[ -d "/tmp/hypr" ]]; then
+          XDG_RUNTIME_DIR=''${XDG_RUNTIME_DIR:-/run/user/$(id -u)}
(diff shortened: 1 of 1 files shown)


modules/services/window-managers/hyprland.nix
```diff
@@ -237,7 +237,8 @@ in {
 
       onChange = lib.mkIf (cfg.package != null) ''
         ( # Execute in subshell so we don't poision environment with vars
```
fufexan, line 239: This comment can also be removed, since we're no longer setting `$HIS` a bunch of times.

Misterio77: Nice! Addressed.

fufexan (approved):
LGTM, thanks!

rycee:

jtrees:
Could we backport this to 24.05? I'm not familiar with the process but I can open a PR with the fix commit against `release-24.05` if that's all it takes.

rycee:
@jtrees Good point, I've cherry-picked this to the 24.05 branch. Please give it a try.

jtrees:
@rycee Thanks! Reloading works again. ❤️

Issue #5384, opened 2024-05-07
Support `zed-editor` 


a99984b1799:
### Description

[Zed](https://zed.dev/) was added to Nixpkgs last month, support for configuration/extensions/themes would be great.

libewa:
working on this in #5455

pxp9:
#5455 got merged , should this issue be closed ?

Ekleog:
FWIW, if anyone has the same issues as I did and comes here: if when trying to enable github copilot zed is doing nonsense, it's because settings.json is read-only and it's trying to edit it.

I temporarily replaced it with its contents, started zed, went through the graphical configuration process, and then copy the added parameters to my nix file and start using home-manager again.

Unfortunately zed seems to not support having multiple config files and only one read-write one, so I'm not sure what could be done better :'(

libewa:
@a99984b1799 please close

PR #5455, opened 2024-05-26
Zed editor

libewa:
### Description

Add a module for the Zed editor.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -1770,6 +1770,16 @@ in {
           ideas from mutt.
         '';
       }
+
+      {
+        time = "2024-10-17T13:07:55+00:00";
+        message = ''
+          A new module is available: 'programs.zed-editor'.
+
+          Zed is a fast text editor for macOS and Linux.
+          See https://zed.dev for more.
+        '';
+      }
     ];
   };
 }
--- a/modules/programs/zed-editor.nix
+++ b/modules/programs/zed-editor.nix
@@ -1,0 +1,89 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.programs.zed-editor;
+  jsonFormat = pkgs.formats.json { };
+
+  mergedSettings = cfg.userSettings // {
+    # this part by @cmacrae
+    auto_install_extensions = lib.genAttrs cfg.extensions (_: true);
+  };
+in {
+  meta.maintainers = [ hm.maintainers.libewa ];
+
+  options = {
+    # TODO: add vscode option parity (installing extensions, configuring
+    # keybinds with nix etc.)
+    programs.zed-editor = {
+      enable = mkEnableOption
+        "Zed, the high performance, multiplayer code editor from the creators of Atom and Tree-sitter";
+
+      package = mkPackageOption pkgs "zed-editor" { };
+
+      userSettings = mkOption {
+        type = jsonFormat.type;
+        default = { };
+        example = literalExpression ''
+          {
+            features = {
+              copilot = false;
+            };
+            telemetry = {
+              metrics = false;
+            };
+            vim_mode = false;
+            ui_font_size = 16;
+            buffer_font_size = 16;
+          }
+        '';
+        description = ''
+          Configuration written to Zed's {file}`settings.json`.
+        '';
+      };
+
+      userKeymaps = mkOption {
+        type = jsonFormat.type;
+        default = { };
+        example = literalExpression ''
+          [
+            {
+              context = "Workspace";
+              bindings = {
+                ctrl-shift-t = "workspace::NewTerminal";
+              };
+            };
+          ]
+        '';
+        description = ''
+          Configuration written to Zed's {file}`keymap.json`.
+        '';
+      };
+
+      extensions = mkOption {
+        type = types.listOf types.str;
+        default = [ ];
+        example = literalExpression ''
+          [ "swift" "nix" "xy-zed" ]
+        '';
+        description = ''
+          A list of the extensions Zed should install on startup.
+          Use the name of a repository in the [extension list](https://github.com/zed-industries/extensions/tree/main/extensions).
+        '';
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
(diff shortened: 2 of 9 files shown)


libewa:
closes #5384

hervyqa:
GREAT. i can't wait...

modules/programs/zed-editor.nix
```diff
@@ -0,0 +1,97 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.programs.zed-editor;
+  jsonFormat = pkgs.formats.json { };
+
+  configDir = if pkgs.stdenv.hostPlatform.isDarwin then
+    "Library/Application Support/zed" # TODO: see if this is actually the correct directory
```
joefiorini, line 10: Many apps support using XDG dirs even on Mac. I'm pretty sure Zed does since I have a `.config/zed` immediately after downloading and running it. Can this just use the xdg dirs regardless of platform and allow an override if needed?

linuxmobile: I am waiting for the fix for this to be merged, but I see no activity from the mantainer. Can someone else take over or are we just waiting?

tymscar: I can confirm that config directory is in `~/.config/zed` both on linux and mac

modules/programs/zed-editor.nix
```diff
@@ -0,0 +1,97 @@
...
+  else
+    "${config.xdg.configHome}/zed";
+  configFilePath = "${configDir}/settings.json";
+  keymapFilePath = "${configDir}/keymap.json";
+  mergedSettings = cfg.userSettings // {
+    auto_install_extensions = cfg.extensions;
```
cmacrae, line 16: the `auto_install_extensions` configuration option doesn't work like this. it is a map of k/v pairs where the key is the extension name and the value is a boolean that determines if the extension is installed on startup.

so, if you want to keep this `extensions` option for the module as a QoL feature, since you're taking this merge approch already, you could instead use an expression to iterate over the provided extensions and provide the correct syntax. here's what i'm doing with a module i started work on before realising this PR existed :)

```nix
xdg.configFile."zed/settings.json" = mkIf (cfg.settings != { }) {
  source = (jsonFormat.generate "zed.json" (cfg.settings // {
    auto_install_extensions = lib.listToAttrs
      (map (ext: lib.nameValuePair ext true) cfg.extensions);
  }));
};
```

libewa: thanks! do you want me to add you as a coauthor on that commit @cmacrae?

cmacrae: very considerate of you, thanks! sure, that would be nice :)

libewa: yw

MangoIV:
hi @libewa for me the biggest problem with zed on linux is that it actually doesn't start up properly at all. e.g. opening a project folder is blocked on a timeout, I think this is because zed tries to side load a couple of dynamically linked executables. Can we fix this somehow?

modules/programs/zed-editor.nix
```diff
@@ -0,0 +1,89 @@
...
+            "features": {
+              "copilot": false
+            },
+            "telemetry": {
+              "metrics": false
+            },
+            "vim_mode": false,
+            "ui_font_size": 16,
+            "buffer_font_size": 16
+          }
```
uncenter, line 42: This example is in actual JSON, it should be Nix here.

modules/programs/zed-editor.nix
```diff
@@ -0,0 +1,89 @@
...
+        default = { };
+        example = literalExpression ''
+          [
+            {
+              "context" = "Workspace";
+              "bindings" = {
+                "ctrl-shift-t" = "workspace::NewTerminal";
+              };
+            };
+          ]
```
isabelroses, line 59: same as @uncenter 's comment

nxy7:
Hello, sorry to ping you directly @libewa but did you lose interest in pushing this PR? Asking because I'm not sure if I should try to fork it or patiently wait since it seems to me like all suggestions are pretty minor and should be easy to implement.

tymscar:
I would also love if this goes through!

modules/programs/zed-editor.nix
libewa: this looks weird because the keys are encased in `"`, but unlike the other example, this is actually valid nixlang (however, this is not how you write it in most cases, so i'm changing that too

cmacrae (approved):

modules/misc/news.nix
```diff
@@ -1656,6 +1656,17 @@ in {
           See https://codeberg.org/dnkl/yambar for more.
         '';
       }
+
+      {
+        time = "2024-05-26T14:43:44+00:00";
+        message = ''
+          A new module is available: 'programs.zed-editor'.
+
+          Zed is a fast text editor for Mac and Linux.
```
uncenter, line 1665:
```suggestion
          Zed is a fast text editor for macOS and Linux.
```

uncenter (approved):

pxp9:
Sorry for pings.

why is taking so long @libewa ?

@sumnerevans could you re-review plz ?

libewa:
@pxp9 yeah gh is still waiting for @sumnerevans' rereview

modules/programs/zed-editor.nix
```diff
@@ -0,0 +1,82 @@
...
+with lib;
+
+let
+  cfg = config.programs.zed-editor;
+  jsonFormat = pkgs.formats.json { };
+
+  mergedSettings = cfg.userSettings // {
+    # this part by @cmacrae
+    auto_install_extensions = lib.listToAttrs
+      (map (ext: lib.nameValuePair ext true) cfg.extensions);
```
h7x4, line 12:
```suggestion
    auto_install_extensions = lib.genAttrs cfg.extensions (_: true);
```

tymscar:
Are there any blockers I can help with for this?

rycee:
Thanks! I squashed all the commits, cleaned up the tests a bit, and merged to master 🙂

PR #5456, opened 2024-05-26
mpd: allow using urls for musicDirectory

TotalChaos05:
### Checklist

- [x ] Change is backwards compatible.

- [ x] Code formatted with `./format`.

- [x ] Code tested through `nix-shell --pure tests -A run.all` or `nix develop --ignore-environment .#all` using Flakes.

- [x ] Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef). (none needed)

- [x ] Commit messages are formatted like

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@rembo10 
@emilazy

### DIFF
--- a/modules/services/mpd.nix
+++ b/modules/services/mpd.nix
@@ -51,7 +51,8 @@ in {
       };
 
       musicDirectory = mkOption {
-        type = with types; either path str;
+        type = with types;
+          either path (strMatching "(http|https|nfs|smb)://.+");
         defaultText = literalExpression ''
           ''${home.homeDirectory}/music    if state version < 22.11
           ''${xdg.userDirs.music}          if xdg.userDirs.enable == true


rembo10:
LGTM :+1:

teto:
I am not sure I get the change: do you just check the string is an URI ?

TotalChaos05:
@teto it would only allow a path (or something convertible to it) earlier
this allows urls too

TotalChaos05:
I forgot to commit after using `./format`
It should be fixed now

rycee:
With `type = with types; either path str;`, shouldn't it already allow for strings such as `https://…`?

But perhaps it helps documentation to have a stricter type, but then would have to allow for `~/` as well considering the test error:

    error: A definition for option `services.mpd.musicDirectory' is not of type `path or string matching the pattern (http|https|nfs|smb)://.+'. Definition values:
    - In `/home/runner/work/home-manager/home-manager/tests/modules/programs/ncmpcpp-linux/ncmpcpp-issue-3560.nix': "~/music"

TotalChaos05:
I think I was just really tired when I wrote this PR, because you are correct about it already allowing strings.
Apparently my issue was with mpdris2, because it only allows paths, and it gets auto-set with the value from `services.mpd.musicDirectory`

rycee:
Fair enough, glad it got resolved in any case 🙂

PR #5462, opened 2024-05-27
home-manager: check FQDN for '--flake .' attribute

gauravjuvekar:
### Description

Since nixpkgs commit 993baa587c4b82e791686f6ce711bcd4ee8ef933, `networking.hostName` is not allowed to be a FQDN.

Add `hostname --fqdn` to the default flake attribute names that are searched.

If

    netorking.hostname = "hostname";
    networking.domain = "example.com";

is set in the system NixOS configuration, this allows defining

    homeConfigurations."username@hostname.example.com" = ...

and still use

    home-manager switch --flake .

instead of having to type out

    home-manager switch --flake .#$(whoami)@$(hostname --fqdn)

### Checklist

    - Technically no if someone happens to have both `homeConfigurations."username@hostname"` and `homeConfigurations."username@hostname.example.com"`.

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/home-manager/home-manager
+++ b/home-manager/home-manager
@@ -198,9 +198,9 @@ function setFlakeAttribute() {
                 ;;
             *)
                 local name="$USER"
-                # Check both long and short hostnames; long first to preserve
+                # Check FQDN, long, and short hostnames; long first to preserve
                 # pre-existing behaviour in case both happen to be defined.
-                for n in "$USER@$(hostname)" "$USER@$(hostname -s)"; do
+                for n in "$USER@$(hostname --fqdn)" "$USER@$(hostname)" "$USER@$(hostname -s)"; do
                     if [[ "$(nix eval "$flake#homeConfigurations" --apply "x: x ? \"$n\"")" == "true" ]]; then
                         name="$n"
                         if [[ -v VERBOSE ]]; then


rycee:
Thanks, looks good! Merged to master now 🙂

ilyagr:
After this PR (I believe), I'm getting the following error on Mac OS:

```console
$ home-manager switch --impure
hostname: illegal option -- -
usage: hostname [-fs] [name-of-host]
```

gauravjuvekar:
Yep, this PR is the likely cause.
I'll change it to `hostname -f` as seems like macos doesn't support the long style option name `--fqdn` (based on the `usage: line`)

gauravjuvekar:
https://github.com/nix-community/home-manager/pull/5471

bryango:
If the hostname is not set in `/etc/hosts`, the command may fail with:
```console
$ hostname -f
hostname: Unknown host
```
This shows up in the `home-manager switch` printout and is mildly annoying. Could it be changed to:
```bash
hostname -f 2>/dev/null
```
to suppress this inconsequential warning? Thank you!

gauravjuvekar:
Ack, I'll make a PR some time this week. Could you file a separate issue for this and tag me?

PR #5463, opened 2024-05-27
tests: fix Darwin failure

midchildan:
### Description

This fixes a broken overlay in the mpv tests. There were multiple errors and none of them were specific to Darwin despite the tests currently succeeding on Linux.

### Checklist

  I only ran `mpv-example-settings` and `mpv-invalid-settings`, since this PR only affects those.

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/tests/modules/programs/mpv/mpv-stubs.nix
+++ b/tests/modules/programs/mpv/mpv-stubs.nix
@@ -1,25 +1,26 @@
-{ pkgs, ... }:
-
 {
   nixpkgs.overlays = [
-    (self: super: {
-      mpvScript = pkgs.runCommandLocal "mpvScript" { scriptName = "something"; }
+    (final: prev: {
+      mpvScript = prev.runCommandLocal "mpvScript" { scriptName = "something"; }
         "mkdir $out";
 
-      mpv-unwrapped = super.mpv-unwrapped.overrideAttrs {
-        builder = pkgs.writeShellScript "dummy" ''
-          PATH=${pkgs.coreutils}/bin
+      mpv-unwrapped = let
+        lua = prev.emptyDirectory.overrideAttrs {
+          luaversion = "0";
+          passthru.withPackages = pkgsFn: prev.emptyDirectory;
+        };
+        mpv-unwrapped' = prev.mpv-unwrapped.override { inherit lua; };
+      in mpv-unwrapped'.overrideAttrs {
+        buildInputs = [ ];
+        nativeBuildInputs = [ ];
+        builder = prev.writeShellScript "dummy" ''
+          PATH=${final.coreutils}/bin
           mkdir -p $dev $doc $man $out/bin $out/Applications/mpv.app/Contents/MacOS
           touch $out/bin/{mpv,umpv} \
                 $out/Applications/mpv.app/Contents/MacOS/{mpv,mpv-bundle}
           chmod +x $out/bin/{mpv,umpv} \
                    $out/Applications/mpv.app/Contents/MacOS/{mpv,mpv-bundle}
         '';
-      };
-
-      lua = pkgs.emptyDirectory.overrideAttrs {
-        luaversion = "0";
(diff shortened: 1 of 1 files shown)


tests/modules/programs/mpv/mpv-stubs.nix
```diff
@@ -1,26 +1,27 @@
-{ pkgs, ... }:
```
midchildan, line 1: The overlay was referencing `pkgs`, which lead to an infinite recursion between `pkgs` and the overlay.

tests/modules/programs/mpv/mpv-stubs.nix
```diff
@@ -1,26 +1,27 @@
...
                 $out/Applications/mpv.app/Contents/MacOS/{mpv,mpv-bundle}
           chmod +x $out/bin/{mpv,umpv} \
                    $out/Applications/mpv.app/Contents/MacOS/{mpv,mpv-bundle}
         '';
       };
-
-      lua = pkgs.emptyDirectory.overrideAttrs {
-        luaversion = "0";
-        withPackages = ps: pkgs.emptyDirectory;
-      };
```
midchildan, line 23: Replacing Lua globally broke other packages.

tests/modules/programs/mpv/mpv-stubs.nix
```diff
@@ -1,26 +1,27 @@
...
-          PATH=${pkgs.coreutils}/bin
+      mpv-unwrapped = let
+        lua = prev.emptyDirectory.overrideAttrs {
+          luaversion = "0";
+          passthru.withPackages = pkgsFn: prev.emptyDirectory;
+        };
+        mpv-unwrapped' = prev.mpv-unwrapped.override { inherit lua; };
+      in mpv-unwrapped'.overrideAttrs {
+        buildInputs = [ ];
+        nativeBuildInputs = [ ];
```
midchildan, line 15: I cleared `buildInputs` and `nativeBuildInputs` to reduce bloat not required for the tests.

tests/modules/programs/mpv/mpv-stubs.nix
```diff
@@ -1,26 +1,27 @@
...
+      mpvScript = prev.runCommandLocal "mpvScript" { scriptName = "something"; }
         "mkdir $out";
 
-      mpv-unwrapped = super.mpv-unwrapped.overrideAttrs {
-        builder = pkgs.writeShellScript "dummy" ''
-          PATH=${pkgs.coreutils}/bin
+      mpv-unwrapped = let
+        lua = prev.emptyDirectory.overrideAttrs {
+          luaversion = "0";
+          passthru.withPackages = pkgsFn: prev.emptyDirectory;
```
midchildan, line 10: I put `withPackages` inside `passthru` to avoid another evaluation failure.

```
           19|         luaversion = "0";
           20|         withPackages = pkgsFn: prev.emptyDirectory;
             |         ^
           21|       };

       error: cannot coerce a function to a string
```

tests/modules/programs/mpv/mpv-stubs.nix
```diff
@@ -1,26 +1,27 @@
...
+        };
+        mpv-unwrapped' = prev.mpv-unwrapped.override { inherit lua; };
+      in mpv-unwrapped'.overrideAttrs {
+        buildInputs = [ ];
+        nativeBuildInputs = [ ];
+        builder = prev.writeShellScript "dummy" ''
+          PATH=${prev.coreutils}/bin
```
rycee, line 17: I think you'll want to refer to the final coreutils, otherwise it'll refer to some intermediate version that causes a rebuild.
```suggestion
          PATH=${final.coreutils}/bin
```

rycee:
Awesome! I've been meaning to look into this failing test but this looks better than anything I could have come up with 😀

midchildan:
Thanks! I applied the suggestion.

PR #5467, opened 2024-05-29
kdeconnect: fix service with 24.05 package version

Liassica:
### Description

The 24.05 update for KDE Connect moved the kdeconnectd binary from `/libexec` to `/bin`, so this fix will check the version of the package used and set the path accordingly.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@adisbladis

### DIFF
--- a/modules/services/kdeconnect.nix
+++ b/modules/services/kdeconnect.nix
@@ -48,7 +48,12 @@ in {
 
         Service = {
           Environment = "PATH=${config.home.profileDirectory}/bin";
-          ExecStart = "${cfg.package}/libexec/kdeconnectd";
+          ExecStart =
+            if strings.versionAtLeast (versions.majorMinor cfg.package.version)
+            "24.05" then
+              "${cfg.package}/bin/kdeconnectd"
+            else
+              "${cfg.package}/libexec/kdeconnectd";
           Restart = "on-abort";
         };
       };


modules/services/kdeconnect.nix
```diff
@@ -48,7 +47,12 @@ in {
 
         Service = {
           Environment = "PATH=${config.home.profileDirectory}/bin";
-          ExecStart = "${cfg.package}/libexec/kdeconnectd";
+          ExecStart =
+            if strings.versionAtLeast (versions.majorMinor cfg.package.version)
+            "24.05" then
+              "${cfg.package}/bin/kdeconnectd"
+            else
+              "${cfg.package}/libexec/kdeconnectd";
```
rycee, line 55: I suggest simply doing

```suggestion
          ExecStart = "${cfg.package}/bin/kdeconnectd";
```

since it appears to me that version 24.05 has propagated to all the unstable channels, so no need to consider older versions.

rycee:
Thanks for the contribution, I think you can simplify this a bit as per my inline comment 🙂

modules/services/kdeconnect.nix
Liassica: Alright :)

Liassica: Actually--I just remembered why I did it like that in the first place. The Qt5 version of KDE Connect is still available even in unstable *and* is still the default for this module in Home Manager, so if I did it like you suggested it'd actually leave it broken by default.

rycee: Fair enough, merged as-is 🙂

Issue #4807, opened 2023-12-24
bug: starship wrong path if `use-xdg-base-directories = true` is set


hakan-demirli:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Removing `use-xdg-base-directories = true` solves the issue. However, this is just a workaround.
Error:
```bash: /home/USERNAME/.nix-profile/bin/starship: No such file or directory```

home.nix
```nix
  programs.direnv.enable = true;
  programs.bash.enable = true;
  programs.starship.enable = true;
```
 .config/nix/nix.conf 
```bash
experimental-features = nix-command flakes
max-jobs = auto
use-xdg-base-directories = true
```
Another user also mentioned encountering a similar issue with the fish shell.
https://github.com/nix-community/home-manager/issues/2316#issuecomment-1712025562

### Maintainer CC

@rycee @marsam 

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.1.69, NixOS, 24.05 (Uakari), 24.05.20231222.6df37dc`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.18.1`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`
```

hakan-demirli:
Looks like this PR https://github.com/nix-community/home-manager/pull/4422#issuecomment-1859149515 addressed the issue.
However, I am using the latest commit so maybe the problem is smth different.
```json

    "home-manager": {
      "inputs": {
        "nixpkgs": [
          "nixpkgs"
        ]
      },
      "locked": {
        "lastModified": 1703368619,
        "narHash": "sha256-ZGPMYL7FMA6enhuwby961bBANmoFX14EA86m2/Jw5Jo=",
        "owner": "nix-community",
        "repo": "home-manager",
        "rev": "a2523ea0343b056ba240abbac90ab5f116a7aa7b",
        "type": "github"
      },
...
```

hakan-demirli:
I just noticed It also created a symlink from my home dir to nixstore
```~/.profile -> /nix/store/m9y2zla1ywzbwyrszym20p1rmfdrgi76-home-manager-files.config```

I think this shouldn't have happened since `use-xdg-base-directories = true` enforces xdg base specification and profile should be in `~/.local/state/nix`.

[...]

PR #5468, opened 2024-05-29
starship: fix broken path with XDG dirs

madsrumlenordstrom:
### Description

This fixes the issue described in https://github.com/nix-community/home-manager/issues/4807.
The starship modules uses the executable ```~/.nix-profile/bin/starship``` even when ```nix.use-xdg-base-directories = true;```. Now it just uses the absolute path.

I did a quick grep on the code base and I noticed the pattern ```"${config.home.profileDirectory}/bin/..."```
 does show up a few times and my guess is that they will also break when using XDG directories. Maybe this should be fixed in the future?
### Checklist

### DIFF
--- a/modules/programs/starship.nix
+++ b/modules/programs/starship.nix
@@ -8,7 +8,7 @@ let
 
   tomlFormat = pkgs.formats.toml { };
 
-  starshipCmd = "${config.home.profileDirectory}/bin/starship";
+  starshipCmd = lib.getExe cfg.package;
 
 in {
   meta.maintainers = [ ];


rycee:
Unfortunately this won't work since it will break existing shells when switching. Basically if you have a long running shell then it will refer to the starship binary that was present at the time the shell started. Eventually that binary may be GC'd and the existing shell will become unusable.

So for Starship (and similar shell prompt tools) we need to refer to a stable path, which historically has been the purpose of `home.profileDirectory`. And in principle it should work also when using Nix XDG directories, see https://github.com/nix-community/home-manager/blob/29c69d9a466e41d46fd3a7a9d0591ef9c113c2ae/modules/home-environment.nix#L528-L534

Specifically, your configuration needs to have

```nix
nix.enable = true;
nix.settings.use-xdg-base-directories = true;
```

madsrumlenordstrom:
Thanks for the explanation!
I now also realize that the problem I am experiencing with starship is because I am using two separate configurations for NixOS and home-manager. This means my home-manager configuration does not see my option ```config.nix.use-xdg-base-directories = true;``` that I set in my NixOS configuiration. However home-manager still uses the XDG directories and when I run:
```which starship```
I get:
```/home/rumle/.local/state/nix/profile/bin/starship```
Which is what I want. However my```.config/fish/config.fish``` still has these lines:
```fish
    if test "$TERM" != dumb
        eval (/home/rumle/.nix-profile/bin/starship init fish)

    end
```

If I set ```nix.use-xdg-base-directories = false;```  in my NixOS config and rebuild my NixOS config and then rebuild my home-manager config I then get the ```~/.nix-profile``` directory with all my binaries etc. and my starship prompt works as intended.

Doesn't this mean that a home-manager configuration basically is not stateless as it depends on the state of nix settings? Is there anything I can do about this? Or can I help solve this issue?

rycee:
Unfortunately there is no way for HM to, in general, know at configuration build time what the Nix system configuration is at configuration application time. This is why it attempts to guess those values. For your setup it guesses wrong because it, as you point out, cannot see that Nix is having the XDG configuration on the system level.

If you are using the NixOS module then it would be possible to access the system Nix configuration, check if it is set to use XDG directories, and forcibly set the profile path accordingly:

``` nix
home.profileDirectory = lib.mkForce "${config.xdg.stateHome}/nix/profile"
```

If you are using a standalone installation, you would have to set the above line directly in your HM configuration.

madsrumlenordstrom:
Thanks a lot!
I found another temporary fix, that I think is slightly cleaner than using ```lib.mkForce```. You can actually just set:
```nix
nix = {
  enable = true;
  package = pkgs.nix;
  settings.use-xdg-base-directories = true;
};
```
in your HM config when you use a standalone HM installation.

The problem with both solutions is that ```nix.settings.use-xdg-base-directories``` needs to be the same in the HM config and NixOS config, when using standalone installation, which is annoying for maintenance.

> If you are interested, then I encourage you to make an attempt at implementing the fix in the NixOS module. I think it basically boils down to setting the option if the system configuration is set to XDG and this is not overridden in the HM Nix configuration.

I will have a look to see if it is possible to implement a fix. And by NixOS module you mean ```home-manager/modules/misc/nix.nix``` right?

I will close this PR.

PR #5482, opened 2024-06-02
programs.waybar: remove modules-* from defaults

teto:
I prefer to handle modules-* imperatively for more flexibility, I do this via waybar "includes" (https://github.com/Alexays/Waybar/wiki/Configuration). Current HM defaults modules-* to `[]` instead of null for most other settings, so they get written to ~/.config/waybar/config even though I've configured nothing.

And they take precedence over the impartives ones because of how waybar's includes work:

====
Paths to additional configuration files.
Each file can contain a single object with any of the bar configuration options. In case of duplicate options, the first defined value takes precedence, i.e. including file -> first included file -> etc. Nested includes are permitted, but make sure to avoid circular imports. For a multi-bar config, the include directive affects only current bar configuration object. ====

Since the default is [], they dont bring naything so defaulting to null is a better choice here.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/waybar.nix
+++ b/modules/programs/waybar.nix
@@ -69,8 +69,8 @@ let
         };
 
         modules-left = mkOption {
-          type = listOf str;
-          default = [ ];
+          type = nullOr (listOf str);
+          default = null;
           description = "Modules that will be displayed on the left.";
           example = literalExpression ''
             [ "sway/workspaces" "sway/mode" "wlr/taskbar" ]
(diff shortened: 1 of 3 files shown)


teto:
cc @berbiche

berbiche (approved):
LGTM, but I wonder if this could break anyone's configuration 🤔

teto:
In my short test waybar doesn't enable any modules by default, which is equal to our current default `[]` so I dont think it should break anyone's config. If anything, if they included a manual config like I want, with a `modules-*` defined, their waybar might show modules that were previously ignored by waybar because of our previous default  `[]`.

rycee:
Fair enough, merged to master now.

Issue #5481, opened 2024-06-02
bug: bat cache should not always be generated


dawidd6:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

`programs.bat` adds an activation script which is run always when this module is enabled.
This script, in fact, should only be run when custom syntaxes or themes are specified in the module, there is no need to run `bat cache --build` if user does not specify said options in their configuration.

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.6.32, NixOS, 24.05 (Uakari), 24.05.20240531.805a384`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.18.2`
 - nixpkgs: `/nix/store/i1aw9jjgxcvyd642s12kw3iasmarwd42-source`
```

PR #5487, opened 2024-06-03
bat: skip activation script if not needed

dawidd6:
### Description

Closes: https://github.com/nix-community/home-manager/issues/5481

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/bat.nix
+++ b/modules/programs/bat.nix
@@ -162,14 +162,15 @@ in {
       # NOTE: run `bat cache --build` in an empty directory to work
       # around failure when ~/cache exists
       # https://github.com/sharkdp/bat/issues/1726
-      home.activation.batCache = hm.dag.entryAfter [ "linkGeneration" ] ''
-        (
-          export XDG_CACHE_HOME=${escapeShellArg config.xdg.cacheHome}
-          verboseEcho "Rebuilding bat theme cache"
-          cd "${pkgs.emptyDirectory}"
-          run ${lib.getExe cfg.package} cache --build
-        )
-      '';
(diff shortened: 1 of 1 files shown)


modules/programs/bat.nix
```diff
@@ -162,14 +162,15 @@ in {
...
-          export XDG_CACHE_HOME=${escapeShellArg config.xdg.cacheHome}
-          verboseEcho "Rebuilding bat theme cache"
-          cd "${pkgs.emptyDirectory}"
-          run ${lib.getExe cfg.package} cache --build
-        )
-      '';
+      home.activation.batCache = mkIf (cfg.syntaxes != { } || cfg.themes != { })
+        (hm.dag.entryAfter [ "linkGeneration" ] ''
+          (
+            export XDG_CACHE_HOME=${escapeShellArg config.xdg.cacheHome}
```
ambroisie, line 168: Drive-by suggestion: this should ideally use `lib.hm.shell.export`, which I just now noticed doesn't do any form of escaping, unlike `lib.strings.toShellVar`...

quatquatt (approved):
Looks great, would love it to be merged since this output has been annoying me for awhile.

rycee:
I think this is probably OK but I worry a bit about the case where you go from a configuration with syntaxes or themes set to a configuration with them both unset. Wouldn't that leave the, now stale syntax or theme, in the cache?

teto:
could you address comments (the last one especially) and if the code can avoid a stale cache rebase ?

dawidd6:
I no longer use home-manager, so I'm gonna close the PR and let others make the changes right.

PR #5492, opened 2024-06-03
yazi: Ensure plugin suffix `.yazi`

### LINKED ISSUE #5478: Automatically append `.yazi` to Yazi plugin names
Yazi plugins need to have the following structure: `$XDG_CONFIG_DIR/yazi/plugins/<plugin-name>.yazi/init.lua`.

It would be helpful if home-manager would automatically append the `.yazi` part to the plugin directory, so you can do this...

```nix
programs.yazi.plugins = {
    smart-enter = ./plugins/smart-enter;
};
```

...instead of this:

```nix
programs.yazi.plugins = {
    "smart-enter.yazi" = ./plugins/smart-enter;
};
```

Another thing that would nice to have is an assertion that the path given is a directory, and not a Lua file. The directory structure won't be correct unless it's the former. The documentation should reflect this as well.

lordkekz:
Fixes #5478 
I would also like to try backporting this PR and the new yazi plugin & flavor options to `release-24.05`. Should I send a PR for that?

### Description
- Append suffix `.yazi` to symlink targets of yazi plugins, if needed.
- Improve some docs, especially links to upstream docs.
- Update tests to make use of this feature.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@XYenon

### DIFF
--- a/modules/programs/yazi.nix
+++ b/modules/programs/yazi.nix
@@ -164,7 +164,7 @@ in {
       default = { };
       description = ''
         Pre-made themes.
-        Values should be a package or path containing an `init.lua` file.
+        Values should be a package or path containing the required files.
         Will be linked to {file}`$XDG_CONFIG_HOME/yazi/flavors/<name>.yazi`.
 
         See <https://yazi-rs.github.io/docs/flavors/overview/> for documentation.
@@ -228,26 +228,37 @@ in {
     ]);
 
     assertions = let
-      mkAsserts = opt:
+      mkAsserts = opt: requiredFiles:
         mapAttrsToList (name: value:
           let
             isDir = pathIsDirectory "${value}";
             msgNotDir = optionalString (!isDir)
               "The path or package should be a directory, not a single file.";
-            hasInitLua = pathExists "${value}/init.lua"
-              && !(pathIsDirectory "${value}/init.lua");
-            msgNoInitLua = optionalString (!hasInitLua)
-              "The path or package must contain a file `init.lua`.";
+            isFileMissing = file:
+              !(pathExists "${value}/${file}")
+              || pathIsDirectory "${value}/${file}";
+            missingFiles = filter isFileMissing requiredFiles;
+            msgFilesMissing = optionalString (missingFiles != [ ])
+              "The ${singularOpt} is missing these files: ${
+                toString missingFiles
+              }";
             singularOpt = removeSuffix "s" opt;
           in {
-            assertion = isDir && hasInitLua;
+            assertion = isDir && missingFiles == [ ];
             message = ''
               Value at `programs.yazi.${opt}.${name}` is not a valid yazi ${singularOpt}.
               ${msgNotDir}
-              ${msgNoInitLua}
+              ${msgFilesMissing}
               Evaluated value: `${value}`
             '';
           }) cfg.${opt};
-    in (mkAsserts "flavors") ++ (mkAsserts "plugins");
+    in (mkAsserts "flavors" [
+      "flavor.toml"
+      "tmtheme.xml"
+      "README.md"
+      "preview.png"
+      "LICENSE"
+      "LICENSE-tmtheme"
+    ]) ++ (mkAsserts "plugins" [ "init.lua" ]);
   };
 }
--- a/modules/programs/yazi.nix
+++ b/modules/programs/yazi.nix
@@ -209,12 +209,31 @@ in {
           else
             "yazi/plugins/${pluginName}.yazi";
 
-        mkPluginLink = pluginName: pluginPackage: {
+        mkPluginLink = pluginName: pluginPackageOrPath: {
           name = ensureSuffix pluginName;
-          value.source = pluginPackage;
+          value.source = pluginPackageOrPath;
         };
 
         pluginLinks = mapAttrs' mkPluginLink cfg.plugins;
       in pluginLinks);
+
+    assertions = (mapAttrsToList (pluginName: pluginPackageOrPath:
+      let
+        isDir = pathIsDirectory "${pluginPackageOrPath}";
+        hasInitLua = pathExists "${pluginPackageOrPath}/init.lua"
+          && !(pathIsDirectory "${pluginPackageOrPath}/init.lua");
+      in {
+        assertion = isDir && hasInitLua;
+        message =
+          "Value at `programs.yazi.plugins.${pluginName}` is not a valid yazi plugin."
+          + (optionalString (!isDir) ''
+
+            The path or package should be a directory, not a single file.'')
+          + (optionalString (!hasInitLua) ''
+
+            The path or package must contain a file `init.lua`.'') + ''
+
+              Evaluated value: `${pluginPackageOrPath}`'';
+      }) cfg.plugins);
   };
 }
--- a/modules/programs/yazi.nix
+++ b/modules/programs/yazi.nix
@@ -144,7 +144,9 @@ in {
       type = with types; attrsOf (oneOf [ path package ]);
       default = { };
       description = ''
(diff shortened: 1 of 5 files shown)


lordkekz:
Ah, shit, forgot the assertion about the link target being a folder.

lordkekz:
_Now_, it's done :)

modules/programs/yazi.nix
```diff
@@ -199,9 +199,41 @@ in {
...
-        cfg.flavors);
+      (name: value: nameValuePair "yazi/flavors/${name}" { source = value; })
+      cfg.flavors) // (let
+        # Make sure that the directory ends in `.yazi`, to comply with specification.
+        # `pluginName` is essential, it's needed to apply config in yazi's `init.lua`
+        ensureSuffix = pluginName:
+          if lib.hasSuffix ".yazi" pluginName then
+            "yazi/plugins/${pluginName}"
+          else
+            "yazi/plugins/${pluginName}.yazi";
```
sxyazi, line 210:
```suggestion
            "yazi/plugins/${pluginName}.yazi";
```

`foo.yazi` is the full package name, while `foo` is the plugin name.

It's best to only support the plugin name to stay consistent with Yazi's behavior. For example, use the `plugin` command as `plugin foo` instead of `plugin foo.yazi`. Yazi's package manager also uses `use = "owner/foo"` instead of `use = "owner/foo.yazi"`.

tests/modules/programs/yazi/settings.nix
```diff
@@ -70,7 +70,10 @@
       };
     };
     initLua = ./init.lua;
-    plugins = { "test.yazi" = ./plugin; };
+    plugins = {
+      "test.yazi" = ./plugin;
+      "anotherTest" = ./plugin;
```
sxyazi, line 75:
```suggestion
      "test" = ./plugin;
      "another-test" = ./plugin;
```

Yazi's plugin/package names should only include kebab-case characters, so it should be `another-test` instead of `anotherTest` here.

> Each plugin is a directory with a [kebab-case](https://developer.mozilla.org/en-US/docs/Glossary/Kebab_case) name, ending in .yazi

modules/programs/yazi.nix
lordkekz: I agree, but it's a breaking change, so I'd also like to add at least a warning if a pluginName ends in `.yazi`.

btw, thanks for your work on yazi ^^

tests/modules/programs/yazi/settings.nix
```diff
@@ -87,6 +90,8 @@
       ${./init.lua}
     assertFileContent home-files/.config/yazi/plugins/test.yazi/init.lua \
       ${./plugin/init.lua}
+    assertFileContent home-files/.config/yazi/plugins/anotherTest.yazi/init.lua \
```
XYenon, line 93:
```suggestion
    assertFileContent home-files/.config/yazi/plugins/another-test.yazi/init.lua \
```

XYenon (changes requested):
I think we can also apply the same rule to flavors?

poperigby:
Now that I think about it, perhaps home-manager could be smart and if you give it a single file, automatically put it into a directory?

lordkekz:
> I think we can also apply the same rule to flavors?

Yes, the config file also references flavors without the suffix, see [docs](https://yazi-rs.github.io/docs/flavors/overview#usage).

lordkekz:
> Now that I think about it, perhaps home-manager could be smart and if you give it a single file, automatically put it into a directory?

I'm not totally against it, but I wonder how @sxyazi feels about this. It might be a bad idea, e.g. if plugins could at some point require multiple files or sth.

I personally package plugins as folders, see [this repo](https://github.com/lordkekz/nix-yazi-plugins/) (very incomplete plugin collection, I might expand it to include at the ones from [yazi's overview](https://yazi-rs.github.io/docs/resources) when I have time).

lordkekz:
I applied the suggestions, except for the automatic directories (for now).

The assertions generated by the commented-out test cases look like this:
> Value at `programs.yazi.flavors.empty-dir-flavor` is not a valid yazi flavor.
> 
> The path or package must contain a file `init.lua`.
> Evaluated value: `/nix/store/fyjnhaxnpmd8wcvqlxipwyqrd9qqhggc-empty`
> 
> --
> Value at `programs.yazi.flavors.single-file-flavor` is not a valid yazi flavor.
> The path or package should be a directory, not a single file.
> The path or package must contain a file `init.lua`.
> Evaluated value: `/nix/store/ywqcihlnqpacpnmyjcxqk2qfnwy0fj7k-init.lua`
> 
> --
> Value at `programs.yazi.plugins.empty-dir-plugin` is not a valid yazi plugin.
> 
> The path or package must contain a file `init.lua`.
> Evaluated value: `/nix/store/fyjnhaxnpmd8wcvqlxipwyqrd9qqhggc-empty`
> 
> --
> Value at `programs.yazi.plugins.single-file-plugin` is not a valid yazi plugin.
> The path or package should be a directory, not a single file.
> The path or package must contain a file `init.lua`.
> Evaluated value: `/nix/store/16x7m6lvcyshl4miw4b9sggy7b1fg45p-init.lua`

And the warnings like this:
> Flavors like `programs.yazi.flavors."flavor-with-suffix.yazi"` should no longer have the suffix ".yazi" in their attribute name.
> The flavor will be linked to `$XDG_CONFIG_HOME/yazi/flavors/flavor-with-suffix.yazi.yazi`.
> You probably want to rename it to `programs.yazi.flavors."flavor-with-suffix"`.
> 
> --
> Plugins like `programs.yazi.plugins."plugin-with-suffix.yazi"` should no longer have the suffix ".yazi" in their attribute name.
> The plugin will be linked to `$XDG_CONFIG_HOME/yazi/plugins/plugin-with-suffix.yazi.yazi`.
> You probably want to rename it to `programs.yazi.plugins."plugin-with-suffix"`.

XYenon (approved):
LGTM

modules/programs/yazi.nix
```diff
@@ -161,8 +164,10 @@ in {
       default = { };
       description = ''
         Pre-made themes.
+        Values should be a package or path containing an `init.lua` file.
```
sxyazi, line 167: There is no `init.lua` file in flavors

lordkekz:
Fixed the asserts for flavors, so that they check for the files listed in the [docs](https://yazi-rs.github.io/docs/flavors/overview/#structure).
They look like this now:
> Value at `programs.yazi.flavors.empty-dir-flavor` is not a valid yazi flavor.
> 
> The flavor is missing these files: flavor.toml tmtheme.xml README.md preview.png LICENSE LICENSE-tmtheme
> Evaluated value: `/nix/store/fyjnhaxnpmd8wcvqlxipwyqrd9qqhggc-empty`
> 
> --
> Value at `programs.yazi.flavors.single-file-flavor` is not a valid yazi flavor.
> The path or package should be a directory, not a single file.
> The flavor is missing these files: flavor.toml tmtheme.xml README.md preview.png LICENSE LICENSE-tmtheme
> Evaluated value: `/nix/store/wc2assmqd02x9x4rvs4jdg7a7n1fb532-flavor.toml`
> 
> --
> Value at `programs.yazi.plugins.empty-dir-plugin` is not a valid yazi plugin.
> 
> The plugin is missing these files: init.lua
> Evaluated value: `/nix/store/fyjnhaxnpmd8wcvqlxipwyqrd9qqhggc-empty`
> 
> --
> Value at `programs.yazi.plugins.single-file-plugin` is not a valid yazi plugin.
> The path or package should be a directory, not a single file.
> The plugin is missing these files: init.lua
> Evaluated value: `/nix/store/16x7m6lvcyshl4miw4b9sggy7b1fg45p-init.lua`

XYenon (approved):

rycee:

PR #5500, opened 2024-06-04
nushell: plugin support

JoaquinTrinanes:
### Description

This PR adds plugin support via `programs.nushell.plugins`.

To achieve it, a barebones nushell instance is launched and then the plugins are registered. The updated plugin file is then linked to the default location.

The plugin file and commands needed to register them differs between nushell versions, so that's handled accordingly. For reference, the new `plugin add` commands work for nushell `>=0.93.0`, the version on `nixpkgs-24.05`. `nixpkgs-23.11` has nushell version `0.87.0`.

I still there are some questions to answer before this becomes mergeable:
- Does it makes sense to support older nushell versions? Mixing an older nixpkgs stable release with home-manager unstable is AFAIK not supported
- As per [the docs](https://www.nushell.sh/book/plugins.html#downloading-and-installing-a-plugin) the plugin installation workflow is subject to changes. Is it worth it to allow this feature now, only to break in the future?
    - Maybe we can keep the interface consistent, but update the implementation. It would require more checks for specific version ranges tho 🥲 
 - It's also possible to [configure plugins](https://www.nushell.sh/book/plugins.html#downloading-and-installing-a-plugin). So instead of a list maybe an attrset with fields like `enabled`, `package`, `settings`... might be a better option

Happy to hear/read your feedback!

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@Philipp-M

### DIFF
--- a/modules/programs/nushell.nix
+++ b/modules/programs/nushell.nix
@@ -39,7 +39,7 @@ let
       };
     });
 in {
-  meta.maintainers = [ maintainers.Philipp-M ];
+  meta.maintainers = with maintainers; [ Philipp-M joaquintrinanes ];
 
   imports = [
     (mkRemovedOptionModule [ "programs" "nushell" "settings" ] ''
@@ -152,6 +152,13 @@ in {
         An attribute set that maps an environment variable to a shell interpreted string.
       '';
     };
+
+    plugins = mkOption {
+      type = types.listOf types.package;
+      default = [ ];
+      example = lib.literalExpression "[ pkgs.nushellPlugins.formats ]";
+      description = "List of nushell plugins ";
+    };
   };
 
   config = mkIf cfg.enable {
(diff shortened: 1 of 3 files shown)


Philipp-M:
> Does it makes sense to support older nushell versions? Mixing an older nixpkgs stable release with home-manager unstable is AFAIK not supported

I mean I'm not using nushell, but I think it's fine, when the relevant branches are only compatible with their respective upstream `nixpkgs` branches (though I think this master branch should be compatible with `nixpkgs/nixos-unstable`). Having to support all kinds of old versions where a lot of breaking changes can occur is cumbersome and I don't think worth the trouble.

> As per [the docs](https://www.nushell.sh/book/plugins.html#downloading-and-installing-a-plugin) the plugin installation workflow is subject to changes. Is it worth it to allow this feature now, only to break in the future?

As you've added yourself to the maintainers list (which I'm happy with, because as noted I'm not effectively maintaining it, since I'm not using nushell currently), if you want to adjust for the coming changes, go for it.

> It's also possible to [configure plugins](https://www.nushell.sh/book/plugins.html#downloading-and-installing-a-plugin). So instead of a list maybe an attrset with fields like enabled, package, settings... might be a better option

You may have more insight in this, when it's super structured, requires few extra code, *maybe* worth it

JoaquinTrinanes:
I removed support for older versions (harder to maintain and it would require someone using unstable HM with nushell from a stable/older nixpkgs.

I also kept it simple as a list of plugins. We don't do any kind of nu code generation right now and it's perfectly doable in `extraConfig`, which is what we'd be doing in the background anyway, so a list of derivations is the simplest way of handling this right now. And I'm not a heavy plugin user so it will lead to less unnoticed breakage 😌 If #3831 gets merged it would be something interesting to explore. 

If we want more options other than the plugin package in the future we should be able to detect it. For example, If `plugins` is a list, show a warning and instruct to migrate to an attrset, and transform the option to the new format internally.

Thanks a lot for the input @Philipp-M ! 😄

nyabinary:
Is this ready to merge? Some tests failed :P

JoaquinTrinanes:
Superseded by #6176

PR #5524, opened 2024-06-10
mpv: fix package if wrapMpv not found

matt1432:
### Description

`wrapMpv` was removed from nixpkgs since this PR: https://github.com/NixOS/nixpkgs/pull/304349

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/mpv.nix
+++ b/modules/programs/mpv.nix
@@ -55,8 +55,13 @@ let
 
   mpvPackage = if cfg.scripts == [ ] then
     cfg.package
+  else if hasAttr "wrapMpv" pkgs then
+    pkgs.wrapMpv pkgs.mpv-unwrapped { scripts = cfg.scripts; }
   else
-    pkgs.wrapMpv pkgs.mpv-unwrapped { scripts = cfg.scripts; };
+    pkgs.mpv-unwrapped.wrapper {
+      mpv = pkgs.mpv-unwrapped;
+      scripts = cfg.scripts;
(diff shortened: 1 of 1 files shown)


modules/programs/mpv.nix
```diff
@@ -55,8 +55,13 @@ let
...
   mpvPackage = if cfg.scripts == [ ] then
     cfg.package
+  else if hasAttr "wrapMpv" pkgs then
+    pkgs.wrapMpv pkgs.mpv-unwrapped { scripts = cfg.scripts; }
   else
-    pkgs.wrapMpv pkgs.mpv-unwrapped { scripts = cfg.scripts; };
+    pkgs.mpv-unwrapped.wrapper {
+      mpv = pkgs.mpv-unwrapped;
+      scripts = cfg.scripts;
+    };
```
marsam, line 64: I believe this can be:
```suggestion
    pkgs.mpv.override { scripts = cfg.scripts; };
```
Besides that, LGTM. Thanks!

matt1432: From what I could tell, `mpv-unwrapped.wrapper` replaces `wrapMpv` which seems to be necessary.

 I didn't really look into it that much, but if a simple override works, why isn't that the first thing we do here?

marsam: I'm not sure why, they yield the same derivation

rycee:

PR #5535, opened 2024-06-13
mu: add support for setting muhome

### LINKED ISSUE #5534: bug: MUHOME not respected by mu activation

### Issue description

The database path can be set by the user, this is useful for keeping the Maildir and the Mu Xapian cache together without affecting `XDG_CACHE_HOME`. It can be set in a couple methods:
> The default location for that is ~/.cache/mu/xapian, but you can change this using the --muhome option, and remember to pass that to the other commands as well. Alternatively, you can use an environment variable MUHOME.

Here the db location is set to `XDG_CACHE_HOME/mu`:
https://github.com/nix-community/home-manager/blob/master/modules/programs/mu.nix#L56

### Maintainer CC

@KarlJo

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.6.32, NixOS, 24.05 (Uakari), 24.05.20240612.cc54fb4`
 - multi-user?: `yes`
 - sandbox: `yes`
 - nixpkgs: `/nix/store/aa0zsahvyqnvzkicsh29jirm9hwv95pg-source`
```

jdek:
### Description

The database path can be set by the user, this is useful for keeping
the Maildir and the Mu Xapian cache together without affecting
XDG_CACHE_HOME. Fixes #5534.

Note that this is untested apart from the test suite. I'm not entirely sure how to easily test a home-manager change when it's a module within my NixOS configuration temporarily.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@Karljo

### DIFF
--- a/modules/programs/mu.nix
+++ b/modules/programs/mu.nix
@@ -31,6 +31,15 @@ in {
 
       package = mkPackageOption pkgs "mu" { };
 
+      home = mkOption {
+        type = with types; nullOr str;
+        default = null;
+        example = "\${config.home.homeDirectory}/Maildir/.mu";
+        description = ''
+          Directory to store Mu's database.
+        '';
+      };
+
       # No options/config file present for mu, and program author will not be
       # adding one soon. See https://github.com/djcb/mu/issues/882 for more
       # information about this.
@@ -46,9 +55,13 @@ in {
   config = mkIf cfg.enable {
     home.packages = [ cfg.package ];
 
+    home.sessionVariables = filterAttrs (n: v: v != null) {
+      MUHOME = cfg.home;
+    };
+
     home.activation.runMuInit = let
       maildirOption = genCmdMaildir config.accounts.email.maildirBasePath;
-      dbLocation = config.xdg.cacheHome + "/mu";
+      dbLocation = if cfg.home != null then cfg.home else (config.xdg.cacheHome + "/mu");
     in hm.dag.entryAfter [ "writeBoundary" ] ''
       # If the database directory exists, then `mu init` should NOT be run.
       # In theory, mu is the only thing that creates that directory, and it is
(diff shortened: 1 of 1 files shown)


modules/programs/mu.nix
```diff
@@ -31,6 +31,15 @@ in {
 
       package = mkPackageOption pkgs "mu" { };
 
+      home = mkOption {
+        type = with types; nullOr str;
+        default = null;
```
rycee, line 36: Maybe the default should be `"${config.xdg.cacheHome}/mu"` and `--muhome` always used? Note, if we do this then also need to add `defaultText` to avoid making documentation depend on the configuration.

rycee: Since the code already uses `dbLocation = config.xdg.cacheHome + "/mu"` we effectively already specify a default value. So I would suggest using something like

``` nix
type = types.path;
default = config.xdg.cacheHome + "/mu";
defaultText = literalExpression ''config.xdg.cacheHome + "/mu"'';
```

in the option

and `--muhome ${escapeShellArg cfg.home}` in the command.

I assume it is unlikely that mu changes the default directory again, if they do then we'll have to guard it behind a `stateVersion` to allow a sensible upgrade path.

modules/programs/mu.nix
```diff
@@ -46,17 +55,21 @@ in {
...
+
     home.activation.runMuInit = let
       maildirOption = genCmdMaildir config.accounts.email.maildirBasePath;
-      dbLocation = config.xdg.cacheHome + "/mu";
+      dbLocation = if cfg.home != null then cfg.home else (config.xdg.cacheHome + "/mu");
     in hm.dag.entryAfter [ "writeBoundary" ] ''
       # If the database directory exists, then `mu init` should NOT be run.
       # In theory, mu is the only thing that creates that directory, and it is
       # only created during the initial index.
       if [[ ! -d "${dbLocation}" ]]; then
```
rycee, line 69: Can take the opportunity to use `escapeShellArg` here as well.

PR #5543, opened 2024-06-16
systemd: Include `home.sessionVariables` in default env

nicoonoclaste:
### Description

Default [`systemd.user.settings.Manager.DefaultEnvironment`][default-env] to `home.sessionVariables`.

[default-env]: https://www.freedesktop.org/software/systemd/man/latest/systemd-user.conf.html#DefaultEnvironment=

### Checklist

  `lib.mkDefault` is used to ensure users' settings take priority, so configs that set `DefaultEnvironment.*` should still eval.
- [x] Code tested through `nix-shell --pure tests -A run.$t` for the following tests:
  - `systemd-{empty-,}user-config`
  - `systemd-services`
  - `systemd-session-variables`

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -1703,6 +1703,21 @@ in {
           one place. See https://github.com/glanceapp/glance for more.
         '';
       }
+
+      {
+        time = "2024-08-19T16:57:24Z";
+        condition = hostPlatform.isLinux;
+        message = ''
+          `home.sessionVariables` is now added to the environment used by `systemd` units.
+
+          `systemd` does not support the whole [POSIX parameter expansion] syntax,
+          but only `${FOO}`, `${FOO:-DEFAULT_VALUE}` and `${FOO:+ALTERNATE_VALUE}`;
+          in particular, substring processing (such as removing prefixes and suffices)
+          is not supported.  See `environment.d(5)`.
+
+          [POSIX parameter expansion]: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02
+        '';
+      }
     ];
   };
 }


ambroisie (approved):
Looks reasonable IMO.

Adding a news item might be good, if only to alert users that use syntax which is not supported by systemd in their configuration (e.g: `$(...)` command substitution) that something funky might happen.

EDIT: to be clear, the manual states one should use "POSIX.2 variable references" (which I interpret as anything mentionned by section 2.6.2 in the [shell command language](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html) spec). The _user_ would be in fault for breaking the contract, but [Hyrum's law](https://www.hyrumslaw.com/) is a thing...

EDIT2: just re-read `man environment.d` and found out that most of the variable reference syntax is not actually supported, just the `:-` and `:+` substitution (i.e: no prefix/suffix removal).

modules/systemd.nix
```diff
@@ -314,6 +314,11 @@ in {
       settings
     ];
 
+    # Make `home.sessionVariables` part of the default environment
+    # using `mkDefault` so users' settings take priority.
+    systemd.user.settings.Manager.DefaultEnvironment =
```
kirasok, line 319: A type of [home.sessionVariables](https://nix-community.github.io/home-manager/options.xhtml#opt-home.sessionVariables) is `lazy attribute set of (string or absolute path or signed integer or floating point number)`

While type of [systemd.user.settings.Manager.DefaultEnvironment](https://nix-community.github.io/home-manager/options.xhtml#opt-systemd.user.settings.Manager.DefaultEnvironment) is `attribute set of (null or string or absolute path or package)`

systemd option type is more restrictive. It cannot be assigned directly. It should either be mapped with `builtins.toString` or [function that generates DefaultEnvironment option](https://github.com/nix-community/home-manager/blob/83053e1d337f33e0b48250588006e4b9df2f0d9d/modules/systemd.nix#L377) changed to accept integers and floating point numbers

kirasok: Also, I noticed that option DefaultEnvironment doesn't expand nested environment variables. So, if such env var would be present, like `SSH_AUTH_SOCK`

```nix
systemd.user.settings.Manager.DefaultEnvironment.SSH_AUTH_SOCK =
  "$XDG_RUNTIME_DIR/${config.services.ssh-agent.socket}";
```

Then user service with `ExecStart=echo $SSH_AUTH_SOCK $XDG_RUNTIME_DIR` would output

```
echo[15730]: $XDG_RUNTIME_DIR/ssh-agent /run/user/1000
```

I found that there are [specifiers](https://www.freedesktop.org/software/systemd/man/258/systemd-system.conf.html#Specifiers) which could be used

```nix
systemd.user.settings.Manager.DefaultEnvironment.SSH_AUTH_SOCK =
  "/run/user/%U/${config.services.ssh-agent.socket}";
```

```
echo[22251]: /run/user/1000/ssh-agent /run/user/1000
```

But this makes definitions for these options non-interchangeable. So, I don't think this PR should be merged. Each module that needs env var in some systemd service should define it itself.

PR #5416, opened 2024-05-19
jujutsu: support darwin guidelines for config placement

mrnossiom:
Follow up to #5207, fixing jujutsu module on darwin targets.

Closes #5544 (duplicate)

### Description

If the module is evaluated on a Darwin platform, place config in `Library/Application Support` instead of unsupported XDG directory.

If someone could test it on Darwin, it would be great (e.g. @bnjmnt4n) :

```sh
nix develop --ignore-environment .#jujutsu-example-config
nix develop --ignore-environment .#jujutsu-empty-config
```

### Checklist

- [x] Change is backwards compatible.

- [x] Code formatted with `./format`.

- [x] Code tested through `nix-shell --pure tests -A run.all` or `nix develop --ignore-environment .#all` using Flakes.

- [x] Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).

- [x] Commit messages are formatted like

    ```
    {component}: {description}

    {long description}
    ```

    See [CONTRIBUTING](https://nix-community.github.io/home-manager/#sec-commit-style) for more information and [recent commit messages](https://github.com/nix-community/home-manager/commits/master) for examples.

#### Maintainer CC

@shikanime

PR #5544, opened 2024-06-16
jujutsu: fix config file location on Darwin

Aehmlo:
### Description

The `jj` tool loads its configuration from [a platform-dependent location](https://github.com/martinvonz/jj/blob/a9953b3fb0713b792541930bf880f6e0d93ede91/docs/config.md#user-config-file). On Darwin, it looks in `~/Library/Application Support/jj`. This revision updates the `programs.jujutsu` module to serialize the generated configuration to the correct location.

### Checklist

Test system information:

```bash session
 - system: `"aarch64-darwin"`
 - host os: `Darwin 22.6.0, macOS 13.6.7`
 - multi-user?: `yes`
 - sandbox: `no`
 - channels(user): `""`
 - channels(root): `"nixpkgs-24.05"`
```

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@shikanime

### DIFF
--- a/modules/programs/jujutsu.nix
+++ b/modules/programs/jujutsu.nix
@@ -6,6 +6,10 @@ let
 
   cfg = config.programs.jujutsu;
   tomlFormat = pkgs.formats.toml { };
+  configDir = if pkgs.stdenv.isDarwin then
+    "Library/Application Support"
+  else
+    config.xdg.configHome;
 
 in {
   meta.maintainers = [ maintainers.shikanime ];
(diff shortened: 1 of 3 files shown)


huwaireb:
Looks like a duplicate of #5416.
cc @mrnossiom

p.s jj through `programs.jujutsu` on Darwin has been using the incorrect(linux-only) config path since #5207 (~3 months ago) so it would be really great if we can merge either one of the PRs.

mrnossiom:
It seems to be indeed

PR #5554, opened 2024-06-19
qcal: configurable package

corite:
qcal: introduce a `package` option
### Description

Introduce a `package` option for qcal.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@antonmosich

### DIFF
--- a/home-manager/po/vi.po
+++ b/home-manager/po/vi.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: Home Manager\n"
 "Report-Msgid-Bugs-To: https://github.com/nix-community/home-manager/issues\n"
 "POT-Creation-Date: 2024-04-17 23:19+0200\n"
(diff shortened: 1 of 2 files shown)


lafrenierejm (approved):

antonmosich:
This certainly works, and I'm not against it being merged. Just wanted to clarify why I didn't add this option right away: The NixOS [manual](https://nixos.org/manual/nixos/stable/#sec-option-declarations-util-mkPackageOption) section for writing modules recommends not creating a package option for all modules, as they could just be overwritten with overlays.

corite:
Oh ok, I did not know this was not recommended. I just checked the home-manager docs, there they recommend using the package option over overlays if one exists (I'm guessing mostly due to it being simpler to use). Honestly I only made this package option because I couldn't get my overlay to work, and just introducing this option was easier for me :D.

PR #5556, opened 2024-06-19
ssh-agent: use POSIX-conform test for session variables

mgdbbrt:
### Description

According to the documentation, hm-session-vars.sh may be sourced by POSIX compatible shells. Therefore, home.sessionVariablesExtra should only include code that conforms to POSIX.

### Checklist


rycee:

PR #5563, opened 2024-06-21
gtk: explicitly set default font size

### LINKED ISSUE #5562: bug: GNOME Shell 46 requires `gtk.font.size` to be set when `gtk.font.name` is set

### Issue description

After upgrading to NixOS 24.05, I noticed that my fonts started breaking on the machines I've setup with gnome. I tracked down the problem to my gtk config - which doesn't explicitly set a size for the font:

#### Without explicit size:

image

#### With `gtk.font.size = 10`:

image

It seems like the simplest fix would just be to just default to 10 (the default in gnome shell). Though I'm not sure how that would impact non-gnome systems. I haven't noticed any difference when using gtk apps in sway.

It looks like the original implementation of the module required a font size, but that was changed in https://github.com/nix-community/home-manager/pull/1848.

### Maintainer CC

@rycee @terlar

### System information

```markdown
- system `"x86_64-linux"`
- host os `Linux 6.6.33, NixOS, 24.05 (Uakari), 24.05.20240620.83cc467`
- multi-user?: `yes`
- sandbox: `yes`
- version `nix-env (Nix) 2.18.2`
- nixpkgs `/nix/store/d23d49vx6ifpppxp1b87aiicn2hx9l6r-source`
```

kira-bruneau:
### Description

Fixes #5562

I originally thought about changing the default of `hm.types.fontType` to 10, but just decided to do a null check in the gtk module since the new default wouldn't apply outside of gtk.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/misc/gtk.nix
+++ b/modules/misc/gtk.nix
@@ -221,10 +221,9 @@ in {
 
   config = mkIf cfg.enable (let
     gtkIni = optionalAttrs (cfg.font != null) {
-      gtk-font-name = let
-        fontSize =
-          optionalString (cfg.font.size != null) " ${toString cfg.font.size}";
-      in "${cfg.font.name}" + fontSize;
+      gtk-font-name =
+        let fontSize = if cfg.font.size != null then cfg.font.size else 10;
+        in "${cfg.font.name} ${toString fontSize}";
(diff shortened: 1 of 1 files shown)


rycee:

PR #5564, opened 2024-06-21
htop: use attrsOf type instead of attrs as the settings type

Luflosi:
### Description
Read https://github.com/NixOS/nixpkgs/issues/85508 for why it's a bad idea to use the attrs type.
I noticed this problem while trying to use `mkDefault` for one of the options, which did not work and instead failed with the unhelpful error message `error: cannot coerce a set to a string`.

The new type is a little more strict than the old type. I could have probably used something like `attrsOf anything` but making the type a little more explicit seems like a good idea. Let me know if 100% backwards compatibility is more important.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@bjpbakker @somasis

### DIFF
--- a/modules/programs/htop.nix
+++ b/modules/programs/htop.nix
@@ -113,7 +113,8 @@ in {
     enable = mkEnableOption "htop";
 
     settings = mkOption {
-      type = types.attrs;
+      type = with types;
+        attrsOf (oneOf [ bool int str (listOf (oneOf [ int str ])) ]);
       default = { };
       example = literalExpression ''
         {


bjpbakker:
Nice, thanks for the improvement :)

bjpbakker:
@rycee could you merge this please?

bjpbakker:
> Maintaining my own home-manager fork is getting really tiresome...

I understand and I agree. As maintainer I can review but not merge these changes. There's 3 PRs waiting for being merged for a long time now.

@rycee if you still want this htop module in home-manage, please take a moment to merge these changed. If not, we should consider other means to manage htop configurations.

rycee:

Luflosi:
Awesome, thanks!

name-snrl:
@Luflosi That's cool, but for me it broke the overriding part of the config using `lib.mkAfter`.

here is a definition in the common module:

https://github.com/name-snrl/nixos-configuration/blob/master/modules/home-manager/profiles/common/htop.nix

and I override this for a specific `homeConfiguration`:

https://github.com/name-snrl/nixos-configuration/blob/df805810b6690f3300a7da536a3343087d86f40a/modules/home-manager/configurations/yusup/default.nix#L26-L29

Since this has been merged, my solution doesn't work:

```
fish 2s > nix build --print-out-paths --no-link .#homeConfigurations.yusup.activationPackage
error:
       … while evaluating 'strict' to select 'drvPath' on it
         at /builtin/derivation.nix:1:552:
         at /builtin/derivation.nix:1:208:

       error: The option `programs.htop.settings.left_meters' is defined multiple times while it's expected to be unique.

       Definition values:
       - In `/nix/store/aj0jvcfb50ips7h5sgd74z8wscggl8m4-source/modules/home-manager/profiles/common/htop.nix':
           [
             "LeftCPUs"
             "Blank"
             "DateTime"
             "Uptime"
           ...
       - In `/nix/store/aj0jvcfb50ips7h5sgd74z8wscggl8m4-source/modules/home-manager/configurations/yusup/default.nix': "LeftCPUs4 Blank DateTime Uptime LoadAverage Tasks Blank Swap Memory"
       Use `lib.mkForce value` or `lib.mkDefault value` to change the priority on any of these definitions.
```

name-snrl:
> let me know

I think this is a bit simpler

https://github.com/name-snrl/nixos-configuration/commit/667495a34bf141db73fcf1029fc4b016987ba409#diff-41489789006ed24b44970ea370c0dd891d3aed42b0e06c4c2fb39ccfd9c226ef

PR #5566, opened 2024-06-22
herbstluftwm: Make herbstclient alias optional

### LINKED ISSUE #5213: herbstluftwm extraconfig dose not support bash functions

### Issue description

Hi, i am in the process of porting my existing herbstlfutwm config to home-manager/nixos using the home-manager herbstluftwm module.
currently the function for generating the final autostart bash script dose not allow for bash functions in the extraconfig string.
my question is, why did you implement it in this way?
this part at the beginning:
```
shopt -s expand_aliases

# shellcheck disable=SC2142
alias herbstclient='set -- "$@" ";"'
set -- 
```
and this part at the end:
```
${cfg.package}/bin/herbstclient chain ";" "$@"
```
don't allow bash scripting, removing them restores the expected behaviour of an normal hebstluftwm autoconfig file.

to me these parts seem redundant since the autoconfig file it self is already been executed.

i know that as mentioned in #2884, you can port the bash function into the let ... in section of the nix config, but since i use some of the functions in an second bash script for monitor handling, i dont want to maintain two versions of the same code.

my suggestion would be to add an option to disable this behaviour by omitting the above mentioned code blocks from the config generator.

best
tincotema

### Maintainer CC

@olmokramer

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.7.10, NixOS, 24.05 (Uakari), 24.05pre-git`
 - multi-user?: `no`
 - sandbox: `yes`
 - nixpkgs: `/nix/store/7ay2dd0w41iqvrnkx19v4vf4f5kkxvc2-source`
```

olmokramer:
### Description

The `herbstclient` alias made it impossible to use bash functions in the autostart script for `herbstluftwm`, as described in #5213.

This makes the `herbstclient` alias optional by adding an extra `herbstclientAlias` option on the herbstluftwm configuration. The new option defaults to `false` as to not confuse newcomers to the herbstluftwm module, which is not a breaking change because it was only an optimization.

Fixes #5213

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


olmokramer:
The test failures on macos don't seem to be related to this PR, these changes should only affect Linux.

olmokramer:
Still relevant.

modules/services/window-managers/herbstluftwm.nix
```diff
@@ -119,6 +119,22 @@ in
         {file}`$XDG_CONFIG_HOME/herbstluftwm/autostart`.
       '';
     };
+
+    herbstclientAlias = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
```
teto: set it to true for backwards compatibility.

teto:
willing to merge if tests pass.

modules/services/window-managers/herbstluftwm.nix
khaneliman:
```suggestion
    enableAlias = lib.mkOption {
      type = lib.types.bool;
      default = true;
```

Already in the module so it feels redundant and it follows convention used elsewhere.

olmokramer:
Thanks for the reviews! I've made the requested changes and added a news entry.

PR #5568, opened 2024-06-22
vscode: add support for codeoss-cloudworkstations

terlar:
### Description

This adds support for configuring the editor that comes with Google Cloud Workstations. On these machines vscode is preinstalled and configured to read from `~/.codeoss-cloudworkstations`.

In order to get this working I had to add:
```nix
{ pkgs, ... }:

{
  programs.vscode = {
    enable = true;
    package = (pkgs.writeShellScriptBin "code" ''
      exec /opt/code-oss/bin/remote-cli/code-oss-cloud-workstations "$@"
    '') // {
      inherit (pkgs.vscodium) pname version;
    };

    extensionsPath = ".codeoss-cloudworkstations/extensions";
    userDir = ".codeoss-cloudworkstations/data/Machine";
  };
}
```

I guess another option is to expose something else specifically for Cloud Workstation that will configure these values. At lease exposing the `extensionsPath` and `userDir` gives more flexibility if there happen to be other variants that utilize different places.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/vscode.nix
+++ b/modules/programs/vscode.nix
@@ -25,19 +25,11 @@ let
     "openvscode-server" = "openvscode-server";
   }.${vscodePname};
 
-  userDir = if pkgs.stdenv.hostPlatform.isDarwin then
-    "Library/Application Support/${configDir}/User"
-  else
-    "${config.xdg.configHome}/${configDir}/User";
+  configFilePath = "${cfg.userDir}/settings.json";
+  tasksFilePath = "${cfg.userDir}/tasks.json";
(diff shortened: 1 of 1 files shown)


modules/programs/vscode.nix
```diff
@@ -223,6 +230,12 @@ in {
   };
 
   config = mkIf cfg.enable {
+    programs.vscode.userDir = lib.mkDefault
```
terlar, line 233: I put the default here, due to a default for the option will cause early evaluation, I got issues where it complained about `home.stateVersion` not being set, while it was set in a later part of the module configuration.

PR #5571, opened 2024-06-23
yazi: rename shell wrappers from `ya` to `yy`

eljamm:
### Description

The program shell wrappers have been renamed from `ya` to `yy` since the
introduction of the new CLI tool, which took the name of `ya`.

See https://github.com/sxyazi/yazi/issues/914

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@XYenon


rycee:
@eljamm Thanks! About the test failure, could you rebase this PR on the latest master? Should fix at least the Linux tests.

The change itself looks good to me but I'll let @XYenon have the last say.

modules/misc/news.nix
```diff
@@ -1679,6 +1679,14 @@ in {
           https://github.com/rafaelmardojai/blanket for more.
         '';
       }
+
+      {
+        time = "2024-06-23T14:34:28+00:00";
+        condition = config.programs.yazi.enable;
```
XYenon, line 1685: Only show this when shell integration is enabled.

```suggestion
        condition = with config.programs.yazi; enable && (enableBashIntegration || enableZshIntegration || enableFishIntegration || enableNushellIntegration);
```

XYenon (changes requested):

modules/misc/news.nix
eljamm: Nice catch!

XYenon (approved):

sxyazi:
This might be a bit off-topic, but is it possible to allow users to set their preferred shell wrapper names? There are two benefits:

- Some users switching from Ranger/lf to Yazi might be more accustomed to their previous names (like [`ra`](https://github.com/search?q=ra%3D%22yazi%22&type=code), [`lf`](https://github.com/search?q=lf%3D%22yazi%22&type=code) etc.).
- It would give users the ability to further customize it, such as using `y` instead of `yy` as it's shorter, or using other characters instead of `yy` to resolve conflicts.

sxyazi:
> That said, users can still use [shell aliases](https://home-manager-options.extranix.com/?query=shellAlias&release=master) to do the same thing, so I'm not certain if we should add this option.

The shell wrapper and an alias are different, and the former is a function that provides additional functionalities (side effects) - switching directories on exit. Some users may want this behavior, while others may not, so they can choose either an alias or a shell wrapper, but being able to change their names should be equally important.

I suggest using `cfg.wrapperName` or `cfg.shellWrapperName` instead of `cfg.shellAlias` as the option name because it is essentially not an alias.

eljamm:
You're right, I see what you mean, now. However, users can still choose to alias the wrapper, so something like `alias ra=yy` should also work, right?

But it might also be worth it to provide the option to change the wrapper name from the module as well.

modules/programs/yazi.nix
```diff
@@ -40,13 +40,22 @@ let
...
     enable = mkEnableOption "yazi";
 
     package = mkPackageOption pkgs "yazi" { };
 
+    shellWrapperName = mkOption {
+      type = types.str;
+      default = "yy";
+      example = "ra";
+      description = ''
+        Yazi alias for the shell integration wrappers.
```
sxyazi, line 55:
```suggestion
      example = "y";
      description = ''
        Name of the shell wrapper to be called.
```

sxyazi:
> However, users can still choose to alias the wrapper, so something like alias ra=yy should also work, right?

Yup that's right. It just creates a new alias for a function, and once we allow changing the shell wrapper's name, this won't be necessary anymore.

modules/programs/yazi.nix
eljamm: I forgot to change this, thanks!

sxyazi (approved):
LGTM!

XYenon (approved):
LGTM too!

rycee:

Issue #5552, opened 2024-06-19
bug: home-manager Failed to build switch plan after recent update 


sakshatshinde:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

home-manager switch worked fine until my last switch. I referred the nix-starter config. Don’t believe I have changed any paths.

`home-manager switch --flake .` stopped working giving the following error

```sh
[sakshat@nixos:~/nix-config]$ home-manager switch --flake .
warning: Git tree '/home/sakshat/nix-config' is dirty
warning: Git tree '/home/sakshat/nix-config' is dirty
warning: Git tree '/home/sakshat/nix-config' is dirty
warning: Git tree '/home/sakshat/nix-config' is dirty
Starting Home Manager activation
Activating checkFilesChanged
Activating checkLinkTargets
Activating writeBoundary
Activating installPackages
replacing old 'home-manager-path'
installing 'home-manager-path'
Activating linkGeneration
Cleaning up orphan links from /home/sakshat
No change so reusing latest profile generation 6
Creating home file links in /home/sakshat
Activating onFilesChange
Activating reloadSystemd
Error: Error switching

Caused by:
    0: Failed to build switch plan
    1: No such file or directory (os error 2)```

### Maintainer CC

_No response_

### System information

```markdown
nix-shell -p nix-info --run "nix-info -m"
 - system: `"x86_64-linux"`
 - host os: `Linux 6.6.33, NixOS, 24.11 (Vicuna), 24.11.20240616.b60ebf5`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.18.2`
 - nixpkgs: `/nix/store/5w7nhyvrgsa81h0kyifmbl6j97hvlzh5-source`
```

radek-stasta:
Also here: https://discourse.nixos.org/t/home-manager-failing-after-recent-update/47211

yswtrue:
same here

radek-stasta:
Can we get any info about this or if there is something we can do to fix it please? Thanks a lot

sakshatshinde:
> Can we get any info about this or if there is something we can do to fix it please? Thanks a lot

Following commits seem to work.

Home-manager: `a7117efb3725e6197dd95424136f79147aa35e5b`
[...]

PR #5573, opened 2024-06-24
systemd: fix sd-switch failure when no user units are managed by HM

kkoniuszy:
After update to sd-switch 0.5.0, `home-manager switch` failed with the following error message:

> Error: Error switching
>
> Caused by:
>     0: Failed to build switch plan
>     1: No such file or directory (os error 2)

This was caused by a nonexisting target unit directory. This change adds a .hm-keep file there, so that sd-switch does not fail when HM didn't create any user units.

Closes: https://github.com/nix-community/home-manager/issues/5552

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


clhodapp:
Wouldn't it be cleaner to have an empty directory, rather than putting a hidden `.hm-keep` in there?

That kind of stuff often confuses people down the road. For example, maybe someday a future version of systemd freaks out if there are things that don't look like unit files in its directory.

kkoniuszy:
@clhodapp It's not possible to create an empty *non-symlinked* directory neither with `xdg.configFile` nor with `home.file`. These mechanisms are supposed to symlink files or directories while creating the parent dirs as needed. They don't allow creating an arbitrary empty dir, which wouldn't make much sense anyways as an empty directory couldn't be tied to the nix store at all (unless it was a symlink).

Symlinking an empty directory would not work here because we actually need `.config/systemd/user` to be writable. The only way to get an empty-ish directory that isn't a link is to make an empty `.keep` file. I don't think it would cause any problems, it's a standard way of dealing with tools that think in terms of file paths rather than treating directories and files as separate – for instance, you cannot store an empty directory in a `git` repo and the common workaround is to make an empty file which is conventionally named `gitkeep`.

Besides, it would be really weird if some random dotfiles with the wrong extension caused systemd to freak out, as it's really common to have random vim .swp files lying around. There's no way it doesn't filter them by supported extensions.

There are a some instances of .keep files being used by home-manager:
```
modules/programs/zsh.nix
718:      home.file."${config.xdg.cacheHome}/oh-my-zsh/.keep".text = "";

modules/programs/firefox.nix
33:      (pkgs.writeTextDir "lib/mozilla/native-messaging-hosts/.keep" "")
765:      "${profilesPath}/${profile.path}/.keep".text = "";

modules/misc/xdg.nix
125:        { "${cfg.cacheHome}/.keep".text = ""; }

modules/misc/dconf.nix
80:    xdg.configFile."dconf/.keep".source = builtins.toFile "keep" "";
```

clhodapp:
Sorry, I meant a (visible) empty directory.

Based on the current spec, the reason that systemd won't choke on your dotfile is the fact that it doesn't end in an extension for one of the existing types of systemd units.

So, for example, if they added a type of unit called a `keep` (e.g. `my-unit.keep`), there might be issues. 

I don't know if any of this is that big of a deal, through, given that there's an extending pattern.

clhodapp:
@kkoniuszy interestingly, I tried adding ` xdg.configFile."systemd/user/.hm-keep".text = "";` to my config (emulating ths fix) and it did not work.

Instead, I had to use `home.file.".config/systemd/user/.keep"` (note that my xdg config home is not `~/.config`) . It seems that resolution is not strictly following the xdg spec.

clhodapp:
Ah ha! I found the culprit: https://github.com/nix-community/home-manager/blob/6b1f90a8ff92e81638ae6eb48cd62349c3e387bb/modules/systemd.nix#L339

This code is not remotely new, so not a problem with your fix. I'll send a PR to address it.

kkoniuszy:
@clhodapp 

> Based on the current spec, the reason that systemd won't choke on your dotfile is the fact that it doesn't end in an extension for one of the existing types of systemd units.
> 
> So, for example, if they added a type of unit called a `keep` (e.g. `my-unit.keep`), there might be issues.

Good point. I made an assumption that systemd would ignore it because it's a dotfile, but that might not have been the case. Fortunately, it looks like it is. 

This function reads unit files from a directory
```c
// https://github.com/systemd/systemd/blob/632094641148f85d226b39ce168cf1db40d822fb/src/basic/unit-file.c#L380-L385
int unit_file_build_name_map(
                const LookupPaths *lp,
                uint64_t *cache_timestamp_hash,
                Hashmap **unit_ids_map,
                Hashmap **unit_names_map,
                Set **path_cache) {
```

and when iterating over them, it will just skip the ones with an invalid unit name.

```c
// https://github.com/systemd/systemd/blob/632094641148f85d226b39ce168cf1db40d822fb/src/basic/unit-file.c#L475-L481
if (de->d_type == DT_REG) {

        /* Accept a regular file whose name is a valid unit file name. */
        if (!unit_name_is_valid(de->d_name, UNIT_NAME_ANY))
                continue;

}
```

This is `unit_name_is_valid` 

``` c
// https://github.com/systemd/systemd/blob/632094641148f85d226b39ce168cf1db40d822fb/src/basic/unit-name.c#L42
bool unit_name_is_valid(const char *n, UnitNameFlags flags) {
```

and fortunately, it will reject all dotfiles.

```c
// https://github.com/systemd/systemd/blob/632094641148f85d226b39ce168cf1db40d822fb/src/basic/unit-name.c#L56-L58
e = strrchr(n, '.');
if (!e || e == n)
        return false;
```

`strrchr(s, c)` returns a pointer to the last occurence of the character `c` in the string `s`, so `e==n` will be true for a dotfile as the last occurence of a dot is at the beginning of the string.

rycee:
Thanks for the fix! Perhaps it may be a bit nicer to make the activation script a bit more robust instead? Something like

``` patch
diff --git a/modules/systemd.nix b/modules/systemd.nix
index f209f65d..75b8db92 100644
--- a/modules/systemd.nix
+++ b/modules/systemd.nix
@@ -333,10 +333,17 @@ in {
           else
             "";
         in ''
+          newUnitsDir="$newGenPath/home-files/.config/systemd/user"
+          if [[ ! -d $newUnitsDir ]]; then
+            newUnitsDir=${pkgs.emptyDirectory}
+          fi
+
           ${pkgs.sd-switch}/bin/sd-switch \
             ''${DRY_RUN:+--dry-run} $VERBOSE_ARG ${timeoutArg} \
             ''${oldGenPath:+--old-units $oldGenPath/home-files/.config/systemd/user} \
-            --new-units $newGenPath/home-files/.config/systemd/user
+            --new-units "$newUnitsDir"
+
+          unset newUnitsDir
         '';
       };
```

perhaps.

kkoniuszy:
That's much better! I didn't know we had `pkgs.emptyDirectory` :)

rycee:
Thanks! I made a small fix and merged to master 🙂

Issue #4786, opened 2023-12-18
bug: `programs.starship.settings` raises error for 'list of lists' setting type


dkimbel:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

As of last year, when [Starship added a C module](https://github.com/starship/starship/commit/0863146f072ae8382be63db26dcf9ddeff967aea), Starship has had a single instance of a 'list of lists' configuration option: the C module's `commands` setting (documented [here](https://starship.rs/config/#c)).

This seems to violate Home Manager's expectations for `programs.starship.settings`, resulting in an error. For instance, this configuration:
```
        starship = {
          enable = true;
          settings = {
            c = {
              commands = [[ "clang" "--version" ] [ "gcc" "--version" ] [ "cc" "--version" ]];
            };
          };
        };
```
Will lead to this error:
```
       error: A definition for option `home-manager.users.dk.programs.starship.settings.c.commands."[definition 1-entry 1]"' is not of type `boolean or signed integer or string or attribute set of (boolean or signed integer or string)'. Definition values:
       - In `/nix/store/a8gjlzrlss739zskkfrfbv807p9d5661-source/darwin/home-manager.nix':
           [
             "clang"
             "--version"
           ]
```

The same problem comes up when using TOML directly. For instance, you'll get the same error by setting `starship.settings = pkgs.lib.importTOML ../starship.toml;` when these are the contents of `starship.toml`:
```
[c]
commands = [ [ 'clang', '--version' ], [ 'gcc', '--version' ], [ 'cc', '--version' ] ]
```

On a more meta note: thank you for maintaining home-manager! It's been a joy to use. Sorry if this is the wrong place or format for this issue report.

Also -- my 'system information' has a rather alarming claim of 'nixpkgs: `not found`' (related to my using a flake? or nix-darwin?), but generally my setup is working.

### System information

```markdown
- system: `"aarch64-darwin"`
 - host os: `Darwin 23.2.0, macOS 14.2`
 - multi-user?: `no`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.19.2`
 - nixpkgs: `not found`
```

rennsax:
Same issues. No workaround now?

PR #5575, opened 2024-06-24
starship: fix type of settings to allow all valid value

rennsax:
### Description

This PR will close https://github.com/nix-community/home-manager/issues/4786.

<details>
<summary>Original comment</summary>
<blockquote>
Type checking for `programs.starship.settings` is nearly a redundant, since:
- It can hardly prevent invalid options in advance, as we cannot copy the whole schema from starship.
- The settings will be passed to `tomlFormat.generate`, which does check the type.
- It can be incomplete and refuse some valid options, e.g., `listOf (listOf str)`.
Just keep the type check as simple as possible, and `starship` will emit warings/errors when it sees invalid options.
</blockquote>
</details>

Changes the original `type` to `tomlFormat.type`, so `programs.starship.settings` can accept all valid TOML-style values.

### Checklist

- [x] (skipped) Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).

    ```
    {component}: {description}

    {long description}
    ```

  - [x] (skipped) Added myself as module maintainer. See [example](https://github.com/nix-community/home-manager/blob/068ff76a10e95820f886ac46957edcff4e44621d/modules/programs/lesspipe.nix#L6).

### DIFF
--- a/modules/programs/starship.nix
+++ b/modules/programs/starship.nix
@@ -24,14 +24,7 @@ in {
     };
 
     settings = mkOption {
-      type = with types;
-        let
-          prim = either bool (either int str);
-          primOrPrimAttrs = either prim (attrsOf prim);
-          entry = either prim (listOf primOrPrimAttrs);
-          entryOrAttrsOf = t: either entry (attrsOf t);
-          entries = entryOrAttrsOf (entryOrAttrsOf entry);
-        in attrsOf entries // { description = "Starship configuration"; };
+      type = tomlFormat.type;
       default = { };
       example = literalExpression ''
         {


modules/programs/starship.nix
```diff
@@ -24,14 +24,7 @@ in {
...
     settings = mkOption {
-      type = with types;
-        let
-          prim = either bool (either int str);
-          primOrPrimAttrs = either prim (attrsOf prim);
-          entry = either prim (listOf primOrPrimAttrs);
-          entryOrAttrsOf = t: either entry (attrsOf t);
-          entries = entryOrAttrsOf (entryOrAttrsOf entry);
-        in attrsOf entries // { description = "Starship configuration"; };
+      type = with types; attrsOf anything;
```
jian-lin, line 27: https://nixos.org/manual/nixos/unstable/#sec-settings-options

```suggestion
      type = tomlFormat.type;
```

jian-lin (changes requested):

modules/programs/starship.nix
rennsax: Good idea. I changed to this.

jian-lin (approved):
LGTM if you also update the commit message (and other things such as PR title and description).

rennsax:
> LGTM if you also update the commit message (and other things such as PR title and description).

How about the title/commit message "starship: simplify `type` field of `programs.starship.settings`"?

jian-lin:
I would go with `startship: fix type of settings to allow all valid value`

rennsax:
I've modified my commit. Thanks for your review! @jian-lin

rennsax:
I have no idea why the test failed.

rycee:

PR #5576, opened 2024-06-24
systemd: don't try to restart templates

me-and:
If the user has template services in their systemd configuration, these can't be restarted, and will produces warnings during the activation phase.  Avoid those warnings by skipping any uninstantiated templates when looking for services to start or stop.

### Description

If a user has template files in their systemd configuration `unit@.service` rather than `unit.service` or `unit@instance.service`, these will produce error messages during the Home Manager activation phase, as they can't be activated or queried; systemctl will only let you activate/query instances of those templates. Avoid this by ignoring such files when looking for units that might want to be stopped or started.

This is fundamentally only a cosmetic change; the warnings are benign, they're just also unnecessary and unhelpful. I'm seeing them because I'm migrating my config files into Home Manager gradually, and in the interim I have a bunch of template systemd units that are stored in a separate Git repository.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/systemd-activate.sh
+++ b/modules/systemd-activate.sh
@@ -34,7 +34,7 @@ function systemdPostReload() {
     touch "$oldServiceFiles"
   else
     find "$oldUserServicePath" \
-      -maxdepth 1 -name '*.service' -exec basename '{}' ';' \
+      -maxdepth 1 -name '*.service' \! -name '*@.service' -exec basename '{}' ';' \
       | sort \
       > "$oldServiceFiles"
   fi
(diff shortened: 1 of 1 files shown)


me-and:
I marked this as a draft while I ran the tests. The tests are failing for me, with the following error, but they're failing the same without my fix, so as best I can tell everything is fine…

```
error: executing shell '/nix/store/rdd4pnr4x9rqc9wgbibhngv217w2xvxl-bash-interactive-5.2p26/bin/bash': Argument list too long
```

rycee:

PR #5585, opened 2024-06-28
eww: string based config

b3nb5n:
### Description

change eww options to allow for configuration via nix. the current implementation sets the entire config directory from a source path which makes it so the entire config must be static files, and the config cannot be split up/appended to from other modules. setting the entire directory also makes it so that fixing this must break backwards compatibility (as far as i can tell). I dont know what the policy is for breaking changes, but im happy to update with any necessary changes

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@mainrs

### DIFF
--- a/modules/programs/eww.nix
+++ b/modules/programs/eww.nix
@@ -22,18 +22,48 @@ in {
       '';
     };
 
-    configDir = mkOption {
-      type = types.path;
-      example = literalExpression "./eww-config-dir";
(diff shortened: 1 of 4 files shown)


trueNAHO:
For reference, this is the exact opposite of https://github.com/nix-community/home-manager/pull/4595.

b3nb5n:
closing this in favor of #5979

PR #5596, opened 2024-07-02
yazi: allow literal string for `initLua`

uncenter:
### Description

Currently `programs.yazi.initLua` only accepts a path/module which seems unnecessarily constraining and I doubt most users will be creating a separate file for what is likely very little lines of Lua code. This change makes the `initLua` option work for literal strings as well.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@eljamm @XYenon

### DIFF
--- a/modules/programs/yazi.nix
+++ b/modules/programs/yazi.nix
@@ -141,7 +141,7 @@ in {
     };
 
     initLua = mkOption {
-      type = with types; nullOr path;
+      type = with types; nullOr (either path lines);
       default = null;
       description = ''
         The init.lua for Yazi itself.
@@ -210,7 +210,12 @@ in {
       "yazi/theme.toml" = mkIf (cfg.theme != { }) {
         source = tomlFormat.generate "yazi-theme" cfg.theme;
       };
-      "yazi/init.lua" = mkIf (cfg.initLua != null) { source = cfg.initLua; };
+      "yazi/init.lua" = mkIf (cfg.initLua != null)
+        (if builtins.isPath cfg.initLua then {
+          source = cfg.initLua;
+        } else {
+          text = cfg.initLua;
+        });
     } // (mapAttrs' (name: value:
       nameValuePair "yazi/flavors/${name}.yazi" { source = value; })
       cfg.flavors) // (mapAttrs' (name: value:
--- a/tests/modules/programs/yazi/default.nix
+++ b/tests/modules/programs/yazi/default.nix
@@ -1,5 +1,6 @@
 {
   yazi-settings = ./settings.nix;
+  yazi-init-lua-string = ./init-lua-string.nix;
   yazi-bash-integration-enabled = ./bash-integration-enabled.nix;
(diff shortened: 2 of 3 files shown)


XYenon:
Could you add a test case for this?

eljamm:
Should the other config files be changed as well or would that add too much complexity?

modules/programs/yazi.nix
```diff
@@ -141,7 +141,7 @@ in {
     };
 
     initLua = mkOption {
-      type = with types; nullOr path;
+      type = with types; nullOr (either path str);
```
sumnerevans, line 144: I think that `lines` is more correct than `str`.

```suggestion
      type = with types; nullOr (either path lines);
```

sumnerevans (changes requested):
I concur with @XYenon that this needs tests.

modules/programs/yazi.nix
isabelroses: Had to check source for this one, https://github.com/NixOS/nixpkgs/blob/d8724afca4565614164dd81345f6137c4c6eab21/lib/types.nix#L473 it should be good

tests/modules/programs/yazi/default.nix
```diff
@@ -1,5 +1,6 @@
 {
   yazi-settings = ./settings.nix;
+  yazi-init-lua-string = ./init-lua-string.nix;
```
eljamm, line 3: We should make another test `init-lua-path.nix` for when the input is a path:

```suggestion
  yazi-init-lua-path = ./init-lua-path.nix;
  yazi-init-lua-string = ./init-lua-string.nix;
```

```nix
# init-lua-string.nix
{ ... }: {
  programs.yazi = {
    enable = true;

    initLua = ./init.lua;
  };

  test.stubs.yazi = { };

  nmt.script = ''
    assertFileContent home-files/.config/yazi/init.lua \
      ${./init.lua}
  '';
}
```

eljamm (changes requested):
There is still one test case missing as we're only checking when the input is a string but not when it's a path.

Don't know if it's possible to include both test cases in the same file, but if not then we need to make a new one.

tests/modules/programs/yazi/default.nix
uncenter: This is already covered by https://github.com/nix-community/home-manager/blob/36e2f9da91ce8b63a549a47688ae60d47c50de4b/tests/modules/programs/yazi/settings.nix#L72 though right?

eljamm: Yup, you're right. I didn't see that before.

eljamm (approved):
LGTM!

uncenter:
Tests on macOS are failing but it doesn't seem to be anything remotely related to this PR's changes... 🤔

eljamm:
Yeah, it apparently has something to do to mpv or swift, but I don't think we have to worry about that.

rycee:

PR #5600, opened 2024-07-03
home-manager: add repl subcommand

me-and:
I keep finding myself wanting to inspect the computed result of my Home Manager config. NixOS has `nixos-rebuild repl`, which launches the REPL with the NixOS configuration already loaded; it's essentially a shortcut for `nix repl --file '<nixos/nixos>'`.  This PR adds similar function to `home-manager`.

No tests, as this is fundamentally an interactive user-facing tool command that doesn't lend itself well to automated testing. All the existing tests pass, except for the "Argument list too long" error I also saw in #5576, which (a) I'm confident isn't related and (b) I'm now investigating separately.

### Description

Similar to the `nixos-rebuild repl` command, `home-manager repl` will launch the Nix read-evaluate-print-loop environment with the Home Manager configuration loaded.

To make that more useful, also add the pkgs and options attributes from the generated Home Manager configuration to the environment.

This doesn't currently work with flakes, because I don't use them and I'm not confident I could safely test that function.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

No obvious maintainer, so tagging @rycee as the person whose name keeps coming up in the history logs.

### DIFF
--- a/home-manager/home-manager
+++ b/home-manager/home-manager
@@ -637,6 +637,28 @@ function doBuild() {
     presentNews
 }
 
+function doRepl() {
+    setFlakeAttribute
+    if [[ -v FLAKE_CONFIG_URI ]]; then
+        _i 'home-manager repl does not (yet) support flakes' >&2
+        return 1
+    fi
+
+    setConfigFile
+
+    extraArgs=()
+    for p in "${EXTRA_NIX_PATH[@]}"; do
+        extraArgs+=(-I "$p")
+    done
+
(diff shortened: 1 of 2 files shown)


me-and:
The REPL this generates includes `pkgs` and therefore `pkgs.lib`, but that `lib` doesn't include the `lib.hm` extras that Home Manager provides, and ideally it would. I don't think that blocks this PR, but it would be a very useful enhancement...

PR #5601, opened 2024-07-03
sway: fix systemd variables example

sg-qwt:
same as #5211


rycee:

PR #5603, opened 2024-07-04
home-manager: set FLAKE_ARG with HOME_MANAGER_PATH

usertam:
### Description

I use a flake setup with home-manager. Currently setting `programs.home-manager.path` will still result in:

```
No configuration file found. Please create one at ~/.config/home-manager/home.nix
```

And so I have to either `ln -s <dotfiles> ~/.config/home-manager` or each time do `home-manager switch --flake <dotfiles>` which kinda defeats the purpose of having `programs.home-manager.path`.

I think it makes sense for home-manager to set `FLAKE_ARG` automatically when both `programs.home-manager.path` and `$path/flake.nix` exist.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/home-manager/home-manager
+++ b/home-manager/home-manager
@@ -100,6 +100,10 @@ function setHomeManagerNixPath() {
     local path="@HOME_MANAGER_PATH@"
 
     if [[ -n "$path" ]] ; then
+        if [[ -e "$path/flake.nix" ]] ; then
+            # Set FLAKE_ARG if not set by --flake already
+            FLAKE_ARG=${FLAKE_ARG:-$path}
+        fi
         if [[ -e "$path" || "$path" =~ ^https?:// ]] ; then
             EXTRA_NIX_PATH+=("home-manager=$path")
             return


home-manager/home-manager
```diff
@@ -100,6 +100,10 @@ function setHomeManagerNixPath() {
     local path="@HOME_MANAGER_PATH@"
 
     if [[ -n "$path" ]] ; then
+        if [[ -e "$path/flake.nix" ]] ; then
+            # Set FLAKE_ARG if not set by --flake already
+            FLAKE_ARG=${FLAKE_ARG:-$path}
```
ambroisie, line 105:
```suggestion
            : "${FLAKE_ARG:=$path}"
```

usertam:
I just picked the edit here, let me know if a squash is needed.

usertam:
I think I might have misunderstood `programs.home-manager.path`. Originally I thought it was like [environment.darwinConfig](https://daiderd.com/nix-darwin/manual/index.html#opt-environment.darwinConfig) but then it was actually used for the `home-manager` package itself, and not the config. Both my usage and this PR are wrong.

In this case I think an extra option like `programs.home-manager.config` would be more proper.

Issue #4049, opened 2023-06-02
Add declarative config options for Syncthing


Anomalocaridid:
### Description
The syncthing service can be declaratively configured at the OS level using the various nixpkgs options under `services.syncthing`. However, Home Manager lacks equivalents for all except for the bare minimum, which makes declaratively configuring syncthing from userspace significantly more difficult.

charlot-shaw:
I also would appreciate better configuration for syncthing via home-manager.

Thymelizabeth:
Syncthing has just changed to using $XDG_STATE_DIR by default, which could change things.

MattSturgeon:
Looks like NixOS does this fairly _imperatively_;

At build time it [generates](https://github.com/NixOS/nixpkgs/blob/3cb4ae6689d2aa3f363516234572613b31212b78/nixos/modules/services/networking/syncthing.nix#L48-L144) an `updateConfig` shell script, which does most of its work by calling Syncthing's REST API (at `localhost:8384`) using `curl`.

This shell script then gets [run by](https://github.com/NixOS/nixpkgs/blob/3cb4ae6689d2aa3f363516234572613b31212b78/nixos/modules/services/networking/syncthing.nix#L713) one of the systemd services.

I don't see any reason why the NixOS solution couldn't be ported to home-manager more-or-less as-is?

TheHans255:
+1 for this. In particular, I can't set up Syncthing as a system level service because I need to have two separate instances of Syncthing running on the same computer for different users.

HPRIOR:
> Just as a heads up for those of you (like @HPRIOR) that use synching also on Darwin: PR #6104 expands the linux solution for a declarative config from the recently merged PR #5616 to Darwin. I've been using it for the past two months in an environment with Linux and Darwin systems without problems. I'd be happy over any reviews on the PR or general feedback if you run into problems.

Great thanks. I'll check it out when I get a chance

PR #5616, opened 2024-07-07
syncthing: expand declarative config

karaolidis:
### Description

This PR expands the Syncthing config to allow declarative settings. Code mostly pulled from [Nixpkgs module](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/networking/syncthing.nix). Addresses #4049.

### Changes to Nix Module

Removed the following options:

- `user`, `group`, `systemService`: Unnecessary since Syncthing always runs as the user declaring the config.

- `dataDir` `configDir`, `databaseDir`: Pointed to ~/.local/state/syncthing, the default Syncthing directory.

- `openDefaultPorts`: We don't have access to the system firewall.

Furthermore, multiple changes to `systemd` services were made to maintain consistency with other `home-manager` modules, sandboxing options might need to be reviewed further.

### Checklist

### Notes

I have marked the PR as a draft while I test it over the next week or so.

#### Maintainer CC

@rycee

### DIFF
--- a/modules/services/syncthing.nix
+++ b/modules/services/syncthing.nix
@@ -5,44 +5,584 @@ with lib;
 let
 
   cfg = config.services.syncthing;
+  settingsFormat = pkgs.formats.json { };
+  cleanedConfig =
+    converge (filterAttrsRecursive (_: v: v != null && v != { })) cfg.settings;
+
+  isUnixGui = (builtins.substring 0 1 cfg.guiAddress) == "/";
+
+  # Syncthing supports serving the GUI over Unix sockets. If that happens, the
+  # API is served over the Unix socket as well.  This function returns the correct
+  # curl arguments for the address portion of the curl command for both network
+  # and Unix socket addresses.
+  curlAddressArgs = path:
+    if isUnixGui
+    # if cfg.guiAddress is a unix socket, tell curl explicitly about it
+    # note that the dot in front of `${path}` is the hostname, which is
+    # required.
+    then
+      "--unix-socket ${cfg.guiAddress} http://.${path}"
+      # no adjustements are needed if cfg.guiAddress is a network address
+    else
+      "${cfg.guiAddress}${path}";
+
+  devices = mapAttrsToList (_: device: device // { deviceID = device.id; })
+    cfg.settings.devices;
+
+  folders = mapAttrsToList (_: folder:
+    folder // {
+      devices = map (device:
+        if builtins.isString device then {
+          deviceId = cfg.settings.devices.${device}.id;
+        } else
+          device) folder.devices;
+    }) (filterAttrs (_: folder: folder.enable) cfg.settings.folders);
+
+  jq = "${pkgs.jq}/bin/jq";
+  sleep = "${pkgs.coreutils}/bin/sleep";
+  printf = "${pkgs.coreutils}/bin/printf";
+  cat = "${pkgs.coreutils}/bin/cat";
+  curl = "${pkgs.curl}/bin/curl";
+  install = "${pkgs.coreutils}/bin/install";
+  syncthing = "${pkgs.syncthing}/bin/syncthing";
+
+  updateConfig = pkgs.writers.writeBash "merge-syncthing-config" (''
+    set -efu
+
+    # be careful not to leak secrets in the filesystem or in process listings
+    umask 0077
+
+    curl() {
+        # get the api key by parsing the config.xml
+        while
+            ! ${pkgs.libxml2}/bin/xmllint \
+                --xpath 'string(configuration/gui/apikey)' \
+                ''${XDG_STATE_HOME:-$HOME/.local/state}/syncthing/config.xml \
+                >"$RUNTIME_DIRECTORY/api_key"
+        do ${sleep} 1; done
+        (${printf} "X-API-Key: "; ${cat} "$RUNTIME_DIRECTORY/api_key") >"$RUNTIME_DIRECTORY/headers"
+        ${curl} -sSLk -H "@$RUNTIME_DIRECTORY/headers" \
+            --retry 1000 --retry-delay 1 --retry-all-errors \
+            "$@"
+    }
+  '' +
+
+    /* Syncthing's rest API for the folders and devices is almost identical.
+       Hence we iterate them using lib.pipe and generate shell commands for both at
+       the same time.
+    */
+    (lib.pipe {
+      # The attributes below are the only ones that are different for devices /
+      # folders.
+      devs = {
+        new_conf_IDs = map (v: v.id) devices;
+        GET_IdAttrName = "deviceID";
+        override = cfg.overrideDevices;
+        conf = devices;
+        baseAddress = curlAddressArgs "/rest/config/devices";
+      };
+      dirs = {
+        new_conf_IDs = map (v: v.id) folders;
+        GET_IdAttrName = "id";
+        override = cfg.overrideFolders;
+        conf = folders;
+        baseAddress = curlAddressArgs "/rest/config/folders";
+      };
+    } [
+      # Now for each of these attributes, write the curl commands that are
+      # identical to both folders and devices.
+      (mapAttrs (conf_type: s:
+        # We iterate the `conf` list now, and run a curl -X POST command for each, that
+        # should update that device/folder only.
+        lib.pipe s.conf [
+          # Quoting https://docs.syncthing.net/rest/config.html:
+          #
+          # > PUT takes an array and POST a single object. In both cases if a
+          # given folder/device already exists, it’s replaced, otherwise a new
+          # one is added.
+          #
+          # What's not documented, is that using PUT will remove objects that
+          # don't exist in the array given. That's why we use here `POST`, and
+          # only if s.override == true then we DELETE the relevant folders
+          # afterwards.
+          (map (new_cfg: ''
+            curl -d ${
+              lib.escapeShellArg (builtins.toJSON new_cfg)
+            } -X POST ${s.baseAddress}
+          ''))
+          (lib.concatStringsSep "\n")
+        ]
+        /* If we need to override devices/folders, we iterate all currently configured
+           IDs, via another `curl -X GET`, and we delete all IDs that are not part of
+           the Nix configured list of IDs
+        */
+        + lib.optionalString s.override ''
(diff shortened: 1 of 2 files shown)


karaolidis:
Things appear to be working so far :)

zorrobert:
This looks great! I'm using syncthing as well and it would be great to manage it via home-manager. Thanks for your effort!

bphenriques:
I was looking for this as well! This is great, thank you!

qenya:
Ah, this is exactly what I was looking for too! Thank you for writing this!

karaolidis:
Updated branch to master and made syncthing folders relative to the user's home :)

wrvsrx:
Users might have syncthing folders whose paths are not in home.

karaolidis:
Hm, I guess that's fair. I just thought that making the options consistent with e.g. `home.file` would be the way to go. 

Should I undo that change?

wrvsrx:
Thanks for you work. I think original version would be better.

Besides, I was wondering if you can add an option to use a local file to specify password, so that hashed password won't appear in nix store. 

Here's my patch:
https://github.com/nix-community/home-manager/commit/34ea327a46219ade90fc6689b5c8ee6918d36fae

karaolidis:
Done, thanks for the feedback :)

Changes made:

- Add `passwordFile` option
- Revert home directory change
- Add optional `$XDG_STATE_HOME` support if it is set, see https://docs.syncthing.net/users/config.html
- Update to master

Testing now and will mark as Ready for review after that.

karaolidis:
Would adding a `gui` section in the options that allows users to set both `user` and either a `password` or `passwordFile` make sense, or am I overthinking this?

cc @wrvsrx

pitkling:
@karaolidis Thanks for the work on this PR! I'm using it in my private home-manager fork and it works rather well so far. One question concerning the handling of the `guiAddress` setting (which actually also applies to the original NixOS module): Is there a reason for having this as a separate option that is given to the synching service via the command line as opposed to setting it via `settings.gui.address` and have it handled "natively" by synching via its configuration file?

karaolidis:
> Is there a reason for having this as a separate option that is given to the synching service via the command line as opposed to setting it via `settings.gui.address` and have it handled "natively" by syncthing via its configuration file?

I am assuming that since the module sets the settings semi-imperatively using `curl`s, we need to pass in the GUI address in the command line, otherwise Syncthing will start with its default listen address. I think that _technically_ this could also be set through `settings.gui.address` instead of `guiAddress`, but the underlying implementation would remain the same.

pitkling:
> > Is there a reason for having this as a separate option that is given to the synching service via the command line as opposed to setting it via `settings.gui.address` and have it handled "natively" by syncthing via its configuration file?
> 
> I am assuming that since the module sets the settings semi-imperatively using `curl`s, we need to pass in the GUI address in the command line, otherwise Syncthing will start with its default listen address. I think that _technically_ this could also be set through `settings.gui.address` instead of `guiAddress`, but the underlying implementation would remain the same.

I see, thanks. I had a look at the implementation and I think you're right. In particular, the very first time the synching configuration is run, when there is no configuration file yet, the command line argument is required to ensure that syncthing starts with the correct address.

While this is probably out of scope for this PR, it would be nice to avoid the command-line argument. It can be quite confusing that the address given in the config file is completely ignored by the systemd service (that's how I stumbled upon this). E.g., one could create a minimal configuration if there is no config file present that has the correct gui address before starting syncthing. This would potentially also avoid other artifacts of the current implementation, like a "~/Sync" folder being created (by the default configuration of synching) on the very first boot, even though the declarative syncthing configuration does not have such a folder.

Anyway, thanks again for the PR!

wrvsrx (approved):

karaolidis:
> This would potentially also avoid other artifacts of the current implementation, like a "~/Sync" folder being created (by the default configuration of synching) on the very first boot, even though the declarative syncthing configuration does not have such a folder.

I know that this doesn't really solve the problem but, FYI, you can work around this with `extraOptions = [ "-no-default-folder" ];` :)

pitkling:
> > This would potentially also avoid other artifacts of the current implementation, like a "~/Sync" folder being created (by the default configuration of synching) on the very first boot, even though the declarative syncthing configuration does not have such a folder.
> 
> I know that this doesn't really solve the problem but, FYI, you can work around this with `extraOptions = [ "-no-default-folder" ];` :)

Good tip, thanks! Actually, shouldn't this be a default command-line argument in a declarative syncthing configuration (that specifies at least one folder)?

karaolidis:
> > > This would potentially also avoid other artifacts of the current implementation, like a "~/Sync" folder being created (by the default configuration of synching) on the very first boot, even though the declarative syncthing configuration does not have such a folder.
> > 
> > 
> > I know that this doesn't really solve the problem but, FYI, you can work around this with `extraOptions = [ "-no-default-folder" ];` :)
> 
> Good tip, thanks! Actually, shouldn't this be a default command-line argument in a declarative syncthing configuration (that specifies at least one folder)?

The `nixpkgs` module doesn't add it, so I didn't add it here for consistency. I did think about making a PR for the original module, but that wouldn't really be backwards compatible if someone was relying on the default behavior :/

modules/services/syncthing.nix
```diff
@@ -78,10 +611,28 @@ in {
...
+                    } "$syncthing_dir/key.pem"
+                  ''}
+                ''
+              }";
             ExecStart = escapeShellArgs syncthingArgs;
             Restart = "on-failure";
             SuccessExitStatus = [ 3 4 ];
             RestartForceExitStatus = [ 3 4 ];
+            Environment =
+              mkIf (cfg.all_proxy != null) { inherit (cfg) all_proxy; };
```
pitkling, line 635: The original NixOS module adds `config.networking.proxy.envVars` to the environment, which seems to [contain not only `all_proxy` but also other, global proxy variables](https://github.com/NixOS/nixpkgs/blob/787623bca83b9d067a1eac881b24548b4003e5eb/nixos/modules/config/networking.nix#L210-L226), at least some of which (like `http_proxy` and `https_proxy`) are seemingly also [used by Syncthing](https://docs.syncthing.net/users/proxying.html). I'm not really familiar with proxy settings, but is there a reason not to add the remaining proxy variables like the original module does?

modules/services/syncthing.nix
```diff
@@ -5,52 +5,585 @@ with lib;
...
     "-logflags=0"
   ];
 
   syncthingArgs = defaultSyncthingArgs ++ cfg.extraOptions;
-
 in {
   meta.maintainers = [ maintainers.rycee ];
 
-  options = {
+  options = with types; {
```
pitkling, line 170: At the moment, this PR enables the declarative Syncthing configuration only on Linux, not on Darwin. The documentation of the module should mention this at some place. Or how is this handled in other Home Manager modules?

As a related side note: I used your (very nice!) PR as a base in [my own home-manger fork](https://github.com/pitkling/home-manager/tree/syncthing-declarative) to add support for Darwin. The first 5 additional commits are mostly some refactoring to make it possible to reuse some of the code. The [last commit](https://github.com/nix-community/home-manager/commit/e40308638db88656394adef26970cc667b00076e) adds two launchd agents akin to the two systemd services you define in your PR.

I assume it might be cleaner to make the Darwin support a separate PR? Anyway, I'm still testing those commits, although so far everything seems to work fine :).

pitkling (changes requested):

modules/services/syncthing.nix
karaolidis: In the `home-manager` module, I think `config` does not give us access to `networking.proxy.envVars` as it points to the `home-manager` configuration itself, and I am not quite sure how to access items in the root config. I would appreciate any ideas on how to do this.

modules/services/syncthing.nix
karaolidis: I definitely think it should be a separate PR, I don't have a Darwin device to test things on and I wouldn't want to push untested code ;)

modules/services/syncthing.nix
pitkling: There is the `osConfig` module argument (see [Home Manager manual](https://nix-community.github.io/home-manager/index.xhtml#sec-install-nixos-module)). ~Although I'm not sure how this is handled if Home Manager is used standalone.~ In fact, some searching revealed that it's simply not available in a standalone installation.

I guess one could bind the module parameters via something like `@args{ config, lib, pkgs, ... }` and check for `osConfig` via `args ? osConfig` to include the system proxy variables if available.

karaolidis: I force-pushed a version with `osConfig`, but a quick search doesn't appear to show any other modules using it :thinking:

pitkling: > I force-pushed a version with `osConfig`, but a quick search doesn't appear to show any other modules using it 🤔

I would assume that's because it's little known (I think the installation part of the manual is the only place where it is mentioned, and it's not the first place one would look for it) and since it works only if HM is used as a nixOS or nix-darwin module.

Personally I think it's ok to use as long as it's safe-guarded via something like `args ? osConfig`, to make sure the module does not break on standalone HM, but not sure how important the system proxy settings are.

pitkling (approved):
I've been using this to configure several systems during the past few weeks and didn't find any problems. Would be great if this gets merged eventually 🙂.

AshleyYakeley:
Would love to see this. What's the status of this PR?

karaolidis:
It's been good to go for a while, I have simply been updating the PR as new commits come into main.

cc: @rycee

AshleyYakeley:
If this takes too long, one option would be to separate it out as a flake. I do this with [NixVirt](/AshleyYakeley/NixVirt), which is a flake that provides a Home Manager module and a NixOS module.

rycee:
Thanks, great stuff! Merged to master now 🙂

karaolidis:
Much appreciated 🙏

leoperegrino:
@karaolidis Is there any reason to remove `dataDir` and `configDir`? It's not obvious to me that this has to be hardcoded just because the syncthing defaults are the xdg dirs. If I pass `-data`/`-config` via `extraOptions` then the `syncthing-init` service fails because is unable to make the API calls to ensure `overrideFolders`/`overrideDevices`.

karaolidis:
Back when I made the PR, I did not know as much about Nix as I do now, and it felt like the safest option at the time. Feel free to make an extension PR if you'd like, as I can see this being useful. :)

PR #5624, opened 2024-07-09
xdg-desktop-entries: allow `terminal` to be null

iynaix:
### Description

`terminal` should not be set for desktop entries of `type = Directory`, as per the [Desktop Entry Specification](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#value-types)

> Some keys only make sense in the context when another particular key is also present and set to a specific value. Those keys should not be used if the particular key is not present or not set to the specific value. For example, the Terminal key can only be used when the value of the Type key is Application.

Testing of this behavior is already enforced by `desktop-file-validate` during the `checkPhase` of `make-desktopitem` in nixpkgs; it errors if you currently try to create a desktop entry of type Directory.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@ncfavier

### DIFF
--- a/modules/misc/xdg-desktop-entries.nix
+++ b/modules/misc/xdg-desktop-entries.nix
@@ -51,7 +51,7 @@ let
 
       terminal = mkOption {
         description = "Whether the program runs in a terminal window.";
-        type = types.bool;
+        type = types.nullOr types.bool;
         default = false;
       };
 


Mag1cByt3s:
This pull request fixes the issue that .directory entries cannot be created by home-manager since home-manager always expects the key 'terminal' to be present which is not allowed for .directory entries.

rycee:

PR #5627, opened 2024-07-10
Add river.systemd.runInService option

zoriya:
### Description

Add a systemd.inService option to run river in systemd (via `systemctl --user start river`). This means that start & STOP events work for river in systemd. Services that bind to `graphical-session.target` will get stopped when river stops or when the computer shutdown. Without this, services never stops, and it leads to a "Stop session timeout" when shutting down. Stopping & restarting river without this option also does not restart graphic services (and they still refer to the old DISPLAY/WAYLAND_DISPLAY vars).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@GaetanLepage

### DIFF
--- a/modules/services/window-managers/river.nix
+++ b/modules/services/window-managers/river.nix
@@ -1,16 +1,14 @@
 { config, lib, pkgs, ... }:
-
 let
   inherit (lib) types;
   cfg = config.wayland.windowManager.river;
 
   # Systemd integration
   variables = builtins.concatStringsSep " " cfg.systemd.variables;
-  extraCommands = builtins.concatStringsSep " "
-    (map (f: "&& ${f}") cfg.systemd.extraCommands);
-  systemdActivation = ''
-    ${pkgs.dbus}/bin/dbus-update-activation-environment --systemd ${variables} ${extraCommands}
-  '';
+  systemdActivation = builtins.concatStringsSep " && " ([
+    "${pkgs.dbus}/bin/dbus-update-activation-environment --systemd ${variables}"
+  ] ++ (lib.optional cfg.systemd.runInService "systemd-notify --ready")
+    ++ cfg.systemd.extraCommands);
 
   toValue = val:
     if lib.isString val || lib.isDerivation val then
@@ -28,7 +26,6 @@ let
 
   # Intermediary function that converts some value (attrs, str, ...) to one or several commands.
   toArgs = path: value:
-
     let
       stringValue = lib.concatStringsSep " " (path ++ [ (toValue value) ]);
       finalValue = if lib.isAttrs value then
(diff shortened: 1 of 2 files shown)


modules/services/window-managers/river.nix
```diff
@@ -180,28 +191,52 @@ in {
       ### SHELL VARIABLES ###
       ${config.lib.shell.exportAll cfg.extraSessionVariables}
 
+    '' + (lib.optionalString cfg.systemd.enable ''
```
zoriya, line 194: I moved systemd activation before configuration because we want `graphical-session.target` services to be started before `settings.spawm` apps are.

zoriya:
I'm on a limited internet, so I can't run tests, is there a way to pull minimal dependencies for running tests (or can someone approve the workflows, so I at least can use the CI for that)?

modules/services/window-managers/river.nix
```diff
@@ -1,16 +1,14 @@
...
   # Systemd integration
   variables = builtins.concatStringsSep " " cfg.systemd.variables;
-  extraCommands = builtins.concatStringsSep " "
-    (map (f: "&& ${f}") cfg.systemd.extraCommands);
-  systemdActivation = ''
-    ${pkgs.dbus}/bin/dbus-update-activation-environment --systemd ${variables} ${extraCommands}
-  '';
+  systemdActivation = builtins.concatStringsSep " && " ([
+    "${pkgs.dbus}/bin/dbus-update-activation-environment --systemd ${variables}"
+  ] ++ (if cfg.systemd.runInService then [ "systemd-notify --ready" ] else [ ])
```
GaetanLepage, line 10:
```suggestion
  ] ++ (lib.optional cfg.systemd.runInService "systemd-notify --ready")
```

zoriya: I can't find the doc for this, is it a shortcut for lib.lists.optional?

GaetanLepage: Yes :)

modules/services/window-managers/river.nix
```diff
@@ -100,10 +110,11 @@ in {
 
       extraCommands = lib.mkOption {
         type = types.listOf types.str;
-        default = [
+        default = if !cfg.systemd.runInService then [
           "systemctl --user stop river-session.target"
           "systemctl --user start river-session.target"
-        ];
+        ] else
+          [ ];
```
GaetanLepage, line 117: You can use `lib.optionals (!cfg.systemd.runInService) [ ... ]` here.

GaetanLepage (approved):
LGTM !

GaetanLepage:
@rycee: This PR looks ready to me. Feel free to have a look whenever you can/want :)

modules/services/window-managers/river.nix
zoriya: I don't understand why, but tests don't pass with this change. It passes if I leave the if/else tho.

zoriya:
@rycee (i can't ask for a review so i'm pinging instead)

this PR is still something i think should be merged (and im using it since i made it).
I rebased it to latest master, lmk if i need to change anything.

zoriya:
Recently learned about uwsm which does a better job than this PR. I recommend anyone that want to run their compositor in systemd to give it a shot instead of this PR.

PR #5628, opened 2024-07-10
hypridle: Use After graphical-session.target instead of pre target

zoriya:
### Description

Before this, hypridle was running before graphical-session.target was started so it was missing the WAYLAND_DISPLAY var & could not start successfully. This PR adds a constraint to make hypridle start after the compositor, making sure WAYLAND_DISPLAY is available.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@khaneliman @fufexan


rycee:
Are you certain this is necessary. I tried [setting up hypridle](https://git.sr.ht/~rycee/configurations/tree/8e1a3552cc25357500a3932b1fa26cb8eeea5377/item/user/common.nix#L1539-1568) and it seems to start just fine as-is. That said, for me the graphical-session.target is started by Hyprland itself using an `exec-once` in `hyprland.conf`. How do you start your compositor?

zoriya:
My compositor is bound to graphical-session.target, I start it using `systemctl --user start river` using the new option I added in #5627

khaneliman (approved):

zoriya:
@rycee How could we get this merged?

fufexan:
Imo https://github.com/nix-community/home-manager/pull/6086 achieves the same result and is more ergonomic in case people use UWSM-managed sessions.

PR #5631, opened 2024-07-11
dircolors: respect preferXdgDirectories if set

svrana:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/dircolors.nix
+++ b/modules/programs/dircolors.nix
@@ -54,7 +54,16 @@ in {
     };
   };
 
-  config = mkIf cfg.enable (mkMerge [
+  config = let
+    dircolorsPath = if config.home.preferXdgDirectories then
+      "${config.xdg.configHome}/dir_colors"
+    else
+      "~/.dir_colors";
+
+    dircolorsConfig = concatStringsSep "\n" ([ ]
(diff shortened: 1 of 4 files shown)


teto (approved):
I dont know if the duplication of the code to just to change the path is better or worse than just passing a different path ? might make the module harder to maintain.

svrana:
I'm open to just changing the path too, but thought some might prefer the current placement. Thanks for taking a look.

teto:
sry what I meant is that you could have a function returning the same code accepting the path in argument :
```
mkDirColorsConfig = destFolder:
        eval $(${pkgs.coreutils}/bin/dircolors -b ${destFolder}/dir_colors)
      '';

      programs.fish.shellInit = mkIf cfg.enableFishIntegration ''
        eval (${pkgs.coreutils}/bin/dircolors -c ${destFolder}/dir_colors)
      '';

    })
    
```

and then call

```
if xdg then mkDirColorsConfig config.xdg.configHome else mkDirColorsConfig config.home
```

svrana:
Tried cleaning this up as suggested, but you have to account for the filename changing as well (from .dir_colors when in $HOME and dir_colors in XDG) and still had the big if block to construct the files differently, so it seemed like a wash. But maybe someone with better nix foo can get rid of the duplication.

PR #5632, opened 2024-07-11
firefox: add preConfig

dani0854:
### Description

I run into a problem, when using opinionated default user.js such as [yokoffing/Betterfox](https://github.com/yokoffing/Betterfox). They define a bunch of preferences, and then you can additionally overwrite them if you want.

In order to use it I added user.js to `extraConfig`. But since it `extraConfig` comes after `settings`, it seems to be impossible to overwrite values using `settings`. While it is still possible to append lines to `extraConfig` after default user.js, using `settings` seems more logical.

I thought of a few solutions:
1. Swap `extraConfig` and `settings` in the generated user.js file. Since `settings` are user defined, there aren't really many reasons to overwrite something in `extraConfig` after. But that will break backwards compatibility if someone uses that as a feature.
2. Add `preConfig` (I am happy to discuss naming), which acts like `extraConfig`, but placed before `settings`. This solution I currently implemented in this PR, since it keeps backward compatibility, although I would still like to discuss if solution 1 is possible.
3. Add a separate options for those predefined default user.js, like `extensions`. But then there would need to be some standard with regard to how they packaged, and I think it will just overcomplicate things. Also, 2 and 3 perfectly allow to easily use any user.js with `fetchgit`, or a simple package, and also allow a bunch more use cases where there is some default predefined config, and then some overwrites in `settings`

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee @kira-bruneau


nyabinary:
cc @rycee

dani0854:
As suggested in #6196, added news message.

PR #5637, opened 2024-07-14
udiskie: enable udisks2

BhasherBEL:
### Description

Enable udisks2 by default for udiskie, to prevent dbus error:
```
gi.repository.GLib.GError: g-dbus-error-quark: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name org.freedesktop.UDisks2 was not provided by any .service files (2)
```

Inspired by: https://discourse.nixos.org/t/udiskie-no-longer-runs/23768

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/services/udiskie.nix
+++ b/modules/services/udiskie.nix
@@ -112,6 +112,8 @@ in {
         cfg.settings
       ]);
 
+    udisks2.enable = true;
+
     systemd.user.services.udiskie = {
       Unit = {
         Description = "udiskie mount daemon";


modules/services/udiskie.nix
```diff
@@ -112,6 +112,8 @@ in {
         cfg.settings
       ]);
 
+    udisks2.enable = true;
```
ambroisie, line 115: This is a NixOS option.

teto:
@ambroisie shall we close then ?

PR #5639, opened 2024-07-15
nnn: add cd on quit option

giang-nghg:
- add bash, zsh, fish, and nushell integration options
- add quitcd configuration for bash, zsh, fish, and nushell

### Description

Configure [cd on quit](https://github.com/jarun/nnn/wiki/Basic-use-cases#configure-cd-on-quit)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/nnn.nix
+++ b/modules/programs/nnn.nix
@@ -104,6 +104,46 @@ in {
         '';
         default = { };
       };
+
+      enableBashIntegration = mkOption {
+        default = true;
+        type = types.bool;
+        description = ''
+          Whether to enable Bash integration.
+        '';
+      };
+
+      enableZshIntegration = mkOption {
+        default = true;
+        type = types.bool;
+        description = ''
+          Whether to enable Zsh integration.
+        '';
+      };
+
+      enableFishIntegration = mkOption {
+        default = true;
+        type = types.bool;
+        description = ''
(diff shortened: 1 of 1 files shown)


modules/programs/nnn.nix
```diff
@@ -104,6 +104,38 @@ in {
...
         default = { };
       };
+
+      enableBashIntegration = mkOption {
+        default = true;
+        type = types.bool;
+        description = ''
+          Whether to enable Bash integration.
+        '';
+      };
```
sumnerevans, line 114: All of these should be `mkEnableOption`

```suggestion
      enableBashIntegration = mkEnableOption "Bash integration" // {
        default = true;
      };
```

modules/programs/nnn.nix
```diff
@@ -104,6 +104,38 @@ in {
...
+
+      quitcd = mkOption {
+        default = false;
+        type = types.bool;
+        description = ''
+          Whether to enable cd on quit.
+        '';
+      };
```
sumnerevans, line 138:
```suggestion
      quitcd = mkEnableOption "change directory on quit";
```

sumnerevans (changes requested):

giang-nghg:
@sumnerevans Hi, could you take another look? Thanks!

PR #5641, opened 2024-07-16
keepassxc: add module

d-brasher:
### Description

Add module for [KeePassXC](https://keepassxc.org/), a popular password manager.

Implements handling of config file.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/keepassxc.nix
+++ b/modules/programs/keepassxc.nix
@@ -1,0 +1,56 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+let
+  cfg = config.programs.keepassxc;
+
+  iniFormat = pkgs.formats.ini { };
+in
+{
+  meta.maintainers = [ lib.maintainers.d-brasher ];
+
+  options.programs.keepassxc = {
+    enable = lib.mkEnableOption "keepassxc";
+
+    package = lib.mkPackageOption pkgs "keepassxc" { };
+
+    settings = lib.mkOption {
+      type = iniFormat.type;
+      default = { };
+      example = lib.literalExpression ''
+        {
+          Browser.Enabled = true;
+
+          GUI = {
+            AdvancedSettings = true;
+            ApplicationTheme = "dark";
+            CompactMode = true;
+            HidePasswords = true;
+          };
+
+          SSHAgent.Enabled = true;
+        }
+      '';
+      description = ''
+        Configuration written to
+        {file}`$XDG_CONFIG_HOME/keepassxc/keepassxc.ini`.
+
+        See <https://github.com/keepassxreboot/keepassxc/blob/647272e9c5542297d3fcf6502e6173c96f12a9a0/src/core/Config.cpp#L49-L223>
+        for the full list of options.
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+    xdg.configFile = {
+      "keepassxc/keepassxc.ini" = lib.mkIf (cfg.settings != { }) {
+        source = iniFormat.generate "keepassxc-settings" cfg.settings;
+      };
+    };
+  };
+}
--- a/modules/programs/keepassxc.nix
+++ b/modules/programs/keepassxc.nix
@@ -16,7 +16,7 @@ in
   options.programs.keepassxc = {
     enable = lib.mkEnableOption "keepassxc";
 
-    package = lib.mkPackageOption pkgs "keepassxc" { };
+    package = lib.mkPackageOption pkgs "keepassxc" { nullable = true; };
 
     settings = lib.mkOption {
       type = iniFormat.type;
@@ -46,7 +46,7 @@ in
   };
 
   config = lib.mkIf cfg.enable {
-    home.packages = [ cfg.package ];
+    home.packages = lib.mkIf (cfg.package != null) [ cfg.package ];
     xdg.configFile = {
       "keepassxc/keepassxc.ini" = lib.mkIf (cfg.settings != { }) {
         source = iniFormat.generate "keepassxc-settings" cfg.settings;
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -152,6 +152,7 @@ let
       ./programs/just.nix
       ./programs/k9s.nix
       ./programs/kakoune.nix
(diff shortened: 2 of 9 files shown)


d-brasher:
While testing this I noticed that during the app startup, a non-invasive red banner shows up for a few seconds, saying "_Access error for config file .config/keepassxc/keepassxc.ini_"

That is expected, since the config file will now be read-only and any setting change done through the GUI will not be saved.

I don't think there is a way to prevent that banner from being displayed.

mainrs:
> While testing this I noticed that during the app startup, a non-invasive red banner shows up for a few seconds, saying "_Access error for config file .config/keepassxc/keepassxc.ini_"

You can override your `keepassxc` from packages and add a patch that removes this line (or comments it out): 

https://github.com/keepassxreboot/keepassxc/blob/e48ef80e9c81a16ab971ca0e40e0ab4cfb39defb/src/gui/ApplicationSettingsWidget.cpp#L480

You have to locally compile the program, but it'll remove the error message.

mainrs (approved):
LGTM.

mainrs:
You can add a news item to `modules/misc.news.nix` that mentions the addition of the module. Just take a look at the file for examples.

mainrs:
> If you contribute a module then you don’t need to add this entry, the merger will create an entry for you.

Ahhh, I didn't know that. Well, then you don't have to. 

Regarding your other question, I don't think so, no. I mentioned it to help you out and in the case others try to search something similar.

d-brasher:
> > While testing this I noticed that during the app startup, a non-invasive red banner shows up for a few seconds, saying "_Access error for config file .config/keepassxc/keepassxc.ini_"
> 
> You can override your `keepassxc` from packages and add a patch that removes this line (or comments it out):
> 
> https://github.com/keepassxreboot/keepassxc/blob/e48ef80e9c81a16ab971ca0e40e0ab4cfb39defb/src/gui/ApplicationSettingsWidget.cpp#L480
> 
> You have to locally compile the program, but it'll remove the error message.

In case somebody is interested, I wrote a patch to change some error messages to be more explicative (and remove the error on startup):

<details>

<summary><em>readonly-config-messages.patch</em></summary>

```
Change "read-only config" error messages to mention Home Manager.
Remove "Access error for config file" error on startup. 
--- a/src/gui/ApplicationSettingsWidget.cpp
+++ b/src/gui/ApplicationSettingsWidget.cpp
@@ -197,7 +197,8 @@
 void ApplicationSettingsWidget::loadSettings()
 {
     if (config()->hasAccessError()) {
-        showMessage(tr("Access error for config file %1").arg(config()->getFileName()), MessageWidget::Error);
+	 showMessage(tr("Config file is managed declaratively by Home Manager. "
+                       "Any setting changed here will not be saved."), MessageWidget::Error);
     }
 
 #ifdef QT_DEBUG
@@ -343,7 +344,8 @@
 void ApplicationSettingsWidget::saveSettings()
 {
     if (config()->hasAccessError()) {
-        showMessage(tr("Access error for config file %1").arg(config()->getFileName()), MessageWidget::Error);
+        showMessage(tr("Config file is managed declaratively by Home Manager. "
+                       "Settings can't be changed from here."), MessageWidget::Error);
         // We prevent closing the settings page if we could not write to
         // the config file.
         return;
@@ -477,7 +479,8 @@
     }
 
     if (config()->hasAccessError()) {
-        showMessage(tr("Access error for config file %1").arg(config()->getFileName()), MessageWidget::Error);
+        showMessage(tr("Config file is managed declaratively by Home Manager. "
+                       "Settings can't be reset from here."), MessageWidget::Error);
         // We prevent closing the settings page if we could not write to
         // the config file.
         return;
--- a/src/gui/MainWindow.cpp
+++ b/src/gui/MainWindow.cpp
@@ -623,8 +623,6 @@
     connect(&m_trayIconTriggerTimer, SIGNAL(timeout()), SLOT(processTrayIconTrigger()));
 
     if (config()->hasAccessError()) {
-        m_ui->globalMessageWidget->showMessage(tr("Access error for config file %1").arg(config()->getFileName()),
-                                               MessageWidget::Error);
     }
 
     // Properly shutdown on logoff, restart, and shutdown
```

</details>

It can be used by specifying
```
package = pkgs.keepassxc.overrideAttrs (oldAttrs: {
  patches = (oldAttrs.patches or []) ++ [ ./readonly-config-messages.patch ];
};
```
inside this module config.

d-brasher:
I was thinking if it was possible to integrate the patch in the module somehow (maybe with a setting that enables the `overrideAttrs` to the package) but the issues could be:
- It requires a local rebuild (but this can be specified in the setting description)
- The patch needs to be kept up to date (and I don't think there is a way to check that the patch works every time the package is updated (?))
Any opinions?

mainrs:
I feel like Pat he's like these are better suited as part of nixpkgs. Although I think this particular patch would probably not fit there well.

I am no export by any means. The packages that included patches usually do so to enable nix functionality or include fixes for erroneous software.

I think it is best if everybody hosts their own patch file.

modules/programs/keepassxc.nix
```diff
@@ -0,0 +1,56 @@
...
+
+  iniFormat = pkgs.formats.ini { };
+in
+{
+  meta.maintainers = [ lib.maintainers.d-brasher ];
+
+  options.programs.keepassxc = {
+    enable = lib.mkEnableOption "keepassxc";
+
+    package = lib.mkPackageOption pkgs "keepassxc" { };
```
khaneliman, line 19:
```suggestion
    package = lib.mkPackageOption pkgs "keepassxc" { nullable = true; };
```

modules/programs/keepassxc.nix
```diff
@@ -0,0 +1,56 @@
...
+        {file}`$XDG_CONFIG_HOME/keepassxc/keepassxc.ini`.
+
+        See <https://github.com/keepassxreboot/keepassxc/blob/647272e9c5542297d3fcf6502e6173c96f12a9a0/src/core/Config.cpp#L49-L223>
+        for the full list of options.
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    home.packages = [ cfg.package ];
```
khaneliman, line 49:
```suggestion
    home.packages = lib.mkIf (cfg.package != null) [ cfg.package ];
```

PR #5643, opened 2024-07-18
programs.nix-your-shell: add module

terlar:
### Description

This is alternative to #4645 which stalled. More along the lines of how starship and zoxide is implemented.

Closes #4645

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/nix-your-shell.nix
+++ b/modules/programs/nix-your-shell.nix
@@ -1,0 +1,55 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.programs.nix-your-shell;
+
+in {
+  meta.maintainers = [ maintainers.terlar ];
+
+  options.programs.nix-your-shell = {
+    enable = mkEnableOption ''
+      {command}`nix-your-shell`, a wrapper for `nix develop` or `nix-shell`
+      to retain the same shell inside the new environment'';
+
+    package = mkPackageOption pkgs "nix-your-shell" { };
+
+    enableFishIntegration = mkEnableOption "Fish integration" // {
+      default = true;
+    };
+
+    enableNushellIntegration = mkEnableOption "Nushell integration" // {
+      default = true;
+    };
+
+    enableZshIntegration = mkEnableOption "Zsh integration" // {
+      default = true;
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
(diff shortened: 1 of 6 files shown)


modules/programs/nix-your-shell.nix
```diff
@@ -0,0 +1,56 @@
...
+  cfg = config.programs.nix-your-shell;
+
+in {
+  meta.maintainers = [ maintainers.terlar ];
+
+  options.programs.nix-your-shell = {
+    enable = mkEnableOption ''
+      {command}`nix-your-shell`, a wrapper for `nix develop` or `nix-shell`
+      to retain the same shell inside the new environment.
+    '';
```
sumnerevans, line 16: the trailing period is added by `mkEnableOption`

```suggestion
      to retain the same shell inside the new environment'';
```

tests/modules/programs/nix-your-shell/enable-shells.nix
```diff
@@ -0,0 +1,52 @@
...
+      enableFishIntegration = true;
+      enableNushellIntegration = true;
+      enableZshIntegration = true;
+    };
+    fish.enable = true;
+    nushell.enable = true;
+    zsh.enable = true;
+  };
+
+  # Needed to avoid error with dummy fish package.
```
sumnerevans, line 16: where is `fish` being stubbed?

sumnerevans (changes requested):

tests/modules/programs/nix-your-shell/enable-shells.nix
terlar: Seems fish ran into other issues when being stubbed (complain about `fish_indent`), and I didn't see any other tests stubbing it, so I removed that.

modules/programs/nix-your-shell.nix
sumnerevans: putting `''` on the same line was intentional (a space is added before the `.` otherwise)

terlar: Sorry, missed that, pushed a fix.

terlar:
@sumnerevans sorry to waste your time with the previous roundtrip and unfixed suggestion. Hopefully all should be good now.

dawidd6 (approved):

terlar:
@rycee Do you think we could get this merged?

rycee:

PR #5645, opened 2024-07-19
eww: add terminal integration options

Jeansidharta:
### Description

Use eww's shell-completions command to generate completions for bash, zsh and fish.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@mainrs

### DIFF
--- a/modules/programs/eww.nix
+++ b/modules/programs/eww.nix
@@ -5,6 +5,7 @@ with lib;
 let
 
   cfg = config.programs.eww;
+  ewwCmd = "${cfg.package}/bin/eww";
 
 in {
   meta.maintainers = [ hm.maintainers.mainrs ];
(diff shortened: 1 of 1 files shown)


modules/programs/eww.nix
```diff
@@ -30,10 +31,34 @@ in {
         {file}`$XDG_CONFIG_HOME/eww`.
       '';
     };
+
+    enableBashIntegration = mkEnableOption "Bash integration";
```
mainrs, line 35: Can you enable these completion options by default, please?

modules/programs/eww.nix
```diff
@@ -30,10 +31,34 @@ in {
...
+
+    enableZshIntegration = mkEnableOption "Zsh integration";
```
mainrs, line 37: Same.

modules/programs/eww.nix
```diff
@@ -30,10 +31,34 @@ in {
...
+
+    enableFishIntegration = mkEnableOption "Fish integration";
```
mainrs, line 39: Same.

mainrs (changes requested):
It would be good to see if eww also has completions for other shells like nushell. If we're already at it.

Jeansidharta:
When running `eww shell-completions --shell` on version 0.6.0 I got this message:

```
[❯] eww shell-completions --shell
error: a value is required for '--shell <SHELL>' but none was supplied
  [possible values: bash, elvish, fish, powershell, zsh]

For more information, try '--help'.
```

- elvish is not in the home-manager modules
- powershell is windows only.

So, no nushell or anything else to implement, unfortunately.

Jeansidharta:
Just enabled them by default. I didn't do this at first with a fear of breaking some users environment. But this should be fine? Probably?

mainrs (approved):
> Just enabled them by default. I didn't do this at first with a fear of breaking some users environment. But this should be fine? Probably?

Good question. I just took a look at other modules and most of them enabled them by default. I guess theoretically this will override existing completions.

LGTM. I think this is ready to be merged.

rycee:

PR #5647, opened 2024-07-19
Fix yazi keymap example to use "run"

404Wolf:
### Description

Yazi now uses `run` instead of `exec` for commands. I just updated the example to reflect this. They will be removing "exec" in the near future.

#### Maintainer CC

@XYenon @eljamm

### DIFF
--- a/tests/modules/programs/yazi/keymap-expected.toml
+++ b/tests/modules/programs/yazi/keymap-expected.toml
@@ -1,27 +1,27 @@
 [[input.keymap]]
-exec = "close"
 on = ["<C-q>"]
+run = "close"
 
 [[input.keymap]]
-exec = "close --submit"
 on = ["<Enter>"]
+run = "close --submit"
 
 [[input.keymap]]
-exec = "escape"
 on = ["<Esc>"]
+run = "escape"
 
 [[input.keymap]]
-exec = "backspace"
 on = ["<Backspace>"]
+run = "backspace"
 
 [[manager.keymap]]
-exec = "escape"
 on = ["<Esc>"]
+run = "escape"
 
 [[manager.keymap]]
-exec = "quit"
 on = ["q"]
+run = "quit"
 
 [[manager.keymap]]
-exec = "close"
 on = ["<C-q>"]
+run = "close"
--- a/modules/programs/yazi.nix
+++ b/modules/programs/yazi.nix
@@ -70,15 +70,15 @@ in {
       example = literalExpression ''
         {
           input.keymap = [
-            { exec = "close"; on = [ "<C-q>" ]; }
-            { exec = "close --submit"; on = [ "<Enter>" ]; }
-            { exec = "escape"; on = [ "<Esc>" ]; }
-            { exec = "backspace"; on = [ "<Backspace>" ]; }
+            { on = [ "<C-q>" ]; run = "close"; }
+            { on = [ "<Enter>" ];  run = "close --submit"; }
+            { on = [ "<Esc>" ];  run = "escape"; }
+            { on = [ "<Backspace>" ]; run = "backspace"; }
           ];
           manager.keymap = [
-            { exec = "escape"; on = [ "<Esc>" ]; }
-            { exec = "quit"; on = [ "q" ]; }
-            { exec = "close"; on = [ "<C-q>" ]; }
+            {  on = [ "<Esc>" ]; run = "escape"; }
+            {  on = [ "q" ]; run = "quit"; }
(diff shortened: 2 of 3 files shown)


eljamm:
Thanks @404Wolf!

There appears to be more occurrences of this, can you correct them as well?

Screenshot from 2024-07-19 20-10-23

I think it would be better to change them all right now and not just do it for the example.

Also, can you rename the commit to something that follows the [project commit style](https://nix-community.github.io/home-manager/index.xhtml#sec-commit-style) better? Perhaps something like:

```txt
yazi: use `run` instead of `exec` in keymaps

{long description}
```

You can optionally add a `{long description}` to explain the change or add more details.

404Wolf:
> Thanks @404Wolf!
> 
> There appears to be more occurrences of this, can you correct them as well?
> 
> Screenshot from 2024-07-19 20-10-23
> 
> I think it would be better to change them all right now and not just do it for the example.
> 
> Also, can you rename the commit to something that follows the [project commit style](https://nix-community.github.io/home-manager/index.xhtml#sec-commit-style) better? Perhaps something like:
> 
> ```
> yazi: use `run` instead of `exec` in keymaps
> 
> {long description}
> ```
> 
> You can optionally add a `{long description}` to explain the change or add more details.

Of course! I think I made all the changes you requested :)

XYenon (approved):
LGTM

eljamm (approved):
This is just a small nitpick, but I don't think the commit subject should be capitalized.

Aside from that, this looks good to me as well 😄

404Wolf:
Okay, I changed it to lowercase

XYenon (approved):

tests/modules/programs/yazi/keymap-expected.toml
```diff
@@ -1,27 +1,27 @@
 [[input.keymap]]
-exec = "close"
+run = "close"
 on = ["<C-q>"]
```
rycee, line 3: Will have to reorder the keys to account for alphabetic ordering. I.e., doing

```suggestion
on = ["<C-q>"]
run = "close"
```

for all examples.

rycee: You can use `nix build -L .#test-yazi-settings` to check the test.

404Wolf: Okay I've made sure it's alphabetical and that the test passes. Just out of curiosity, how does it know to generate in alphabetical order?

eljamm: I think the reason is that in Nix, all attributes are evaluated to an alphabetical order, no matter in what order you define them. For example:

```
nix-repl> test = { b = "world"; a = "hello"; }

nix-repl> test
{
  a = "hello";
  b = "world";
}
```

As such, the toml file generated from the keymap attrset contains keys which are sorted in an alphabetical order, which I assume is because the toml generator takes the inputs from first to last. In the example above, it starts with a, then b.

However, the keys in the expected toml file in the tests are not sorted, so a conflict occurs and the test fails.

404Wolf: That makes sense, thanks! I'm used to hashset orders being non-deterministic, I like this nix solution.

eljamm:
@rycee I think this is good to merge. Is there something blocking this?

404Wolf:
Didn't realize this was still hanging, thanks for bumping it :)

Adda0 (approved):
LGTM. I was surprised to see the `exec` when looking for what configuration options there are in hm for Yazi. Are there any blocking issues to merging the PR?

eljamm:
> I was surprised to see the exec

Yeah, it seems that `exec` has been outright removed since v3.0.0:

```
Error: TOML parse error at line 8, column 1
  |
8 | [[manager.prepend_keymap]]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^
missing field `run`
```

> Are there any blocking issues to merging the PR?

I don't really think so.

@rycee If there are no issues, can this be merged, please?

teto:
the issue is CI. Macos is broken (as usual), probably not because of this PR. which prevents merging (from me at least).

eljamm:
@404Wolf Can you rebase on top of master? Perhaps that will make CI pass.

Adda0:
These changes have already been applied in a different PR. This PR can be closed.

sumnerevans:
This appears to have been superseded by by https://github.com/nix-community/home-manager/pull/6111. Closing.

PR #5648, opened 2024-07-19
eza: fix fish completion

Samuel-Martineau:
### Description

Currently, when using eza with `enableFishIntegration` set to true, ls is aliased to eza. eza's completions are also correctly added to ls. However, the builtin ls completions are currently *not* removed. This means that some flags, like `--dired`, are still suggested, even though they do not exist on eza.

This pull request introduces a small fix via `programs.fish.plugins` which disables the builtin ls completions if and only if  `enableFishIntegration` is true, leaving only the eza completions. 

### Checklist

- [] Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@cafkafk

### DIFF
--- a/modules/programs/eza.nix
+++ b/modules/programs/eza.nix
@@ -93,6 +93,10 @@ with lib;
 
     programs.fish.shellAliases = optionsAlias
       // optionalAttrs cfg.enableFishIntegration aliases;
+    programs.fish.plugins = mkIf cfg.enableFishIntegration [{
+      name = "eza-remove-default-ls-completion";
+      src = pkgs.writeTextDir "completions/ls.fish" "";
+    }];
 
     programs.ion.shellAliases = optionsAlias
       // optionalAttrs cfg.enableIonIntegration aliases;


teto:
ain't that a fish bug ? not recognizing the alias takes priority over the binary ? if you `unalias ls`, then you lose the completion for ls so I am not sure it's the proper fix. Maybe remove the fish support ?

cafkafk (changes requested):
Thanks for fixing this, I never noticed this before, but now that I do, it should obviously be solved somehow. 

I think this fix is okay, but could you add some comments just explaining what it is that we are doing to this diff, just so we don't forget why we are doing this in the future.

(also, the conflicts must be resolved before merge)

teto:
so if you want to use native ls (`\ls` on bash, dunno how it works in fish) one would lose completion so we trade one problem for another ?

PR #5657, opened 2024-07-23
nix-gc: call nix-gc in a shell script to match upstream behavior

Ninja3047:
### Description

This will match the behavior in the upstream service which allows the user to set options to something that uses shell syntax.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@shivaraj-bh

### DIFF
--- a/modules/services/nix-gc.nix
+++ b/modules/services/nix-gc.nix
@@ -110,9 +110,10 @@ in {
       systemd.user.services.nix-gc = {
         Unit = { Description = "Nix Garbage Collector"; };
         Service = {
-          ExecStart = "${nixPackage}/bin/nix-collect-garbage ${
+          ExecStart = toString (pkgs.writeShellScript "nix-gc" ''
+            exec "${nixPackage}/bin/nix-collect-garbage ${
               lib.optionalString (cfg.options != null) cfg.options
-            }";
+            }"'');
         };
       };
       systemd.user.timers.nix-gc = {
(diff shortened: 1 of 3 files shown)


shivaraj-bh (approved):
Diff LGTM

Ninja3047:
also cc @rycee

modules/services/nix-gc.nix
```diff
@@ -110,9 +110,10 @@ in {
       systemd.user.services.nix-gc = {
         Unit = { Description = "Nix Garbage Collector"; };
         Service = {
-          ExecStart = "${nixPackage}/bin/nix-collect-garbage ${
+          ExecStart = ''
+            ${pkgs.bash}/bin/bash -c "${nixPackage}/bin/nix-collect-garbage ${
               lib.optionalString (cfg.options != null) cfg.options
-            }";
+            }"'';
```
rycee, line 116: It would probably be better to create a proper script for this to avoid potential quote escaping issues. Something like

``` nix
ExecStart = toString (pkgs.writeShellScript "nix-gc" ''
  exec ${nixPackage}/bin/nix-collect-garbage ${
    lib.optionalString (cfg.options != null) cfg.options
  }";
'');
```

should pretty closely match what is inside the NixOS module.

rycee:
Thanks for the contribution. It's a bit odd to me to use Bash to evaluate expressions like that when you could do it in Nix and directly splice in the desired value.

In any case, I added a comment that I think is worth considering.

Also, please rebase the branch ontop of the latest master to resolve the CI (Linux) error.

PR #5659, opened 2024-07-23
firefox: make NixOS wiki search engine icon visible

jirafey:
Before it was referencing: `https://wiki.nixos.org/favicon.png` which is not existing. 
Now both of the icons for NixOS packages and NixOS wiki have the same icon.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

Before:
image
After:
image

### DIFF
--- a/modules/programs/firefox.nix
+++ b/modules/programs/firefox.nix
@@ -546,7 +546,7 @@ in {
 
                     "NixOS Wiki" = {
                       urls = [{ template = "https://wiki.nixos.org/index.php?search={searchTerms}"; }];
-                      iconUpdateURL = "https://wiki.nixos.org/favicon.png";
+                      iconUpdateURL = "https://wiki.nixos.org/nixos.png";
                       updateInterval = 24 * 60 * 60 * 1000; # every day
                       definedAliases = [ "@nw" ];
                     };


kira-bruneau (commented):
Thanks @jirafey! I think it'd make sense to also update the tests that reference the old url (even though it's never actually fetched there).

jirafey:
> Thanks @jirafey! I think it'd make sense to also update the tests that reference the old url (even though it's never actually fetched there).

We should probably change this `iconUpdateURL` to the URL used for Nix Packages as the example suggests, but maybe showing 2 distinct ways of passing the URL are ideal here.

Visible here:
https://github.com/nix-community/home-manager/blob/master/modules/programs/firefox.nix#L555

Highlighted line:
`icon = "''${pkgs.nixos-icons}/share/icons/hicolor/scalable/apps/nix-snowflake.svg";`

kira-bruneau:
> We should probably change this iconUpdateURL to the URL used for Nix Packages as the example suggests, but maybe showing 2 distinct ways of passing the URL are ideal here.

Oh I intended for it to show both ways when I added that example. `icon` was just a convenience attribute that I added that gets translated into `iconURL`: https://github.com/nix-community/home-manager/blob/26059c1477116d6a3ea486eed3bb4e806ff5aaec/modules/programs/firefox.nix#L816-L818

> EDIT: I don't know how to update or even run the tests if there is someone that can do that for me that would be appreciated. Also providing an explanation or a link to such resource that would teach me to that myself would be maybe even better. My time is limited though.

Here's the documentation on testing: https://github.com/nix-community/home-manager/blob/26059c1477116d6a3ea486eed3bb4e806ff5aaec/docs/manual/contributing/tests.md

kira-bruneau:
@jirafey Were you cleaning up your repos and accidentally deleted your home-manager fork? It looks like it autoclosed this PR.

jirafey:
> @jirafey Were you cleaning up your repos and accidentally deleted your home-manager fork? It looks like it autoclosed this PR.

yes, precisely the case. I don't have a NixOS machine right now and don't plan on using nix in the future. Just an icon shouldn't be too hard for someone to change so please mark it as a `good-first-issue` as it seems not to be that important. Consider this my resignation from this issue

PR #5663, opened 2024-07-24
starship: add `enableInteractive` option for fish

### LINKED ISSUE #5445: Add programs.starship.interactiveOnly
### Description

I would like to have `programs.starship.interactiveOnly` option, the same as in NixOS: https://github.com/NixOS/nixpkgs/blob/5710852ba686cc1fd0d3b8e22b3117d43ba374c2/nixos/modules/programs/starship.nix#L33.

It is useful in conjunction with `async-prompt` fish plugin. See: https://github.com/acomagu/fish-async-prompt/issues/54.

eljamm:
### Description

- Some fish plugins such as [fish-async-prompt](https://github.com/acomagu/fish-async-prompt) require that starship be initialized as non-interactive.

When the `programs.starship.enableInteractive` option is enabled, starship is initialized at the end of the init script, outside the interactive block.

**Note:** Don't know if other shells need this, so I only added this for fish for now.

- Also changed the initialization method from `eval` to `source` as per the [starship installation](https://starship.rs/guide/#%F0%9F%9A%80-installation) guide.

Closes #5445 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -1847,6 +1847,20 @@ in {
           output.
         '';
       }
+
+      {
+        time = "2024-12-04T20:00:00+00:00";
+        condition = let
+          sCfg = config.programs.starship;
+          fCfg = config.programs.fish;
+        in sCfg.enable && sCfg.enableFishIntegration && fCfg.enable;
+        message = ''
+          A new option 'programs.starship.enableInteractive' is available for
+          the Fish shell that only enables starship if the shell is interactive.
+
+          Some plugins require this to be set to 'false' to function correctly.
+        '';
+      }
     ];
   };
 }
--- a/modules/programs/starship.nix
+++ b/modules/programs/starship.nix
@@ -119,7 +119,7 @@ in {
 
     programs.fish.${initFish} = mkIf cfg.enableFishIntegration ''
       if test "$TERM" != "dumb"
-        eval (${starshipCmd} init fish)
+        ${starshipCmd} init fish | source
         ${lib.optionalString cfg.enableTransience "enable_transience"}
       end
     '';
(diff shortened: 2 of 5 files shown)


mainrs (commented):
Did this change make  the async prompt fish plugin you mentioned in the PR work properly? 

There is the argument that, since this is for fish only, one can rename the new option to `enableInteractiveFish` or something similar.  I think this might be better, since we can always rename it once we require it for different shells. Better be more explicit than replying on documentation to mention that it's fish only

mainrs:
The failing CI job seems to be related to Swift.

eljamm:
> Did this change make the async prompt fish plugin you mentioned in the PR work properly?

Yup. I've been running this for a while with the fish-async-prompt with no issues.

> There is the argument that, since this is for fish only, one can rename the new option to `enableInteractiveFish` or something similar.

It was actually named something like that initially, but I changed it to this to be more consistent with the `enableTransience` option, which is also for fish only. Should we perhaps also rename that as well?

mainrs:
> It was actually named something like that initially, but I changed it to this to be more consistent with the `enableTransience` option, which is also for fish only. Should we perhaps also rename that as well?

Nah, leave it then. I didn't see the already existing field in the diff view (it was cut off).

mainrs (approved):
LGTM.

eljamm:
Resolved merge conflicts.

dawidd6 (approved):

eljamm:
Resolved merge conflict again with the news file.

eljamm:
I ran all the tests locally (x86_64-linux) and they all passed, so I guess the failing CI task is unrelated to changes in this PR.

rycee:

dudeofawesome:
Would it be possible to get this backported to 24.11?

PR #5670, opened 2024-07-25
firefox: Enable userChrome in about:config

### LINKED ISSUE #5666: Enable userChrome in about:config if userChrome is used
### Description

Setting `programs.firefox.profiles..userChrome` does nothing by default and requires manually setting `toolkit.legacyUserProfileCustomizations.stylesheets` to true.

Module should enable this option in `programs.firefox.profiles..settings` automatically if `userChrome` is not empty.

MrQubo:
### Description

Enable "toolkit.legacyUserProfileCustomizations.stylesheets" in about:config if userChrome is not empty.
Resolves #5666

Change is not backwards compatible. It will break configs where `userChrome` is set to non-empty string and `settings."toolkit.legacyUserProfileCustomizations.stylesheets"` is set to false. It seems unlikely but I've added assertion nevertheless.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee @kira-bruneau

### DIFF
--- a/modules/programs/firefox/mkFirefoxModule.nix
+++ b/modules/programs/firefox/mkFirefoxModule.nix
@@ -975,6 +975,10 @@ in
                 ]
               ) (lib.attrsToList config.extensions.settings))
               ++ config.bookmarks.assertions;
+
+              settings."toolkit.legacyUserProfileCustomizations.stylesheets" = mkIf (
+                config.userChrome != "" || config.userContent != ""
+              ) true;
             };
           }
         )
--- a/tests/modules/programs/firefox/profiles/userchrome/default.nix
+++ b/tests/modules/programs/firefox/profiles/userchrome/default.nix
@@ -62,6 +62,16 @@ in
           assertFileContent \
             "home-files/${cfg.profilesPath}/derivation-file/chrome/userChrome.css" \
             ${./chrome/userChrome.css}
+
+          assertFileContains \
+            "home-files/${cfg.profilesPath}/lines/user.js" \
+            'user_pref("toolkit.legacyUserProfileCustomizations.stylesheets", true)'
+          assertFileContains \
+            "home-files/${cfg.profilesPath}/file/user.js" \
+            'user_pref("toolkit.legacyUserProfileCustomizations.stylesheets", true)'
+          assertFileContains \
+            "home-files/${cfg.profilesPath}/derivation-file/user.js" \
+            'user_pref("toolkit.legacyUserProfileCustomizations.stylesheets", true)'
         '';
     }
   );


MrQubo:
Here's a more friendly diff without indentation changes:
```
diff --git a/modules/programs/firefox.nix b/modules/programs/firefox.nix
index 3bee6e0b..f946f1d2 100644
--- a/modules/programs/firefox.nix
+++ b/modules/programs/firefox.nix
@@ -301,7 +301,9 @@ in {
       };
 
       profiles = mkOption {
-        type = types.attrsOf (types.submodule ({ config, name, ... }: {
+        type = types.attrsOf (types.submodule ({ config, name, ... }:
+          let profile = config;
+          in {
           options = {
             name = mkOption {
               type = types.str;
@@ -704,7 +706,12 @@ in {
             };
 
           };
-        }));
+
+            config = {
+              settings."toolkit.legacyUserProfileCustomizations.stylesheets" =
+                mkIf (profile.userChrome != "") true;
+            };
+          }));
         default = { };
         description = "Attribute set of Firefox profiles.";
       };
@@ -757,6 +764,19 @@ in {
       }
 
       (mkNoDuplicateAssertion cfg.profiles "profile")
+
+      (let
+        assertProfile = profile:
+          profile.userChrome != ""
+          -> (profile.settings."toolkit.legacyUserProfileCustomizations.stylesheets"
+            != false);
+      in {
+        assertion = any assertProfile (attrValues cfg.profiles);
+        message = ''
+          'userChrome' won't work with
+          'settings."toolkit.legacyUserProfileCustomizations.stylesheets"' set to false.
+        '';
+      })
     ] ++ (mapAttrsToList
       (_: profile: mkNoDuplicateAssertion profile.containers "container")
       cfg.profiles);
diff --git a/tests/modules/programs/firefox/profile-settings-expected-userChrome-user.js b/tests/modules/programs/firefox/profile-settings-expected-userChrome-user.js
new file mode 100644
index 00000000..b0f8619c
--- /dev/null
+++ b/tests/modules/programs/firefox/profile-settings-expected-userChrome-user.js
@@ -0,0 +1 @@
+user_pref("toolkit.legacyUserProfileCustomizations.stylesheets", true);
diff --git a/tests/modules/programs/firefox/profile-settings-expected-userChrome-userChrome.css b/tests/modules/programs/firefox/profile-settings-expected-userChrome-userChrome.css
new file mode 100644
index 00000000..a2819786
--- /dev/null
+++ b/tests/modules/programs/firefox/profile-settings-expected-userChrome-userChrome.css
@@ -0,0 +1 @@
+#example { default: none; }
diff --git a/tests/modules/programs/firefox/profile-settings.nix b/tests/modules/programs/firefox/profile-settings.nix
index 8b781552..ceac20a9 100644
--- a/tests/modules/programs/firefox/profile-settings.nix
+++ b/tests/modules/programs/firefox/profile-settings.nix
@@ -153,6 +153,11 @@
           };
         };
       };
+
+      profiles.userChrome = {
+        id = 6;
+        userChrome = builtins.readFile ./profile-settings-expected-userChrome-userChrome.css;
+      };
     };
 
     nmt.script = ''
@@ -203,6 +208,14 @@
       assertFirefoxSearchContent \
         home-files/.mozilla/firefox/searchWithoutDefault/search.json.mozlz4 \
         ${./profile-settings-expected-search-without-default.json}
+
+      assertFileContent \
+        home-files/.mozilla/firefox/userChrome/user.js \
+        ${./profile-settings-expected-userChrome-user.js}
+
+      assertFileContent \
+        home-files/.mozilla/firefox/userChrome/chrome/userChrome.css \
+        ${./profile-settings-expected-userChrome-userChrome.css}
     '';
   };
 }
```

MrQubo:
Here's a more friendly diff without indentation changes:
<details>

```diff
diff --git a/modules/programs/firefox.nix b/modules/programs/firefox.nix
index 3bee6e0b..f946f1d2 100644
--- a/modules/programs/firefox.nix
+++ b/modules/programs/firefox.nix
@@ -301,7 +301,9 @@ in {
       };
 
       profiles = mkOption {
-        type = types.attrsOf (types.submodule ({ config, name, ... }: {
+        type = types.attrsOf (types.submodule ({ config, name, ... }:
+          let profile = config;
+          in {
           options = {
             name = mkOption {
               type = types.str;
@@ -704,7 +706,12 @@ in {
             };
 
           };
-        }));
+
+            config = {
+              settings."toolkit.legacyUserProfileCustomizations.stylesheets" =
+                mkIf (profile.userChrome != "") true;
+            };
+          }));
         default = { };
         description = "Attribute set of Firefox profiles.";
       };
@@ -757,6 +764,19 @@ in {
       }
 
       (mkNoDuplicateAssertion cfg.profiles "profile")
+
+      (let
+        assertProfile = profile:
+          profile.userChrome != ""
+          -> (profile.settings."toolkit.legacyUserProfileCustomizations.stylesheets"
+            != false);
+      in {
+        assertion = any assertProfile (attrValues cfg.profiles);
+        message = ''
+          'userChrome' won't work with
+          'settings."toolkit.legacyUserProfileCustomizations.stylesheets"' set to false.
+        '';
+      })
     ] ++ (mapAttrsToList
       (_: profile: mkNoDuplicateAssertion profile.containers "container")
       cfg.profiles);
diff --git a/tests/modules/programs/firefox/profile-settings-expected-userChrome-user.js b/tests/modules/programs/firefox/profile-settings-expected-userChrome-user.js
new file mode 100644
index 00000000..b0f8619c
--- /dev/null
+++ b/tests/modules/programs/firefox/profile-settings-expected-userChrome-user.js
@@ -0,0 +1 @@
+user_pref("toolkit.legacyUserProfileCustomizations.stylesheets", true);
diff --git a/tests/modules/programs/firefox/profile-settings-expected-userChrome-userChrome.css b/tests/modules/programs/firefox/profile-settings-expected-userChrome-userChrome.css
new file mode 100644
index 00000000..a2819786
--- /dev/null
+++ b/tests/modules/programs/firefox/profile-settings-expected-userChrome-userChrome.css
@@ -0,0 +1 @@
+#example { default: none; }
diff --git a/tests/modules/programs/firefox/profile-settings.nix b/tests/modules/programs/firefox/profile-settings.nix
index 8b781552..ceac20a9 100644
--- a/tests/modules/programs/firefox/profile-settings.nix
+++ b/tests/modules/programs/firefox/profile-settings.nix
@@ -153,6 +153,11 @@
           };
         };
       };
+
+      profiles.userChrome = {
+        id = 6;
+        userChrome = builtins.readFile ./profile-settings-expected-userChrome-userChrome.css;
+      };
     };
 
     nmt.script = ''
@@ -203,6 +208,14 @@
       assertFirefoxSearchContent \
         home-files/.mozilla/firefox/searchWithoutDefault/search.json.mozlz4 \
         ${./profile-settings-expected-search-without-default.json}
+
+      assertFileContent \
+        home-files/.mozilla/firefox/userChrome/user.js \
+        ${./profile-settings-expected-userChrome-user.js}
+
+      assertFileContent \
+        home-files/.mozilla/firefox/userChrome/chrome/userChrome.css \
+        ${./profile-settings-expected-userChrome-userChrome.css}
     '';
   };
 }
```
</details>

MrQubo:
I've tried running tests with `nix-shell --pure tests -A run.all` but I'm getting error `error: executing shell '/nix/store/p5441mp6wwpdh65qamaixvd092cwr45h-bash-interactive-5.2p26/bin/bash': Argument list too long`. :/

modules/programs/firefox.nix
```diff
@@ -301,410 +301,418 @@ in {
...
-              '';
             };
 
-          };
-        }));
+            config = {
+              settings."toolkit.legacyUserProfileCustomizations.stylesheets" =
+                mkIf (profile.userChrome != "") true;
+            };
+          }));
```
kira-bruneau, line 715: Thanksl! :blush: 

I also just verified that `userContent` also needs this option, so I think it should be included here.

MrQubo: It's the same options?? o.0 Cause it says "stylesheets" there.

MrQubo: Ah, nvm, I was thinking that userContent refers to `userContent.js`.

tests/modules/programs/firefox/profile-settings.nix
```diff
@@ -1,6 +1,8 @@
 { config, lib, pkgs, ... }:
 
-{
+let userChromeContent = "#example { display: none; }";
+
+in {
```
IldenH, line 5:
```suggestion
let 

  userChromeContent = "#example { display: none; }";

in {
```

modules/programs/firefox.nix
```diff
@@ -301,410 +301,418 @@ in {
...
-                to
-                [{option}`programs.firefox.profiles.<profile>.settings`](#opt-programs.firefox.profiles._name_.settings)
-              '';
             };
 
-          };
-        }));
+            config = {
+              settings."toolkit.legacyUserProfileCustomizations.stylesheets" =
+                mkIf (profile.userChrome != "") true;
```
IldenH, line 713:
```suggestion
                mkIf (profile.userChrome != "" || profile.userContent != "") true;
```

modules/programs/firefox.nix
```diff
@@ -301,410 +301,418 @@ in {
...
-        type = types.attrsOf (types.submodule ({ config, name, ... }: {
-          options = {
-            name = mkOption {
-              type = types.str;
-              default = name;
-              description = "Profile name.";
-            };
+        type = types.attrsOf (types.submodule ({ config, name, ... }:
+          let profile = config;
+          in {
```
IldenH, line 306:
```suggestion
          let

            profile = config;

          in {
```

modules/programs/firefox.nix
```diff
@@ -757,6 +765,19 @@ in {
...
+      (let
+        assertProfile = profile:
+          profile.userChrome != ""
+          -> (profile.settings."toolkit.legacyUserProfileCustomizations.stylesheets"
+            != false);
+      in {
+        assertion = any assertProfile (attrValues cfg.profiles);
+        message = ''
+          'userChrome' won't work with
+          'settings."toolkit.legacyUserProfileCustomizations.stylesheets"' set to false.
```
IldenH, line 778:
```suggestion
          {option}`settings."toolkit.legacyUserProfileCustomizations.stylesheets"` set to false.
```

tests/modules/programs/firefox/profile-settings.nix
```diff
@@ -153,6 +155,11 @@
           };
         };
       };
+
+      profiles.userChrome = {
+        id = 6;
+        userChrome = userChromeContent;
+      };
```
IldenH, line 162:
```suggestion
        userChrome = userChromeContent;
        userContent = userChromeContent;
      };
```

tests/modules/programs/firefox/profile-settings.nix
```diff
@@ -203,6 +210,14 @@
...
         ${./profile-settings-expected-search-without-default.json}
+
+      assertFileContains \
+        home-files/.mozilla/firefox/userChrome/user.js \
+        'user_pref("toolkit.legacyUserProfileCustomizations.stylesheets", true)'
+
+      assertFileContains \
+        home-files/.mozilla/firefox/userChrome/chrome/userChrome.css \
+        '${userChromeContent}'
     '';
```
IldenH, line 221:
```suggestion

      assertFileContains \
        home-files/.mozilla/firefox/userChrome/chrome/userContent.css \
        '${userChromeContent}'
    '';
```

modules/programs/firefox.nix
```diff
@@ -757,6 +765,19 @@ in {
       }
 
       (mkNoDuplicateAssertion cfg.profiles "profile")
+
+      (let
+        assertProfile = profile:
+          profile.userChrome != ""
```
IldenH, line 771:
```suggestion
          profile.userChrome != "" || profile.userContent != ""
```

modules/programs/firefox.nix
```diff
@@ -757,6 +765,19 @@ in {
...
+          -> (profile.settings."toolkit.legacyUserProfileCustomizations.stylesheets"
+            != false);
+      in {
+        assertion = any assertProfile (attrValues cfg.profiles);
+        message = ''
+          'userChrome' won't work with
```
IldenH, line 777:
```suggestion
          'userChrome' and/or 'userContent' don't work with
```
Unsure on the wording here

IldenH (changes requested):
Really like this change as it is confusing having the stylesheets not work despite being set.

MrQubo:
I hate github, I was replying you, but it was "pending" so I think you didn't see any of that. ;_;

modules/programs/firefox.nix
MrQubo: I added separate assertions with separate messages. This will be more clear for the users.

modules/programs/firefox.nix
MrQubo: I also added `{option}` to `userChrome`. Not sure if it works? I don't know how to check how it looks like rendered.

IldenH (approved):

modules/programs/firefox.nix
MrQubo: At least here, it's not being handled at all. .-.
https://home-manager-options.extranix.com/?query=services.polybar.config&release=master

kira-bruneau:
It looks like this needs to rebased since: #5128

MrQubo:
Rebased, I've made some additional changes, could you review again?

tests/modules/programs/firefox/profile-settings.nix
```diff
@@ -154,12 +157,18 @@ in {
       id = 5;
       containers = {
         "shopping" = {
-          id = 6;
```
kira-bruneau, line 157: Wait what was wrong with the containers test? Each container should have its own unique id set, so I think it makes sense to keep one explicitly set in the test.

modules/programs/firefox.nix
kira-bruneau: I don't think it makes sense to use `{option}` for assertions. I'm not 100% sure how it's handled for the docs, but assertions don't get rendered like the docs do. They just get dumped out verbatim on the terminal.

tests/modules/programs/firefox/profile-settings.nix
MrQubo: It was a bit confusing, because it's container id, not profile id. It looks like someone set it to 6 because it's the next number after 5, but these ids are unrelated. I feel like there's no need to set container id here (apart from making this test failproof against changing the default container id), and it's not testing anything.

I could set it explicitly to 1, but using 6 here is a bit confusing, because it looks like it's related to this 5.

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -801,7 +812,28 @@ in {
...
+          profile.${userChromeAttr} != ""
+          -> (profile.settings."toolkit.legacyUserProfileCustomizations.stylesheets"
+            != false);
+
+        mkAssertion = userChromeAttr: {
+          assertion =
+            all (assertProfile userChromeAttr) (attrValues cfg.profiles);
+          message = ''
+            {option}`${userChromeAttr}` won't work with
+            {option}`settings."toolkit.legacyUserProfileCustomizations.stylesheets"` set to false.
```
kira-bruneau, line 830: I'm not sure if you saw my other comment on this since it was in a resolved thread, but `` {option}`x` `` shouldn't be used in assertions, because they don't get rendered like the docs do.

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -343,410 +343,421 @@ in {
...
-                "browser.bookmarks.showMobileBookmarks" = true;
-                "browser.newtabpage.pinned" = [{
-                  title = "NixOS";
-                  url = "https://nixos.org";
-                }];
-              }
-            '';
-            description = ''
-              Attribute set of ${name} preferences.
+          profile = config;
```
kira-bruneau, line 349: nit: It would cleanup the diff a lot if you just referred to `config` directly rather than giving it an alias of profile, I could see why you'd want to do this to avoid confusion with the top-level `config` though.

rycee: Agreed, a decent middle way may be to introduce the `profile = config` binding inside the `config` attribute. I.e.,

```nix
config = let profile = config; in {
  settings."toolkit.legacyUserProfileCustomizations.stylesheets" =
    mkIf (profile.userChrome != "" || profile.userContent != "") true;
};
```

MrQubo: You're maintainers, so what do you think it's better for maintainability? Using it only in `config` means, that if in the future it won't be possible to refer to `profile` in `settings` in case it's needed e.g. for some `default`.

The bad things are that it will mess up any other PR's that are open right now, and it will also mess up `git blame` (though, there is `git blame -w`, which ignores those indentation changes, not idea whether github uses it).

modules/programs/firefox/mkFirefoxModule.nix
MrQubo: Ok. I left the backticks characters.

modules/programs/firefox/mkFirefoxModule.nix
rycee: I'd say move the let to the config attribute as in my previous comment. We can deal with the (probably unlikely) event that the profile variable needs to be referenced in the options attribute when it happens 🙂

MrQubo: Done

kira-bruneau (approved):

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -833,27 +867,27 @@ in {
...
+      "${cfg.profilesPath}/${profile.path}/chrome/userChrome.css" =
         mkIf (profile.userChrome != "") { text = profile.userChrome; };
 
-      "${profilesPath}/${profile.path}/chrome/userContent.css" =
+      "${cfg.profilesPath}/${profile.path}/chrome/userContent.css" =
         mkIf (profile.userContent != "") { text = profile.userContent; };
 
-      "${profilesPath}/${profile.path}/user.js" = mkIf (profile.settings != { }
-        || profile.extraConfig != "" || profile.bookmarks != [ ]) {
+      "${cfg.profilesPath}/${profile.path}/user.js" = mkIf (profile.settings
```
MrQubo, line 878: I think it hurts readability how `./format` broke the line before `!=`. :/

MrQubo:
I (hopefully) fixed the tests on macos.

MrQubo:
Maybe it would be easier to fix the tests in https://github.com/nix-community/home-manager/pull/5698 and keep this PR self-contained.

kira-bruneau:
> Maybe it would be easier to fix the tests in https://github.com/nix-community/home-manager/pull/5698 and keep this PR self-contained.

Thanks @MrQubo I'll push the fix to https://github.com/nix-community/home-manager/pull/5698 if it passes here.

kira-bruneau:
It looks like it's still failing here

MrQubo:
@kira-bruneau I can't understand why it's failing.

MrQubo:
Seems like the macos tests where fixed in master.

MrQubo:
Now ubuntu tests are broken in master. ;_;

MrQubo:
The code moved from `default.nix` to `mkFirefoxModule.nix`, but the changes are still the same.

MrQubo:
No idea what is the error now.
It says "NOTE: If your Manual build passes locally and you see this message in CI, you probably need a rebase.", but I already rebased onto master.

nyabinary:
> No idea what is the error now. It says "NOTE: If your Manual build passes locally and you see this message in CI, you probably need a rebase", but I already rebased onto master.

Is this working at its current state, or are there still errors?

MrQubo:
Tests on this PR need fixing: https://github.com/nix-community/home-manager/pull/6403

MrQubo:
I decided to not add an assertion after all. The error message about conflicting values of `settings."toolkit.legacyUserProfileCustomizations.stylesheets"` should be enough for the users.
Not having an assertion means users can use `lib.mkForce` to still set it to false if they need to (e.g. maybe someone wants to enable it dynamically with some other method).

MrQubo:
I came to fix conflicts and someone merge exactly the same feature. #9532
.-.

PR #5672, opened 2024-07-25
sheldon: add module for zsh, bash and fish

Kyure-A:
### Description
Add module for https://github.com/rossmacarthur/sheldon

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@rycee


mainrs (changes requested):
`sheldon` also provides the `sheldon completions --shell <bash|zsh>` command that prints the completion script to stdout.

Might as well add the configuration flags to the module and add the `eval` lines to bashrc/zshrc. Take a look at the `eww` module and how that is done.

modules/programs/sheldon.nix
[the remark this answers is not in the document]
Kyure-A: I thought it might be too much for me too, so I changed it to make it simpler. Thanks.

Kyure-A:
I didn't know Sheldon had a completion script function. Thank you very much! I have added it as a setting.

modules/programs/sheldon.nix
```diff
@@ -0,0 +1,73 @@
...
+
+    settings = mkOption {
+      inherit (tomlFormat) type;
+      default = { };
+      description = "";
+      example = literalExpression "";
+    };
+
+    enableZshCompletions = mkEnableOption "Zsh completions" // {
+      default = false;
```
mainrs, line 30: These three should be `default = true`.

mainrs (changes requested):
Once changed, the PR should be ready to be merged.

Kyure-A:
I would like to add you as the maintainer of this package, if you don't mind.

Kyure-A:
@mainrs

mainrs:
> The reason I have been making minor changes is probably because there is an error in the initExtra area of the shell:
> 
> ```nix
> error:
>        … while calling the 'head' builtin
> 
>          at /nix/store/qxf6anli54ij0q1sdlnlgx9hyl658a4v-source/lib/attrsets.nix:1575:11:
> 
>          1574|         || pred here (elemAt values 1) (head values) then
>          1575|           head values
>              |           ^
>          1576|         else
> 
>        … while evaluating the attribute 'value'
> 
>          at /nix/store/qxf6anli54ij0q1sdlnlgx9hyl658a4v-source/lib/modules.nix:821:9:
> 
>           820|     in warnDeprecation opt //
>           821|       { value = addErrorContext "while evaluating the option `${showOption loc}':" value;
>              |         ^
>           822|         inherit (res.defsFinal') highestPrio;
> 
>        (stack trace truncated; use '--show-trace' to show the full trace)
> 
>        error: cannot coerce a set to a string
> ```
> 
> I don't know how to deal with this error.

I think concatStrings does not handle the case when you pass it an empty list. And if both expressions are optional, that is what you get, because `mkIf` returns an empty list if the condition is false IIRC.

You can rely on the `lib.optionalString`. This is untested:

```nix
concatStringsSep "\n" [
  optionalString (cfg.settings != { })
    ''
      eval "$(sheldon source)"
    ''
  optionalString cfg.enableFishCompletions
    ''
      if test "$TERM" != "dumb"
         eval "$(${cmd} completions --shell=fish)"
      end
    ''
]
```

Kyure-A:
I have fixed this bug. Thanks for posting the code for reference.

Kyure-A:
It will probably work fine now.

Kyure-A:
@mainrs plz review it

modules/programs/sheldon.nix
```diff
@@ -0,0 +1,82 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.programs.sheldon;
+  tomlFormat = pkgs.formats.toml { };
+  sheldonCmd = "${config.home.profileDirectory}/bin/sheldon";
```
mainrs, line 8: What exactly is the difference between this and using `${cfg.package}/bin/sheldon`?

mainrs (commented):
Rest is LGTM.

onedaxter:
Hi @Kyure-A , @mainrs 
would really like to try out this package.
anything holding off despite just doing the merge ?
thank you

Kyure-A:
> Hi @Kyure-A , @mainrs would really like to try out this package. anything holding off despite just doing the merge ? thank you

 I completely missed @mainrs's comment. There's no particular reason.

modules/programs/sheldon.nix
Kyure-A: I wrote this based on a Nix expression for a certain package, so I'm not sure about the exact difference.

khaneliman: This doesn't seem right, we should be pointing at the configured package's binary.

modules/programs/sheldon.nix
```diff
@@ -0,0 +1,82 @@
...
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    xdg.configFile."sheldon/plugins.toml" = mkIf (cfg.settings != { }) {
+      source = tomlFormat.generate "sheldon-config" cfg.settings;
+    };
+
+    programs.bash.initExtra = ''
+      ${optionalString (cfg.settings != { }) ''
+        eval "$(sheldon source)"
```
khaneliman, line 51: Why are we relying on `PATH` here but using the `sheldonCmd` for the completions?

modules/programs/sheldon.nix
```diff
@@ -0,0 +1,82 @@
...
+      ''}
+      ${optionalString cfg.enableBashCompletions ''
+        if [[ $TERM != "dumb" ]]; then
```
khaneliman, line 54: What are these `$TERM != "dumb"` checks for?

snaeil:
@Kyure-A Are there any plans on continuing with your PR?

elanora96:
Hi I forked this PR and believe I implemented the changes suggested in the reviews [in this commit](https://github.com/elanora96/home-manager/commit/a9bd52ccaa05c7b9c2fe07f39593194313381c27). Honestly I'm a bit uncertain how to best contribute this without the ability to push to @Kyure-A 's branch. I don't want to end up sniping this and creating my own PR... but if there isn't a better way... I mean I do want Sheldon to be in Home Manager...

edit: The linked commit remains the most relevant one. After it I ran `./format` (my bad!) and then resolved the merge conflicts with master

(Off-topic, I like how the Copilot PR that references this PR ends up with Copilot just removing Sheldon lol)

khaneliman:
> Hi I forked this PR and believe I implemented the changes suggested in the reviews [in this commit](https://github.com/elanora96/home-manager/commit/a9bd52ccaa05c7b9c2fe07f39593194313381c27). Honestly I'm a bit uncertain how to best contribute this without the ability to push to @Kyure-A 's branch. I don't want to end up sniping this and creating my own PR... but if there isn't a better way... I mean I do want Sheldon to be in Home Manager...
> 
> edit: The linked commit remains the most relevant one. After it I ran `./format` (my bad!) and then resolved the merge conflicts with master
> 
> (Off-topic, I like how the Copilot PR that references this PR ends up with Copilot just removing Sheldon lol)

Just create a new PR that adds `Kyure-A` as a coauthor with the changes you made.

PR #5674, opened 2024-07-27
starship: use direct derivation binary path instead of referencing ~/nix-profile

tulilirockz:
### Description

Using the lib.getExe path from starship is better in this case because if your nix settings specify the "use-xdg-directories" variable, the default `$HOME/.nix-profile` path that is supposed to exist in the starshipCmd variable. It seems to be working fine on my system

### Checklist

Honestly since this is a single line change I dont know if all this makes sense?

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/starship.nix
+++ b/modules/programs/starship.nix
@@ -8,8 +8,7 @@ let
 
   tomlFormat = pkgs.formats.toml { };
 
-  starshipCmd = "${config.home.profileDirectory}/bin/starship";
-
+  starshipCmd = "${lib.getExe pkgs.starship}";
 in {
   meta.maintainers = [ ];
 


rycee:
Thanks for the contribution! The existing behavior is intentional, however, since referring to the executable directly will cause errors when you have a shell that lives long enough for the path to be garbage collected.

I think the only other possible solution that is somewhat reliable would be to simply do

``` bash
eval "$(starship init bash --print-full-init)"
```

and hope that `PATH` is set up such that the correct starship binary is executed. I'm pretty skeptical about this, though.

I think the most reliable is to continue with the existing setup, which unfortunately requires you to manually set `home.profileDirectory`, or let the Nix module do it by adding something like

``` nix
nix = {
  package = pkgs.nix;
  settings.use-xdg-base-directories = true;
};
```

to your configuration.

PR #5677, opened 2024-07-27
programs.wlr-which-key: init

### LINKED ISSUE #4396: wlr-which-key
I'd love to see https://github.com/MaxVerevkin/wlr-which-key.

It's in nixpkgs already.

minijackson:
### Description

wlr-which-key is a keymap manager for wlroots-based compositors, inspired by which-key.nvim.

closes #4396

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@mightyiam, @hervyqa any of you interested in becoming co-maintainer of this module?

### DIFF
--- a/modules/programs/wlr-which-key.nix
+++ b/modules/programs/wlr-which-key.nix
@@ -1,0 +1,104 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+let
+  cfg = config.programs.wlr-which-key;
+  yamlFormat = pkgs.formats.yaml { };
+in
+{
+  meta.maintainers = with lib.maintainers; [
+    LilleAila
+    mightyiam
+    minijackson
+  ];
+
+  options.programs.wlr-which-key = {
+    enable = lib.mkEnableOption "wlr-which-key";
+
+    package = lib.mkPackageOption pkgs "wlr-which-key" { nullable = true; };
+
+    commonSettings = lib.mkOption {
+      type = lib.types.submodule { freeformType = yamlFormat.type; };
+
+      default = { };
+      example = {
+        anchor = "center";
+        background = "#282828d0";
+        border = "#8ec07c";
+        color = "#fbf1c7";
+      };
+
+      description = ''
+        Settings to be applied to every configuration under the `configs` option.
+      '';
+    };
+
+    configs = lib.mkOption {
+      type = lib.types.attrsOf (
+        lib.types.submodule {
+          freeformType = yamlFormat.type;
+
+          options.menu = lib.mkOption {
+            description = "The various menus to display";
+            type = lib.types.attrs;
+          };
+
+          config = cfg.commonSettings;
+        }
+      );
+
+      default = { };
+      example.config = {
+        anchor = "center";
+        menu = {
+          p = {
+            desc = "Power";
+            submenu = {
+              o = {
+                cmd = "poweroff";
+                desc = "Off";
+              };
+              r = {
+                cmd = "reboot";
+                desc = "Reboot";
+              };
+              s = {
+                cmd = "systemctl suspend";
+                desc = "Sleep";
+              };
+            };
+          };
+        };
+      };
+
+      description = ''
+        Various configurations for wlr-which-key.
+
+        Each configuration `configs.''${name}` is installed into
+        `~/.config/wlr-which-key/''${name}.yaml`.
+
+        This enables you to run: `wlr-which-key ''${name}`.
+
+        To set the default menu, use `programs.wlr-which-key.configs.config`.
+
+        For more information on available options, see:
+        <https://github.com/MaxVerevkin/wlr-which-key/>
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    home.packages = lib.mkIf (cfg.package != null) [ cfg.package ];
+
+    xdg.configFile = lib.mapAttrs' (
+      name: value:
+      lib.nameValuePair "wlr-which-key/${name}.yaml" {
+        source = yamlFormat.generate "wlr-which-key-${name}.yaml" value;
+      }
+    ) cfg.configs;
+  };
+}
--- a/modules/misc/news/2025-04-20_16-43-56.nix
+++ b/modules/misc/news/2025-04-20_16-43-56.nix
@@ -1,0 +1,12 @@
+{
+  time = "2025-04-20T21:43:56+00:00";
+  condition = true;
+  message = ''
+    A new module is available: 'programs.wlr-which-key'.
+
+    wlr-which-key is a keymap manager for wlroots-based compositors,
+    inspired by which-key.nvim.
+
+    See https://github.com/MaxVerevkin/wlr-which-key for more.
+  '';
+}
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -295,6 +295,7 @@ let
       ./programs/waybar.nix
       ./programs/wezterm.nix
       ./programs/wlogout.nix
+      ./programs/wlr-which-key.nix
       ./programs/wofi.nix
       ./programs/xmobar.nix
       ./programs/xplr.nix
--- a/tests/default.nix
+++ b/tests/default.nix
@@ -473,6 +473,7 @@ import nmtSrc {
       ./modules/programs/wallust
       ./modules/programs/watson
       ./modules/programs/wezterm
+      ./modules/programs/wlr-which-key
       ./modules/programs/yazi
       ./modules/programs/zed-editor
       ./modules/programs/zellij
--- a/tests/modules/programs/wlr-which-key/default.nix
+++ b/tests/modules/programs/wlr-which-key/default.nix
@@ -1,0 +1,1 @@
+{ wlr-which-key = ./wlr-which-key.nix; }
--- a/tests/modules/programs/wlr-which-key/expected-config.yaml
+++ b/tests/modules/programs/wlr-which-key/expected-config.yaml
@@ -1,0 +1,68 @@
+anchor: center
+background: '#282828d0'
+border: '#8ec07c'
+border_width: 2
+color: '#fbf1c7'
+corner_r: 10
+font: JetBrainsMono Nerd Font 12
+margin_bottom: 0
+margin_left: 0
+margin_right: 0
+margin_top: 0
+menu:
+  l:
+    desc: Laptop Screen
+    submenu:
+      s:
+        desc: Scale
+        submenu:
+          '1':
+            cmd: wlr-randr --output eDP-1 --scale 1
+            desc: Set Scale to 1.0
+          '2':
+            cmd: wlr-randr --output eDP-1 --scale 1.1
+            desc: Set Scale to 1.1
+          '3':
+            cmd: wlr-randr --output eDP-1 --scale 1.2
+            desc: Set Scale to 1.2
+          '4':
+            cmd: wlr-randr --output eDP-1 --scale 1.3
+            desc: Set Scale to 1.3
+      t:
+        cmd: toggle-laptop-display.sh
+        desc: Toggle On/Off
+  p:
+    desc: Power
+    submenu:
+      o:
+        cmd: poweroff
+        desc: 'Off'
+      r:
+        cmd: reboot
+        desc: Reboot
+      s:
+        cmd: systemctl suspend
+        desc: Sleep
+  t:
+    desc: Theme
+    submenu:
+      d:
+        cmd: dark-theme on
+        desc: Dark
+      l:
+        cmd: dark-theme off
+        desc: Light
+      t:
+        cmd: dark-theme toggle
+        desc: Toggle
+  w:
+    desc: WiFi
+    submenu:
+      c:
+        cmd: kitty --class nmtui-connect nmtui-connect
+        desc: Connections
+      t:
+        cmd: wifi_toggle.sh
+        desc: Toggle
+padding: 15
+separator: ' ➜ '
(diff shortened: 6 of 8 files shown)


mightyiam:
Sure, I can be a co-maintainer.

LilleAila:
How about doing something like this? (taken from my hm config)
```nix
{
  config,
  lib,
  pkgs,
  ...
}:
let
  cfg = config.programs.wlr-which-key;
in
{
  options.settings.wlr-which-key = {
    enable = lib.mkEnableOption "wlr-which-key";
    settings = lib.mkOption {
      type = yamlFormat.type;
      default = { };
      example =
        lib.literalExpression # nix
          ''
            {
              font = "JetBrainsMono Nerd Font 12";
              background = "#282828cc";
              color = "#ebdbb2";
              border = "#3c3836";
              border_width = 2;
              corner_r = 12;
              padding = 16;
              separator = " -> ";

              anchor = "center"; # One of center, left, right, top, bottom, bottom-left, top-left, etc.
              margin_right = 0; # Only relevant when not center
              margin_bottom = 0;
              margin_left = 0;
              margin_top = 0;
            }
          '';
    };
    menus = lib.mkOption {
      type = lib.types.attrsOf yamlFormat.type;
      default = { };
      example =
        lib.literalExpression # nix
          ''
            {
              default = {
                w = {
                  desc = "WiFi";
                  submenu = {
                    t = {
                      desc = "Toggle";
                      cmd = "rfkill toggle wifi";
                    };
                    c = {
                      desc = "Connect";
                      cmd = "kitty -e nmtui-connect";
                    };
                  };
                };
                l = {
                  desc = "Lock";
                  cmd = "pidof swaylock || swaylock";
                };
              };
            }
          '';
    };
  };

    home.packages = [ pkgs.wlr-which-key ];
    xdg.configFile = lib.attrsets.mapAttrs' (
      name: menu:
      let
        filename = if name == "default" then "config" else name;
      in
      lib.attrsets.nameValuePair "wlr-which-key/${filename}.yaml" {
        source = (yamlFormat.generate "${filename}.yaml" (cfg.settings // { inherit menu; }));
      }
    ) cfg.menus;
  };
}
```

From `wlr-which-key --help`, it says that different menus can be opened from different config files. This does that, and keeps the settings the same between them while allowing for defining different menus, e.g. `wlr-which-key power`, `wlr-which-key firefox` etc.

LilleAila:
> @LilleAila aha! Hadn't thought about that particular usecase, thanks! I've opted for a slightly different implementation. Let me know what you think.

LGTM! But maybe `commonSettings` should be renamed to just `settings`, to comply with the naming scheme of most other modules. I guess i could also be a co-maintainer.

LilleAila (approved):

minijackson:
> But maybe `commonSettings` should be renamed to just `settings`, to comply with the naming scheme of most other modules.

I'm not completely sure, this `settings` option doesn't map to a config file. If some user just does `programs.wlr-which-key.settings = { ... }`, it wouldn't do anything.

> I guess i could also be a co-maintainer.

Awesome! I'll add you

minijackson:
Hmmm, I've tried adding you @LilleAila but home-manager doesn't seem to know you:

```
       error: undefined variable 'LilleAila'

       at /home/minijackson/Documents/Dev/dev/github.com/nix-community/home-manager/modules/programs/wlr-which-key.nix:7:46:

            6| in {
            7|   meta.maintainers = with lib.maintainers; [ LilleAila mightyiam minijackson ];
             |                                              ^
            8|
```

LilleAila:
> Hmmm, I've tried adding you @LilleAila but home-manager doesn't seem to know you:
> 
> ```
>        error: undefined variable 'LilleAila'
> 
>        at /home/minijackson/Documents/Dev/dev/github.com/nix-community/home-manager/modules/programs/wlr-which-key.nix:7:46:
> 
>             6| in {
>             7|   meta.maintainers = with lib.maintainers; [ LilleAila mightyiam minijackson ];
>              |                                              ^
>             8|
> ```

Right, i forgot the PR adding me isn’t merget yet. I think the one in nixpkgs maintainers is though, otherwise it’s okay not to be added.

minijackson:
I think home-manager can use the maintainer list from nixpkgs. If you've added yourself in nixpkgs recently, might be that home-manager's nixpkgs is not recent enough.

minijackson:
@LilleAila after investigating a bit more, I think this is a "me" issue, since the way I tested it was using `import <nixpkgs>`. I've added you, and I guess we'll see if this goes well!

modules/programs/wlr-which-key.nix
```diff
@@ -0,0 +1,88 @@
+{ config, lib, pkgs, ... }:
+
+let
+  cfg = config.programs.wlr-which-key;
+  yamlFormat = pkgs.formats.yaml { };
+in {
+  meta.maintainers = with lib.maintainers; [ LilleAila mightyiam minijackson ];
+
+  options.programs.wlr-which-key = {
+    enable = lib.mkEnableOption "wlr-which-key configuration";
```
IldenH, line 10:
```suggestion
    enable = lib.mkEnableOption "wlr-which-key";
```

modules/programs/wlr-which-key.nix
```diff
@@ -0,0 +1,88 @@
...
+
+      description = ''
+        Various configurations for wlr-which-key.
+
+        Each configuration `configs.''${name}` is installed into
+        `~/.config/wlr-which-key/''${name}.yaml`.
+
+        This enables you to run: `wlr-which-key ''${name}`
+
+        The default configuration is named 'config'.
```
IldenH, line 72: This line is confusing and I think it is unnecessary as those who only use one config will just run wlr-which-key without anything else.

modules/programs/wlr-which-key.nix
```diff
@@ -0,0 +1,88 @@
...
+        border = "#8ec07c";
+        color = "#fbf1c7";
+      };
+
+      description = ''
+        Settings to be applied to every configuration under the `configs` option.
+      '';
+    };
+
+    configs = lib.mkOption {
```
IldenH, line 28:
```suggestion
    menus = lib.mkOption {
```

modules/programs/wlr-which-key.nix
```diff
@@ -0,0 +1,88 @@
...
+        };
+      };
+
+      description = ''
+        Various configurations for wlr-which-key.
+
+        Each configuration `configs.''${name}` is installed into
+        `~/.config/wlr-which-key/''${name}.yaml`.
+
+        This enables you to run: `wlr-which-key ''${name}`
```
IldenH, line 70:
```suggestion
        This enables you to run: `wlr-which-key ''${name}`.
```

modules/programs/wlr-which-key.nix
```diff
@@ -0,0 +1,88 @@
...
+let
+  cfg = config.programs.wlr-which-key;
+  yamlFormat = pkgs.formats.yaml { };
+in {
+  meta.maintainers = with lib.maintainers; [ LilleAila mightyiam minijackson ];
+
+  options.programs.wlr-which-key = {
+    enable = lib.mkEnableOption "wlr-which-key configuration";
+
+    commonSettings = lib.mkOption {
```
IldenH, line 12:
```suggestion
    settings = lib.mkOption {
```

modules/programs/wlr-which-key.nix
```diff
@@ -0,0 +1,88 @@
...
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    home.packages = [ pkgs.wlr-which-key ];
+
+    xdg.configFile = lib.mapAttrs' (name: value:
+      lib.nameValuePair "wlr-which-key/${name}.yaml" {
+        source = yamlFormat.generate "wlr-which-key-${name}.yaml" value;
+      }) cfg.configs;
```
IldenH, line 86: I think a config.yaml should be generated as well. Such that `settings.menu` works.

modules/programs/wlr-which-key.nix
```diff
@@ -0,0 +1,88 @@
...
+      default = { };
+      example = {
+        anchor = "center";
+        background = "#282828d0";
+        border = "#8ec07c";
+        color = "#fbf1c7";
+      };
+
+      description = ''
+        Settings to be applied to every configuration under the `configs` option.
```
IldenH, line 24:
```suggestion
        Settings to be applied to every menu specified under the `menus` option.
```

IldenH (changes requested):
I think allowing users to ignore the `configs` or `menus` option as I've renamed it to and only use the normal `settings` is good.

minijackson:
@IldenH I'm not sold on making `settings` both the configuration for the default file, and also apply them to the other `menus`. I think it's best to keep one option for common settings, and one option that map to files.

I've reworded the sentence explaining the default file, please tell me if it's still confusing.

I also prefer keeping `configs` since I feel that with `menus` it can be confused with the `menu` setting inside the config.

I agree that `programs.wlr-which-key.configs.config` feels clunky to write, but the name `config.yaml` is upstream's choice and I don't think we should hide it.

Catmaniscatlord:
This PR seems stale, but I have some feedback that might move things forward.

Packing the menus in a single variable and keeping the settings separate, unless you choose to override the settings for a specific menu is a complicated and confusing layout to configure this package. It seems that this was chosen so that settings can be shared between different menus. The user can do this in their own config by creating a local variable for the config ` let myStyle = { ... }; in ` and then using unions to override changed values with ` style = myStyle // { ... }`.

Putting the config that gets called by default into the `config` variable is intuitive. Also, any extra configs being in `extraConfigs` would make sense that you need to pass an argument to call those configs explicitly, without the help of a description.

``` nix
{
  programs.wlr-which-key = 
  let
   myStyle = {
    anchor = "center";
    background = "#282828d0";
    border = "#8ec07c";
    border_width = 2;
    color = "#fbf1c7";
    corner_r = 10;
    font = "JetBrainsMono Nerd Font 12";
    margin_bottom = 0;
    margin_left = 0;
    margin_right = 0;
    margin_top = 0;
    padding = 15;
    separator = " ➜ ";
   };
  in {
  enable = true;
  
  # Default Style
  config = {
    style = myStyle
    #Default config
    menu = {
      ...
    };
  };

  # Extra Configs if the user desires
  extraConfigs = {
    firefox = {
      style = myStyle // { border = "#ff0000"; };
      menu = {
         ...
      };
    };
  };
};
```

hervyqa:
no stale

khaneliman:
> This PR seems stale, but I have some feedback that might move things forward.
> 
> Packing the menus in a single variable and keeping the settings separate, unless you choose to override the settings for a specific menu is a complicated and confusing layout to configure this package. It seems that this was chosen so that settings can be shared between different menus. The user can do this in their own config by creating a local variable for the config `let myStyle = { ... }; in` and then using unions to override changed values with ` style = myStyle // { ... }`.
> 
> Putting the config that gets called by default into the `config` variable is intuitive. Also, any extra configs being in `extraConfigs` would make sense that you need to pass an argument to call those configs explicitly, without the help of a description.
> 
> ```nix
> {
>   programs.wlr-which-key = 
>   let
>    myStyle = {
>     anchor = "center";
>     background = "#282828d0";
>     border = "#8ec07c";
>     border_width = 2;
>     color = "#fbf1c7";
>     corner_r = 10;
>     font = "JetBrainsMono Nerd Font 12";
>     margin_bottom = 0;
>     margin_left = 0;
>     margin_right = 0;
>     margin_top = 0;
>     padding = 15;
>     separator = " ➜ ";
>    };
>   in {
>   enable = true;
>   
>   # Default config
>   config = {
>     style = myStyle
>     menu = {
>       ...
>     };
>   };
> 
>   # Extra Configs if the user desires
>   extraConfigs = {
>     firefox = {
>       style = myStyle // { border = "#ff0000"; };
>       menu = {
>          ...
>       };
>     };
>   };
> };
> ```

Yeah, I was reviewing module a bit and this did stand out as a weird UX. We should strive for as much of a nix -> config format transparency as possible. If the desire is just to deduplicate code in a user's config, they can leverage variables for that.

minijackson:
Sorry all, I've lost interest in becoming maintainer of this module. If anyone wants to pick this PR backup, please feel free!

mightyiam:
Sorry, a bit unrelated; is there an alternative to this that is already in home-manager?

rachitvrma:
```nix
    {
      config,
      lib,
      pkgs,
      ...
    }:
    let
      inherit (lib) mkIf mkOption types;
      cfg = config.programs.wlr-which-key;
    in
    {
      options.programs.wlr-which-key = {
        enable = lib.mkEnableOption "wlr-which-key";

        package = lib.mkPackageOption pkgs "wlr-which-key" { };

        settings = mkOption {
          inherit (yamlFormat) type;
          default = { };
          description = "Main wlr-which-key configuration written to config.yml.";
        };

        inheritSettings = mkOption {
          type = types.bool;
          default = false;
          description = "Whether extra menus inherit top-level settings from settings.";
        };

        extraMenus = mkOption {
          type = types.attrsOf yamlFormat.type;
          default = { };
          description = "Additional named menus, each written to wlr-which-key/<name>.yml.";
        };
      };

      config = mkIf cfg.enable {

        xdg.configFile = lib.mkMerge [
          (mkIf (cfg.settings != { }) {
            "wlr-which-key/config.yml".source = yamlFormat.generate "wlr-which-key-main" cfg.settings;
          })
          (lib.mapAttrs' (
            name: value:
            lib.nameValuePair "wlr-which-key/${name}.yml" {
              source = yamlFormat.generate "wlr-which-key-${name}" (
                lib.optionalAttrs cfg.inheritSettings cfg.settings // value
              );
            }
          ) cfg.extraMenus)
        ];
      };
    }
```

How about this? We can have an extra boolean option to whether inherit the main settings or not.

PR #5683, opened 2024-07-29
home-manager: ignore hostname -f lookup errors

### LINKED ISSUE #5665: bug: "hostname: Unknown host" error

### Issue description

Note: migrated from https://github.com/nix-community/home-manager/pull/5462#issuecomment-2247723873.

After #5462, if the hostname is not set in `/etc/hosts`, the fqdn command `hostname -f` may fail with:
```console
$ hostname -f
hostname: Unknown host
```
This shows up in the `home-manager switch` printout and is mildly annoying. Could it be changed to:
```bash
hostname -f 2>/dev/null
```
to suppress this inconsequential warning? Thank you!

### Maintainer CC

@gauravjuvekar

### System information

```markdown
$ nix run nixpkgs#nix-info -- -m                               
 - system: `"x86_64-linux"`
 - host os: `Linux 6.6.32-1-MANJARO, Manjaro Linux, noversion, rolling`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.18.5`
 - nixpkgs: `/home/bryan/.nix-defexpr/channels/nixpkgs`
```

gauravjuvekar:
### Description

`hostname -f` could fail depending on the resolver. Discard any stderr
and test for the exit status before using the value for flake attribute
lookup.

I was unable to repro the exact bad exit status in #5665.
With
  - nscd disabled,
  - nsswitch.conf pointing to 'files',
  - hostname entry removed from /etc/hosts
`hostname -f` from inetutils-2.5 fell back to showing just the nodename
from `uname(2)`. Injecting an empty string into the
`(struct utsname).nodename` field of `uname(2)` using strace still
exited with empty output and 0 exit-status.

Fixes #5665

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/home-manager/home-manager
+++ b/home-manager/home-manager
@@ -198,9 +198,19 @@ function setFlakeAttribute() {
                 ;;
             *)
                 local name="$USER"
+
+                local hostnameArray=()
+                # FQDN lookup can fail depending on the resolver.
+                local fqdn
+                fqdn="$(hostname -f 2> /dev/null)"
+                if [[ $? -eq 0 ]]; then
+                    hostnameArray+=( "$USER@$fqdn"  )
+                fi
                 # Check FQDN, long, and short hostnames; long first to preserve
                 # pre-existing behaviour in case both happen to be defined.
-                for n in "$USER@$(hostname -f)" "$USER@$(hostname)" "$USER@$(hostname -s)"; do
+                hostnameArray+=( "$USER@$(hostname)" "$USER@$(hostname -s)" )
+
+                for n in "${hostnameArray[@]}"; do
                     if [[ "$(nix eval "$flake#homeConfigurations" --apply "x: x ? \"$n\"")" == "true" ]]; then
                         name="$n"
                         if [[ -v VERBOSE ]]; then


bryango:
It might depend on which `hostname` we use. In my case,
- `hostname` from `inetutils` works fine, but
- `hostname` from `net-tools` (also `nixpkgs#hostname`, also the one used by home-manager) fails

```console
$ nix run nixpkgs#hostname -- -f
hostname: Unknown host

$ nix shell nixpkgs#inetutils --command hostname -f
btrsamsung
```

rycee:
Pretty wild how difficult it is to reliably get the hostname. Thanks for the contribution, I've merged it to master now!

I'm not sure its worth switching to the net-tools `hostname`, maybe it has other corner cases?

bryango:
Thank you very much! :fireworks: 

> I'm not sure its worth switching to the inetutils `hostname`, maybe it has other corner cases?

For now, I think it's not necessary. As always, the GNU variant is more powerful but more bloated and less portable.

gauravjuvekar:
True, I think "it works" is fine for now.

Eventually, we might want to think about using specific versions of utilities from a nix store instead of relying on those in the invocation shell environment. I'm not sure how well that would work for stand alone home-manager but probably worth exploring if we see similar issues for other utilities.

rycee:
@gauravjuvekar The home-manager tool should already use a `hostname` tool provided by Nixpkgs, never the host one. See the `unixtools.hostname` input in https://github.com/nix-community/home-manager/blob/4fcd54df7cbb1d79cbe81209909ee8514d6b17a4/home-manager/default.nix#L31-L43

The tool provided is different for Linux and Darwin, though. So if the GNU inetutils hostname behaves the same on Linux and Darwin, then that would definitely make it interesting to use instead of `unixtools.hostname`.

bryango:
Ah we are already using a bunch of GNU tools, then indeed we might as well switch to `inetutils` from GNU. One thing I was worrying about is the availability across platforms, but it should be fine according to the hydra build log:
- [x] https://hydra.nixos.org/job/nixpkgs/trunk/inetutils.aarch64-linux
- [x] https://hydra.nixos.org/job/nixpkgs/trunk/inetutils.aarch64-darwin
- [x] https://hydra.nixos.org/job/nixpkgs/trunk/inetutils.x86_64-darwin
- [x] https://hydra.nixos.org/job/nixpkgs/trunk/inetutils.x86_64-linux

The closure size difference should be insignificant since dependencies can be shared between GNU utilities.

bryango:
Ah sorry for not testing this more thoroughly, I was too busy back then... Apparently, this is not working due to `set -e`, see #5711. I have opened #5738 to fix this.

PR #5685, opened 2024-07-29
firefox: read the appName for search engine disclaimer from package

kkoniuszy:
### Description

Using a fixed application name in the salt for the search engine name hash can break with minor branding changes. For example, LibreWolf 127 used the application name "LibreWolf", but in version 128 it's "Firefox".

The proper name can be found in about:support -> Application Basics.

Because it doesn't have to be related to the product name visible in most of the browser (for example in the window title and help menus), we shouldn't rely on cfg.name for that.

The application name can be read from lib/*/application.ini and we can use that if the browser was installed via home-manager. If not, we can fall back to cfg.name.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee @brckd @kira-bruneau

### DIFF
--- a/modules/programs/firefox/mkFirefoxModule.nix
+++ b/modules/programs/firefox/mkFirefoxModule.nix
@@ -553,6 +553,7 @@ in {
                 inherit (args) config;
                 inherit lib pkgs;
                 appName = cfg.name;
+                package = cfg.finalPackage;
                 modulePath = modulePath ++ [ "profiles" name "search" ];
                 profilePath = config.path;
               });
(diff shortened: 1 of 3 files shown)


kira-bruneau (approved):
Thanks @kkoniuszy!

donovanglover (approved):
This fixed changing the default search engine for me with LibreWolf and https://github.com/nix-community/home-manager/pull/5684. Looking forward to the merge!

MithicSpirit (approved):
Been using this and #5684 for a couple of days now, and it's been working very well.

donovanglover:
Merge conflict is due to https://github.com/nix-community/home-manager/pull/5697

kkoniuszy:
Conflict resolved

teto:
we got kira-bruneau's approval so it's good to go, just need a rebase to fix ci

rycee:

PR #5689, opened 2024-07-30
neovim: add `extraPrefixedPackages` option

carschandler:
### Description

`extraPackages` is nice to have, but there are scenarios where a package can already exist in `PATH` before it is wrapped in, so the `--suffix` option that `extraPackages` uses will have no effect. It used to use `--prefix`, and I'd like to bring back the option to choose that.

### Checklist

- [x] Code tested through `nix-shell --pure tests -A run.all` or `nix develop --ignore-environment .#all` using Flakes. **Hanging on `building pubkey.txt: Warning: Problem : timeout` but I don't think this is related to me.**

- [ ] Test cases updated/added. **No previous tests existed for `extraPackages`, so I have not yet added any for `extraPrefixedPackages`, but am happy to add a test for both if needed.**

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

Missing?

@teto @misumisumi

### DIFF
--- a/modules/programs/neovim.nix
+++ b/modules/programs/neovim.nix
@@ -65,8 +65,10 @@ let
   luaPackages = cfg.finalPackage.unwrapped.lua.pkgs;
   resolvedExtraLuaPackages = cfg.extraLuaPackages luaPackages;
 
-  extraMakeWrapperArgs = lib.optionalString (cfg.extraPackages != [ ])
-    ''--suffix PATH : "${lib.makeBinPath cfg.extraPackages}"'';
+  extraMakeWrapperArgs = lib.optionalString (cfg.extraPackages != [ ]) ''
+    --suffix PATH : "${lib.makeBinPath cfg.extraPackages}"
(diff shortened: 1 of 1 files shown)


teto:
you can already use `programs.neovim.extraWrapperArgs` for this

carschandler:
Guess that's fair. Just cleaner and better for discoverability to have an option for it, but I can understand not wanting more bloat. Feel free to close if desired.

teto:
This feature is common to much software, not only neovim so I think it's better to show how to achieve this by example rather than adding cruft. I've introduced a wiki entry to https://github.com/nix-community/home-manager/wiki/Neovim as this is a recurrent demand. It could probably live in `extraPackages` documentation as well. My wiki entry is not explicit enough but I am in a hurry and just wanted to get the ball rolling. Ty for your effort. Feel free to edit the wiki

carschandler:
I think this is a good alternative. Thanks!

PR #5697, opened 2024-08-01
thunderbird: support setting search engines

kira-bruneau:
### Description

- Splits firefox search engine configuration into a separate submodule so it can be reused in thunderbird
- Removes myself maintainer for all of firefox, and adds myself as a maintainer for just the search engine submodule

### Checklist

  Submodule already tested in `firefox-profile-search` tests.

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

Firefox:
@rycee @brckd

Thunderbird:
@d-dervishi

### DIFF
--- a/modules/programs/firefox/profiles/search.nix
+++ b/modules/programs/firefox/profiles/search.nix
@@ -1,0 +1,251 @@
+{ config, lib, pkgs, appName, modulePath, profilePath }:
+
+with lib;
+
+let
+  jsonFormat = pkgs.formats.json { };
+
+  # Map of nice field names to internal field names.
+  # This is intended to be exhaustive and should be
+  # updated at every version bump.
+  internalFieldNames = (genAttrs [
+    "name"
+    "isAppProvided"
+    "loadPath"
+    "hasPreferredIcon"
+    "updateInterval"
+    "updateURL"
+    "iconUpdateURL"
+    "iconURL"
+    "iconMapObj"
+    "metaData"
+    "orderHint"
+    "definedAliases"
+    "urls"
+  ] (name: "_${name}")) // {
+    searchForm = "__searchForm";
+  };
+
+  processCustomEngineInput = input:
+    (removeAttrs input [ "icon" ]) // optionalAttrs (input ? icon) {
+      # Convenience to specify absolute path to icon
+      iconURL = "file://${input.icon}";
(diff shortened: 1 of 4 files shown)


kira-bruneau:
I was also thinking of splitting out `user.js` & `chrome` from firefox so changes there would also be shared in thunderbird, but decided to keep this focused on search to simplify the review.

kira-bruneau:
I also just pushed a change that also organizes the firefox tests by submodule.

kira-bruneau:
Actually, I just decided to split off test refactoring into it's own PR, because the change is easy to treat separately: #5698

brckd (approved):
Thanks @kira-bruneau! No issues from my side, since it doesn't affect the functionality of the Firefox module. The formatting only conflicts with #5685 so far, which would become your responsibility when you become the maintainer of this module.

modules/programs/firefox/profiles/search.nix
```diff
@@ -0,0 +1,251 @@
...
+
+  meta.maintainers = with maintainers; [ kira-bruneau ];
+
+  options = {
+    enable = mkOption {
+      type = with types; bool;
+      default = config.default != null || config.privateDefault != null
+        || config.order != [ ] || config.engines != { };
+      internal = true;
+    };
```
kira-bruneau, line 151: I added this internal `enable` option to make the implicit enable behaviour reusable, but I think ideally users should be required to set `enable` explicitly. I just didn't want to introduce any breaking changes in this PR.

I plan on folding `force` into `enable` in a follow-up PR.

khaneliman:
Getting this after this commit.

```bash
error: The option `home-manager.users.khaneliman.home.file.".mozilla/firefox/khaneliman/search.json.mozlz4".enable' has conflicting definition values:
       - In `/nix/store/i9vkk0yd40vsvn73a1rfi60p2p0dhc8m-source/modules/programs/firefox.nix': true
       - In `/nix/store/i9vkk0yd40vsvn73a1rfi60p2p0dhc8m-source/modules/programs/firefox.nix': false
```

khaneliman:
Hmm.. I can double check my config.. but I didn't spend too much time looking originally because both files referenced were an internal home-manager module.

khaneliman:
Appears to be issue with multiple profiles. One profile doesn't set the search option, the other uses it. This conflict is thrown.

PR #5699, opened 2024-08-02
home-cursor: add hyprcursor support

arminius-smh:
add the option to enable hyprcursor support by setting the relevant environment variables

### Description

"hyprcursor is a new cursor theme format that has many advantages over the widely used xcursor." - https://wiki.hyprland.org/Hypr-Ecosystem/hyprcursor/

This pull request is for adding an option to home-cursor to set the two variables needed for hyprcursor to set a theme, `HYPRCURSOR_THEME` and `HYPRCURSOR_SIZE`

I think the specific size option should be included, as that is specifically one benefit of hyprcursors, since all sizes are supported and not just a few common ones like in xcursor, but if this is not desired I can just remove it (I also don't know if this is the best way to make the size optional so that if it is not set to just use the size from xcursor)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@league

### DIFF
--- a/modules/config/home-cursor.nix
+++ b/modules/config/home-cursor.nix
@@ -45,6 +45,17 @@ let
           gtk config generation for {option}`home.pointerCursor`
         '';
       };
+
+      hyprcursor = {
+        enable = mkEnableOption "hyprcursor config generation";
+
+        size = mkOption {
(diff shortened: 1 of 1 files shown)


rycee:

PR #5702, opened 2024-08-03
Enable nix-direnv option in direnv by default

jla2000:
### Description

Enabling this option does speed up the usage of nix flakes together with direnv a lot.
I think it should be enabled by default as things don't change and it's just faster than the builtin `use flake`.
Most users probably just want it to be fast out of the box, without digging in and finding this option and enabling it manually.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/programs/direnv.nix
+++ b/modules/programs/direnv.nix
@@ -8,7 +8,8 @@ let
 
   tomlFormat = pkgs.formats.toml { };
 
-in {
+in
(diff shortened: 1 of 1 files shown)


jla2000:
note: the change in line 11/12 was introduced by the call to `./format`

PR #5705, opened 2024-08-04
Repeat kdl nodes

shimunn:
Supercedes #5272 

> Replaces the list attr -> KDL conversion logic with a more flexible approach, allowing for multiple nodes with the same name in a scope. This unfortunately also breaks the existing JSON-in-KDL semantics in favor of ergonomics. As far as I can tell though, zellij is the only program using it, and it doesn't accept JiK anyway.
> 
> For example, the following KDL was previously impossible to generate, since nix attrs were mapped 1:1 to KDL nodes:
> 
> ```
> resize {
> 	bind "k" "Up" { Resize "Increase Up"; }
> 	bind "j" "Down" { Resize "Increase Down"; }
> }
> ```
> 
> Now, this can be achieved with the nix expression:
> 
> ```nix
> resize.bind = [
> 	{ _args = ["k" "Up"]; Resize = "Increase Up"; }
> 	{ _args = ["j" "Down"]; Resize = "Increase Down"; }
> ];
> ```
> 
> which would previously have generated the not-very-useful:
> 
> ```
> resize {
> 	bind {
> 		- "k" "Up" { Resize "Increase Up"; }
> 		- "j" "Down" { Resize "Increase Down"; }
> 	}
> }
> ```
> 
> which, in turn, can now be generated via:
> 
> ```nix
> resize.bind."-" = [
> 	{ _args = ["k" "Up"]; Resize = "Increase Up"; }
> 	{ _args = ["j" "Down"]; Resize = "Increase Down"; }
> ];
> ```
> 
> ### Description
> ### Checklist
> 
>     * [ ]  ~Change is backwards compatible.~
>       It's explicitly not. I'm not sure that the previous behavior that's being broken was actually usable, so it doesn't seem worth keeping intact at the cost of ergonomics.
> 
>     * [x]  Code formatted with `./format`.
> 
>     * [x]  Code tested through `nix-shell --pure tests -A run.all` or `nix develop --ignore-environment .#all` using Flakes.
> 
>     * [x]  Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).
> 
>     * [x]  Commit messages are formatted like
>       ```
>       {component}: {description}
>       
>       {long description}
>       ```
>       
>       
>           
>             
>           
>       
>             
>           
>       
>           
>         
>       See [CONTRIBUTING](https://nix-community.github.io/home-manager/#sec-commit-style) for more information and [recent commit messages](https://github.com/nix-community/home-manager/commits/master) for examples.
> 
> 
> #### Maintainer CC
> 
> @h7x4 @mainrs


mainrs:
Could you elaborate: what changes did you do and why? Please also give examples. Thank you!

shimunn:
Added the tests you have requested in the original PR

mainrs:
@shimunn why did you force push changes for Dunst and fastfetch onto this PR? Was it an accident?

shimunn:
Rebased from master

shimunn:
Is there actually any interest in getting this merged?

budimanjojo:
I think another problem with `lib.generators.toKDL` is not allowing type of `attrset` inside `_args`. Because right now there's no way to have ordered struct.
For example:
```nix
keybinds = {
  RenameTab = [
    {
      bind = {
        _args = [ "Esc" ];
        UndoRenameTab = [];
        SwitchToMode = "normal";
      };
    }
  ];
}
```
With this PR will result in:
```kdl
keybinds {
  RenameTab {
    bind "Esc" {
      SwitchToMode "normal"
      UndoRenameTab 
    }
  }
}
```

Notice that `SwitchToMode` is now above `UndoRenameTab`? This will cause problem.
I think the logical way to fix this is to allow `attrset` inside `_args` so we can do it like so:
```nix
keybinds = {
  RenameTab = [
    {
      bind = {
        _args = [
          "Esc"
          { UndoRenameTab = []; }
          { SwitchToMode = "normal"; }
        ];
      };
    }
  ];
}
```
And the expected result is this:
```kdl
keybinds {
  RenameTab {
    bind "Esc" {
      UndoRenameTab 
      SwitchToMode "normal"
    }
  }
}
```
Not sure if this is easily doable though as I'm not really understanding the current `toKDL` function even after reading it many times :cry: .

PR #5706, opened 2024-08-04
Fix zellij settings parsing to KDL

ymnejmi:
### Description

fixed zellij settings by adding required semicolons when parsing toKDL (semicolons are optional in [kdl spec](https://github.com/kdl-org/kdl/blob/main/SPEC.md#node) and a new line is enough however).
added a more helpful example for unintuitive essential use cases.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/lib/generators.nix
+++ b/modules/lib/generators.nix
@@ -221,7 +221,7 @@
         lib.throwIf (name == "") "Directive must not be empty"
         (let vType = typeOf value;
         in if elem vType [ "int" "float" "bool" "string" ] then
-          "${name} ${literalValueToString value}"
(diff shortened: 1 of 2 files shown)


ymnejmi:
I dont know how or when it happened since neither hm nor zellij pushed any commits, but the issue that I tried to fix (which DEFINITELY happened) with this pr (which it DID fix) is no longer an issue; I am back now to using the master branch of hm which results in node statements with no semicolon and zellij no longer spits an error at node statements with no semicolon...

I guess now this is useless on part of the fix so ill close it, however a better example could definitely be provided..

PR #5710, opened 2024-08-05
dunst: only use `-config` flag when `configFile` is set

Anomalocaridid:
### Description

This ensures that in dunst's systemd service, the `-config` flag is only used when `configFile` is set. This is because when passing `-config` to dunst, dunst does not look for config files other than the specified path. This breaks dunst's drop-in functionality, where files can be placed in `~/.config/dunst/dunstrc.d` to override settings in `~/.config/dunst/dunstrc`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/services/dunst.nix
+++ b/modules/services/dunst.nix
@@ -65,9 +65,8 @@ in {
       };
 
       configFile = mkOption {
-        type = with types; either str path;
-        default = "${config.xdg.configHome}/dunst/dunstrc";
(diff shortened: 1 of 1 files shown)


rycee:
Thanks for the contribution!

Anomalocaridid:
It was my pleasure!

Issue #2451, opened 2021-11-07
Allow installing of fish plugins from nixpkgs


legendofmiracles:
Nixpkgs has a handful or so fish plugins packaged, but enabling them is tedious and has to be done with some wrapper.
HM has options for plugins in fish, but it doesn't seem to support the `mkFishPlugin` format used by nixpkgs:
```
error: The option `home-manager.users.nix.programs.fish.plugins.[definition 1-entry 2].__ignoreNulls' does not exist. Definition values:
```

All the plugins, and their builder are [here](https://github.com/NixOS/nixpkgs/tree/master/pkgs/shells/fish/plugins)
Instructions how to install with some wrapper here: https://github.com/NixOS/nixpkgs/blob/8284fc30c84ea47e63209d1a892aca1dfcd6bdf3/doc/builders/packages/fish.section.md#fish-wrapper-sec-fish-wrapper

Ideally home-manager would just support dropping them into programs.fish.plugins, and I think it would also be best to drop the current way plugins are made, and force to user to use mkFishPlugin, similarly to how vim does it with its [buildVimPlugin](https://github.com/legendofmiracles/dotnix/blob/7aa5c17c1545339d083310575618cb0ef7e967f8/hm/nvim.nix#L10N)

samhh:
There are issues with plain `wrapFish` as well, because then Home Manager is unable to generate completions - `tree` the output in `/nix/store/` to see what I mean.

As a workaround you can override the `fish` package with a `wrapFish` executable :exploding_head: https://github.com/samhh/dotfiles/commit/dd8b0c37dea5cd53124c3d131202f81ccf1faa8d#diff-adb51b0bca750896f46c5ffb6529ba7604d2f15b9453ee4faf61051861eb658fR15

teto:
started trying fish and it's annoying. Can't we coordinate with nixpkgs to have something working ? seems odd we can't do better

PR #5714, opened 2024-08-07
Fish copy plugins

diniamo:
### Description

Fish plugin files are now installed directly into the config directory instead of being sourced. This is needed because sourcing breaks attaching event listeners using `functions --query`, which sponge requires for example.

Breaking changes:
- `programs.fish.plugins` not takes a list of `path`s, instead of a submodule, because the name is no longer needed, only the source
- init.fish and key_bindings.fish are no longer sourced. While there is no official plugin spec, they are not a part of the [fisher](https://github.com/jorgebucaran/fisher) spec, which is what everyone uses outside of Nix.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

Uhh no clue

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -6,29 +6,6 @@ let
 
   cfg = config.programs.fish;
 
-  pluginModule = types.submodule ({ config, ... }: {
-    options = {
-      src = mkOption {
-        type = types.path;
-        description = ''
-          Path to the plugin folder.
-
-          Relevant pieces will be added to the fish function path and
-          the completion path. The {file}`init.fish` and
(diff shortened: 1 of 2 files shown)


diniamo:
The test fail seems to be irrelevant (?).

teto:
I have enabled fish on my laptop for fun and was bummed by this https://github.com/nix-community/home-manager/issues/2451 . 
How does this compare with that issue that seems better in UX terms (HM would behave like nixpkgs, accepting derivations as plugins).

diniamo:
This isn't really related to that issue I believe.

I'm also closing this, since there has been no maintainer interest, and I don't use HM anymore.

PR #5716, opened 2024-08-08
jqp: add module

Demianeen:
### Description

I added `jqp` module with the ability to set the `jqp` configuratin within nix. I have tested the module and it creates the `~/.jqp.yaml` file with the nix config settings

### Checklist

- [x] Code tested through `nix-shell --pure tests -A run.all` or `nix develop --ignore-environment .#all` using Flakes. (although it fails, please see below)

    ```
    {component}: {description}

    {long description}
    ```

#### Additional info

I added a test case based on the Polybar example mentioned above but still couldn't manage to get the test working after half a day. I only found in the Home Manager manual how to run them. Is there any documentation available on how to write these tests with comments?

### DIFF
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -140,6 +140,7 @@ let
     ./programs/java.nix
     ./programs/jetbrains-remote.nix
     ./programs/jq.nix
+    ./programs/jqp.nix
     ./programs/jujutsu.nix
     ./programs/joshuto.nix
     ./programs/joplin-desktop.nix
--- a/tests/default.nix
+++ b/tests/default.nix
@@ -171,6 +171,7 @@ in import nmtSrc {
     ./modules/programs/irssi
     ./modules/programs/jujutsu
     ./modules/programs/joplin-desktop
+    ./modules/programs/jqp
     ./modules/programs/k9s
     ./modules/programs/kakoune
     ./modules/programs/khal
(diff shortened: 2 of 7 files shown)


tests/default.nix
```diff
@@ -110,6 +110,7 @@ in import nmtSrc {
     ./modules/programs/ncmpcpp
     ./modules/programs/ne
     ./modules/programs/neomutt
+    ./modules/programs/jqp
```
Alexdelia, line 113: you should probably sort it by `alphabetic` order

Demianeen: Sure, but there is the issue that created test are not passing. Will it help?

Alexdelia: I do not think so

I sadly can't help. I am a new comer to `nix` and `home-manager`, and started learning by checking the `PR` and found yours

I hope someone with a proper experience will be able to help (as I would also love to learn why it is going wrong)

Demianeen:
So, what do we do?

Demianeen:
I would be happy to finish it but I would need some help with tests. Is there any good docs for them or should I close it for now?

modules/modules.nix
```diff
@@ -140,6 +140,7 @@ let
     ./programs/java.nix
     ./programs/jetbrains-remote.nix
     ./programs/jq.nix
+    ./programs/jqp.nix
```
khaneliman, line 143: This was missing.

Demianeen: Wow, thanks! I don't know how I missed that)

PR #5720, opened 2024-08-09
gpg-agent: add quietShellIntegration option

redxtech:
### Description

This adds an option to `services.gpg-agent`: `quietShellIntegration`.

**Rationale:** it is annoying to see the verbose output on every ssh connection attempt.

If enabled, this adds `--quiet` to `gpg-connect-agent` when it is invoked as per `enable<Shell>Integration`.

I have left it disabled by default to not modify any existing behavior. 

### Checklist

- [ ] ~~Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).~~

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@rycee

### DIFF
--- a/modules/services/gpg-agent.nix
+++ b/modules/services/gpg-agent.nix
@@ -9,8 +9,10 @@ let
 
   homedir = config.programs.gpg.homedir;
 
+  gpgConnectAgentOpts = optionalString cfg.quietShellIntegration "--quiet";
+
   gpgSshSupportStr = ''
-    ${gpgPkg}/bin/gpg-connect-agent updatestartuptty /bye > /dev/null
+    ${gpgPkg}/bin/gpg-connect-agent ${gpgConnectAgentOpts} updatestartuptty /bye > /dev/null
   '';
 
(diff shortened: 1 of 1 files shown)


redxtech:
Actually it turns out this doesn't resolve the issue, it was something else that I had changed locally.

I'll mark this as WIP for now until I can determine if it's necessary or not.

redxtech:
Turns out only part of the output was unrelated. Without `--quiet`, I was still getting some output. This fixes that.

redxtech:
Also: I'm fairly certain the failing test (macos) is unrelated. The build logs seem to point towards `mpv`, and nothing that I've touched should lead to a build error.

sumnerevans (approved):

JohnRTitor (approved):
Otherwise looks good to me.

@khaneliman @sumnerevans could you perhaps merge/force push with this change then merge?

Issue #5717, opened 2024-08-08
bug: Firefox no longer launches with custom profiles


dairyisscary:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

I have a home module that looks something like this:

```nix
  programs.firefox = {
    enable = true;
    package = lib.mkIf isDarwin (pkgs.firefox-bin.overrideAttrs (prevAttrs: {
      postInstall = ''
        folder="$out/Applications/Firefox.app/Contents/Resources/distribution"
        mkdir -p "$folder"
        echo '{
          "policies": {
            "AppAutoUpdate": false,
            "DisableAppUpdate": true
          }
        }' > "$folder/policies.json"
      '';
    }));
    profiles.${config.home.username} = {
      settings = {
        "toolkit.legacyUserProfileCustomizations.stylesheets" = true;
      };
      userChrome = builtins.readFile ./firefox-userchrome.css;
      search = {
        force = true;
        default = "Kagi";
        engines = {
           # elided...
        };
      };
    };
  };
```

After #5128, firefox no longer opens on aarch64-darwin (I have not tested any other platform). I have verified that the commit _before_ is not affected.

I noticed that `profiles.ini` file looks exactly the same between the two profiles, with #5128 and without, except for this notable line:
```ini
[General]
StartWithLastProfile=1
Version=2 # <- this one.

[Profile0]
Default=1
IsRelative=1
Name=eric
Path=Profiles/eric
```

And that seems to be new configuration in #5128. If i just manually link this file to the older profile without this `Version` line, but keep the newer profile, firefox does seem to open correctly. I can't find any mozilla documentation on this property.

### System information

```markdown
nix run nixpkgs#nix-info -- -m
 - system: `"aarch64-darwin"`
 - host os: `Darwin 23.6.0, macOS 14.6.1`
 - multi-user?: `yes`
 - sandbox: `no`
 - version: `nix-env (Nix) 2.18.5`
 - nixpkgs: `/nix/store/4cpakzyvfw1rmm9v5i3387x6jd2h1v86-source`
```

khaneliman:
Same behavior here.. I saw it immediately after that got merged too. My workaround for the time being has been to create a rw profiles.ini with the same contents after a flake switch...

brckd:
[...]

PR #5723, opened 2024-08-10
firefox: expose profiles version as an option. So firefox can be lanuched ono darwin

yuanwang-wf:
### Description
PR aims to fix https://github.com/nix-community/home-manager/issues/5717 

add an option for profiles general version 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/firefox/mkFirefoxModule.nix
+++ b/modules/programs/firefox/mkFirefoxModule.nix
@@ -61,7 +61,8 @@ let
     }) // {
       General = {
         StartWithLastProfile = 1;
-        Version = 2;
+      } // lib.optionalAttrs (cfg.profileVersion == null) {
+        Version = cfg.profileVersion;
       };
     };
 
@@ -341,6 +342,13 @@ in {
         BlockAboutConfig = true;
       };
     });
+
+    profileVersion = mkOption {
+      internal = true;
+      type = types.nullOr types.ints.unsigned;
+      default = if isDarwin then null else 2;
+      description = "profile version, set null for nix-darwin";
+    };
 
     profiles = mkOption {
       inherit visible;


brckd (commented):
Thank you! Can someone confirm this doesn't break other FF derivations on Darwin? Does FF on linux work with Version = 2 set?

khaneliman:
> Thank you! Can someone confirm this doesn't break other FF derivations on Darwin? Does FF on linux work with Version = 2 set?

I use the same firefox config for darwin and linux. Darwin doesn't launch and linux launches, fine.

brckd:
So one could also add an option to the mkFirefoxModule method, since the version config seems to be related to the Firefox derivative, not to the OS. This PR might serve as a fix in the meantime though.

khaneliman (approved):
LGTM, appreciate being able to set this.

khaneliman:
Although... wouldn't it make sense to default it to null if it detects darwin?

khaneliman (approved):
Looks better, thanks!

yuanwang-wf:
> Although... wouldn't it make sense to default it to null if it detects darwin?

i updated it, it seems to work on my machine

```
❯ bat /nix/store/h5123pw39rjpfx4mcbirfw5j469laqfl-hm_LibraryApplicationSupportFirefoxprofiles.ini
───────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: /nix/store/h5123pw39rjpfx4mcbirfw5j469laqfl-hm_LibraryApplicationSupportFirefoxprofiles.ini
───────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ [General]
   2   │ StartWithLastProfile=1
   3   │
   4   │ [Profile0]
   5   │ Default=1
   6   │ IsRelative=1
   7   │ Name=home
   8   │ Path=Profiles/home
───────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

```

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -341,6 +340,13 @@ in {
       };
     });
 
+    profileVersion = mkOption {
+      inherit visible;
```
brckd, line 344:
```suggestion
      internal = true;
```

It would make sense to mark this as internal.

brckd (approved):
Perfect otherwise! It's both effective and configurable.

yuanwang-wf:
hey @brckd, anything I should do to get this merged ? btw thanks for all the helps

brckd:
Just be patient and wait for rycee to review it. There is nothing we can do in the meantime afaik.

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -341,6 +340,13 @@ in {
       };
     });
 
+    profileVersion = mkOption {
+      internal = true;
+      type = types.nullOr types.ints.unsigned;
+      default = if isDarwin then null else 2;
```
booxter, line 346: I am not sure this is correct (but I agree it will help Darwin users). `Version=2` means that `[Install]` entries are saved in `profiles.ini` and `installs.ini` is never loaded (as a backup). I think it breaks any firefox package that won't also set `MOZ_LEGACY_PROFILES=1` (for example, `firefox-bin` or a Firefox installed by `brew`.) I captured what I think is happening [here](https://github.com/nix-community/home-manager/pull/5724#issuecomment-2323655812).

---

I think the proper way to fix this is remove `Version=2` for **all** platforms (which will enforce legacy, v1 profiles behavior - the one that `firefox` `nixpkgs` package already enforces with `MOZ_LEGACY_PROFILES=1`, set for the binary wrapper).

booxter: UPD: I stand corrected on `firefox-bin`: I think it's also [wrapped](https://github.com/NixOS/nixpkgs/blob/d80a3129b239f8ffb9015473c59b09ac585b378b/pkgs/top-level/all-packages.nix#L30124) with `wrapFirefox`, which will guarantee that `MOZ_LEGACY_PROFILES=1` is set (meaning, v1 logic is implied - using `installs.ini`.)

So the only versions of firefox affected are externally installed. This is, largely, just Darwin installations. Still, I don't think removal of `Version=2` line for all platform should affect anyone (happy to learn otherwise); so introducing a new option to set the version seems unnecessary.

yuanwang-wf: hey @brckd when you have time, do you want to weight in ?

brckd: I just tried both wrapped and unwrapped LibreWolf versions with the following `profile.ini` with `Version = 2` removed and they seem to run fine. Perhaps the issue was with another derivative, so sorry for the confusion.
```ini
[General]
StartWithLastProfile=1

[Profile0]
Default=1
IsRelative=1
Name=Bricked
Path=bricked```

**Comment by khaneliman** [comment]
@yuanwang-wf @rycee  looks like this doesn't actually filter out the null value so it didn't fix darwin. Now the profile has `Version=null` so we see the same issue. Looks like it needs to be `lib.mkIf stdenv.isLinux 2` instead if we don't filter out nulls.

File: modules/programs/firefox/mkFirefoxModule.nix [hunk 3]
```diff
@@ -61,7 +61,8 @@ let
     }) // {
       General = {
         StartWithLastProfile = 1;
-        Version = 2;
+      } // lib.optionalAttrs (cfg.profileVersion == null) {
```
khaneliman, line 64: this should be `cfg.profileVersion != null`, didn't notice that at first review.

rycee:
Apologies! This error was introduced by me when doing the merge. I
switched the explicit `if` expression to use `optionalAttrs` and missed
inverting the condition. I pushed a quick fix to master that does invert
the condition, hopefully it'll work better!

PR #5724, opened 2024-08-10
fix: firefox launching on darwin

### LINKED ISSUE #5717: bug: Firefox no longer launches with custom profiles

HyunggyuJang:
### Description

Fixes #5717, by removing `Version` property in profiles.ini, as observed in #5717.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee, @kira-bruneau, @bricked

### DIFF
--- a/modules/programs/firefox/mkFirefoxModule.nix
+++ b/modules/programs/firefox/mkFirefoxModule.nix
@@ -61,8 +61,7 @@ let
     }) // {
       General = {
         StartWithLastProfile = 1;
-        Version = 2;
-      };
+      } // (if isDarwin then { } else { Version = 2; });
     };
 
   profilesIni = generators.toINI { } profiles;


booxter:
I can confirm that this fixes startup of Firefox on Darwin when using an externally installed Firefox (from homebrew).

booxter:
I was curious as to why this works. Here is what I think is happening.

1. `installs.ini` is a regular file under `~/Library/Application Support/Firefox/installs.ini`.
2. `profiles.ini` is a symlink to a read-only nix store file.
3. When Firefox (or Thunderbird, see another PR I posted, link above) starts:
a) If it's `Version=2`, `[Install]` sections are expected to be present in the `profiles.ini` file. Firefox does *not* load them from `installs.ini`, which makes the startup fail.
b) on the other hand, when `Version` is missing, it [restores installed profiles from backup](https://github.com/mozilla/gecko-dev/blob/5f641898e12dafdfb7d54a93f4f3164d2c1642b2/toolkit/profile/nsToolkitProfileService.cpp#L1019) (`installs.ini`). Later, it [attempts](https://github.com/mozilla/gecko-dev/blob/5f641898e12dafdfb7d54a93f4f3164d2c1642b2/toolkit/profile/nsToolkitProfileService.cpp#L2317) to write the `profiles.ini` file with `Version=2` and the `[Install]` entries, but [fails](https://github.com/mozilla/gecko-dev/blob/5f641898e12dafdfb7d54a93f4f3164d2c1642b2/toolkit/profile/nsToolkitProfileService.cpp#L2364) to do so (because it's read-only). Still, the app works because Firefox [ignores](https://github.com/mozilla/gecko-dev/blob/5f641898e12dafdfb7d54a93f4f3164d2c1642b2/xpcom/base/nsDebug.h#L136) these errors.

I was also wondering why `Version=2` works for Linux, since I'd expect it to fail the same way there too. This seems to be because we pass:

```
            --set MOZ_LEGACY_PROFILES 1 \
            --set MOZ_ALLOW_DOWNGRADE 1 \
```

when building the firefox wrapper from `nixpkgs`. (Though not for `firefox-bin`! - perhaps it's also broken now?)

With that, I am not even sure we should have `Version=2` set for any platform: this version implies that `install` data is dynamically added to the `profiles.ini` file during the app use; this is obviously impossible with a read-only nix store. Perhaps we should remove the `Version=` line unconditionally, regardless of the platform?

booxter:
I think this was done in https://github.com/nix-community/home-manager/pull/5723 already. I don't think the linked PR is the best approach and proposed an alternative that will fix Thunderbird too here: https://github.com/nix-community/home-manager/pull/5801 but regardless, this PR here is no longer needed I think.

HyunggyuJang:
Fair enough. Let me close this PR. I think #5723 has been made almost the same time I made this! I haven’t noticed that.

PR #5742, opened 2024-08-16
hyprland: make package nullable

diniamo:
### Description

Makes `wayland.windowManager.hyprland.package` nullable.

Should xwayland still be installed, if the package is null?

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@fufexan

### DIFF
--- a/modules/services/window-managers/hyprland.nix
+++ b/modules/services/window-managers/hyprland.nix
@@ -56,7 +56,11 @@ in {
       '';
     };
 
-    package = lib.mkPackageOption pkgs "hyprland" { };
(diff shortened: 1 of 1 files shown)


fufexan:
If the package is null it is assumed the user manages xwayland somewhere else too. So let's add ` && cfg.package != null`.

diniamo:
Sure, I was unsure about that.

fufexan (approved):

diniamo:
@rycee could we get this merged please?

khaneliman (approved):

PR #5745, opened 2024-08-16
git: add config option for gpg.format

ezrizhu:
### Description

git provides three options for signing key format, openpgp(default), ssh, and x509. This commit provides the option to specify one of the other two available formats.

Motivation: currently, we already have signing.pgpPath and signing.Key, with the addition of signing.Format, we also have support for ssh signing, which most major git servers also support. 

see https://git-scm.com/docs/git-config#Documentation/git-config.txt-gpgformat

### Checklist

- [X] Code tested through `nix-shell --pure tests -A run.all` or `nix develop --ignore-environment .#all` using Flakes. **I ran `nix-shell --pure tests -A run.git-with-most-options`**

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@rycee

### DIFF
--- a/modules/programs/git.nix
+++ b/modules/programs/git.nix
@@ -38,6 +38,17 @@ let
         defaultText = "\${pkgs.gnupg}/bin/gpg2";
         description = "Path to GnuPG binary to use.";
       };
+
+      gpgFormat = mkOption {
+        type = types.str;
+        default = "openpgp";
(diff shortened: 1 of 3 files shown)


yzx9:
This might be a duplicate of #5516.

khaneliman:
Implemented in https://github.com/nix-community/home-manager/pull/5516

PR #5750, opened 2024-08-17
kitty: Removed IFD

### LINKED ISSUE #5110: bug: IFD is used to set Kitty themes

### Issue description

The Kitty theme option added in #2710 uses import-from-derivation to find the theme file. In later revisions, an error is also thrown if the theme does not exist.

IFD slows down evaluation and makes instantiation harder in many deployment scenarios, and should be avoided where possible.

Either the `themes.json` from `kitty-themes` should be vendored, or the file finding logic should be moved into the activation script instead.

### Maintainer CC

@NelsonJeppesen

### System information

```markdown
- system: `"x86_64-linux"`
- host os: `Linux 6.7.5-zen1, NixOS, 24.05 (Uakari), 24.05pre585741.b98a4e1746ac`
- multi-user?: `yes`
- sandbox: `yes`
- channels(root): `"home-manager, impermanence, nixos, sops-nix"`
```

alejandro-angulo:
Removes IFD from kitty module. Theme check happens at activation time now.

Fixes #5110 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/kitty.nix
```diff
@@ -218,6 +210,13 @@ in {
       '';
     };
 
+    home.activation.kittyThemeChecker = hm.dag.entryAnywhere ''
+      if [ ! -f ${pkgs.kitty-themes}/share/kitty-themes/themes/${cfg.theme}.conf ]; then
```
alejandro-angulo, line 214: Should this be checking the contents of themes.json too :thinking:

modules/programs/kitty.nix
```diff
@@ -218,6 +210,20 @@ in {
...
 
+    home.activation.kittythemechecker = mkIf (cfg.theme != null)
+      (hm.dag.entryAnywhere ''
+        theme_path=$(${pkgs.jq}/bin/jq '.[] | select(.name=="${cfg.theme}" | .file' ${pkgs.kitty-themes}/share/kitty-themes/themes/themes.json)
+        if [ -z "$theme_path" ]; then
+          echo 'kitty-themes does not contain a theme named ${cfg.theme}' &>2
+          exit 1
+        fi
+
+        if [ ! -f "${pkgs.kitty-themes}/share/kitty-themes/themes/$theme_path" ]; then
```
alejandro-angulo, line 221: This check might be redundant (I think this is testing that `kitty-theme`s is packaged correctly). Should I leave this in?

modules/programs/kitty.nix
```diff
@@ -218,6 +210,20 @@ in {
       '';
     };
 
+    home.activation.kittythemechecker = mkIf (cfg.theme != null)
```
alejandro-angulo, line 213: I thought about writing a test for this check but I'm not sure how to test this. Should this even be tested (most of the tests I found check that config files are setup correctly rather than checking activation logic)?

modules/programs/kitty.nix
```diff
@@ -193,15 +193,7 @@ in {
...
-          include ${pkgs.kitty-themes}/share/kitty-themes/${
-            let
-              matching = filter (x: x.name == cfg.theme) (builtins.fromJSON
-                (builtins.readFile
-                  "${pkgs.kitty-themes}/share/kitty-themes/themes.json"));
-            in throwIf (length matching == 0)
-            "kitty-themes does not contain a theme named ${cfg.theme}"
-            (head matching).file
-          }
+          include ${pkgs.kitty-themes}/share/kitty-themes/themes/${cfg.theme}.conf
```
rycee, line 196: Unfortunately this won't work and is not compatible with the activation script block.

For example, if you have `programs.kitty.theme = "Space Gray Eighties";` in your HM configuration. Then the activation script block will work OK but this will produce

    include /nix/store/…/themes/Space Gray Eighties.conf

while the actual file is `/nix/store/…/themes/SpaceGray_Eighties.conf`.

rycee: So, I think you either have to keep the IFD, or change the meaning of `programs.kitty.theme` such that the user is expected to put the file name. I do not want to include the `themes.json` in the HM repository.

rycee: Perhaps remove the `theme` option and instead add a `themeFile` option like so:

```nix
themeFile = mkOption {
  default = null;
  example = "SpaceGray_Eighties";
  description = ''
    Apply a Kitty color theme. This option takes the file name of a theme in
    a `kitty-themes`, without the `.conf` suffix. See
    <https://github.com/kovidgoyal/kitty-themes/tree/master/themes> for a
    list of themes.
  '';
};
```

Then use `mkChangedOptionModule` to convert existing instances of `programs.kitty.theme` using IFD into the new `themeFile` option.

modules/programs/kitty.nix
```diff
@@ -57,6 +54,25 @@ let
     '';
   };
 in {
+  imports = [
```
alejandro-angulo, line 57: @rycee I went with your suggestion and replaced `theme` with `themeFile`.

modules/programs/kitty.nix
```diff
@@ -1,9 +1,6 @@
 { config, lib, pkgs, ... }:
-
```
alejandro-angulo, line 2: I'm not sure why these formatting changes were made. I ran the format script at the root of the repo. I can clean this up if we want to omit the formatting changes.

tests/integration/standalone/kitty.nix
```diff
@@ -0,0 +1,73 @@
+{ pkgs, ... }: {
```
alejandro-angulo, line 1: Thanks for adding this in @rycee :pray: 

Out of curiosity (and, tbh still somewhat of a nix noob), how can I run this test on my machine (without running the entire test suite)?

modules/misc/news.nix
```diff
@@ -1703,6 +1703,19 @@ in {
           one place. See https://github.com/glanceapp/glance for more.
         '';
       }
+
+      {
```
alejandro-angulo, line 1707: Figure this might be good to call out.

tests/integration/standalone/kitty.nix
rycee: If you are using Nix Flakes then you can run it using

    nix build --option sandbox false -L .#integration-test-kitty

modules/programs/kitty.nix
```diff
@@ -211,13 +218,23 @@ in {
...
+      ];
     } // optionalAttrs pkgs.stdenv.hostPlatform.isLinux {
       onChange = ''
         ${pkgs.procps}/bin/pkill -USR1 -u $USER kitty || true
       '';
     };
 
+    home.activation.checkKittyTheme = mkIf (cfg.themeFile != null) (let
+      themePath =
+        "${pkgs.kitty-themes}/share/kitty-themes/${cfg.themeFile}.conf";
```
rycee, line 230: I think you need

```suggestion
        "${pkgs.kitty-themes}/share/kitty-themes/themes/${cfg.themeFile}.conf";
```

modules/programs/kitty.nix
```diff
@@ -57,6 +54,25 @@ let
...
+    ] (config:
+      let value = getAttrFromPath [ "programs" "kitty" "theme" ] config;
+      in if value != null then
+        (let
+          matching = filter (x: x.name == value) (builtins.fromJSON
+            (builtins.readFile
+              "${pkgs.kitty-themes/share/kitty-themes/themes.json}"));
+        in throwIf (length matching == 0)
+        "kitty-themes does not contain a theme named ${value}"
+        (head matching).file)
```
rycee, line 71: I think you'll need to remove the `themes/` prefix and the `.conf` suffix from this value to be compatible with the expected format of `themeFile`.

tests/modules/programs/kitty/theme-to-themeFile.nix
```diff
@@ -0,0 +1,21 @@
+{ config, lib, pkgs, options, ... }: {
+  config.programs.kitty = {
+    enable = true;
+    theme = "Space Gray Eighties";
+
+    test.stubs.kitty = { };
+
+    test.asserts.warnings.enable = false;
+    test.asserts.warnings.expected = [''
```
alejandro-angulo, line 9: This test is failing locally for me with this message

> Expected home-files/asserts/warnings.actual to be same as /nix/store/arjnippklyvws49lxwgi1y7hb0yjayh5-warnings.expected but were different:

I checked the expected warnings file and it's empty for some reason.

```
❯ bat /nix/store/arjnippklyvws49lxwgi1y7hb0yjayh5-warnings.expected
───────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: /nix/store/arjnippklyvws49lxwgi1y7hb0yjayh5-warnings.expected   <EMPTY>
───────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

```

Need to look into this later

tests/modules/programs/kitty/theme-to-themeFile.nix
```diff
@@ -0,0 +1,21 @@
+{ config, lib, pkgs, options, ... }: {
+  config.programs.kitty = {
+    enable = true;
+    theme = "Space Gray Eighties";
+
+    test.stubs.kitty = { };
+
+    test.asserts.warnings.enable = true;
+    test.asserts.warnings.expected = [''
```
alejandro-angulo, line 9: This test is failing locally for me with this message

> Expected home-files/asserts/warnings.actual to be same as /nix/store/arjnippklyvws49lxwgi1y7hb0yjayh5-warnings.expected but were different:

I checked the expected warnings file and it's empty for some reason.

```
❯ bat /nix/store/arjnippklyvws49lxwgi1y7hb0yjayh5-warnings.expected
───────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: /nix/store/arjnippklyvws49lxwgi1y7hb0yjayh5-warnings.expected   <EMPTY>
───────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

```

Need to look into this later

alejandro-angulo: Ah this was nested under config.programs.kitty :facepalm:

modules/programs/kitty.nix
alejandro-angulo: Glad you caught this. I ended up writing `tests/modules/programs/kitty/theme-to-themeFile.nix` to test this. 

There was also another mistake I made a couple lines up (curly brace is closed too late when I tried to refer to `pkgs.kitty-themes`) that the test caught.

rycee:

PR #5763, opened 2024-08-23
programs.texlive: extend documentation of options

Zocker1999NET:
The options were barely documented until now, making it unnecessarily hard for people even somewhat familiar with other, similar modules to use this one. This is especially because, intuitively, `packageSet` & `extraPackages` can be misunderstood and neither of their descriptions described their advantages. The enable option did not explain the module‘s purpose. I heavily expanded the option’s descriptions which hopefully makes this module useable to newcomers without resorting to look into the Nix code of the module.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/texlive.nix
+++ b/modules/programs/texlive.nix
@@ -14,12 +14,25 @@ in {
 
   options = {
     programs.texlive = {
-      enable = mkEnableOption "TeX Live";
+      enable = mkEnableOption ''
+        TeX Live package customization.
+
+        This module allows you to select which texlive packages you want to install.
+        Start by exteding {option}`programs.texlive.extraPackages`.
+      '';
 
       packageSet = mkOption {
         default = pkgs.texlive;
-        defaultText = literalExpression "pkgs.texlive";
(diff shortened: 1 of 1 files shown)


modules/programs/texlive.nix
```diff
@@ -14,21 +14,55 @@ in {
...
+      enable = mkEnableOption ''
+        TeX Live package customization.
+
+        This module allows you to select which `pkgs.texlivePackages.*` you want to install.
+        Start by editing {option}`programs.texlive.extraPackages`.
+
+        If one of the pre-built common environments already suits you
+        (e.g. `pkgs.texliveFull` or `pkgs.texliveMedium`),
+        you do not need to use this module.
+      '';
```
rycee, line 26: The `mkEnableOption` function automatically appends a `.` to this string so you can omit that (and the newline character to avoid a space between module and `.`: "module ."

```suggestion
        you do not need to use this module'';
```

rycee (commented):
Thanks! Looks good and much more helpful.

PR #5765, opened 2024-08-24
hyprland: make package nullable

nathanregner:
### Description

Allow setting `wayland.windowManager.hyprland.package` to null for standalone home-manager installations. The hyprland package is not needed if it's already installed via the NixOS module.

At some point, I let my package versions get out of sync and encountered some weirdness trying to invoke `hyprctl`. The module code already seems to check for null; this just fixes the following error:

```
   error: A definition for option `wayland.windowManager.hyprland.package' is not of type `package'. Definition values:
       - In `/nix/store/...': null
```

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@fufexan

### DIFF
--- a/modules/services/window-managers/hyprland.nix
+++ b/modules/services/window-managers/hyprland.nix
@@ -56,7 +56,13 @@ in {
       '';
     };
 
-    package = lib.mkPackageOption pkgs "hyprland" { };
+    package = lib.mkPackageOption pkgs "hyprland" {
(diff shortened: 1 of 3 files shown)


fufexan:
https://github.com/nix-community/home-manager/pull/5742 ?

nathanregner:
My bad, totally missed that!

PR #5766, opened 2024-08-24
nix: add nix.package to home.packages

BarrOff:
### Description

Add nix.package to home.packages if nix.package is not null.
This fixes the following use case:

- set `nix.package` to a version newer than the default system nix version
- `nix.settings` or `nix.extraOptions` contains some options which work for `nix.package` but not the older system nix version
- after creating a new home-manager generation nix commands fail because of the incompatible options 

In my opinion it makes sense to install the nix package compatible to the configuration settings.
However the change might not be backwards compatible because it potentially installs a new package to existing setups.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@polykernel

### DIFF
--- a/modules/misc/nix.nix
+++ b/modules/misc/nix.nix
@@ -276,6 +276,8 @@ in {
   };
 
   config = mkIf cfg.enable (mkMerge [
+    (mkIf (cfg.package != null) { home.packages = [ cfg.package ]; })
+
     (mkIf (cfg.nixPath != [ ] && !cfg.keepOldNixPath) {
       home.sessionVariables.NIX_PATH = "${nixPath}";
     })


drupol:
I was looking for the same thing and finally did: https://code.europa.eu/ecphp/devs-profile/-/commit/d992404513ca8a8909b9116538ad618f28fef1b7

khaneliman:
> However the change might not be backwards compatible because it potentially installs a new package to existing setups.

This is the only thing I wasn't sure about. 

For me just testing this branch real quick: 

```bash
ADDED
[A.] boehm-gc       8.2.8-dev
[A.] boost          1.87.0-dev
[A.] libarchive     3.8.1, 3.8.1-dev
[A*] nix            2.31.1-dev
[A.] nix-cmd        2.31.1-dev
[A.] nix-expr       2.31.1-dev
[A.] nix-expr-c     2.31.1, 2.31.1-dev
[A.] nix-fetchers   2.31.1-dev
[A.] nix-fetchers-c 2.31.1, 2.31.1-dev
[A.] nix-flake      2.31.1-dev
[A.] nix-flake-c    2.31.1, 2.31.1-dev
[A.] nix-main       2.31.1-dev
[A.] nix-main-c     2.31.1, 2.31.1-dev
[A.] nix-perl       2.31.1
[A.] nix-store      2.31.1-dev
[A.] nix-store-c    2.31.1, 2.31.1-dev
[A.] nix-util       2.31.1-dev
[A.] nix-util-c     2.31.1, 2.31.1-dev
[A.] nlohmann_json  3.11.3

SIZE: 23.3 GiB -> 23.5 GiB
DIFF: 151 MiB
```

Before:

```
trace-which nix
/nix/store/69lc6ybbzjn6y0n3j6fvfnabknxcyjpf-nix-2.31.1/bin/nix
```

After:

```
trace-which nix
/nix/store/yqngwidm05f5cbk6sssazy9a0xqkqs95-home-manager-path/bin/nix
```

teto:
> In my opinion it makes sense to install the nix package compatible to the configuration settings.

yes current behavior is surprising. As @khaneliman mentioned it would be a breaking change (one that could annoy lots of people). It could be done adding a check on home.stateVersion (check mpd.nix for an example) and a mention in release-notes. We could get this in before 25.11

ambroisie:
Bumping this back up to get merged at some point, we've already slipped past 25.11.

PR #5768, opened 2024-08-25
kanshi: add support for output aliases

octvs:
- Add `services.kanshi.profiles.<name>.outputs.*.alias` to support new alias directive from kanshi [1].
- Add relevant coverage by modifying the existing "new-configuration" test.

[1]: https://git.sr.ht/~emersion/kanshi/commit/1ed86ce523553240ceb81e9d466b88d3554a66bb

### Description

_Refer to the commit msg_

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@nurelin

### DIFF
--- a/modules/services/kanshi.nix
+++ b/modules/services/kanshi.nix
@@ -123,6 +123,15 @@ let
         '';
       };
 
+      alias = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        example = "laptopMonitor";
+        description = ''
+          Defines an alias for the output
+        '';
+      };
+
       adaptiveSync = mkOption {
         type = types.nullOr types.bool;
         default = null;
@@ -135,15 +144,16 @@ let
     };
   };
 
-  outputStr =
-    { criteria, status, mode, position, scale, transform, adaptiveSync, ... }:
+  outputStr = { criteria, status, mode, position, scale, transform, adaptiveSync
+    , alias, ... }:
     ''output "${criteria}"'' + optionalString (status != null) " ${status}"
     + optionalString (mode != null) " mode ${mode}"
     + optionalString (position != null) " position ${position}"
     + optionalString (scale != null) " scale ${toString scale}"
     + optionalString (transform != null) " transform ${transform}"
     + optionalString (adaptiveSync != null)
-    " adaptive_sync ${if adaptiveSync then "on" else "off"}";
+    " adaptive_sync ${if adaptiveSync then "on" else "off"}"
+    + optionalString (alias != null) " alias \$${alias}";
 
   profileModule = types.submodule {
     options = {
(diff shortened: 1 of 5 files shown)


rycee:
Looks good as far as I can tell (as a person not using Kanshi). Is @nurelin's comment resolved in the current version of this PR?

nurelin:
Not yet.

octvs:
Now it should be. My apologies for the delay, I already had the patch ready but
I was busy with other entanglements.

Please take a look at the assertion that I added, I'm still rough on the edges
with nix.

-- 
O. Cenk Taskin

octvs:
Just bumped into a note I made for myself while checking the kanshi docs
previously.

> Current test case is faulty since kanshi also doesn't allow wildcards on
> global scope. 

I've fixed that as well. Let me know if we should also add an assertion for
that.

About the tests: Sadly I've no idea why the it is failing with

```txt
kanshi-new-configuration: FAILED
Expected home-files/asserts/assertions.actual to be same as /nix/store/y73rids8ybzgg3a1rakvpi3fva8lh706-assertions.expected but were different:
--- actual
+++ expected
@@ -1 +0,0 @@
-Output kanshi.*.output.alias can only be defined on global scope
\ No newline at end of file
For further reference please introspect /nix/store/4wql2ffzp97xhn38za068ya1h9c56pvc-nmt-report-kanshi-new-configuration
```

Is it formatting or should I have declared the assertions somewhere else as
well?

nurelin:
Thank you for fixing the mistake with the wildcard! I do not think that we really need to enforce this with an assertion.

For the test failure, I think that is due to [this line](https://github.com/nix-community/home-manager/pull/5768/files#diff-2522539054f62379308b3d12076ed8b009b157f9db62b911b12164e9891e6d4aR128) that force the creation of the attribute with a value of null. You should update the assertion to check both the existence of the attribute `alias` and that the attribute is not `null`.

Could you also add the test below to check that the assertion actually raised:

```diff
diff --git a/tests/modules/services/kanshi/alias-assertion.nix b/tests/modules/services/kanshi/alias-assertion.nix
new file mode 100644
index 00000000..ba6b4760
--- /dev/null
+++ b/tests/modules/services/kanshi/alias-assertion.nix
@@ -0,0 +1,23 @@
+{ config, pkgs, ... }: {
+  config = {
+    services.kanshi = {
+      enable = true;
+      package = config.lib.test.mkStubPackage { };
+      settings = [
+        {
+          profile.name = "nomad";
+          profile.outputs = [
+            {
+              criteria = "eDP-1";
+              alias = "test";
+            }
+          ];
+        }
+      ];
+    };
+
+    test.asserts.assertions.expected = [
+      ''Output kanshi.*.output.alias can only be defined on global scope''
+    ];
+  };
+}
diff --git a/tests/modules/services/kanshi/default.nix b/tests/modules/services/kanshi/default.nix
index b7704b11..05e224aa 100644
--- a/tests/modules/services/kanshi/default.nix
+++ b/tests/modules/services/kanshi/default.nix
@@ -1,4 +1,5 @@
 {
   kanshi-basic-configuration = ./basic-configuration.nix;
   kanshi-new-configuration = ./new-configuration.nix;
+  kanshi-alias-assertion = ./alias-assertion.nix;
 }
```

With all of that, we should be good to go.

octvs:
> For the test failure, I think that is due to [this line]

Thanks for the pointer, now I modified the check however the fail is still
there. 

> Could you also add the test below to check that the assertion actually

Done. However this new test also fails with the same error as with the other
test.

nurelin:
The tests passed so it ok for me :+1:

rycee:
Thanks to both of you! Merged to master now 🙂

PR #5772, opened 2024-08-26
git: add module for `git maintenance`

PerchunPak:
### Description

Adds module for https://git-scm.com/docs/git-maintenance. Feel free to propose any grammar nitpicks, I am not a native, so will gladly accept them.

Example usage:

```nix
programs.git = {
  maintenance = {
    enable = true;
    repositories = [ "/home/perchun/dev/nixpkgs" ];
  };
}
```

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/git.nix
+++ b/modules/programs/git.nix
@@ -214,6 +214,41 @@ in {
         };
       };
 
+      maintenance = {
+        enable = mkEnableOption "" // {
+          description = ''
+            Enable the automatic {command}`git maintenance`.
+
+            See <https://git-scm.com/docs/git-maintenance>.
+          '';
+        };
+
+        repositories = mkOption {
+          type = with types; listOf str;
+          default = [ ];
+          description = ''
+            Repositories on which {command}`git maintenance` should run.
+
+            Should be a list of absolute paths.
+          '';
+        };
+
+        timers = mkOption {
+          type = types.attrsOf types.str;
+          default = {
+            hourly = "*-*-* 1..23:53:00";
+            daily = "Tue..Sun *-*-* 0:53:00";
+            weekly = "Mon 0:53:00";
+          };
+          description = ''
+            Systemd timers to create for scheduled {command}`git maintenance`.
+
+            Key is passed to `--schedule` argument in {command}`git maintenance run`
+            and value is passed to `Timer.OnCalendar` in `systemd.user.timers`.
+          '';
+        };
+      };
+
       diff-highlight = {
         enable = mkEnableOption "" // {
           description = ''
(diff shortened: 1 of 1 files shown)


modules/programs/git.nix
```diff
@@ -214,6 +214,41 @@ in {
...
+          '';
+        };
+
+        repositories = mkOption {
+          type = with types; listOf str;
+          default = [ ];
+          description = ''
+            Repositories on which {command}`git maintenance` should run.
+
+            Should be an absolute path.
```
teto, line 232: `a list of absolute paths.`

salko-ua:
good suggestion to add this module!

modules/programs/git.nix
```diff
@@ -479,6 +514,45 @@ in {
...
+        Service = {
+          Type = "oneshot";
+          ExecStart = let exe = lib.getExe cfg.package;
+          in ''
+            "${exe}" --exec-path="${exe}" for-each-repo --config=maintenance.repo maintenance run --schedule=%i
+          '';
+          LockPersonality = "yes";
+          MemoryDenyWriteExecute = "yes";
+          NoNewPrivileges = "yes";
+          RestrictAddressFamilies = "AF_UNIX AF_INET AF_INET6 AF_VSOCK";
```
PerchunPak, line 532: Forgot to mention, that this is the exact copy of service created by `git maintaince` by default

modules/programs/git.nix
```diff
@@ -479,6 +514,45 @@ in {
...
+
+            Timer = {
+              OnCalendar = time;
+              Persistent = true;
+            };
+
+            Install.WantedBy = [ "timers.target" ];
+          };
+      in lib.attrsets.mapAttrs' toSystemdTimer cfg.maintenance.timers;
+    })
```
h7x4, line 554: Maybe you could consider templating the timer as well? It doesn't carry a lot of benefit, but it avoids copy pasting logic and lets users start ad-hoc timers with an `OnCalendar` field of their choosing. We could also potentially drop the `OnCalendar` overrides and let the template handle it, since it takes `daily`, `weekly`, etc. just fine on it's own.

```suggestion
      systemd.user.timers = {
        "git-maintenance@" = {
          Unit.Description = "Optimize Git repositories data";
          
          Timer = {
            OnCalendar = "%i";
            Persistent = true;
          };
          
          Install.WantedBy = [ "timers.target" ];
        };
      }
      // 
      lib.attrsets.mapAttrs'
        (n: v: lib.attrsets.nameValuePair "git-maintenance@${n}" {
          Timer.OnCalendar = v;
        })
        cfg.maintenance.timers;
```

modules/programs/git.nix
```diff
@@ -479,6 +514,45 @@ in {
         };
     })
 
+    (mkIf cfg.maintenance.enable {
+      programs.git.iniContent.maintenance.repo = cfg.maintenance.repositories;
+
+      systemd.user.services."git-maintenance@" = {
+        Unit.Description = "Optimize Git repositories data";
```
h7x4, line 521:
```suggestion
        Unit = {
          Description = "Optimize Git repositories data";
          Documentation = [ "man:git-maintenance(1)" ];
        };
```

modules/programs/git.nix
PerchunPak: Good idea! Done [here](https://github.com/nix-community/home-manager/compare/39432cd042226d331207ccf22392264f7f0ed3c7..bcf8990114a66bfd992c4dce8da2c6d6ed8ae5f3)

modules/programs/git.nix
PerchunPak: [Done](https://github.com/nix-community/home-manager/compare/39432cd042226d331207ccf22392264f7f0ed3c7..bcf8990114a66bfd992c4dce8da2c6d6ed8ae5f3)

modules/programs/git.nix
```diff
@@ -511,6 +547,56 @@ in {
...
+          Install.WantedBy = [ "timers.target" ];
+        };
+      }
+      //
+      lib.attrsets.mapAttrs'
+        (n: v: lib.attrsets.nameValuePair "git-maintenance@${n}" {
+          Timer.OnCalendar = v;
+        })
+        cfg.maintenance.timers;
+>>>>>>> bcf89901 (git: add module for `git maintenance`)
```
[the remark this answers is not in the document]
PerchunPak: Oops, fixed

modules/programs/git.nix
PerchunPak: Looks like this produces an error `Unit name git-maintenance@.timer is missing the instance name.` because `sd-switch` tries to start `git-maintenance@.timer` and not `git-maintenance@something.timer`. Reported this to `sd-switch`'s [bug tracker](https://lists.sr.ht/~rycee/public-inbox/%3CCADPR+QweDcpom2h-6xv1+zRDjybcv7KMTagqQ-CBZJJrMEx7=A@mail.gmail.com%3E)

PR #5773, opened 2024-08-26
nix-direnv: Override the `nix-direnv` package with `nix` from `config.nix.package`

shivaraj-bh:
### Description

Override the `nix-direnv` package with `nix` from `config.nix.package` to match the behaviour with upstream nixpkgs: https://github.com/NixOS/nixpkgs/blob/6a9d18e057c1dbd6acefa6fe094de8f3c68051a2/nixos/modules/programs/direnv.nix#L64

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@rycee

### DIFF
--- a/modules/programs/direnv.nix
+++ b/modules/programs/direnv.nix
@@ -92,7 +92,17 @@ in {
         [nix-direnv](https://github.com/nix-community/nix-direnv),
         a fast, persistent use_nix implementation for direnv'';
 
-      package = mkPackageOption pkgs "nix-direnv" { };
+      package = mkOption {
+        default = pkgs.nix-direnv.override {
+          nix =
+            if config.nix.package == null then pkgs.nix else config.nix.package;
+        };
+        defaultText = "pkgs.nix-direnv";
+        type = types.package;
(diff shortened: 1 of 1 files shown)


Malix-Labs:
Should this PR still be merged?

modules/programs/direnv.nix
```diff
@@ -92,7 +92,17 @@ in {
         [nix-direnv](https://github.com/nix-community/nix-direnv),
         a fast, persistent use_nix implementation for direnv'';
 
-      package = mkPackageOption pkgs "nix-direnv" { };
+      package = mkOption {
+        default = pkgs.nix-direnv.override {
+          nix =
+            if config.nix.package == null then pkgs.nix else config.nix.package;
```
lafrenierejm, line 98:
```suggestion
        default = pkgs.nix-direnv.override {
          # Override the nix-direnv package with nix from config.nix.package to match the behaviour of nix-direnv NixOS module
          nix =
            if config.nix.package == null then pkgs.nix else config.nix.package;
```

PR #5776, opened 2024-08-27
Add Glab

shikanime:
Add Glab


shikanime:
Implemented by https://github.com/nix-community/home-manager/pull/8066

PR #5781, opened 2024-08-30
jujutsu: support darwin guidelines for config placement

jkachmar:
### Description

backports changes from https://github.com/nix-community/home-manager/pull/5416 to the `release-24.05` branch.

#### Maintainer CC

@shikanime

### DIFF
--- a/modules/programs/jujutsu.nix
+++ b/modules/programs/jujutsu.nix
@@ -7,6 +7,11 @@ let
   cfg = config.programs.jujutsu;
   tomlFormat = pkgs.formats.toml { };
 
+  configDir = if pkgs.stdenv.isDarwin then
+    "Library/Application Support"
+  else
+    config.xdg.configHome;
+
 in {
   meta.maintainers = [ maintainers.shikanime ];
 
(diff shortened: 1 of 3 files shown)


shikanime (approved):

jkachmar:
how does merging work for `home-manager` anyway? @rycee merged the original PR, but the UI says we should be able to merge this with an approval from a maintainer.

rycee:
Thanks! Rebased to the 24.05 branch in 3288b74c61b9cc09dd2d10854f5876b10d9f9383.

puppe:
I appreciate this change in the grand scheme of things. With jj ignoring XDG_CONFIG_HOME, this is definitely the right thing to do. But as I already had a workaround in place by setting JJ_CONFIG, this just broke my setup. It took me embarassingly long to find the cause  (in part because I had forgotten about setting JJ_CONFIG).

Personally, I would have preferred if this would not have been backported. If this had come with the next release with an accompanying entry in the release notes, that would probably have saved me a lot of time. (I hope I do not sound too cranky. It is getting late in UTC+0200.)

rycee:
@puppe Sorry about the wasted time! I think you have a good point about this being a breaking change and it shouldn't have made it into the stable branch.

Maybe it still would have been valuable to bring into the stable branch but perhaps with a news entry conditioned on jujutsu being enabled, then you at least would be notified about the change.

In any case, I will keep this in mind for the future.

Issue #5310, opened 2024-04-20
Upgrade `programs.taskwarrior` to use `taskwarrior3`


cloudlena:
Currently, `programs.taskwarrior.package` defaults to `pkgs.taskwarrior`. A new version of Taskwarrior was released, so we should update the default to `pkgs.taskwarrior3`.

aktaboot:
since the `taskwarrior` package is still in nixpkgs and people might not have migrated yet, I think we should still keep it to `pkgs.taskwarrior`. It can easily be set manually to taskwarrior3.

I don't think a breaking change should be imposed on users without notice

cloudlena:
Thanks, @aktaboot. I see your point.

I'm not sure how this is handled usually. Would you just wait for some time and then change it later, or how would you proceed?

aktaboot:
Yes, eventually it will roll out, but really unsure when.

mdietrich16:
What about `services.taskwarrior-sync`? AFAIK (still a newbie to Nix) there is no option to explicitly set the used taskwarrior pkg to `taskwarrior3` instead of `taskwarrior`.

aktaboot:
@mdietrich16  for taskwarrior3, you can just set `programs.taskwarrior.package = pkgs.taskwarrior3;`
for the sync server module, I have a thing that I should upstream soon

PR #5782, opened 2024-08-30
taskwarrior: support taskwarrior3 migration

trueNAHO:
### Description

Support taskwarrior3 migration, following the approach in https://github.com/NixOS/nixpkgs/pull/303632 to avoid breaking changes.

Closes: https://github.com/nix-community/home-manager/issues/5310

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

- `taskwarrior-sync.nix`
  - @minijackson
  - @pacien

- `taskwarrior.nix`
  - @Leixb
  - @x10an14

### DIFF
--- a/modules/programs/taskwarrior.nix
+++ b/modules/programs/taskwarrior.nix
@@ -85,7 +85,8 @@ in {
         '';
       };
 
-      package = mkPackageOption pkgs "taskwarrior" { };
+      package =
+        mkPackageOption pkgs "taskwarrior" { example = "pkgs.taskwarrior3"; };
     };
   };
 
--- a/modules/services/taskwarrior-sync.nix
+++ b/modules/services/taskwarrior-sync.nix
@@ -12,7 +12,8 @@ in {
   options.services.taskwarrior-sync = {
     enable = mkEnableOption "Taskwarrior periodic sync";
 
-    package = mkPackageOption pkgs "taskwarrior" { };
+    package =
+      mkPackageOption pkgs "taskwarrior" { example = "pkgs.taskwarrior3"; };
 
     frequency = mkOption {
       type = types.str;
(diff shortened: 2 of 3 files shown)


trueNAHO:
> The default package should perhaps be set to `taskwarrior3` for `stateVersion`s more recent than the current stable?

Users not already on taskwarrior3 receive the following warning:

```
trace: evaluation warning: taskwarrior was replaced by taskwarrior3, which requires manual transition from taskwarrior 2.6, read upstram's docs: https://taskwarrior.org/docs/upgrade-3/
```

Avoiding the breaking change of setting `pkgs.taskwarrior3` as the new default aligns with the approach in https://github.com/NixOS/nixpkgs/pull/303632. Until Nixpkgs hard deprecates taskwarrior2, we should follow their approach.

> This would be more consistent with what is being tested in the module test too.

Are you referring to the following:

```patch
diff --git a/tests/modules/programs/taskwarrior/taskwarrior.nix b/tests/modules/programs/taskwarrior/taskwarrior.nix
index f21198f5732b..ea6cde3a1316 100644
--- a/tests/modules/programs/taskwarrior/taskwarrior.nix
+++ b/tests/modules/programs/taskwarrior/taskwarrior.nix
@@ -16,6 +16,7 @@ with lib;
         include /my/stuff
         urgency.user.tag.test.coefficient=-42.42
       '';
+      package = pkgs.taskwarrior3;
     };

     test.stubs.taskwarrior = { };
```

I added this to verify that `pkgs.taskwarrior3` still exists if it ever gets replaced with `pkgs.taskwarrior` in the future. This makes it easier to spot when to revert this patch.

Leixb:
LGTM, I agree with following the nixpkgs approach.

I am not sure if setting a taskwarrior3 in the test is a good idea. In my opinion, it's better to ensure that the default option is valid than checking that the example option is.

rycee:
Thanks! I've adjusted for the previously merged taskwarrior-sync PR and merged this PR to master now 🙂

PR #5783, opened 2024-08-30
Qutebrowser: allow config to output Python dict

giodamelio:
### Description

Add a new special `__isDict` attribute to the qutebrowser settings code to allow it to output Python dicts.

I am not sure I love the solution really, but it works and is backwards compatable.

I think it would be great to instead serilize the config to something like toml or yaml then write some Python config to load that, but that is non trivial.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

There are no maintainers in the meta @somasis and @rycee it looks like you are recent commiters.

I also went ahead and added myself as a maintainer.

### DIFF
--- a/modules/programs/qutebrowser.nix
+++ b/modules/programs/qutebrowser.nix
@@ -6,25 +6,34 @@ let
 
   cfg = config.programs.qutebrowser;
 
+  formatValue = v:
+    if v == null then
+      "None"
+    else if builtins.isBool v then
+      (if v then "True" else "False")
+    else if builtins.isString v then
+      ''"${v}"''
+    else if builtins.isList v then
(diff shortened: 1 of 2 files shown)


giodamelio:
I am also not sure why the tests are not kicking off for this PR. Is there something I can do to manually trigger them?

somasis:
This looks good, though I do wish we could do it a little nicer... I know for GVariant we've got `mkDictionaryEntry`. If there was any other precedent for a module needing to output a dictionary like this I'd say a function to set `__isDict` or something would be nice, but I think in lieu of that, this is fine.

Maybe we could just do `isDict` even, since qutebrowser uses snake_case for options anyway?

PR #5788, opened 2024-09-01
process-compose: add module

dudeofawesome:
### Description

Adding a module to configure [process-compose](https://f1bonacc1.github.io/process-compose/).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


dudeofawesome:
@rycee Not sure if you're the right person to ping, but wanted to make sure this was seen.

modules/programs/process-compose.nix
```diff
@@ -0,0 +1,83 @@
...
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home = {
+      packages = [ cfg.package ];
+
+      file.process-compose-settings = {
+        enable = !builtins.isNull cfg.settings;
```
teto, line 63: `!= null` might be shorter

modules/programs/process-compose.nix
```diff
@@ -0,0 +1,77 @@
...
+        Check [the documentation](https://f1bonacc1.github.io/process-compose/tui/#tui-themes) for available options.
+      '';
+    };
+  };
+
+  config = let
+    configDir = if (pkgs.stdenv.targetPlatform.isDarwin) then
+      "Library/Application Support/process-compose/"
+    else
+      "$XDG_CONFIG_HOME/process-compose/";
```
dudeofawesome, line 51: It looks like the [linux tests are failing](https://github.com/nix-community/home-manager/actions/runs/13173123866/job/36766898135?pr=5788). I wonder if it could be caused by the `XDG_CONFIG_HOME` var not working here?

PR #5789, opened 2024-09-01
sesh: add module

michaelvanstraten:
### Description

Sesh is a CLI that helps you create and manage tmux sessions quickly and easily using zoxide. See https://github.com/joshmedeski/sesh for more.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/sesh.nix
+++ b/modules/programs/sesh.nix
@@ -1,0 +1,86 @@
+{ pkgs, lib, config, ... }:
+
+let
+  inherit (lib) mkEnableOption mkIf mkMerge mkOption mkPackageOption types;
+
+  cfg = config.programs.sesh;
+  tomlFormat = pkgs.formats.toml { };
+in {
+  meta.maintainers = [ lib.hm.maintainers.michaelvanstraten ];
+
+  options.programs.sesh = {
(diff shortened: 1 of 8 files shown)


michaelvanstraten:
@KostaGorod, could you review this?

KostaGorod (approved):
:3 @michaelvanstraten once conflicts resolved lemme know

michaelvanstraten:
:ping @KostaGorod

modules/programs/sesh.nix
```diff
@@ -0,0 +1,72 @@
+{ pkgs, lib, config, ... }:
+with lib;
```
khaneliman, line 2:
```suggestion
```

bbigras:
Did you folks consider using `--icons` with `sesh list` by default?

cronyakatsuki:
> Did you folks consider using `--icons` with `sesh list` by default?

Same thought, using` --icons` by default or adding an option for allowing icons would be nice, also the preview window in fzf which is very useful in some cases.

michaelvanstraten:
> Did you folks consider using `--icons` with `sesh list` by default?

No, sorry. I did not know about it when adding the module. I will make a PR to add it as a default.

Thank you for pointing it out.

Issue #5483, opened 2024-06-02
add golden test tooling to help update tests


teto:
I was preparing a PR that updates the test and now I wonder how people update the tests because I find it tedious. Either you copy the generated result back into the folder or you edit the snapshot right into the tree. If the tests change a lot, it can become tedious.

What we do is called "snapshot testing" or "golden testing". There are some tools to help with that for instance you are presented with a diff and can choose the generated file as the new reference snapshot.

For a personal project I was quite happy with: https://github.com/ReplicaTest/REPLica 
More recently in the nix area, there is https://github.com/nix-community/namaka but I wonder if it fits our scenario.

PR #5793, opened 2024-09-02
Move aliases after initializations in fish

FaustXVI:
### Description

Move the aliases declaration in fish configuration to appear after the initialization as some programs (like zoxide) may need initialization before being used in an alias.

### Checklist

#### Maintainer CC

No maintainer @rycee ? If so I may apply (it would improve my `nix`)

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -471,11 +471,11 @@ in {
           # Abbreviations
           ${abbrsStr}
 
-          # Aliases
-          ${aliasesStr}
-
           # Interactive shell initialisation
           ${cfg.interactiveShellInit}
+
+          # Aliases
+          ${aliasesStr}
 
         end
 


FaustXVI:
I would like to add a test on that but I have no idea of how to generate the golden file.

My ideal would be : 
 - write the test I want and see it fail
 - the failure has generated the current state of the file, get it, fix it and make it the new golden file
 - see the test pass

@rycee is there a way to do that ?
Thanks

FaustXVI:
does anyone has an idea of why the checks doesn't run ?

dawidd6:
Might be some GitHub hiccup. Try rebasing on master and force pushing.

teto:
> I would like to add a test on that but I have no idea of how to generate the golden file.
>     * write the test I want and see it fail
>     * the failure has generated the current state of the file, get it, fix it and make it the new golden file

to the best of my knowledge that's how it's done so far see https://github.com/nix-community/home-manager/issues/5483.

> as some programs (like zoxide) may need initialization before being used in an alias.

fish can't declare an alias for an existing program ? curious to see an example that this PR fixes.

PR #5795, opened 2024-09-02
nushell: Allow arbitrary env vars

JoaquinTrinanes:
### Description

This PR allows setting arbitrary values as nushell's environment variables.

This started as a fix for [some value assigments now requiring quotes](https://www.nushell.sh/blog/2024-08-20-nushell_0_97_1.html#more-consistent-parsing-of-assignment-toc) (including `$env.FOO = bar`, which now needs to be `$env.FOO = "bar"`). Allowing any JSON value is just a secondary effect :)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@Philipp-M

### DIFF
--- a/modules/programs/nushell.nix
+++ b/modules/programs/nushell.nix
@@ -145,11 +145,17 @@ in {
     };
 
     environmentVariables = mkOption {
-      type = types.attrsOf types.str;
+      type = types.attrsOf (pkgs.formats.json { }).type;
       default = { };
-      example = { FOO = "BAR"; };
+      example = {
+        FOO = "BAR";
+        BOOLEAN_VAR = true;
+        NUMERIC_VAR = 4;
+        LIST_VAR = [ "elem1" 2 ];
+        OBJECT_VAR = { some = "field"; };
+      };
       description = ''
-        An attribute set that maps an environment variable to a shell interpreted string.
+        Environment variables that will be set.
       '';
     };
   };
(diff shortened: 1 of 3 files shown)


modules/programs/nushell.nix
```diff
@@ -145,11 +145,11 @@ in {
     };
 
     environmentVariables = mkOption {
-      type = types.attrsOf types.str;
+      type = types.attrsOf (pkgs.formats.json { }).type;
       default = { };
       example = { FOO = "BAR"; };
```
Philipp-M, line 150: I think you could add something like, so that it's obvious that more than just a string is supported:
```suggestion
      example = {
        FOO = "BAR";
        BOOLEAN_VAR = true;
        NUMERIC_VAR = 4;
        LIST_VAR = [ "elem1" 2 ];
      };
```

(could also be something more descriptive/real-world-use)

Philipp-M (approved):
Sorry, was on holidays.

I guess similar as in the other PR (sorry that I haven't reviewed that yet), you could add yourself as maintainer.

Otherwise I think it looks good, thanks.

Environment variables not being strings seems weird at first, but I see that nushell supports this, so I think it's good to support this via nix as well.

JoaquinTrinanes:
I just realized this is a breaking change: embedding nu expressions into the config isn't possible anymore.

So this isn't valid anymore, because it will be just a string:
```nix
environmentVariables.PROMPT_INDICATOR = "{|| '> ' }";
```

I think a possible fix for this is creating a `mkNushellInline` and `toNushell` lib functions (like [`nixpkgs`'s lua helpers](https://github.com/NixOS/nixpkgs/blob/1da2d8d992771d5a141765db06193df16f44f13e/lib/generators.nix#L771)) that handle the nix -> nu transformation.

The other trivial fix is not doing anything, but some (most?) configurations that use string variables expecting them to become strings and not raw nushell code will break.

nyabinary:
> I just realized this is a breaking change: embedding nu expressions into the config isn't possible anymore.
> 
> So this isn't valid anymore, because it will be just a string:
> 
> ```nix
> environmentVariables.PROMPT_INDICATOR = "{|| '> ' }";
> ```
> 
> I think a possible fix for this is creating a `mkNushellInline` and `toNushell` lib functions (like [`nixpkgs`'s lua helpers](https://github.com/NixOS/nixpkgs/blob/1da2d8d992771d5a141765db06193df16f44f13e/lib/generators.nix#L771)) that handle the nix -> nu transformation.
> 
> The other trivial fix is not doing anything, but some (most?) configurations that use string variables expecting them to become strings and not raw nushell code will break.

I mean, this is still pretty important and shouldn't be left to rot imo

rycee:
With #5905 merged, I think it is safe to close this? Please comment if that is a mistake.

PR #5801, opened 2024-09-03
Set MOZ_* session variables for firefox and thunderbird modules

booxter:
- **thunderbird: set MOZ_* variables for legacy profiles.ini**
- **firefox: set MOZ_* variables for legacy profiles.ini**

### Description

This PR includes the following changes:

1. Make thunderbird set MOZ_* session variables same as in nixpkgs wrapped package;
2. Make firefox do the same;
3. Deprecate the darwin warning message from thunderbird module about the need to set the variables, since now it's happening automatically;
4. Introduce a launchd service that will apply the session variables to GUI session environment with `launchctl setenv`.
5. Now that non-nixpkgs firefox can now work with Version=2 in profiles.ini thanks to MOZ_* variables, revert patches that introduced a new special option for Darwin to unset Version in profiles config file.

Closes: #5717

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee @kira-bruneau @bricked @d-dervishi @jkarlson

### DIFF
--- a/modules/launchd/default.nix
+++ b/modules/launchd/default.nix
@@ -209,5 +209,21 @@ in {
           setupLaunchAgents
         '';
     })
+
+    (mkIf isDarwin (let
+      launchctl-setenv = pkgs.writeShellScriptBin "launchctl-setenv"
+        (concatStringsSep "\n" (mapAttrsToList
+          (name: val: "/bin/launchctl setenv ${name} ${toString val}")
+          config.home.sessionVariables));
+    in {
+      launchd.agents.launchctl-setenv = {
+        enable = true;
+        config = {
+          ProgramArguments = [ "${launchctl-setenv}/bin/launchctl-setenv" ];
+          KeepAlive.SuccessfulExit = false;
+          RunAtLoad = true;
+        };
+      };
+    }))
   ];
 }
--- a/modules/programs/firefox/mkFirefoxModule.nix
+++ b/modules/programs/firefox/mkFirefoxModule.nix
@@ -62,7 +62,7 @@ let
     }) // {
       General = {
         StartWithLastProfile = 1;
-      } // lib.optionalAttrs (cfg.profileVersion != null) {
+      } // lib.optionalAttrs (cfg.profileVersion == null) {
         Version = cfg.profileVersion;
       };
     };
--- a/modules/programs/firefox/mkFirefoxModule.nix
+++ b/modules/programs/firefox/mkFirefoxModule.nix
@@ -62,8 +62,7 @@ let
     }) // {
       General = {
         StartWithLastProfile = 1;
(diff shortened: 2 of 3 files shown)


booxter:
Ah this is not working either, because the variables are not exposed to GUI apps. (So it fixes firefox only when executed from a shell it seems.)

booxter:
Combined with https://github.com/nix-community/home-manager/pull/5808 this works fine.

booxter:
@rycee This PR reverts some of firefox patches for profiles.ini Version= field removal on Darwin that were merged recently. Please take a look if this is a good approach.

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -799,6 +791,12 @@ in {
           force = true;
         };
     }));
+
+    # Mimic nixpkgs package environment for read-only profiles.ini management
+    home.sessionVariables = {
+      MOZ_LEGACY_PROFILES = 1;
+      MOZ_ALLOW_DOWNGRADE = 1;
```
rycee, line 798: I'm not entirely certain this is safe since it will apply across the entire user session. E.g., it would presumably also apply for a secondary Firefox install that is not managed through Home Manager. Which perhaps causes unexpected behavior.

Setting these variables in a wrapper like in the main Nixpkgs firefox package seems safer.

modules/launchd/default.nix
```diff
@@ -209,5 +209,28 @@ in {
...
+        (concatStringsSep "\n" (mapAttrsToList
+          (name: val: "/bin/launchctl setenv ${name} ${toString val}")
+          config.home.sessionVariables));
+      launchctl-setenv = pkgs.stdenv.mkDerivation {
+        name = "setup-script";
+        buildCommand = ''
+          mkdir -p $out/bin
+          cp ${setenv-script} $out/bin/launchctl-setenv
+        '';
+      };
```
rycee, line 224: Should be replaceable by

```suggestion
      launchctl-setenv = pkgs.writeShellScriptBin "launchctl-setenv"
        (concatStringsSep "\n" (mapAttrsToList
          (name: val: "/bin/launchctl setenv ${name} ${toString val}")
          config.home.sessionVariables));
      };
```

modules/programs/firefox/mkFirefoxModule.nix
booxter: That may be correct for nixpkgs derivation (though there is no wrapped flavor for Darwin, at the moment at least). For homebrew installations that a lot of people use (nixpkgs ports for Darwin are very much new and still not complete - like the wrapper part that is missing), nixpkgs solution wouldn't help.

booxter:
With wrapped firefox and thunderbird supported by nixpkgs, this becomes less needed. I am no longer interested to pursue this.

PR #5806, opened 2024-09-05
eza: add support for fish abbreviations

HeitorAugustoLN:
### Description

Added support for fish abbreviations instead of aliases in `enableFishIntegration`

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@cafkafk

### DIFF
--- a/modules/programs/eza.nix
+++ b/modules/programs/eza.nix
@@ -91,8 +91,17 @@ with lib;
     programs.zsh.shellAliases = optionsAlias
       // optionalAttrs cfg.enableZshIntegration aliases;
 
-    programs.fish.shellAliases = optionsAlias
-      // optionalAttrs cfg.enableFishIntegration aliases;
+    programs.fish = mkMerge [
+      (mkIf (!config.programs.fish.preferAbbrs) {
+        shellAliases = optionsAlias
+          // optionalAttrs cfg.enableFishIntegration aliases;
+      })
+
+      (mkIf config.programs.fish.preferAbbrs {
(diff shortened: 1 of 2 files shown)


modules/programs/eza.nix
```diff
@@ -29,8 +29,16 @@ with lib;
       default = true;
     };
 
-    enableFishIntegration = mkEnableOption "Fish integration" // {
-      default = true;
+    enableFishIntegration = mkOption {
+      type = types.enum [ "no" "aliases" "abbreviations" ];
```
rycee, line 33: Can keep backwards compatibility

```suggestion
      type = with types; either bool (enum [ "no" "aliases" "abbreviations" ]);
      apply = v: if isBool v then if v then "aliases" else "no" else v;
```

rycee:
Thanks for the contribution! I wonder if this is specific for this module or should there be an option under `programs.fish` for specifying preference?

Something like `programs.fish.preferAbbrevs`?

HeitorAugustoLN:
> Thanks for the contribution! I wonder if this is specific for this module or should there be an option under `programs.fish` for specifying preference?
> 
> Something like `programs.fish.preferAbbrevs`?

That is indeed an better approach. Will adapt to it.

This is an better approach

modules/programs/eza.nix
HeitorAugustoLN: Change is now backwards-compatible.

HeitorAugustoLN:
I will adapt other modules that have fish aliases in another PR.

PR #5808, opened 2024-09-06
darwin: Set launchd.user.envVariables from home.sessionVariables

booxter:
This allows to have consistent environment for both (gui) apps started
by launchd and shell programs.

Closes: https://github.com/LnL7/nix-darwin/issues/1056
Signed-off-by: Ihar Hrachyshka <ihar.hrachyshka@gmail.com>

### Description

In [some](https://github.com/nix-community/home-manager/pull/5801) scenarios, a
module may want to inject special variables in app runtime. This is usually
achieved through setting `sessionVariables`. But on MacOS, the setting doesn't
apply to apps when they are started using Mac UI (not from a terminal). For the
setting to apply there, we have to also configure user `launchd` variables.

With this patch, and the patches linked above, we can now run Firefox and
Thunderbird on Darwin without any additional configuration: the needed MOZ_*
variables are injected into `sessionVariables` by the app modules; and then
Darwin specific integration copies these values into `launchd` environment.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/nix-darwin/default.nix
+++ b/nix-darwin/default.nix
@@ -11,21 +11,28 @@ in {
 
   config = mkMerge [
     { home-manager.extraSpecialArgs.darwinConfig = config; }
-    (mkIf (cfg.users != { }) {
-      system.activationScripts.postActivation.text = concatStringsSep "\n"
-        (mapAttrsToList (username: usercfg: ''
-          echo Activating home-manager configuration for ${username}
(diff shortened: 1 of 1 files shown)


booxter:
The work is adopted and included in the PR that fixes Thunderbird: https://github.com/nix-community/home-manager/pull/5808

PR #5812, opened 2024-09-08
firefox: correct lastUserContextId output from firefox containers

Faeranne:
### Description

lastUserContextId should match the highest context id from the containers set in a given profile.  mapAttrsToList uses the container name for sorting.  foldlAttrs ensures we get the largets commit id.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@brckd @rycee @kira-bruneau

### DIFF
--- a/modules/programs/firefox/mkFirefoxModule.nix
+++ b/modules/programs/firefox/mkFirefoxModule.nix
@@ -102,7 +102,8 @@ let
       ${builtins.toJSON {
         version = 4;
         lastUserContextId =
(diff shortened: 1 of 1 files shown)


Faeranne:
this should fix #5812
I didn't see any tests that this should match up with, but all existing tests pass.  If I need to add a test for this I can try, though I'm not sure how important that is for this single element.

rycee:

brckd:
Closes #5812.

PR #5817, opened 2024-09-08
swayidle: make -w optional

leoTlr:
### Description

The option -w causes swayidle to wait on timeouts/events to finish. This can cause problems with certain timeout/event commands (see https://github.com/swaywm/swaylock/issues/86#issuecomment-662702180).

This PR lets the user choose if he wants to use the option. It defaults to the current behavior.

### Checklist

-> swayidle-basic-configuration: OK

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@c0deaddict

### DIFF
--- a/docs/release-notes/rl-2411.md
+++ b/docs/release-notes/rl-2411.md
@@ -7,7 +7,16 @@ is therefore not final.
 
 This release has the following notable changes:
 
-- No changes.
+- The swayidle module behavior has changed. Specifically, swayidle was
+  previously always called with a `-w` flag. This flag is now moved to
+  the default
+  [services.swayidle.extraArgs](#opt-services.swayidle.extraArgs)
+  value to make it optional.
+
+  Your configuration may break if you already set this option and also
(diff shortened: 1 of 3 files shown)


teto:
oh I noticed the double lock sometimes too but too lazy to lock into. Rather than adding maintainance overhead with new options, I would just default cfg.extraArgs to `[ "-w" ]` with possibly a comment and a notification in news.nix.
`

leoTlr:
Applied suggestions.

I even had it with extraArgs before but then decided against it to not break user configs. 
With the news entry it should be fine - didnt think of that.

teto:
@rycee maybe we can make macos optional ? it feels like it fails more often than other platforms (true for nixpkgs as well). Or maybe better make it possible to merge even with failures (often the PR is not responsible for the failures)?

rycee:

@teto Yeah, the Darwin tests are a big pain. I was hoping somebody would have a look at fixing them but perhaps it would be best to simply disable them for now.

PR #5822, opened 2024-09-11
swayidle: Fix service ordering

e-tho:
### Description

Fix service ordering by ensuring that the `swayidle` unit is started after the compositor.

### Checklist

```
{component}: {description}

    {long description}
```

#### Maintainer CC

@c0deaddict


Pacman99:
I think Requisite=graphical-session.target is needed to. I've noticed that the After dependency doesn't have an effect for me without setting Requisite. I'm not entirely sure why that's the case.

PR #5826, opened 2024-09-12
evil-helix: add module

chetgurevitch:
Copied from helix

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@Philipp-M This removes the comment about the language-server option requiring the master branch from the upstream helix module since it was released in 23.10

### DIFF
--- a/modules/lib/maintainers.nix
+++ b/modules/lib/maintainers.nix
@@ -107,6 +107,12 @@
     githubId = 32838899;
     name = "Daniel Wagenknecht";
   };
+  gurevitch = {
+    name = "Chet Gurevitch";
+    email = "chet@gurevitch.net";
+    github = "chetgurevitch";
+    githubId = 25513187;
+  };
   henrisota = {
     email = "henrisota@users.noreply.github.com";
     github = "henrisota";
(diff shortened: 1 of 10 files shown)


Philipp-M:
I did a diff between helix and this module, and it seems that the only change is, the name of the module.

Do you think it would make sense to just add an example in the helix module maybe something like:

```nix
...
    package = mkOption {
      type = types.package;
      default = pkgs.helix;
      defaultText = literalExpression "pkgs.helix";
      example = literalExpression "pkgs.evil-helix"; # <--- here
      description = "The package to use for helix.";
    };
...
```

to show how to use this with evil-helix?

I don't think it's a good idea to have this as a separate module, when `evil-helix` is otherwise compatible with the helix module. (having to sync these modules, duplicate maintenance etc.)

So in your case it may be enough to just do `programs.helix.package = pkgs.evil-helix` in your home-manager config to achieve the same.

Thanks for the removal of that comment, that should likely be merged, I forgot about that.

chetgurevitch:
That seems like a better solution, missed that option somehow. I'll update the PR to just remove the comment

Issue #5849, opened 2024-09-16
A secret/unintuitive behavior of xdg.*Home options (they can set XDG_*_HOME env vars)


Andrew15-5:
Hi, I had a question about XDG env vars that I asked at https://matrix.to/#/#users:nixos.org. I wanted to know the best way to set, e.g., `xdg.configHome` from the set `environment.sessionVariables.XDG_CONFIG_HOME`. But then people quickly told me a secret that `xdg.configHome` itself does this already! I was shocked, especially because [the documentation](https://github.com/nix-community/home-manager/blob/a9c9cc6e50f7cbd2d58ccb1cd46a1e06e9e445ff/modules/misc/xdg.nix#L50) doesn't say _anything_ about that. Only the [`xdg.enable`](https://github.com/nix-community/home-manager/blob/a9c9cc6e50f7cbd2d58ccb1cd46a1e06e9e445ff/modules/misc/xdg.nix#L25) says _something_ related to the XDG.

I think that it would be great if the concrete options a user searches for explain/mention this behavior. Then there should be no confusion and the `environment.sessionVariables.XDG_CONFIG_HOME` should be unset (generally speaking).

PR #5850, opened 2024-09-16
docs: add XDG_*_HOME mentions to xdg.*Home options

Andrew15-5:
### Description

Fixes https://github.com/nix-community/home-manager/issues/5849.

One thing that I'm not user I should include in the docs is when the XDG var is actually applied. I tried changing the env vars a few times in a GNOME Wayland VM and the majority of the time they are only applied after system reboot. I think there were a couple of times when they are applied after re-login.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/xdg.nix
+++ b/modules/misc/xdg.nix
@@ -30,6 +30,8 @@ in {
       apply = toString;
       description = ''
         Absolute path to directory holding application caches.
+
+        Sets `XDG_CACHE_HOME` for the user if `xdg.enable` is set `true`.
       '';
     };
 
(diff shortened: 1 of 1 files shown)


teto (approved):

teto (commented):
I wonder if we want to add more info about how this works (needs session reloading, source ~/.profile etc) in the overall level.

We could improve the description of:
```
enable = mkEnableOption "management of XDG base directories";
```
to explain what it does globally with a link towards XDG spec

rycee:
I agree that it would be possible to improve this quite a lot. One thing I've been thinking about occasionally is to make the activation script detect when a session variable has changed and print a warning saying that the user needs to relog to fully apply the change.

In any case, I think this change by itself is quite useful so will merge as-is. Additional improvements could be made in separate PRs 🙂

PR #5880, opened 2024-09-22
direnv: update for new nushell behavior

K900:
### Description

Nushell 0.98 makes it so external commands returning non-zero is considered an exception. Wrap the direnv invocation in a try-catch, so that even if direnv fails, we get the message from direnv, but not the traceback from nushell.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/direnv.nix
+++ b/modules/programs/direnv.nix
@@ -140,9 +140,15 @@ in {
             $env.config.hooks.pre_prompt?
             | default []
             | append {||
-                let direnv = (${getExe cfg.package} export json
-                | from json
-                | default {})
+                let direnv = (
+                    try {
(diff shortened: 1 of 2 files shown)


K900:
CC @happysalada as resident nushell person I guess

K900:
The test failure seems unrelated?

happysalada:
Indeed the failure looks like a compilation failure on darwin arm, so unrelated.

The PR looks good to me, thank you for this.

sumnerevans (approved):

sumnerevans:
I'm unable to merge due to the test failure.

K900:
Let's see if it helps if I rebase it on latest

K900:
It did, all green now.

PR #5881, opened 2024-09-22
fix: respect username setting of home-manager in nix-darwin activation script

Lynnaut:
#### Maintainer CC
@ElvishJerricco @ncfavier @toonn

### DIFF
--- a/nix-darwin/default.nix
+++ b/nix-darwin/default.nix
@@ -14,9 +14,9 @@ in {
     (mkIf (cfg.users != { }) {
       system.activationScripts.postActivation.text = concatStringsSep "\n"
         (mapAttrsToList (username: usercfg: ''
-          echo Activating home-manager configuration for ${username}
-          sudo -u ${username} --set-home ${
-            pkgs.writeShellScript "activation-${username}" ''
+          echo Activating home-manager configuration for ${usercfg.home.username}
+          sudo -u ${usercfg.home.username} --set-home ${
+            pkgs.writeShellScript "activation-${usercfg.home.username}" ''
               ${lib.optionalString (cfg.backupFileExtension != null)
               "export HOME_MANAGER_BACKUP_EXT=${
                 lib.escapeShellArg cfg.backupFileExtension


toonn (commented):
How does the `username` argument differ from `usercfg.home.username`? Is the former an attribute key from the `home-manager.users` set?

I'm not sure what issue this is trying to address. On the surface it seems to be that the behavior might be useful, you could pass the same home-manager configuration for different users. If that's not the case I feel like an assert that `username` and `usercfg.home.username` are equal might be appropriate, potentially inheriting the `username` into the `usercfg` if it's not set there.

Lynnaut:
Yes indeed, the current `username` argument is derived from the `home-manager.users` attribute set instead of using the `usercfg.home.username` attribute which defaults to the name of the set.
This change allows reusing an existing config over multiple machines with just overriding the username and homeDirectory from inside home-manager.

toonn:
@Lynnaut, you seem to imply that the behavior is opposite to my intuition. I'm talking about passing a single `usercfg.home` to multiple `home-manager.users`, so having the latter override the username as it does now seems useful.

AFAIU this PR would override the username from the set in `home-manager.users` with the one in the `usercfg.home` and therefore make that pattern create multiple users with the same username?

leona-ya:
This is quite interesting. I agree to your use case @tooon, but it's a bit confusing that `home.username` isn't respected here in contrast to almost anywhere. 

In `nixos/common.nix:34` it's defined that `home.username` defaults to the attrset key.  I _think_ it would be more consistent to use `home.username` (as this PR proposes)?

toonn:
@leona-ya, does that mean leaving `usercfg.home.username` unset falls back to using the key from `home-manager.users`?

networkException:
I'm in favor of treating `home.username` as the source of truth for the username here, from what I can tell the main (?) activation script does the same: https://github.com/nix-community/home-manager/blob/09a0c0c02953318bf94425738c7061ffdc4cba75/modules/home-environment.nix#L716

> In nixos/common.nix:34 it's defined that home.username defaults to the attrset key. I think it would be more consistent to use home.username (as this PR proposes)?

https://github.com/nix-community/home-manager/blob/09a0c0c02953318bf94425738c7061ffdc4cba75/nixos/common.nix#L34

This line is saying that when using home-manager from NixOS, the `home.username` attribute defaults to NixOS' [`users.users.<name>.name`](https://search.nixos.org/options?channel=24.05&show=users.users.%3Cname%3E.name&from=0&size=50&sort=relevance&type=packages&query=users.users.%3Cname%3E.name), so the user defined in the system. nix-darwin also has [such an option](https://github.com/LnL7/nix-darwin/blob/a60ac02f9466f85f092e576fd8364dfc4406b5a6/modules/users/user.nix#L7) to which `home.username` should probably default to, but I feel like that's out of scope for what this pr is trying to achieve.

toonn (approved):
Consensus seems to be in favor of the change despite my concern.

toonn:
@toodeluna, since you are currently interested in this, maybe you can answer the question. Would the following still work?

```nix
let user-config = "a full config that's intented to be re-used for multiple users";
in
{ users.users = {
    user1 = user-config;
    user2 = user-config;
}
```

toodeluna:
Found some more issues in nix-darwin while testing this but along with those changes (which can be found in [this PR](https://github.com/LnL7/nix-darwin/pull/1355)) this should now be working properly.

I tested the changes using this code snippet:

```nix
users.users =
let
  userConfig = { createHome = true; shell = pkgs.zsh; };
in
{
  user1 = userConfig // { uid = 502; home = "/Users/user1"; };
  user2 = userConfig // { uid = 503; home = "/Users/user2"; };
  yetAnotherUser = userConfig // { name = "user3"; uid = 504; home = "/Users/user3"; };
};

users.knownUsers = [ "user1" "user2" "user3" ];
nix.settings.trusted-users = [ "user1" "user2" "user3" ];

home-manager = {
  users.user1.home.stateVersion = "25.05";
  users.user2.home.stateVersion = "25.05";
  users.yetAnotherUser.home.stateVersion = "25.05";
};
```

This creates three users: "user1", "user2" and "user3". User3 is managed under `users.users.yetAnotherUser` and `home-manager.users.yetAnotherUser`. The setup for user1 and user2 will still work without any extra changes, but the "yetAnotherUser" example will only work once both this and the nix-darwin PR have been merged.

This isn't exactly like the example you specified since as far as I know there's no way to have a single shared user config in nix-darwin because every user needs to have fields like `uid` and `home` set, but I made it as close to what you wrote as possible.

PR #5886, opened 2024-09-23
zsh: Add extraPackages option

benjamb:
### Description

Certain plugins such as `bgnotify` expect certain binaries from system packages to be available on zsh's PATH. Add support for 'extraPackages' so additional packages can be added as needed.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@teto

### DIFF
--- a/modules/programs/zsh.nix
+++ b/modules/programs/zsh.nix
@@ -288,6 +288,13 @@ in
 
       package = mkPackageOption pkgs "zsh" { };
 
+      extraPackages = mkOption {
+        type = with types; listOf package;
+        example = literalExpression "with pkgs; [ libnotify ]";
+        description = "Extra packages available to zsh.";
+        default = [ ];
+      };
+
       autocd = mkOption {
         default = null;
         description = ''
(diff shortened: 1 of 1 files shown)


teto:
I dont think it adds much over just using `home.packages` so I would rather avoid the overhead.

benjamb:
@teto Fair enough. I was originally thinking of doing something along the lines of what `helix` does, where it creates a wrapper with the relative packages added to `PATH`, since I don't like adding things to the bin directories on my default PATH that I'm unlikely to ever explicitly call via the terminal (e.g. unnecessary noise for tab completion), but I couldn't get that to work with the plugin I was enabling this functionality for.

teto:
we have a similar problem (soon solved) with neovim plugins. Ideally this would be solved at the nixpkgs level, with zsh plugins setting their required runtime inputs. Wrapping zsh with those paths will leak to almost every program, equivalent to adding the runtime dependencies to home.packages so except if you have a very strong usecase/motivation, home.packages looks good enough

PR #5895, opened 2024-09-26
direnv: even better nushell fix

K900:
### Description

Avoid do | complete, because it swallows stderr, which can contain direnv debug output.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/direnv.nix
+++ b/modules/programs/direnv.nix
@@ -143,9 +143,9 @@ in {
                 let direnv = (
                     # We want to get the stdout from direnv even if it exits with non-zero,
                     # because it will have the DIRENV_ internal variables defined.
-                    do { ${getExe cfg.package} export json }
-                    | complete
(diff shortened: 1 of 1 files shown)


modules/programs/direnv.nix
```diff
@@ -143,9 +143,7 @@ in {
                 let direnv = (
                     # We want to get the stdout from direnv even if it exits with non-zero,
                     # because it will have the DIRENV_ internal variables defined.
-                    do { ${getExe cfg.package} export json }
-                    | complete
-                    | get stdout
+                    do --ignore-program-errors { ${getExe cfg.package} export json }
```
rycee, line 146: Looks OK to me but CI is complaining about formatting:

    /home/runner/work/home-manager/home-manager/modules/programs/direnv.nix: not formatted

Issue #5527, opened 2024-06-10
bug: nushell module configuration overwrites default basic configuration


nyabinary:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

So the home-manager module for nushell is actually destructive, which is not an idea as it overwrites the original nushell configuration which most users probably don't want since it breaks nushell. Rather, users want to most likely append the config to add their config.
![image](https://github.com/nix-community/home-manager/assets/97130632/0c0aa63d-b47a-4d67-a916-c2429e8a38a4)
![image](https://github.com/nix-community/home-manager/assets/97130632/8466d1c1-7892-4db5-a5b0-d59110b7112e)
My config for clarity is:
```nix
    programs.nushell = {
      enable = true;
      environmentVariables = {
        NIXOS_OZONE_WL = "1";
        ELECTRON_OZONE_PLATFORM_HINT = "auto";
        EDITOR = "hx";
        VISUAL = "hx";
        TERM = "rio";
      };
      shellAliases = {
        switch = "sudo nixos-rebuild switch";
        boot = "sudo nixos-rebuild boot";
      };
      extraConfig = ''
        $env.config = {
          show_banner: false,
        }
      '';
    };
```

### Maintainer CC

@Philipp-M 

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.9.3, NixOS, 24.11 (Vicuña), 24.11.20240607.051f920`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.23.0pre20240526_7de033d6`
 - nixpkgs: `/nix/store/3dr5pyja36lvvrszhzffww1jwyrx6i09-source`
```
```

Philipp-M:
I do think this is intended behavior as the name of either `programs.nushell.{envFile,configFile}`, suggests that the whole file is written, and we cannot cleanly (side-effects) just extend the default file at that path, which may not even exist when nushell was not yet started, and the user confirmed creation of such files.

But it's a little bit unfortunate, that nushell somewhat "relies" on a default file at that path as well to provide the expected UX.
Instead of using its own (internally hardcoded) defaults that can be overwritten with the user-specified file...
It's also not optimal that `enableNushellIntegration` in a few other programs seems to set `extraEnv/extraConfig` which in turn leads to writing the `envFile/configFile`.

[...]

PR #5896, opened 2024-09-26
direnv: hopefully final nushell fix

K900:
### Description

I hate this.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/direnv.nix
+++ b/modules/programs/direnv.nix
@@ -130,46 +130,64 @@ in {
         ${getExe cfg.package} hook fish | source
       '');
 
-    programs.nushell.extraConfig = mkIf cfg.enableNushellIntegration (
+    programs.nushell.extraConfig = mkIf cfg.enableNushellIntegration (let
+      # We want to get the stdout from direnv even if it exits with non-zero,
+      # because it will have the DIRENV_ internal variables defined.
+      #
+      # However, nushell's current implementation of try-catch is subtly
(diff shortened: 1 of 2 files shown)


rycee:
I fixed the test case and merged to master now. Hope it works better now! Have to go sleep so will check back tomorrow.

nyabinary:
This breaks nushell as it overrides the default env.nu breaking the behavior of nushell in general isn't there a way to just append the direnv stuff instead of overwriting the entire env.nu file?

K900:
No it doesn't? That's why the mkAfter is there.

nyabinary:
> No it doesn't? That's why the mkAfter is there.

https://github.com/nix-community/home-manager/issues/5527
https://github.com/nushell/nushell/issues/13174

Does for me I just want to add a alias and stuff in my config and it overrides the entire default env.nu already in place instead of appending to it.

K900:
That's unrelated to this change though?

nyabinary:
> That's unrelated to this change though?

But the direnv thing is broken *in general* because if you enable the direnv support for nushell it currently breaks nushell:
image
Pops up a nu::shell::non_zero_exit_code when normally its not supposed to do that :P

K900:
That has nothing to do with direnv, and it _is_ supposed to do that starting with Nushell 0.98 (see the original chain of PRs)

Issue #3864, opened 2023-04-11
bug: `services.gpg-agent` not supported for Apple Silicon M1


shackra:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

trying to install home-manager on an Macbook Pro with Apple Silicone M1 chip:

```shell
% nix-shell '<home-manager>' -A install --show-trace
The file /Users/jorge/.config/home-manager/home.nix already exists, leaving it unchanged...

Creating initial Home Manager generation...

error:
       Failed assertions:
       - The module services.gpg-agent does not support your platform. It only supports

         - aarch64-linux
         - armv5tel-linux
         - armv6l-linux
         - armv7a-linux
         - armv7l-linux
         - i686-linux
         - m68k-linux
         - microblaze-linux
         - microblazeel-linux
         - mips64el-linux
         - mipsel-linux
         - powerpc64-linux
         - powerpc64le-linux
         - riscv32-linux
         - riscv64-linux
         - s390-linux
         - s390x-linux
         - x86_64-linux
(use '--show-trace' to show detailed location information)
Uh oh, the installation failed! Please create an issue at

    https://github.com/nix-community/home-manager/issues

if the error seems to be the fault of Home Manager.
```

### System information

```markdown
- system: `"aarch64-darwin"`
 - host os: `Darwin 21.6.0, macOS 12.6.3`
 - multi-user?: `yes`
 - sandbox: `no`
 - version: `nix-env (Nix) 2.13.3`
 - channels(jorge): `"darwin, home-manager, nixpkgs"`
 - nixpkgs: `/Users/jorge/.nix-defexpr/channels/nixpkgs`
```
```

berbiche:
Hi,

It seems that your existing Home Manager configuration uses services that are only configured for Linux at the moment.

You can set `services.gpg-agent.enable = pkgs.hostPlatform.isLinux` in the meantime.

This is likely going to be an issue for other "services" that don't/won't support Darwin/Mac in your configuration.

shackra:
> You can set `services.gpg-agent.enable = pkgs.hostPlatform.isLinux` in the meantime.
> 
> This is likely going to be an issue for other "services" that don't/won't support Darwin/Mac in your configuration.

yes, I just did that. Sharing configuration between GNU/Linux and macOS is not that hard tbh

dogayuksel:
[...]

PR #5901, opened 2024-09-27
Fix gpg-agent sockets on darwin

aymanbagabas:
### Description

This fixes a couple of issues with https://github.com/nix-community/home-manager/pull/5786 that uses launchd on darwin to start the gpg-agent daemon and its sockets. In particular, the issues outlined by @cmacrae.

1. Always export `$SSH_AUTH_SOCK` when using gpg-agent as ssh agent.
2. Use canonical socket paths on darwin
3. Replace deprecated `--supervised` option with `--daemon` using `gpgconf --launch` on darwin

Fixes: https://github.com/nix-community/home-manager/issues/3864

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/services/gpg-agent.nix
+++ b/modules/services/gpg-agent.nix
@@ -358,7 +358,8 @@ in {
           enable = true;
           config = {
             # macOS doesn't like the "--supervised" option
-            ProgramArguments = [ "${gpgPkg}/bin/gpgconf" "--launch" "gpg-agent" ]
+            ProgramArguments =
+              [ "${gpgPkg}/bin/gpgconf" "--launch" "gpg-agent" ]
               ++ optionals cfg.verbose [ "--verbose" ];
             EnvironmentVariables = { GNUPGHOME = homedir; };
             KeepAlive = {
--- a/modules/services/gpg-agent.nix
+++ b/modules/services/gpg-agent.nix
@@ -43,7 +43,9 @@ let
       else
         "d.${hash}/${dir}";
     in if pkgs.stdenv.isDarwin then
-      "/private/var/run/org.nix-community.home.gpg-agent/${subdir}"
+    # macOS GnuPG sockets always use the canonical path.
+    # See #3864
+      "${homedir}/${subdir}"
     else
       "%t/gnupg/${subdir}";
 
@@ -355,8 +357,9 @@ in {
         launchd.agents.gpg-agent = {
           enable = true;
           config = {
-            ProgramArguments = [ "${gpgPkg}/bin/gpg-agent" "--supervised" ]
-              ++ optional cfg.verbose "--verbose";
+            # macOS doesn't like the "--supervised" option
+            ProgramArguments = [ "${gpgPkg}/bin/gpgconf" "--launch" "gpg-agent" ]
+              ++ optionals cfg.verbose [ "--verbose" ];
             EnvironmentVariables = { GNUPGHOME = homedir; };
             KeepAlive = {
               Crashed = true;
--- a/modules/services/gpg-agent.nix
+++ b/modules/services/gpg-agent.nix
@@ -280,8 +280,11 @@ in {
           "pinentry-program ${lib.getExe cfg.pinentryPackage}"
           ++ [ cfg.extraConfig ]);
 
+      # Make sure we export GnuPG agent socket for SSH
+      # https://www.gnupg.org/documentation/manuals/gnupg/Agent-Examples.html#Agent-Examples
       home.sessionVariablesExtra = optionalString cfg.enableSshSupport ''
-        if [[ -z "$SSH_AUTH_SOCK" ]]; then
+        unset SSH_AGENT_PID
+        if [ "''${gnupg_SSH_AUTH_SOCK_by:-0}" -ne $$ ]; then
           export SSH_AUTH_SOCK="$(${gpgPkg}/bin/gpgconf --list-dirs agent-ssh-socket)"
         fi
       '';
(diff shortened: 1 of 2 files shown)


modules/services/gpg-agent.nix
```diff
@@ -280,8 +282,11 @@ in {
...
 
+      # Make sure we export GnuPG agent socket for SSH
+      # https://www.gnupg.org/documentation/manuals/gnupg/Agent-Examples.html#Agent-Examples
       home.sessionVariablesExtra = optionalString cfg.enableSshSupport ''
-        if [[ -z "$SSH_AUTH_SOCK" ]]; then
+        unset SSH_AGENT_PID
+        if [ "''${gnupg_SSH_AUTH_SOCK_by:-0}" -ne $$ ]; then
           export SSH_AUTH_SOCK="$(${gpgPkg}/bin/gpgconf --list-dirs agent-ssh-socket)"
         fi
       '';
```
cmacrae, line 292: not sure i understand this `gnupg_SSH_AUTH_SOCK_by` part -- how is this used?

modules/services/gpg-agent.nix
```diff
@@ -352,8 +357,10 @@ in {
         launchd.agents.gpg-agent = {
           enable = true;
           config = {
-            ProgramArguments = [ "${gpgPkg}/bin/gpg-agent" "--supervised" ]
-              ++ optional cfg.verbose "--verbose";
+            # macOS doesn't like the "--supervised" option
+            ProgramArguments =
+              [ "${gpgPkg}/bin/gpgconf" "--launch" "gpg-agent" ]
+              ++ optionals cfg.verbose [ "--verbose" ];
```
cmacrae, line 363: nice, thanks for updating this - i'm not sure if we'll need to look at changing how launchd manages/expects the lifecycle of the process to behave, given that `gpgconf --launch` basically schedules a separate process, then exits. so, perhaps `ProcessType = "Background"` isn't applicable here

cmacrae (commented):
hey @aymanbagabas 👋 thanks for your continued efforts! and thank you for including me as a co-author on some of the changes, that was kind of you :)

these changes mostly look good, particularly around the use of launchd sockets. just so i understand, is the intention that launchd manage the sockets and gpg-agent just uses them?

left some other comments for a couple of the other changes

folliehiyuki:
Upstream gnupg has deprecated both `--supervised` flag and having gpg-agent run under a service file (https://dev.gnupg.org/T6336#166815).

I suggest we remove both systemd unit and launchd service here, and attempt to start gpg-agent in user shell's init script.

Additional context: https://github.com/NixOS/nixpkgs/issues/213794

modules/services/gpg-agent.nix
folliehiyuki: Seems like the snippet is taken from https://wiki.archlinux.org/title/Gpg#Set_SSH_AUTH_SOCK

jkachmar:
so while @folliehiyuki is correct & upstream has deprecated init-based supervision[^1], it looks like both [the nixpkgs `gnupg` module](https://github.com/NixOS/nixpkgs/blob/nixos-24.11/nixos/modules/programs/gnupg.nix) & other distros [such as Arch](https://wiki.archlinux.org/title/GnuPG#gpg-agent) are going to step in and maintain their own service & socket definitions.

fwiw, I extended my config with changes roughly along the lines of this module & everything appears to work as expected; if upstream nixpkgs & other distributions are maintaining their own supervised `gpg-agent` services, it doesn't seem unreasonable for `home-manager` to provide one as well.

[^1]: some additional context on the commit: https://dev.gnupg.org/rGeae28f1bd4a5632e8f8e85b7248d1c4d4a10a5ed

r4v3n6101:
Any news here guys? Come from here: https://github.com/nix-community/home-manager/issues/5997

aymanbagabas:
> EDIT: spoke too soon, looks like something about the `launchd` socket setup doesn't work correctly & it locks up...

Reached the same conclusion :/ Need to investigate this move.

modules/services/gpg-agent.nix
Yunir: looks like it's from official docs also: https://www.gnupg.org/documentation/manuals/gnupg/Agent-Examples.html.
Either way leaving just `SSH_AUTH_SOCK` on if-clause won't work with macOS because macOS provides it's own native LaunchAgent ssh-agent `/System/Library/LaunchAgents/com.openssh.ssh-agent.plist` that prefills this `SSH_AUTH_SOCK` env variable with something like `/private/tmp/com.apple.launchd.wS4zQCeDFk/Listeners`. So it's never empty, only if user won't do `unset` or equals to empty string before.

upd. Actually looks like this part was partially upstreamed and enhanced on master already.

modules/services/gpg-agent.nix
Yunir: For me `--supervised` also does not work. Here is the logs:
```
~ $ sudo launchctl debug gui/$(id -u)/org.nix-community.home.gpg-agent --stdout --stderr
Service configured for next launch.
gpg-agent (GnuPG) 2.4.8 starting in supervised mode.
no LISTEN_PID environment variable found in --supervised mode (ignoring)
no LISTEN_FDS or LISTEN_FDNAMES environment variables found in --supervised mode (assuming 1 active descriptor)
Fatal: file descriptor 3 must be valid in --supervised mode if LISTEN_FDNAMES is not set
```

upd. noticed that `--supervised` is even deprecated.

Yunir: just for the reference, nix-darwin's launchctl service worked: https://github.com/nix-darwin/nix-darwin/blob/d76299b2cd01837c4c271a7b5186e3d5d8ebd126/modules/programs/gnupg.nix#L34.

Issue #5889, opened 2024-09-24
bug: Megasync service core dump.


knokelmaat:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Megasync is enabled in Home Manager using `services.megasync.enable`, but does not launch correctly on login.

Relevant logs:
```
Sep 24 09:39:40 nixos-B450M systemd[2091]: Started megasync.
Sep 24 09:39:40 nixos-B450M megasync[2177]: Avoiding wayland
Sep 24 09:39:40 nixos-B450M megasync[2177]: Error for "xrdb -query" command: "xrdb: Can't open display ''\n"
Sep 24 09:39:40 nixos-B450M megasync[2177]: qt.qpa.xcb: could not connect to display
Sep 24 09:39:40 nixos-B450M megasync[2177]: qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
Sep 24 09:39:40 nixos-B450M megasync[2177]: This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.
Sep 24 09:39:40 nixos-B450M megasync[2177]: Available platform plugins are: wayland-egl, wayland, wayland-xcomposite-egl, wayland-xcomposite-glx, eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, xcb.
Sep 24 09:39:40 nixos-B450M systemd-coredump[2221]: Process 2177 (.megasync-wrapp) of user 1000 terminated abnormally with signal 6/ABRT, processing...
Sep 24 09:39:40 nixos-B450M systemd-coredump[2229]: Process 2177 (.megasync-wrapp) of user 1000 dumped core.
                                                    
                                                    Module libxcb-util.so.1 without build-id.
                                                    Module libexpat.so.1 without build-id.
                                                    Module libxkbcommon.so.0 without build-id.
                                                    Module libxkbcommon-x11.so.0 without build-id.
                                                    Module libICE.so.6 without build-id.
                                                    Module libSM.so.6 without build-id.
                                                    Module libxcb-xinput.so.0 without build-id.
                                                    Module libxcb-xkb.so.1 without build-id.
                                                    Module libxcb-xinerama.so.0 without build-id.
                                                    Module libxcb-xfixes.so.0 without build-id.
                                                    Module libxcb-sync.so.1 without build-id.
[...]

PR #5902, opened 2024-09-27
megasync: fix issue with service failing to launch

knokelmaat:
### Description
This fixes the problem described in #5889, where enabling the `megasync` service resulted in a failed systemd unit.

It was seemingly solved by delaying the launching of the service to after `graphical-session.target` instead of `graphical-session-pre.target`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@GaetanLepage

### DIFF
--- a/modules/services/megasync.nix
+++ b/modules/services/megasync.nix
@@ -28,7 +28,7 @@ in {
     systemd.user.services.megasync = {
       Unit = {
         Description = "megasync";
-        After = [ "graphical-session-pre.target" ];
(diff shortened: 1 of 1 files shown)


GaetanLepage (approved):
Thanks for fixing this ! I am not using this module anymore.
LGTM.

cc @rycee

knokelmaat:
Also fixes #5910.

rycee:

PR #5905, opened 2024-09-28
Nushell generator

JoaquinTrinanes:
### Description

This PR supersedes #5795.

It adds a nushell generator that works with basic primitive types, records and lists. It also adds a `nushell-inline` to allow generating nushell code instead of strings.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@Philipp-M @rycee

### DIFF
--- a/modules/lib/default.nix
+++ b/modules/lib/default.nix
@@ -14,4 +14,5 @@ rec {
 
   shell = import ./shell.nix { inherit lib; };
   zsh = import ./zsh.nix { inherit lib; };
+  nushell = import ./nushell.nix { inherit lib; };
 }
--- a/modules/lib/nushell.nix
+++ b/modules/lib/nushell.nix
@@ -1,0 +1,65 @@
+{ lib }: rec {
+  mkNushellInline = expr: lib.setType "nushell-inline" { inherit expr; };
+
+  toNushell = { indent ? "", multiline ? true, asBindings ? false }@args:
+    v:
+    let
+      innerIndent = "${indent}    ";
+      introSpace = if multiline then ''
+
+        ${innerIndent}'' else
+        " ";
+      outroSpace = if multiline then ''
+
+        ${indent}'' else
+        " ";
+      innerArgs = args // {
+        indent = if asBindings then indent else innerIndent;
+        asBindings = false;
+      };
+      concatItems = lib.concatStringsSep introSpace;
+      isNushellInline = lib.isType "nushell-inline";
+
+      generatedBindings = assert lib.assertMsg (badVarNames == [ ])
+        "Bad Nushell variable names: ${
+          lib.generators.toPretty { } badVarNames
+        }";
+        lib.concatStrings (lib.mapAttrsToList (key: value: ''
+          ${indent}let ${key} = ${toNushell innerArgs value}
+        '') v);
+
+      isBadVarName = name:
+        # Extracted from https://github.com/nushell/nushell/blob/ebc7b80c23f777f70c5053cca428226b3fe00d30/crates/nu-parser/src/parser.rs#L33
+        # Variables with numeric or even empty names are allowed. The only requisite is not containing any of the following characters
+        let invalidVariableCharacters = ".[({+-*^/=!<>&|";
+        in lib.match "^[$]?[^${lib.escapeRegex invalidVariableCharacters}]+$"
+        name == null;
+      badVarNames = lib.filter isBadVarName (builtins.attrNames v);
+    in if asBindings then
+      generatedBindings
+    else if v == null then
+      "null"
+    else if lib.isInt v || lib.isFloat v || lib.isString v || lib.isBool v then
+      lib.strings.toJSON v
+    else if lib.isList v then
+      (if v == [ ] then
+        "[]"
+      else
+        "[${introSpace}${
+          concatItems (map (value: "${toNushell innerArgs value}") v)
+        }${outroSpace}]")
+    else if lib.isAttrs v then
+      (if isNushellInline v then
+        "(${v.expr})"
+      else if v == { } then
+        "{}"
+      else if lib.isDerivation v then
+        toString v
+      else
+        "{${introSpace}${
+          concatItems (lib.mapAttrsToList (key: value:
+            "${lib.strings.toJSON key}: ${toNushell innerArgs value}") v)
+        }${outroSpace}}")
+    else
+      abort "nushell.toNushell: type ${lib.typeOf v} is unsupported";
+}
--- a/modules/programs/nushell.nix
+++ b/modules/programs/nushell.nix
@@ -145,11 +145,24 @@ in {
     };
 
     environmentVariables = mkOption {
-      type = types.attrsOf types.str;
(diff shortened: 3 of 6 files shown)


modules/programs/nushell.nix
```diff
@@ -145,11 +145,24 @@ in {
...
-      type = types.attrsOf types.str;
+      type = types.attrsOf hm.types.nushellValue;
       default = { };
-      example = { FOO = "BAR"; };
+      example = literalExpression ''
+        {
+          FOO = "BAR";
+          LIST_VALUE = [ "foo" "bar" ];
+          NU_LIB_DIRS = lib.concatStringsSep ":" [ ./scripts ];
+          PROMPT_COMMAND = lib.hm.nushell.mkNushellInline "{|| \"> \"}";
```
rycee, line 155: It is reasonably idiomatic to use `''` strings if they help avoid escaping:

```suggestion
          PROMPT_COMMAND = lib.hm.nushell.mkNushellInline ''{|| "> "}'';
```

modules/programs/nushell.nix
```diff
@@ -145,11 +145,24 @@ in {
...
+            from_string = lib.hm.nushell.mkNushellInline "{|s| $s | split row (char esep) }";
+            to_string = lib.hm.nushell.mkNushellInline "{|v| $v | str join (char esep) }";
+          };
+        }
+      '';
       description = ''
-        An attribute set that maps an environment variable to a shell interpreted string.
+        Environment variables to be set.
+
+        Inline values can be set with `lib.hm.gvariant.mkUint32`.
```
rycee, line 165: Are you sure about this?

modules/lib/nushell.nix
```diff
@@ -0,0 +1,63 @@
...
+        " ";
+      innerArgs = args // {
+        indent = if asBindings then indent else innerIndent;
+        asBindings = false;
+      };
+      concatItems = lib.concatStringsSep introSpace;
+      isNushellInline = lib.isType "nushell-inline";
+
+      generatedBindings = assert lib.assertMsg (badVarNames == [ ])
+        "Bad Nushell var names: ${lib.generators.toPretty { } badVarNames}";
```
rycee, line 24:
```suggestion
        "Bad Nushell variable names: ${lib.generators.toPretty { } badVarNames}";
```

rycee (commented):
Thanks! Just added a few minor comments.

modules/lib/default.nix
```diff
@@ -9,6 +9,7 @@ rec {
   generators = import ./generators.nix { inherit lib; };
   gvariant = import ./gvariant.nix { inherit lib; };
   maintainers = import ./maintainers.nix;
+  nushell = import ./nushell.nix { inherit lib; };
```
Philipp-M, line 12: Nit: maybe put it below to the other shells (i.e. rather semantic ordering than lexicographic, as this is not really ordered lexicographically anyways)

modules/lib/nushell.nix
```diff
@@ -0,0 +1,63 @@
+{ lib }: rec {
+  mkNushellInline = expr: lib.setType "nushell-inline" { inherit expr; };
+
+  toNushell = { indent ? "", multiline ? true, asBindings ? false }@args:
```
Philipp-M, line 4: I guess `multiline`, `indent` (and vars `outroSpace`, `introSpace`) is purely for aesthetic reasons?

`asBindings` isn't used anywhere as well (or rather is false everywhere), for what is this used, is something planned with this?

(No need for action, just questions)

modules/lib/nushell.nix
```diff
@@ -0,0 +1,63 @@
...
+        "[${introSpace}${
+          concatItems (map (value: "${toNushell innerArgs value}") v)
+        }${outroSpace}]")
+    else if lib.isAttrs v then
+      (if isNushellInline v then
+        "(${v.expr})"
+      else if v == { } then
+        "{}"
+      else if lib.isDerivation v then
+        toString v
```
Philipp-M, line 55: Does this make sense anyhow, is this useful anywhere? Should this be just an error as below?

modules/programs/nushell.nix
```diff
@@ -39,7 +39,7 @@ let
       };
     });
 in {
-  meta.maintainers = [ maintainers.Philipp-M ];
+  meta.maintainers = [ maintainers.Philipp-M maintainers.joaquintrinanes ];
```
Philipp-M, line 42: I think you also need to add yourself in the `maintainers.nix` file.

(Btw. I think you can name yourself as first maintainer, I'm not really effectively maintaining nushell anyways, you seem to be more into nushells guts)

modules/programs/nushell.nix
Philipp-M, line 165: Is this accurate? Shouldn't this be:
```suggestion
        Inline values can be set with `lib.hm.nushell.mkNushellInline`.
```

Philipp-M (approved):
Apart from questions and nits, looks good to me. Nice improvement, thanks

modules/programs/nushell.nix
rycee: It shouldn't be necessary to add yourself to the HM maintainers file if you are already in the Nixpkgs' one.

modules/programs/nushell.nix
JoaquinTrinanes: Had to use `'''` for escaping, but agree

modules/lib/default.nix
JoaquinTrinanes: Done! I like it better now

modules/lib/nushell.nix
JoaquinTrinanes: This is straight up yoinked from [the `nixpkgs` lua generator](https://github.com/NixOS/nixpkgs/blob/8637e879e425363b14aa23f3ec1f69e32c9f512f/lib/generators.nix#L704). I thought about removing some unused features it but they work so why not keep them.

`multiline` et al. are just aesthetic, but I like that it's the default to avoid having incredibly long lines in the config files.

`asBindings` could be used by consumers of the lib but it is unused by the module as well.

modules/lib/nushell.nix
JoaquinTrinanes: I think it makes sense to coerce a derivation to a string, it would allow thing like this:

```nix
programs.nushell.environmentVariables = {
    RIPGREP_CONFIG_PATH = pkgs.writeText "ripgreprc" ''
      --smart-case
      --auto-hybrid-regex
  '';
};
```

It's also what the official lua generator does

abhisheksingh0x558:
Both [zsh](https://github.com/nix-community/home-manager/blob/2a4fd1cfd8ed5648583dadef86966a8231024221/modules/programs/zsh.nix#L584) and [bash](https://github.com/nix-community/home-manager/blob/2a4fd1cfd8ed5648583dadef86966a8231024221/modules/programs/bash.nix#L215) source home.sessionVariables apart from sourcing the shell specific environment. Would it serve as a sensible default for nushell as well?

abhisheksingh0x558 (approved):

rycee:

Issue #3317, opened 2022-10-10
mbsync will be getting XDG support in 1.5


peterhoeg:
### Description

hm currently drops `.mbsyncrc` into `$HOME` but with the upcoming 1.5, it will gain proper XDG support:

https://sourceforge.net/p/isync/feature-requests/14/

rycee:
Cool 🙂 PR very welcome after the new version is available in Nixpkgs unstable.

peterhoeg:
1.5 hasn't hit nixpkgs yet.

sumnerevans:
Is there a Nixpkgs PR to track?

peterhoeg:
No, because the reason it isn’t in nixpkgs is because upstream hasn’t released it yet… ;-)

peterhoeg:
1.5 has not been released yet.

nekowinston:
No new release yet.

felipeqq2:
1.5 was [released](https://sourceforge.net/projects/isync/files/isync/1.5.0/)! :tada: 

Related: https://github.com/NixOS/nixpkgs/pull/331813

teto:
done

PR #5918, opened 2024-09-30
isync/mbsync: update module for 1.5.0 changes

### LINKED ISSUE #5884: mbsync: update module for v1.5.0
`isync` had a new release v1.5.0 in August, and the update has been merged into nixpkgs master a few days ago (see https://github.com/NixOS/nixpkgs/pull/331813). The update introduces several breaking changes, the details of which can be seen on https://sourceforge.net/projects/isync/files/isync/1.5.0/. See #3317 as well for an issue specific to one of those changes.

Maintainer ping: @KarlJoad

KarlJoad:
### Description
Update isync to use new features in isync 1.5.0.

Closes #5884 and #3317.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/tests/modules/programs/mbsync/mbsync-expected.conf
+++ b/tests/modules/programs/mbsync/mbsync-expected.conf
@@ -57,7 +57,7 @@ CertificateFile /etc/ssl/certs/ca-certificates.crt
 Host imap.example.com
 PassCmd password-command
 TLSType IMAPS
-SSLVersions TLSv1.3 TLSv1.2
+TLSVersions +1.3 +1.2 -1.1
 User home.manager
 
 IMAPStore hm@example.com-remote
--- a/tests/modules/programs/mbsync/mbsync-expected.conf
+++ b/tests/modules/programs/mbsync/mbsync-expected.conf
@@ -4,7 +4,7 @@ IMAPAccount hm-account
 CertificateFile /etc/ssl/certs/ca-certificates.crt
 Host imap.example.org
 PassCmd "password-command 2"
-SSLType IMAPS
+TLSType IMAPS
 User home.manager.jr
 
 IMAPStore hm-account-remote
@@ -56,7 +56,7 @@ IMAPAccount hm@example.com
 CertificateFile /etc/ssl/certs/ca-certificates.crt
 Host imap.example.com
 PassCmd password-command
-SSLType IMAPS
+TLSType IMAPS
 SSLVersions TLSv1.3 TLSv1.2
 User home.manager
 
--- a/tests/modules/programs/mbsync/mbsync.nix
+++ b/tests/modules/programs/mbsync/mbsync.nix
@@ -79,8 +79,8 @@ with lib;
     test.stubs.isync = { };
 
     nmt.script = ''
-      assertFileExists home-files/.mbsyncrc
-      assertFileContent home-files/.mbsyncrc ${./mbsync-expected.conf}
+      assertFileExists home-files/.config/isyncrc
+      assertFileContent home-files/.config/isyncrc ${./mbsync-expected.conf}
     '';
   };
 }
--- a/tests/modules/programs/mbsync/mbsync.nix
+++ b/tests/modules/programs/mbsync/mbsync.nix
@@ -21,7 +21,7 @@ with lib;
     accounts.email.accounts = {
       "hm@example.com".mbsync = {
         enable = true;
-        extraConfig.account.SSLVersions = [ "TLSv1.3" "TLSv1.2" ];
+        extraConfig.account.TLSVersions = [ "+1.3" "+1.2" "-1.1" ];
         groups.inboxes = {
           channels = {
             inbox1 = {
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -1861,6 +1861,35 @@ in {
           Some plugins require this to be set to 'false' to function correctly.
         '';
       }
+
+      {
+        time = "2024-12-08T17:22:13+00:00";
+        condition = let
+          usingMbsync = any (a: a.mbsync.enable)
+            (attrValues config.accounts.email.accounts);
+        in usingMbsync;
+        message = ''
+          isync/mbsync 1.5.0 has changed several things.
+
+          isync gained support for using $XDG_CONFIG_HOME, and now places
+          its config file in '$XDG_CONFIG_HOME/isyncrc'.
+
+          isync changed the configuration options SSLType and SSLVersion to
+          TLSType and TLSVersion respectively.
+
+          All instances of
+          'accounts.email.accounts.<account-name>.mbsync.extraConfig.account'
+          that use 'SSLType' or 'SSLVersion' should be replaced with 'TLSType'
+          or 'TLSVersion', respectively.
+
+          TLSType options are unchanged.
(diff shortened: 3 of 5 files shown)


tests/modules/programs/mbsync/mbsync-expected.conf
```diff
@@ -56,8 +56,8 @@ IMAPAccount hm@example.com
 CertificateFile /etc/ssl/certs/ca-certificates.crt
 Host imap.example.com
 PassCmd password-command
-SSLType IMAPS
-SSLVersions TLSv1.3 TLSv1.2
+TLSType IMAPS
+TLSVersions TLSv1.3 TLSv1.2
```
toothbrush, line 60:
```suggestion
TLSVersions +1.3 +1.2
```

Sorry about the drive-by comment, but i just got bitten by this change in isync/mbsync. The `SSLVersions` and `TLSVersions` config keys don't take the same arguments, it's not a simple rename.

I have found references here https://sourceforge.net/projects/isync/files/isync/1.5.0/ and here https://isync.sourceforge.io/mbsync.html but frustratingly i haven't found actual documentation for permitted arguments.  By trial and error i managed to make things work with the `+` as shown above.

Disclosure/disclaimer is that i'm using isync/mbsync on macOS via Homebrew – i am just making a comment here before this change wastes other people's time more than necessary. I found this PR because i was looking for working examples.

KarlJoad: That's a good catch! Thanks! I will have to mark `SSLVersions` as completely deprecated and inform the user about `TLSVersions` along with the `+`/`-` change, I guess.

>  ...but frustratingly i haven't found actual documentation for permitted arguments.

I found the source file that implements this parsing (`src/drv_imap.c:3788`). They also support a `-` symbol to disable a particular version. I suspect this was an undocumented change to support isync 1.5.0's default-disabling of TLS 1.0/1.1.

I will document this in a release note. But because `SSLVersions`/`TLSVersions` is part of `extraConfig`, I doubt I will be able to automatically catch this misconfiguration.

dbaynard: > isync 1.5.0's default-disabling of TLS 1.0/1.1

This is the only reason I used `SSLVersions` so I could delete the line. (I probably won't, though, in case there are future changes.)

rycee: Is this something that needs to be resolved before merging?

tests/modules/programs/mbsync/mbsync.nix
```diff
@@ -21,7 +21,7 @@ with lib;
     accounts.email.accounts = {
       "hm@example.com".mbsync = {
         enable = true;
-        extraConfig.account.SSLVersions = [ "TLSv1.3" "TLSv1.2" ];
+        extraConfig.account.TLSVersions = [ "TLSv1.3" "TLSv1.2" ];
```
dbaynard, line 24:
```suggestion
        extraConfig.account.TLSVersions = [ "TLSv1.3" "TLSv1.2" ];
        extraConfig.account.TLSVersions = [ "+1.3" "+1.2" ];
```

Needed in addition to https://github.com/nix-community/home-manager/pull/5918#discussion_r1816360564

tests/modules/programs/mbsync/mbsync-expected.conf
rycee: Thanks for the clarification @dbaynard.

NovaViper:
Hey can you update the PR to match master? Several stuff has changed in master since this PR was made and makes it impossible to test it in my flake

NovaViper:
@KarlJoad Hey can you update the PR to match master? Several stuff has changed in master since this PR was made and makes it impossible to test it in my flake

KarlJoad:
@NovaViper, the rebase should be done. I just jumped to the tip of main. Let me know if things are still not working for you.

KarlJoad:
This should be an MVP for getting isync 1.5.0 changes pushed out the door. A news entry has been added so that mbsync users are informed about this change, since I don't see a way to have Nix automatically detect and warn the user about this change (since `TLSType` and `TLSVersions` are placed in an `extraConfig` slot).

NovaViper:
Hey @KarlJoad , thank you for the rebase! I just gave it a test and the `TLSVersions` and TLSType` changes work. And mbsync is properly reading from the new `~/.config/isyncrc` path. Looks good to merge!

KarlJoad:
Thanks for checking that stuff still works @NovaViper. I have switched off home-manager since originally writing this module and have no good way to check it anymore.

NovaViper:
No problem, glad I could help in testing! I saw this a bit ago but hadn't had the time to test it then 😞

octvs:
Just tested. Diff btw the generated isync config and the one created from the current
master is as follows

```diff
10c10
< SSLType IMAPS
---
> TLSType IMAPS
37c37
< SSLType IMAPS
---
> TLSType IMAPS
64c64
< SSLType IMAPS
---
> TLSType IMAPS
```

It also fetches my mails as expected, from gmail and two other mail
providers.

octvs (approved):

KarlJoad:
@octvs, your diff should also have included `TLSVersions`. Did your config change that too? Or is the diff not complete?

octvs:
> @octvs, your diff should also have included `TLSVersions`. Did your config
> change that too? Or is the diff not complete?

The diff is complete, I never had `TLSVersions` on my previous `~/.mbsyncrc`,
since back when I wrote the config by hand. Therefore I did not expect to see
it on the new one. Maybe someone else who uses that variable can help out.

dbaynard:
The tests touched by this PR include `TLSVersions` but the module itself does not generate such a field unless a user enters it explicitly. Users who have `SSLVersions` in their `extraConfig` will see the deprecation warning from isync.

Well, the warning will be displayed. `isync` has been very much "set and forget" for me, so it was only by chance that I saw it. It would be great to have the warning available during a nixpkgs/home-manager upgrade. It probably belongs on the upstream package, though, rather than this. (I wish I'd thought of this earlier.)

KarlJoad:
@dbaynard I agree and wanted to do that with `mkRenamedOptionModule`, but couldn't (see https://github.com/nix-community/home-manager/pull/5918#issuecomment-2526277506). That's why I left it in `news.nix`. Then at least everyone using mbsync will see it, and if they do not use `SSLVersions` at all, then they do not have to do anything.

mbsync already warns you about the `SSLType -> TLSType` warning, so I am not sure it will do it for `SSLVersions -> TLSVersions` too.

When I was testing yesterday, I wanted to warn users with `mkRenamedOptionModule` but couldn't, because the option module where users provide this information is marked as an `extraConfig` typed slot, which allows the user to pass almost anything.

This module will need to go through a bit of a refactor if we want to match against some key variables that need to be set (like `TLSType` and `TLSVersion`).

alarsyo:
Could this get back-ported to `release-24.11`? isync 1.5.0 is in `nixos-24.11` as well :)

KarlJoad:
Sure thing @alarsyo. I have never done a backport before though, so I do not know the procedure.

alarsyo:
Me neither. I tried something in #6194 :)

Issue #5102, opened 2024-03-07
bug: Linux Mint 21.3 w/Home-Manager Crashes to Fallback & systemd-coredump


davidjmeier:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

New to Nix.  Installed Nix via Determinate systems, made no changes.  Went directly to installing Home-Manager stand-alone via the Flakes instructions.  Upon logout / login or reboot immediately after enabling the default Home-Manager configuration (no changes) Cinnamon now crashes and goes to fallback mode.

If I force a Cinnamon refresh (Ctrl-Alt-ESC) I get this line immediately in `/var/log/dmesg`: `syslog:Mar  7 11:00:16 myhost systemd-coredump[8997]: Process 8941 (cinnamon) of user 1000 dumped core.`

TIA!

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.5.0-21-generic, Linux Mint, 21.3 (Virginia), nobuild`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.20.3`
 - nixpkgs: `/home/username/.nix-defexpr/channels/nixpkgs`
```

rycee:
Sorry to hear about the crashing. I'm not exactly certain what Cinnamon is but the default HM configuration doesn't do very much at all. It is more or less limited to installing the `man` tool and a few basic symlinks like `~/.cache/.keep`. Does it work OK if you don't use Home Manager? For example, just installing some package using `nix profile install`?

Also, any other hint in the systemd journal or something? Just knowing that it crashes unfortunately isn't very actionable.

davidjmeier:
Thanks for the response @rycee!

Yeah, I really wasn't sure exactly where to start.  Just installing Nix and tossing a few packages into the system works fine.  I removed everything and basically made one change at a time and then rebooted until the desktop environment (Cinnamon) started crashing again.  It's immediately after HM is installed with the default configuration - which, agreed, doesn't have much to it.  I was wondering if there was chance HM was taking over something specific that Cinnamon no longer had access to, was removed or moved.

Is there anything specific I can look for in logs that could help pinpoint this?

Thanks much!

DanSM-5:
[...]

Issue #4941, opened 2024-01-23
bug: 


andersrh:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

When I install home-manager Dolphin freezes, when I go to my home folder.
When I uninstall home-manager the issue gets resolved.

I haven't been able to find others, who have exactly the same issue.

My nixpkgs is pointed to unstable and I have added the master branch for home-manager.

### System information

```markdown
this path will be fetched (0.00 MiB download, 0.00 MiB unpacked):
  /nix/store/v8p3m2vgkcd3ykrdylm5vi9alrxbwhip-nix-info
copying path '/nix/store/v8p3m2vgkcd3ykrdylm5vi9alrxbwhip-nix-info' from 'https://cache.nixos.org'...
 - system: `"x86_64-linux"`
 - host os: `Linux 6.7.0-cb1.0.fc39.lto.x86_64, Fedora Linux, 39.20240120.0 (Kinoite), nobuild`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.19.2`
 - channels(root): `"nixpkgs"`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixpkgs`
```

n-a-m-e:
Happens to me too. Running home-manager on Ubuntu.

I don't know why but any directory containing .js files crashes dolphin. If I open a Folder without any .js files it works just fine.
There may be other file extensions which also cause a crash but I haven't found any.

I also can't open the KDE editor "Kate" as it also instantly crashes.

I think it is probably something to do with KDE apps

I was previously able to fix it by changing changing from master and setting nixpkgs.url to "github:nixos/nixpkgs/nixos-23.11" and home-manager url to "github:nix-community/home-manager/release-23.11"

But after the last time I updated a couple of days ago the issue is back.

yeus:
same issue here!  Not sure whats causing it, but resetting dolphin to default settings helped:

`rm -r .local/share/dolphin/ .config/session/dolphin_* .config/dolphinrc`

yeus:
some more investigation shows that kate always freezes when I call the open/close dialog...   trying to reset that as well right now :)

yeus:
hm.  now I have a similar problem to @n-a-m-e :

- dolphin freezes whenever I enter a directory with a json file.
[...]

Issue #4682, opened 2023-11-20
bug: `targets.genericLinux` breaks Plasma on Arch Linux


metiulekm:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

After a recent update of Arch and my HM config, I found that (X11) Plasma would not start. Some Plasma systemd services would start, but the initial splash screen would never end, except fading into black in some cases. This happened on all my computers. After some debugging, I found that removing `targets.genericLinux.enable = true;` from my config, switching the config and stopping any remaining background programs before attempting to login in again would fix the issue.

I understand that `targets.genericLinux` provides config that is appropriate for non-NixOS installations. It's not crucial to me, so for now I have just commented it out in my config.

I believe that the issue happened after Arch update and not HM/nixpkgs update, since I have also reproduced the issue with the ~100 days old versions of HM and nixpkgs.

Here's a minimal `flake.nix` that I have been able to reproduce the issue on a fresh user account with:
```nix
{
  description = "Home Manager configuration";

  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
    home-manager = {
      url = "github:nix-community/home-manager";
      inputs.nixpkgs.follows = "nixpkgs";
    };
  };

  outputs = inputs: let
    system = "x86_64-linux";
    pkgs = import inputs.nixpkgs {
      inherit system;
      allowAliases = false;
    };
  in {
    homeConfigurations.metiulekm = inputs.home-manager.lib.homeManagerConfiguration {
      inherit pkgs;
      modules = [
        (
          {
            config,
            ...
          }: {
            home = {
              username = "test";
              homeDirectory = "/home/${config.home.username}";
              stateVersion = "22.05";
            };

            targets.genericLinux.enable = true;
          }
        )
      ];
    };

    devShell.${system} = pkgs.mkShell {
      buildInputs = [inputs.home-manager.packages.${system}.home-manager];
    };
  };
}
```

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.6.1-zen1-1-zen, Arch Linux, noversion, rolling`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.8.1`
 - channels(root): `"nixpkgs"`
 - channels(metiulekm): `""`
[...]

PR #5920, opened 2024-10-01
xdg-mime type package options

### LINKED ISSUE #4955: bug: xdg.mime.enabled = true causes crash when encountering a .json file

### Intro and Workaround
Not sure if I'm supposed to add information to issue 4682 or create a new one.
Sorry in case I chose the wrong approach. (sounds like a different problem, even though the same fix/workaround can be used)

_Originally posted by @hexclover in https://github.com/nix-community/home-manager/issues/4682#issuecomment-1841994950_
> For me setting `xdg.mime.enable = false;` in home-manager also works

### Problem description and steps to reproduce:
encountering[^1] a .json file crashes a lot of programs[^2] in at least Gnome and KDE
[^1]: (with 'encountering' I mean clicking it in file explorer apps or opening it in editor apps)
[^2]: Explicitly tested Nautilus, gedit, new gnome Text Editor (no idea what the name is), thunar and Dolphin.

I first encountered the problem on bare metal PopOS 22.04 
and then reproduced it with (quickemu) VMs (Kubuntu 23.10, Ubuntu 23.10, Fedora 39)

For the tests in the VMs I used the following code (but also tested once with a multiuser install)

```
echo 1 >> ~/Downloads/foo.json
sudo apt install -y curl
sh <(curl -L https://nixos.org/nix/install) --no-daemon --yes
. "/home/${USER}/.nix-profile/etc/profile.d/nix.sh"
nix-channel --add https://github.com/nix-community/home-manager/archive/master.tar.gz home-manager
nix-channel --update
nix-shell '<home-manager>' -A install
```
Then rebooted and tried to access/open the .json file

### Maintainer CC

_No response_

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.6.10-76060610-generic, Pop!_OS, 22.04 LTS, nobuild`
 - multi-user?: `no`
 - sandbox: `yes`
 - nixpkgs: `/home/foobar/.nix-defexpr/channels/nixpkgs`
```

notalltim:
### Description

This PR adds package options for `shared-mime-info` and `desktop-file-utils` for the purpose of allowing users to override the version in nixpkgs. This with overridden version of `shared-mime-info` should  resolve #4955, #5102, #4682, and possibly #4941. I also added a couple basic tests for the module. An example of using the override can be found at notalltim/home-manager-config#4. I did not see a maintainer let me know if I missed one.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/xdg-mime.nix
+++ b/modules/misc/xdg-mime.nix
@@ -5,33 +5,50 @@ with lib;
 let
 
   cfg = config.xdg.mime;
+  inherit (lib) getExe getExe';
 
 in {
   options = {
-    xdg.mime.enable = mkOption {
-      type = types.bool;
-      default = pkgs.stdenv.hostPlatform.isLinux;
-      defaultText =
-        literalExpression "true if host platform is Linux, false otherwise";
-      description = ''
-        Whether to install programs and files to support the
-        XDG Shared MIME-info specification and XDG MIME Applications
-        specification at
-        <https://specifications.freedesktop.org/shared-mime-info-spec/shared-mime-info-spec-latest.html>
-        and
-        <https://specifications.freedesktop.org/mime-apps-spec/mime-apps-spec-latest.html>,
-        respectively.
-      '';
+    xdg.mime = {
+      enable = mkOption {
+        type = types.bool;
+        default = pkgs.stdenv.hostPlatform.isLinux;
+        defaultText =
+          literalExpression "true if host platform is Linux, false otherwise";
+        description = ''
+          Whether to install programs and files to support the
+          XDG Shared MIME-info specification and XDG MIME Applications
+          specification at
+          <https://specifications.freedesktop.org/shared-mime-info-spec/shared-mime-info-spec-latest.html>
+          and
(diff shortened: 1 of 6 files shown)


teto:
though it's possible to achieve that via an overlay, I can see how this could be easier to use.

modules/misc/xdg-mime.nix
```diff
@@ -46,12 +63,12 @@ in {
       if [[ -w $out/share/mime && -w $out/share/mime/packages && -d $out/share/mime/packages ]]; then
         XDG_DATA_DIRS=$out/share \
         PKGSYSTEM_ENABLE_FSYNC=0 \
-        ${pkgs.buildPackages.shared-mime-info}/bin/update-mime-database \
+        ${getExe cfg.sharedMimeInfoPackage} \
```
Artturin, line 66: These are run during the build right?, so this breaks cross. Reported by @colemickens

Artturin: @notalltim 

> I also switched from using `buildPackages` to `pkgs` to improve
cross-compilation compatibility.

This does not improve cross-compilation compatibility, you need to use the package from buildPackages  or an emulator.

PR #5923, opened 2024-10-01
modules/services/dropbox.nix: fix dropbox service

tph5595:
### Description

The dropbox service does not work because it uses ```dropbox-cli``` and not ```dropbox```. This changes the service to use ```dropbox``` only. Additionally, the old maintainer is not active (https://github.com/NixOS/nixpkgs/issues/235345#issuecomment-1618603377) so I added myself as the maintainer.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@eyJhb


modules/services/dropbox.nix
```diff
@@ -33,48 +33,42 @@ in {
...
+    home.packages = [ pkgs.dropbox ];
 
     systemd.user.services.dropbox = {
       Unit = { Description = "dropbox"; };
 
       Install = { WantedBy = [ "default.target" ]; };
 
       Service = {
-        Environment = [ "HOME=${homeBaseDir}" "DISPLAY=" ];
+        Environment = [ "HOME=${homeBaseDir}" "LD_LIBRARY_PATH=${pkgs.libGL}/lib/"];
```
SuperSandro2000, line 44: This should be fixed in the package

tph5595: How do I do that? Sorry this is my first contribution.

tph5595: @ttuegel Can you help me with this?

tph5595: > This should be fixed in the package
@SuperSandro2000 

This has been fixed. Is this ready to be merged?

modules/services/dropbox.nix
```diff
@@ -6,11 +6,11 @@ let
 
   cfg = config.services.dropbox;
   baseDir = ".dropbox-hm";
-  dropboxCmd = "${pkgs.dropbox-cli}/bin/dropbox";
+  dropboxCmd = "${pkgs.dropbox}/bin/dropbox";
   homeBaseDir = "${config.home.homeDirectory}/${baseDir}";
 
 in {
-  meta.maintainers = [ maintainers.eyjhb ];
+  meta.maintainers = [ maintainers.tph5595 ];
```
rycee, line 13: Since you are not in the Nixpkgs repo maintainer list, you will have to add yourself to the `modules/lib/maintainers.nix` file and change this line to

```suggestion
  meta.maintainers = [ hm.maintainers.tph5595 ];
```

No need to add real name if you feel uncomfortable with that, your GitHub user ID is 6002071:

```console
$ curl -s https://api.github.com/users/tph5595 | jq .id
6002071
```

tph5595:
I believe this is ready to be merged. @rycee @SuperSandro2000

modules/services/dropbox.nix
```diff
@@ -33,48 +33,42 @@ in {
         lib.platforms.linux)
     ];
 
-    home.packages = [ pkgs.dropbox-cli ];
+    home.packages = [ pkgs.dropbox ];
 
     systemd.user.services.dropbox = {
       Unit = { Description = "dropbox"; };
 
       Install = { WantedBy = [ "default.target" ]; };
```
mactep, line 41: I'm not sure if this is a Gnome problem, but I got Dropbox to properly show on the system tray by changing this:
```suggestion
      Unit = {
        Description = "dropbox";
        After = [ "graphical-session.target" ];
      };

      Install = { WantedBy = [ "graphical-session.target" ]; };
```

PR #5929, opened 2024-10-03
kubecolor: add module

ajgon:
### Description

Add support for [kubecolor](https://github.com/kubecolor/kubecolor) program.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/kubecolor.nix
+++ b/modules/programs/kubecolor.nix
@@ -1,0 +1,87 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.programs.kubecolor;
+  yamlFormat = pkgs.formats.yaml { };
+  inherit (pkgs.stdenv.hostPlatform) isDarwin;
+
+in {
+  meta.maintainers = with maintainers; [ ajgon ];
+
+  options.programs.kubecolor = {
+    enable = mkEnableOption "kubecolor - Colorize your kubectl output";
+
+    package = mkPackageOption pkgs "kubecolor" { };
+
+    enableAlias = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        When set to true, it will create an alias for kubectl pointing to
+        kubecolor, thus making kubecolor the default kubectl client.
+      '';
+    };
+
+    settings = mkOption {
+      type = yamlFormat.type;
+      default = { };
+      example = literalExpression ''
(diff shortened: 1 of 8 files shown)


modules/programs/kubecolor.nix
```diff
@@ -0,0 +1,82 @@
...
+    # https://github.com/kubecolor/kubecolor/pull/145
+    configPathSuffix = if cfg.package.pname == "kubecolor"
+    && lib.strings.toInt (lib.versions.major cfg.package.version) == 0
+    && lib.strings.toInt (lib.versions.minor cfg.package.version) < 4 then
+      "kube/"
+    else
+      "kube/color.yaml";
+
+  in mkIf cfg.enable {
+    home.enableNixpkgsReleaseCheck = false;
```
rycee, line 57: Need to remove this.

modules/programs/kubecolor.nix
```diff
@@ -0,0 +1,82 @@
...
+        Configuration written to {file}`$XDG_CONFIG_HOME/kube/color.yaml`
+        (Linux) or {file}`Library/Application Support/kube/color.yaml` (Darwin),
+        See <https://kubecolor.github.io/reference/config/> for supported
+        values.
+      '';
+    };
+  };
+
+  config = let
+    enableXdgConfig = !isDarwin || config.xdg.enable;
```
rycee, line 46: Overriding the default configuration path with an XDG path should only happen when `config.home.preferXdgDirectories == true`.

When `config.home.preferXdgDirectories == false` I would expect the default configuration path to be used, i.e., the `KUBECOLOR_CONFIG` environment variable being unnecessary.

modules/programs/kubecolor.nix
```diff
@@ -0,0 +1,82 @@
...
+
+in {
+  meta.maintainers = with maintainers; [ ajgon ];
+
+  options.programs.kubecolor = {
+    enable = mkEnableOption "kubecolor - Colorize your kubectl output";
+
+    package = mkPackageOption pkgs "kubecolor" { };
+
+    enableAlias = mkOption {
```
rycee, line 19: Note, I rebased this branch on latest master and changed this option name to `enableAlias` to better match the naming in other modules.

ajgon:
@rycee Thanks for the review. I adjusted the PR, and added changes in separate commit.

rycee:

PR #5930, opened 2024-10-04
espanso: add wayland support

phanirithvij:
Tested on
- kde plasma x11 
- kde plasma wayland
- niri wm (wayland)
- xfce (x11)

### DIFF
--- a/modules/misc/news/2025-04-26_18-56-13.nix
+++ b/modules/misc/news/2025-04-26_18-56-13.nix
@@ -1,0 +1,11 @@
+{ pkgs, config, ... }:
+{
+  time = "2025-04-26T13:26:13+00:00";
+  condition = pkgs.hostPlatform.isLinux && config.services.espanso.enable;
+  message = ''
+    `services.espanso` now supports wayland.
+    This is enabled by default on Linux as `services.espanso.waylandSupport = true;`.
+    Depending on your graphical session type, you may disable one of `services.espanso.x11Support` and `services.espanso.waylandSupport` to reduce the closure size of espanso on your system.
+    Both x11 and wayland versions come enabled by default on Linux.
+  '';
+}
--- a/modules/services/espanso.nix
+++ b/modules/services/espanso.nix
@@ -34,6 +34,7 @@ in
     maintainers.bobvanderlinden
     lib.hm.maintainers.liyangau
     maintainers.n8henrie
+    maintainers.phanirithvij
   ];
   options = {
     services.espanso = {
--- a/modules/services/espanso.nix
+++ b/modules/services/espanso.nix
@@ -7,6 +7,7 @@
 let
   inherit (lib)
     mkOption
+    mkPackageOption
     mkEnableOption
     mkIf
     maintainers
@@ -16,8 +17,24 @@ let
     versionAtLeast
     ;
 
+  inherit (pkgs.stdenv.hostPlatform) isLinux;
+
   cfg = config.services.espanso;
   espansoVersion = cfg.package.version;
+
+  package-bin =
+    if isLinux && cfg.x11Support && cfg.waylandSupport then
+      pkgs.writeShellScriptBin "espanso" ''
+        if [ -n "$WAYLAND_DISPLAY" ]; then
+          ${lib.meta.getExe cfg.package-wayland} "$@"
+        else
+          ${lib.meta.getExe cfg.package} "$@"
+        fi
+      ''
+    else if isLinux && cfg.waylandSupport then
+      cfg.package-wayland
+    else
+      cfg.package;
 
   yaml = pkgs.formats.yaml { };
 in
@@ -45,6 +62,29 @@ in
         description = "Which espanso package to use";
         default = pkgs.espanso;
         defaultText = literalExpression "pkgs.espanso";
+      };
+
+      package-wayland =
+        mkPackageOption pkgs "espanso-wayland" {
+          nullable = true;
+          extraDescription = "Which `espanso` package to use when running under wayland.";
+        }
+        // {
+          default = if isLinux && cfg.waylandSupport then pkgs.espanso-wayland else null;
+        };
+
+      x11Support = mkOption {
+        type = types.bool;
+        description = "Whether to enable x11 support on linux";
+        default = isLinux;
+        defaultText = "`true` on linux";
+      };
+
+      waylandSupport = mkOption {
+        type = types.bool;
+        description = "Whether to enable wayland support on linux";
+        default = isLinux;
+        defaultText = "`true` on linux";
       };
 
       configs = mkOption {
@@ -126,9 +166,16 @@ in
           The services.espanso module only supports Espanso version 2 or later.
         '';
       }
+      {
+        assertion = isLinux -> (cfg.x11Support || cfg.waylandSupport);
+        message = ''
+          In services.espanso at least one of x11 or wayland support must be enabled on linux.
+        '';
+      }
     ];
 
-    home.packages = [ cfg.package ];
+    # conflicting to have cfg.package and cfg.package-wayland
+    home.packages = [ package-bin ];
 
     xdg.configFile =
       let
@@ -152,7 +199,7 @@ in
         Description = "Espanso: cross platform text expander in Rust";
       };
       Service = {
-        ExecStart = "${cfg.package}/bin/espanso launcher";
+        ExecStart = "${lib.meta.getExe package-bin} launcher";
         Restart = "on-failure";
         RestartSec = 3;
       };
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -62,7 +62,7 @@ let
     let
       imported = import (./news + "/${newsFile}");
     in
-    if builtins.isFunction imported then imported { inherit pkgs; } else imported
+    if builtins.isFunction imported then imported { inherit config lib pkgs; } else imported
   ) newsFiles;
 in
 {
--- a/tests/modules/services/espanso/basic-configuration.nix
+++ b/tests/modules/services/espanso/basic-configuration.nix
@@ -1,3 +1,6 @@
+espansoExtraArgs:
+{ config, ... }:
+
 {
   services.espanso = {
     enable = true;
@@ -42,7 +45,7 @@
         ];
       };
     };
-  };
+  } // espansoExtraArgs;
 
   nmt.script = ''
     serviceFile=home-files/.config/systemd/user/espanso.service
--- a/tests/modules/services/espanso/basic-configuration.nix
+++ b/tests/modules/services/espanso/basic-configuration.nix
@@ -46,8 +46,14 @@
 
   nmt.script = ''
     serviceFile=home-files/.config/systemd/user/espanso.service
+    expectedServiceFile=${./basic-configuration.service}
     assertFileExists "$serviceFile"
-    assertFileContent "$serviceFile" ${./basic-configuration.service}
+    # ensure espanso is in ExecStart
+    assertFileRegex "$serviceFile" 'ExecStart=.*/bin/espanso launcher'
+    # rest of the file should be identical
+    grep -v "/bin/espanso launcher" "$(_abs $serviceFile)" > espanso-service.actual
+    grep -v "/bin/espanso launcher" "$expectedServiceFile" > espanso-service.expected
+    assertFileContent "$(realpath espanso-service.actual)" espanso-service.expected
 
     configFile=home-files/.config/espanso/config/default.yml
     assertFileExists "$configFile"
--- a/tests/modules/services/espanso/default.nix
+++ b/tests/modules/services/espanso/default.nix
@@ -1,1 +1,11 @@
-{ espanso-basic-configuration = ./basic-configuration.nix; }
+{
+  espanso-basic-configuration = import ./basic-configuration.nix { };
+  espanso-basic-configuration-wayland = import ./basic-configuration.nix {
+    waylandSupport = true;
+    x11Support = false;
+  };
+  espanso-basic-configuration-x11 = import ./basic-configuration.nix {
+    waylandSupport = false;
+    x11Support = true;
+  };
+}


phanirithvij:
So, this requires both wayland and x11 binaries installed. I guess that's why it wasn't implemented before.

I guess it's a dud then.

phanirithvij:
Came up with a weird hack. I don't know if it belongs here.
One thing is clear, for it to work seamlessly without explicitly requiring user to choose wayland or x11 the only solution is for the upstream espanso package should provide a `fat` binary (not possible currently).

In this pr the fat binary is `package-bin` which brings in both the verisons of espanso (x11, wayland) into the store.

But I guess on darwin there is no such thing as wayland and thus the `pkgs.espanso` should be chosen always. I don't have access to a darwin machine to debug/test this.

phanirithvij:
so I've tried `espanso-wayland = callPackage ../applications/office/espanso { ... };` in own nixpkgs instance instead of using override and got
```shell
$ nix build .#test-espanso-basic-configuration
> --- actual
> +++ expected
> @@ -2,7 +2,7 @@
>  WantedBy=default.target
>
> [Service]
> -ExecStart=/nix/store/nq21hfj3v74hd640ggzvmdd2zgmqv51y-espanso/bin/espanso daemon
> +ExecStart=@espanso@/bin/espanso daemon
> Restart=on-failure
```

phanirithvij:
Fixed the test (with the `@espanso@` stub) but I don't know how to fix the `espanso.override attribute missing` error.

and the ci failed because of something unrelated.

phanirithvij:
@rycee I need some help, please have a look at it when you're free.

```shell
$ nix build .#test-espanso-basic-configuration
error:
         at <nix/derivation-internal.nix>:34:12:
           33|
           34|   strict = derivationStrict drvAttrs;
             |            ^
           35|

       … while evaluating derivation 'nmt-test-espanso-basic-configuration'
         whose name attribute is located at /nix/store/riqkpszjqk02bi1wppfg8ip5xvh102qd-source/pkgs/stdenv/generic/make-derivation.nix:336:7

       … while evaluating attribute 'buildCommand' of derivation 'nmt-test-espanso-basic-configuration'
         at /nix/store/riqkpszjqk02bi1wppfg8ip5xvh102qd-source/pkgs/build-support/trivial-builders/default.nix:59:17:
             |                 ^

       … while evaluating the option `home.activation.installPackages.data':

       … while evaluating definitions from `/nix/store/vv3xqpi8hwcciy7labjysw3s6qwafch4-source/modules/home-environment.nix':

       … while evaluating the option `services.espanso.package-wayland':

       error: attribute 'override' missing
       at /nix/store/riqkpszjqk02bi1wppfg8ip5xvh102qd-source/pkgs/top-level/all-packages.nix:4857:21:
         4856|   };
         4857|   espanso-wayland = espanso.override {
             |                     ^
         4858|     x11Support = false;
```

phanirithvij:
The eval error was fixed in 7a3f0b3b8df1b913890f5fad04d34f2af91da858 not sure how exactly (git bisect on home-manger)

phanirithvij:
@n8henrie @lucasew @liyangau @bobvanderlinden

lucasew (approved):

modules/services/espanso.nix
```diff
@@ -16,9 +16,25 @@ let
...
+    if isLinux && cfg.x11Support && cfg.waylandSupport then
+      pkgs.writeShellScriptBin "espanso" ''
+        if [ -n "$WAYLAND_DISPLAY" ]; then
+          ${lib.meta.getExe cfg.package-wayland} "$@"
+        else
+          ${lib.meta.getExe cfg.package} "$@"
+        fi
+      ''
+    else if isLinux && cfg.waylandSupport then
+      cfg.package-wayland
```
phanirithvij, line 34: Missed this previously. will not allow user to override `package-wayland` but works because the default is still correct.

modules/services/espanso.nix
```diff
@@ -42,10 +59,31 @@ in
...
+        default = if isDarwin || cfg.x11Support then pkgs.espanso else pkgs.espanso-wayland;
         defaultText = literalExpression "pkgs.espanso";
       };
 
+      package-wayland = mkOption {
+        type = types.package;
+        description = "Which espanso package to use when running under wayland on linux";
+        default = if isLinux && cfg.waylandSupport then pkgs.espanso-wayland else pkgs.espanso;
+        defaultText = literalExpression "pkgs.espanso-wayland";
+      };
```
khaneliman, line 71: This makes no sense, to me. We already have a package option. You can change the desired package there.

modules/services/espanso.nix
```diff
@@ -16,9 +16,25 @@ let
...
 
+  package-bin =
+    if isLinux && cfg.x11Support && cfg.waylandSupport then
+      pkgs.writeShellScriptBin "espanso" ''
+        if [ -n "$WAYLAND_DISPLAY" ]; then
+          ${lib.meta.getExe cfg.package-wayland} "$@"
+        else
+          ${lib.meta.getExe cfg.package} "$@"
+        fi
+      ''
```
khaneliman, line 32: Why do we need to create a script to choose which binary to launch? Just set the package you want and just use it.

modules/services/espanso.nix
```diff
@@ -42,10 +59,31 @@ in
       package = mkOption {
         type = types.package;
         description = "Which espanso package to use";
-        default = pkgs.espanso;
+        default = if isDarwin || cfg.x11Support then pkgs.espanso else pkgs.espanso-wayland;
```
khaneliman, line 62: Would make more sense, to me, to keep the safe default and show the `espanso-wayland` package as an example alternative.

modules/services/espanso.nix
phanirithvij: what would be your suggestion, for espanso to seamlessly work on both x11 and wayland by default.

modules/services/espanso.nix
phanirithvij: what would be your suggestion, for espanso to seamlessly work on both x11 and wayland by default?
Say If I switch from kde plasma x11 to wayland by logging out and logging in.

khaneliman: `espanso-wayland` doesn't run on X11?

phanirithvij: No

n8henrie: > espanso-wayland doesn't run on X11?

The `espanso` crate's `wayland` feature activates different logic for reading keyboard input.

khaneliman: Hmm.. just feels like a lot of module options for a niche case of someone not committing to wayland or x11... 
Could change the example to be the script here, really...

modules/services/espanso.nix
phanirithvij: This is to customise the wayland version of the package, say a patched version of espanso-wayland.

modules/services/espanso.nix
```diff
@@ -125,9 +163,16 @@ in
           The services.espanso module only supports Espanso version 2 or later.
         '';
       }
+      {
+        assertion = isLinux -> (cfg.x11Support || cfg.waylandSupport);
+        message = ''
+          In services.espanso at least one of x11 or wayland support must be enabled on linux.
+        '';
+      }
```
khaneliman, line 171: This isn't true, is it? You have a if else condition that falls back to the original behavior of using the normal package.

modules/services/espanso.nix
```diff
@@ -42,10 +59,31 @@ in
...
+        type = types.bool;
+        description = "Whether to enable x11 support on linux";
+        default = isLinux;
+        defaultText = literalExpression "enabled on linux";
+      };
+
+      waylandSupport = mkOption {
+        type = types.bool;
+        description = "Whether to enable wayland support on linux";
+        default = isLinux;
```
khaneliman, line 83: This will cause all linux users to unknowingly have a wayland package, all of a sudden. But, it does seem like a well intentioned change after the previous discussion.

modules/services/espanso.nix
phanirithvij: It allows anyone to run espanso by default is how I would look at it, without doing any debugging on why the espanso service is not starting/failing.

khaneliman: That's fair.. just wanted to challenge the ideas here because it feels very weird.

modules/services/espanso.nix
phanirithvij: If the user (accidentally?) sets both to be false then it is better to advice the user, it makes no sense to disable both x11 and wayland.

If we wish to remove this warning. The overall expression does default to the wayland version if both are set to false. (`cfg.package` resolves to `pkgs.espanso-wayland` and `package-bin` resolves to `cfg.package`)

 but I think it is best to leave it in because its an invalid state.

khaneliman: We shouldn't be changing the default package to wayland version if it doesn't support X11, that'd be an unexpected breaking change. You added a wayland package specifically for that reason.

modules/services/espanso.nix
phanirithvij: I added an explaination in the news entry describing this. 

Yeah, this is needed for the service to start properly. Is there a better way to communicate to users that `waylandSupport` needs to be decided on without breaking eval?

maybe `lib.warn if waylandSupport == null "services.espanso.waylandSupport needs to be set to either true or false"`

modules/services/espanso.nix
khaneliman: This isn't a warning, it's an assertion. Which means the eval cannot continue, but there's no reason for it not to since we handle the configuration fine. If both new options are set to false, it should default to the legacy behavior and issue a warning, imo.

phanirithvij: yeah meant to say assertion. and since it is an invalid state to be in disabling both x11 and wayland support IF we are on linux then better to break eval

modules/services/espanso.nix
khaneliman: > I added an explaination in the news entry describing this.

The news entry will only be seen by those who use that functionality, though. Which, I don't know how many actually do.. 

>Yeah, this is needed for the service to start properly. Is there a better way to communicate to users that waylandSupport needs to be decided on without breaking eval?
>
>maybe lib.warnIf waylandSupport == null "services.espanso.waylandSupport needs to be set to either true or false"

Since this is all new confguration, I don't see why we would force users to have to change their configuration to silence a warning that might not apply to them. But, without checking against all the possible wayland platforms enabled, I'm not sure of a better way to gracefully enable it for those who might expect it to work...

modules/services/espanso.nix
khaneliman: I think this should just be strictly a default of `espanso-wayland`, at least. Makes no sense for the wayland package to not be wayland compatible by default, ever.

phanirithvij: yes these `default` conditions for both `package` and `package-wayland` are actually redundant I realised. the package-bin does handle all of them fine.

modules/services/espanso.nix
```diff
@@ -46,6 +63,27 @@ in
         defaultText = literalExpression "pkgs.espanso";
       };
 
+      package-wayland = mkOption {
+        type = types.package;
+        description = "Which espanso package to use when running under wayland on linux";
+        default = pkgs.espanso-wayland;
+        defaultText = literalExpression "pkgs.espanso-wayland";
+      };
```
khaneliman, line 71:
```suggestion
      package-wayland = mkPackageOption pkgs "espanso-wayland {
        nullable = true;
        extraDescription = "Which `espanso` package to use when running under wayland.";
      } // { 
        default = if isLinux && cfg.waylandSupport then pkgs.espanso-wayland else null; };
```

We should utilize `mkPackageOption` and perhaps just null it out on darwin, since it never applies.

modules/services/espanso.nix
```diff
@@ -46,6 +63,26 @@ in
...
+        nullable = true;
+        extraDescription = "Which `espanso` package to use when running under wayland.";
+      } // { 
+        default = if isLinux && cfg.waylandSupport then pkgs.espanso-wayland else null; };
+
+      x11Support = mkOption {
+        type = types.bool;
+        description = "Whether to enable x11 support on linux";
+        default = isLinux;
+        defaultText = literalExpression "enabled on linux";
```
khaneliman, line 76:
```suggestion
        defaultText = literalExpression "`true` on linux";
```

modules/services/espanso.nix
```diff
@@ -46,6 +63,26 @@ in
...
+      };
+
+      waylandSupport = mkOption {
+        type = types.bool;
+        description = "Whether to enable wayland support on linux";
+        default = isLinux;
+        defaultText = literalExpression "enabled on linux";
```
khaneliman, line 83:
```suggestion
        defaultText = literalExpression "`true` on linux";
```

khaneliman (commented):
While combing through again, was wondering if we could simplify even further... Would it be a better UX to just have the 2 package options and not use the `support` options. We default there to be a wayland package enabled for linux users and then can just null it out when they don't need that support. This way, we can decide to use the wrapper when both package options are set and disabling either of them indicates support.

phanirithvij:
this is how the UX looks like as of now.

```nix
services.espanso.enable = true;
services.espanso = {
    enable = true;
    waylandSupport = false;
};
services.espanso = {
    enable = true;
    x11Support = false;
};
```

Let me know if I misunderstood something but you're suggesting it to be

```nix
services.espanso.enable = true; # UNCHANGED
services.espanso = {
    enable = true;
    package = null;
};
services.espanso = {
    enable = true;
    package-wayland = null;
};
```

is that right?

if so I'd prefer the first one because I can tell right away what's happening.

khaneliman:
Alright, sounds good. Annoying you have to do all this just because of their packaging/runtime issues.

modules/misc/news/2025-04-26_18-56-13.nix
```diff
@@ -0,0 +1,11 @@
+{ pkgs, config, ... }:
+{
+  time = "2025-04-26T13:26:13+00:00";
+  condition = pkgs.hostPlatform.isLinux && config.services.espanso.enable;
```
zowoq, line 4: `pkgs.hostPlatform` is an alias. https://github.com/nix-community/home-manager/pull/6931

n8henrie: @zowoq for my learning, why is using an alias a problem here? Actually, NM -- looks like that's made pretty clear [here](https://github.com/NixOS/nixpkgs/blob/master/pkgs/top-level/aliases.nix). Did you pick this up just from memory? Or is there a linter or something that might help me notice this going forward?

zowoq: nixpkgs has an [allowAliases](https://github.com/NixOS/nixpkgs/blob/88b87b7a62ef7c1c70510097f01f656a428d293e/pkgs/top-level/config.nix#L86) option.

n8henrie: Thanks!

PR #5934, opened 2024-10-06
zathura: Fix the type for config options

akshettrj:
### Description

Changed the type for `programs.zathura.options` to accept float values as well. You can check their man page (https://manpages.ubuntu.com/manpages/trusty/man5/zathurarc.5.html) as well. In the beginning, they mention:

```
  set - Changing options
       In  addition  to  the  build-in :set command zathura offers more options to be changed and
       makes those changes permanent. To overwrite  an  option  you  just  have  to  add  a  line
       structured like the following

          set <option> <new value>

       The option field has to be replaced with the name of the option that should be changed and
       the new value field has to be replaced with the new value the option should get. The  type
       of the value can be one of the following:

       • INT - An integer number

       • FLOAT - A floating point number

       • STRING - A character string

       • BOOL - A boolean value ("true" for true, "false" for false)
```

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rprospero

### DIFF
--- a/modules/programs/zathura.nix
+++ b/modules/programs/zathura.nix
@@ -31,7 +31,7 @@ in {
 
     options = mkOption {
       default = { };
-      type = with types; attrsOf (either str (either bool (either float int)));
+      type = with types; attrsOf (oneOf [ str bool int float ]);
       description = ''
         Add {option}`:set` command options to zathura and make
         them permanent. See
--- a/modules/programs/zathura.nix
+++ b/modules/programs/zathura.nix
@@ -31,7 +31,7 @@ in {
 
     options = mkOption {
       default = { };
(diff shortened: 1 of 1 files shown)


rprospero (approved):
Yeah, it should have been written this way in the first place.  Sorry I missed the floats!

modules/programs/zathura.nix
```diff
@@ -31,7 +31,7 @@ in {
 
     options = mkOption {
       default = { };
-      type = with types; attrsOf (either str (either bool int));
+      type = with types; attrsOf (either str (either bool (either float int)));
```
teto, line 34: rather than a cascade of either, using `oneOf ` would be more readable

PR #5940, opened 2024-10-07
eza: add options

refaelsh:
> Add more commonly use options

### Description

The following options were added:
1. --all
2. --ling
3. --extended
4. --header
5. --classify

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@cafkafk

### DIFF
--- a/modules/programs/eza.nix
+++ b/modules/programs/eza.nix
@@ -64,14 +64,61 @@ with lib;
       '';
     };
 
+    all = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Show hidden and 'dot' files.
+      '';
+    };
+
+    long = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Display extended details and attributes.
+      '';
+    };
+
+    extended = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        List each file’s extended attributes and sizes.
+      '';
+    };
+
+    header = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Add a header row to each column.
+      '';
+    };
+
+    clasiffy = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Display type indicator by file names.
+      '';
+    };
     package = mkPackageOption pkgs "eza" { };
   };
 
   config = let
     cfg = config.programs.eza;
 
-    args = escapeShellArgs (optional cfg.icons "--icons"
-      ++ optional cfg.git "--git" ++ cfg.extraOptions);
+    args = escapeShellArgs (
+      optional cfg.icons "--icons"
+      ++ optional cfg.git "--git"
+      ++ optional cfg.all "--all"
+      ++ optional cfg.long "--long"
+      ++ optional cfg.extended "--extended"
+      ++ optional cfg.header "--header"
+      ++ optional cfg.clasiffy "--classify"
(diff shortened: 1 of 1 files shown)


modules/programs/eza.nix
```diff
@@ -64,14 +64,61 @@ with lib;
...
+      '';
+    };
+
+    clasiffy = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Display type indicator by file names.
+      '';
+    };
```
cafkafk, line 105: `--classify` takes auto/always/never, so it would make more sense to turn this into an enum, and then default it to auto

cafkafk (changes requested):
Thanks for adding these. 

The classify option should be turned into an enum instead, and default to `auto`, since that should be the default.

modules/programs/eza.nix
refaelsh: Good suggestion. Done.

teto (changes requested):
My opinion is that it should be out of scope, mapping every option of each program doesn't scale, it's just too much. In documentation etc. A list of string is enough

refaelsh:
Thank you very much for your comment @teto. I agree with your statement on a sentimental level.
When you say: `A list of string is enough`, what exactly do you mean? Please provide an example in the context of the current PR.

cafkafk (commented):
Thanks for changing the default to `auto`

---

Wouldn't it be fine to keep this around for users that want it if it was gated behind a config option? I do think that yes, keeping this up to date may end up being annoying and brittle, but for users that want this, having the ability to use these options gated behind some `enable = true` would make sense. 

Consider this behavior has existed since exa since feb 2023 (introduced in https://github.com/nix-community/home-manager/commit/4bac34186886f8d484791702f32ec371f8c7a81e)

modules/programs/eza.nix
```diff
@@ -64,14 +64,61 @@ with lib;
...
+
+    header = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Add a header row to each column.
+      '';
+    };
+
+    clasiffy = mkOption {
```
rycee, line 99:
```suggestion
    classify = mkOption {
```

rycee:
Overall I agree with @teto that we should be restrictive about the number of options as it introduces maintenance complexity. These flags seem somewhat restricted in scope, though, so I could live with it as long as @cafkafk is OK with it and eza upstream don't consider these flags unstable or experimental.

cafkafk:
> Overall I agree with @teto that we should be restrictive about the number of options as it introduces maintenance complexity. These flags seem somewhat restricted in scope, though, so I could live with it as long as @cafkafk is OK with it and eza upstream don't consider these flags unstable or experimental.

They're not "experimental", but may be broken in future breaking releases. They will be marked as breaking changes in both our changelogs and by the `major` bumps (0.y.z, where y is acting as major here), as we conform to a stricter semver that gives the guarantee that we will not introduce breaking changes in patch versions, even if we're not 1.0 yet.

That said, it may be the case that these do incur too big of a maintenance overhead, I perhaps suggest that module maintainers should get final say on this, or that the author of the PR commits to potentially fixing breakage.

(even then, I wonder how this will be kept in sync when there is breaking changes, as eza updates come from nixpkgs, the nescesarry module option changes can't be "transactionally" added with new eza versions, the interface does seem somewhat brittle)

PR #5946, opened 2024-10-08
atuin: configure daemon using systemd and launchd

### LINKED ISSUE #5797: Atuin daemon option
### Description

Add an option to manage the Atuin daemon with a user service.

[daemon](https://docs.atuin.sh/reference/daemon/#atuin-daemon)

water-sucks:
### Description

This configures the atuin daemon for Linux and Darwin systems using systemd and launchd, respectively. For systemd, a socket is also automatically configured to exist at atuin's default socket location.

I also added myself as a maintainer of this module.

Closes #5797.

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@hawkw

### DIFF
--- a/modules/programs/atuin.nix
+++ b/modules/programs/atuin.nix
@@ -145,11 +145,19 @@ in {
 
     (mkIf cfg.daemon.enable (mkMerge [
       {
-        assertions = [{
-          assertion = isLinux || isDarwin;
-          message =
-            "The atuin daemon can only be configured on either Linux or macOS.";
-        }];
+        assertions = [
+          {
+            assertion = versionAtLeast cfg.package.version "18.2.0";
+            message = ''
+              The atuin daemon requires at least version 18.2.0 or later.
+            '';
+          }
+          {
+            assertion = isLinux || isDarwin;
+            message =
+              "The atuin daemon can only be configured on either Linux or macOS.";
+          }
+        ];
 
         programs.atuin.settings = { daemon = { enabled = true; }; };
       }
--- a/modules/programs/atuin.nix
+++ b/modules/programs/atuin.nix
@@ -8,6 +8,7 @@ let
 
   tomlFormat = pkgs.formats.toml { };
 
+  inherit (pkgs.stdenv) isLinux isDarwin;
 in {
   meta.maintainers = [ maintainers.hawkw ];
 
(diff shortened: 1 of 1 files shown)


shymega:
LGTM, but I would add a deprecation warning when the key `programs.atuin.settings.daemon.enabled` is set, and advise the user to use the new setting. 

Alternatively, but perhaps the new `programs.atuin.daemon` key should become an attrset itself, and, if `programs.atuin.settings.daemon` exists, then the new setting should just copy over the `programs.atuin.settings.daemon` attrset into itself?

Also, the Atuin daemon was only [merged](https://github.com/atuinsh/atuin/pull/2006) on the 8th of May, and this was before Atuin v18.2.0. In Nixpkgs 24.05, the version of Atuin is v18.2.0.

Therefore, the PR should perform an assertion to check the version of Atuin used before enabling the daemon - otherwise users will run into errors.

water-sucks:
> Also, the Atuin daemon was only https://github.com/atuinsh/atuin/pull/2006 on the 8th of May, and this was before Atuin v18.2.0. In Nixpkgs 24.05, the version of Atuin is v18.2.0.

I have added the assertion for the version, thanks for that tip.

> LGTM, but I would add a deprecation warning when the key programs.atuin.settings.daemon.enabled is set, and advise the user to use the new setting.

I'm not sure about adding a warning for this, because the `daemon.enabled` key does have to be set in the atuin configuration for it to be using the daemon; in fact, I do set this in the config when the `cfg.daemon.enable` option is set, so I'm not sure how I would warn the user about them setting it manually, or if that's even possible.

shymega:
> > Also, the Atuin daemon was only [atuinsh/atuin#2006](https://github.com/atuinsh/atuin/pull/2006) on the 8th of May, and this was before Atuin v18.2.0. In Nixpkgs 24.05, the version of Atuin is v18.2.0.
> 
> I have added the assertion for the version, thanks for that tip.

np

> 
> > LGTM, but I would add a deprecation warning when the key programs.atuin.settings.daemon.enabled is set, and advise the user to use the new setting.
> 
> I'm not sure about adding a warning for this, because the `daemon.enabled` key does have to be set in the atuin configuration for it to be using the daemon; in fact, I do set this in the config when the `cfg.daemon.enable` option is set, so I'm not sure how I would warn the user about them setting it manually, or if that's even possible.

OK, so in the Atuin struct for that table, there's an alias - it can either be `enabled` or `alias`. But that's on HEAD, so I'm not certain.

I think for now it's OK.

modules/programs/atuin.nix
```diff
@@ -94,49 +95,117 @@ in {
...
+            Description = "atuin daemon";
+            Requires = [ "atuin-daemon.socket" ];
+          };
+          Install = {
+            Also = [ "atuin-daemon.socket" ];
+            WantedBy = [ "default.target" ];
+          };
+          Service = {
+            ExecStart = "${lib.getExe cfg.package} daemon";
+            Environment = [ "ATUIN_LOG=info" ];
```
shymega, line 178: I would add a `cfg` key behind `lib.optionals` for this - people might not like Atuin's daemon being too chatty.

modules/programs/atuin.nix
```diff
@@ -94,49 +95,117 @@ in {
...
+        ];
+
+        programs.atuin.settings = { daemon = { enabled = true; }; };
+      }
+      (mkIf isLinux {
+        programs.atuin.settings = { daemon = { systemd_socket = true; }; };
+
+        systemd.user.services.atuin-daemon = {
+          Unit = {
+            Description = "atuin daemon";
```
shymega, line 169: 'atuin should be 'Atuin'.

modules/programs/atuin.nix
```diff
@@ -94,49 +95,117 @@ in {
...
+            ExecStart = "${lib.getExe cfg.package} daemon";
+            Environment = [ "ATUIN_LOG=info" ];
+            Restart = "on-failure";
+            RestartSteps = 3;
+            RestartMaxDelaySec = 6;
+          };
+        };
+
+        systemd.user.sockets.atuin-daemon = {
+          Unit = { Description = "atuin daemon socket"; };
```
shymega, line 186: 'atuin should be 'Atuin'.

modules/programs/atuin.nix
```diff
@@ -94,49 +95,117 @@ in {
...
+            RemoveOnStop = true;
+          };
+        };
+      })
+      (mkIf isDarwin {
+        launchd.agents.atuin-daemon = {
+          enable = true;
+          config = {
+            ProgramArguments = [ "${lib.getExe cfg.package}" "daemon" ];
+            EnvironmentVariables = { ATUIN_LOG = "info"; };
```
shymega, line 200: I would add a cfg key behind lib.optionals for this - people might not like Atuin's daemon being too chatty.

shymega (changes requested):
Looks good so far, just a few more nitpicks.

shymega (approved):
LGTM. Can't wait to get this landed!

water-sucks:
Hmm, I'm not sure why this Ubuntu test is failing. Seems odd, since it failed twice with two different error messages when I just pushed up the same commit with a different hash just for testing.

shymega:
Have you rebased against master?

rycee:

modules/programs/atuin.nix
```diff
@@ -94,49 +96,132 @@ in {
...
+            RestartSteps = 3;
+            RestartMaxDelaySec = 6;
+          };
+        };
+
+        systemd.user.sockets.atuin-daemon = {
+          Unit = { Description = "Atuin daemon socket"; };
+          Install = { WantedBy = [ "sockets.target" ]; };
+          Socket = {
+            ListenStream = "%h/.local/share/atuin/atuin.sock";
```
ambroisie, line 202: Shouldn't this use `%D` to refer to `$XDG_DATA_HOME`?

shymega: We could well change this in the future. Perhaps you could make a PR to fix this, and link it to this PR?

mmlb: https://github.com/nix-community/home-manager/pull/6187

shymega: Thanks 👍🏻

PR #5947, opened 2024-10-08
eza: Fix icons option

### LINKED ISSUE #5800: bug: eza completions not working

### Issue description

The shell completions for `eza` in zsh are not working correctly. 

with `enableZshIntegration = true`, this only happens for `ls`, not for i.e. `ll` or others.

If I run `ls`, I get the follwing output
```
ls <TAB>
always auto automatic never
``` 
This options match the `--classify` options, what ever that means.

With `enableZshIntegration = false`, I have the alias `eza --icons` in my config. Running `eza <TAB>` show the same, wrong completions while `eza --icons <TAB>` shows correct completions (using the same alias of course).

### Maintainer CC

@stasjok 

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.8.0-40-generic, Ubuntu, 22.04.4 LTS (Jammy Jellyfish), nobuild`
 - multi-user?: `yes`
 - sandbox: `yes`
 - channels(root): `"nixpkgs"`
```

kuznetsss:
### Description

Fix icons option for eza which was breaking completion in zsh.

Fixes #5800.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@cafkafk

### DIFF
--- a/modules/programs/eza.nix
+++ b/modules/programs/eza.nix
@@ -49,10 +49,13 @@ with lib;
     };
 
     icons = mkOption {
-      type = types.bool;
-      default = false;
+      type = types.enum [ null true false "auto" "always" "never" ];
+      default = null;
       description = ''
         Display icons next to file names ({option}`--icons` argument).
+
+        Note, the support for Boolean values is deprecated.
+        Setting this option to `true` corresponds to `--icons=auto`.
       '';
     };
 
@@ -70,8 +73,15 @@ with lib;
   config = let
     cfg = config.programs.eza;
 
-    args = escapeShellArgs (optional cfg.icons "--icons"
-      ++ optional cfg.git "--git" ++ cfg.extraOptions);
+    iconsOption = let
+      v = if isBool cfg.icons then
+        (if cfg.icons then "auto" else null)
+      else
+        cfg.icons;
+    in optionals (v != null) [ "--icons" v ];
+
+    args = escapeShellArgs
+      (iconsOption ++ optional cfg.git "--git" ++ cfg.extraOptions);
 
     optionsAlias = optionalAttrs (args != "") { eza = "eza ${args}"; };
 
@@ -83,6 +93,12 @@ with lib;
       lla = "eza -la";
     };
   in mkIf cfg.enable {
+    warnings = optional (isBool cfg.icons) ''
+      Setting programs.eza.icons to a Boolean is deprecated.
+      Please update your configuration so that
+
+        programs.eza.icons = ${if cfg.icons then ''"auto"'' else "null"}'';
+
     home.packages = [ cfg.package ];
 
     programs.bash.shellAliases = optionsAlias
--- a/docs/release-notes/rl-2411.md
+++ b/docs/release-notes/rl-2411.md
@@ -18,6 +18,11 @@ This release has the following notable changes:
   add `-w` to your assignment of
   [services.swayidle.extraArgs](#opt-services.swayidle.extraArgs).
 
+- Support for Boolean values in the option
+  [programs.eza.icons](#opt-programs.eza.icons) is deprecated for
+  future removal. The new value for `true` is `"auto"`, and for
+  `false` it is `null`.
+
 ## State Version Changes {#sec-release-24.11-state-version-changes}
 
 The state version in this release includes the changes below. These


modules/programs/eza.nix
```diff
@@ -49,10 +49,10 @@ with lib;
     };
 
     icons = mkOption {
-      type = types.bool;
+      type = types.either types.bool (types.enum [ "auto" "always" "never" ]);
```
cafkafk, line 52: It may be worth it to just remove the bool type and have users set this option explicitly, I think it could be confusing in some cases that it defaults to auto and users then wondering why the icons don't show up when piped. Alternatively keep the bool as a type but use it to send a deprecation notice to the user that the bool option is being deprecated and will be removed next release? (I'm not sure about home-managers deprecation process)

kuznetsss: I agree and I'm happy to implement any of the options you suggested. @stasjok could you, please, advice about breaking changes here?

stasjok: > I agree and I'm happy to implement any of the options you suggested. @stasjok could you, please, advice about breaking changes here?

I'm not a maintainer here, so I don't know either. But https://nix-community.github.io/home-manager/index.xhtml#sec-guidelines-back-compat states that the changes should be backward compatible. I think a warning should be enough, see examples in the codebase: https://github.com/search?q=repo%3Anix-community%2Fhome-manager+warnings+deprecated&type=code. There are a couple of warnings when usages of options in the old format are deprecated in favor of the new format. I think it's the best course of action here.

rycee: I would propose something along the lines of the following patch. Note, I added `null` as a possible choice, indicating that the command line option should be omitted entirely.

``` patch
diff --git a/docs/release-notes/rl-2411.md b/docs/release-notes/rl-2411.md
index 6f7a6b7b0..626624640 100644
--- a/docs/release-notes/rl-2411.md
+++ b/docs/release-notes/rl-2411.md
@@ -18,6 +18,11 @@ This release has the following notable changes:
   add `-w` to your assignment of
   [services.swayidle.extraArgs](#opt-services.swayidle.extraArgs).
 
+- Support for Boolean values in the option
+  [programs.eza.icons](#opt-programs.eza.icons) is deprecated for
+  future removal. The new value for `true` is `"auto"`, and for
+  `false` it is `null`.
+
 ## State Version Changes {#sec-release-24.11-state-version-changes}
 
 The state version in this release includes the changes below. These
diff --git a/modules/programs/eza.nix b/modules/programs/eza.nix
index fd88ab150..679f53499 100644
--- a/modules/programs/eza.nix
+++ b/modules/programs/eza.nix
@@ -49,10 +49,13 @@ with lib;
     };
 
     icons = mkOption {
-      type = types.either types.bool (types.enum [ "auto" "always" "never" ]);
-      default = false;
+      type = types.enum [ null true false "auto" "always" "never" ];
+      default = null;
       description = ''
-        Display icons next to file names ({option}`--icons` argument). Setting this option to true uses `--icons=auto`
+        Display icons next to file names ({option}`--icons` argument).
+
+        Note, the support for Boolean values is deprecated.
+        Setting this option to `true` corresponds to `--icons=auto`.
       '';
     };
 
@@ -69,12 +72,13 @@ with lib;
 
   config = let
     cfg = config.programs.eza;
-    iconsOption = if isBool cfg.icons then
-      optionals cfg.icons [ "--icons" "auto" ]
-    else [
-      "--icons"
-      cfg.icons
-    ];
+
+    iconsOption = let
+      v = if isBool cfg.icons then
+        (if cfg.icons then "auto" else null)
+      else
+        cfg.icons;
+    in optionals (v != null) [ "--icons" v ];
 
     args = escapeShellArgs
       (iconsOption ++ optional cfg.git "--git" ++ cfg.extraOptions);
@@ -89,6 +93,12 @@ with lib;
       lla = "eza -la";
     };
   in mkIf cfg.enable {
+    warnings = optional (isBool cfg.icons) ''
+      Setting programs.eza.icons to a Boolean is deprecated.
+      Please update your configuration so that
+
+        programs.eza.icons = ${if cfg.icons then ''"auto"'' else "null"}'';
+
 
     programs.bash.shellAliases = optionsAlias
```

kuznetsss: Applied the path. Thanks everyone for reviewing!

rycee:

PR #5948, opened 2024-10-09
vscode: Get paths from product.json

RossBrunton:
### Description

The VSCode packages contain a product.json which is used to determine the extension and config dirs. Instead of hardcoding the package name, this change parses this file (which may appear in one of two places) and uses that to generate the paths.

This hopefully allows more VSCode derivatives to "just work" without home-manager updates.

### Checklist

- [x] Change is backwards compatible. (Verified there is no change in directories using `pkgs.vscode`, `pkgs.vscodium` and `pkgs.openvscodeserver`)

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@Rucadi
@rycee
@ncfavier

This package has no listed maintainer, so apologies if none of you are that familiar with it.


terlar:
Nice, I am pretty sure this will make my PR obsolete as well #5568

terlar:
Yeah, that could be another PR. However having another look at this, perhaps the value should be an option, because now this is switching to IFD, right? and there is no way to opt-out.

RossBrunton:
Apologies for the delay, but just pushed an update which adds `nameShort` and `dataFolderName` options (so named to match the appropriate values in the product.json). Let me know if the descriptions make sense.

@0x50F1A Noticed that you added a new VSCode variant. I'm assuming that Windsurf's `product.json` is in the same place as vanilla VSCode and VSCodium, so this should also just work for that as well.

PR #5953, opened 2024-10-11
pls: fixed perm argument to pass via pls

pwnwriter:
The argument should be perm not perms

### Description

The argument should be `perm` instead of `perms`. 

Screenshot 2024-09-28 at 5 15 08 PM

I changed the argument from `perms` to `perm`. Just a little typo. 

### Checklist

  - [x] Test cases run successfully.
  - [x] No errors or warnings encountered during testing.

- [] Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).
  - [] Coverage for edge cases considered.

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@arjan-s

### DIFF
--- a/modules/programs/pls.nix
+++ b/modules/programs/pls.nix
@@ -9,7 +9,7 @@ let
   aliases = {
     ls = "${cfg.package}/bin/pls";
     ll =
-      "${cfg.package}/bin/pls -d perms -d user -d group -d size -d mtime -d git";
+      "${cfg.package}/bin/pls -d perm -d user -d group -d size -d mtime -d git";
   };
 
 in {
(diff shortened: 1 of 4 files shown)


pwnwriter:
I didn't check the test cases, It should pass now @arjan-s Could you please verify once :3

rycee:
I think this is uncontroversial so I've merged this now, hope you don't mind @arjan-s. Thanks for the contribution @pwnwriter!

arjan-s:
The change looks good to me. I didn't test it this time around because I haven't used pls in a long time. Because of that I would also be happy to be removed as maintainer.

rycee:
@arjan-s You can remove yourself in a small PR. Thanks for the help 🙂

Issue #5954, opened 2024-10-11
bug: fzf bash completion fails in combination with zoxide --cmd cd


talhaHavadar:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Hello all,

When home-manager configures as below:
```
  programs = {
    home-manager.enable = true;
    starship.enable = true;
    zoxide = {
      enable = true;
      options = [
        "--cmd cd"
      ];
    };
    fzf.enable = true;
};
```

It generates a bashrc as below: 
```
...
if [[ ! -v BASH_COMPLETION_VERSINFO ]]; then
  . "/nix/store/k0ag90vp8qszsd1y85b0vq8dwvkxzjz9-bash-completion-2.14.0/etc/profile.d/bash_completion.sh"
fi

if [[ :$SHELLOPTS: =~ :(vi|emacs): ]]; then
  eval "$(/nix/store/2f27qfzb3b3s09rjfz0g06g03igibnx1-fzf-0.55.0/bin/fzf --bash)"
fi

eval "$(/nix/store/x2v8rifhny7l55rfddaxgcaplaw0r2yy-zoxide-0.9.6/bin/zoxide init bash --cmd cd)"
...
```

This is somehow causing a conflict which results as fzf completions not working at all.

I am not able to use cd <path>/** or kill -9 ** to trigger fzf so it makes ' ** ' unusable. But when I change the order of eval blocks as below:
```
...
if [[ ! -v BASH_COMPLETION_VERSINFO ]]; then
  . "/nix/store/k0ag90vp8qszsd1y85b0vq8dwvkxzjz9-bash-completion-2.14.0/etc/profile.d/bash_completion.sh"
fi

eval "$(/nix/store/x2v8rifhny7l55rfddaxgcaplaw0r2yy-zoxide-0.9.6/bin/zoxide init bash --cmd cd)"

if [[ :$SHELLOPTS: =~ :(vi|emacs): ]]; then
  eval "$(/nix/store/2f27qfzb3b3s09rjfz0g06g03igibnx1-fzf-0.55.0/bin/fzf --bash)"
fi
...
```

So it seams like we need to source after zoxide, I am total noobie in regards to inner mechanics of home-manager and totally open for any suggestions

### System information

```markdown
❯ nix-shell -p nix-info --run "nix-info -m"
[...]

PR #5955, opened 2024-10-11
fzf: fix fzf bash-completion conflict

talhaHavadar:
### Description

Initializing fzf before zoxide is causing fzf completions to fail it seems like either we need to make zoxide initialize earlier than fzf or make fzf later. I just added mkAfter to fzf to test the theory and indeed initializing fzf after zoxide fixes the issue but we need to decide how to make it. I feel like mkAfter may break things, it seems like fzf is intentionally put earlier due to some bindings inside it.

Another alternative solution can be making zoxide earlier which can be seen in this diff: https://github.com/nix-community/home-manager/compare/master...talhaHavadar:home-manager:fix-fzf-bash-completion?expand=1

I have tested both solutions and both works fine and I am able to use ' ** ' trigger of fzf

Longer description is here in #5954 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


rycee:
Unfortunately fzf needs to be early. So I think the only possible solution would be to place zoxide even earlier. Can use `mkOrder 150` to make it come between the bash completion initialization and fzf initialization.

talhaHavadar:
@rycee I just updated the commit accordingly and here is the resulting .bashrc
```
if [[ ! -v BASH_COMPLETION_VERSINFO ]]; then
  . "/nix/store/k0ag90vp8qszsd1y85b0vq8dwvkxzjz9-bash-completion-2.14.0/etc/profile.d/bash_completion.sh"
fi

eval "$(/nix/store/x2v8rifhny7l55rfddaxgcaplaw0r2yy-zoxide-0.9.6/bin/zoxide init bash --cmd cd)"

if [[ :$SHELLOPTS: =~ :(vi|emacs): ]]; then
  eval "$(/nix/store/2f27qfzb3b3s09rjfz0g06g03igibnx1-fzf-0.55.0/bin/fzf --bash)"
fi

if [[ $TERM != "dumb" ]]; then
  eval "$(/home/ubuntu/.nix-profile/bin/starship init bash --print-full-init)"
fi
```
it placed the zoxide correctly and I tested " ** " trigger of fzf is working fine now.
image

rycee:
Thanks! Rebased to master in fe56302339bb28e3471632379d733547caec8103 🙂

PR #5957, opened 2024-10-12
espanso: add sandboxing for systemd service

talal:
### Description

- Add sandboxing for systemd service

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@n8henrie @lucasew @liyangau

### DIFF
--- a/modules/services/espanso.nix
+++ b/modules/services/espanso.nix
@@ -124,6 +124,15 @@ in {
         Type = "exec";
         ExecStart = "${cfg.package}/bin/espanso daemon";
         Restart = "on-failure";
+
+        # Sandboxing.
+        LockPersonality = true;
+        MemoryDenyWriteExecute = true;
+        NoNewPrivileges = true;
+        PrivateUsers = true;
+        RestrictNamespaces = true;
+        SystemCallArchitectures = "native";
+        SystemCallFilter = "@system-service";
       };
       Install = { WantedBy = [ "default.target" ]; };
     };
--- a/tests/modules/services/espanso/basic-configuration.service
+++ b/tests/modules/services/espanso/basic-configuration.service
@@ -3,7 +3,14 @@ WantedBy=default.target
 
 [Service]
 ExecStart=@espanso@/bin/espanso daemon
+LockPersonality=true
+MemoryDenyWriteExecute=true
+NoNewPrivileges=true
+PrivateUsers=true
 Restart=on-failure
+RestrictNamespaces=true
+SystemCallArchitectures=native
+SystemCallFilter=@system-service
 Type=exec
 
 [Unit]


modules/services/espanso.nix
```diff
@@ -122,8 +122,17 @@ in {
       Unit = { Description = "Espanso: cross platform text expander in Rust"; };
       Service = {
         Type = "exec";
```
rycee, line 124: The upstream source doesn't specify `Type` so presumably you could remove this line?

talal: Done. I've also added `RestartSec` as per source.

n8henrie:
Have you tested on Wayland by chance? Might be tricky since it's not working as is on nixos.

@pitkling and I have an ongoing discussion on the best way to manage the capabilities that are required for the evdev backend, in the context of espanso's forking and good faith effort to drop capabilities as early as possible. We have a couple open PRs, and I'm particularly interested in compatibility of these systemd changes with his PR, which I think is what we'll eventually go with for Wayland compatibility.

n8henrie:
Interesting. I haven't been able to get espanso to work on Wayland without `setcap` (on non-nixos) or running as root. Is your config available somewhere?

talal:
Ahh, you're right. I'm sorry, I forgot about the `setcap` since I did that step before I switched to home-manager.

I ran the command as mentioned in https://espanso.org/docs/install/linux#adding-the-required-capabilities

But, please feel free to correct me if I'm wrong, I don't think this is within the scope of home-manager—changes like capabilities are not a part of home-manager, right?

For example, some packages require adding a udev rule and so they end up adding a new file under `/etc/udev/rules.d` but this is not done by home-manager or does home-manager also make such changes? I actually have no idea about this because I manage my udev rules manually on my non-NixOS home-manager setup.

pitkling:
> Ahh, you're right. I'm sorry, I forgot about the `setcap` since I did that step before I switched to home-manager.
> 
> I ran the command as mentioned in https://espanso.org/docs/install/linux#adding-the-required-capabilities
> 
> But, please feel free to correct me if I'm wrong, I don't think this is within the scope of home-manager—changes like capabilities are not a part of home-manager, right? […]

No, you're completely right, the problem @n8henrie is referring to cannot be fixed by Home Manager. But we have some PRs in NixOS (namely [#328890 ](https://github.com/NixOS/nixpkgs/pull/328890) and [#339594 ](https://github.com/NixOS/nixpkgs/pull/339594)) that try to fix the Wayland issue. The `sudo setcap …` command you're referring to is actually not supposed to work on NixOS, since the nix store is usually read-only. I assume you circumvented that manually by remounting it read-write or something like that? In general, NixOS uses a wrapper mechanism (see [`security.wrappers`](https://search.nixos.org/options?channel=24.05&show=security.wrappers&from=0&size=50&sort=relevance&type=packages&query=security.wrappers) to allow elevated binaries (root permissions via `setuid`, linux capabilities via `setcap`, …). Unfortunately this doesn't work for Espanso out of the box due to reasons outlined in [#328890 ](https://github.com/NixOS/nixpkgs/pull/328890).

I assume @n8henrie mostly was curious about the compatibility of your PR with those fixes. Since the Wayland issues are related to capabilities, sandboxing is likely to affect them.

I tried your PR together with my PR [#328890 ](https://github.com/NixOS/nixpkgs/pull/328890) for the Wayland issue and indeed I ran into some problems due to the sandboxing. I did only some very quick testing so far. I expected that something like `NoNewPrivileges = true;` might cause problems (since upstream Espanso is relying on regaining capabilities after dropping them early on, and also `security.wrappers` is doing some magic with capabilities). However, it turns out I could not use **any** of the sandboxing options from your PR. I'm honestly not sure why, since something like `SystemCallArchitectures = "native";` seems pretty unrelated and safe. But then, I don know exactly how the sandboxing options work. I'll investigate further when I have some time in the coming days.

I didn't yet test @n8henrie's PR [#339594 ](https://github.com/NixOS/nixpkgs/pull/339594) together with your PR. It should involve changing the Home Manager service to also run `/run/wrappers/bin/espanso worker` similar to his PR and set up the corresponding wrapper via `security.wrappers` in the NixOS configuration. Would be interesting to see whether that also causes problems together with the sandboxing options.

talal:
I see. Since I'm using home-manager with Fedora (GNOME Wayland), I was able to use the `sudo setcap ...` command along with the sandboxing (as suggested in this PR) and Espanso works fine for me but I understand that this certainly needs further investigation in order for the home-manager service to work on all supported platforms—NixOS and non-NixOS.

n8henrie:
Thanks for the elaboration @pitkling, spot on. Sorry, traveling / on mobile.

pitkling:
@talal I looked further into this on my system, and even with the unmodified espanso-wayland package (version 2.2.1, from nixpkgs-unstable) the sandboxing options cause problems for me. More exactly, I did the following:

* copied the binary from espanso-wayland (`${pkgs.unstable.espanso-wayland}/bin/.espanso-wrapped}`, which is the unmodified, unwrapped espanso binary) to `~/Desktop/espanso`
* gave that file the necessary capabilities via the `sudo setcap …` call from the Espanso documentation
* adapted the user service file generated from your PR to point to the binary at `~/Desktop/espanso` and reloaded it

Starting the espanso user service now fails, with the logs mentioning permissions problems. If I remove the sandboxing options from the user service file and reload the service, everything works fine.

Could you possibly post the output of `systemctl --user status espanso.service` (showing some infos about the service) and

```
journalctl _SYSTEMD_INVOCATION_ID=`systemctl --user show -p InvocationID --value espanso.service`
```

(showing the log output of the last invocation of the espanso user service)? Maybe I can find a hint why the sandboxing works on your system but not on mine.

pitkling:
Thanks @talal, that helped a lot!

Your espanso log indicates that the capabilities you granted via `sudo setcap …` to the espanso binary are not in effect (see the warning starting with the line `[WARN] EVDEV backend is being used, but without enabling linux capabilities.` in the user service's log). In that case, espanso needs some other ways to access `/dev/input/*` and `/dev/uinput` under Wayland. This is usually achieved by running it as root (when it is run as a system service) or by adding your user to a group that has access to those device files. Any chance your user account is part of a group like `input`, granting it access to `/dev/input/*` and `/dev/uinput`? Can you post the output of `groups` and `ls -l /dev/uinput /dev/input/`?

I tested this locally by adding myself to the input group (and temporarily changing the group/access rights of `/dev/uinput`, which is only readable by root on NixOS by default). After that, the espanso user service with your sandboxing options (and an unmodified espanso-wayland binary) worked fine.

Now, if I'm right then it might be difficult to add sandboxing to the service, at least in the case of espanso-wayland. Not only mine but also @n8henrie's [NixOS PR #339594](https://github.com/NixOS/nixpkgs/pull/339594) does not work with your sandboxing options. I don't understand it enough to be sure, but the [documentation of `PrivateUsers`](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#PrivateUsers=) (which seems to be required for basically any sandboxing ability in user services) indicates that when enabled, the user service has pretty much no capabilities. That would explain the behavior. Maybe one can reenable certain capabilities via `CapabilityBoundingSet` and/or `AmbientCapabilities`. I didn't yet have time to play around with that.

talal:
You are right, I forgot about the `input` group. I also use xremap which can be run without sudo by enabling the `uinput` module, adding the required udev rules, and adding user to the `input` group. Kindly see [xremap's README](https://github.com/xremap/xremap?tab=readme-ov-file#running-xremap-without-sudo) for reference, you might get some hints from there.

```
$ groups
talal wheel input

$ lsmod | grep uinput
uinput                 20480  3

$ cat /etc/udev/rules.d/99-input.rules
KERNEL=="uinput", GROUP="input", TAG+="uaccess"
```

pitkling:
Ok, that explains why your espanso service had access to the input devices, thanks! 🙂

Unfortunately it seems sandboxing an espanso-wayland user service via Home Manager won't be possible without extra intervention on the system level (via the NixOS configuration or manual `sudo`-actions on other distributions). So I guess at best sandboxing could be an opt-in with hints in the documentation[^1] about how to enable corresponding access rights at least in NixOS? I guess on non-wayland system one could also think about activating sandboxing by default (although I haven't tested this). Any thoughts, @n8henrie?

Concerning the compatibility of this with the NixOS PRs [#328890](https://github.com/NixOS/nixpkgs/pull/328890) and [#339594](https://github.com/NixOS/nixpkgs/pull/339594): Any solution based on `CAP_DAC_OVERRIDE` (the ["officially" suggested way](https://espanso.org/docs/install/linux/#install-on-wayland) to run Espanso as a non-privileged user on Wayland, used by both PRs) won't be compatible with a sandboxed user service. After some reading about systemd sandboxing and user namespaces it seems that:
* most **user service** sandboxing options [require `PrivateUsers = true`](https://man.archlinux.org/man/systemd.exec.5.en#SANDBOXING)
* for user services, `PrivateUsers = true` creates a user namespace that maps **only** the corresponding user to the parent namespace (see [documentation of `PrivateUsers`](https://man.archlinux.org/man/systemd.exec.5.en#SANDBOXING))
* file related capabilities (like `CAP_DAC_OVERRIDE`) do only work in user namespaces [if the corresponding file's owner and group are mapped to the parent namespace](https://man.archlinux.org/man/user_namespaces.7.en#Operation_of_file-related_capabilities)
* since `/dev/uinput` is owned by root and root is not mapped in the user namespace created by `PrivateUsers`, `CAP_DAC_OVERRIDE` won't have any effect in the user namespace

[^1]: For NixOS systems, enabling sandboxing for an espanso user service would require something like
  `config.hardware.uinput.enable = true;`
  `users.groups.input.members = [ "<username>" ];`
  `users.groups.uinput.members = [ "<username>" ];`

talal:
@rycee Since sandboxing for Espanso is tricky and requires further investigation. I've removed  _non-sandboxing_  changes from this PR and cherry-picked them into a separate PR (https://github.com/nix-community/home-manager/pull/5975) so that they don't get blocked and can be merged independently.

@n8henrie @pitkling feel free to add to this PR or close it if sandboxing doesn't work out.

n8henrie:
@talal I don't have access to close this, but I think it would make sense to close for the time being.

PR #5960, opened 2024-10-13
kitty: allow not setting shell_integration

### LINKED ISSUE #5809: Kitty allow not setting shell_integration=no_rc
### Description

I would appreciate it if there was a way to disable setting `shell_integration = no-rc` in kitty. This is because I don't have home-manager managing my shell config, and so I want to use kitty's built-in shell integration method, but if `no-rc` is always enabled this is impossible. It's possible to work around this by setting `settings.shell_integration` since that makes it appear twice in the config, and it seems like the latter entry overrides the former, but this feels janky.

The solution I've come up with is that manually setting `shellIntegration.mode = "";` should make that not show up at all in the config. I believe this (untested) patch should make it behave as such. I can do some testing and submit this as a PR if desired.
```patch
diff --git a/modules/programs/kitty.nix b/modules/programs/kitty.nix
index 48ec13da..ffc66c5b 100644
--- a/modules/programs/kitty.nix
+++ b/modules/programs/kitty.nix
@@ -144,20 +144,20 @@ in {
     shellIntegration = {
       mode = mkOption {
         type = types.str;
         default = "no-rc";
         example = "no-cursor";
         apply = (o:
-          let
+          optionalString o (let
             modes = splitString " " o;
             filtered = filter (m: m != "no-rc") modes;
-          in concatStringsSep " " (concatLists [ [ "no-rc" ] filtered ]));
+          in concatStringsSep " " (concatLists [ [ "no-rc" ] filtered ])));
         description = ''
           Set the mode of the shell integration. This accepts the same options
           as the `shell_integration` option of Kitty. Note that
-          `no-rc` is always implied. See
+          `no-rc` is always implied if nonempty. See
           <https://sw.kovidgoyal.net/kitty/shell-integration>
           for more details.
         '';
       };
 
       enableBashIntegration = mkEnableOption "Kitty Bash integration"
@@ -200,16 +200,16 @@ in {
                   "${pkgs.kitty-themes}/share/kitty-themes/themes.json"));
             in throwIf (length matching == 0)
             "kitty-themes does not contain a theme named ${cfg.theme}"
             (head matching).file
           }
         '')
-        ''
+        (optionalString cfg.shellIntegration.mode ''
           # Shell integration is sourced and configured manually
           shell_integration ${cfg.shellIntegration.mode}
-        ''
+        '')
         (toKittyConfig cfg.settings)
         (toKittyKeybindings cfg.keybindings)
         (toKittyEnv cfg.environment)
         cfg.extraConfig
       ]);
     } // optionalAttrs pkgs.stdenv.hostPlatform.isLinux {
```

MithicSpirit:
### Description

This allows better support for using kitty's built-in shell integration
method by setting `shellIntegration.mode` to `null`, which makes it not
be set at all in the config file and disables modification to shell
initialization scripts by default. If this is used, the user should set
`settings.shell_integration` instead (if shell integration is desired),
which receives no special handling.

I'm not sure whether the `mode != null` should just be part of the default options or if it should be forced.

Fixes #5809

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@teto


MithicSpirit:
At this rate, maybe I should just add myself as a maintainer.

khaneliman:
> At this rate, maybe I should just add myself as a maintainer.

Thought always crosses my mind lmao

sumnerevans (approved):

khaneliman (approved):

PR #3817, opened 2023-03-29
Revert "programs.neovim: link packpath dir in XDG_DATA_HOME"

ncfavier:
Reverts nix-community/home-manager#3717

PR #5964, opened 2024-10-15
neovim xdg plugins not wrapped

teto:
### Description

2nd attempt at:
- **Reapply "programs.neovim: link packpath dir in XDG_DATA_HOME (#3717)" (#3817)**

HM wraps the neovim executable with its config, which is a pain because nvim
behaves a bit different with `-u`, also it prevents neovim GUIs to load your
config unless you wrap them too (painful).

Requires https://github.com/NixOS/nixpkgs/pull/348740

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


c3n21:
Would it be possible to make this change opt-in?

teto:
> Would it be possible to make this change opt-in?

why ? no. The current way makes no sense IMO, i.e., it just installs "one" neovim, regardless of anything user-specific. It surprises all users and is the source of a lot of confusion.

Note that in the past week, the nixpkgs documentation for the neovim wrapper has improved significantly and creating your custom neovim should be a lot easier. One could even use `NVIM_APPNAME` to keep some imperative config different from the default neovim folders.

teto:
done in https://github.com/nix-community/home-manager/pull/8586 and https://github.com/nix-community/home-manager/pull/8606

PR #5965, opened 2024-10-15
firefox: Avoid unnecessarily overriding package

amarshall:
When `cfg.package` is already wrapped, and wrapped without the `ExtensionSettings` key set, this would always add that key, even if its value was blank. This would result in `cfg.finalPackage` being a functionally-identical, but differently-input-addressed package. This is generally undesirable as it may result in multiple derivations being built, and also if the value of `cfg.package` is expected to be unchanged by the user (e.g. because they want it to be consistent between NixOS and HM configuration).

Add a test to ensure this does not regress in the default case. Only test on newish stateVersion since the logic for `isWrapped` differs on older versions.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee @brckd

### DIFF
--- a/modules/programs/firefox/mkFirefoxModule.nix
+++ b/modules/programs/firefox/mkFirefoxModule.nix
@@ -799,12 +799,13 @@ in {
     finalPackage = wrapPackage cfg.package;
 
     policies = {
-      ExtensionSettings = listToAttrs (map (lang:
-        nameValuePair "langpack-${lang}@firefox.mozilla.org" {
-          installation_mode = "normal_installed";
-          install_url =
-            "https://releases.mozilla.org/pub/firefox/releases/${cfg.package.version}/linux-x86_64/xpi/${lang}.xpi";
-        }) cfg.languagePacks);
+      ExtensionSettings = lib.mkIf (cfg.languagePacks != [ ]) (listToAttrs (map
+        (lang:
+          nameValuePair "langpack-${lang}@firefox.mozilla.org" {
+            installation_mode = "normal_installed";
+            install_url =
+              "https://releases.mozilla.org/pub/firefox/releases/${cfg.package.version}/linux-x86_64/xpi/${lang}.xpi";
+          }) cfg.languagePacks));
     };
   });
 }
--- a/tests/modules/programs/firefox/final-package.nix
+++ b/tests/modules/programs/firefox/final-package.nix
@@ -1,0 +1,25 @@
+modulePath:
+{ config, lib, pkgs, ... }:
+
+let
+
+  cfg = lib.getAttrFromPath modulePath config;
+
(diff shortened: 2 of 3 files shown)


nyabinary:
@rycee bump :p

brckd:
Thanks for your contribution! Sorry I didn't catch it before. I'll review it asap.

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -799,12 +799,13 @@ in {
     finalPackage = wrapPackage cfg.package;
 
     policies = {
-      ExtensionSettings = listToAttrs (map (lang:
-        nameValuePair "langpack-${lang}@firefox.mozilla.org" {
-          installation_mode = "normal_installed";
-          install_url =
-            "https://releases.mozilla.org/pub/firefox/releases/${cfg.package.version}/linux-x86_64/xpi/${lang}.xpi";
-        }) cfg.languagePacks);
+      ExtensionSettings = lib.mkIf (cfg.languagePacks != [ ]) (listToAttrs (map
```
brckd, line 802: Are you sure `lib.mkIf` works inside the `policies` option? The current tests don't seem to test this policy yet, so kindly check whether the `ExtensionSettings` policy will indeed be unset.

tests/modules/programs/firefox/final-package.nix
```diff
@@ -0,0 +1,25 @@
...
+  cfg = lib.getAttrFromPath modulePath config;
+
+  firefoxMockOverlay = import ./setup-firefox-mock-overlay.nix modulePath;
+
+in {
+  imports = [ firefoxMockOverlay ];
+
+  config = lib.mkIf config.test.enableBig
+    (lib.setAttrByPath modulePath { enable = true; } // {
+      home.stateVersion = "19.09";
```
brckd, line 15: Why exactly is it using this older HM version? Does this test the older `isWrapped` logic?

brckd (commented):
Overall a promising optimization! Isn't the package still being wrapped though? Perhaps `wrapPackage` should check whether any policies are set and wrap / not wrap accordingly.

brckd:
Ok thanks, since you mentioned that `package == finalPackage` my concerns should be invalid.

tests/modules/programs/firefox/final-package.nix
brckd: Thanks for clarifying. I was a bit confused because the other modules use `23.05`, but that's an arbitrary choice and shouldn't have any impact.

teto:
@brckd were all your worries adressed ? is this good to go ?

brckd:
Yea, no concerns from my side. I haven't tested the change yet, but @amarshall has. @rycee will need to merge it AFAIK.

brckd (approved):

rycee:

PR #5970, opened 2024-10-17
firefox, thunderbird: conditional search file

khaneliman:
### Description
When using multiple profiles that share the same path we still try to evaluate the file even when search customization isn't enabled. 
Follow up to https://github.com/nix-community/home-manager/pull/5697#issuecomment-2411177495 to only evaluate when the profile wants a search.json.mozlz4 file generated. 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@teto @rycee @kira-bruneau

### DIFF
--- a/modules/programs/firefox/mkFirefoxModule.nix
+++ b/modules/programs/firefox/mkFirefoxModule.nix
@@ -777,11 +777,12 @@ in {
           force = profile.containersForce;
         };
 
-      "${profilesPath}/${profile.path}/search.json.mozlz4" = {
-        enable = profile.search.enable;
-        force = profile.search.force;
-        source = profile.search.file;
-      };
+      "${profilesPath}/${profile.path}/search.json.mozlz4" =
+        mkIf (profile.search.enable) {
+          enable = profile.search.enable;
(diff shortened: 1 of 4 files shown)


khaneliman:
I can try and add a test for this scenario, too.

khaneliman:
Alright, added a test to firefox for the shared path. Verified it throws error without the change and with the change the test passes.

khaneliman:
Looks like unrelated test failure due to recent neovim PR 
>neovim-plugin-config: FAILED
>Expected /build/tmp.MOi8v4T8TN to contain HM_PLUGINS_CONFIG but it did not.

khaneliman:
Rebased and should be good now.

rycee:

teto:
thanks for the fix.

> Looks like unrelated test failure due to recent neovim https://github.com/NixOS/nixpkgs/pull/344541

It's annoying that the tests are not run against the flake nixpkgs, it means we cant provide lockstep updates and any nixos-unstable bump can break HM's CI/configs

PR #5971, opened 2024-10-17
direnv/nushell: simplify, work around nushell/nushell#14112

K900:
nushell 0.99 does not like early returns in hooks. So, what if we just didn't? Rewrite the entire hook to work as one single pipeline.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/direnv.nix
+++ b/modules/programs/direnv.nix
@@ -162,28 +162,20 @@ in {
           $env.config.hooks.pre_prompt?
           | default []
           | append {||
-              let direnv = (
-                  ${direnvWrapped}
-                  | from json --strict
-                  | default {}
-              )
-              if ($direnv | is-empty) {
-                  return
-              }
-              $direnv
+              ${direnvWrapped}
+              | from json --strict
+              | default {}
               | items {|key, value|
-                  {
-                      key: $key
-                      value: (do (
-                          $env.ENV_CONVERSIONS?
-                          | default {}
-                          | get -i $key
-                          | get -i from_string
-                          | default {|x| $x}
-                      ) $value)
-                  }
+                  let value = do (
+                      $env.ENV_CONVERSIONS?
+                      | default {}
+                      | get -i $key
+                      | get -i from_string
+                      | default {|x| $x}
+                  ) $value
+                  return [ $key $value ]
               }
-              | transpose -ird
+              | into record
               | load-env
           }
       )
--- a/modules/programs/direnv.nix
+++ b/modules/programs/direnv.nix
@@ -130,39 +130,16 @@ in {
         ${getExe cfg.package} hook fish | source
       '');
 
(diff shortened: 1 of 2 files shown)


modules/programs/direnv.nix
```diff
@@ -162,28 +162,20 @@ in {
...
-                      ) $value)
-                  }
+                let value = do (
+                  $env.ENV_CONVERSIONS?
+                  | default {}
+                  | get -i $key
+                  | get -i from_string
+                  | default {|x| $x}
+                ) $value
+                return [ $key $value ]
```
JoaquinTrinanes, line 176: I was about to ask the reason for the explicit `return`, but I guess explicitness doesn't hurt :)

modules/programs/direnv.nix
```diff
@@ -162,28 +162,20 @@ in {
...
-                          | default {|x| $x}
-                      ) $value)
-                  }
+                let value = do (
+                  $env.ENV_CONVERSIONS?
+                  | default {}
+                  | get -i $key
+                  | get -i from_string
+                  | default {|x| $x}
+                ) $value
```
JoaquinTrinanes, line 175: Indentation goes from 4 spaces to 2 spaces over here

```suggestion
                  let value = do (
                      $env.ENV_CONVERSIONS?
                      | default {}
                      | get -i $key
                      | get -i from_string
                      | default {|x| $x}
                ) $value
```

JoaquinTrinanes (approved):
LGTM, thanks! I wish there was a simpler way of handling the empty-record-transposes-to-empty-array special case, but meh.

Unrelated-ish, but as you're the author of the fix: the issue mentioned above (https://github.com/nushell/nushell/issues/13868) is closed. Can the wrapper be removed? I'm not sure how to reproduce it and never encountered it in the wild 🤔

modules/programs/direnv.nix
khaneliman: I would always prefer explicit returns. But, I'm not familiar with nu standards.

K900: I don't have a strong preference, it just had a bit too many sigils in a row for my taste without it.

K900:
> LGTM, thanks! I wish there was a simpler way of handling the empty-record-transposes-to-empty-array special case, but meh.
> 
> Unrelated-ish, but as you're the author of the fix: the issue mentioned above (https://github.com/nushell/nushell/issues/13868) is closed. Can the wrapper be removed? I'm not sure how to reproduce it and never encountered it in the wild 🤔 

Probably can, yes. I'll try that in a few hours.

JoaquinTrinanes (approved):

K900:
neovim-plugin-config test also fails on master :(

rycee:

Yeah, unfortunately the neovim test started failing. I'm guessing due to some change in Nixpkgs. Unfortunately I don't know enough about neovim to resolve it myself.

PR #5975, opened 2024-10-18
espanso: use `launcher` command on Linux

talal:
### Description

- Use `launcher` command instead of `daemon`. Additionally we remove `Type` and add `RestartSec` as defined in the source: https://github.com/espanso/espanso/blob/b421bcf73fa13506938d62425459d6c16c6a8d0a/espanso/src/res/linux/systemd.service#L5-L7C1

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@n8henrie @lucasew @liyangau


rycee:

PR #5979, opened 2024-10-20
eww: added improvements

Eveeifyeve:
### Description

- Added systemd option to run eww as a service inspired by `programs.waybar.systemd`
  - Removes the need of this pr: #5136
- Deprecated `configDir`.
- Added string config option for both yuck and scss implements this propperly #5585
  - Made all options require and `<yuck, configDir, scss>.enable` to work.
  - Made assertions to avoid overriding issues with scss x yuck and config dir.
- Made `yuck`, `scss` optional. which does this: #4595

closes https://github.com/nix-community/home-manager/pull/5136

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@mainrs


Eveeifyeve:
This closes prs #5585 & #5136

Eveeifyeve:
Also @mainrs, please provide an actual email not a GitHub nore-reply because I can't contact you sept for here about this.

Eveeifyeve:
@rycee maintainer maybe abandoned hm?

modules/programs/eww.nix
```diff
@@ -45,26 +107,73 @@ in {
...
+        }
+        {
+          assertion = !(cfg.yuck.enable && cfg.configDir.enable);
+          inherit message;
+        }
+      ];
+
+      home.packages = [ cfg.package ];
+
+      programs.bash.initExtra = let ewwCmd = "${cfg.package}/bin/eww";
```
mainrs, line 127: Why are you re-defining the command here? It's already in-scope on line 9.

mainrs (changes requested):
Rest is looking good to me! Thanks for bringing all the PRs together! And sorry for everybody who was waiting on them. I'll make sure to communicate better next time!

modules/programs/eww.nix
Eveeifyeve: Ops I forgot to remove that.

mainrs (approved):

mainrs:
@Eveeifyeve I just had this thought: wouldn't it also be possible to simply check for the existence of the values in yuck, scss and configDir without the enable flags? It would de-clutter user configurations. I also haven't found any configuration definition in home-manager that uses the enable pattern for exclusive sets (I have to admit, I did a fast, rudimentary search only).

modules/programs/eww.nix
```diff
@@ -23,12 +21,74 @@ in {
       '';
     };
 
-    configDir = mkOption {
-      type = types.path;
-      example = literalExpression "./eww-config-dir";
+    yuck = {
+      enable = mkEnableOption ""
+      mkOption {
```
rycee, line 26: This looks like a mistake to me. Perhaps you mean to add a semicolon on the previous line and define a named option on this line?

modules/programs/eww.nix
```diff
@@ -23,12 +21,74 @@ in {
...
+        type = types.nullOr types.lines;
+        default = null;
```
rycee, line 28:
```suggestion
        type = types.lines;
        default = "";
```

Same applies for the `scss` option.

tests/modules/programs/eww/basic-configuration.nix
```diff
@@ -0,0 +1,38 @@
+{ config, pkgs, ... }:
+
+{
+  config = {
+    programs.eww = {
+      enable = true;
+      systemd.enable = true;
+      configYuck = ''
```
rycee, line 8: I guess the `configYuck` and `configScss` are mistakes? Maybe need a separate yuck and scss test, since they are mutually exclusive?

modules/programs/eww.nix
```diff
@@ -23,12 +21,74 @@ in {
...
+        description = ''
+          The content that gets symlinked to
+          {file} `$XDG_CONFIG_HOME/eww/eww.yuck`.
+        '';
+      };
+    };
+
+    scss = {
+      enable = mkEnableOption "Enables string based scss config";
+      config = mkOption {
```
rycee, line 52: I would suggest `extraConfig` as name instead. The problem with `config` is that it is a very overloaded term and the convention is to use the `extraConfig` name (even though there are no other options).

modules/programs/eww.nix
```diff
@@ -23,12 +21,74 @@ in {
...
+          }
+        '';
+        description = ''
+          The directory that gets symlinked to
+          {file} `$XDG_CONFIG_HOME/eww/eww.scss`.
+        '';
+      };
+    };
+
+    configDir = {
```
rycee, line 67: I'm not fond of this option and would suggest removing it. if a user really wants to do this I think they can set `xdg.configFile.eww.source` directly…

rycee: Hehe, just noticed this option actually being present before. In that case I guess we have to keep it 😀

modules/programs/eww.nix
rycee: Actually, looking at it closer I agree with @mainrs, just make single `scssConfig` and `yuckConfig` options of type `lines`, default it to `""` and skip creating the file if the string is empty.

tests/modules/programs/eww/basic-configuration.nix
rycee: Nevermind, I was a bit confused. See my other suggestion about having `yuckConfig` and `scssConfig`.

modules/programs/eww.nix
Eveeifyeve: I was recommend by someone named `just_midi` in discord to do this as they said they were maintainer of multiple modules in home manager, can't find their gh username unfortantly

Eveeifyeve: > Actually, looking at it closer I agree with @mainrs, just make single `scssConfig` and `yuckConfig` options of type `lines`, default it to `""` and skip creating the file if the string is empty.

wouldn't null just be better then? if the type is not null then create the file else not.

modules/programs/eww.nix
Eveeifyeve: I have replied but I think I will keep that because of this comment: https://github.com/nix-community/home-manager/pull/5979?notification_referrer_id=NT_kwDOBUkEqrQxMjk1NTI0MzIyMDo4ODY3MTQwMg#discussion_r1826307253

modules/programs/eww.nix
Eveeifyeve: Ops a classic mistake that got missed.

rycee:
Thanks! Looks good in general 🙂

About the

> BREAKING CHANGE configDir isprograms.eww.configDir = ./path to programs.eww.configDir.config = ./path .

I would suggest keeping the old option for backwards compatibility, just make it nullable with `null` as default value. Can add a sentence in the description discouraging use of `configDir`. Perhaps even add a warning. Something like

``` nix
warnings = optional (cfg.configDir != null) ''
  The option programs.eww.configDir is deprecated.
  Use programs.eww.yuck or programs.eww.scss instead.'';
```

perhaps.

Eveeifyeve:
> Thanks! Looks good in general 🙂
> 
> About the
> 
> > BREAKING CHANGE configDir isprograms.eww.configDir = ./path to programs.eww.configDir.config = ./path .
> 
> I would suggest keeping the old option for backwards compatibility, just make it nullable with `null` as default value. Can add a sentence in the description discouraging use of `configDir`. Perhaps even add a warning. Something like
> 
> ```nix
> warnings = optional (cfg.configDir != null) ''
>   The option programs.eww.configDir is deprecated.
>   Use programs.eww.yuck or programs.eww.scss instead.'';
> ```
> 
> perhaps.

I mean I am still going to keep config dir because backwards compatibility but the problem is in the future say if someone wants to pull a random eww config off of the internet that uses the `(import ./folder/eww.yuck)` method is going to become an issue.

https://github.com/HynDuf/dotfiles/tree/9be198a9ab5307ebd55633fe248607a9956bc074/.config/eww

modules/programs/eww.nix
Eveeifyeve: Null is better to avoid overriding

Eveeifyeve:
@rycee & @mainrs Ready to review again!

Eveeifyeve:
Oh tests failed.

modules/programs/eww.nix
```diff
@@ -45,26 +96,60 @@ in {
...
-        eval "$(${ewwCmd} shell-completions --shell fish)"
-      end
-    '';
-  };
+  config = mkIf cfg.enable (mkMerge [
+    {
+      assertions = [
+          {
+            assertion = !(cfg.scssConfig == null && cfg.configDir == null) or !(cfg.yuckConfig == null && cfg.configDir == null);
+            message = "You can have either yuck & scss or configDir enabled"
```
dixslyf, line 104:
```suggestion
            assertion = !(cfg.scssConfig == null && cfg.configDir == null) || !(cfg.yuckConfig == null && cfg.configDir == null);
            message = "You can have either yuck & scss or configDir enabled";
```
Seems like a language mishap.

Eveeifyeve: Yeah I am aware, but I have been so busy focusing other things than this pr.

mainrs (approved):
LGTM

modules/programs/eww.nix
```diff
@@ -45,26 +96,60 @@ in {
...
+          Description = "ElKowars wacky widgets daemon";
+          Documentation = "https://elkowar.github.io/eww/";
+          After = [ "graphical-session.target" ];
+          PartOf = [ "graphical-session.target" ];
+        };
+
+        Service = {
+          ExecStart = "${cfg.package} daemon --no-daemonize";
+          ExecStop = "${cfg.package} kill";
+          ExecReload = "${cfg.package} reload";
```
dixslyf, line 148:
```suggestion
          ExecStart = "${ewwCmd} daemon --no-daemonize";
          ExecStop = "${ewwCmd} kill";
          ExecReload = "${ewwCmd} reload";
```
`${cfg.package}` is just the Nix store path for Eww, not the `eww` executable.

tests/modules/programs/eww/basic-configuration.nix
dixslyf, line 8:
```suggestion
      yuckConfig = ''
```

tests/modules/programs/eww/basic-configuration.nix
```diff
@@ -0,0 +1,38 @@
...
+                                 :width "90%"
+                                 :height "30px"
+                                 :anchor "top center")
+             :stacking "fg"
+             :reserve (struts :distance "40px" :side "top")
+             :windowtype "dock"
+             :wm-ignore false
+          "example content")
+      '';
+      configScss = ''
```
dixslyf, line 22:
```suggestion
      scssConfig = ''
```

tests/modules/programs/eww/basic-configuration.service
```diff
@@ -0,0 +1,13 @@
+[Install]
+WantedBy=graphical-session.target
+
+[Service]
+ExecReload=@eww@ reload
+ExecStart=@eww@ daemon --no-daemonize
+ExecStop=@eww@ kill
```
dixslyf, line 7:
```suggestion
ExecReload=@eww@/bin/eww reload
ExecStart=@eww@/bin/eww daemon --no-daemonize
ExecStop=@eww@/bin/eww kill
```
`@eww@` is a stub for the Nix store path for Eww if I understand correctly.

dixslyf (changes requested):
There are some more syntax and logic errors. Otherwise, the rest looks good to me.

modules/programs/eww.nix
```diff
@@ -45,26 +96,60 @@ in {
...
+
+      programs.fish.interactiveShellInit = mkIf cfg.enableFishIntegration ''
+        if test "$TERM" != "dumb"
+          eval "$(${ewwCmd} shell-completions --shell fish)"
+        end
+      '';
+    }
+
+    (mkIf cfg.configDir
+      == null { xdg.configFile."eww".source = cfg.configDir; })
```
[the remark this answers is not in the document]
mainrs: I missed this one. I didn't look too close at the syntax because the PR had tests. I just assumed they would fail in the case of errors. Wasn't that the case?

tests/modules/programs/eww/basic-configuration.service
mainrs: That I didn't know, thank you for having a look at it. I learned something new too :)

modules/programs/eww.nix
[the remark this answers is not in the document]
dixslyf: To be honest, I only noticed the syntax error after looking at the CI results.

modules/programs/eww.nix
```diff
@@ -45,26 +96,60 @@ in {
...
-      if test "$TERM" != "dumb"
-        eval "$(${ewwCmd} shell-completions --shell fish)"
-      end
-    '';
-  };
+  config = mkIf cfg.enable (mkMerge [
+    {
+      assertions = [
+          {
+            assertion = !(cfg.scssConfig == null && cfg.configDir == null) || !(cfg.yuckConfig == null && cfg.configDir == null);
```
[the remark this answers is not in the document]
Eveeifyeve: I feel like you made a mistake I. This suggestion as it's only ment to be made if config dir is not null and the string based config is not null, because I want to avoid an overriding issue.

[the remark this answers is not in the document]
dixslyf: Just to make sure we're on the same page, the intention of the assertion is to keep `configDir` _mutually exclusive_ with `scssConfig` and `yuckConfig`, right? I.e., if the user specifies `configDir`, they should _not_ be able to specify `scssConfig` and/or `yuckConfig`, and vice versa.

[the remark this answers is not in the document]
Eveeifyeve: Yes that is correct.

modules/programs/eww.nix
Eveeifyeve: ops my bad

modules/programs/eww.nix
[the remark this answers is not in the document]
Eveeifyeve: it should be `cfg.configDir != null || (cfg.scssConfig != null && cfg.yuckConfig != null);`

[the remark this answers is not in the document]
Eveeifyeve: it should be this:
```suggestion
            assertion = cfg.configDir != null && (cfg.scssConfig != null || cfg.yuckConfig != null);
```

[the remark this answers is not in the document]
dixslyf: > should it be this?

I don't think so. That assertion will always fail whenever `configDir` is not set (`configDir = null`), which is not right.

I suspect there's some confusion about when an assertion is triggered — an assertion fails when the boolean condition evaluates to `false`. Conversely, there's no error when the condition evaluates to `true`. With that in mind, I'm pretty sure my suggested change (`cfg.configDir == null || (cfg.scssConfig == null && cfg.yuckConfig == null)`) is correct:
- If `configDir` isn't set (i.e., `configDir = null`), the whole expression is `true`. This is good because, when `configDir = null`, `scssConfig` and `yuckConfig` should be allowed to be anything (including both being set to `null`).
- Similarly, when both `scssConfig` and `yuckConfig` are `null` (both are unspecified), the whole expression is `true`. This is also good because it doesn't matter what `configDir` is when they're both `null`.
- Now, if say `configDir` and `scssConfig` are *not* `null`, then the assertion fails since it evaluates to `false`. The same thing happens if you had specified `yuckConfig` instead of `scssConfig`, or even specified both of them.
- When all three (`configDir`, `yuckConfig` and `scssConfig`) are `null`, the expression still evaluates to `true`. This case might not be useful for the user, but it should still be valid.

I'll test this on my configuration to make sure when I find the time.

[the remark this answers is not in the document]
Eveeifyeve: I would reccomend seeing this: https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/misc/assertions.nix

tests/modules/programs/eww/basic-configuration.service
Eveeifyeve: Same.

modules/programs/eww.nix
```diff
@@ -45,26 +96,59 @@ in {
...
-      if test "$TERM" != "dumb"
-        eval "$(${ewwCmd} shell-completions --shell fish)"
-      end
-    '';
-  };
+  config = mkIf cfg.enable (mkMerge [
+    {
+      assertions = [
+          {
+            assertion = cfg.configDir != null && (cfg.scssConfig != null || cfg.yuckConfig != null);
```
dixslyf, line 103: Gonna resume the [conversation](https://github.com/nix-community/home-manager/pull/5979#discussion_r1937769271) here.

> Well based off https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/misc/assertions.nix it seems to me if the `assetion == true` then it works but if it's `false` then it doesn't.

Yes, that is exactly what I said.

Regardless, I've managed to test it on my own config in [this branch](https://github.com/dixslyf/dotfiles/commits/hm-eww-improvements-test/). The latest commits in that branch each tests one combination of setting `configDir`, `yuckConfig` and `scssConfig`. If you look at the CI results and logs for each commit, you'll see that your assertion does the exact opposite of what we want — it does not trigger when `configDir` is set alongside `yuckConfig` and/or `scssConfig` even though it should (which would then lead to an error when activating the HM configuration due to the conflict), but triggers in the other possible combinations where it shouldn't.

So, we just need to negate the whole current condition to get the right result:
```suggestion
            assertion = !(cfg.configDir != null && (cfg.scssConfig != null || cfg.yuckConfig != null));
```
Now, if you do some boolean algebra, the negation turns out to be equivalent to what I had suggested:
```suggestion
            assertion = cfg.configDir == null || (cfg.scssConfig == null && cfg.yuckConfig == null);
```

dixslyf (changes requested):
CI is failing due to the code not being properly formatted. Could you run the [`format`](https://github.com/nix-community/home-manager/blob/801ddd8693481866c2cfb1efd44ddbae778ea572/format) script located at the root of the HM repo (just run it without any arguments)?

modules/programs/eww.nix
```diff
@@ -45,26 +96,59 @@ in {
...
+        if [[ $TERM != "dumb" ]]; then
+          eval "$(${ewwCmd} shell-completions --shell zsh)"
+        fi
+      '';
+
+      programs.fish.interactiveShellInit = mkIf cfg.enableFishIntegration ''
+        if test "$TERM" != "dumb"
+          eval "$(${ewwCmd} shell-completions --shell fish)"
+        end
+      '';
```
khaneliman, line 126:
```suggestion
```
Looks like shell completions are installed in the package upstream.

modules/programs/eww.nix
```diff
@@ -23,15 +21,68 @@ in {
...
     configDir = mkOption {
-      type = types.path;
+      type = types.nullOr types.path;
+      default = null;
       example = literalExpression "./eww-config-dir";
       description = ''
         The directory that gets symlinked to
         {file}`$XDG_CONFIG_HOME/eww`.
       '';
     };
```
khaneliman, line 69: Would rather just remove this option and use a single option that supports paths and lines.

Eveeifyeve: @rycee your thoughts?

eveeifyeve: > @mainrs your thoughts?

CC @mainrs.

khaneliman: Can probably just refactor that later since this is an existing option and name can't be reused.

eveeifyeve: Okay Well I will make an assertion. But since my grammar is not great what would be a great message to notify the deprecated option?

eveeifyeve: Removed option!

eveeifyeve:
Now ready to review! 

Sorry for the wait.

mdaniels5757 (changes requested):
Looks like this broke some tests.

eveeifyeve:
> Looks like this broke some tests.

This issue should be fixed, however I had to remove the news in this process 😞

eveeifyeve:
@mdaniels5757 is there anyway to make the tests ignore spacing or tabs in the test? Because this is really annoying when it looks exactly right and it doesn't work because it's so "picky" about the spacing/tab usage.

mdaniels5757:
Not that I'm aware of.

rycee:
I was having a look at this and it was a bit odd in places. I don't understand how it works for you since, e.g., the service file refers to the package dir, not the executable inside the package.

I attempted to clean things up, please have a look:

``` patch
diff --git a/modules/programs/eww.nix b/modules/programs/eww.nix
index 9d46f24b0..59f6a3259 100644
--- a/modules/programs/eww.nix
+++ b/modules/programs/eww.nix
@@ -54,7 +54,7 @@ in
       '';
       description = ''
         The content that gets symlinked to
-        {file} `$XDG_CONFIG_HOME/eww/eww.yuck`.
+        {file}`$XDG_CONFIG_HOME/eww/eww.yuck`.
       '';
     };
 
@@ -68,35 +68,34 @@ in
       '';
       description = ''
         The directory that gets symlinked to
-        {file} `$XDG_CONFIG_HOME/eww/eww.scss`.
+        {file}`$XDG_CONFIG_HOME/eww/eww.scss`.
       '';
     };
 
     systemd = {
       enable = lib.mkEnableOption "Launches Eww Daemon";
+
       target = lib.mkOption {
         type = lib.types.str;
-        default = "graphical-session.target";
-        example = "sway-session.target";
+        default = config.wayland.systemd.target;
+        defaultText = lib.literalExpression "config.wayland.systemd.target";
         description = ''
-          The systemd target that will automatically start the Eww service.
-
-          When setting this value to `"sway-session.target"`,
-          make sure to also enable {option}`wayland.windowManager.sway.systemd.enable`,
-          otherwise the service may never be started.
+          Systemd target to bind to.
         '';
       };
     };
   };
 
-  config = mkIf cfg.enable (mkMerge [
-    {
-    }
-    (mkIf (cfg.yuckConfig != null) { xdg.configFile."eww/eww.yuck".text = cfg.yuckConfig; })
-    (mkIf (cfg.scssConfig != null) { xdg.configFile."eww/eww.scss".text = cfg.scssConfig; })
-    (mkIf cfg.systemd.enable {
-      systemd.user.services.eww = {
+  config = mkIf cfg.enable {
+
+    xdg.configFile = {
+      "eww/eww.yuck" = mkIf (cfg.yuckConfig != null) { text = cfg.yuckConfig; };
+      "eww/eww.scss" = mkIf (cfg.scssConfig != null) { text = cfg.scssConfig; };
+    };
+
+    systemd = mkIf cfg.systemd.enable {
+      user.services.eww = {
         Unit = {
           Description = "ElKowars wacky widgets daemon";
           Documentation = "https://elkowar.github.io/eww/";
@@ -104,16 +103,20 @@ in
         };
 
-        Service = {
-          ExecStart = "${cfg.package} daemon --no-daemonize";
-          ExecStop = "${cfg.package} kill";
-          ExecReload = "${cfg.package} reload";
-        };
+        Service =
+          let
+            exe = lib.getExe cfg.package;
+          in
+          {
+            ExecStart = "${exe} daemon --no-daemonize";
+            ExecStop = "${exe} kill";
+            ExecReload = "${exe} reload";
+          };
 
         Install = {
           WantedBy = [ cfg.systemd.target ];
         };
       };
-    })
-  ]);
+    };
+  };
 }
diff --git a/tests/modules/programs/eww/basic-configuration.nix b/tests/modules/programs/eww/basic-configuration.nix
index 93f4450f6..f8d5e128b 100644
--- a/tests/modules/programs/eww/basic-configuration.nix
+++ b/tests/modules/programs/eww/basic-configuration.nix
@@ -1,86 +1,84 @@
 {
-  config = {
-    programs.eww = {
-      enable = true;
-      systemd.enable = true;
-      yuckConfig = ''
-        (defwindow powermenu
-               :monitor 0
-               :geometry (geometry
-                   :anchor "center"
-               )
-               ; Widgets
-               (box
-                   :spacing 5
-                   :class "powermenu"
-                   :space-evenly true
-                   :orientation "vertical"
-                   :halign "left"
-                   :valign "center"
+  programs.eww = {
+    enable = true;
+    systemd.enable = true;
+    yuckConfig = ''
+      (defwindow powermenu
+          :monitor 0
+          :geometry (geometry
+              :anchor "center"
+          )
+          ; Widgets
+          (box
+              :spacing 5
+              :class "powermenu"
+              :space-evenly true
+              :orientation "vertical"
+              :halign "left"
+              :valign "center"
 
-                   ; Contents
-                   (button
-                       :onclick "eww open confirm-command --arg action=poweroff --arg command=poweroff"
-                       "Poweroff"
-                   )
-                   (button
-                       :onclick "eww open confirm-command --arg action=reboot --arg command=reboot"
-                       "Reboot"
-                   )
-                   (button
-                       :onclick "eww close powermenu"
-                       "Cancel"
-                   )
-               )
-           )
+              ; Contents
+              (button
+                  :onclick "eww open confirm-command --arg action=poweroff --arg command=poweroff"
+                  "Poweroff"
+              )
+              (button
+                  :onclick "eww open confirm-command --arg action=reboot --arg command=reboot"
+                  "Reboot"
+              )
+              (button
+                  :onclick "eww close powermenu"
+                  "Cancel"
+              )
+          )
+      )
 
-           (defwindow confirm-command [action command]
-               :monitor 0
-               :geometry (geometry
-                   :anchor "center"
-                   :width "13%"
-                   :height "8%"
-               )
-               ; Widgets
-               (box
-                   :class "powermenu"
-                   :orientation "vertical"
-                   :halign "center"
-                   :valign "center"
-                   "Are you sure you want to ''${action}"
-                   (box
-                       :orientation "horizontal"
-                       :spacing 10
-                       (button
-                           :onclick "eww close confirm-command"
-                           "No"
-                       )
-                       (button
-                           :onclick command
-                           "Yes"
-                       )
-                   )
-               )
-           )
-      '';
-      scssConfig = ''
-                .powermenu {
-          padding: 10px 10px;
-          border-radius: 10px;
-        }
-      '';
-    };
+      (defwindow confirm-command [action command]
+          :monitor 0
+          :geometry (geometry
+              :anchor "center"
+              :width "13%"
+              :height "8%"
+          )
+          ; Widgets
+          (box
+              :class "powermenu"
+              :orientation "vertical"
+              :halign "center"
+              :valign "center"
+              "Are you sure you want to ''${action}"
+              (box
+                  :orientation "horizontal"
+                  :spacing 10
+                  (button
+                      :onclick "eww close confirm-command"
+                      "No"
+                  )
+                  (button
+                      :onclick command
+                      "Yes"
+                  )
+              )
+          )
+      )
+    '';
+    scssConfig = ''
+      .powermenu {
+        padding: 10px 10px;
+        border-radius: 10px;
+      }
+    '';
+  };
 
-    nmt.script = ''
-      yuckDir=home-files/.config/eww
-      serviceFile=$(normalizeStorePaths home-files/.config/systemd/user/eww.service)
+  nmt.script = ''
+    yuckDir=home-files/.config/eww
+    serviceFile=$(normalizeStorePaths home-files/.config/systemd/user/eww.service)
 
-      assertFileExists $yuckDir/eww.yuck
-      assertFileExists $yuckDir/eww.scss
+    assertFileExists $yuckDir/eww.yuck
+    assertFileExists $yuckDir/eww.scss
 
-      assertFileContent $yuckDir/eww.yuck ${./expected.yuck}
-      assertFileContent $yuckDir/eww.scss ${./expected.scss}
-      assertFileContent "$serviceFile" ${./expected.service}
-    '';
-  };
+    assertFileContent $yuckDir/eww.yuck ${./expected.yuck}
+    assertFileContent $yuckDir/eww.scss ${./expected.scss}
+    assertFileContent "$serviceFile" ${./expected.service}
+  '';
 }
diff --git a/tests/modules/programs/eww/empty-settings.nix b/tests/modules/programs/eww/empty-settings.nix
index 8fcd252b1..45677d9b8 100644
--- a/tests/modules/programs/eww/empty-settings.nix
+++ b/tests/modules/programs/eww/empty-settings.nix
@@ -1,15 +1,13 @@
 {
-  config = {
-    programs.eww = {
-      enable = true;
-    };
-
-    nmt.script = ''
-      yuckDir=home-files/.config/eww
-      serviceFile=$(normalizeStorePaths home-files/.config/systemd/user/eww.service)
-      assertPathNotExists $yuckDir/eww.yuck
-      assertPathNotExists $yuckDir/eww.scss
-      assertPathNotExists $serviceFile
-    '';
+  programs.eww = {
+    enable = true;
   };
+
+  nmt.script = ''
+    yuckDir=home-files/.config/eww
+    serviceFile=home-files/.config/systemd/user/eww.service
+    assertPathNotExists $yuckDir/eww.yuck
+    assertPathNotExists $yuckDir/eww.scss
+    assertPathNotExists $serviceFile
+  '';
 }
diff --git a/tests/modules/programs/eww/expected.yuck b/tests/modules/programs/eww/expected.yuck
index c3a88c6f4..f7fd7d5c0 100644
--- a/tests/modules/programs/eww/expected.yuck
+++ b/tests/modules/programs/eww/expected.yuck
@@ -41,7 +41,7 @@
         :orientation "vertical"
         :halign "center"
         :valign "center"
-        "Are you sure you want to ''${action}"
+        "Are you sure you want to ${action}"
         (box
             :orientation "horizontal"
             :spacing 10
```

eveeifyeve:
> I was having a look at this and it was a bit odd in places. I don't understand how it works for you since, e.g., the service file refers to the package dir, not the executable inside the package.
> 
> I attempted to clean things up, please have a look:
> 
> ```diff
> diff --git a/modules/programs/eww.nix b/modules/programs/eww.nix
> index 9d46f24b0..59f6a3259 100644
> --- a/modules/programs/eww.nix
> +++ b/modules/programs/eww.nix
> @@ -54,7 +54,7 @@ in
>        '';
>        description = ''
>          The content that gets symlinked to
> -        {file} `$XDG_CONFIG_HOME/eww/eww.yuck`.
> +        {file}`$XDG_CONFIG_HOME/eww/eww.yuck`.
>        '';
>      };
>  
> @@ -68,35 +68,34 @@ in
>        '';
>        description = ''
>          The directory that gets symlinked to
> -        {file} `$XDG_CONFIG_HOME/eww/eww.scss`.
> +        {file}`$XDG_CONFIG_HOME/eww/eww.scss`.
>        '';
>      };
>  
>      systemd = {
>        enable = lib.mkEnableOption "Launches Eww Daemon";
> +
>        target = lib.mkOption {
>          type = lib.types.str;
> -        default = "graphical-session.target";
> -        example = "sway-session.target";
> +        default = config.wayland.systemd.target;
> +        defaultText = lib.literalExpression "config.wayland.systemd.target";
>          description = ''
> -          The systemd target that will automatically start the Eww service.
> -
> -          When setting this value to `"sway-session.target"`,
> -          make sure to also enable {option}`wayland.windowManager.sway.systemd.enable`,
> -          otherwise the service may never be started.
> +          Systemd target to bind to.
>          '';
>        };
>      };
>    };
>  
> -  config = mkIf cfg.enable (mkMerge [
> -    {
> -      home.packages = [ cfg.package ];
> -    }
> -    (mkIf (cfg.yuckConfig != null) { xdg.configFile."eww/eww.yuck".text = cfg.yuckConfig; })
> -    (mkIf (cfg.scssConfig != null) { xdg.configFile."eww/eww.scss".text = cfg.scssConfig; })
> -    (mkIf cfg.systemd.enable {
> -      systemd.user.services.eww = {
> +  config = mkIf cfg.enable {
> +    home.packages = [ cfg.package ];
> +
> +    xdg.configFile = {
> +      "eww/eww.yuck" = mkIf (cfg.yuckConfig != null) { text = cfg.yuckConfig; };
> +      "eww/eww.scss" = mkIf (cfg.scssConfig != null) { text = cfg.scssConfig; };
> +    };
> +
> +    systemd = mkIf cfg.systemd.enable {
> +      user.services.eww = {
>          Unit = {
>            Description = "ElKowars wacky widgets daemon";
>            Documentation = "https://elkowar.github.io/eww/";
> @@ -104,16 +103,20 @@ in
>            PartOf = [ "graphical-session.target" ];
>          };
>  
> -        Service = {
> -          ExecStart = "${cfg.package} daemon --no-daemonize";
> -          ExecStop = "${cfg.package} kill";
> -          ExecReload = "${cfg.package} reload";
> -        };
> +        Service =
> +          let
> +            exe = lib.getExe cfg.package;
> +          in
> +          {
> +            ExecStart = "${exe} daemon --no-daemonize";
> +            ExecStop = "${exe} kill";
> +            ExecReload = "${exe} reload";
> +          };
>  
>          Install = {
>            WantedBy = [ cfg.systemd.target ];
>          };
>        };
> -    })
> -  ]);
> +    };
> +  };
>  }
> diff --git a/tests/modules/programs/eww/basic-configuration.nix b/tests/modules/programs/eww/basic-configuration.nix
> index 93f4450f6..f8d5e128b 100644
> --- a/tests/modules/programs/eww/basic-configuration.nix
> +++ b/tests/modules/programs/eww/basic-configuration.nix
> @@ -1,86 +1,84 @@
>  {
> -  config = {
> -    programs.eww = {
> -      enable = true;
> -      systemd.enable = true;
> -      yuckConfig = ''
> -        (defwindow powermenu
> -               :monitor 0
> -               :geometry (geometry
> -                   :anchor "center"
> -               )
> -               ; Widgets
> -               (box
> -                   :spacing 5
> -                   :class "powermenu"
> -                   :space-evenly true
> -                   :orientation "vertical"
> -                   :halign "left"
> -                   :valign "center"
> +  programs.eww = {
> +    enable = true;
> +    systemd.enable = true;
> +    yuckConfig = ''
> +      (defwindow powermenu
> +          :monitor 0
> +          :geometry (geometry
> +              :anchor "center"
> +          )
> +          ; Widgets
> +          (box
> +              :spacing 5
> +              :class "powermenu"
> +              :space-evenly true
> +              :orientation "vertical"
> +              :halign "left"
> +              :valign "center"
>  
> -                   ; Contents
> -                   (button
> -                       :onclick "eww open confirm-command --arg action=poweroff --arg command=poweroff"
> -                       "Poweroff"
> -                   )
> -                   (button
> -                       :onclick "eww open confirm-command --arg action=reboot --arg command=reboot"
> -                       "Reboot"
> -                   )
> -                   (button
> -                       :onclick "eww close powermenu"
> -                       "Cancel"
> -                   )
> -               )
> -           )
> +              ; Contents
> +              (button
> +                  :onclick "eww open confirm-command --arg action=poweroff --arg command=poweroff"
> +                  "Poweroff"
> +              )
> +              (button
> +                  :onclick "eww open confirm-command --arg action=reboot --arg command=reboot"
> +                  "Reboot"
> +              )
> +              (button
> +                  :onclick "eww close powermenu"
> +                  "Cancel"
> +              )
> +          )
> +      )
>  
> -           (defwindow confirm-command [action command]
> -               :monitor 0
> -               :geometry (geometry
> -                   :anchor "center"
> -                   :width "13%"
> -                   :height "8%"
> -               )
> -               ; Widgets
> -               (box
> -                   :class "powermenu"
> -                   :orientation "vertical"
> -                   :halign "center"
> -                   :valign "center"
> -                   "Are you sure you want to ''${action}"
> -                   (box
> -                       :orientation "horizontal"
> -                       :spacing 10
> -                       (button
> -                           :onclick "eww close confirm-command"
> -                           "No"
> -                       )
> -                       (button
> -                           :onclick command
> -                           "Yes"
> -                       )
> -                   )
> -               )
> -           )
> -      '';
> -      scssConfig = ''
> -                .powermenu {
> -          padding: 10px 10px;
> -          border-radius: 10px;
> -        }
> -      '';
> -    };
> +      (defwindow confirm-command [action command]
> +          :monitor 0
> +          :geometry (geometry
> +              :anchor "center"
> +              :width "13%"
> +              :height "8%"
> +          )
> +          ; Widgets
> +          (box
> +              :class "powermenu"
> +              :orientation "vertical"
> +              :halign "center"
> +              :valign "center"
> +              "Are you sure you want to ''${action}"
> +              (box
> +                  :orientation "horizontal"
> +                  :spacing 10
> +                  (button
> +                      :onclick "eww close confirm-command"
> +                      "No"
> +                  )
> +                  (button
> +                      :onclick command
> +                      "Yes"
> +                  )
> +              )
> +          )
> +      )
> +    '';
> +    scssConfig = ''
> +      .powermenu {
> +        padding: 10px 10px;
> +        border-radius: 10px;
> +      }
> +    '';
> +  };
>  
> -    nmt.script = ''
> -      yuckDir=home-files/.config/eww
> -      serviceFile=$(normalizeStorePaths home-files/.config/systemd/user/eww.service)
> +  nmt.script = ''
> +    yuckDir=home-files/.config/eww
> +    serviceFile=$(normalizeStorePaths home-files/.config/systemd/user/eww.service)
>  
> -      assertFileExists $yuckDir/eww.yuck
> -      assertFileExists $yuckDir/eww.scss
> +    assertFileExists $yuckDir/eww.yuck
> +    assertFileExists $yuckDir/eww.scss
>  
> -      assertFileContent $yuckDir/eww.yuck ${./expected.yuck}
> -      assertFileContent $yuckDir/eww.scss ${./expected.scss}
> -      assertFileContent "$serviceFile" ${./expected.service}
> -    '';
> -  };
> +    assertFileContent $yuckDir/eww.yuck ${./expected.yuck}
> +    assertFileContent $yuckDir/eww.scss ${./expected.scss}
> +    assertFileContent "$serviceFile" ${./expected.service}
> +  '';
>  }
> diff --git a/tests/modules/programs/eww/empty-settings.nix b/tests/modules/programs/eww/empty-settings.nix
> index 8fcd252b1..45677d9b8 100644
> --- a/tests/modules/programs/eww/empty-settings.nix
> +++ b/tests/modules/programs/eww/empty-settings.nix
> @@ -1,15 +1,13 @@
>  {
> -  config = {
> -    programs.eww = {
> -      enable = true;
> -    };
> -
> -    nmt.script = ''
> -      yuckDir=home-files/.config/eww
> -      serviceFile=$(normalizeStorePaths home-files/.config/systemd/user/eww.service)
> -      assertPathNotExists $yuckDir/eww.yuck
> -      assertPathNotExists $yuckDir/eww.scss
> -      assertPathNotExists $serviceFile
> -    '';
> +  programs.eww = {
> +    enable = true;
>    };
> +
> +  nmt.script = ''
> +    yuckDir=home-files/.config/eww
> +    serviceFile=home-files/.config/systemd/user/eww.service
> +    assertPathNotExists $yuckDir/eww.yuck
> +    assertPathNotExists $yuckDir/eww.scss
> +    assertPathNotExists $serviceFile
> +  '';
>  }
> diff --git a/tests/modules/programs/eww/expected.yuck b/tests/modules/programs/eww/expected.yuck
> index c3a88c6f4..f7fd7d5c0 100644
> --- a/tests/modules/programs/eww/expected.yuck
> +++ b/tests/modules/programs/eww/expected.yuck
> @@ -41,7 +41,7 @@
>          :orientation "vertical"
>          :halign "center"
>          :valign "center"
> -        "Are you sure you want to ''${action}"
> +        "Are you sure you want to ${action}"
>          (box
>              :orientation "horizontal"
>              :spacing 10
> ```

~~Let's just try it out.~~ It works!

eveeifyeve:
Recent commit adds news entry and `lib.mkRemovedOptionModule`.

rycee:

schrottkatze:
cool
this entirely breaks my 10+ file config 
thanks very nice that i now have to work around an unnecessary config change removing perfectly good features from a years long unmaintained program! thank you!! (**/s**)

PR #5992, opened 2024-10-23
launchd: wrap agents in .app bundles

Samasaur1:
This causes agents to appear in System Settings with a fancy name (and optionally icon), rather than the file name of the executable. This is particularly desirous when the "executable" is a shell script that waits for the Nix store to be available before running the true agent, because those show up in System Settings as `sh`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/launchd/default.nix
+++ b/modules/launchd/default.nix
@@ -38,6 +38,27 @@ let
 
   enabledAgents = filterAttrs (n: v: v.enable) cfg.agents;
 
+  toAgent = config: pkgs.writeText "${config.Label}.plist" (toPlist { } config);
+
+  agentPlists =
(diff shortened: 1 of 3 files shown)


Samasaur1:
LaunchAgents wrapped in app bundles fail to register unless `home-manager switch` is run from a graphical context (e.g., it fails over SSH), so I think I'm giving up on this.

Samasaur1:
FWIW there is the `launchctl managername` command. Checking this at runtime would potentially allow a version of this PR to be merged

Enzime:
Does it work if you have Full Disk Access turned on for SSH?

Samasaur1:
It doesn't appear to, no

PR #5994, opened 2024-10-23
granted:.assume-wrapped no longer present

wcarlsen:
.assume-wrapped is no longer present, use assume instead

### Description

.assume-wrapped is no longer present see error message:

`assume:source:2: no such file or directory: /nix/store/7gp6i2pdvyk7cc01g4w8wz1ymjpxgcqx-granted-0.34.1/bin/.assume-wrapped`

Using `assume` instead seems to work as intended.

As a small remark I'm not sure how backwards compatible this is, since this comes from an change in upstream packaging.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

It doesn't make sense that I review this alone, since I'm maintaining it. But adding myself anyways @wcarlsen

### DIFF
--- a/modules/programs/granted.nix
+++ b/modules/programs/granted.nix
@@ -28,7 +28,7 @@ in {
     programs.zsh.initExtra = mkIf cfg.enableZshIntegration ''
       function assume() {
         export GRANTED_ALIAS_CONFIGURED="true"
-        source ${package}/bin/.assume-wrapped "$@"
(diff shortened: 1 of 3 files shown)


rycee:
Thanks! Rebased to master in 454e8d6b15aafb02fd22bba0edc4cc0b06dd0f41 🙂

PR #6006, opened 2024-10-26
nixpkgs-disabled: correct assertion

sellout:
### Description

Should only pass if _both_ attributes are `null`.

I recently found myself in a confusing situation when I added my usual
`homeConfiguration` to a `nixosConfiguration` that had
`home-manager.useGlobalPackages = true`.

`nixos-rebuild` was just claiming that some attributes were missing from `pkgs`,
rather than asserting that my `nixpkgs.overlays` was set when it couldn’t be.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@thiagokokada

### DIFF
--- a/modules/misc/nixpkgs-disabled.nix
+++ b/modules/misc/nixpkgs-disabled.nix
@@ -64,7 +64,7 @@ in {
 
   config = {
     assertions = [{
-      assertion = cfg.config == null || cfg.overlays == null;
+      assertion = cfg.config == null && cfg.overlays == null;
       message = ''
         `nixpkgs` options are disabled when `home-manager.useGlobalPkgs` is enabled.
       '';


sellout:
I tried to add some new test cases around this, but struggled to get `asserts.nix` working in the integration tests. Would be happy to add them if I could get a pointer of how to use `test.asserts.assertions.expected` there.

Also, I ran into a test (targets-darwin/user-defaults.nix) that was failing regardless of my changes. I can push a fix in this PR, or make a separate one for that.

And finally, I’m curious why `nixpkgs.overlays` isn’t available with `useGlobalPkgs = true`. Can’t something like `_module.args.pkgs = pkgs.appendOverlays cfg.overlays` work?

sellout:
I this this would be good to get in, and the review should be trivial …

(I think all of my followup questions are optional)

khaneliman:
Handled in another PR

sellout:
@khaneliman Can you point to the PR that handles this? The only one I found was #4228, and while that does eliminate this issue, the overall change seems fairly controversial and far from guaranteed to get in; while this PR is a minimal bug-fix.

That PR does have a lot of discussion around _why_ nixpkgs-disabled is the way it is, so that is very helpful.

khaneliman:
https://github.com/nix-community/home-manager/pull/6172 and then https://github.com/nix-community/home-manager/pull/6466

sellout:
Thanks! I had checked to see if master contained the fix, but it didn’t (which makes sense in the context of #6466), so I had assumed the relevant PR was still open.

PR #6009, opened 2024-10-26
mako: add max-history option

istudyatuni:
### Description

Add `max-history` option for mako

### Checklist

  - If someone add `max-history` to extraConfig, then mako doesn't work because cannot parse the config, so this case can be ignored

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@onny

### DIFF
--- a/modules/services/mako.nix
+++ b/modules/services/mako.nix
@@ -33,6 +33,15 @@ in {
         '';
       };
 
+      maxHistory = mkOption {
+        default = 5;
+        type = types.nullOr types.int;
+        description = ''
+          Set maximum number of expired notifications to keep in the history
+          buffer. Set 0 to disable history.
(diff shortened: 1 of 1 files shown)


istudyatuni:
There was no any comments

onny:
Looks good for me, thank you :+1:

modules/services/mako.nix
```diff
@@ -33,6 +33,14 @@ in {
         '';
       };
 
+      maxHistory = mkOption {
+        default = 5;
```
sumnerevans, line 37: For future us, default is documented here: https://github.com/emersion/mako/blob/master/doc/mako.5.scd#global-configuration-options

sumnerevans (approved):

PR #6014, opened 2024-10-27
git-credential-oauth: fix ensure git extraConfig is last in list

KostaGorod:
### Description

when adding storage helpers via git.extraConfig,
git-credential-oauth is not last is `.git/config`, which is the desired state: [ git-credential-oauth must be configured last](https://github.com/hickford/git-credential-oauth?tab=readme-ov-file#how-it-works).
without the correct order, the storage helper is never called.

added mkAfter in git.extraConfig to ensure our oauth is last, as required to work with additional helpers

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@tomodachi94

### DIFF
--- a/modules/programs/git-credential-oauth.nix
+++ b/modules/programs/git-credential-oauth.nix
@@ -29,7 +29,7 @@ in {
   config = lib.mkIf cfg.enable {
     home.packages = [ cfg.package ];
 
-    programs.git.extraConfig.credential.helper = [
+    programs.git.extraConfig.credential.helper = lib.mkAfter [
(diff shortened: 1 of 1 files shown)


KostaGorod:
@tomodachi94 could you please review this request?
I had issues while trying to add `libsecret` as helper but ensuring that oauth is the last, 
initially I used mkBefore on my additional config but I think added `mkAfter` here is the better solution overall

tomodachi94 (approved):
Lgtm, didn't even think about the ordering 😅 thank you!

rycee:

PR #6017, opened 2024-10-29
nixos: add `key` to shared module to allow disabling it

Enzime:
### Description

By adding `key`, this allows users to disable this shared module or they can choose to not disable this shared module (by filtering by key before disabling)

This means users can disable all shared modules if all modules are paths or attrsets with a key:

`configuration.nix`:

```nix
{ config, ... }:

{
  home-manager.users.enzime = { ... }: {
    disabledModules = config.home-manager.sharedModules;
  };
}
```

Or disabling just this module specifically:

```nix
{ ... }:

{
  home-manager.users.enzime = { ... }: {
    disabledModules = [ { key = "home-manager#nixos-shared-module"; } ];
  };
}
```

Or disabling all modules when you have modules you can't disable (like lambdas):

```nix
{ ... }:

{
  home-manager.users.enzime = { ... }: {
    disabledModules = lib.filter (v: lib.isString v || lib.isPath v || (lib.isAttrs v && v ? key)) config.home-manager.sharedModules;
  };
}
```

https://nixos.org/manual/nixos/unstable/#sec-replace-modules

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

cc @rycee

### DIFF
--- a/nixos/default.nix
+++ b/nixos/default.nix
@@ -19,13 +19,17 @@ in {
         extraSpecialArgs.nixosConfig = config;
 
         sharedModules = [{
-          # The per-user directory inside /etc/profiles is not known by
-          # fontconfig by default.
-          fonts.fontconfig.enable = lib.mkDefault
-            (cfg.useUserPackages && config.fonts.fontconfig.enable);
+          key = "home-manager#nixos-shared-module";
 
-          # Inherit glibcLocales setting from NixOS.
-          i18n.glibcLocales = lib.mkDefault config.i18n.glibcLocales;
+          config = {
+            # The per-user directory inside /etc/profiles is not known by
+            # fontconfig by default.
+            fonts.fontconfig.enable = lib.mkDefault
+              (cfg.useUserPackages && config.fonts.fontconfig.enable);
+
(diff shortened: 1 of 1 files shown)


roberth:
The `key` is good, but I don't understand why `sharedModules` is a list, and it probably doesn't need to be.
List indexes aren't good identifiers, so I would recommend to change it to `coercedTo (listOf raw) (ms: { imports = ms; }) deferredModule`.
Useful, stable identifiers for this module / these modules can be provided in the form of flake output attributes, whose values (modules) each have a unique `key`.

roberth (approved):
Good first step, and [further improvements](https://github.com/nix-community/home-manager/pull/6017#issuecomment-2446947105) possible.

Enzime:
Changing `sharedModules` is a lot bigger of a refactor than the currently proposed changes, I think it would be best to do that in a different PR

Enzime:
cc @teto :)

teto:
I dont know this aspect of the module system but the approval of roberth is good enough for me

PR #6026, opened 2024-11-01
Sbt credentials scoping to build rather than projects.

zarthross:
### Description

In Sbt (prior to 2.0) a sbt key not scoped to `ThisBuild` will always be at the project level, rather than at the build level.
In order to do some things, like use a private repository configured by sbt with sbt-scalafix, we need the credentials scope to the build, otherwise the credentials show as non-existant.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@kubukoz

### DIFF
--- a/modules/programs/sbt.nix
+++ b/modules/programs/sbt.nix
@@ -25,7 +25,7 @@ let
     in
     ''
       lazy val ${symbol} = "${cred.passwordCommand}".!!.trim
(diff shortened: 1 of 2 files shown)


kubukoz (approved):

zarthross:
@rycee Can you give this a look please? 🥺

zarthross:
Ping @rycee, this should be an easy quick merge.

PR #6030, opened 2024-11-02
gpg: add dirmngrSettings and gpgsmSettings options + remove obsolete default values

folliehiyuki:
### Description

The PR adds additional options to manage `dirmngr.conf` and `gpgsm.conf` files. The option I'm most interested in is `keyserver`, which has been deprecated in `gpg.conf` and should now be set in `dirmngr.conf`.

While at it, I've also deleted some obsolete values from the default gpg settings. The reasons are listed in the 2nd commit message.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

### DIFF
--- a/modules/programs/gpg.nix
+++ b/modules/programs/gpg.nix
@@ -18,6 +18,16 @@ let
     listsAsDuplicateKeys = true;
   } cfg.scdaemonSettings;
 
+  dirmngrCfgText = generators.toKeyValue {
+    inherit mkKeyValue;
+    listsAsDuplicateKeys = true;
+  } cfg.dirmngrSettings;
+
+  gpgsmCfgText = generators.toKeyValue {
+    inherit mkKeyValue;
+    listsAsDuplicateKeys = true;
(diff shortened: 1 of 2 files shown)


modules/programs/gpg.nix
```diff
@@ -293,6 +342,10 @@ in
 
     home.file."${cfg.homedir}/scdaemon.conf".text = scdaemonCfgText;
 
+    home.file."${cfg.homedir}/dirmngr.conf".text = dirmngrCfgText;
+
+    home.file."${cfg.homedir}/gpgsm.conf".text = gpgsmCfgText;
```
teto: seems like this always create the files. Can it create them only when settings are actually set ? it makes hybrid setups easier

teto:
could you update the test to exert those new options ?

modules/programs/gpg.nix
folliehiyuki: I was blindly following how the existing `scdaemonSettings` option was handled.

Issue #6020, opened 2024-10-30
Support for `streamlink`


SymphonySimper:
### Description
It would be great if `streamlink` can be configured with `home-manager`.

- Nix Package: `streamlink`
- Source: https://github.com/streamlink/streamlink

SymphonySimper:
@berbiche @rycee @sumnerevans 

Sorry for the mention, can anyone please check the PR. Thanks!

PR #6031, opened 2024-11-02
streamlink: add module

folliehiyuki:
### Description

Closes https://github.com/nix-community/home-manager/issues/6020.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

### DIFF
--- a/modules/programs/streamlink.nix
+++ b/modules/programs/streamlink.nix
@@ -1,0 +1,145 @@
+{ config, lib, pkgs, ... }:
+
+let
+
+  cfg = config.programs.streamlink;
+
+  renderSettings = settings:
+    lib.concatLines (lib.remove "" (lib.mapAttrsToList (name: value:
+      if (builtins.isBool value) then
+        if value then name else ""
+      else if (builtins.isList value) then
+        lib.concatStringsSep "\n"
+        (builtins.map (item: "${name}=${builtins.toString item}") value)
+      else
+        "${name}=${builtins.toString value}") settings));
+
+  pluginType = lib.types.submodule {
+    options = {
+      src = lib.mkOption {
+        type = with lib.types; nullOr (either path lines);
+        default = null;
+        description = ''
+          Source of the custom plugin. The value should be a path to the
+          plugin file, or the text of the plugin code. Will be linked to
+          {file}`$XDG_DATA_HOME/streamlink/plugins/<name>.py` (linux) or
+          {file}`Library/Application Support/streamlink/plugins/<name>.py` (darwin).
+        '';
+        example = lib.literalExpression "./custom_plugin.py";
+      };
+
+      settings = lib.mkOption {
+        type = with lib.types;
+          attrsOf
+          (oneOf [ bool int float str (listOf (oneOf [ int float str ])) ]);
+        default = { };
+        example = lib.literalExpression ''
+          {
+            quiet = true;
+          }
+        '';
+        description = ''
+          Configuration for the specific plugin, written to
+          {file}`$XDG_CONFIG_HOME/streamlink/config.<name>` (linux) or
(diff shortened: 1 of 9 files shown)


folliehiyuki:
@SymphonySimper can you try this PR?

SymphonySimper:
> @SymphonySimper can you try this PR?

It works just fine, thanks!

modules/programs/streamlink.nix
```diff
@@ -0,0 +1,143 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
```
khaneliman, line 3:
```suggestion
```

modules/programs/streamlink.nix
```diff
@@ -0,0 +1,143 @@
...
+              quiet = true;
+            };
+          };
+        }
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
```
khaneliman, line 101:
```suggestion
    home.packages = lib.mkIf (cfg.package != null) [ cfg.package ];
```

modules/programs/streamlink.nix
```diff
@@ -0,0 +1,143 @@
...
+    };
+  };
+
+in {
+  meta.maintainers = [ hm.maintainers.folliehiyuki ];
+
+  options.programs.streamlink = {
+    enable = mkEnableOption "streamlink";
+
+    package = mkPackageOption pkgs "streamlink" { };
```
khaneliman, line 56:
```suggestion
    package = mkPackageOption pkgs "streamlink" { nullable = true; };
```

modules/programs/streamlink.nix
```diff
@@ -0,0 +1,143 @@
...
+  cfg = config.programs.streamlink;
+
+  renderSettings = mapAttrsToList (name: value:
+    if (builtins.isBool value) then
+      if value then name else ""
+    else if (builtins.isList value) then
+      concatStringsSep "\n"
+      (builtins.map (item: "${name}=${builtins.toString item}") value)
+    else
+      "${name}=${builtins.toString value}");
```
khaneliman, line 16: This looks like it's re-implementing [lib.generators.toKeyValue](https://github.com/nixos/nixpkgs/blob/master/lib/generators.nix#L174-L174) ? Similar to https://github.com/nix-community/home-manager/pull/6674

modules/programs/streamlink.nix
```diff
@@ -0,0 +1,143 @@
...
+
+    xdg.dataFile = mkIf pkgs.stdenv.hostPlatform.isLinux (mapAttrs'
+      (name: value:
+        nameValuePair "streamlink/plugins/${name}.py" (mkIf (value.src != null)
+          (if (builtins.isPath value.src) then {
+            source = value.src;
+          } else {
+            text = value.src;
+          }))) cfg.plugins);
+  };
```
khaneliman, line 142: Would be easier to read if we use `mkMerge` and a `lib.generators` function. Would also recommend moving the config  path out into a variable outside of the scope of this section so you don't need to have the check inside here on the path.

modules/programs/streamlink.nix
folliehiyuki: `toKeyValue` can't handle option key without value (the `=` separator will always render). You can see `player-no-close` option in the test case for an example.

folliehiyuki: I can, of course, tweak the generator `mkKeyValue` function, but it feels just doing the same thing as writing my own function here.

khaneliman:
Looks like it needs to be rebased on master, still.

khaneliman:
Alright, looks like darwin failure was due to needing to wrap the path in quotes in the assertion, to handle the space.

PR #6032, opened 2024-11-02
git: add default value null to programs.git.signing.key

folliehiyuki:
### Description

Right now, if other signing options are defined without `key` specified, home-manager switch will fail.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/programs/git.nix
+++ b/modules/programs/git.nix
@@ -18,6 +18,7 @@ let
     options = {
       key = mkOption {
         type = types.nullOr types.str;
+        default = null;
         description = ''
           The default GPG signing key fingerprint.
 


Adda0 (approved):
Seems reasonable to me. When the `key` is `null`, GnuPG will decide what signing key to use based on the author of the commit, which in my opinion should be the default behaviour. When one wants to overwrite that, they can set `key` to a specific signing key.

dawidd6 (approved):

teto:
LGTM. Can you rebase to fix CI plz (can't merge without green CI)

PR #6037, opened 2024-11-03
zellij: Add additional options for shell integration

Adda0:
### Description

This PR adds additional options for integrating Zellij with shells, namely:
1. autostarting Zellij on shell start.
2. automatically attaching to an existing session on Zellij autostart if such session exists.
3. automatically exiting the shell when the Zellij session is terminated.

The new Zellij configurations can look like this:
```nix
  programs.zellij = {
    enable = true;

    autostartOnShellStart = {
      enable = true;
      attachExistingSession = false;
      exitShellOnExit = true;
    };

    enableZshIntegration = true;
    enableFishIntegration = true;
    enableBashIntegration = true;
  };
``` 

Currently, the options are implemented such a single configuration, which can get applied to shells by using `enable<Shell>Integration`. This means that one cannot have one configuration for, Zsh, another for Bash, and completely different one for Fish.
- Would separate configurations for each shell be useful? For example, use autostart in one shell, but load completions only in another?

The PR is supposed to supersede #5333 as it implements the same integration with Fish, albeit with different options. Closes #5333.

I tested the PR primarily on Zsh and Bash. Everything works for Bash and Zsh as expected.

~~However, Fish seems to be having issues. Namely, the completions are loaded, but for each shell instance, one must manually activate completions on `zellij` command by pressing `Tab` to allow for the functions (such as `ze`, `zfr` etc.) to register and the completions to load. Autostart for Fish does not work at all, as it seems that the `interactiveShellInit` does not write to the config file at all. Needs further investigation. Maybe a similar workaround as for Zsh is needed?~~ 
Fish integration works well, too.

I need to perform some more testing on Fish shell. When this is completed, I will open the PR for merging. Nevertheless, I ask for a review for the PR now, especially regarding the possibility of allowing different configurations for each shell (basically just having `autostartOn<Shell>Start` module for each shell instead of a shared `autostartOnShellStart`). But then the `enable<Shell>Integration` seems kinda pointless.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@mainrs

### DIFF
--- a/modules/programs/zellij.nix
+++ b/modules/programs/zellij.nix
@@ -8,18 +8,39 @@ let
   yamlFormat = pkgs.formats.yaml { };
   zellijCmd = getExe cfg.package;
 
+  autostartOnShellStartModule = types.submodule {
+    options = {
+      enable = mkEnableOption "" // {
+        description = ''
+          Whether to autostart Zellij session on shell creation.
+        '';
+      };
+
+      attachExistingSession = mkEnableOption "" // {
+        description = ''
+          Whether to attach to the default session after being autostarted if a Zellij session already exists.
+        '';
+      };
+
+      exitShellOnExit = mkEnableOption "" // {
+        description = ''
+          Whether to exit the shell when Zellij exits after being autostarted.
+        '';
+      };
+    };
+  };
 in {
   meta.maintainers = [ hm.maintainers.mainrs ];
 
   options.programs.zellij = {
-    enable = mkEnableOption "zellij";
+    enable = mkEnableOption "Zellij";
 
     package = mkOption {
       type = types.package;
       default = pkgs.zellij;
       defaultText = literalExpression "pkgs.zellij";
       description = ''
-        The zellij package to install.
+        The Zellij package to install.
       '';
     };
 
@@ -41,6 +62,15 @@ in {
 
         See <https://zellij.dev/documentation> for the full
         list of options.
+      '';
+    };
+
+    autostartOnShellStart = mkOption {
+      type = types.nullOr autostartOnShellStartModule;
+      default = null;
+      description = ''
+        Options related to autostarting Zellij on shell creation.
+        Requires enable<Shell>Integration to apply to the respective <Shell>.
       '';
     };
 
@@ -69,17 +99,32 @@ in {
         text = lib.hm.generators.toKDL { } cfg.settings;
       };
 
-    programs.bash.initExtra = mkIf cfg.enableBashIntegration (mkOrder 200 ''
-      eval "$(${zellijCmd} setup --generate-auto-start bash)"
-    '');
+    programs.zsh.initExtra = mkIf (cfg.enableZshIntegration)
+      (if cfg.autostartOnShellStart.enable then ''
+        eval "$(${zellijCmd} setup --generate-auto-start zsh)"
+      '' else
+        "");
 
-    programs.zsh.initExtra = mkIf cfg.enableZshIntegration (mkOrder 200 ''
-      eval "$(${zellijCmd} setup --generate-auto-start zsh)"
-    '');
-
-    programs.fish.interactiveShellInit = mkIf cfg.enableFishIntegration
-      (mkOrder 200 ''
+    programs.fish.interactiveShellInit = mkIf (cfg.enableFishIntegration)
+      (if cfg.autostartOnShellStart.enable then ''
         eval (${zellijCmd} setup --generate-auto-start fish | string collect)
-      '');
+      '' else
+        "");
+
+    programs.bash.initExtra = mkIf (cfg.enableBashIntegration)
+      (if cfg.autostartOnShellStart.enable then ''
+        eval "$(${zellijCmd} setup --generate-auto-start bash)"
+      '' else
+        "");
+
+    home.sessionVariables = mkIf cfg.autostartOnShellStart.enable {
+      ZELLIJ_AUTO_ATTACH =
+        if cfg.autostartOnShellStart.attachExistingSession then
+          "true"
+        else
+          "false";
+      ZELLIJ_AUTO_EXIT =
+        if cfg.autostartOnShellStart.exitShellOnExit then "true" else "false";
+    };
   };
 }
--- a/modules/programs/zellij.nix
+++ b/modules/programs/zellij.nix
@@ -3,32 +3,8 @@
 with lib;
 
 let
-
   cfg = config.programs.zellij;
   yamlFormat = pkgs.formats.yaml { };
-  zellijCmd = getExe cfg.package;
-
-  autostartOnShellStartModule = types.submodule {
-    options = {
-      enable = mkEnableOption "" // {
-        description = ''
-          Whether to autostart Zellij session on shell creation.
-        '';
-      };
-
-      attachExistingSession = mkEnableOption "" // {
-        description = ''
-          Whether to attach to the default session after being autostarted if a Zellij session already exists.
-        '';
-      };
-
-      exitShellOnExit = mkEnableOption "" // {
-        description = ''
-          Whether to exit the shell when Zellij exits after being autostarted.
-        '';
-      };
-    };
-  };
 in {
   meta.maintainers = [ hm.maintainers.mainrs ];
 
@@ -65,12 +41,23 @@ in {
       '';
     };
 
-    autostartOnShellStart = mkOption {
-      type = types.nullOr autostartOnShellStartModule;
-      default = null;
+    attachExistingSession = mkOption {
+      type = types.bool;
+      default = false;
       description = ''
-        Options related to autostarting Zellij on shell creation.
-        Requires enable<Shell>Integration to apply to the respective <Shell>.
+        Whether to attach to the default session after being autostarted if a Zellij session already exists.
+
+        Requires shell integration to be enabled to have effect.
+      '';
+    };
+
+    exitShellOnExit = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Whether to exit the shell when Zellij exits after being autostarted.
+
+        Requires shell integration to be enabled to have effect.
       '';
     };
 
@@ -84,7 +71,10 @@ in {
       lib.hm.shell.mkZshIntegrationOption { inherit config; };
   };
 
-  config = mkIf cfg.enable {
+  config = let
+    shellIntegrationEnabled = (cfg.enableBashIntegration
+      || cfg.enableZshIntegration || cfg.enableFishIntegration);
+  in mkIf cfg.enable {
     home.packages = [ cfg.package ];
 
     # Zellij switched from yaml to KDL in version 0.32.0:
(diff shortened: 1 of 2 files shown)


Adda0:
The Fish integration seems to work well. I forget to enable Fish shell in Home Manager during my testing. Therefore, I am opening the PR from its draft state. The remaining question is whether to enable shell-specific autostart options, or not (see above).

Dietr1ch:
I was interested in this and merged it into my testing branch[^0] (after having successfully built my config from it),

I'm adding the fish bits from your new config example,

```nix
  programs.zellij = {
        # New settings from https://github.com/nix-community/home-manager/pull/6037
        enableShellCompletions = true;

        autostartOnShellStart = {
          enable = true;
          attachExistingSession = false;
          exitShellOnExit = true;
        };
        enableFishIntegration = true;
  };
```

And then got this build failure (I'm using flakes and suspect you might not)


```fish
home-manager switch --no-out-link --keep-going --flake ~/.config/home-manager
```

``` text
error:
         at <nix/derivation-internal.nix>:34:12:
           33|
           34|   strict = derivationStrict drvAttrs;
             |            ^
           35|

       … while evaluating derivation 'home-manager-generation'
         whose name attribute is located at /nix/store/0sxsfzswhjck7f4sbsvznz285s4i7y64-source/pkgs/stdenv/generic/make-derivation.nix:336:7

       … while evaluating attribute 'buildCommand' of derivation 'home-manager-generation'
         at /nix/store/0sxsfzswhjck7f4sbsvznz285s4i7y64-source/pkgs/build-support/trivial-builders/default.nix:59:17:
             |                 ^

       … while evaluating the option `home.activation.installPackages.data':

       … while evaluating definitions from `/nix/store/8r5zrd713yknnc39ynz7qkwfflhhmfh0-source/modules/home-environment.nix':

       … while evaluating the option `home.sessionVariables':

       … while evaluating definitions from `/nix/store/8r5zrd713yknnc39ynz7qkwfflhhmfh0-source/modules/programs/zellij.nix':

       error: expected a set but found null: null


Compilation exited abnormally with code 1 at Wed Nov 13 21:25:56
```

The `--show-trace` error is quite long to post here, but points to issue with `ZELLIJ_AUTO_ATTACH` in [modules/programs/zellij.nix](https://github.com/nix-community/home-manager/compare/master...Adda0:home-manager:zellij-settings#diff-ef086b65a6c43990c41a9b685a11b95a0aab7d8ae9974484372946c6e6463f93)


```text
       … while selecting an attribute
         at /nix/store/8r5zrd713yknnc39ynz7qkwfflhhmfh0-source/modules/programs/zellij.nix:177:34:
          176|
          177|     home.sessionVariables = mkIf cfg.autostartOnShellStart.enable {
             |                                  ^
          178|       ZELLIJ_AUTO_ATTACH =

       error: expected a set but found null: null

```

It seems that you'd be better of inconditionally setting `ZELLIJ_AUTO_ATTACH` to reflect `programs.zellij.attachExistingSession` even if the value is being defaulted to false.

---

[^0]: https://github.com/Dietr1ch/home-manager/tree/test

Adda0:
Aha. I remember that I encountered this issue when implementing this and have thought I already fixed it. Glad I am not crazy, looking at the code now and thinking that this is already resolved :)

Dietr1ch:
BTW, it seems this lets you shoot you in the foot because of a zellij bug, but there's nothing to change over here.


https://github.com/zellij-org/zellij/issues/3773

modules/programs/zellij.nix
```diff
@@ -69,17 +107,44 @@ in {
...
+      (mkOrder 200 ((if cfg.autostartOnShellStart.enable then ''
         eval (${zellijCmd} setup --generate-auto-start fish | string collect)
-      '');
+      '' else
+        "") + (if cfg.enableShellCompletions then ''
+          eval (${zellijCmd} setup --generate-completion fish | string collect)
+        '' else
+          "")));
+
+    programs.bash.initExtra = mkIf (cfg.enableBashIntegration) (mkOrder 200
```
mainrs, line 131: Why exactly is the mkOrder 200 required?

mainrs (commented):
Besides that this looks good to me. Tests are also included. I think for the whole YAML to KDL issue it might be better to create a final PR that includes all the KDL code instead of having multiples like right now.

Some of them do more than others, some are already superseeded. Would be great if you could help out on that front! I don't have access to a nix machine that easily right now, so it's hard to test all the stuff.

modules/programs/zellij.nix
Adda0: It is not required. Some packages in hm use this to place the generated completions/aliases a bit earlier in the config files for each respective shell so that the generated options can be easily modified by manual configuration or other programs. However, it is not necessary at all and I can remove them.

Adda0:
What about whether to support per-shell autostart options, i.e. using `autostartOn<Shell>Start` submodules instead of the generic `autostartOnShellStart` option? Any thoughts?

> I think for the whole YAML to KDL issue it might be better to create a final PR that includes all the KDL code instead of having multiples like right now.

I agree that it might be necessary. I intend to work on this. First, I will need to investigate each PR, and then I will see what makes the most sense to me. I would welcome any and all suggestions.

mainrs:
> What about whether to support per-shell autostart options, i.e. using `autostartOn<Shell>Start` submodules instead of the generic `autostartOnShellStart` option? Any thoughts?

I'm not sure what the best approach would be. Maybe rycee can chime in before merging the PR!

Adda0:
I was made aware of the fact that shell completions are already loaded in the nixpkgs `zellij` package and should never be an option in home manager modules. The issue with the accompanying aliases is that they are not loaded until a first Zellij autocomplete is executed. It is an issue with the upstream Zellij. Therefore, I will perform the following steps:
1. remove the option to enable autocompletion (which also includes the aliases mentioned above) from this PR.
2. create a PR for a temporary fix the inclusion of the aliases in the nixpkgs package `zellij`, as per https://github.com/zellij-org/zellij/issues/1933#issuecomment-2274464004, and
3. convert this PR into a draft until this issue is resolved.

KamWithK:
Would be cool to add Nushell as another shell option

Dietr1ch:
> Would be cool to add Nushell as another shell option

You might want upstream support from Zellij first,

```fish
zellij setup --generate-auto-start nushell
zellij setup --generate-auto-start nu
```

```text
Unsupported shell: nushell
Unsupported shell: nu
```

KamWithK:
> > Would be cool to add Nushell as another shell option
> 
> You might want upstream support from Zellij first,
> 
> ```fish
> zellij setup --generate-auto-start nushell
> zellij setup --generate-auto-start nu
> ```
> 
> ```text
> Unsupported shell: nushell
> Unsupported shell: nu
> ```

Ah I'll look into that then!

KamWithK:
> > Would be cool to add Nushell as another shell option
> 
> You might want upstream support from Zellij first,
> 
> ```fish
> zellij setup --generate-auto-start nushell
> zellij setup --generate-auto-start nu
> ```
> 
> ```text
> Unsupported shell: nushell
> Unsupported shell: nu
> ```

Looks like there is currently a PR for it awaiting review:
https://github.com/zellij-org/zellij/pull/3206

Adda0:
> 1. remove the option to enable autocompletion (which also includes the aliases mentioned above) from this PR.

Done. I consider this first PR finished and ready for reviews (and if deemed suitable, merging). I think it would be wise to first merge this PR, and only in a next PR try to fix the completions. The initial experiment to fix this issue with the zellij package in nixpkgs failed.

khaneliman:
> I have rebased the PR to reflect the changes to shell integrations in #6358.
> 
> The tests need to be rerun when the nushell issue with a hash of the vendored dependencies is resolved. It is not caused by this PR.

That was just fixed if you want to rebase again.

khaneliman:
The changes don't work in the current form, as I was fixing them I noticed a logic issue with the change, too. This basically makes shell integration a 2 step process minimum, which is kinda weird UX. I would leave the `enableShellIntegration` stuff alone and that's the decision on whether it auto attaches still. The other 2 are just ancillary options to fine tune the auto attach behavior, so they can be moved up and just add a warning about them not having an effect by themselves. 

Pushed a commit with that method, let me know what you think.

Adda0:
I will do a proper review, but does this not always force the user to automatically launch Zellij on shell start? That was the purpose of the `<autostartModule>.enable` option. This is now automatically enabled when the shell integration is enabled? That would not be right. The shell integration should enable completions, add Zellij specific functions etc., but autostarting Zellij should definitely be a separate option, dependent on the shell integration.

modules/programs/zellij.nix
```diff
@@ -70,16 +90,33 @@ in {
...
       '');
+
+    home.sessionVariables = mkIf shellIntegrationEnabled {
+      ZELLIJ_AUTO_ATTACH =
+        if cfg.attachExistingSession then "true" else "false";
+      ZELLIJ_AUTO_EXIT = if cfg.exitShellOnExit then "true" else "false";
+    };
+
+    warnings =
+      optional (cfg.attachExistingSession -> !shellIntegrationEnabled) ''
```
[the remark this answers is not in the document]
khaneliman: You're right, thanks

Adda0:
> > I will do a proper review later, but does this not always force the user to automatically launch Zellij on shell start when the shell integration is enabled? That was the purpose of the `<autostartModule>.enable` option. This is now automatically enabled when the shell integration is enabled? That would not be right. The shell integration should enable completions, add Zellij specific functions etc., but autostarting Zellij should definitely be a separate option, dependent on the shell integration.
> 
> There was never any concept of that in this module. The only purpose of the integration has always been auto attaching on shell start.

There are two features in Zellij:
1. *Autostart* Zellij when the shell session starts.
2. *Auto-attach* to an existing session (same name, different session to the shell session) in the Zellij system, that is, after starting Zellij, automatically attach a session.

The way Zellij does the autostart of itself is adding a shell script which either only start Zellij, or starts Zellij with the auto-attach option enabled. That is, you can enable autostart (that is what I currently use), but then decide whether to also automatically enable auto-attach (that is what I have disabled in my configuration, for example), and whether to exist the shell session when the autostarted Zellij session exits (still in this one shell script). This is exactly what was possible in this autostart module and how I use it.

> > > The changes don't work in the current form
> > 
> > 
> > What did not work for you? I use the original version daily, and it works for me fine.
> 
> It broke existing configs that didn't use the new options. You defined a submodule with a null default value and then accessed suboptions without null checking.

It should not. Everything was disabled by default, so it behaved like previously. Only when the autostart option was enabled, the shell script was injected, and when the additional options were enabled (simply by setting environment variables, the behaviour of the shell script was modified). If this was not the case for you, there was indeed an issue somewhere.

Adda0:
> > And the module made it obvious to not ever get into an invalid state (so the warnings were not even necessary) by hiding the additional options, which require the autostart of Zellij on shell start to be used, behind the `<autostartModule>.enable` option. It made more sense to me as it was, to be honest.
> 
> It required 2 options to be set to auto start. Both `enableXIntegration` and `enableautostart`. `enableXintegration` did nothing on its own and there was no warning, just an eval error when you didn't set any of the new options. The only new behavior here are the 2 environment variables for auto attach and exit.

This PR has not added the `enableShellIntegration` options. The enable module only reused the existing enable shell integration options (since that makes sense to me: only enable autostart for the shells you enabled integration for), but in order to enable the autostart, since that is not a default desired feature for every user to use by default, one has to explicitly enable the autostart module, and if the autostart module is enabled, one has to explicitly choose the env vars to customize the autostart behaviour.

khaneliman:
> There are two features in Zellij:
> 
>     1. _Autostart_ Zellij when the shell session starts.
> 

This is the behavior of the current module and `enableXIntegration`.

>     2. _Auto-attach_ to an existing session (same name, different session to the shell session) in the Zellij system, that is, after starting Zellij, automatically attach a session.
> 

This is the new environment variable, from what I could see. Looking at upstream code, it is only checked inside the `auto-start` script https://github.com/search?q=repo%3Azellij-org%2Fzellij%20ZELLIJ_AUTO_ATTACH&type=code and https://github.com/search?q=repo%3Azellij-org%2Fzellij+ZELLIJ_AUTO_EXIT&type=code

> It should not. Everything was disabled by default, so it behaved like previously. Only when the autostart option was enabled, the shell script was injected, and when the additional options were enabled (simply by setting environment variables, the behaviour of the shell script was modified). If this was not the case for you, there was indeed an issue somewhere.

You can observe it easily by updating the test case. `autostartOnShellStart` is an option with a null default value. If you just comment out the new attributes added to the `zellij-enable-shells` test (which everyone who isn't aware of this change would have). You get a `expected a set but found a null`.

>This PR has not added the enableShellIntegration options. The enable module only reused the existing enable shell integration options (since that makes sense to me: only enable autostart for the shells you enabled integration for), but in order to enable the autostart, since that is not a default desired feature for every user to use by default, one has to explicitly enable the autostart module, and if the autostart module is enabled, one has to explicitly choose the env vars to customize the autostart behaviour.

Exactly, they were already there to enable auto starting on shell start. So people who had them enabled and expected that functionality would have it broken because it does nothing by itself with the original change. But, that means those options do nothing at all because you still had to configure the new options to have any effect. I think there is just a misunderstanding of the current state of the module.

Adda0:
> I think there is just a misunderstanding of the current state of the module.

That is entirely possible, as what I am talking about is the state of the Zellij module at the time of writing this PR. If the module changed in this regard (which I have then missed, as I am not aware of any such change), then what I am talking about might not be true any more. I will check this out during my proper review.

> I would be in favor of changing enableXIntegration to inject

This is what this PR originally did as well, but then I found out that this should be the responsibility of the nixpkgs package for Zellij (this is done for all packages in nixpkgs with completions, btw.), only upstream Zellij generates incorrect completions which require to be run at least once in each shell session to function. Therefore, I removed this from this PR and intend to fix this upstream in the Zellij repo. Fixing this in nixpkgs Zellij package as a workaround before Zellij upstream fixes this issue failed so far.

PR #6039, opened 2024-11-04
podman: install package if enabled and create config files

dawidd6:
### Description

Copied some useful options from NixOS podman module and adapted the configuration to home-manager.

Also setting `services.podman.enable = true` will now add `podman` to `PATH`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@bamhm182 @n-hass

### DIFF
--- a/modules/services/podman-linux/default.nix
+++ b/modules/services/podman-linux/default.nix
@@ -11,61 +11,63 @@ in {
   options.services.podman = {
     enable = lib.mkEnableOption "Podman, a daemonless container engine";
 
-    containersConf.settings = lib.mkOption {
-      type = toml.type;
-      default = { };
-      description = "containers.conf configuration";
-    };
-
-    storage.settings = lib.mkOption {
-      type = toml.type;
-      description = "storage.conf configuration";
-    };
-
-    registries = {
-      search = lib.mkOption {
-        type = lib.types.listOf lib.types.str;
-        default = [ "docker.io" ];
-        description = ''
-          List of repositories to search.
-        '';
+    config = {
+      containers.settings = lib.mkOption {
+        type = toml.type;
+        default = { };
+        description = "containers.conf configuration";
       };
 
-      insecure = lib.mkOption {
-        default = [ ];
-        type = lib.types.listOf lib.types.str;
-        description = ''
-          List of insecure repositories.
-        '';
+      storage.settings = lib.mkOption {
+        type = toml.type;
+        description = "storage.conf configuration";
       };
 
-      block = lib.mkOption {
-        default = [ ];
-        type = lib.types.listOf lib.types.str;
-        description = ''
-          List of blocked repositories.
-        '';
-      };
-    };
+      registries = {
+        search = lib.mkOption {
+          type = lib.types.listOf lib.types.str;
+          default = [ "docker.io" ];
+          description = ''
+            List of repositories to search.
+          '';
+        };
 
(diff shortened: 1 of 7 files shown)


bamhm182:
I like it and it looks mostly good. For some reason ubuntu-latest is failing here, but they seem unrelated and I don't seem to be getting them locally. That said, it does appear that the actual output of registries.conf has an extra `[registries]` section at the top that shouldn't be there.

n-hass:
LGTM. @bamhm182 `registries.conf` default settings gives this, which should be fine
```
[registries.block]
registries = []

[registries.insecure]
registries = []

[registries.search]
registries = ["docker.io"]
```

bamhm182:
I just had a few spare minutes yesterday, so I opted to clone the repo and run the tests while reviewing it and hadn't had time to dig deep into it. I spun up a VM to mess with it, and I can't recreate the issue I was seeing in the test, so I'm thinking there must have just been a glitch when I was running it yesterday. I am running all the tests again on a new fully up to date VM that I am using without issues. Will report back on how that turns out.

dawidd6:
Hmm. Different nixpkgs revisions? I was using latest master for the evaluation using nix-shell.

bamhm182:
Just went through and made sure that everything I could find was up to date. Now my test results match those of the automatic pipeline. Sorry for the confusion!

@n-hass, I'm not totally sure I understand what that suggestion is trying to avoid. Where would we run into clashes with the implementation by dawidd6?

n-hass:
> Where would we run into clashes with the implementation by dawidd6?

Eg having to name `services.podman.containersConf` because there is already `services.podman.containers`. Actual resources vs podman configuration files.

bamhm182:
Ah, yeah. I see what you're saying. I'm also wondering if this belongs under the podman configuration, or if it ought to be under a services.containers like it is for the main nixpkgs. I know right now there's only podman in HM, but rootless docker is technically a thing and uses the same files.

bamhm182:
Went ahead and submitted a PR on @dawidd6's fork [here](https://github.com/dawidd6/home-manager/pull/1). I just put everything under `services.podman.config` as @n-hass had suggested.

dawidd6:
Isn't it a little too deep now? `services.podman.config.storage.settings`, maybe only `services.podman.config.storage` would suffice?

bamhm182:
Yeah. Very fair point. Made the change to the PR I made.

n-hass (approved):

modules/services/podman-linux/default.nix
```diff
@@ -1,17 +1,99 @@
...
+in {
   meta.maintainers = with lib.hm.maintainers; [ bamhm182 n-hass ];
 
   imports =
     [ ./containers.nix ./install-quadlet.nix ./networks.nix ./services.nix ];
 
   options.services.podman = {
     enable = lib.mkEnableOption "Podman, a daemonless container engine";
+
+    config = {
```
rycee, line 14: I think this is probably best renamed to `settings` to avoid the overloaded term `config`. It's not quite an RFC-42 style settings but I think it is close enough.

rycee:
Thanks! Looks good to me in general, just one minor comment.

modules/services/podman-linux/default.nix
dawidd6: Yea, I agree. Done.

rycee:

PR #6042, opened 2024-11-05
Fix git-sync crash when whitespace in path

RedEtherbloom:
### Description

We fixed the home-manager git-sync to properly escape the arguments that get written into the users git-sync-\<name\>.service.

We additionally bumped the nixpkgs-unstable version to include our recently merged update PR for git-sync.
git-sync was 2 years out of date and also had a bug with repo paths containing whitespaces.

### Checklist

- [x] Code tested through `nix-shell --pure tests -A run.all` or `nix develop --ignore-environment .#all` using Flakes.(Excluding the four broken, non-related tests, in #6049)

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@colonelpanic8 @cab404 @ryane

### DIFF
--- a/modules/services/git-sync.nix
+++ b/modules/services/git-sync.nix
@@ -16,9 +16,9 @@ let
         "PATH=${
           lib.makeBinPath (with pkgs; [ openssh git ] ++ repo.extraPackages)
         }"
-        "GIT_SYNC_DIRECTORY=${repo.path}"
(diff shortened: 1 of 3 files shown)


RedEtherbloom:
I sadly found another git-sync bug (soft) blocking this: https://github.com/simonthum/git-sync/issues/33

rycee:

PR #6045, opened 2024-11-05
vivid: add module

arunoruto:
### Description

Add module for https://github.com/sharkdp/vivid.
Previous PRs #2194 and #3705 addressed this, but never merged.

The `themes` and `filetypes` section were taken from #2194.
The shell integrations are used since the variable name `LS_COLORS` needs to be set to the output of the command `vivid generate <theme>`. One could utilize what #2194 used:
```nix
home.sessionVariables = mkIf (cfg.theme != null) {
    LS_COLORS = "$(${cfg.package}/bin/vivid generate ${cfg.theme})";
};
```
but this wouldn't work for `fish` and `nushell`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/vivid.nix
+++ b/modules/programs/vivid.nix
@@ -1,0 +1,95 @@
+{ config, lib, pkgs, ... }:
+let yaml = pkgs.formats.yaml { };
+in with lib; {
+  meta.maintainers = with maintainers; [ arunoruto ];
+
+  options.programs.vivid = {
+    enable = mkEnableOption ''
+      vivid - A themeable LS_COLORS generator with a rich filetype database
+      <link xlink:href="https://github.com/sharkdp/vivid" />
+    '';
+
+    package = mkPackageOption pkgs "vivid" { };
+
+    theme = mkOption {
+      type = with types; nullOr str;
+      default = null;
+      example = "molokai";
+      description = ''
+        Color theme to enable.
+        Run `vivid themes` for a list of available themes.
+      '';
+    };
+
+    filetypes = mkOption {
+      type = yaml.type;
+      default = { };
+      example = literalExpression ''
+        {
+          core = {
+            regular_file = [ "$fi" ];
+            directory = [ "$di" ];
+          };
+          text = {
+            readme = [ "README.md" ];
+            licenses = [ "LICENSE" ];
+          };
+        }
+      '';
+      description = ''
+        Configuration written to
+        <filename>~/.config/vivid/filetypes.yml</filename>.
+        Visit <link xlink:href="https://github.com/sharkdp/vivid/tree/master/config/filetypes.yml" />
+        for a reference file.
+      '';
+    };
+
+    themes = mkOption {
+      type = types.attrsOf (yaml.type);
+      default = { };
+      example = literalExpression ''
+        {
+          mytheme = {
+            colors = {
+              blue = "0000ff";
+            };
+            core = {
+              directory = {
+                foreground = "blue";
+                font-style = "bold";
+              };
+            };
+          };
+        }
+      '';
+      description = ''
+        Theme files written to
+        <filename>~/.config/vivid/themes/<mytheme>.yml</filename>.
+        Visit <link xlink:href="https://github.com/sharkdp/vivid/tree/master/themes" />
+        for references.
+      '';
+    };
+  };
+
+  config = let
+    cfg = config.programs.vivid;
+    lsColors = builtins.readFile (pkgs.runCommand "vivid-ls-colors" { } ''
+      ${lib.getExe cfg.package} generate ${cfg.theme} > $out
+    '');
+  in mkIf cfg.enable {
+    home = {
+      packages = [ cfg.package ];
+      sessionVariables = { LS_COLORS = "${lsColors}"; };
+    };
+
+    xdg.configFile = {
+      "vivid/filetypes.yml" =
+        mkIf (builtins.length (builtins.attrNames cfg.filetypes) > 0) {
+          source = yaml.generate "filetypes.yml" cfg.filetypes;
+        };
+    } // mapAttrs' (name: value:
+      nameValuePair "vivid/themes/${name}.yml" {
+        source = yaml.generate "${name}.yml" value;
+      }) cfg.themes;
+  };
+}
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -264,6 +264,7 @@ let
     ./programs/vim-vint.nix
     ./programs/vim.nix
     ./programs/vinegar.nix
+    ./programs/vivid.nix
     ./programs/vscode.nix
     ./programs/vscode/haskell.nix
     ./programs/pywal.nix
--- a/tests/default.nix
+++ b/tests/default.nix
@@ -246,6 +246,7 @@ in import nmtSrc {
     ./modules/programs/vifm
     ./modules/programs/vim-vint
     ./modules/programs/vinegar
+    ./modules/programs/vivid
     ./modules/programs/vscode
     ./modules/programs/watson
     ./modules/programs/wezterm
--- a/tests/modules/programs/vivid/default.nix
+++ b/tests/modules/programs/vivid/default.nix
@@ -1,0 +1,5 @@
+{
+  vivid-empty-settings = ./empty.nix;
+  vivid-filetypes = ./filetypes-example.nix;
+  vivid-themes = ./themes-example.nix;
+}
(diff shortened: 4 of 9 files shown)


modules/programs/vivid.nix
```diff
@@ -0,0 +1,141 @@
+{ config, lib, pkgs, ... }:
+let
+  yaml = pkgs.formats.yaml { };
+  bashLine = theme:
+    ''export LS_COLORS="$(${lib.getExe pkgs.vivid} generate ${theme})"'';
+  fishLine = theme:
+    "set -Ux LS_COLORS $(${lib.getExe pkgs.vivid} generate ${theme})";
```
folliehiyuki, line 7:
```suggestion
    "set -gx LS_COLORS (${lib.getExe pkgs.vivid} generate ${theme})";
```

I'd prefer `set -gx`, as the line is loaded on each shell session anyway. Also, please remove the `$` sign, as it's not "native" fish syntax (it's supported only for compatibility).

arunoruto: Thanks for the heads up! I changed the flag and removed the `$` sign before the command.

arunoruto:
@folliehiyuki Hej, I just wanted to check if the PR is alright. It is my first in the `hm` repo, so feedback is always appreciated!

modules/programs/vivid.nix
```diff
@@ -0,0 +1,141 @@
+{ config, lib, pkgs, ... }:
+let
+  yaml = pkgs.formats.yaml { };
+  bashLine = theme:
+    ''export LS_COLORS="$(${lib.getExe pkgs.vivid} generate ${theme})"'';
+  fishLine = theme:
+    "set -gx LS_COLORS (${lib.getExe pkgs.vivid} generate ${theme})";
+  nushellLine = theme: "${lib.getExe pkgs.vivid} generate ${theme}";
+  zshLine = bashLine;
+in {
```
folliehiyuki, line 10: You can put a `with lib;` declaration here, so usages of `lib` afterward can be shorten.

modules/programs/vivid.nix
```diff
@@ -0,0 +1,141 @@
...
+      default = true;
+      type = lib.types.bool;
+      description = ''
+        Whether to enable Zsh integration.
+        Adds `${zshLine "<theme>"}` to Zsh.
+      '';
+    };
+
+    theme = lib.mkOption {
+      type = lib.types.nullOr lib.types.str;
```
folliehiyuki, line 58:
```suggestion
      type = with types; nullOr str;
```

assuming that `with lib;` was declared above.

folliehiyuki:
Can you also write a couple of assert tests for `filetypes.yml` and a sample theme file?

modules/programs/vivid.nix
folliehiyuki: The same goes for the other options as well.

arunoruto: I have rewritten the module with `with lib;`.
I checked it multiple times, so there shouldn't be anything missing.

arunoruto:
> Can you also write a couple of assert tests for `filetypes.yml`

Do you have any examples on how to write those tests? I am not that experienced.

> and a sample theme file?

In Nix and put it in the `literalExample` field?

folliehiyuki:
You can look into https://github.com/nix-community/home-manager/tree/master/tests/modules/programs for referenced assert test cases and how to write them.

> > and a sample theme file?
>
> In Nix and put it in the literalExample field?

I mean defining them in an assert test. You write the sample theme inside Nix and assert the generated theme file with expected value.

arunoruto:
> You can look into https://github.com/nix-community/home-manager/tree/master/tests/modules/programs for referenced assert test cases and how to write them.
> 
> > > and a sample theme file?
> > 
> > 
> > In Nix and put it in the literalExample field?
> 
> I mean defining them in an assert test. You write the sample theme inside Nix and assert the generated theme file with expected value.

I looked into the new ghostty PR and took that init as a base to design the tests (plus some other modules as you suggested!). All of the tests pass on my device, except the `enable-shells.nix` one. I guess it can't replace `@vivid@` with the correct path? Or am I doing something wrong here...

I also went a bit overboard with the config, but I wanted to see if it works correctly! Hit me up if I need to reduce it :)

arunoruto:
@folliehiyuki sorry for the tag, but I wanted finish this PR so it can be integrated until the `25.05` release.

arunoruto:
@rycee Hi, sorry for the ping, but I wanted to ask if someone could review the PR in march/april :) maybe you know someone, who has time on their hands, since Hoang seems to be busy.

arunoruto:
@sedlund Thanks for the feedback!

What would you recommend to focus the testing on? I am still new in that regard to nixos and home-manager.

I just implemented the recommended changes using `pkgs.runCommand` and it worked, albeit needing a restart of the current user session. I implement the changes upstream when I am at home.

Just tried to do it at work and nuked my branch with a reset... But I should still have the files on my laptop at home! Afterwards I will reopen the PR or create a new one if I can't reopen this one!

arunoruto:
Reopening it

arunoruto:
Sorry for the spam!
I just recovered my branch and pushed the suggested changes :)

As for the tests, I would refactor them if you have some suggestions or even an example. Should I just make the files smaller, like a minimal working example, and use that as a test?

modules/programs/vivid.nix
```diff
@@ -0,0 +1,95 @@
+{ config, lib, pkgs, ... }:
+let yaml = pkgs.formats.yaml { };
+in with lib; {
+  meta.maintainers = with maintainers; [ arunoruto ];
+
+  options.programs.vivid = {
+    enable = mkEnableOption ''
+      vivid - A themeable LS_COLORS generator with a rich filetype datebase
```
sedlund, line 8: *database

arunoruto: Done :)

sedlund:
> Should I just make the files smaller, like a minimal working example, and use that as a test?

My thought is that the current test is only testing if `lib.yaml.generate` generates yaml correctly - which should be tested upstream.

I tested creating a small `filetypes.yml` and a theme that references a filetype that does not exist in the filetypes.yml and running:

```sh
vivid generate mytheme
```
this tests the `filetypes.yml` file and theme yml as a working set together verifies and correctly tells me:

```cli
Error: Could not find style for category 'programming.source.cxx'
```

and exits 1 

This would be output now during build with your recent changes -  and I think the output would be more useful.

Nothing stands out to me what more could be tested.

Anomalocaridid:
> I implement it like this:
> 
> ```nix
> sessionVariables =
>   let
>     lsColors = builtins.readFile (
>       pkgs.runCommand "vivid-ls-colors" { } ''
>         ${lib.getExe pkgs.vivid} generate dracula > $out
>       ''
>     );
>   in
>   {
>     LS_COLORS = "${lsColors}";
>   }
> ```
> 
> this allows nix build to run the vivid command to output LS_COLORS and stores it. it runs once on the build machine, not requiring the vivid binary to be copied to each host and to be run each time a login shell is executed.
> 
> also this uses `sessionVariables` so it is shell agnostic. you could do the same and get rid of the extra shell specific options.
> 
> hope this helps

Not a reviewer either, but wouldn't this cause unnecessary [import-from-derivation (IFD)](https://nix.dev/manual/nix/2.26/language/import-from-derivation)? This would slow down evaluation and make the module unusable for people that have `allow-import-from-derivation` set to `false`.

arunoruto:
I guess this means I will revert it back to the shell options?
This is usually how most shell programs import, like direnv with its init function. And vivid is implemented in rust, to the bottleneck shouldn't be there, I hope.
Either way, maybe the upstream project could implement a caching function, where the output could be stored in `~/.cache` and loaded from there, if a theme exists.

sedlund:
I really don't think IFD is an issue here at all.  It is good to be aware of when doing it for large evals (package builds, etc).

But in this context it only blocks eval when the vivid derivation or the config file changes.  The eval takes 3ms on my machine. The time it would take to distribute the binary to various machines would take far longer.  So the doing this is still preferred computationally and network wise.

I'd be happy to hear why this would not be true.

sedlund:
deleted that suggestion, it was not correct

tuxiqae:
Vivid support was added here https://github.com/nix-community/home-manager/pull/7772
Do we still need this PR open?

arunoruto:
> Vivid support was added here #7772 Do we still need this PR open?

Thanks for the update! I will close this PR now :)
P.S. maybe look at what the comments about testing in this PR, since it was the main reason it got delayed so far...

tuxiqae:
FYI I took no part in #7772, I just noticed that it was merged and decided to notify you :)

PR #6046, opened 2024-11-06
ssh: Allow forwardAgent to be set to null

tebriel:
### Description

Allow setting the forwardAgent to `null` to prevent it from being set in `Host *`.

In situations where you use `CanonicalizeHostname` and `CanonicalDomains` the ssh config is parsed twice. THe first time a value is set, typically means it's the final value. This means that if `programs.ssh.forwardAgent` is set to `false` you cannot have `ForwardAgent = yes` apply to a canonical match (`Match canonical *.my.tld`) as `Host *` will be parsed first.

If we consider the config where `programs.ssh.forwardAgent = false;`:

```ssh_config
CanonicalizeHostanme yes
CanonicalizeDomains my.tld
CanonicalizeMaxDots 1

Match canonical Host *.my.tld
  ForwardAgent yes

Host *
  ForwardAgent no
```

`ssh myhost` will first pass through `Host *` and set `ForwardAgent no`, canonicalize the name to myhost.my.tld, then match `Match canonical *.my.tld` and try to set `ForwardAgent` to `yes` but since it's already `no` it will be ignored.

The same is a problem in reverse if `programs.ssh.forwardAgent = true;` and the canonical match is `ForwardAgent no`.

The default for `forwardAgent` is already `false` so there's not a need to always explicitly set it, we allow `forwardAgent = null` in match blocks, we can set it to `null` in the root as well.

#### Note

This does add a blank line if `forwardAgent` is set to null in the `Host *` block. Mabye not ideal, can probably fix it if it's really a problem.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/ssh.nix
+++ b/modules/programs/ssh.nix
@@ -351,7 +351,7 @@ in {
 
     forwardAgent = mkOption {
       default = false;
-      type = types.bool;
+      type = types.nullOr types.bool;
       description = ''
         Whether the connection to the authentication agent (if any)
         will be forwarded to the remote machine.
@@ -533,7 +533,10 @@ in {
         '') ++ (map (block: matchBlockStr block.name block.data) matchBlocks))}
 
       Host *
-        ForwardAgent ${lib.hm.booleans.yesNo cfg.forwardAgent}
+        ${
+          optionalString (cfg.forwardAgent != null)
+          "ForwardAgent ${lib.hm.booleans.yesNo cfg.forwardAgent}"
+        }
         AddKeysToAgent ${cfg.addKeysToAgent}
         Compression ${lib.hm.booleans.yesNo cfg.compression}
         ServerAliveInterval ${toString cfg.serverAliveInterval}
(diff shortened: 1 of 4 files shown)


modules/programs/ssh.nix
```diff
@@ -533,7 +533,10 @@ in {
         '') ++ (map (block: matchBlockStr block.name block.data) matchBlocks))}
 
       Host *
-        ForwardAgent ${lib.hm.booleans.yesNo cfg.forwardAgent}
+        ${
```
teto, line 536: this is why I like freeform settings: we could have just removed the nix option instead. Is this something you would be interested in impleementing ? 

Now that it exists it makes sense to allow a "null" but with the current implementation it adds an empty line (a recurrent issue throughout nixpkgs). Could you fix that ?

teto: you can have a look at the sshd module for instance https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/networking/ssh/sshd.nix

tebriel:
too tired to try to continue to work to get more reviews/help

Issue #3965, opened 2023-05-08
Recommendation regarding zsh compinit


nh2:
Hi, a hit an issue with `zsh` that starting a new shell (by opening a new terminal, typing `zsh`, or benchmarking with `command time zsh -i -l -c ':'`) took ~300 ms, causing a noticeable, annoying delay.

Using `zprof` (e.g. as described [here](https://htr3n.github.io/2018/07/faster-zsh/)) showed that `compinit` and `compaudit` took > 85% of that time, and that they were called repeatedly (3 times for me). Both are related to `zsh`'s completion system.

After a bit of reasearch, e.g. mainly https://dev.to/djmoch/zsh-compinit--rtfm-47kg, I learned that during `zsh` startup:

* You need to initialise the `zsh` completion system once by calling `compinit`
* By default (when not calling it with `-C`), it also creates a cache file `.zcompdump`; this takes extra long
* Any time you add some completions afterwards during zsh startup (e.g. via `home-manager`'s plugins for other shell-integrating software like `atuin`, `fzf`, or `skim`) which modifies the `fpath` (if I understood it right), zsh automatitically has to re-run `compinit`.

Here is my concrete setup:

* Using global NixOS config in `/etc/nixos/configuration.nix` (where I load the `grml` zsh config)
* A bit of extra user-specific config in home-manager's `home.nix`

Because of that, I get `compinit` calls in:

* GRML's [own config](https://github.com/grml/grml-etc-core/blob/74a8efd0172e2468ee055a019f14706276758c7d/etc/zsh/zshrc#L732-L743) that I include (and which apparently cannot easily be turned off)
* the global NixOS `zsh` config (NixOS [module setting](https://github.com/NixOS/nixpkgs/blob/fd40cef8d797670e203a27a91e4b8e6decf0b90c/nixos/modules/programs/zsh/zsh.nix#L149-L160), [code that does the call](https://github.com/NixOS/nixpkgs/blob/fd40cef8d797670e203a27a91e4b8e6decf0b90c/nixos/modules/programs/zsh/zsh.nix#L251-L254))
* home-manager `zsh` module ([module setting](https://github.com/nix-community/home-manager/blob/78ceec68f29ed56d6118617e9f0f588bf164067f/modules/programs/zsh.nix#L290-L306), [code that does the call](https://github.com/nix-community/home-manager/blob/78ceec68f29ed56d6118617e9f0f588bf164067f/modules/programs/zsh.nix#L519-L524))

[...]

PR #6063, opened 2024-11-10
zsh: fix completion when zplug is enabled

taha-yassine:
### Description

`zplug` already calls `compinit` so there is no need to explicitly call it when it's enabled. It was also brought up in https://github.com/nix-community/home-manager/issues/3965#issuecomment-1742207733.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@ThinkChaos

### DIFF
--- a/modules/programs/zsh.nix
+++ b/modules/programs/zsh.nix
@@ -632,10 +632,10 @@ in
         '') cfg.plugins))
 
         ''
-        # Oh-My-Zsh/Prezto calls compinit during initialization,
+        # Oh-My-Zsh/Prezto/zplug calls compinit during initialization,
         # calling it twice causes slight start up slowdown
         # as all $fpath entries will be traversed again.
-        ${optionalString (cfg.enableCompletion && !cfg.oh-my-zsh.enable && !cfg.prezto.enable)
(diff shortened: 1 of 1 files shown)


ThinkChaos (approved):

Adda0 (approved):

teto (commented):
I am not sure I liked having options autodisable other options.

Maybe having an assert instead or the default of enableCompletion depend on oh-my-zsh.enable and   prezto.enable ?

taha-yassine:
> I am not sure I liked having options autodisable other options.
> 
> Maybe having an assert instead or the default of enableCompletion depend on oh-my-zsh.enable and prezto.enable ?

Strictly speaking the option is indeed disabled, but it's just to avoid initializing completion twice since the listed modules do that anyway.

teto:
I dont like that we have all these ZSH frameworks intertwined with the ZSH module. each framework should live in its own file and should configure things through programs.zsh without custom entrypoints ideally.

khaneliman:
> I dont like that we have all these ZSH frameworks intertwined with the ZSH module. each framework should live in its own file and should configure things through programs.zsh without custom entrypoints ideally.

Yeah, I think this module could still use some more refactoring, but we can at least fix this issue in the current form.

PR #6064, opened 2024-11-10
nix: make sure extra- settings are applied last

alois31:
All extra- settings should come after their unprefixed counterpart, because the entries of the config file are processed from top to bottom and appending should happen after overwriting.

This is a port of https://github.com/NixOS/nixpkgs/pull/278064 .

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


teto:
I hit this a few years ago so I understand the need or this. Thanks for referencing the nixpkgs PR. Could you rebase to fix ci

teto (approved):

alois31:
Locally the test fails now because of nixpkgs and home-manager version mismatch, let's hope the nixpkgs in CI is too old for that.

ambroisie:
@alois31 would you like to rebase the change?

PR #6072, opened 2024-11-12
mopidy: restart service on config changes

jwillikers:
### Description

The Mopidy service is not restarted when the config changes.
This PR adds `X-Restart-Triggers` to the unit to ensure that the service is restarted on config file changes.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@foo-dogsquared


rycee:
Thanks! I extended the test a bit and merged to master now 🙂

PR #6074, opened 2024-11-13
feh: add themes option

gmvar:
### Description
- added themes option
- added themes test
- updated broken man page links

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
No `meta.maintainers` list for this module, so @-ing those who have last touched it.
@sumnerevans 
@ncfavier 
@rycee

### DIFF
--- a/modules/programs/feh.nix
+++ b/modules/programs/feh.nix
@@ -8,6 +8,12 @@ let
 
   bindingsOf = t: with types; attrsOf (nullOr (either t (listOf t)));
 
+  renderThemes = options:
+    let
+      render =
+        mapAttrsToList (theme: options: "${theme} ${escapeShellArgs options}");
+    in concatStringsSep "\n" (render options);
+
   renderBindings = bindings:
     let
       enabled = filterAttrs (n: v: v != null) bindings;
@@ -41,7 +47,7 @@ in {
         Override feh's default mouse button mapping. If you want to disable an
         action, set its value to null. If you want to bind multiple buttons to
         an action, set its value to a list.
-        See <https://man.finalrewind.org/1/feh/#x425554544f4e53> for
+        See <https://man.finalrewind.org/1/feh/#BUTTONS_CONFIG_SYNTAX> for
         default bindings and available commands.
       '';
     };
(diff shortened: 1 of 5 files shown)


modules/programs/feh.nix
```diff
@@ -79,5 +96,8 @@ in {
     xdg.configFile."feh/keys" = mkIf (cfg.keybindings != { }) {
       text = renderBindings cfg.keybindings + "\n";
     };
+
+    xdg.configFile."feh/themes" =
+      mkIf (cfg.themes != { }) { text = renderBindings cfg.themes + "\n"; };
```
ncfavier, line 101: Why use `renderBindings`? Themes don't use the same syntax. Options should be escaped with `escapeShellArgs`.

gmvar: Themes use the syntax `theme options ...`, which is similar to key bindings which are `action key1 key2 ...`

From [fehs theme config syntax man page](https://man.finalrewind.org/1/feh/#THEMES_CONFIG_SYNTAX):
> Note that the option parser does not behave like a normal shell: filename expansion and backslash escape sequences are not supported and passed to feh's option parser as-is. However, quoting of arguments is respected and can be used for arguments with whitespace.

I believe `escapeShellArgs` would be incorrect here.
> nix-repl> lib.strings.escapeShellArgs ["-rVq --thumb-width 40 --thumb-height 30 --index-info '%n\\\\n%wx%h'"]
> "'-rVq --thumb-width 40 --thumb-height 30 --index-info '\\\\''%n\\\\n%wx%h'\\\\'''"

You can see the unnecessary escape sequences in the output.

- - - 
> (And your example should be "--slideshow-delay" "20".)

I did that on purpose to make it clear that the option can be set with a `string` or `list of string`.
`"--slideshow-delay" "20"` and `"--slideshow-delay 20"` produce the same result, it is just a stylistic choice to keep the key value pair together in the same quote.

ncfavier: You would only use `escapeShellArgs` if the input is a list of strings, not a string. If we accept lists of strings then the usual convention is that those strings represent individual arguments, and are not just concatenated with whitespace. Otherwise you can't have whitespace in an argument without double-quoting, which is silly.

gmvar: Ah, I see now. 

I will change to only accept a list of strings and will use `escapeShellArgs`.

modules/programs/feh.nix
```diff
@@ -8,6 +8,13 @@ let
 
   bindingsOf = t: with types; attrsOf (nullOr (either t (listOf t)));
 
+  renderThemes = options:
+    let render = mapAttrsToList renderTheme;
+    in concatStringsSep "\n" (render options);
+
+  renderTheme = func: key:
+    if key == null then func else "${func} ${escapeShellArgs key}";
```
ncfavier, line 16: How could `key` be `null` here?

gmvar: Fixed.

ncfavier (approved):

PR #6077, opened 2024-11-13
uv: init module

### LINKED ISSUE #5154: programs.uv
### Description

Would be cool to have `programs.uv` for the new python dev tool [uv](https://github.com/astral-sh/uv) (with all of the available `enable*Integration`s)

The number of stars is probably mostly carry-over hype from `ruff`, but it does have a lot going for it.

If I wasn't so busy atm, I'd try writing my first home-manager module for this, but I figured it'd at least be worth creating an issue for

mirkolenz:
### Description

Fixes #5154 

Adds a new module for the python package manager uv. Includes support for writing settings to a config file and enabling shell integrations for bash/zsh/fish. I also added multiple tests (based on those for `atuin`).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/uv.nix
+++ b/modules/programs/uv.nix
@@ -1,0 +1,56 @@
+{
+  pkgs,
+  config,
+  lib,
+  ...
+}:
+
+let
+
+  inherit (lib)
+    mkEnableOption
+    mkPackageOption
+    mkOption
+    literalExpression
+    ;
+
+  tomlFormat = pkgs.formats.toml { };
+  cfg = config.programs.uv;
+
+in
+{
+  meta.maintainers = with lib.maintainers; [ mirkolenz ];
+
+  options.programs.uv = {
+    enable = mkEnableOption "uv";
+
+    package = mkPackageOption pkgs "uv" { };
+
+    settings = mkOption {
+      type = tomlFormat.type;
+      default = { };
+      example = literalExpression ''
+        {
+          python-downloads = "never";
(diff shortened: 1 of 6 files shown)


modules/programs/uv.nix
```diff
@@ -0,0 +1,74 @@
...
+
+    programs.zsh.initExtra = lib.mkIf cfg.enableZshIntegration ''
+      eval "$(${lib.getExe cfg.package} generate-shell-completion zsh)"
+      eval "$(${lib.getExe' cfg.package "uvx"} --generate-shell-completion zsh)"
+    '';
+
+    programs.fish.interactiveShellInit = lib.mkIf cfg.enableFishIntegration ''
+      ${lib.getExe cfg.package} generate-shell-completion fish | source
+      ${lib.getExe' cfg.package "uvx"} --generate-shell-completion fish | source
+    '';
```
ambroisie, line 72: The shell completion should be managed by the package, _not_ home-manager.

In fact, it [already does](https://github.com/NixOS/nixpkgs/blob/dc460ec76cbff0e66e269457d7b728432263166c/pkgs/by-name/uv/uv/package.nix#L50-L56).

mirkolenz: Thanks, didn't know that. I've removed the integration options from the module.

mirkolenz:
@rycee Is there anything I can do to move this forward? Thanks in advance 😄

Issue #4659, opened 2023-11-13
bug: zellij: It is not clear how to configure keyboard bindings


brianmay:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

From the docs zellij keyboard bindings look like:

```
    // bind the Alt-n to open a new pane
    bind "Alt n" { NewPane; }
    // bind both the "h" key and the left-arrow key to move pane focus left
    bind "h" "Left" { MoveFocus "Left"; }
    // bind the "f" key to toggle the focused pane full-screen and switch to normal mode
    bind "f" { ToggleFocusFullscreen; SwitchToMode "Normal"; }
```

But it is far from clear what I have to set the `settings` to to generate this. And I can't see any documentation either.

I have seen bug #4054, but this hasn't really helped.

### Maintainer CC

@mainr

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.1.61, NixOS, 23.05 (Stoat), 23.05.20231105.aeefe20`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.13.6`
 - channels(brian): `""`
 - channels(root): `"nixos-23.05"`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`
```

brianmay:
After writing this I found #4160, looks like there isn't a solution yet. But when the solution is found, it should be clearly documented...

brianmay:
Oh, that explains my confusion. I wasn't aware zellij changed formats.

pluiedev:
Yeah they did, in [version 0.32.0](https://github.com/zellij-org/zellij/releases/tag/v0.32.0), although we should probably document this in the module somewhere as it can break things, and users should be aware of that

gotcha:
I have added documentation: https://github.com/nix-community/home-manager/pull/4665

Feedback welcome !

brianmay:
I think this is still a concern. There is a proposal in #4465 to resolve, but looks like that has stalled too :-(.

b-m-f:
Still an issue

pluiedev:
Still a thing.

Dietr1ch:
Can we get a textual `extraSettings` out to allow configuring things?

I guess that figuring out the best way to transform nix values into kdl configs in a more generic way is going to take a while.

[...]

PR #6078, opened 2024-11-13
Add extraConfig to programs.zellij

Dietr1ch:
### Description

This allows working around [#4659](https://github.com/nix-community/home-manager/issues/4659)
While fixing `toKDL` might be fun, I just don't think it's a good use of anyone's time.

This is still untested as I'm not yet familiar with testing changes here.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@mainrs


Dietr1ch:
Tested locally after tracking my own fork.

```nix
  inputs = {
    nixpkgs = {
      url = "github:nixos/nixpkgs/nixos-unstable";
    };
    home-manager = {
      url = "github:Dietr1ch/home-manager/test";
      inputs.nixpkgs.follows = "nixpkgs";
    };
  };
```


```kdl
/// (Config added by regular programs.zellij.settings)
default_layout "compact"
default_mode "locked"
theme "nord"

// extraConfig

keybinds {
    // keybinds are divided into modes
    normal {
        // bind instructions can include one or more keys (both keys will be bound separately)
        // bind keys can include one or more actions (all actions will be performed with no sequential guarantees)
        bind "Ctrl g" { SwitchToMode "locked"; }
        bind "Ctrl p" { SwitchToMode "pane"; }
        bind "Alt n" { NewPane; }
        bind "Alt h" "Alt Left" { MoveFocusOrTab "Left"; }
    }
    pane {
        bind "h" "Left" { MoveFocus "Left"; }
        bind "l" "Right" { MoveFocus "Right"; }
        bind "j" "Down" { MoveFocus "Down"; }
        bind "k" "Up" { MoveFocus "Up"; }
        bind "p" { SwitchFocus; }
    }
    locked {
        bind "Ctrl g" { SwitchToMode "normal"; }
    }
}
```

Dietr1ch:
I'd not mind adding a fixed output test, but I'm not sure exactly where it should live.

Are there any examples in the repo I can mimic and learn a bit?

Adda0:
Thanks for the PR. I will definitely investigate it later.

> I'd not mind adding a fixed output test, but I'm not sure exactly where it should live.

Just a non-authoritative suggestion: Have a look at #6037. You can extend the test file similarly to what I did. I believe that is what you want in this case.

mainrs:
Honestly, this seems like the best approach right now. This would at least make it _work_ in all use cases and having a working `toKDL` function can be a goal for the future.

mainrs (approved):
Although the PR is more of a fix for the intermediate future, I think it's better than having a non-working module that relies on explicit `home.file` calls by the user. We've seen multiple PRs now that try to fix the `toKDL` function. The last two ones have not received any responses from their creators. I also don't use zellij anymore and neither have the motivation nor time to actively work on the PRs to make them work. The generator requires some good testing, since there are a bunch of serialization edge cases.

I'd rather have this merged in the short term and remove it once `toKDL` works properly.

LGTM.

Dietr1ch:
This went stale. Is the "short-term" fix of just offering `extraConfig` good enough to be merged?

I think I'll end up fixing this merge conflict anyways to keep my systems updated, but would love to avoid dealing with it in the future.

Dietr1ch:
Updated again, got tripped by new formatting changes in between rebases, so the first CI run failed.

tests/modules/programs/zellij/config.nix
```diff
@@ -0,0 +1,37 @@
...
+{
+  programs = {
+    zellij = {
+      enable = true;
+
+      settings = {
+        default_layout = "welcome";
+      };
+      extraConfig = ''
+        This_could_have_been_json {
```
mainrs, line 12: I do love this haha!

Dietr1ch:
I rebased into a fresh master again since some unrelated tests were failing

Dietr1ch:
Hey, I've been running this patch for almost a year now, but I'd like to see this merged. It's a trivial escape hatch to config woes that should be in even if `toKDL` worked flawlessly in case KDL gets any extension that would require updates to `toKDL` to even try out.

modules/programs/zellij.nix
```diff
@@ -248,7 +281,16 @@ in
...
               {
-                text = lib.hm.generators.toKDL { } cfg.settings;
+                text =
+                  (lib.hm.generators.toKDL { } cfg.settings)
+                  + lib.optionalString (cfg.extraConfig != "") (
+                    ''
+
+                      // extraConfig
+
+                    ''
```
khaneliman, line 291: Not sure we really need a comment explaining which option the config came from.

khaneliman (commented):
I've been changing my mind on `extraConfig` options recently... I think it's good as a way for people to configure in normal text instead of being forced to use a generator if the `nix` config way is too complex to configure through a generator.

Dietr1ch:
I guess we don't, it's just that in case there's conflicts between the generated config and the "extra" config, it'll be a bit more obvious that the bottom half of the config comes from the "extra" part.

I've found having this reminder helpful in cases where the generated config grows as the nix module grows and starts configuring something I had configured myself in the "extra" section before. This makes the weird conflict-related error that points me into the config file easier to debug.

I don't feel too strongly about having to keep it in the module. It's just something that in my mind may be useful and I'll do on my end anyway as I also like including the paths to the relevant files and docs.

modules/programs/zellij.nix
Dietr1ch: I guess we don't, it's just that in case there's conflicts between the generated config and the "extra" config, it'll be a bit more obvious that the bottom half of the config comes from the "extra" part.

I've found having this reminder helpful in cases where the generated config grows as the nix module grows and starts configuring something I had configured myself in the "extra" section before. This makes the weird conflict-related error that points me into the config file easier to debug.

I don't feel too strongly about having to keep it in the module. It's just something that in my mind may be useful and I'll do on my end anyway as I also like including the paths to the relevant files and docs.

PR #6089, opened 2024-11-16
zsh: path handling - remove `$HOME` reliance & make relative/absolute path logic consistent

### LINKED ISSUE #5100: bug: home manager zsh option "dotDir" is mangling zdotdir path and causing ".config/zsh" to be created in any directory the user is in

### Issue description

Im having the exact same issue as this issue here [#219](https://github.com/nix-community/home-manager/issues/219) and the suggested solution brings an even more annoying issue where ".config/zsh" is created in literally any directory relative to the current.

So for if example if the current working directory is ```/etc/nixos``` home-manager creates ```/etc/nixos/.config/zsh/...``` or if I am in ~/.config/home-manager and open a terminal, it creates ~/.config/home-manager/.config/zsh. So far the only other file it creates is the zsh history file.

Its very annoying. The problem is that I can literally only set programs.zsh.dotDir to ".config/zsh" and cant use "${config.xdg.configHome}"/zsh or even an absolute path like "/home/xxx/.config/zsh" because "$HOME" is prepended to dotDir and the .zshenv in the home directory (which sources a .zshenv in dotDir)

so it looks like this "$HOME/'.config/zsh'" (with ".config/zsh")
and it looks like when using any absolute path like the xdg option: "$HOME//home/xxx/.config/zsh"
So for some reason in certain cases $HOME isn't being evaluated and it creates the directory structure relative to whatever directory the user is in.

the solution I would guess is to allow the user to specify the full path and not prepend '$HOME' to the .zshenv file that is symlinked to home

### Maintainer CC

_No response_

### System information

```markdown
❯ nix-shell -p nix-info --run "nix-info -m"
 - system: `"x86_64-linux"`
 - host os: `Linux 6.6.19, NixOS, 24.05 (Uakari), 24.05.20240303.b8697e5`
 - multi-user?: `yes`
 - sandbox: `yes`
 - channels(root): `"nixos, nixos-unstable"`
```
```

### PR DESCRIPTION
**Pull request by kjkent**
### Description

Previously, `config.programs.zsh.dotDir` prepended strings with `$HOME`. This caused issues where `$HOME` is inconsistently resolved in time for the evaluation of the option. The handling of this variable is also inconsistent with how paths are handled elsewhere, including within the same module, where `config.programs.zsh.history.path` does not mutate the supplied string.

This change prepends `config.home.homeDirectory` to relative paths, while assigning absolute paths unchanged. Tests for both cases are added.

Closes #5100 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### CODE REVIEW

File: modules/programs/zsh.nix [hunk 1]
```diff
@@ -22,7 +21,10 @@ let
     mapAttrsToList (k: v: ''hash -d ${k}="${v}"'') cfg.dirHashes
   );
 
-  zdotdir = "$HOME/" + lib.escapeShellArg cfg.dotDir;
+  # Absolute paths are assigned unmutated, relative paths
+  # are prepended with the user's home directory.
+  zdotdir = (strings.optionalString (!strings.hasPrefix "/" cfg.dotDir)
+    config.home.homeDirectory) + escapeShellArg cfg.dotDir;
```
teto, line 27: I had `dotDir = ".config/zsh";`  in my config and I think it generated a ZDOTDIR value of
 
```
echo $ZDOTDIR
/home/teto.config/zsh
```

kjkent: @ thanks for raising this -- looking at the code, yeah, I can see where I missed that. Curiously, it didn't cause the test to fail, which is doubly concerning -- let me made a change and test and I'll push an update :+1:

rycee:
Looks like the CI is a bit unhappy with this 🙂 But perhaps not far off from working?

kjkent:
@rycee yeah -- apologies for the delay on this; I started, then realised there was more to the fix than I anticipated and life got in the way a bit. I'm actually doing nothing right now so will see if I can get it boxed off :)

kjkent:
@rycee @teto I believe this is good to go. Thanks for your patience with it.

I've been descriptive in the commit messages, but, in addition to original PR description I found that when setting `HISTFILE` in `.zshrc`, `$HOME` was also prepended to any custom `programs.zsh.history.path` value, if state version was < 20.03 (newer state versions only work with absolute paths). This wasn't conveyed in the description strings. Additionally, station versions >= 20.03 had a default path of the home dir while older versions defaulted to ZDOTDIR, which is (in my own view) more logical -- if settings are left at their defaults, it's the same behavior, but if a user makes the choice to move their ZDOTDIR out of home, then it follows that the histfile should move with.

The first commit removes all reliance on `$HOME` which should fix the race condition issues seen, and allows both options to accept relative paths (which are interpreted as relative to `home.homeDirectory`) and absolute paths. The second commit simplifies the history file path handling + restores the sane, pre-20.03 default location.

If the second commit isn't desired I can revert it, otherwise, it poses a minor breaking change to those who have set a custom ZDOTDIR but not a custom HISTFILE location. I'm unsure of the best way to communicate/mitigate this (via state version logic?), though I'd wager those installs are few -- why go to the effort of moving _most_ .z* files out of `~/`, but leave one?

khaneliman:
Curious if you're still interested in pursuing this? We've had a bit of refactoring since this was made.

khaneliman:
Rebased on master and resolved all the merge conflicts and fixed some issues that were caught in newer tests.

andrevmatos:
Just FYI, this broke usecases where users had environment variables in `programs.zsh.history.path`; e.g. I used to have `\${XDG_STATE_HOME:-$HOME/.local/state}/zsh/history` in there, but this now adds some `'` single quotes plus wrong variable expansion inside, which end up creating a path like `/home/user/\'/home/user/.local/state/zsh/history\'`
Fix is to use `${config.xdg.stateHome}` (or other xdg path) interpolated directly

gepbird:
Same, after updating with `programs.zsh.history.path = "$ZDOTDIR/.zsh_history";` (which now renders to `HISTFILE="'/home/gep/$ZDOTDIR/.zsh_history'"`), this directory structure gets creates when running `zsh`:

```
~/foo ❯ find
.
~/foo ❯ zsh
~/foo ❯ find
.
./'
./'/home
./'/home/gep
./'/home/gep/home
./'/home/gep/home/gep
./'/home/gep/home/gep/.config
./'/home/gep/home/gep/.config/zsh
./'/home/gep/home/gep/.config/zsh/.zsh_history'
```

I agree with this simpler handling, but some warning, release note or hiding it behind a new `stateVersion` would be nice.

khaneliman:
@andrevmatos @gepbird thanks for identifying those issues, created https://github.com/nix-community/home-manager/pull/7561 to handle that better. Would be helpful to test with your previous config to verify it works, as expected.

lzpreslav:
JFYI a trailing slash in my homeDirectory path caused a zsh initialization failure.

My nix-darwin configuration:
```
users.users."${currentSystemUser}".home = "/Users/${currentSystemUser}/";
```
After updating and rebuilding I got this `.zshenv` file which failed zsh initialization with "job table full or recursion limit exceeded":
```
❯ cat ~/.zshenv        
source /Users/lzpreslav/.zshenv
```
Removing the trailing slash from the path fixed the issue.

pecigonzalo:
I was just "surprised" by the changes in this PR and my files now no longer respecting `xdg.enable`. I believe this is a regresion in the behaviour to be honest, and its inconsistent with other shells.
I do understand wanting to get away from inconsistent behaviours, but that should not have meant in my opinion, to no longer respect `xdg.enable` like we do in other shells.

khaneliman:
> I was just "surprised" by the changes in this PR and my files now no longer respecting `xdg.enable`. I believe this is a regresion in the behaviour to be honest, and its inconsistent with other shells. I do understand wanting to get away from inconsistent behaviours, but that should not have meant in my opinion, to no longer respect `xdg.enable` like we do in other shells.

This is an oversight, for sure. I do think we are just naively prepending `homeDirectory` instead of checking if we should use `xdg.configHome`. I am about to go on vacation, but can see if i can make time to fix this. It should just be updating the default `dotDir` to respect the `xdg.enable` like you said.

khaneliman:
@pecigonzalo Added a change to https://github.com/nix-community/home-manager/pull/8378 if you want to test

pecigonzalo:
Thanks @khaneliman , and sorry for the late reply here. It seems to be working 🎉, thanks again for taking a look.

Issue #5952, opened 2024-10-10
bug: tmux with 'sensibleOnTop' option no longer uses the correct shell


cixel:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

The upstream version of `tmux` was updated last week to 3.5a ([commit](https://github.com/NixOS/nixpkgs/commit/7facbffaf151f76629f1d67a67d109232ddb0b52)). One of the [changes](https://github.com/tmux/tmux/blob/tmux-3.5a/CHANGES#L7) in this version (from 3.5) is that `/bin/sh` is used for run-shell.

With this version of tmux, all panes/windows/etc use `sh` instead of the configured default shell (in my case, `zsh`).

My rough guess is that this change in tmux has introduced some ordering issues when `sensibleOnTop` is set, as `run-shell` will now use `sh` regardless of how tmux's default-shell is set.

The problem goes away if I do:
```
  package = pkgs.tmux.overrideAttrs (old: rec {
    version = "3.5";
    src = pkgs.fetchFromGitHub {
      owner = "tmux";
      repo = "tmux";
      rev = version;
      hash = "sha256-8CRZj7UyBhuB5QO27Y+tHG62S/eGxPOHWrwvh1aBqq0=";
    };
  });
```

or if I set `sensibleOnTop = false;`

### Maintainer CC

@jorsn @rycee (cc based on https://github.com/nix-community/home-manager/commit/a7cdfaa32585fce404cf8890bae099348e53e0ad).

### System information

```markdown
- system: `"aarch64-darwin"`
 - host os: `Darwin 23.6.0, macOS 14.7`
 - multi-user?: `yes`
 - sandbox: `no`
 - version: `nix-env (Nix) 2.18.8`
 - channels(ehdens): `"home-manager, nixpkgs"`
 - nixpkgs: `/Users/ehdens/.nix-defexpr/channels/nixpkgs`
```
```

x123:
I'm also experiencing the same issue. The notes for tmux 3.5a here https://github.com/tmux/tmux/issues/4162 say the following:

"Note that a fix in 3.5 and 3.5a to correctly set SHELL for run-shell and if-shell reveals a bug in the tmux-sensible script, causing it to set default-shell to /bin/sh. This can be worked around by adding this to .tmux.conf after the tmux-sensible script is invoked:"

gshpychka:
[...]

PR #6091, opened 2024-11-16
tmux: Disable tmux-sensible by Default (bug fix)

gen740:
### Description
In #5952, it was pointed out that tmux has changed how it handles shells, and because of this, the tmux.shell option doesn’t work properly unless tmux-sensible is turned off.

However, tmux-sensible hasn’t been updated for over two years, as mentioned [here](https://github.com/tmux-plugins/tmux-sensible/pull/70#issuecomment-2127666779). Since it’s no longer actively maintained, keeping it enabled by default is becoming more of a problem.

This PR suggests turning off tmux-sensible by default to avoid these issues and improve usability.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/programs/tmux.nix
+++ b/modules/programs/tmux.nix
@@ -239,7 +239,7 @@ in {
 
       sensibleOnTop = mkOption {
         type = types.bool;
-        default = true;
+        default = false;
         description = ''
           Run the sensible plugin at the top of the configuration. It
           is possible to override the sensible settings using the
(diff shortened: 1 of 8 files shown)


gen740:
@rycee

This is my first contribution to home-manager. Since tmux.nix lacks meta.maintainers, I wasn't sure who to CC.

I noticed there hasn't been any activity on this PR for about a week, so just checking if there's anything else I can do.

Thanks!

rycee:

PR #6092, opened 2024-11-16
version: allow 25.05 as state version

oneingan:
nixpkgs updated upstream: https://github.com/NixOS/nixpkgs/commit/90fcf3aa7eba70f18dad15b634e9336b997d25ac

### DIFF
--- a/modules/misc/version.nix
+++ b/modules/misc/version.nix
@@ -22,6 +22,7 @@ in {
         "23.11"
         "24.05"
         "24.11"
+        "25.05"
       ];
       description = ''
         It is occasionally necessary for Home Manager to change


rycee:

i-am-logger:
not sure this is complete... I still get the warning home is running on 24.05 even though `home.stateVerson = "25.05";`

image

i-am-logger:
might be the `release.json` needs an update?

https://github.com/nix-community/home-manager/blob/5056a1cf0ce7c2a08ab50713b6c4af77975f6111/release.json#L2

muradbu:
> not sure this is complete... I still get the warning home is running on 24.05 even though `home.stateVerson = "25.05";`
> 
> image

I'm running into the same issue currently

oneingan:
This change is the minimal required to avoid evaluation errors with the latest nixpkgs. If you want to proceed, you can use the home.enableNixpkgsReleaseCheck = false; option as suggested.

I assume that bumping home-manager requires a stricter release process than just a simple string change.

i-am-logger:
> This change is the minimal required to avoid evaluation errors with the latest nixpkgs. If you want to proceed, you can use the home.enableNixpkgsReleaseCheck = false; option as suggested.
> 
> I assume that bumping home-manager requires a stricter release process than just a simple string change.

i am not sure what is the release process for home-manager and what is required to change the `release.json` 

More work is required to properly release a 25.05 home-manager

PR #6100, opened 2024-11-18
mu: run initialization command when personal addresses change

VojtechStep:
### Description

When the user changes which addresses mu should consider 'personal', mu's store should be reinitialized.

After this change, the activation script parses the previously configured list of addresses and compares it with the new one. If they differ, it runs the init command even when the store has already been initialized.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@KarlJoad

### DIFF
--- a/modules/programs/mu.nix
+++ b/modules/programs/mu.nix
@@ -9,17 +9,22 @@ let
   # Used to generate command line arguments that mu can operate with.
   genCmdMaildir = path: "--maildir=" + path;
 
-  # Takes the list of accounts with mu.enable = true, and generates a
-  # command-line flag for initializing the mu database.
-  myAddresses = let
+  # Sorted list of personal email addresses to register
+  sortedAddresses = let
     # Set of email account sets where mu.enable = true.
     muAccounts =
       filter (a: a.mu.enable) (attrValues config.accounts.email.accounts);
     addrs = map (a: a.address) muAccounts;
     # Construct list of lists containing email aliases, and flatten
     aliases = flatten (map (a: a.aliases) muAccounts);
-    # Prefix --my-address= to each account's address AND all defined aliases
-    addMyAddress = map (addr: "--my-address=" + addr) (addrs ++ aliases);
+    # Sort the list
+  in sort lessThan (addrs ++ aliases);
+
+  # Takes the list of accounts with mu.enable = true, and generates a
+  # command-line flag for initializing the mu database.
+  myAddresses = let
+    # Prefix --my-address= to each account's address and all defined aliases
+    addMyAddress = map (addr: "--my-address=" + addr) sortedAddresses;
   in concatStringsSep " " addMyAddress;
 
 in {
@@ -49,14 +54,19 @@ in {
     home.activation.runMuInit = let
       maildirOption = genCmdMaildir config.accounts.email.maildirBasePath;
       dbLocation = config.xdg.cacheHome + "/mu";
+      muExe = getExe cfg.package;
     in hm.dag.entryAfter [ "writeBoundary" ] ''
-      # If the database directory exists, then `mu init` should NOT be run.
+      # If the database directory exists and registered personal addresses remain the same,
+      # then `mu init` should NOT be run.
       # In theory, mu is the only thing that creates that directory, and it is
       # only created during the initial index.
-      if [[ ! -d "${dbLocation}" ]]; then
-        run ${
-          getExe cfg.package
-        } init ${maildirOption} ${myAddresses} $VERBOSE_ARG;
+      MU_SORTED_ADDRS=$((${muExe} info store | ${
+        getExe pkgs.gawk
+      } '/personal-address/{print $4}' | LC_ALL=C sort | paste -sd ' ') || exit 0)
+      if [[ ! -d "${dbLocation}" || ! "$MU_SORTED_ADDRS" = "${
+        concatStringsSep " " sortedAddresses
+      }" ]]; then
+        run ${muExe} init ${maildirOption} ${myAddresses} $VERBOSE_ARG;
       fi
     '';
   };
--- a/modules/programs/mu.nix
+++ b/modules/programs/mu.nix
@@ -55,14 +55,13 @@ in {
       maildirOption = genCmdMaildir config.accounts.email.maildirBasePath;
       dbLocation = config.xdg.cacheHome + "/mu";
       muExe = getExe cfg.package;
+      gawkExe = getExe pkgs.gawk;
     in hm.dag.entryAfter [ "writeBoundary" ] ''
       # If the database directory exists and registered personal addresses remain the same,
       # then `mu init` should NOT be run.
       # In theory, mu is the only thing that creates that directory, and it is
       # only created during the initial index.
-      MU_SORTED_ADDRS=$((${muExe} info store | ${
-        getExe pkgs.gawk
-      } '/personal-address/{print $4}' | LC_ALL=C sort | paste -sd ' ') || exit 0)
+      MU_SORTED_ADDRS=$((${muExe} info store | ${gawkExe} '/personal-address/{print $4}' | LC_ALL=C sort | paste -sd ' ') || exit 0)
       if [[ ! -d "${dbLocation}" || ! "$MU_SORTED_ADDRS" = "${
         concatStringsSep " " sortedAddresses
       }" ]]; then
--- a/tests/integration/default.nix
+++ b/tests/integration/default.nix
@@ -12,6 +12,7 @@ let
   tests = {
     home-with-symbols = runTest ./standalone/home-with-symbols.nix;
     kitty = runTest ./standalone/kitty.nix;
+    mu = runTest ./standalone/mu;
     nixos-basics = runTest ./nixos/basics.nix;
     standalone-flake-basics = runTest ./standalone/flake-basics.nix;
     standalone-standard-basics = runTest ./standalone/standard-basics.nix;
--- a/tests/integration/standalone/mu/config-account-without-mu.nix
+++ b/tests/integration/standalone/mu/config-account-without-mu.nix
@@ -1,0 +1,5 @@
+{ lib, ... }: {
+  imports = [ ./config-two-accounts.nix ];
+
+  accounts.email.accounts.example2.mu.enable = lib.mkForce false;
+}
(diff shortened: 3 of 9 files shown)


VojtechStep:
> Does this also handle the case where you add/remove only _some_ addresses/aliases?

On activation it asks `mu` for the current list of aliases, so any observable change to that (modulo order) should trigger a reinit.

> I would like to codify the expected behavior of changing a subset of the addresses as an actual test that Nix/home-manager can run if we do go down this road.

I agree, though it would involve testing the actual side-effects of activation over multiple runs, and I don't know if the current test infra allows that.

VojtechStep:
> I agree, though it would involve testing the actual side-effects of activation over multiple runs, and I don't know if the current test infra allows that.

I tried to get the integration tests working, but I can't get a working internet connection in the VMs. In any case I'm pretty sure the patch works, since I've been using it for the past couple weeks, during which I manipulated my email addresses (adding new aliases and accounts), and `mu` always ended up in the expected state.

teto:
willing to merge but can't review. @KarlJoad could you test/review and confirm it works ?

KarlJoad (approved):

VojtechStep:
I'm not sure if I got the usage of `run` in the activation script right. On dry runs, should the check for currently registered email addresses run, so that the log correctly reflects if reinit would run or not, or should the check also not run?

KarlJoad:
In my opinion, a dry-run should do everything a normal run should do, except for any actual modifications, that way you (as the user) can see exactly what will happen, without anything actually changing.

So dry-runs _should_ check for currently-registered addresses.

VojtechStep:
> So dry-runs should check for currently-registered addresses.

I agree; fixed the check in dry-run mode and rebased on master.

KarlJoad (approved):

VojtechStep:
This is hopefully the last change, sorry. I'm setting up a new device today, and the current address detection failed because the store has never been initialized.

KarlJoad (changes requested):
> the current address detection failed because the store has never been initialized.

Honestly, I would have expected this to be caught by a test. Could you add a test for this too?

VojtechStep:
> Honestly, I would have expected this to be caught by a test.

Yeah, but it needs to be an integration test, and I still haven't managed to run those

VojtechStep:
I figured out how to run integration tests (in hindsight it makes sense that the VM can't access the internet when it's running in sandbox, so I had to turn that off; also I was getting `error: path /home/alice/.nix-defexpr/channels/home-manager is a symlink` because `home.revision` tries to read from it, but I didn't dig further, just set the default to null).

Hopefully now it should be mergeable.

KarlJoad:
I don't seem to be able to run the test locally, for some reason (Podman on my desktop). What is the exact `run.??` that I can use on the command-line to run these integration tests?

If I get this to test successfully locally, I'll gladly approve this again! 

Thanks for all your hard work on this @VojtechStep.

VojtechStep:
> What is the exact run.?? that I can use on the command-line to run these integration tests?

With flakes it's `nix build -L .#integration-test-mu`, otherwise `nix-build --arg pkgs 'import <nixpkgs> {}' tests/integration -A mu`, since AFAICT integration tests are not exposed in `tests/default.nix` (the explicit `pkgs` instantiation is necessary because it doesn't have a default value). If you get a `couldn't download ...` error in the VM then you need to turn off the sandbox, and if you get a `path ... is a symlink` error, then change `modules/misc/version.nix:78` from `default = let ...` to `default = null;` (I don't know why that's broken now, but all `standalone` integration tests are broken by it).

KarlJoad:
I actually ended up with a different error, which is probably just operator error.
```
$ nix build -L .#integration-test-mu --no-sandbox
...
machine # building '/nix/store/xjkydxc0n24mwxp8kh4wn5jq0fppga9k-bootstrap-tools.tar.xz.drv'...
machine # warning: error: unable to download 'http://tarballs.nixos.org/stdenv/x86_64-unknown-linux-gnu/82b583ba2ba2e5706b35dbe23f31362e62be2a9d/bootstrap-tools.tar.xz': Could not resolve hostname (6); retrying in 258 ms
machine # warning: error: unable to download 'http://tarballs.nixos.org/stdenv/x86_64-unknown-linux-gnu/82b583ba2ba2e5706b35dbe23f31362e62be2a9d/bootstrap-tools.tar.xz': Could not resolve hostname (6); retrying in 591 ms
machine # warning: error: unable to download 'http://tarballs.nixos.org/stdenv/x86_64-unknown-linux-gnu/82b583ba2ba2e5706b35dbe23f31362e62be2a9d/bootstrap-tools.tar.xz': Could not resolve hostname (6); retrying in 1174 ms
machine # warning: error: unable to download 'http://tarballs.nixos.org/stdenv/x86_64-unknown-linux-gnu/82b583ba2ba2e5706b35dbe23f31362e62be2a9d/bootstrap-tools.tar.xz': Could not resolve hostname (6); retrying in 2383 ms
machine # error:
machine #        … writing file '/nix/store/2pizl7lq4awa7p9bklr8037yh1sca0hg-bootstrap-tools.tar.xz'
machine #
machine #        error: unable to download 'http://tarballs.nixos.org/stdenv/x86_64-unknown-linux-gnu/82b583ba2ba2e5706b35dbe23f31362e62be2a9d/bootstrap-tools.tar.xz': Could not resolve hostname (6)
machine # error: builder for '/nix/store/xjkydxc0n24mwxp8kh4wn5jq0fppga9k-bootstrap-tools.tar.xz.drv' failed with exit code 1;
machine #        last 4 log lines:
machine #        > error:
machine #        >        … writing file '/nix/store/2pizl7lq4awa7p9bklr8037yh1sca0hg-bootstrap-tools.tar.xz'
machine #        >
machine #        >        error: unable to download 'http://tarballs.nixos.org/stdenv/x86_64-unknown-linux-gnu/82b583ba2ba2e5706b35dbe23f31362e62be2a9d/bootstrap-tools.tar.xz': Could not resolve hostname (6)
machine #        For full logs, run 'nix-store -l /nix/store/xjkydxc0n24mwxp8kh4wn5jq0fppga9k-bootstrap-tools.tar.xz.drv'.
machine # error: 1 dependencies of derivation '/nix/store/05q48dcd4lgk4vh7wyk330gr2fr082i2-bootstrap-tools.drv' failed to build
machine # error: 1 dependencies of derivation '/nix/store/s63zivn27i8qv5cqiy8r5hf48r323qwa-bash-5.2p37.drv' failed to build
machine # building '/nix/store/026i5h3a3pg28dxgv1x90117z6diha3x-byacc-20240109.tgz.drv'...
machine # error: 1 dependencies of derivation '/nix/store/by983vw5x0b8i6cbcz5y2hkd0n3n4a9b-coreutils-9.5.drv' failed to build
machine # error: 1 dependencies of derivation '/nix/store/zrp1z93l9cj9ia4n6s9cfs0f3s37r8v1-gcc-wrapper-13.3.0.drv' failed to build
machine # error: 1 dependencies of derivation '/nix/store/qhdvi3qcn60vrapyhsxxpbw0q63gmfz8-glibc-2.40-36.drv' failed to build
machine # error: 1 dependencies of derivation '/nix/store/vxckchzd4ny3dni980qf570fmfc3q5m6-stdenv-linux.drv' failed to build
machine # error: 1 dependencies of derivation '/nix/store/ws0w1jw9y3wwmvm305m18ivjil4nayls-stdenv-linux.drv' failed to build
machine # error: build of '/nix/store/7samcjpvkdln5jk24i3mmfl8blxnjmw8-home-manager.drv', '/nix/store/ws0w1jw9y3wwmvm305m18ivjil4nayls-stdenv-linux.drv' failed
machine # [  111.963817] su[1025]: pam_unix(su:session): session closed for user alice
machine: output:
cleanup
kill machine (pid 9)
qemu-system-x86_64: terminating on signal 15 from pid 6 (/nix/store/zv1kaq7f1q20x62kbjv6pfjygw5jmwl6-python3-3.12.7/bin/python3.12)
kill vlan (pid 7)
(finished: cleanup, in 0.00 seconds)
Traceback (most recent call last):
  File "/nix/store/ylz95zwxwfihzfwyv5pylkf92hn1yxd6-nixos-test-driver-1.1/bin/.nixos-test-driver-wrapped", line 9, in <module>
    sys.exit(main())
             ^^^^^^
  File "/nix/store/ylz95zwxwfihzfwyv5pylkf92hn1yxd6-nixos-test-driver-1.1/lib/python3.12/site-packages/test_driver/__init__.py", line 146, in main
    driver.run_tests()
  File "/nix/store/ylz95zwxwfihzfwyv5pylkf92hn1yxd6-nixos-test-driver-1.1/lib/python3.12/site-packages/test_driver/driver.py", line 174, in run_tests
    self.test_script()
  File "/nix/store/ylz95zwxwfihzfwyv5pylkf92hn1yxd6-nixos-test-driver-1.1/lib/python3.12/site-packages/test_driver/driver.py", line 166, in test_script
    exec(self.tests, symbols, None)
  File "<string>", line 41, in <module>
  File "<string>", line 22, in succeed_as_alice
  File "/nix/store/ylz95zwxwfihzfwyv5pylkf92hn1yxd6-nixos-test-driver-1.1/lib/python3.12/site-packages/test_driver/machine.py", line 612, in succeed
    raise Exception(f"command `{command}` failed (exit code {status})")
Exception: command `su -l alice --shell /bin/sh -c $'export XDG_RUNTIME_DIR=/run/user/$UID ; nix-shell "<home-manager>" -A install'` failed (exit code 1)
```

VojtechStep:
I tried that to disable the sandbox at first, but got a warning that I'm not a trusted user, so the setting is ignored. I ended up temporarily disabling it from my NixOS config. Could something similar be happening to you?

KarlJoad:
It could be. I have not added my user to `/etc/nix.conf` as a `trustedUser`. I can try adding my user to that list, but it will take me a while. I have not updated my computer in a while, and I have been planning to combine an upgrade with my next reconfigure.

KarlJoad:
I have added myself to `trustedUsers` in `/etc/nix/nix.conf`, but I still cannot build the integration tests. Perhaps @rycee could offer some insight on how to get integration tests building?

rycee:
I usually just run

    nix build --option sandbox false -L .#integration-test-all

and have `trusted-users = root root @wheel` in `/etc/nix/nix.conf`.

In general the setup should pretty closely mirror the test setup in NixOS (with the exception that the sandbox unfortunately needs to be off).

rycee:
Merged to master now 🙂

KarlJoad:
Yep, those changes made integration tests work for me too. Thanks for clearing that up @rycee! Thanks for your work on this @VojtechStep!

PR #6104, opened 2024-11-19
syncthing: expand declarative config to Darwin

### LINKED ISSUE #4049: Add declarative config options for Syncthing

pitkling:
### Description

This PRs extends the recently merged PR #5616, which expanded the Synching config to allow declarative settings under Linux, such that it also works under Darwin.

To simplify the review process, I left split the PR into several commits for now:
* The first five commits perform some minor refactoring tasks of the module code such that parts of it can be reused in the Darwin specific changes.
* Commit six implements the actual Darwin specific changes. It does the following:
  * Update the module's `syncthing` launchd agent to copy the synching key/certificate before starting syncthing, analogously to the systemd service from the above mentioned PR #5616.
  * Adds an `syncthing-init`launchd agent (analogously to the systemd service `synching-init` from the above mentioned PR #5616) that updates the configuration files. Since this must be run after the syncthing service started, we use a `WatchPath` to coordinate both launchd agents.
* The last commit just fixes the formatting according to HM's `format` tool.

There might be better ways to coordinate the two launchd agents I'm not aware of. But I have had this running flawlessly for over two months now (on four Darwin and two Linux machines sharing various folders).

Fixes #4049 for Darwin systems.

### Checklist

- [x] Commit messages are formatted (see commit 6; I'll squash the other commits into it after the review process)

#### Maintainer CC

@karaolidis (pinging as author of the original (linux-specific) PR #5616)
@rycee (maintainer)

### DIFF
--- a/modules/services/syncthing.nix
+++ b/modules/services/syncthing.nix
@@ -687,15 +687,17 @@ in {
         };
       };
 
-      launchd.agents.syncthing = {
-        enable = true;
-        config = {
-          ProgramArguments = syncthingArgs;
-          KeepAlive = {
-            Crashed = true;
-            SuccessfulExit = false;
+      launchd.agents = {
+        syncthing = {
+          enable = true;
+          config = {
+            ProgramArguments = syncthingArgs;
+            KeepAlive = {
+              Crashed = true;
+              SuccessfulExit = false;
+            };
+            ProcessType = "Background";
           };
-          ProcessType = "Background";
         };
       };
     })
--- a/modules/services/syncthing.nix
+++ b/modules/services/syncthing.nix
@@ -50,6 +50,7 @@ let
   cat = lib.getExe' pkgs.coreutils "cat";
   curl = lib.getExe pkgs.curl;
   install = lib.getExe' pkgs.coreutils "install";
+  mktemp = lib.getExe' pkgs.coreutils "mktemp";
   syncthing = lib.getExe cfg.package;
 
   copyKeys = pkgs.writers.writeBash "syncthing-copy-keys" ''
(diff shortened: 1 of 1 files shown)


karaolidis (approved):

modules/services/syncthing.nix
```diff
@@ -11,6 +11,12 @@ let
 
   isUnixGui = (builtins.substring 0 1 cfg.guiAddress) == "/";
 
+  # syncthing's configuration directory (see https://docs.syncthing.net/users/config.html)
+  syncthing_dir = if pkgs.stdenv.isDarwin then
+    "$HOME/Library/Application Support/Syncthing"
+  else
+    "\${XDG_STATE_HOME:-$HOME/.local/state}/syncthing";
```
rycee, line 18: Quick question. Will the Darwin directory be used even if the XDG_STATE_HOME variable is set?

rycee (commented):
Thanks for the contribution, looks OK to me! Especially thanks for splitting the change in multiple commits with explanations 🙂

Just added one comment.

modules/services/syncthing.nix
pitkling: I think so, that's at least how I understood the documentation.

Anyway, I remember that I was also a bit surprised by this when I wrote the module. So just to be sure, this time I took a quick look at the source code (release tag v1.28.0, as used in 24.11). Indeed, `XDG_STATE_HOME` is only considered when [not on Windows/Darwin](https://github.com/syncthing/syncthing/blob/6d64daaba326ba8378c700e00a31e425c7c90579/lib/locations/locations.go#L178-L189) and for Darwin [the location is hardcoded](https://github.com/syncthing/syncthing/blob/6d64daaba326ba8378c700e00a31e425c7c90579/lib/locations/locations.go#L208-L210) as `$HOME/Library/Application Support/Syncthing`.

ameyp:
Could a reviewer please look into this?

DamienCassou:
The branch has some conflicts.

pitkling:
> The branch has some conflicts.

Thanks @DamienCassou, I didn't notice the recent refactoring by @khaneliman. I just updated the PR to resolve the conflicts.

khaneliman (approved):
LGTM, thanks!

Issue #6110, opened 2024-11-20
bug: home-manager-manual.drv fails to build


knightpp:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

After update

```diff
       "locked": {
-        "lastModified": 1731235328,
-        "narHash": "sha256-NjavpgE9/bMe/ABvZpyHIUeYF1mqR5lhaep3wB79ucs=",
+        "lastModified": 1732025103,
+        "narHash": "sha256-qjEI64RKvDxRyEarY0jTzrZMa8ebezh2DEZmJJrpVdo=",
         "owner": "nix-community",
         "repo": "home-manager",
-        "rev": "60bb110917844d354f3c18e05450606a435d2d10",
+        "rev": "a46e702093a5c46e192243edbd977d5749e7f294",
         "type": "github"
       },
```

I can no longer build new generations

```
$ home-manager build
error: builder for '/nix/store/qfxwfqsfs0y9d5ipjc7y5sx6hpsl7w3y-home-manager-manual.drv' failed with exit code 1;
       last 25 log lines:
       >     - ch-contributing
       >     - sec-release-19.03
       >     - sec-release-24.05
       >     - sec-code-style
       >     - sec-option-types-gvariant-mkUchar
       >     - sec-guidelines-commit-message-style
       >     - sec-release-24.05-state-version-changes
       >     - sec-release-23.11-highlights
       >     - ch-3rd-party
       >     - sec-release-21.05
       >     - sec-commit-style
       >     - sec-usage-gpu-non-nixos
       >     - sec-option-types-dag-entryBefore
       >     - sec-release-21.05-state-version-changes
       >     - sec-release-21.05-highlights
       >     - sec-updating
       >     - sec-guidelines-add-tests
       >     - sec-usage-graphical
       >     - sec-option-types-gvariant-mkString
       >     - sec-release-23.11-state-version-changes
       >
       >     This can happen when an identifier was added or renamed.
       >     Please update doc/redirects.json or nixos/doc/manual/redirects.json!
       >
       > NOTE: If your Manual build passes locally and you see this message in CI, you probably need a rebase.
       For full logs, run 'nix log /nix/store/qfxwfqsfs0y9d5ipjc7y5sx6hpsl7w3y-home-manager-manual.drv'.
[...]

PR #6111, opened 2024-11-20
yazi: update keymap config

GaetanLepage:
### Description

Update description and tests for the `yazi.keymap` option to reflect upstream changes.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

cc @XYenon @eljamm

### DIFF
--- a/modules/programs/yazi.nix
+++ b/modules/programs/yazi.nix
@@ -77,16 +77,16 @@ in {
       default = { };
       example = literalExpression ''
         {
-          input.keymap = [
-            { exec = "close"; on = [ "<C-q>" ]; }
-            { exec = "close --submit"; on = [ "<Enter>" ]; }
-            { exec = "escape"; on = [ "<Esc>" ]; }
-            { exec = "backspace"; on = [ "<Backspace>" ]; }
+          input.prepend_keymap = [
+            { run = "close"; on = [ "<C-q>" ]; }
+            { run = "close --submit"; on = [ "<Enter>" ]; }
+            { run = "escape"; on = [ "<Esc>" ]; }
+            { run = "backspace"; on = [ "<Backspace>" ]; }
           ];
-          manager.keymap = [
-            { exec = "escape"; on = [ "<Esc>" ]; }
-            { exec = "quit"; on = [ "q" ]; }
-            { exec = "close"; on = [ "<C-q>" ]; }
+          manager.prepend_keymap = [
+            { run = "escape"; on = [ "<Esc>" ]; }
+            { run = "quit"; on = [ "q" ]; }
+            { run = "close"; on = [ "<C-q>" ]; }
           ];
         }
       '';
--- a/tests/modules/programs/yazi/keymap-expected.toml
+++ b/tests/modules/programs/yazi/keymap-expected.toml
@@ -1,27 +1,27 @@
-[[input.keymap]]
-exec = "close"
+[[input.prepend_keymap]]
 on = ["<C-q>"]
+run = "close"
 
-[[input.keymap]]
-exec = "close --submit"
+[[input.prepend_keymap]]
 on = ["<Enter>"]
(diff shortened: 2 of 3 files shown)


modules/programs/yazi.nix
```diff
@@ -77,16 +77,16 @@ in {
       default = { };
       example = literalExpression ''
         {
-          input.keymap = [
-            { exec = "close"; on = [ "<C-q>" ]; }
-            { exec = "close --submit"; on = [ "<Enter>" ]; }
-            { exec = "escape"; on = [ "<Esc>" ]; }
-            { exec = "backspace"; on = [ "<Backspace>" ]; }
+          input.keymap_prepend = [
```
eljamm, line 80: I believe the correct attribute to use here is `prepend_keymap` (see [keymap.toml](https://yazi-rs.github.io/docs/configuration/keymap/))

```suggestion
          input.prepend_keymap = [
```

tests/modules/programs/yazi/keymap-expected.toml
```diff
@@ -1,27 +1,27 @@
-[[input.keymap]]
-exec = "close"
+[[input.keymap_prepend]]
+run = "close"
```
eljamm, line 2: `run` and `on` need to be alphabetically ordered, else the `nix build -L .#test-yazi-settings` test fails.

This happens because Nix orders the attributes in alphabetical order, so the generated toml file is ordered as well, while the expected one is not, thus causing a conflict.

eljamm (commented):
Thanks @GaetanLepage. There is a similar PR https://github.com/nix-community/home-manager/pull/5647 which has been open for a while. If all CI checks pass, perhaps we can merge this one, instead.

Before that, there are a few things to fix, here.

modules/programs/yazi.nix
GaetanLepage: Oops. Fixed.

tests/modules/programs/yazi/keymap-expected.toml
GaetanLepage: done.

tests/modules/programs/yazi/settings.nix
```diff
@@ -5,35 +5,35 @@
     enable = true;
 
     keymap = {
-      input.keymap = [
+      input.keymap_prepend = [
```
eljamm, line 8: There is still some `keymap_prepend` left

```suggestion
      input.prepend_keymap = [
```

tests/modules/programs/yazi/settings.nix
```diff
@@ -5,35 +5,35 @@
...
           on = [ "<Esc>" ];
         }
         {
-          exec = "backspace";
+          run = "backspace";
           on = [ "<Backspace>" ];
         }
       ];
-      manager.keymap = [
+      manager.keymap_prepend = [
```
eljamm, line 26: Here as well.

```suggestion
      manager.prepend_keymap = [
```

GaetanLepage: Done.

GaetanLepage:
> Looks good. Hopefully the CI passes so this can be merged.

Thanks for the review !

XYenon (approved):
LGTM. Could you rebase it to master? Maybe it will pass the CI after rebase.

eljamm:
Not only are the tests taking a long time, but they're also consistently failing when building the docs ...

For the former, I think it's because there is no cache set up for the [cachix/install-nix-action](https://github.com/cachix/install-nix-action), so the tests are building everything with each PR, which is wasteful IMO.

For the latter, I believe the docs broke because of recent changes to the redirects in nixpkgs, but there is a fix in the works (see https://github.com/nix-community/home-manager/issues/6110). Until this is addressed or @rycee merges this, I believe there isn't much else to do besides wait.

PR #6115, opened 2024-11-21
rofi: modes option

mightyiam:
### Description

Adds `programs.rofi.modes` option.
For the tiny convenience of supporting the custom mode syntax.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/rofi.nix
+++ b/modules/programs/rofi.nix
@@ -132,6 +132,7 @@ let
     else
       cfg.theme;
 
+  modes = map (mode: if isString mode then mode else "${mode.name}:${mode.path}") cfg.modes;
 in
 {
   options.programs.rofi = {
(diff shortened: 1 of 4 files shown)


teto:
might be more forward-going to convert `configuration` into a freeform option ?

mightyiam:
> might be more forward-going to convert `configuration` into a freeform option ?

I'm trying to squeeze the most value out of home-manager 🤷‍♂️

khaneliman:
> > might be more forward-going to convert `configuration` into a freeform option ?
> 
> I'm trying to squeeze the most value out of home-manager 🤷‍♂️

The comment is more towards shifting from manually declaring all options, at the top-level, possible in the config, to a freeform option that accepts anything and will just convert it to the toRasi generator. You can still declare submodule options like this inside of it to make transformations easier and document them. But, I think that can be done as a refactor effort in this module.

mightyiam:
Thank you for explaining. I am trying to replace my usage of rofi with something else anyway. Maybe kickoff or fuzzel. So I don't think I'll be contributing much more to this module. I think this is slightly useful so I will leave it to whomever wants to make a decision here.

khaneliman:
> I'm using `rofi-wayland` and this PR causes a segfault when starting rofi

What does the generated default config look like? I assumed the generator wouldn't just insert an empty list. If it does, we need to add some check before adding to config.

marijanp:
@khaneliman an empty list appears
```
configuration {
font: "Roboto Mono 10";
location: 0;
modes: [  ];
terminal: "kitty";
xoffset: 0;
yoffset: 0;
}
@theme "nord"
```

marijanp:
@khaneliman can you link the PR?

khaneliman:
> @khaneliman can you link the PR?

https://github.com/nix-community/home-manager/pull/7003

mightyiam:
Sorry!

Issue #5794, opened 2024-09-02
home.stateVersion explained?


tamis-laan:
I can't really find a good explanation of `home.stateVersion` what is what does it do? Some comments I have read say this value should never be touched. But then why is there, and not "hidden"?

Can we have a good explanation and have it added to the manual?

mapperfr:
This snippet from ```configuration.nix``` partly answers the question:

```
  # This value determines the NixOS release from which the default
  # settings for stateful data, like file locations and database versions
  # on your system were taken. It‘s perfectly fine and recommended to leave
  # this value at the release version of the first install of this system.
  # Before changing this value read the documentation for this option
  # (e.g. man configuration.nix or on https://nixos.org/nixos/options.html).
  system.stateVersion = "24.05"; # Did you read the comment?
```

doraeric:
I still don't understand why I need to keep a setting in my config file that I’m not supposed to modify. If it represents the supported version of the home-manager config file, why not use semantic versioning to make it easier to manage? If not, why not let the program handle it with something like a `.lock` file instead?  

The variable name also makes it seem like the installed package would only correspond to the version from that specific date.

> This option defines the first version of NixOS you have installed on this particular machine, and is used to maintain compatibility with application data (e.g. databases) created on older NixOS versions.

This statement confuses me even more. If I install home-manager on different machines at different times—for example, on another machine in the future—they will generate different default configurations and `stateVersion` values.  

Then, if I use Git to sync configurations across machines, do I need to manually change `stateVersion` to match the value from when each machine was initially set up?

gensimone:
> I still don't understand why I need to keep a setting in my config file that I’m not supposed to modify. If it represents the supported version of the home-manager config file, why not use semantic versioning to make it easier to manage? If not, why not let the program handle it with something like a `.lock` file instead?
> 
> The variable name also makes it seem like the installed package would only correspond to the version from that specific date.
> 
> > This option defines the first version of NixOS you have installed on this particular machine, and is used to maintain compatibility with application data (e.g. databases) created on older NixOS versions.
> 
> This statement confuses me even more. If I install home-manager on different machines at different times—for example, on another machine in the future—they will generate different default configurations and `stateVersion` values.
> 
> Then, if I use Git to sync configurations across machines, do I need to manually change `stateVersion` to match the value from when each machine was initially set up?

I would also like to know how to handle the last situation you described.

jeggao:
[...]

PR #6116, opened 2024-11-21
docs: enhance comment for home.stateVersion option

mapperfr:
See issue #5794

### Description

This PR proposes to add a bit of explanation about the home.stateVersion option. It partly copies the comment available on ```configuration.nix``` in NixOS.


rycee:
Thanks! It may be worth copying the version of this text so that it matches the configuration generated when you do a standalone installation:
https://github.com/nix-community/home-manager/blob/a9953635d7f34e7358d5189751110f87e3ac17da/home-manager/home-manager#L355-L362

PR #6124, opened 2024-11-22
home-manager: prepare 24.11 release

rycee:


stelcodes:
Just a heads up, with #6112 unmerged anyone relying on zsh abbreviations will have a partially broken shell upon switching to `release-24.11`.

rycee:
@stelcodes I think it should be resolved by https://github.com/nix-community/home-manager/commit/30a16c91a4b133a4fadbc27bb1ab2c3096c5e08a.

eljamm:
Yazi keymaps are currently broken without #6111 or #5647. I think the former should be merged as it also changes the keymaps to use `prepend_keymap`, which will override individual keys instead of all the default ones (See [keymap.toml](https://yazi-rs.github.io/docs/configuration/keymap/)).

PR #6129, opened 2024-11-23
kdeconnect: Upgrade default version to Plasma6

RedEtherbloom:
Plasma6 has been out for a while now and the service still points to the old, incompatible version.

### Description

Upgrade the default version of kde-connect to use Plasma6.

The current one, based on Plasma5, is 23.08 and incompatible with Plasma6.

### Checklist

- [x] Change is backwards compatible.(Kinda? It may break in Plasma5 setups)

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@adisbladis

### DIFF
--- a/modules/services/kdeconnect.nix
+++ b/modules/services/kdeconnect.nix
@@ -14,8 +14,8 @@ in {
       enable = mkEnableOption "KDE connect";
       package = mkOption {
         type = types.package;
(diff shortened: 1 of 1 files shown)


RedEtherbloom:
It's ready now. The failing test seems to be docs, as it's happening on all other PRs again.

PR #6131, opened 2024-11-23
hypridle: fix service when no config file

sebaszv:
### Description

When `cfg.settings` is the default `{}`, the systemd user service declaration fails since it depends on `config.xdg.configFile."hypr/hypridle.conf".source` for `X-Restart-Triggers`. This causes failure if nothing is set for `cfg.settings`.

Making the addition conditional on `cfg.settings` actually having content, which would mean `xdg.configFile."hypr/hypridle.conf"` does exist, avoids the error.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


sebaszv:
Seems to be fine, but the Linux test is failing on the docs test, which appears to be broken across the board in other PRs as well. This is after rebasing as well.

`nix develop --ignore-environment .#all` still succeeds on my machine.

khaneliman (approved):
LGTM, thanks

PR #6133, opened 2024-11-24
hyprpaper: fix service when no config file

sebaszv:
### Description

> Applies the same fix here from <https://github.com/nix-community/home-manager/pull/6131>.

The the systemd user service depends on `config.xdg.configFile."hypr/hyprpaper.conf".source` for `X-Restart-Triggers`. When `cfg.settings` is the default `{}`, this causes failure since `config.xdg.configFile."hypr/hyprpaper.conf".source` will not exist.

Making the addition conditional on `cfg.settings` actually having content, which would mean `xdg.configFile."hypr/hyprpaper.conf"` does exist, avoids the error.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@fufexan
@khaneliman

### DIFF
--- a/modules/services/hyprpaper.nix
+++ b/modules/services/hyprpaper.nix
@@ -75,7 +75,7 @@ in {
         Description = "hyprpaper";
         After = [ "graphical-session-pre.target" ];
         PartOf = [ "graphical-session.target" ];
-        X-Restart-Triggers =
+        X-Restart-Triggers = mkIf (cfg.settings != { })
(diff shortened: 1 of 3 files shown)


sebaszv:
As mentioned in <https://github.com/nix-community/home-manager/pull/6131>, the docs test appears be failing across all current PRs at the moment, even though this should be just fine.

`nix develop --ignore-environment .#all` still succeeds on my machine.

PR #6137, opened 2024-11-24
podman: Added build, image, and volume quadlet support

bamhm182:
### Description

- Added support for `build`, `image`, and `volume` quadlets
- Resolved test failures due to podman 5.3.0 upgrade
- Replaced several instances of `pkgs.podman` with `services.podman.package`

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@n-hass @bamhm182


bamhm182:
Cursory testing appears to have everything functioning as intended, but submitting PR to get more eyes on it any try to break things.

Also looking for a bit of feedback on:
- [The way I am finding container resources](https://github.com/bamhm182/home-manager/blob/e7a37e4927b6382d2b763b3d6d190b2731836932/modules/services/podman-linux/quadlets/container.nix#L44) is a little hairy. Is there a better way to do this?

Questions for @n-hass:
- Do we still need awaitPodmanUnshare? It seems like we're adding that to make sure that podman actually is available before we try to use it, but if the podman maintainers aren't worrying about anything past `podman-user-wait-network-online.service`, is it actually necessary?
- What was the stuff in [Service.Environment](https://github.com/bamhm182/home-manager/blob/e7a37e4927b6382d2b763b3d6d190b2731836932/modules/services/podman-linux/quadlets/network.nix#L29) for again? I'm not 100% sure when it is necessary and what problem it's resolving.

Other notes:
- There doesn't appear to be a way to tag images, so our standard approach of using `nix.home-manager.managed=true` to find our images doesn't really work. I don't think it's that big of a problem as there is hopefully periodic pruning of images. The images that come from builds are tagged and managed appropriately.

bamhm182:
I just realized I forgot to add in the integration tests for the build quadlet. I am also expecting the container dependencies for build quadlets won't be quite right. Also looks like the docs are not building properly. Will look into all this in a few hours.

bamhm182:
After playing around with this more, I am wondering if we should add an option to `services.podman.volumes.*` to stop the activation script from nuking the volumr automatically when removed? I am thinking like a `preserve` option which adds a tag like `homemanager.preserve=true` that would have the activation script note that it has been queued for deletion, but that flag is set. Images can be rebuilt, contains can be reset, data is valuable and can be easily lost.

The current functionality when setting `services.podman.containers.mysleep.volumes = [ "somevol:/tmp"] ` prior to this PR is that a volume named "somevol" will be created automatically if it doesn't exist, and it will not be deleted when everything is removed from the nix config, which makes me think preserving the volumes would be the default option.

Alternatively, there could be an option like `services.podman.deleteHmVolumes`, which is set to false by default and would result in volumes not being looked at then false, and managed like everything else when true.

n-hass:
still haven’t had time to have a proper play around and review, but I would go with your first option - add a persistent/preserve option to volumes that defaults true, set a label `nix.home-manager.preserve = true`, and adjust the activation script to respect it

bamhm182:
All good. Should be good to go with that change.

bamhm182:
Ran into a couple snags with the Builds quadlets. Fixed an issue where it should be referencing `localhost\homemanager\bld` instead of `homemanager\bld`. Easy enough.

One issue that I'm having a bit of a harder time with is that the Pull option appears to be breaking the ability to build an image because it appears podman is understanding the value set for `--pull` as the context directory. The quadlet builds the following line:
```
$ podman build --pull always --tls-verify --tag homemanager/coder --label nix.home-manager.managed=true --file /nix/store/v9hfyvdv0lr86w3xdkrix30g5dppr6vg-coder-containerfile
Error: context must be a directory: "./newer"
```

It works when executed with an `=`, like this:
```
$ podman build --pull=always --tls-verify --tag homemanager/coder --label nix.home-manager.managed=true --file /nix/store/v9hfyvdv0lr86w3xdkrix30g5dppr6vg-coder-containerfile
```

I am thinking that I may remove the `pull` attribute and expect users who wish to set it to do `builds."bld".extraConfig.Build.extraPodmanArgs = [ "--pull=newer" ];` if they want it set.

modules/services/podman-linux/builds.nix
```diff
@@ -0,0 +1,173 @@
...
+          PodmanArgs = buildDef.extraPodmanArgs;
+          SetWorkingDirectory = buildDef.workingDirectory;
+          TLSVerify = buildDef.tlsVerify;
+        };
+        Install = {
+          WantedBy = (if buildDef.autoStart then [
+            "default.target"
+            "multi-user.target"
+          ] else
+            [ ]);
```
rycee, line 26: Can use `optionals`. Something like
```suggestion
          WantedBy = optionals buildDef.autoStart [
            "default.target"
            "multi-user.target"
          ];
```
ought to work.

modules/services/podman-linux/builds.nix
```diff
@@ -0,0 +1,173 @@
...
+
+      authFile = mkOption {
+        type = with types; nullOr path;
+        default = null;
+        description = "Path of the authentication file.";
+      };
+
+      description = mkOption {
+        type = with types; nullOr str;
+        default = null;
```
rycee, line 73: Can change `types.submodule {` to `types.submodule ({name, ...}: {` and then change this like so:
```suggestion
        type = with types; nullOr str;
        default = "Service for build ${name}";
        defaultText = "Service for build \${name}";
```

modules/services/podman-linux/builds.nix
```diff
@@ -0,0 +1,173 @@
...
+        };
+        Service = {
+          TimeoutStartSec = 300;
+          RemainAfterExit = "yes";
+        };
+        Unit = {
+          Description = (if (builtins.isString buildDef.description) then
+            buildDef.description
+          else
+            "Service for build ${name}");
```
rycee, line 36: The parentheses are unnecessary

```suggestion
          Description = if isString buildDef.description then
            buildDef.description
          else
            "Service for build ${name}";
```

but I think the better solution is to make `description` have type `str` and default the value to `"Service for build ${name}"`. See my other comment.

modules/services/podman-linux/builds.nix
```diff
@@ -0,0 +1,173 @@
...
+  };
+in let
+  buildDefinitionType = types.submodule {
+    options = {
+
+      autoStart = mkOption {
+        type = types.bool;
+        default = true;
+        description =
+          "Whether to start the build on boot. (Requires user lingering)";
```
rycee, line 62:
```suggestion
          "Whether to start the build on boot. Requires user lingering.";
```

modules/services/podman-linux/builds.nix
```diff
@@ -0,0 +1,173 @@
...
+    default = { };
+    description = "Defines Podman build quadlet configurations.";
+  };
+
+  config = let buildQuadlets = mapAttrsToList toQuadletInternal cfg.builds;
+  in mkIf cfg.enable {
+    services.podman.internal.quadletDefinitions = buildQuadlets;
+    assertions = flatten (map (build: build.assertions) buildQuadlets);
+
+    home.file."${config.xdg.configHome}/podman/images.manifest".text =
```
rycee, line 170: Not sure why the podman module ended up using this `home.file` format, seems to me that it would be possible to simply use `xdg.configFile`.

```suggestion
    xdg.configFile."podman/images.manifest".text =
```

modules/services/podman-linux/builds.nix
```diff
@@ -0,0 +1,173 @@
...
+
+  createQuadletSource = name: buildDef:
+    let
+      buildConfig = (podman-lib.deepMerge {
+        Build = {
+          AuthFile = buildDef.authFile;
+          Environment = buildDef.environment;
+          File = buildDef.file;
+          ImageTag = ([ "homemanager/${name}" ] ++ buildDef.tags);
+          Label = (buildDef.labels // { "nix.home-manager.managed" = true; });
```
rycee, line 16:
```suggestion
          ImageTag = [ "homemanager/${name}" ] ++ buildDef.tags;
          Label = buildDef.labels // { "nix.home-manager.managed" = true; };
```

modules/services/podman-linux/builds.nix
```diff
@@ -0,0 +1,173 @@
+{ config, lib, pkgs, ... }:
+with lib;
+let
+  cfg = config.services.podman;
+
+  podman-lib = import ./podman-lib.nix { inherit lib config; };
+
+  createQuadletSource = name: buildDef:
+    let
+      buildConfig = (podman-lib.deepMerge {
```
rycee, line 10:
```suggestion
      buildConfig = podman-lib.deepMerge {
```

modules/services/podman-linux/podman-lib.nix
```diff
@@ -83,8 +85,35 @@ in {
...
+      checkImageTag = extraConfig:
+        let
+          buildSection = if (builtins.hasAttr "Build" extraConfig) then
+            extraConfig.Build
+          else
+            { };
+          imageTags = if (builtins.hasAttr "ImageTag" buildSection) then
+            buildSection.ImageTag
+          else
+            [ ];
```
rycee, line 97: Perhaps this could be simplified to
```suggestion
          imageTags = (extraConfig.Build or {}).ImageTag or [];
```

modules/services/podman-linux/podman-lib.nix
```diff
@@ -83,8 +85,35 @@ in {
...
+            "") == "homemanager/${quadletName}";
+        in if (containsRequiredTag || builtins.length imageTags == 0) then
+          [ ]
+        else [{
+          assertion = false;
+          message = ''
+            In '${quadletName}' config. Build.ImageTag: '[ "${
+              builtins.concatStringsSep ''" "'' imageTags
+            }" ]' does not have 'homemanager/${quadletName}' at index 0.'';
+        }];
```
rycee, line 110: Does this tag really have to be in position 0? Perhaps something like

```suggestion
          containsRequiredTag = builtins.elem "homemanager/${quadletName}" imageTags;
          
          imageTagsStr = concatMapStringsSep " " toString imageTags;
        in [{
          assertion = imageTags == [ ] || containsRequiredTag;
          message = ''
            In '${quadletName}' config. Build.ImageTag: '[ ${imageTagsStr} ]' does not contain 'homemanager/${quadletName}'.'';
        }];
```

would work?

rycee:
Thanks! Looks quite impressive. I've added a few mostly stylistic comments. I cannot comment on the functionality since I don't use the module myself.

modules/services/podman-linux/builds.nix
bamhm182: I went ahead and made the changes in your other comment, then set this as `Description = buildDef.description;`. If it is null, then it is automatically dropped. Much simpler!

modules/services/podman-linux/podman-lib.nix
bamhm182: The reason that I had it forced to position 0 was so that you could easily discern which ones were from Home Manager and which ones were built outside of Home Manager. That said, it probably isn't really that important and this code is a lot more straight forward. Will make the swap.

bamhm182:
Thanks for all of your feedback @rycee. Incorporating it into a new commit now. I went ahead and took the suggestions that  you made towards `build.nix` and applied them to all the new quadlets. As they do not affect the functionality of the old quadlets, I have left them there for now. I have one more pull I want to make to get `pod` and `kube` quadlets in here, then I believe this code base is in need of a refactor just to make sure that everything is standardized across all files and it's all easy to read and maintain. I will incorporate these suggestions (and a handful of others I have noticed) through the rest of the files  into that refactor PR.

Prunkles:
I saw #6228, and I wonder does and how this PR is used with `pkgs.dockerTools.build*` functions?

As I see, it adds a Quadlet Build section support, but it works for native Containerfiles, and it is not possible to retrieve them from `pkgs.dockerTools.build*`, as far as I know. `virtualisation.oci-containers.containers.<name>.imageFile` from NixOS modules uses a custom script which runs `podman load -i $imageFile`, and I don't see anything like this in this PR

bamhm182:
It doesn't. That's not something I was tracking. I can take a look at it, but if dockerTools drops the image in a place that podman can read (is it in the nix store?) Then I imagine you could pass it to the images functionality in this PR. Haven't tried it, though.

bamhm182:
I had forgotten about that ticket, but my thought process at the time was that it would just work, if that's how dockerTools would normally be used. I haven't used dockerTools. I'll try and remember to give it a shot and report back.

n-hass:
You wouldn’t need `pkgs.dockerTools` to use Containerfiles with this PR. If you had a ‘build' Quadlet with a containerfile ([docs](https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html#file)) called `myimage.build` , and the `Image` setting for the ‘container’ Quadlet ([docs](https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html#image)) is set to `myimage.build`, then yes Podman makes it work. The auto-generated dependency of the container on `myimage.build` may be wrong as it will be `myimage-build.service`, while we add a prefix, but thats easily patchable.

This PR won't give the same ‘direct’ behaviour as just specifiying a path to an `imageFile` like in the NixOS module, but you achieve the same thing by using a `build`-type Quadlet.

n-hass:
Thanks. This now looks ready to merge please @rycee

Prunkles:
@n-hass 

> You wouldn’t need `pkgs.dockerTools` to use Containerfiles with this PR.

But what if I want to? I like how images are built using `pkgs.dockerTools` with all its nix ecosystem and approach, and I wouldn't like to use the raw Containerfile for this.

> This PR won't give the same ‘direct’ behaviour as just specifiying a path to an `imageFile` like in the NixOS module, but you achieve the same thing by using a `build`-type Quadlet.

So is it possible to use `dockerTools` with this PR? I didn't quite understand how to do it with Quadlet. Is it just about setting `Image` field in `.image` unit to an output path of `dockerTools.buildImage` derivation (which is a .tar.gz archive of the image)?

n-hass:
@Prunkles 
I'm not a user of `dockerTools`, but some quick reading suggests you can use the `docker-archive` format of the image created with it in a corresponding `Image` Quadlet to be able to use it for a Quadlet container.

Check [this documentation](https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html#id61) for example usage of a docker-archive. tldr something like

```
services.podman.images."my-img" = { image = "${pathToArvhive}"; tag = (actual fqin of image); };

services.podman.containers."my-container".extraConfig.Container.Image = (some name and tag in the registry);

```

Using the image in the container config directly (ie `my-container.image = "my-img.image";`) will require my follow up PR to this one: #6447.

bamhm182:
@rycee 

This should be good to go. Is there anything we can do to help it get merged?

PR #6139, opened 2024-11-25
Bash: move `bash.sessionVariables` into `.bashrc`

### LINKED ISSUE #5474: bug: `home.sessionVariables` and `bash.sessionVariables` end up serving the same purpose: `.profile`

### Issue description

When you set `home.sessionVariables` the variables go into `hm-session-vars.sh`, which is loaded by `.profile`.
When you set `bash.sessionVariables` the variables go directly into `.profile`, serving the same purpose as `home.sessionVariables`.

With that, we're left missing a way to set variables in `.bashrc`, or bash session variables. One would have to script them directly.

`bash.sessionVariables` should not go directly into `.profile`, but into `.bashrc`.

This line should be moved a few lines bellow (into `home.file.".bashrc".source `):

https://github.com/nix-community/home-manager/blob/0eb314b4f0ba337e88123e0b1e57ef58346aafd9/modules/programs/bash.nix#L215

### Maintainer CC

@rycee 

### System information

```markdown
  /nix/store/y11cs26gb9b1irv5hk776mra7hgy80bl-nix-info
copying path '/nix/store/y11cs26gb9b1irv5hk776mra7hgy80bl-nix-info' from 'https://cache.nixos.org'...
 - system: `"x86_64-linux"`
 - host os: `Linux 6.8.0-31-generic, Ubuntu, 24.04 LTS (Noble Numbat), nobuild`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.22.0`
 - channels(root): `"nixpkgs"`
 - nixpkgs: `/home/giggio/.nix-defexpr/channels/nixpkgs`
```
```

### PR DESCRIPTION
**Pull request by giggio**
# Tests are failing in the `master` branch, I can't be sure if my changes are correct.

When you set `home.sessionVariables` the variables go into `hm-session-vars.sh`, which is loaded by `.profile`.
When you set `bash.sessionVariables` the variables go directly into `.profile`, serving the same purpose as `home.sessionVariables`.

With that, we're left missing a way to set variables in `.bashrc`, or bash session variables. One would have to script them directly.

`bash.sessionVariables` should not go directly into `.profile`, but into `.bashrc`.

Closes #5474.

### Description

Moving `bash.sessionVariables` into `.bashrc`.

### Checklist

- [ x ] Code formatted with `./format`.

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/programs/bash.nix
+++ b/modules/programs/bash.nix
@@ -214,8 +214,6 @@ in {
     home.file.".profile".source = writeBashScript "profile" ''
       . "${config.home.profileDirectory}/etc/profile.d/hm-session-vars.sh"
 
-      ${sessionVarsStr}
-
       ${cfg.profileExtra}
     '';
 
@@ -224,6 +222,8 @@ in {
 
       # Commands that should be applied only for interactive shells.
       [[ $- == *i* ]] || return
+
+      ${sessionVarsStr}
 
       ${historyControlStr}
 
--- a/tests/modules/programs/bash/session-variables.nix
+++ b/tests/modules/programs/bash/session-variables.nix
@@ -17,13 +17,21 @@ with lib;
     nmt.script = ''
       assertFileExists home-files/.profile
       assertFileContent \
+        home-files/.bashrc \
+        ${
+          builtins.toFile "session-variables-expected" ''
+
+            export V1="v1"
+            export V2="v2-v1"
+
+
+          ''
+        }
+      assertFileContent \
         home-files/.profile \
         ${
           builtins.toFile "session-variables-expected" ''
             . "/home/hm-user/.nix-profile/etc/profile.d/hm-session-vars.sh"
-
-            export V1="v1"
-            export V2="v2-v1"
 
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by rycee** [comment]
They are intentionally loaded in .profile since they are associated with the login session. Bashrc is sourced for each shell instance. So perhaps it would be useful to have an option like `interactiveInstanceVariables` but I'm not certain it is necessary since it is pretty easy to add variables to bashrc. Something like

```nix
programs.bash.initExtra = config.lib.shell.exportAll {
  FOO = "BAR";
  BAZ = "BOO";
};
```

should work OK. Use `programs.bash.bashrcExtra` instead of you want the variables to apply for non-interactive shells as well.

giggio:
@rycee Don't you think it is confusing to have two variables that do the same thing: `home.sessionVariables` and `bash.sessionVariables`?

I am aware that you can work around it, and I have done so. It is not about that.

PR #6141, opened 2024-11-26
feat(zsh): initExtraLast option

### LINKED ISSUE #6140: Support `initExtraLast` option for Zsh?
We can use `initExtraFirst` to ensure that I start `.zshrc` file with the following lines.

```nix
    initExtraFirst = ''
      echo "# ----------------------------------------------------------------------"
      echo "# Init Extra First...."

      # time ZSH_ZPROF_DEBUG=1 zsh -i -c exit
      if [[ -n ''${ZSH_ZPROF_DEBUG} ]]; then
        zmodload zsh/zprof
      fi
    ''
```

For example, I want these lines to be the last four lines in the generated `.zshrc` file.

```text
# time ZSH_ZPROF_DEBUG=1 zsh -i -c exit
if [[ -n ${ZSH_ZPROF_DEBUG} ]]; then
  zprof
fi
```

How do we make sure that some configurations are set at the end of `.zshrc` file?

sheeeng:
### Description

Fix #6140.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/zsh.nix
+++ b/modules/programs/zsh.nix
@@ -469,6 +469,12 @@ in
         description = "Commands that should be added to top of {file}`.zshrc`.";
       };
 
+      initExtraLast = mkOption {
+        default = "";
+        type = types.lines;
+        description = "Commands that should be added to bottom of {file}`.zshrc`.";
+      };
+
       envExtra = mkOption {
         default = "";
         type = types.lines;
@@ -738,6 +744,10 @@ in
           }
         '')
 
+        cfg.initExtraLast
+
+        # zprof must place last after everything else, since it
+        # benchmarks the shell initialization.
         (optionalString cfg.zprof.enable
         ''
           zprof


teto:
cam't you use mkBefore/mkAfter instead ?

NovaViper:
@teto Doing it that way still doesn't let you add stuff to the very end of the file, it sticks it right above the aliases history subsearch and a couple of other zsh features. See [line 702](https://github.com/nix-community/home-manager/blob/65912bc6841cf420eb8c0a20e03df7cbbff5963f/modules/programs/zsh.nix#L702).

teto:
I was more thinking about `home.file."${relToDotDir ".zshrc"}".text = mkAfter ...`

rather than adding `initExtraBeforeCompInit` and `initExtraLast` . The current generator generates a lot of empty lines. I think it would be better to append to a zsh config with various `mkOrder XXX` such that one can insert its zsh config where more precisely where he wants.

teto:
this should be fixed by #6479

khaneliman:
Verified mkOrder replaces this with 

```nix
        programs.zsh.initContent = lib.mkOrder 5000 ''
          test
        '';
```

PR #6142, opened 2024-11-26
emacs: add proFileDirectory to daemon PATH

klchen0112:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/services/emacs.nix
```diff
@@ -1,9 +1,6 @@
 { config, lib, pkgs, ... }:
-
 with lib;
-
 let
-
```
rycee, line 6: Can revert these formatting changes.

tests/modules/services/emacs-darwin/basic.nix
```diff
@@ -1,6 +1,4 @@
-{ lib, pkgs, ... }:
-
-{
+{ lib, pkgs, ... }: {
```
rycee, line 1: Can revert the unrelated formatting changes.

klchen0112: Done

PR #6143, opened 2024-11-26
mbsync: stop using deprecated SSLType config

pks-t:
### Description

With mbsync 1.5.0, the SSLType configuration has been renamed to TLSType. While the old key still works, it is now deprecated and causes mbsync to print a warning to stderr.

Stop using SSLType in favor of TLSType to fix the warning.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@KarlJoad

### DIFF
--- a/modules/programs/mbsync.nix
+++ b/modules/programs/mbsync.nix
@@ -30,7 +30,7 @@ let
 
   genTlsConfig = tls:
     {
-      SSLType = if !tls.enable then
+      TLSType = if !tls.enable then
         "None"
       else if tls.useStartTls then
         "STARTTLS"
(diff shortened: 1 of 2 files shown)


KarlJoad:
There is more to this change than just a simple configuration option rename.

See the discussion on Issue #5884 and PR #5918.

I have not gotten around to finishing my PR up because of school and work. I plan to finish it by the end of this week.

Testing is slightly hampered because I do not use home-manager anymore, so I would appreciate you testing out the changes in #5918 to see if they work on your system.

pks-t:
Makes sense, let me close this PR then. Thanks!

PR #6148, opened 2024-11-28
tmux: add 'focusEvents'

ambroisie:
With `tmux-sensible` being disabled by default (#6091), add an easy toggle to enable this functionality.

Disabled by default, as in upstream `tmux` (I _could_ see an argument for enabling it by default, as `sensibleOnTop` was enabled by default and enabled this option unconditionally, but historically that's not been the way default option values have been chosen for this module).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/programs/tmux.nix
+++ b/modules/programs/tmux.nix
@@ -95,6 +95,7 @@ let
     ''}
 
     set  -g mouse             ${boolToStr cfg.mouse}
+    set  -g focus-events      ${boolToStr cfg.focusEvents}
     setw -g aggressive-resize ${boolToStr cfg.aggressiveResize}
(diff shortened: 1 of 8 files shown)


ambroisie:
CI failure is on a Podman test.

rycee:

Perhaps it would be good to add a news entry about the removal of the sensible defaults plugin and how to recover various behaviors, such as using your new option?

dbaynard:
> Perhaps it would be good to add a news entry about the removal of the sensible defaults plugin and how to recover various behaviors, such as using your new option?

Did this happen? I checked changlog entries and didn't clock anything. I'd be interested to know what else has been affected.

PR #6149, opened 2024-11-28
lorri: fix ReadWritePaths for new gcroots behavior

kmicklas:
### Description

Since https://github.com/nix-community/lorri/pull/128, lorri no longer uses this legacy path for gcroots, and instead delegates to the Nix daemon. Leaving the path here causes the service to fail to start if the directory does not exist on the user's system:

```
nov 28 11:49:14 meitan (lorri)[50058]: lorri.service: Failed to set up mount namespacing: /nix/var/nix/gcroots/per-user/kmicklas: No such file or directory
nov 28 11:49:14 meitan systemd[2265]: lorri.service: Main process exited, code=exited, status=226/NAMESPACE
nov 28 11:49:14 meitan systemd[2265]: lorri.service: Failed with result 'exit-code'.
```

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/lorri.nix
+++ b/modules/services/lorri.nix
@@ -59,7 +59,6 @@ in {
             "%C/lorri"
             # Needs %C/nix/fetcher-cache-v1.sqlite
             "%C/nix"
-            "/nix/var/nix/gcroots/per-user/%u"
           ];
           CacheDirectory = [ "lorri" ];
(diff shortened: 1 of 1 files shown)


rycee:
Thanks! Merged to master now.

PR #6151, opened 2024-11-28
zsh: make autosuggest strategy accept more options

NovaViper:
### Description
~The current implementation completely overrides any declaration of ZSH_AUTOSUGGEST_STRATEGY, which can break some user's settings (in my case, breaks adding abbr's suggestion strategy). New implementation now allows for users to define more than just `history`, `completion`, and `match_prev_cmd` and will auto append the variable if there are existing declarations of ZSH_AUTOSUGGEST_STRATEGY in the user's zshrc; and if there isn't any existing declaration, then simply just set the variable. Additionally, it can be left to empty (`[  ]`) which would completely disable the accompanying code; allowing users to fully decide where they want to place ZSH_AUTOSUGGEST_STRATEGY. And finally, use `typeset -U` to remove any possible duplicate values in the variable.~ The original scope of this PR was a much broader and complicated approach; simplified to allow `programs.zsh.autosuggestion.strategies` to be set to empty (`[]`) and disable the code. This would still allow for the option to declare `ZSH_AUTOSUGGEST_STRATEGY` in any of the `*Variables` module options available in Home-Manager

Edit: I did test the change in my flake and found that it works without causing issues to other parts of Home-Manager

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@JohnRTitor

### DIFF
--- a/modules/programs/zsh.nix
+++ b/modules/programs/zsh.nix
@@ -427,6 +427,9 @@ in
             - `match_prev_cmd`: Like `history`, but chooses the most recent match whose preceding history item matches
                 the most recently executed command. Note that this strategy won't work as expected with ZSH options that
                 don't preserve the history order such as `HIST_IGNORE_ALL_DUPS` or `HIST_EXPIRE_DUPS_FIRST`.
+
+            Setting the option to an empty list `[]` will make ZSH_AUTOSUGGESTION_STRATEGY not be set automatically,
+            allowing the variable to be declared in {option}`programs.zsh.localVariables` or {option}`programs.zsh.sessionVariables`
           '';
         };
       };
@@ -641,7 +644,10 @@ in
 
         (optionalString cfg.autosuggestion.enable ''
           source ${pkgs.zsh-autosuggestions}/share/zsh-autosuggestions/zsh-autosuggestions.zsh
-          ZSH_AUTOSUGGEST_STRATEGY=(${concatStringsSep " " cfg.autosuggestion.strategy})
+          ${optionalString (cfg.autosuggestion.strategy != []) ''
+            ZSH_AUTOSUGGEST_STRATEGY=(${concatStringsSep " " cfg.autosuggestion.strategy})
+          ''
+          }
         '')
         (optionalString (cfg.autosuggestion.enable && cfg.autosuggestion.highlight != null) ''
           ZSH_AUTOSUGGEST_HIGHLIGHT_STYLE="${cfg.autosuggestion.highlight}"


rycee:
Thanks for the contribution, looks OK to me but I don't use zsh so cannot verify the functionality.

modules/programs/zsh.nix
```diff
@@ -641,7 +641,18 @@ in
 
         (optionalString cfg.autosuggestion.enable ''
           source ${pkgs.zsh-autosuggestions}/share/zsh-autosuggestions/zsh-autosuggestions.zsh
-          ZSH_AUTOSUGGEST_STRATEGY=(${concatStringsSep " " cfg.autosuggestion.strategy})
+          # Make strategy configuration optional in case the user wants absolutely
+          # full control with when and where zsh-autosuggestions' strategy variable
+          # is placed in the zshrc
+          ${optionalString (cfg.autosuggestion.strategy != []) ''
```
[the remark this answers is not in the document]
NovaViper: Done!

NovaViper:
> Thanks for the contribution, looks OK to me but I don't use zsh so cannot verify the functionality.

No problem! Also, I did give it a test with my configs (made my flake use my fork as the home-manager input) and it worked on my system; saw no issues as a result of the changes.

NovaViper:
Updated branch to match home-manager's master branch.

NovaViper:
Gave the branch another test in my flake and the code still works with no issues (enabled zsh, autin, fzf, oh-my-posh-zsh-abbr, bat, zoxide, eza, fd and dircolors). Also tried with mcfly/mcfly fzf, still found no conflicts with other cli plugins

modules/programs/zsh.nix
```diff
@@ -415,7 +415,7 @@ in
         };
 
         strategy = mkOption {
-          type = types.listOf (types.enum [ "history" "completion" "match_prev_cmd" ]);
+          type = types.listOf types.str;
```
JohnRTitor, line 418: Why accept any arbitrary string value?

AFAIK, only these three options are available. https://github.com/zsh-users/zsh-autosuggestions/blob/master/README.md#suggestion-strategy

JohnRTitor:
> Additionally, it can be left to empty (`[ ]`) which would completely disable the accompanying code;

In that case, I think just checking if it's a empty list is good for now. A user should use this option if they want to use any of the built in strategies. Or they can just disable the accompanying code completely, by setting it to an empty  string.

modules/programs/zsh.nix
```diff
@@ -641,7 +641,18 @@ in
 
         (optionalString cfg.autosuggestion.enable ''
           source ${pkgs.zsh-autosuggestions}/share/zsh-autosuggestions/zsh-autosuggestions.zsh
-          ZSH_AUTOSUGGEST_STRATEGY=(${concatStringsSep " " cfg.autosuggestion.strategy})
+          ${optionalString (cfg.autosuggestion.strategy != []) ''
+            # Make strategy configuration optional in case the user wants absolutely
+            # full control with when and where zsh-autosuggestions' strategy variable
+            # is placed in the zshrc
+            if [ -n "''${ZSH_AUTOSUGGEST_STRATEGY:-}" ]; then
+              typeset -U ZSH_AUTOSUGGEST_STRATEGY # Remove any duplicate entries
```
JohnRTitor, line 649: We could use lib.unique for it.

JohnRTitor (commented):
Honestly, the sheer amount of code for this in zshrc not worth it. Duplicates can be removed by using `lib.unique` at nix eval time, and checking if it's an empty list is okay IMO.

modules/programs/zsh.nix
NovaViper: >Because I notice that even the zsh autosuggestion plugin itself actually creates the ZSH_AUTOSUGGEST_STRATEGY variable itself, which adds the duplicate variables if you simply want to append the variable, why I had to add `typeset -U`

This one here in particular I was referring to: https://github.com/zsh-users/zsh-autosuggestions/blob/0e810e5afa27acbd074398eefbe28d13005dbc15/src/config.zsh#L18-L21

Also now I'm considering taking out the entire boolean check altogether since the plugin itself adds the variable in the first place and just leave it as simply this 🤔 
```
typeset -U ZSH_AUTOSUGGEST_STRATEGY # Remove any duplicate entries
ZSH_AUTOSUGGEST_STRATEGY=(${concatStringsSep " " cfg.autosuggestion.strategy} "''${ZSH_AUTOSUGGEST_STRATEGY[@]}")
```

NovaViper:
> Honestly, the sheer amount of code for this in zshrc not worth it. Duplicates can be removed by using `lib.unique` at nix eval time

The duplicates issue sadly isn't something that Nix can handle directly, because its' something that happens as zsh is loaded along with accompanying plugins, not during Nix evaluation time.

> checking if it's an empty list is okay IMO.
That could work too though it would be nice to have it directly in the `strategies` option because then if say a new plugin module gets added and it has a autosuggestion strategy then that module could append that strategy in the list with the option. I just thought about that use case on the spot, am going to see if that's something that's possible to do 🤔

teto:
> The current implementation completely overrides any declaration of ZSH_AUTOSUGGEST_STRATEGY, which can break some user's settings (in my case, breaks adding abbr's suggestion strategy). New implementation now allows for users to define more than just history, completion, and match_prev_cmd and will auto append the variable if there are existing declarations of ZSH_AUTOSUGGEST_STRATEGY in the user's zshrc; and if there isn't any existing declaration, then simply just set the variable. 

so what you want is to not define the value via nix ? what's the point of defining one via nix if you ignore it under some conditions. You lose the determinism. So probably a fix here is to accept null as a possible type ?

NovaViper:
> so what you want is to not define the value via nix ? what's the point of defining one via nix if you ignore it under some conditions. You lose the determinism. So probably a fix here is to accept null as a possible type ?

No, the goal is to allow for the strategies to either use any of the `*variables` module options, directly in `programs.zsh.initExtra*`, or in `programs.zsh.autosuggestion.strategies`. The original way completely disallows for any tweaking of the option outside of `programs.zsh.autosuggestion.strategies`, but that option only lets you put a set number of defaults ("history" "completion" "match_prev_cmd"); it doesn't allow for cases when someone declares a plugin via any of the zsh plugin modules available in Home-Manager (and some of those plugins require you to set `ZSH_AUTOSUGGESTION_STRATEGY` in a particular way.

NovaViper:
Yeah having the option to set it null would also work too, since that would allow users to set the variable using any of the `*variables` modules.

NovaViper:
I was mainly trying to go for the approach of allowing someone to choose how they want to declare the strategies using either the native `programs.zsh.autosuggestion.strategies`, any of the `*variables` modules, or `programs.zsh.initExtra`; basically flexibility in how they want to define that variable. Granted I do admit the code that resulted from my initial attempt does leave alot to be desired 😓

NovaViper:
Took the simpliest approach to the problem; now `programs.zsh.autosuggestion.strategies` can be left empty so the environment variable isn't overwritten anymore and users can go ahead and set `ZSH_AUTOSUGGEST_STRATEGY` in the `*variables` modules 👍🏾

rycee:

PR #6155, opened 2024-11-30
starship: add option to use community presets

kpbaks:
### Description

Presets are community-submitted configurations for starship. See: https://starship.rs/presets/#presets.

This PR adds the option `programs.starship.preset` that can be set to the name of a preset. If `programs.starship.preset != null` it will have priority over `programs.starship.settings`.

The preset is fetched from starships [github repo](https://github.com/starship/starship) by using `pkgs.fetchFromGithub`. I do not think this implementation detail is ideal, as the version fetched from the repo can be out of sync with `programs.starship.package`. This can lead to a bad user experience where a user has seen a preset online and wants to use it, but cannot because the revision is out of date. I am not experienced enough with nix and nixpkgs to know of an alternative. You can get the TOML of the preset with `starship preset <name>` where `<name>` is one of the items in `starship preset --list`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/starship.nix
+++ b/modules/programs/starship.nix
@@ -11,7 +11,7 @@ let
   starshipCmd = "${config.home.profileDirectory}/bin/starship";
 
 in {
-  meta.maintainers = [ ];
+  meta.maintainers = [ maintainers.kpbaks ];
 
   options.programs.starship = {
     enable = mkEnableOption "starship";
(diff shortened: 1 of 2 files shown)


teto:
I would like to have this so I will merge  (once it works:) ).

modules/programs/starship.nix
```diff
@@ -88,9 +118,25 @@ in {
...
+    xdg.configFile."starship.toml" = if cfg.preset != null then
+      let
+        starshipGithub = pkgs.fetchFromGitHub {
+          owner = "starship";
+          repo = "starship";
+          rev = "61c860e1293d446d515203ac44055f7bef77d14a";
+          hash = "sha256-pl3aW4zCM6CcYOL0dUwE56aTC7BJdvdRyo/GudvX7fQ=";
+        };
+      in {
+        text = builtins.readFile
```
ambroisie: Let's avoid the IFD and use `source = ...`?

modules/programs/starship.nix
```diff
@@ -23,6 +23,34 @@ in {
...
+        "jetpack"
+        "nerd-font-symbols"
+        "no-empty-icons"
+        "no-nerd-font"
+        "no-runtime-versions"
+        "pastel-powerline"
+        "plain-text-symbols"
+        "pure-preset"
+        "tokyo-night"
+      ];
```
ambroisie: Does it _need_ to be an exhaustive enum? How often is that list likely to change?

ambroisie (changes requested):

modules/programs/starship.nix
kpbaks: Done

modules/programs/starship.nix
kpbaks: I have changed the type to `types.nullOr types.str`. It appears the number of presents change infrequently, but you are right that it is extra maintenance to keep track of it.

PR #6172, opened 2024-12-04
Fix assertion that `config.nixpkgs` cannot be set if `useGlobalPkgs` is enabled

### LINKED ISSUE #6079: while `useGlobalPkgs=true`, `nixpkgs.config` can be set without error message if `nixpkgs.overlays` is null (and vice versa)
### Description

When `home-manager.useGlobalPkgs = true`, setting `nixpkgs.overlays` in the home configuration has _no effect_. I unknowingly had `useGlobalPkgs` enabled in my config and struggled for days with overlays not applying. There should be a warning emitted if a user tries to add an overlay to nixpkgs in their home configuration while `useGlobalPkgs` is enabled, informing them of this relationship.

justDeeevin:
### Description

When `useGlobalPkgs` is enabled, setting `nixpkgs.config` and `nixpkgs.overlays` in the home config has no effect. There is an assertion in place intended to throw an error if the user tries to set these options in this scenario, but it uses an OR instead of an AND, allowing for _one_ of them to be set. This PR replaces this OR for an AND.

Fixes #6079.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@thiagokokada

### DIFF
--- a/modules/misc/nixpkgs-disabled.nix
+++ b/modules/misc/nixpkgs-disabled.nix
@@ -64,7 +64,7 @@ in {
 
   config = {
     assertions = [{
-      assertion = cfg.config == null || cfg.overlays == null;
+      assertion = cfg.config == null && cfg.overlays == null;
       message = ''
         `nixpkgs` options are disabled when `home-manager.useGlobalPkgs` is enabled.
       '';


thiagokokada (approved):

justDeeevin:
Been a minute... what's left to get this merged?

thiagokokada:
CC @rycee .

Zerodya:
I updated my flake.lock this morning only to get this failed assertions error:
```
error:
         at /nix/store/brjzz8wv5k09bq0yrvhc1j4hd8677k21-source/lib/attrsets.nix:1574:11:
         1574|           head values
             |           ^
         1575|         else

         at /nix/store/brjzz8wv5k09bq0yrvhc1j4hd8677k21-source/lib/modules.nix:853:9:
          852|     in warnDeprecation opt //
          853|       { value = addErrorContext "while evaluating the option `${showOption loc}':" value;
             |         ^
          854|         inherit (res.defsFinal') highestPrio;

       error:
       Failed assertions:
       - alpha profile: `nixpkgs` options are disabled when `home-manager.useGlobalPkgs` is enabled.
```
Now, I haven't found any nixpkgs options in my home config, and we all know how unhelpful nix errors are. So maybe this should've been a warning and not an error, since I can't update my system anymore even though it was not broken before.

Anyway, I set `useGlobalPkgs = false;` in my flake, but now home-manager can't see the `allowUnfreePredicate` I have set in my system and complains that it can't install some of my packages. I couldn't find any documentation about how to allow unfree packages in home-manager so maybe someone here could help me with this?

panchoh:
@Zerodya, if you use stylix, check [this issue](https://github.com/danth/stylix/issues/865).

Zerodya:
Ah, I do. Thank you for pointing me there, I will follow that issue now.

gaelj:
Same problem here (home manager + stylix + hyprpanel). Hyprpanel is impacted as well.

+1 for it to be a warning rather than a hard error, so that we can continue to update home manager until all concerned upstreams have fixed the bug on their side.

tuxiqae:
In my case it broke my `allowUnfree` settings:
```nix
nixpkgs.config.allowUnfree = true;
nixpkgs.config.allowUnfreePredicate = _: true;
```

```
- <myUser> profile: `nixpkgs` options are disabled when `home-manager.useGlobalPkgs` is enabled.
```

Not sure why, since it's not like it was broken before I guess.

brckd:
The error has been removed temporarily removed as of #6466, so updating Home Manager should fix it, but make sure to adapt your configurations nonetheless.

gaelj:
Here is @Silveere 's solution adapted for setups which don't involve `flake-parts`

- set `useGlobalPkgs = false;`
- in each user's home-manager module, add :

```nix
    nixpkgs = {
      config = lib.mapAttrs (n: v: lib.mkDefault v) osConfig.nixpkgs.config;
      # mkOrder 900 is after mkBefore but before default order
      overlays = lib.mkOrder 900 osConfig.nixpkgs.overlays;
    };
```

It silences of the warning, but it really should be fixed in all upstreams projects.

justDeeevin:
I'm very sorry i caused so much of a stirr with this change -- good learning experience to consider the impact of my code, i suppose

brckd:
> I'm very sorry i caused so much of a stirr with this change -- good learning experience to consider the impact of my code, i suppose

Don't worry, this was a learning experience for all of us. You fixed a bug which was clearly not intended and needed to be addressed. Although an intermediate warning would have been nice, this quickly urged configs and libraries to be fixed, which they would have had to anyways. Luckily it seems pretty much only Stylix users were affected by this and it was all resolved within a day or two. But you're right, it really shows what unforeseen consequences a simple change can have!

RossSmyth:
Yeah, how are you supposed to modify the nixpkgs config then? It would be good if that was documented somewhere.

brckd:
@RossSmyth You're using `useGlobalPkgs`, which means that Home Manager will use the Nixpkgs instance NixOS uses. Therefore you have to move `nixpkgs.allowUnfree` and all your other Nixpkgs configurations into your NixOS module.

Issue #5010, opened 2024-02-08
bug: Thunderbird handling multiple Identites wrong


isiko:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

I have a mail account with multiple identities associated with it. Although I have different settings for the different IDs, Home Manager ignores this and just uses the default ID configuration for all IDs. The options I notice not being applied correctly are `accounts.email.accounts.<name>.signature`, `accounts.email.accounts.<name>.smtp`, but there may be more.

Here's an excerpt from my config:
```Nix
"Main" = {
    address = "main@domain1.com";
    primary = true;

    aliases = [
        "alias1@domain1.com"
        "alias2@domain2.com"
    ];

    userName = "***";

    flavor = email-config1.flavor;
    imap = email-config1.imap;
    smtp = email-config1.smtp;

    realName = name.full;
    signature = /*Signature 1 */;
    thunderbird = {
        enable = true;
        profiles = [ "Full" ];
        perIdentitySettings = id : {
            "mail.identity.id_${id}.sign_mail" = true;
            "mail.identity.id_${id}.encryptionpolicy" = 2;
        };
    };
};

"Personal" = {
    address = "alias2@domain2.com";
    primary = false;

    userName = "***";

    flavor = email-config2.flavor;
    smtp = email-config2.smtp;

    realName = name.full;
    signature = /*Signature 2 */;
    thunderbird = {
        enable = true;
        profiles = [ "Full" ];
        perIdentitySettings = id : {
            "mail.identity.id_${id}.sign_mail" = true;
            "mail.identity.id_${id}.encryptionpolicy" = 2;
        };
    };
};
```

### Maintainer CC

From `git blame` I think I should ping @teto @rycee 

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.1.76, NixOS, 24.05 (Uakari), 24.05.20240205.faf912b`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.18.1`
 - channels(root): `"home-manager, nixos, nixos-unstable"`
 - channels(isiko404): `"nixos-stable-23.05, nixos-unstable"`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`
```

teto:
[...]

PR #6177, opened 2024-12-07
thunderbird: set the correct SMTP server for aliases

alois31:
Previously, only the main identity of an account would get the proper SMTP server assigned. Identities corresponding to aliases would not get an SMTP server assigned at all, leading to a (Thunderbird-internal) fallback to the SMTP server associated to the primary account. This is obviously wrong for non-primary accounts having aliases associated to them. Fix it by specifying the SMTP server explicitly for all identities.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/thunderbird.nix
+++ b/modules/programs/thunderbird.nix
@@ -71,6 +71,8 @@ let
       "mail.identity.id_${id}.openpgp_key_id" = account.gpg.key;
       "mail.identity.id_${id}.protectSubject" = true;
       "mail.identity.id_${id}.sign_mail" = account.gpg.signByDefault;
+    } // optionalAttrs (account.smtp != null) {
+      "mail.identity.id_${id}.smtpServer" = "smtp_${account.id}";
     } // account.thunderbird.perIdentitySettings id;
 
   toThunderbirdAccount = account: profile:
@@ -103,7 +105,6 @@ let
       "mail.server.server_${id}.type" = "imap";
       "mail.server.server_${id}.userName" = account.userName;
     } // optionalAttrs (account.smtp != null) {
-      "mail.identity.id_${id}.smtpServer" = "smtp_${id}";
       "mail.smtpserver.smtp_${id}.authMethod" = 3;
       "mail.smtpserver.smtp_${id}.hostname" = account.smtp.host;
       "mail.smtpserver.smtp_${id}.port" =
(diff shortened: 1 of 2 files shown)


teto:
trusting you on this.

zmberber:
I am linking my comment for an issue relevant to this merge:  https://github.com/nix-community/home-manager/issues/5010#issuecomment-2538777893

This change definitely makes things better, but I still wonder why the `if` case in https://github.com/nix-community/home-manager/blob/6e5b2d9e8014b5572e3367937a329e7053458d34/modules/programs/thunderbird.nix#L48-L53 is done.

It doesn't seem necessary in this way.  Maybe I should do another issue on this.

zmberber:
Ok after further looking into it, I seem to have found the reason why this `if` case is done:

We have this part here https://github.com/nix-community/home-manager/blob/6e5b2d9e8014b5572e3367937a329e7053458d34/modules/programs/thunderbird.nix#L79-L82, where we both have the main address, which is in `account.address`, and the aliases, which is in `account.aliases`.

the account name must be associated with the main address, so here this `if` case in https://github.com/nix-community/home-manager/blob/6e5b2d9e8014b5572e3367937a329e7053458d34/modules/programs/thunderbird.nix#L48-L53 checks this, and therefore uses the hash of the account name, so that in the relevant parts, one can use the same id for both the account and the primary address.

I am not sure if I like the way it is done here, it almost seems like a hack for something (that i still cannot find).  But okay, it works.

I do think one could separate it, and only use the hashed account name for settings such as those involving `mail.server.server_${id}` or `mail.account.account_${id}` etc., and always use the hashed address for settings such as those involving `mail.identity.id_${id}`.

If I am not mistaken (since i cannot find where it is needed that the `if` case is true and we need the id of the account to be the same as the one of the "primary" address/identity), the code should completely work without that `if` statement I talked about.

I would need to do some tests.

zmberber:
I see one reason why the `if` could make sense:

If an address listed somewhere as an alias (meaning that hm creates a new identity in thunderbird), one can create a new identity _for the same email address_ by defining a new account in hm, and having that address as the address option (as opposed to being listed there as an alias).  Then there are two identities with the same address.  This is precisely what the `if` case does, for a non-alias defined address, it has an id that is created in such a way that it doesn't clash with an id created via an alias.

However, if one allows this, then one could also think about allowing multiple accounts having the same address listed in their aliases.  This however, does not really work in home-manager that well, because it takes the hash of the email address for this, and these hashes would clash.

Therefore, I also think this `if` is not really that helpful (and makes things more confusing).

I would almost suggest the following:  each alias listed for an account leads to hm creating a thunderbird mail identity with id generated via a hash that involves both the account and the email address.

Not sure if this is reasonable.

PR #6181, opened 2024-12-08
qt: install kio when qt.platformTheme = "kde"

alyraffauf:
### Description

Following discussions at https://github.com/NixOS/nixpkgs/pull/362949, this PR fixes file dialogs when KDE is not installed but `qt.platformTheme = "kde"` (common for theming purposes). Without it, file dialogs fail to show files and folders. Only an issue when KDE is not installed, because KDE installs kio in its own module.

image

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@rycee @thiagokokada


doronbehar:
There is also a NixOS module option `qt.platformTheme`. Perhaps kde should be added there as well?

https://github.com/NixOS/nixpkgs/blob/65484f88b2c2175b3d06a8b5fd199bb852eee5e3/nixos/modules/config/qt.nix#L9

rycee:

PR #6184, opened 2024-12-08
nushell: Structured settings

### LINKED ISSUE #5527: bug: nushell module configuration overwrites default basic configuration

JoaquinTrinanes:
### Description

Closes #5527 (but it should be fixed on new nushell versions regardless of this PR if `programs.nushell.configFile` is set in an adequate way)

- After https://github.com/nushell/nushell/pull/14249 nushell loads the default config before the user config, removing the need to explicitly set it.
- After #5905, it is now possible to convert structured nix values into nushell code.

Hence, this PR reintroduces the `programs.nushell.settings` option. It then flattens the config and assigns each value to `$env.config`. The flattening step is done to avoid overwriting previously defined config values (otherwise, `completions.quick = true` would become `$env.config.completions = { quick: true }`, deleting all other `completions` config).

```nix
{
  show_banner = false;
  completions.external = {
    enable = true;
    max_results = 200;
  };
}
```
becomes
```nushell
$env.config.completions.external.enable = true
$env.config.completions.external.max_results = 200
$env.config.show_banner = false
```

One thing I'm wondering is if the code should have a special cases for lists/records, and append/merge them instead of assigning them. This could be implemented on a followup PR if we see the need 🤔 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@Philipp-M @aidalgol


modules/programs/nushell.nix
```diff
@@ -167,19 +180,35 @@ in {
...
+                    (map (key: unravel (prefixes ++ [ key ]) value.${key})
+                      (builtins.attrNames value))
+                  else
+                    lib.nameValuePair (lib.concatStringsSep "." prefixes) value;
+              in lib.listToAttrs (unravel [ ] settings);
+
+            flattenedSettings = flattenSettings cfg.settings;
+          in lib.mkIf (cfg.settings != { }) (lib.concatLines (lib.mapAttrsToList
+            (key: value: "$env.config.${key} = ${toNushell { } value}")
+            flattenedSettings)))
```
rycee, line 211: An alternative implementation that I personally find slightly more readable and is potentially a bit more efficient since it avoids the intermediate attribute set (although I imagine you'll need many settings for it to be noticeable 🙂).

I haven't tested it beyond the test case and feel free to reject the suggestion if you prefer the original version:

```suggestion

          (let
            flattenSettings = settings:
              let
                joinDot = a: b: "${if a == "" then "" else "${a}."}${b}";
                unravel = prefix: value:
                  if lib.isAttrs value && !isNushellInline value then
                    lib.concatMap
                    (key: unravel (joinDot prefix key) value.${key})
                    (builtins.attrNames value)
                  else
                    [ (lib.nameValuePair prefix value) ];
              in unravel "" settings;

            mkLine = { name, value }: ''
              $env.config.${name} = ${toNushell { } value}
            '';

            settingsLines =
              lib.concatMapStrings mkLine (flattenSettings cfg.settings);
          in lib.mkIf (cfg.settings != { }) settingsLines)

```

JoaquinTrinanes: Ooh thanks! I prefer your implementation, it's _way_ easier to understand 😄

nyabinary:
bump @rycee :3

JoaquinTrinanes:
@nyabinary do you think anything in this PR needs to change due to the release? The only thing I can think of right now is changing some examples to use `const NU_*` instead of `$env.NU_*`.

Also, [according to the new docs](https://www.nushell.sh/book/configuration.html#configuration-overview) using the `env.nu` file is not a "best practice" anymore. I could make it so that env vars are written to `config.nu` instead. But that should be a separate PR 🤔

nyabinary:
> @nyabinary do you think anything in this PR needs to change due to the release? The only thing I can think of right now is changing some examples to use `const NU_*` instead of `$env.NU_*`.
> 
> Also, [according to the new docs](https://www.nushell.sh/book/configuration.html#configuration-overview) using the `env.nu` file is not a "best practice" anymore. I could make it so that env vars are written to `config.nu` instead. But that should be a separate PR 🤔

Would be better to just do it in this PR tbh, maybe rename it to `nushell: refactor`?

JoaquinTrinanes:
@nyabinary it was a simple-enough change that I included it in this PR to avoid the extra overhead ✨

modules/programs/nushell.nix
```diff
@@ -177,39 +193,56 @@ in {
...
 
-    home.file = mkMerge [
+    home.file = lib.mkMerge [
       (let
         writeConfig = cfg.configFile != null || cfg.extraConfig != ""
-          || aliasesStr != "";
+          || aliasesStr != "" || cfg.settings != { };
+
+        aliasesStr = lib.concatLines
+          (lib.mapAttrsToList (k: v: "alias ${k} = ${v}") cfg.shellAliases);
```
Spoonbaker, line 205:
```suggestion
          (lib.mapAttrsToList (k: v: "alias '${k}' = ${v}") cfg.shellAliases);
```
This should make it possible to define aliases with spaces. I figured it might make sense to include in this PR, if that's alright.

JoaquinTrinanes: Good call, I never thought about multi word aliases being valid. Added support in [cf27650](https://github.com/nix-community/home-manager/pull/6184/commits/cf27650610d793c4b17dbc207eac75d4e1feefc7)!

JoaquinTrinanes: I used `toNushell` instead of manually adding quotes because quotes are valid parts of an alias name apparently. They don't work in some cases but that's a nushell issue, not a config issue
```nushell
# Works
alias 'a"a"a' = echo 'closed quotes work'
a"a"a

# Doesn't work
alias 'a"a' = echo 'echo opened quotes dont'

a"a
#   ┬
#   ╰── expected closing "
```

Spoonbaker: Awesome, thanks for including that!

JoaquinTrinanes:
@nyabinary it should be yeah

(Ping @rycee :) )

rycee:

PR #6187, opened 2024-12-09
atuin: Do not hardcode daemon socket path

mmlb:
### Description

Use systemd path specifiers for appropriate path prefixes instead of assuming paths. This PR is future compatible with 18.4.0 preferring XDG_RUNTIME_PATH over XDG_DATA_DIR.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@hawkw @water-sucks

### DIFF
--- a/modules/programs/atuin.nix
+++ b/modules/programs/atuin.nix
@@ -199,7 +199,7 @@ in {
           Unit = { Description = "Atuin daemon socket"; };
           Install = { WantedBy = [ "sockets.target" ]; };
           Socket = {
-            ListenStream = "%h/.local/share/atuin/atuin.sock";
+            ListenStream = "%D/atuin/atuin.sock";
             SocketMode = "0600";
             RemoveOnStop = true;
           };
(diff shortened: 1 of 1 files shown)


ambroisie (approved):

water-sucks (approved):
Thanks! I just resorted to the default location in my initial PR, but I should have tracked the location change for `$XDG_RUNTIME_DIR` in the Atuin issues right then.

rycee:

mmlb:
> Thanks! I just resorted to the default location in my initial PR, but I should have tracked the location change for `$XDG_RUNTIME_DIR` in the Atuin issues right then.

No problem, luckily my curiosity got to me and I wanted to check the implementation since everything I saw documented was for explicit paths not $XDG_DATA_DIR and I looked at code on main branch that had the new behavior and didn't make sense and then I found the PR :D

minhtrancccp:
i got the error of `attribute 'version' missing` when the atuin's `flake.nix` is used, i.e. `package` is set to `[flake inputs].atuin.packages.[system].default`. would there be any fix for this bug? many thanks.

PR #6195, opened 2024-12-11
install: allow customizing the shellHook

nikhil-057:
### Description

`nix-shell "<home-manager>" -A install` fails when a target path to be written by home-manager (eg `~/.profile`) already exists. This causes problems when I try to bootstrap non-nixos machines with home-manager.

There's currently no way to modify the install shellHook, which runs `exec home-manager init --switch --no-flake`. This PR allows customizing the shellHook with an override (for example, changing it to `exec home-manager init --switch --no-flake -b backup`).

I've also added an integration test for illustration, happy to remove if you think it's overkill.

### Checklist

- [x] Commit messages are formatted.

**Maintainer CC**
@rycee

### DIFF
--- a/home-manager/install.nix
+++ b/home-manager/install.nix
@@ -1,4 +1,6 @@
-{ home-manager, runCommand }:
+{ home-manager, runCommand
+, shellHook ? "exec ${home-manager}/bin/home-manager init --switch --no-flake"
+}:
 
 let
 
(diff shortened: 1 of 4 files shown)


home-manager/install.nix
```diff
@@ -1,4 +1,6 @@
-{ home-manager, runCommand }:
+{ home-manager, runCommand
+, shellHook ? "exec ${home-manager}/bin/home-manager init --switch --no-flake"
```
rycee, line 2: I'm not very comfortable allowing wholesale replacement of the shell hook. Perhaps it would be sufficient to add a parameter `extraSwitchOptions` or similar?

tests/integration/standalone/override-install.nix
```diff
@@ -0,0 +1,6 @@
+{ hmInstall ? (import <home-manager> { }).install }:
+hmInstall.override {
+  shellHook = ''
+    exec home-manager init --switch --no-flake -b backup
```
rycee, line 4: I'm surprised this works, looking at the code I don't see any handling of `-b` in the `doInit` function. It may work (mostly accidentally) if the `-b backup` is placed before the `init` sub-command.

PR #6196, opened 2024-12-11
firefox: add preConfig option

bandithedoge:
### Description

Adds a `programs.firefox.profiles.<name>.preConfig` option that mirrors `extraConfig`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee @brckd


bandithedoge:
> Are there any specific cases where this would be useful? Otherwise looks good.

Using user.js templates such as [arkenfox](https://github.com/arkenfox/user.js) or [Betterfox](https://github.com/yokoffing/Betterfox) and overriding their settings.

brckd:
> > Are there any specific cases where this would be useful? Otherwise looks good.
> 
> Using user.js templates such as [arkenfox](https://github.com/arkenfox/user.js) or [Betterfox](https://github.com/yokoffing/Betterfox) and overriding their settings.

Good idea, in that case, it's good to go!

brckd (approved):

HeitorAugustoLN:
> > Are there any specific cases where this would be useful? Otherwise looks good.
> 
> Using user.js templates such as [arkenfox](https://github.com/arkenfox/user.js) or [Betterfox](https://github.com/yokoffing/Betterfox) and overriding their settings.

In that case, might it be better to use [arkenfox-nixos](https://github.com/dwarfmaster/arkenfox-nixos) or [betterfox-nix](https://github.com/HeitorAugustoLN/betterfox-nix)? But this option looks good, though!

bandithedoge:
> > > Are there any specific cases where this would be useful? Otherwise looks good.
> > 
> > 
> > Using user.js templates such as [arkenfox](https://github.com/arkenfox/user.js) or [Betterfox](https://github.com/yokoffing/Betterfox) and overriding their settings.
> 
> In that case, might it be better to use [arkenfox-nixos](https://github.com/dwarfmaster/arkenfox-nixos) or [betterfox-nix](https://github.com/HeitorAugustoLN/betterfox-nix)? But this option looks good, though!

I've never heard of these projects but they do look really useful. I think this PR would enable a cleaner implementation, though, since you won't have to flatten all preferences into JS before home-manager does it anyway.

Kranzes:
This PR seems to duplicate the efforts of #5632, which happens to also have tests.

brckd:
Thanks for pointing that out, that didn't catch my radar! I'll try to mark this PR as a duplicate of #5632. The PR you've linked to seems more complete in general, but we should copy over this PR's news message.

nyabinary:
#5632 has been merged so this should be closed.

PR #6201, opened 2024-12-12
zed-editor: fix always generating settings.json

### LINKED ISSUE #6193: bug: HM always writes settings.json for zed

### Issue description

When using zed-editor, I've configured no settings or extensions but there is still a settings.json file generated.

I've figured out the issue comes from this https://github.com/nix-community/home-manager/blob/c7ffc9727d115e433fd884a62dc164b587ff651d/modules/programs/zed-editor.nix#L10-L11

Later on when the config is checked for being empty it will never be empty because auto_install_extensions will always be in mergedSettings even if there were none defined.

### Maintainer CC

_No response_

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.11.11-xanmod1, NixOS, 25.05 (Warbler), 25.05.20241209.a73246e`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Lix, like Nix) 2.92.0-dev-pre20241210-8ddcdfb
System type: x86_64-linux
Features: gc, signed-caches
User configuration files: /home/racci/.config/nix/nix.conf:/nix/store/dfn5fg6l506vm7n8cchfys27x4wj27iy-stylix-kde-config/nix/store/dfn5fg6l506vm7n8cchfys27x4wj27iy-stylix-kde-config/etc/xdg/nix/nix.conf:/home/racci/.nix-profile/etc/xdg/nix/nix.conf:/nix/profile/etc/xdg/nix/nix.conf:/home/racci/.local/state/nix/profile/etc/xdg/nix/nix.conf:/etc/profiles/per-user/racci/etc/xdg/nix/nix.conf:/nix/var/nix/profiles/default/etc/xdg/nix/nix.conf:/run/current-system/sw/etc/xdg/nix/nix.conf
Store directory: /nix/store
State directory: /nix/var/nix
Data directory: /nix/store/wpr7kirrxzz1l8kb17vwnqggl8dbcly4-lix-2.92.0-dev-pre20241210-8ddcdfb/share`
 - nixpkgs: `/nix/store/yw6kg4rb9v8s3ypjbpspig5r81m4lr5s-source`
```

DaRacci:
### Description

Makes it so settings.json no longer gets made by ensuring it can be empty as expected when there are no extensions or userSettings defined.

Previously settings.json would always be generated because even if there were no extensions defined an empty value would still be made.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@libewa 

Fixes #6193

### DIFF
--- a/modules/programs/zed-editor.nix
+++ b/modules/programs/zed-editor.nix
@@ -6,10 +6,11 @@ let
   cfg = config.programs.zed-editor;
   jsonFormat = pkgs.formats.json { };
 
-  mergedSettings = cfg.userSettings // {
-    # this part by @cmacrae
-    auto_install_extensions = lib.genAttrs cfg.extensions (_: true);
-  };
+  mergedSettings = cfg.userSettings
+    // (lib.optionalAttrs (builtins.length cfg.extensions > 0) {
+      # this part by @cmacrae
+      auto_install_extensions = lib.genAttrs cfg.extensions (_: true);
+    });
 in {
   meta.maintainers = [ hm.maintainers.libewa ];
 


DaRacci:
I've attempted to run the tests but have received the following error which i dont believe has anything to do with my changes but I'm not sure why is failing.

image

rycee:
Thanks! Rebased to master in 66c5d8b62818ec4c1edb3e941f55ef78df8141a8 🙂

PR #6206, opened 2024-12-14
git-worktree-switcher: init module

mateusauler:
### Description

Create new module for [git-worktree-switcher](https://github.com/mateusauler/git-worktree-switcher).

Note: I'm the upstream maintainer.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/lib/maintainers.nix
+++ b/modules/lib/maintainers.nix
@@ -258,6 +258,14 @@
     github = "LucasWagler";
     githubId = 32136449;
   };
+  mateusauler = {
+    name = "Mateus Auler";
+    email = "mateus@auler.dev";
+    github = "mateusauler";
+    githubId = 24767687;
+    keys =
+      [{ fingerprint = "A09D C093 3C37 4BFC 2B5A  269F 80A5 D62F 6EB7 D9F0"; }];
+  };
   matrss = {
     name = "Matthias Riße";
     email = "matrss@users.noreply.github.com";
--- a/modules/programs/git-worktree-switcher.nix
+++ b/modules/programs/git-worktree-switcher.nix
@@ -1,0 +1,52 @@
+{ pkgs, config, lib, ... }:
+
+let
+  inherit (lib) mkEnableOption mkOption mkPackageOption optionalString;
+
+  cfg = config.programs.git-worktree-switcher;
+
+  initScript = shell:
+    if (shell == "fish") then ''
+      ${lib.getExe pkgs.git-worktree-switcher} init ${shell} | source
+    '' else ''
+      eval "$(${lib.getExe pkgs.git-worktree-switcher} init ${shell})"
+    '';
+in {
+  meta.maintainers = [ lib.maintainers.mateusauler ];
+
+  options.programs.git-worktree-switcher = {
+    enable = mkEnableOption "git-worktree-switcher";
+    package = mkPackageOption pkgs "git-worktree-switcher" { };
+    enableBashIntegration = mkOption {
+      type = lib.types.bool;
+      default = true;
+      description = ''
+        Whether to enable git-worktree-switcher's Bash integration.
+      '';
+    };
+    enableFishIntegration = mkOption {
+      type = lib.types.bool;
+      default = config.programs.fish.enable;
+      description = ''
+        Whether to enable git-worktree-switcher's Fish integration.
+      '';
+    };
+    enableZshIntegration = mkOption {
+      type = lib.types.bool;
+      default = config.programs.zsh.enable;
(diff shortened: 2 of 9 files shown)


jiriks74:
You can drop https://github.com/nix-community/home-manager/commit/4c2c109beb112719dfbd0a3437ca58859f491306:

> If you are present in the nixpkgs maintainer list then you can use that entry. If you are not then you can add yourself to modules/lib/maintainers.nix in the Home Manager project.

modules/programs/git-worktree-switcher.nix
```diff
@@ -0,0 +1,52 @@
...
+  cfg = config.programs.git-worktree-switcher;
+
+  initScript = shell:
+    if (shell == "fish") then ''
+      ${lib.getExe pkgs.git-worktree-switcher} init ${shell} | source
+    '' else ''
+      eval "$(${lib.getExe pkgs.git-worktree-switcher} init ${shell})"
+    '';
+in {
+  meta.maintainers = [ lib.maintainers.mateusauler ];
```
jiriks74, line 15: The home-manager maintainer you set up isn't even used here. See how a `home-manager` maintainer is used elsewhere:

https://github.com/nix-community/home-manager/blob/8264bfe3a064d704c57df91e34b795b6ac7bad9e/modules/programs/thefuck.nix#L6

modules/lib/maintainers.nix
```diff
@@ -258,6 +258,14 @@
...
     githubId = 32136449;
   };
+  mateusauler = {
+    name = "Mateus Auler";
+    email = "mateus@auler.dev";
+    github = "mateusauler";
+    githubId = 24767687;
+    keys =
+      [{ fingerprint = "A09D C093 3C37 4BFC 2B5A  269F 80A5 D62F 6EB7 D9F0"; }];
+  };
```
jiriks74, line 268: As mentioned in the [Contributing](https://nix-community.github.io/home-manager/index.xhtml#_add_yourself_as_a_module_maintainer) section of the home-manager manual:

> If you are present in the nixpkgs maintainer list then you can use that entry. If you are not then you can add yourself to modules/lib/maintainers.nix in the Home Manager project.

jiriks74 (changes requested):
You should either remove the `home-manager` maintainer (probably the better course of action) or use it.

modules/lib/maintainers.nix
mateusauler: Missed this. Thank you!

mateusauler:
@jiriks74 Thank you for the review! Would you like to be added as a maintainer?

jiriks74:
> @jiriks74 Thank you for the review! Would you like to be added as a maintainer?

I would! Thanks!

mateusauler:
> I would! Thanks!

Added! :)

mateusauler:
I couldn't find it anywhere, should we explicitly ask someone to review this?

modules/programs/git-worktree-switcher.nix
```diff
@@ -0,0 +1,39 @@
...
+  };
+
+  config = lib.mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+    programs.bash.initExtra =
+      optionalString cfg.enableBashIntegration (initScript "bash");
+    programs.fish.interactiveShellInit =
+      optionalString cfg.enableFishIntegration (initScript "fish");
+    programs.zsh.initExtra =
+      optionalString cfg.enableZshIntegration (initScript "zsh");
```
ambroisie, line 37: This should not be the job of home manager, but instead packaged with `installShellCompletion`.

In fact it [already is](https://github.com/NixOS/nixpkgs/blob/64e75cd44acf21c7933d61d7721e812eac1b5a0a/pkgs/by-name/gi/git-worktree-switcher/package.nix#L47-L49). Making this option redundant.

ambroisie: Opened #6440.

jiriks74: As mentioned in https://github.com/nix-community/home-manager/pull/6440, the integrations are loading a function into the current shell which is necessary for this utility to work (to avoid creating subshells when changing directories) and therefore shouldn't be removed. This module works on the same ideas as others - allowing users to easily enable the utility in their Nix/Home-Manager configurations without creating a custom `.<insert_shell_here>rc` or `programs.<insert_shell_here>.initExtra`.

See
https://github.com/nix-community/home-manager/blob/83bd3a26ac0526ae04fa74df46738bb44b89dcdd/modules/programs/thefuck.nix#L37-L43

PR #6209, opened 2024-12-15
fish: source event handling functions on shell init

MrTipson:
### Description

Functions that contain event handler switches should be sourced during init, otherwise they become active only after being called manually.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -228,6 +228,15 @@ let
       echo "end"
       echo "setup_hm_session_vars") > $out
     '';
+  sourceHandlersStr = let
+    handlerAttrs =
+      [ "onJobExit" "onProcessExit" "onVariable" "onSignal" "onEvent" ];
+    isHandler = name: def:
+      isAttrs def && any (attr: hasAttr attr def) handlerAttrs;
+    handlerFunctions = filterAttrs isHandler cfg.functions;
+    sourceFunction = name: def:
+      "source ${config.xdg.configHome}/fish/functions/${name}.fish";
+  in concatStringsSep "\n" (mapAttrsToList sourceFunction handlerFunctions);
 
 in {
   imports = [
@@ -474,6 +483,9 @@ in {
         set -g __fish_home_manager_config_sourced 1
 
         source ${translatedSessionVariables}
+
+        # Source handler functions
+        ${sourceHandlersStr}
 
         ${cfg.shellInit}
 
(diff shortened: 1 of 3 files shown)


modules/programs/fish.nix
```diff
@@ -228,6 +228,15 @@ let
...
     '';
+  sourceHandlersStr = let
+    handlerAttrs =
+      [ "onJobExit" "onProcessExit" "onVariable" "onSignal" "onEvent" ];
+    isHandler = name: def:
+      isAttrs def && any (attr: hasAttr attr def) handlerAttrs;
+    handlerFunctions = filterAttrs isHandler cfg.functions;
+    sourceFunction = name: def:
+      "source ${config.xdg.configHome}/fish/functions/${name}.fish";
+  in concatStringsSep "\n" (mapAttrsToList sourceFunction handlerFunctions);
```
teto: wont the source trigger errors if files dont exist (like bash does) ?

teto:
Maybe a more radical change like the zsh one https://github.com/nix-community/home-manager/pull/6664 would address this issue (and a lot more) in a more legant manner ?

modules/programs/fish.nix
MrTipson: I don't see how the files wouldn't exist if these are functions generated by HM. I suppose something could change with the naming but I don't know how much its worth worrying about that.

MrTipson:
I'm not that familiar with the other config setups for shells, but that PR seems to be dealing more with merging multiple init* options, which I don't believe applies here.

modules/programs/fish.nix
teto: ok I see that the module creates files for each function so the file should always exist. Is the "1 function <-> 1 file" a common fish practice ? 
I guess it's ok to merge (needs a rebase though). I would welcome a fish user review.

MrTipson: Sorry, I don't know how I missed your response until now. Working on rebasing and testing atm, but I should be done with it soon.

I'm no expert myself, but the [fish docs](https://fishshell.com/docs/current/language.html#autoloading-functions) do mention that
> As we’ve explained, autoload files are loaded by name, so, while you can put multiple functions into one file, the file will only be loaded automatically once you try to execute the one that shares the name.

So I think its necessary for them to be in their own files, as a user will expect them to be available. Event handlers are also mentioned a bit later (that they don't get autoloaded).

teto:
opened in 2024 heh... I just tried the PR locally as a fish noob and looks like it makes sense ty. You can ping me on your possible next fish PR, will try to speed merging a bit :)

PR #6213, opened 2024-12-17
lutris: configuration init

Steinhagen:
### Description

Initial work to bring home-manager configuration to Lutris.
Due to some limitation on how Lutris rewrites on runtime the .conf file, only the runners and the extra packages will be addressed in this patchset.
With this change, you can automatically configure Lutris to execute a number of runners using nixpkgs instead of relying on the Appimages.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/lutris.nix
+++ b/modules/programs/lutris.nix
@@ -1,0 +1,152 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+let
+  cfg = config.programs.lutris;
+
+  cemuConf = ''
+    cemu:
+      runner_executable: ${lib.getExe pkgs.cemu}
+  '';
+  dolphinConf = ''
+    dolphin:
+      runner_executable: ${lib.getExe cfg.runners.dolphin.package}
+  '';
+  duckstationConf = ''
+    duckstation:
+      runner_executable: ${lib.getExe pkgs.duckstation}
+  '';
+  pcsx2Conf = ''
+    pcsx2:
+      runner_executable: ${lib.getExe pkgs.pcsx2}
+      nogui: ${boolToString cfg.runners.pcsx2.noGui}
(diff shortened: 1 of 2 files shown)


modules/lib/maintainers.nix
```diff
@@ -502,6 +502,12 @@
     github = "shimunn";
     githubId = 41011289;
   };
+  rapiteanu = {
+    name = "Viorel-Cătălin Răpițeanu";
+    email = "rapiteanu.catalin@gmail.com";
+    github = "Steinhagen";
+    githubId = 4029937;
+  };
```
rycee, line 510: Since you are already in the Nixpkgs maintainers list you don't need to also add yourself here.
```suggestion
```

modules/programs/lutris.nix
```diff
@@ -0,0 +1,152 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+let
+  cfg = config.programs.lutris;
+
+  cemuConf = ''
+    cemu:
```
rycee, line 8: Doing yaml configurations in this way will prevent users of the module from doing settings that are not provided in the module. Are you sure that it won't be necessary to allow users deeper access?

modules/lib/maintainers.nix
Steinhagen: I will remove that.

modules/programs/lutris.nix
Steinhagen: There will be a need to allow users deeper access to these. I was thinking of adding an extraConf part for cemu, dolphin, duckstation and ppsspp that is user modfiable. If there is a better method for this, please let me know. Also, any example is greatly appreciated!

khaneliman:
This looks like it should use freeform submodule options with common options so we can abstract away the details more to support all different kinds of runners someone might want to configure.

PR #6220, opened 2024-12-20
yazi: Fix example option values for settings

Adda0:
### Description

Yazi changed the allowed values for:
- its [setting `manager.sort_by`](https://yazi-rs.github.io/docs/configuration/yazi/#manager.sort_by). Option `modified` is now called `mtime`.
- its [setting `theme.filetype.rules.mime-type`](https://yazi-rs.github.io/docs/configuration/theme#filetype) to no longer use `x-` prefix for mime types. For example, `application/x-bzip` is not called `appliation/bzip`.

This PR makes the settings suggested by hm to be fully compatible with Yazi 0.4.

### Checklist

  - It is not, but it is also only a change in the example for an option value. And users get a nice warning when the option "modified" is used.

- [x]  Commit messages are formatted like

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@XYenon @eljamm

### DIFF
--- a/modules/programs/yazi.nix
+++ b/modules/programs/yazi.nix
@@ -109,7 +109,7 @@ in {
           };
           manager = {
             show_hidden = false;
-            sort_by = "modified";
+            sort_by = "mtime";
             sort_dir_first = true;
             sort_reverse = true;
           };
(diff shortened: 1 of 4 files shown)


eljamm (commented):
Thanks @Adda0! Could you also remove `x-` from the mime types as well? (`application/x-bzip` -> `application/bzip`). Occurences:
- **modules/programs/yazi.nix**: L137
- **tests/modules/programs/yazi/settings.nix**: L67
- **tests/modules/programs/yazi/theme-expected.toml**: L15

With this the config should be compatible with yazi 0.4

Adda0:
Sure. I did not know about it, but that can be resolved as well while we are at it :+1:

eljamm (approved):
Tests are passing locally. LGTM.

XYenon (approved):

rycee:

PR #6226, opened 2024-12-23
swayr: avoid IFD

mightyiam:
### Description

Avoids IFD in the swayr module.

Confirm IFD on `master` with

```nix
nix build --no-allow-import-from-derivation .\#test-swayr-basic-configuration
```

It fails due to IFD not being allowed.

On this branch the same command succeeds because IFD is not used. 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@9p4


rycee:

PR #6229, opened 2024-12-23
librespot: init module

9ary:
### Description

Librespot is a Spotify client. While there is already a module for spotifyd, which uses Librespot as a library, this adds one for the upstream frontend.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -327,6 +327,7 @@ let
     ./services/kdeconnect.nix
     ./services/keybase.nix
     ./services/keynav.nix
+    ./services/librespot.nix
     ./services/lieer.nix
     ./services/listenbrainz-mpd.nix
     ./services/lorri.nix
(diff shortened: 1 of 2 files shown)


9ary:
This is the config I'm running:
```nix
services.librespot = {
  enable = true;
  settings = {
    disable-audio-cache = true;
    disable-discovery = true;
    enable-oauth = true;
    name = "${nixosConfig.networking.hostName}";
    backend = "pulseaudio";

    # Hands off
    initial-volume = 100;
    volume-ctrl = "fixed";

    bitrate = 320;
    enable-volume-normalisation = true;
    normalisation-method = "basic";
  };
};
```

As far as integration goes, I only bothered to set the default storage paths. Not sure what else there is to do really. Discovery requires firewall configuration at the NixOS level, we can't touch that. Setting the default backend to pulseaudio is a possibility but none of the other similar modules do this.

Issue #6212, opened 2024-12-17
bug: programs.pqiv only writes config file if both 'settings' and 'extraConfig' is defined


EmmChriss:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Configuration only gets written if both 'settings' and 'extraConfig' are defined. Desired behaviour is that config gets written when at least one of the two is defined.

Condition at [programs.pqiv:68](https://github.com/nix-community/home-manager/blob/80b0fdf483c5d1cb75aaad909bd390d48673857f/modules/programs/pqiv.nix#L68) should be `||` instead of '&&'.

### Maintainer CC

@donovanglover @iynaix

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.6.64, NixOS, 24.11 (Vicuna), 24.11.20241213.314e12b`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.24.10`
 - channels(root): `"nixos-24.11, nixos-hardware"`
 - nixpkgs: `/nix/store/0xbni69flk8380w0apw4h640n37wn1i9-source`
```

donovanglover:
Nice catch! Are you interested in making a PR to fix this?

EmmChriss:
Sure, I'll take a look

donovanglover:
Issue can be closed now since the PR was merged

PR #6232, opened 2024-12-25
programs.pqiv.nix: write configuration if either settings or extraConfig are set

EmmChriss:
### Description

Issue: #6212

Due to flawed logic, configuration file for `programs.pqiv` was only written if both `settings` and `extraConfig` were set.

Desired behaviour is that either form of supplied configuration should trigger the creation of the config file. 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@donovanglover @iynaix

### DIFF
--- a/modules/programs/pqiv.nix
+++ b/modules/programs/pqiv.nix
@@ -65,7 +65,7 @@ in {
     home.packages = [ cfg.package ];
 
     xdg.configFile."pqivrc" =
-      mkIf (cfg.settings != { } && cfg.extraConfig != "") {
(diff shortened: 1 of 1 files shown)


iynaix (approved):
LGTM, not sure why CI is failing...

EmmChriss:
Not sure, complains about `euxane` not being defined (as a maintainer), but on nixpkgs-unstable (which ci looks to be using), he is defined for sure, just checked. So..

rycee:

PR #6233, opened 2024-12-25
swaync: Add onChange

LemmusLemmus:
Calls swanyc-client with the argument reload-css/reload-config when config/CSS is changed.

### Description
I have simply added calls to `swaync --reload-css` and `swaync --reload-config` for when the config or CSS are changed, so that changes take effect immediately.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@abayomi185

### DIFF
--- a/modules/services/swaync.nix
+++ b/modules/services/swaync.nix
@@ -83,13 +83,16 @@ in {
       lib.mkIf (cfg.package != null) [ cfg.package pkgs.at-spi2-core ];
 
     xdg.configFile = {
-      "swaync/config.json".source =
-        jsonFormat.generate "config.json" cfg.settings;
+      "swaync/config.json" = {
+        source = jsonFormat.generate "config.json" cfg.settings;
+        onChange = "${cfg.package}/bin/swaync-client --reload-config";
+      };
       "swaync/style.css" = lib.mkIf (cfg.style != null) {
         source = if builtins.isPath cfg.style || lib.isStorePath cfg.style then
           cfg.style
         else
           pkgs.writeText "swaync/style.css" cfg.style;
+        onChange = "${cfg.package}/bin/swaync-client --reload-css";
       };
     };
 


LemmusLemmus:
Hmm, my implementation is probably not the correct way to do this. I guess one should possible use
1. `cfg.package` instead of `pkgs.swaynotificationcenter`, and then also maybe use `lib.mkIf (cfg.package != null)`.
2.  Maybe check if swaync is enabled to begin with.
3. Judging by other code in this repo, maybe one should append ` || true` at the end in case the command fails?

I am not sure what the preferred method is though.

rycee:
Thanks for the contribution, you could give
``` patch
diff --git a/modules/services/swaync.nix b/modules/services/swaync.nix
index 5f813af3..8512d794 100644
--- a/modules/services/swaync.nix
+++ b/modules/services/swaync.nix
@@ -98,6 +98,11 @@ in {
         ConditionEnvironment = "WAYLAND_DISPLAY";
+        X-Restart-Triggers = lib.mkMerge [
+          [ config.xdg.configFile."swaync/config.json".source ]
+          (lib.mkIf (cfg.style != null)
+            [ config.xdg.configFile."swaync/style.css".source ])
+        ];
       };
 
       Service = {
diff --git a/tests/modules/services/swaync/swaync.nix b/tests/modules/services/swaync/swaync.nix
index ceb3bf06..777e0d8c 100644
--- a/tests/modules/services/swaync/swaync.nix
+++ b/tests/modules/services/swaync/swaync.nix
@@ -10,8 +10,11 @@
   };
 
   nmt.script = ''
+    serviceFile=home-files/.config/systemd/user/swaync.service
+    serviceFile=$(normalizeStorePaths $serviceFile)
+
     assertFileContent \
-      home-files/.config/systemd/user/swaync.service \
+      $serviceFile \
       ${
         builtins.toFile "swaync.service" ''
           [Install]
@@ -29,6 +32,7 @@
           Description=Swaync notification daemon
           Documentation=https://github.com/ErikReider/SwayNotificationCenter
           PartOf=graphical-session.target
+          X-Restart-Triggers=/nix/store/00000000000000000000000000000000-config.json
         ''
       }
   '';
```
a try for a more generic solution.

LemmusLemmus:
Sorry for the delay, I am not dismissing the suggestion :)

I might mention that I am using home-manager as a NixOS module (through flakes).

After removing my changes and copying the changes that rycee suggested, nothing seems to happen when I rebuild with a different style?  I am probably doing something wrong though.

I assume that `X-Restart-Triggers` should restart the service. Would reloading it not be better? Although the swaync service seems to be [implemented](https://github.com/ErikReider/SwayNotificationCenter/blob/main/services/systemd/swaync.service.in) such that it simply calls `--reload-config` and `--reload-css`:
```
ExecReload=@bindir@/swaync-client --reload-config ; @bindir@/swaync-client --reload-css
```
This `ExecReload` part is not present in the service definition in home-manager `swaync.nix`.

--------

By the way, the home manager manual seems to suggest that it should be enough to run
`home-manager --override-input home-manager ~/devel/home-manager` to start contributing, however I do not have  the command `home-manager`, and instead I have run `nix flake lock --override-input home-manager ~/devel/home-manager`, as well as changed the `home-manager.url` in my flake.nix to point at the home-manager directory. This seems to work, but is this the correct way to do it? (If so, maybe it would be worth mentioning in the manual).

Lillecarl:
Regarding home-manager overrides to start contributing. There's a thousand ways to do it.

The easiest way when developing a single module might be setting
```nix
disabledModules = [ "path/in/hm/to/module.nix" ];
```
Then copying the entire thing into your own config folder, start hacking away and copy it into HM when it's ready for some kind of review.

I'm LilleCarl at Matrix.org, happy to help anyone who's got the energy to contribute! 😄

khaneliman:
> This might can be better.

Judging by the discussion, I assumed this was tried and didn't work. I can try it myself, later.

YaroKasear:
This commit causes problems with impermanence setups. 

It causes home manager activation to stall, delay boot up to five minutes, then time out, and cause the bind-mounts to stop working.

LemmusLemmus:
A potential reason for it stalling could be if `swaync` is not running, in which case `swaync-client` (regardless of arguments) will apparently wait for a connection until `swaync` starts running, if it ever does.

A possible solution could be to check if any `swaync-client` is running, and only in that case run `--reload-css/config`...

PR #6235, opened 2024-12-26
ghostty: add module

### LINKED ISSUE #6236: Program request: Ghostty
### Description
https://ghostty.org/docs/about
Ghosty is the hot new terminal on the block and should have a hm module. It's config is all `key = value` so I assume it would not be to hard to add.

Example config:
`
background-opacity = 0.8
gtk-titlebar = false
`

HeitorAugustoLN:
### Description

Adds a module for ghostty terminal emulator, depends on NixOS/nixpkgs#368404
Closes #6236

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


anund:
You could also look at module options adding the vim/bat plugins that ship with ghostty. In `$out/share` there are several external application integrations, I think only vim and bat require user intervention to configure.

anund:
```nix
  programs.bat = {
    enable = true;
    syntaxes = {
      ghostty = {
        src = ghostty;
        file = "share/bat/syntaxes/ghostty.sublime-syntax";
      };
    };
    config = {
      map-syntax = [ "*/ghostty/config:Ghostty Config" ];
      # or more explicit
      # map-syntax = [ "${config.xdg.configHome}/ghostty/config:Ghostty Config" ];
    };
  };

  programs.vim.plugins = [
    ghostty.vim
  ]; 
```

HeitorAugustoLN:
> You could also look at module options adding the vim/bat plugins that ship with ghostty. In $out/share there are several 
> external application integrations, I think only vim and bat require user intervention to configure.
> 
> https://github.com/clo4/ghostty-hm-module @clo4 might have additional insight to share

Didn't knew this was a thing, I will look into it.

arunoruto:
Maybe I am missing something, but I am not sure how to add keybindings using the config...
I thought it should be set as an array, but I get an error that it should be of type `atom (null, bool, int, float or string)`. Maybe an additional setting for keybindings could be implemented? The format could be a set with `"shortcut" = "action"` and it get squished down into a string `keybind = shortcut=action`.

HeitorAugustoLN:
> Maybe I am missing something, but I am not sure how to add keybindings using the config... I thought it should be set as an array, but I get an error that it should be of type `atom (null, bool, int, float or string)`. Maybe an additional setting for keybindings could be implemented? The format could be a set with `"shortcut" = "action"` and it get squished down into a string `keybind = shortcut=action`.

```nix
{
  programs.ghostty.settings = {
    keybind = [
      "clear"
      "ctrl+h=goto_split:left"
      "ctrl+l=goto_split:right"
      # etc.
    ];
  };
}
```

###### If this doesn't work for you right now, maybe you should update the input.

I wouldn't really like to create an option solely for keybindings, because I don't think it is really needed.

anund:
Ghostty config format is <key><optional spaces>=<optional spaces><value>. The value must be a ~complete string as if you past it via the command line. (`ghostty --key=value`). https://github.com/ghostty-org/ghostty/issues/3203 has a decent summary explanation.

anund:
https://github.com/ghostty-org/ghostty/issues/3203 has a good summary explanation for the config format.

arunoruto:
> > Maybe I am missing something, but I am not sure how to add keybindings using the config... I thought it should be set as an array, but I get an error that it should be of type `atom (null, bool, int, float or string)`. Maybe an additional setting for keybindings could be implemented? The format could be a set with `"shortcut" = "action"` and it get squished down into a string `keybind = shortcut=action`.
> 
> ```nix
> {
>   programs.ghostty.settings = {
>     keybind = [
>       "clear"
>       "ctrl+h=goto_split:left"
>       "ctrl+l=goto_split:right"
>       # etc.
>     ];
>   };
> }
> ```
> 
> ###### If this doesn't work for you right now, maybe you should update the input.
> 
> I wouldn't really like to create an option solely for keybindings, because I don't think it is really needed. My only concern is the `keybind = clear`, since I don't know if it's position matters on the configuration file or not. But if it does maybe I will need to create a separate option for that.

Fair enough, I just came up at the same solution! Thanks tho!

HeitorAugustoLN:
Added a option for prepending `keybind = clear` before settings so it doesn't reset some custom keybinds

modules/programs/ghostty.nix
```diff
@@ -0,0 +1,71 @@
...
+      inherit (keyValue) type;
+      default = { };
+      example = lib.literalExpression ''
+        {
+          theme = "catppuccin-mocha";
+          font-size = 10;
+        }
+      '';
+      description = ''
+        Configuration written to $XDG_CONFIG_HOME/ghostty/config.
```
rhodes-b, line 25: the default is ~/.config/ghostty/config if `$XDG_CONFIG_HOME` is not set

HeitorAugustoLN: I don't think it is really needed, $XDG_CONFIG_HOME in that case would just be `xdg.configDir` in home-manager. It is common knowledge that `$HOME/.config` is the default `$XDG_CONFIG_HOME`

zoedsoupe:
wow! i was starting to implement this module! awesome!

now question i have is: I don't know how home-manager aims to tackle this or if there's a standard, but some module have extensive nix options representing actual settings items and others have a generic 'settings' option that receive a raw string while others have a mix of 'native' options in addition to the extraConfig option.

based in the standard (or not), does it make sense to try to implement native options in nix for the Ghostty settings? I could tackle this in another pr if this is the case. wdyt?

HeitorAugustoLN:
> now question i have is: I don't know how home-manager aims to tackle this or if there's a standard, but some module have extensive nix options representing actual settings items and others have a generic 'settings' option that receive a raw string while others have a mix of 'native' options in addition to the extraConfig option.

There *is* a standard: [RFC 0042](https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md), which is what I’m following for this PR. But yeah, not all modules implement it.

> Based on the standard (or not), does it make sense to try and implement native options in Nix for Ghostty’s settings? If it does, I could take this on in another PR. What do you think?

Technically, I *could* create a submodule to map out all the Ghostty options and then add a `freeformType` for anything not covered. But honestly, that feels super counterintuitive. It would take a ton of time and still end up with the same result. Plus, over time, as Ghostty adds more options (and it already has a ton), it’ll just get harder to maintain. Check out the [motivation section](https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md#motivation) in the RFC.

That said, there *are* cases where mapping options is worth it, like if the upstream project doesn’t have proper documentation for their settings. But that’s not the case with Ghostty, they’ve got a [solid configuration reference](https://ghostty.org/docs/config/reference)

modules/programs/ghostty.nix
```diff
@@ -0,0 +1,71 @@
...
+      };
+    installVimSyntax = lib.mkEnableOption
+      "installation of ghostty configuration syntax for vim/neovim";
+  };
+
+  config = lib.mkIf cfg.enable (lib.mkMerge [
+    {
+      home.packages = [ cfg.package ];
+
+      xdg.configFile."ghostty/config" = lib.mkIf (cfg.settings != { }) {
```
pope, line 46: This is where Linux users would place their config, but for macos users, that location is `home.file."Library/Application Support/com.mitchellh.ghostty/config"`. 

I know Ghostty for darwin-aarch64 isn't part of nixpkgs, but it would still be very convenient for darwin users to be able to configure things though home. WDYT?

HeitorAugustoLN: Thought it were the same, thanks for the information

zoedsoupe: as far as i know macos also support xdg config, right? ghostty docs says the same and even look on both options

jamesog: Mac will use either, and will check XDG_CONFIG_HOME first: https://ghostty.org/docs/config#file-location

arunoruto:
I got feedback on my PR for [adding ghostty to stylix](https://github.com/danth/stylix/pull/703#pullrequestreview-2524379748). One question was if there was a way to configure custom themes using the HM module. AFAIK the theme can only be set, but not provided. Are there any plans on adding this feature in this PR, or should I open up a new one when this one gets merged?

HeitorAugustoLN:
> I got feedback on my PR for [adding ghostty to stylix](https://github.com/danth/stylix/pull/703#pullrequestreview-2524379748). One question was if there was a way to configure custom themes using the HM module. AFAIK the theme can only be set, but not provided. Are there any plans on adding this feature in this PR, or should I open up a new one when this one gets merged?

I will add an option to add themes

modules/programs/ghostty.nix
pope: Thanks folks. I missed that part where XDG_CONFIG_HOME was supported. My bad and sorry for the noise.

HeitorAugustoLN:
@arunoruto added themes option

modules/programs/ghostty.nix
```diff
@@ -0,0 +1,121 @@
...
+        };
+      };
+      description = ''
+        Custom themes written to $XDG_CONFIG_HOME/ghostty/themes.
+
+        See https://ghostty.org/docs/features/theme#authoring-a-custom-theme for more information.
+      '';
+    };
+
+    clearDefaultKeybinds = lib.mkEnableOption "clear default keybinds" // {
```
rycee, line 68: Since the description is overridden, we can pass an empty string.
```suggestion
    clearDefaultKeybinds = lib.mkEnableOption "" // {
```

tests/modules/programs/ghostty/empty-settings.nix
```diff
@@ -0,0 +1,9 @@
+{
+  config = {
```
rycee, line 2: Can remove the `config` attribute:
```suggestion
```

Similar holds for the other tests.

modules/programs/ghostty.nix
```diff
@@ -0,0 +1,121 @@
...
+              settings =
+                lib.generators.toKeyValue keyValueSettings cfg.settings;
+            in pkgs.writeText "ghostty-config" "${
+              lib.optionalString cfg.clearDefaultKeybinds ''
+                keybind = clear
+
+              ''
+            }${settings}";
+          };
+        }
```
rycee, line 99: Perhaps could do something like

```suggestion
      programs.ghostty.settings = lib.mkIf cfg.clearDefaultKeybinds {
        keybind = lib.mkBefore [ "clear" ];
      };

      xdg.configFile = lib.mkMerge [
        {
          "ghostty/config" = lib.mkIf (cfg.settings != { }) {
            source = keyValue.generate "ghostty.conf" cfg.settings;
          };
        }
```

modules/programs/ghostty.nix
```diff
@@ -0,0 +1,121 @@
...
+
+        (lib.mkIf (cfg.themes != { }) (lib.mapAttrs' (name: value:
+          lib.nameValuePair "ghostty/themes/${name}" {
+            source = keyValue.generate "ghostty-${name}-theme" value;
+          }) cfg.themes))
```
rycee, line 104: Maybe mildly neater with

```suggestion
        (lib.mkIf (cfg.themes != { }) (lib.mapAttrs' (name: value: {
          name = "ghostty/themes/${name}";
          value.source = keyValue.generate "ghostty-${name}-theme" value;
        }) cfg.themes))
```

modules/programs/ghostty.nix
```diff
@@ -0,0 +1,121 @@
...
+            "12=#89b4fa"
+            "13=#f5c2e7"
+            "14=#94e2d5"
+            "15=#a6adc8"
+          ];
+          background = "1e1e2e";
+          foreground = "cdd6f4";
+          cursor-color = "f5e0dc";
+          selection-background = "353749";
+          selection-foreground = "cdd6f4";
```
rycee, line 58: The palette color values all start with `#`, perhaps the same is necessary here? Or vice versa?

modules/programs/ghostty.nix
```diff
@@ -0,0 +1,121 @@
...
+
+        See https://ghostty.org/docs/config/reference for more information.
+      '';
+    };
+
+    themes = lib.mkOption {
+      type = lib.types.attrsOf keyValue.type;
+      default = { };
+      example = {
+        "catppuccin-mocha" = {
```
rycee, line 35:
```suggestion
        catppuccin-mocha = {
```

rycee (commented):
Thanks for the contribution! Looks good in general, I've just added a few minor comments.

modules/programs/ghostty.nix
arunoruto: I am not sure why, but only the palette colors need to be prefixed with `#`, the rest is just a hex code without the shebang.

HeitorAugustoLN: https://ghostty.org/docs/features/theme#example looks like it is not needed.

modules/programs/ghostty.nix
HeitorAugustoLN: Forgot about that one. Thanks!

modules/programs/ghostty.nix
rycee: Good, then no worries :-)

anund:
May want to add an option for an ssh terminfo work around. See https://ghostty.org/docs/help/terminfo#ssh
```nix
  programs.ssh = {
    matchBlocks = {
      "*" = {
        extraOptions = {
          SetEnv = "TERM=xterm-256color";
        };
      };
    };
  };    
```

anund:
Could also include shell-integration flags following the manual integration steps here: https://ghostty.org/docs/help/terminfo#ssh. Similar to kitty, though I don't think you need the extra mangling they to to set `no-rc`.

anund:
Roughly
 
```nix
  programs.zsh.initExtra = ''
    if [[ -n $GHOSTTY_RESOURCES_DIR ]]; then
      source $GHOSTTY_RESOURCES_DIR/shell-integration/zsh/ghostty-integration
    fi
  '';

  programs.bash.bashrcExtra = lib.mkBefore ''
    if [[ -n $GHOSTTY_RESOURCES_DIR ]]; then
      source $GHOSTTY_RESOURCES_DIR/shell-integration/zsh/ghostty-integration
    fi
  '';

  programs.fish.shellInit = ''
    if [[ -n $GHOSTTY_RESOURCES_DIR ]]; then
      source #GHOSTTY_RESOURCES_DIR/shell-integration/fish/vendor_conf.d/ghostty-shell-integration.fish
    fi                                           
  '';    
```

HeitorAugustoLN:
> Roughly
> 
> ```nix
>   programs.zsh.initExtra = ''
>     if [[ -n $GHOSTTY_RESOURCES_DIR ]]; then
>       source $GHOSTTY_RESOURCES_DIR/shell-integration/zsh/ghostty-integration
>     fi
>   '';
> 
>   programs.bash.bashrcExtra = lib.mkBefore ''
>     if [[ -n $GHOSTTY_RESOURCES_DIR ]]; then
>       builtin source "${GHOSTTY_RESOURCES_DIR}/shell-integration/bash/ghostty.bash"
>     fi
>   '';
> 
>   programs.fish.shellInit = ''
>     if [[ -n $GHOSTTY_RESOURCES_DIR ]]; then
>       source #GHOSTTY_RESOURCES_DIR/shell-integration/fish/vendor_conf.d/ghostty-shell-integration.fish
>     fi                                           
>   '';    
> ```

This looks like is already done by the package itself.

anund:
You're correct for most interactions where you launch the shell with ghostty. The above helps if you do launch another shell from the first one ghostty opens. (ie launching one shell and manually starting another)

anund:
Number of bad copy/pastes in the above shell examples. Fixed.

anund:
Valid snippets: (sorry)

```nix
  programs.zsh.initExtra = ''
    if [[ -n $GHOSTTY_RESOURCES_DIR ]]; then
      source $GHOSTTY_RESOURCES_DIR/shell-integration/zsh/ghostty-integration
    fi
  '';

  programs.bash.bashrcExtra = lib.mkBefore ''
    if [[ -n $GHOSTTY_RESOURCES_DIR ]]; then
      builtin source "''${GHOSTTY_RESOURCES_DIR}/shell-integration/bash/ghostty.bash"
    fi
  '';

  programs.fish.shellInit = ''
    if set -q GHOSTTY_RESOURCES_DIR
      source "$GHOSTTY_RESOURCES_DIR/shell-integration/fish/vendor_conf.d/ghostty-shell-integration.fish"
    end
  '';

```

rhodes-b:
> > Roughly
> > ```nix
> >   programs.zsh.initExtra = ''
> >     if [[ -n $GHOSTTY_RESOURCES_DIR ]]; then
> >       source $GHOSTTY_RESOURCES_DIR/shell-integration/zsh/ghostty-integration
> >     fi
> >   '';
> > 
> >   programs.bash.bashrcExtra = lib.mkBefore ''
> >     if [[ -n $GHOSTTY_RESOURCES_DIR ]]; then
> >       builtin source "${GHOSTTY_RESOURCES_DIR}/shell-integration/bash/ghostty.bash"
> >     fi
> >   '';
> > 
> >   programs.fish.shellInit = ''
> >     if [[ -n $GHOSTTY_RESOURCES_DIR ]]; then
> >       source #GHOSTTY_RESOURCES_DIR/shell-integration/fish/vendor_conf.d/ghostty-shell-integration.fish
> >     fi                                           
> >   '';    
> > ```
> 
> This looks like is already done by the package itself. But I am not sure.

to follow up on this ghostty shell integration can break if its not at the very top of the respective file. Also by default ghostty will attempt to do shell integration automatically making this not required to add in your config
https://ghostty.org/docs/features/shell-integration

> Some Ghostty features require integrating with your shell. Ghostty can automatically inject shell integration for bash, zsh, fish, and elvish.

> Technically, the bash shell integration script could be sourced from anywhere, but some bash configurations may interfere with the shell integration. To avoid this, we recommend sourcing the script as early as possible in your shell configuration.

anund:
Believe the top of file warning is specific to bash. (see the `lib.mkBefore` in the snippet)

arunoruto:
Ghostty just got merged into `nixos-unstable`. I wanted to ask what needs to be done, to change the draft status and marge it :)

HeitorAugustoLN:
Some updates:

- I decided to drop the `installVimSyntax` option since the package already takes care of it.]
- Added optional shell integration options

HeitorAugustoLN:
> I wanted to ask what needs to be done, to change the draft status and marge it :)

#6240 needs to get merged, so nixpkgs is updated

surfaceflinger:
I think rycee or someone could manually trigger the flake update workflow, otherwise you'll have to wait until Wednesday and then for a merge

anund:
For the vim syntax: adding `programs.vim.plugins = [ ghostty.vim ];` is required by users. nixpkgs/home-manager vim modules will not automatically add the plugin. Just to confuse things: Neovim will.

HeitorAugustoLN:
> For the vim syntax: adding `programs.vim.plugins = [ ghostty.vim ];` is required by users. nixpkgs/home-manager vim modules will not automatically add the plugin. Just to confuse things: Neovim will.

I completely forgot that vim handles things differently, re-added the option exclusively for vim

modules/programs/ghostty.nix
```diff
@@ -0,0 +1,171 @@
...
+    {
+      home.packages = [ cfg.package ];
+
+      programs.ghostty.settings = lib.mkIf cfg.clearDefaultKeybinds {
+        keybind = lib.mkBefore [ "clear" ];
+      };
+
+      xdg.configFile = lib.mkMerge [
+        {
+          "ghostty/config" = lib.mkIf (cfg.settings != { }) {
```
YorikSar, line 119: On macOS, this config file is in `~/Library/Application Support/com.mitchellh.ghostty/config`. Please add a branch for that.

YorikSar (changes requested):

modules/programs/ghostty.nix
HeitorAugustoLN: This was already discussed in: https://github.com/nix-community/home-manager/pull/6235#discussion_r1898657805. Ghostty will look for both

YorikSar: Ah, sorry. GitHub UI doesn't let you search through comments easily (or at all). Maybe add a comment in code to reference that to avoid any confusion in the future?

HeitorAugustoLN: > Ah, sorry. GitHub UI doesn't let you search through comments easily (or at all). Maybe add a comment in code to reference that to avoid any confusion in the future?

No problem. I don't really think a code comment is needed. But I might add one if I modify the module again.

modules/programs/ghostty.nix
```diff
@@ -0,0 +1,171 @@
...
+  keyValueSettings = {
+    listsAsDuplicateKeys = true;
+    mkKeyValue = lib.generators.mkKeyValueDefault { } " = ";
+  };
+  keyValue = pkgs.formats.keyValue keyValueSettings;
+in {
+  meta.maintainers = [ lib.maintainers.HeitorAugustoLN ];
+
+  options.programs.ghostty = {
+    enable = lib.mkEnableOption "Ghostty";
```
dezren39, line 14: why was this capitalized? it wasn't in a previous commit. i'm not sure how it would affect things, if at all.

HeitorAugustoLN: It doesn't affect anything, it just capitalizes the Ghostty in the description of the option.

modules/programs/ghostty.nix
rhodes-b: https://ghostty.org/docs/config#file-location

its also very well documented on the website

modules/programs/ghostty.nix
```diff
@@ -0,0 +1,171 @@
...
+        if set -q GHOSTTY_RESOURCES_DIR
+          source "$GHOSTTY_RESOURCES_DIR/shell-integration/fish/vendor_conf.d/ghostty-shell-integration.fish"
+        end
+      '';
+    })
+
+    (lib.mkIf cfg.enableZshIntegration {
+      programs.zsh.initExtra = ''
+        if [[ -n $GHOSTTY_RESOURCES_DIR ]]; then
+          source $GHOSTTY_RESOURCES_DIR/shell-integration/zsh/ghostty-integration
```
anund, line 166:
```suggestion
          source "$GHOSTTY_RESOURCES_DIR/shell-integration/zsh/ghostty-integration"
```
Turns out zsh can have issues with the unquoted value in certain cases.

HeitorAugustoLN: Updated.

rycee:

nyabinary:
Does this have nushell integration?

arunoruto:
> Does this have nushell integration?

Ghostty does not provide a nushell integration according to their [docs](https://ghostty.org/docs/features/shell-integration#shell-integration-injection)

anund:
Not currently no. See https://github.com/ghostty-org/ghostty/tree/main/src/shell-integration and perhaps https://github.com/ghostty-org/ghostty/pull/1774/files as a sketch pr for adding it

HeitorAugustoLN:
> Does this have nushell integration?

Some shells such as [Nushell](https://www.nushell.sh/) have built-in support for some of the features that Ghostty provides, so shell integration is not necessary.

This is what they have in their documentation

NeQuissimus:
This module does break when `bat` is enabled. It looks like the syntax files are not as expected by that module.

HeitorAugustoLN:
> This module does break when `bat` is enabled. It looks like the syntax files are not as expected by that module.

Could you provide the error for me? I have bat enabled in my configuration, and it didn't fail to build, last time I updated my configuration

NeQuissimus:
```
Activating checkFilesChanged
Activating checkLaunchAgents
Activating checkLinkTargets
Activating writeBoundary
Activating linkGeneration
Cleaning up orphan links from /Users/nequi
Creating profile generation 85
Creating home file links in /Users/nequi
Activating batCache
No themes were found in '/Users/nequi/.config/bat/themes', using the default set
[bat error]: error finding all the files in a directory: IO error for operation on /Users/nequi/.config/ba
t/syntaxes/ghostty.sublime-syntax: No such file or directory (os error 2)
```

Relevant configs:

```
  programs = {
    bat.enable = true;

    ghostty = {
      enable = true;
      enableZshIntegration = true;

      settings = {
        auto-update = "off";
        background-opacity = 0.8;
        confirm-close-surface = false;
        font-family = "FiraCode Nerd Font Mono";
        font-size = 12;
        theme = "Teerb";
      };
    };
};
```

HeitorAugustoLN:
I see you are on macOS, so my guess is that the bat syntax file is not available in the darwin package for some reason, and looks like the package got marked again as broken on darwin

PR #6238, opened 2024-12-27
copyq: add option to disable XWayland

marcin-serwin:
### Description

The current CopyQ service always forces the use of XWayland, this PR adds option to disable that.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@DamienCassou

### DIFF
--- a/tests/modules/services/copyq/basic-configuration.nix
+++ b/tests/modules/services/copyq/basic-configuration.nix
@@ -1,0 +1,17 @@
+{ ... }:
+
+{
+  config = {
+    services.copyq = {
+      enable = true;
+      systemdTarget = "sway-session.target";
+    };
+
+    test.stubs.copyq = { };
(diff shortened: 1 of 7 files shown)


tests/modules/services/copyq/basic-configuration.nix
```diff
@@ -0,0 +1,15 @@
+{ ... }:
+
+{
+  config = {
+    services.copyq = {
+      enable = true;
+      systemdTarget = "sway-session.target";
+    };
```
rycee, line 8: It's best to avoid pulling in the actual copyq package as a dependency since it'll cause unnecessary downloads. Instead can try stubbing the package:

```suggestion
    };
    
    test.stubs.copyq = { };
```

rycee:

PR #6239, opened 2024-12-28
treewide: standardize Wayland graphical services

thiagokokada:
### Description

Graphical services that are meant for Wayland environment are all over the place:
- Some set `After = [ "graphical-session-pre.target" ]`, that is almost always a mistake ([see](https://github.com/Vladimir-csp/uwsm/issues/40)), others set `After = [ "graphical-session.target" ]`
- Some have `systemd.target` (or `systemdTarget`) to allow changing the default target, some does not
- Some set `ConditionEnvironment = "WAYLAND_DISPLAY"`, some does not

This PR tries to standardize this by:
- Adding `systemd.target` to services missing it, and defaulting it to `graphical-session.target`
- Set `After`, `PartOf` and `Wanted` to `cfg.systemd.target` 
- Adding `ConditionEnvironment = "WAYLAND_DISPLAY"` for services missing it

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/wob.nix
+++ b/modules/services/wob.nix
@@ -37,6 +37,19 @@ in {
 
     systemd = mkEnableOption "systemd service and socket for wob"
       // mkOption { default = true; };
+
+    systemdTarget = mkOption {
+      type = lib.types.str;
+      default = "graphical-session.target";
+      example = "sway-session.target";
+      description = ''
+        The systemd target that will automatically start the Waybar service.
+
+        When setting this value to `"sway-session.target"`,
+        make sure to also enable {option}`wayland.windowManager.sway.systemd.enable`,
+        otherwise the service may never be started.
+      '';
+    };
   };
 
   config = mkIf cfg.enable {
@@ -50,8 +63,8 @@ in {
           Description =
             "A lightweight overlay volume/backlight/progress/anything bar for Wayland";
           Documentation = "man:wob(1)";
-          PartOf = "graphical-session.target";
-          After = "graphical-session.target";
+          PartOf = [ cfg.systemdTarget ];
+          After = [ cfg.systemdTarget ];
           ConditionEnvironment = "WAYLAND_DISPLAY";
         };
         Service = {
@@ -59,7 +72,7 @@ in {
           ExecStart = builtins.concatStringsSep " " ([ (getExe cfg.package) ]
             ++ optional (cfg.settings != { }) "--config ${configFile}");
         };
-        Install.WantedBy = [ "graphical-session.target" ];
+        Install.WantedBy = [ cfg.systemdTarget ];
       };
 
       sockets.wob = {
--- a/modules/programs/waybar.nix
+++ b/modules/programs/waybar.nix
@@ -309,8 +309,9 @@ in {
           Description =
             "Highly customizable Wayland bar for Sway and Wlroots based compositors.";
           Documentation = "https://github.com/Alexays/Waybar/wiki";
-          PartOf = [ "graphical-session.target" ];
-          After = [ "graphical-session-pre.target" ];
+          PartOf = [ cfg.systemd.target ];
+          After = [ cfg.systemd.target ];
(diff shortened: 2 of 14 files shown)


modules/services/wob.nix
```diff
@@ -37,6 +37,19 @@ in {
 
     systemd = mkEnableOption "systemd service and socket for wob"
       // mkOption { default = true; };
+
+    systemdTarget = mkOption {
```
thiagokokada, line 41: `systemd` is sadly already used to enable the service above, so using `systemdTarget` instead to avoid breaking changes.

thiagokokada:
CC @rycee.

Lillecarl:
This is good. If I had a say in it I'd create an option like "services.wayland.systemdTarget" and make all the options default to this one. What do you think? Then we end up with the hardest problem in programming... Naming things.

rycee:
Thanks @thiagokokada for taking on this. It has indeed become quite the mess. I'll have a deeper look at this PR as soon as possible.

I think @Lillecarl's idea to have a common `wayland.systemdTarget` seems quite good. I suspect it would be very rare that anybody would like different targets.

That said, there is something fundamentally wrong somewhere if we cannot use `graphical-session.target` and `graphical-session-pre.target` as-is. The whole point of these targets is to handle the graphical session services! I haven't had time and motivation to look into it since `graphical-session.target` works fine for me even on Wayland (in my case Hyprland).

thiagokokada:
> I think @Lillecarl's idea to have a common `wayland.systemdTarget` seems quite good. I suspect it would be very rare that anybody would like different targets.

I initially had the same idea, but was against it because it would be more work (e.g.: deprecating old options). But if this is the way we want to do things I can redo this work this way instead.

> That said, there is something fundamentally wrong somewhere if we cannot use `graphical-session.target` and `graphical-session-pre.target` as-is. The whole point of these targets is to handle the graphical session services! I haven't had time and motivation to look into it since `graphical-session.target` works fine for me even on Wayland (in my case Hyprland).

To be clear, `graphical-session.target` works fine to me, but `graphical-sesion-pre.target` doesn't since in most cases things that services depends are not ready yet (e.g.: `WAYLAND_DISPLAY`). The only reason I added `systemd.target` for other services is for consistency, but my main reason for this PR is replacing `After = [ "graphical-session-pre.target" ];` with `After = [ "graphical-session.target" ];`.

thiagokokada:
BTW, `graphical-session-pre.target` only makes sense for the compositor (e.g.: Hyprland, Sway, etc.), everything else should use `graphical-session.target`: https://github.com/Vladimir-csp/uwsm/issues/40#issuecomment-2345420159.

Lillecarl:
> That said, there is something fundamentally wrong somewhere if we cannot use `graphical-session.target` and `graphical-session-pre.target` as-is. The whole point of these targets is to handle the graphical session services! I haven't had time and motivation to look into it since `graphical-session.target` works fine for me even on Wayland (in my case Hyprland).

Everything should work just fine on graphical-session.target, but considering the target audience (people who run "not in [ kde gnome ] I think having the option available is a good thing.

>I initially had the same idea, but was against it because it would be more work (e.g.: deprecating old options). But if this is the way we want to do things I can redo this work this way instead.

You can just set "default = config.wayland.systemdTarget" in all options that already exist, you don't need to deprecate them. Then again, they're just one mkForce away if they're deprecated and removed.

thiagokokada:
> You can just set "default = config.wayland.systemdTarget" in all options that already exist, you don't need to deprecate them. Then again, they're just one mkForce away if they're deprecated and removed.

I know, but unless we deprecate the old options the inconsistency will exist and folks will find new and creative ways to abuse the old target (like this one https://github.com/nix-community/home-manager/pull/6241, that I find it would be better done by disabling the `systemd` service instead).

thiagokokada:
Ok, I created a `wayland.systemd.target` and add a few more services. Opened a new PR: https://github.com/nix-community/home-manager/pull/6253.

PR #6242, opened 2024-12-29
home-manager: move profile management

rycee:
### Description

This commit separates profile management (setting profile and creating GC root) from file management (removing and adding managed files within the user's home directory).

This is a step towards deprecating profile management within the activation script, instead relying on the caller of the activation script managing the profile.

### Checklist

- [x] Change is backwards compatible. Almost entirely, hypothetically somebody may rely on the profile switch happening in the link files activation script block but 

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/home-manager/po/home-manager.pot
+++ b/home-manager/po/home-manager.pot
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Home Manager\n"
 "Report-Msgid-Bugs-To: https://github.com/nix-community/home-manager/issues\n"
-"POT-Creation-Date: 2024-04-17 23:19+0200\n"
+"POT-Creation-Date: 2025-01-03 09:09+0100\n"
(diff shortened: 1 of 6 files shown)


rycee:
@blaggacao Hi! Since you did the original support of "nix profile", would you mind testing this PR? It is related to some simplifications I'm working on. In this case the home-manager profile update is moved and simplified. Note in the simplificated code I always use `nix-env` for the profile management.

In my tests this has worked well even when the user uses the modern "nix profile" style profiles. Still would be good with an independent test, though 🙂

blaggacao:
Hi @rycee, I'm taking a break and keeping myself away from Nix setups these days, at least for some time.

Therefore, I hope my validation can be skipped. 

I will maintain the intention to fix anything related to this should I stumble over it: I have a starting point, now, thank you for pinging me!

rycee:
@blaggacao No worries, have a nice vacation away from Nix! 😀

PR #6243, opened 2024-12-30
Allow multiple backups files

colonelpanic8:
…he backup file path

### Description

Allow multiple backup files -- avoid failing the "switch home manager configuration".

### Checklist

- [x ] Change is backwards compatible.

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/files/check-link-targets.sh
+++ b/modules/files/check-link-targets.sh
@@ -33,12 +33,16 @@ for sourcePath in "$@" ; do
     elif [[ ! -L "$targetPath" && -n "$HOME_MANAGER_BACKUP_EXT" ]] ; then
       # Next, try to move the file to a backup location if configured and possible
       backup="$targetPath.$HOME_MANAGER_BACKUP_EXT"
-      if [[ -e "$backup" ]]; then
-        errorEcho "Existing file '$backup' would be clobbered by backing up '$targetPath'"
-        collision=1
-      else
-        warnEcho "Existing file '$targetPath' is in the way of '$sourcePath', will be moved to '$backup'"
-      fi
+
+      # Check for backup existence and iterate until finding an available filename
+      counter=1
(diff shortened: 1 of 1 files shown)


teto:
if you need multiple backup files, I am tempted to think something is wrong with setup so I would lean against the feature. Do you have a specific scenario in mind ?

colonelpanic8:
> if you need multiple backup files, I am tempted to think something is wrong with setup so I would lean against the feature. Do you have a specific scenario in mind ?

It's not that I need multiple backup files -- its just that I find it very annoying that this fails if there was a preexisting backup file.

In my case, my gtkrc files sometimes get overwritten by certain programs and my home manager activation is constantly failing because of this. I don't really care about the files that are getting overwritten and I think trying to aim for the default of almost never failing is a good idea.

teto:
you can set `force = true` to forcefully update the file (check the code to find examples), which solves your issue in a better way IMO.

modules/files/check-link-targets.sh
```diff
@@ -33,12 +33,16 @@ for sourcePath in "$@" ; do
...
-      else
-        warnEcho "Existing file '$targetPath' is in the way of '$sourcePath', will be moved to '$backup'"
-      fi
+
+      # Check for backup existence and iterate until finding an available filename
+      counter=1
+      while [[ -e "$backup" ]]; do
+        backup="$targetPath.$HOME_MANAGER_BACKUP_EXT.$counter"
+        counter=$((counter + 1))
+      done
```
thiagokokada, line 42: `stat` tends to be slow in some systems (especially in macOS), and considering that this is affecting you at every new generation it means you will eventually end up having multiple files and your switch taking longer and longer.

I concur with @teto here, it is better to use `force` in your case.

PR #6244, opened 2024-12-30
firefox: fix starting with non-default profile

spacedentist:
### Description

When home-manager manages the Firefox configuration, the `profiles.ini` file in `~/.mozilla/firefox/` is immutable. Any attempt by Firefox to update it fails accordingly. Currently, when multiple profiles are defined, it is not possible to choose a non-default one by passing a `-Profile` option to Firefox, or by using `firefox -ProfileManager`: if the user chooses a non-default profile, Firefox tries to update the `profile.ini` with this choice, which fails because of the file's immutability, and Firefox refuses to start up. The user is shown an error saying `An unexpected error has prevented your changes from being saved.`.

Currently, the setting to save the profile choice back to `profiles.ini` is hardcoded in home-manager. This Pull Request changes that, so the file is never updated with the user choice.

This fixes the cases where a user attempts to start a non-default profile.

I think this is also matches home-manager philosophy nicely: because we define what the default profile is in our home-manager config, the user should not be able to change that in the firefox gui, because then the active configuration would diverge from what's defined in home-manager. As a matter of fact, the user currently can't change the default anyway, as trying to do so results in an error, but the user should be able to choose a profile for an individual Firefox instance.

The Firefox Profile Manager GUI (that you get by starting Firefox with the `-ProfileManager` option) has a check box to decide whether to "use the selected profile without asking at startup", i.e. make it the new default. This is currently ticked (because it's hardcoded in home-manager as `StartWithLastProfile = 1`). If it stays ticked, choosing any profile other than the default results in the error. Unticking this box will also not help, because now Firefox would attempt to update `profiles.ini` with `StartWithLastProfile=0`.

Either way, when home-manager manages the Firefox config for us, we cannot have Firefox update `profiles.ini`, and therefore we cannot change the default profile through Firefox's GUI.

With this fix, choosing a different profile on the command line (`firefox -Profile foo`) or via the GUI (`firefox -ProfileManager`) now works. The user will still get an error when they tick the checkbox to "use the selected profile without asking at startup", because that results in an attempt to update `profiles.ini`. Also, it would mean that the chosen default profile would diverge from the setting in the user's home-manager config - so not allowing this seems sensible.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@brckd @rycee @offlinehacker


brckd:
Thank you @spacedentist for your contribution shining a light on this issue! I agree there should be a way to select profiles other than the default one. As you said, Home Manager tries to uphold reproducibility and thus a solution that keeps the default profile the default choice and optionally allows to temporarily select a different profile would be idea

Sadly your proposed solution didn't work with neither Firefox nor LibreWolf on my NixOS machine. I set `StartWithLastProfile = 0` and it launched the profile switcher. But attempting to start profiles other than the default one would fail regardless of the preference for `Use the selected profile without asking at startup`. Instead it would fail saying `An unexpected error has prevented your changes from being saved.`
Using the `-Profile` flag didn't work for me and raised `Your Firefox profile cannot be loaded. It may be missing or inaccessible.` regardless of any changes. Perhaps I'm doing something wrong here.

Please check again whether your solution actually achieved or still achieves the described results. Help is appreciated from anyone reading this. Kindly name the operating system and Firefox version you use.

spacedentist:
Yeah, this is not working at all. My guess would be that the only way to make multiple profiles work (through the `-P` option or `--ProfileManager`) would be to make the `firefox.ini` file mutable.

PR #6246, opened 2024-12-30
tex-fmt: add module

WGUNDERWOOD:
tex-fmt is a LaTeX source code formatter written in Rust, and uses a user configuration file in the .toml format.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/tex-fmt.nix
+++ b/modules/programs/tex-fmt.nix
@@ -18,7 +18,7 @@ in {
   options.programs.tex-fmt = {
     enable = mkEnableOption "tex-fmt";
 
-    package = mkPackageOption pkgs "tex-fmt" { };
+    package = mkPackageOption pkgs "tex-fmt" { nullable = true; };
 
     settings = mkOption {
       type = tomlFormat.type;
@@ -43,7 +43,7 @@ in {
   };
 
   config = lib.mkIf cfg.enable {
-    home.packages = [ cfg.package ];
+    home.packages = lib.mkIf (cfg.package != null) [ cfg.package ];
 
     home.file."${configDir}/tex-fmt/tex-fmt.toml" =
       lib.mkIf (cfg.settings != { }) {
--- a/modules/programs/tex-fmt.nix
+++ b/modules/programs/tex-fmt.nix
@@ -1,0 +1,53 @@
+{ pkgs, config, lib, ... }:
+
+let
+
+  inherit (lib) mkEnableOption mkPackageOption mkOption literalExpression;
+
+  configDir = if pkgs.stdenv.isDarwin then
+    "Library/Application Support"
+  else
+    config.xdg.configHome;
+
+  tomlFormat = pkgs.formats.toml { };
+  cfg = config.programs.tex-fmt;
+
+in {
+  meta.maintainers = with lib.maintainers; [ mirkolenz wgunderwood ];
+
+  options.programs.tex-fmt = {
+    enable = mkEnableOption "tex-fmt";
+
+    package = mkPackageOption pkgs "tex-fmt" { };
+
+    settings = mkOption {
+      type = tomlFormat.type;
+      default = { };
+      example = literalExpression ''
+        {
+          wrap = true;
+          tabsize = 2;
+          tabchar = "space";
+          lists = [];
+        }
+      '';
+      description = ''
+        Configuration written to
+        {file}`$XDG_CONFIG_HOME/tex-fmt/tex-fmt.toml` on Linux or
+        {file}`$HOME/Library/Application Support/tex-fmt/tex-fmt.toml` on Darwin.
+        See <https://github.com/WGUNDERWOOD/tex-fmt> and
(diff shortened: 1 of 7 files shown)


WGUNDERWOOD:
This is a draft PR which is waiting for https://github.com/NixOS/nixpkgs/pull/363212 to be merged.

mathjiajia:
any update?

mirkolenz (commented):
Just came across this this PR after creating a tex-fmt module for myself 😄 First of all, thank you very much for this great tool, it is so much better than latexindent!

Regarding the proposed module, I have some suggestions that could improve the interface:

- Instead of using the name `options` for custom settings, I would advocate for the name `settings` instead. The former is used to define nix module options, meaning the name could lead to confusion for some users. Using the name `settings` has also been adopted by nixpkgs: <https://github.com/NixOS/rfcs/pull/42>.
- According to the documentation, tex-fmt does not use XDG directories on macOS but instead uses the application support dir. If so, the current module would only work on Linux.
- Since you're using toml as a config language, you can use the built-in nixpkgs formats that have corresponding option types.
- The use of `with lib` in the top-level has been discouraged.
- It would be good to have some tests making sure that the config file is written correctly.

If you're interested, here is my own module. Feel free to copy snippets of it as needed (you can also add me as a co-maintainer if you'd like to). Please note that it is formatted using the new nixfmt rfc style that is not yet used in home manager.

```nix
{
  pkgs,
  config,
  lib,
  ...
}:

let

  inherit (lib)
    mkEnableOption
    mkPackageOption
    mkOption
    literalExpression
    ;

  configDir = if pkgs.stdenv.isDarwin then "Library/Application Support" else config.xdg.configHome;

  tomlFormat = pkgs.formats.toml { };
  cfg = config.programs.tex-fmt;

in
{
  meta.maintainers = with lib.maintainers; [ mirkolenz ];

  options.programs.tex-fmt = {
    enable = mkEnableOption "tex-fmt";

    package = mkPackageOption pkgs "tex-fmt" { };

    settings = mkOption {
      type = tomlFormat.type;
      default = { };
        {
          wrap = false;
          tabsize = 2;
          tabchar = "space";
          lists = [];
        }
      '';
      description = ''
        Configuration written to
        {file}`$XDG_CONFIG_HOME/tex-fmt/tex-fmt.toml` on Linux or
        {file}`$HOME/Library/Application Support/tex-fmt/tex-fmt.toml` on Darwin.
        See <https://github.com/WGUNDERWOOD/tex-fmt> and
        <https://github.com/WGUNDERWOOD/tex-fmt/blob/master/tex-fmt.toml>
        for more information.
      '';
    };
  };

    home.file."${configDir}/tex-fmt/tex-fmt.toml" = lib.mkIf (cfg.settings != { }) {
      source = tomlFormat.generate "tex-fmt-config" cfg.settings;
    };
  };
}
```

WGUNDERWOOD:
Thanks for these pointers! I will change `options` to `settings`. Note that the config directory is set within tex-fmt using [`dirs::config_dir`](https://docs.rs/dirs/latest/dirs/fn.config_dir.html); perhaps this is a known issue if it disagrees with Nix on MacOS? I am also unsure how much example config to provide: I want to show the user how to pass options in `tex-fmt.toml`, but also want to avoid adding too much to simplify maintenance in the future, should any options change.

mirkolenz:
> perhaps this is a known issue if it disagrees with Nix on MacOS

I think it's not so much related to nix on macOS but rather a difference between macOS and Linux. There are a few other applications (e.g., poetry) that behave the same way. I know that for instance uv supports both XDG config directories and the Application Support directory on macOS, but I don't know if they have custom logic in place. In the end, I don't think it's too much of an issue because this workaround I added here is also used in some other home manager modules.

>  I am also unsure how much example config to provide

I think it's sufficient to have one or two entries in here together with the links to the documentation. I just added my own config as an example here 😄

WGUNDERWOOD:
Great, thanks for this. Could you explain why you have semicolons at the end of each line in the example config? As far as I know, this is not needed in toml.

WGUNDERWOOD:
I see, thanks I just thought they might be removable since it's inside a `literalExpression`, but not if it's being parsed as Nix. I've basically just taken your example and added myself as a maintainer and changed the default settings slightly.

modules/programs/tex-fmt.nix
```diff
@@ -0,0 +1,53 @@
+{ pkgs, config, lib, ... }:
+
+let
+
+  inherit (lib) mkEnableOption mkPackageOption mkOption literalExpression;
+
+  configDir = if pkgs.stdenv.isDarwin then
```
khaneliman, line 7: Have you tested darwin with xdg enabled? Does it still write to `Library/Application Support` ?

khaneliman: Sorry, was reading above and it seems like you were talking about it that they might just ignore XDG on macos. If that's not the case, it can be fixed in a follow up.

PR #6248, opened 2024-12-31
Fix atuin daemon on darwin

utkarshgupta137:
### Description

With atuin 18.4.0, it looks for the socket file in `XDG_RUNTIME_DIR` by default. But the launchd daemon we run doesn't have access to that variable, so it is created at the default `socket_path` which is `XDG_DATA_HOME/atuin`.
To fix this, we provide a default value for the socket path which is `XDG_DATA_HOME/atuin/daemon.sock`

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@hawkw @water-sucks

### DIFF
--- a/modules/programs/atuin.nix
+++ b/modules/programs/atuin.nix
@@ -211,6 +211,13 @@ in {
         };
       })
       (mkIf isDarwin {
+        programs.atuin.settings = {
+          daemon = {
+            socket_path =
+              lib.mkDefault "${config.xdg.dataHome}/atuin/daemon.sock";
+          };
+        };
+
         launchd.agents.atuin-daemon = {
           enable = true;
           config = {


modules/misc/xdg.nix
```diff
@@ -14,6 +14,7 @@ let
   defaultCacheHome = "${config.home.homeDirectory}/.cache";
   defaultConfigHome = "${config.home.homeDirectory}/.config";
   defaultDataHome = "${config.home.homeDirectory}/.local/share";
+  defaultRuntimeHome = "${config.home.homeDirectory}/.local/run";
```
rycee, line 17: Unfortunately it is not possible to allow the user to edit the runtime directory path. It should be set by the user login process. So please revert the changes in this file.

modules/programs/atuin.nix
```diff
@@ -211,6 +211,12 @@ in {
         };
       })
       (mkIf isDarwin {
+        programs.atuin.settings = {
+          daemon = {
+            socket_path = lib.mkDefault "${config.xdg.runtimeHome}/atuin.sock";
```
rycee, line 216: Looking at the Atuin source code it seems it uses the data directory when the runtime directory is unset, so for MacOS, perhaps we could use that path instead? Unfortunately I don't know what path that is for Macs, though. Perhaps @water-sucks knows?

See https://github.com/atuinsh/atuin/blob/0056c72d89c1a01122ffdaebf15e0df6b63c8bf2/crates/atuin-common/src/utils.rs#L93-L95

rycee (commented):
Thanks for the contribution. Added a few comments.

water-sucks:
@rycee In my original PR, I went with letting `atuin` configure the daemon socket path itself at `.local/share/atuin/atuin.sock` due to this very issue of `launchd` not seeing XDG variables.

AFAIK it used to fall back to this path if `XDG_RUNTIME_DIR` is not found, but @utkarshgupta137 is right about `atuin` using `XDG_RUNTIME_DIR` only though from 18.4.0 onward (https://github.com/atuinsh/atuin/blob/05aec6f8d60409af0f58942e1550e1a6a3b3b0ce/crates/atuin-client/src/settings.rs#L730), which is quite the change, but I digress.

water-sucks (approved):

khaneliman (approved):

PR #6249, opened 2024-12-31

3ulalia:
### Description

Re: #4538, wpaperd needs something to run it after being enabled. This commit adds a systemd service to run it, and accordingly moves it to `services.wpaperd`.

In addition, the existing tests have been migrated to `services`, and an entry in the newslist has been created alerting users to this change.

@rycee pls i have deleted and recreated this PR 10 times to make it not conflict as the newsfile is updated ╥_╥

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@Avimitin

### DIFF
--- a/modules/services/copyq.nix
+++ b/modules/services/copyq.nix
@@ -25,6 +25,13 @@ in {
         otherwise the service may never be started.
       '';
     };
+
+    forceXWayland = lib.mkOption {
+      type = lib.types.bool;
+      default = true;
+      example = false;
+      description = "Force the CopyQ to use the X backend on wayland";
+    };
   };
 
   config = lib.mkIf cfg.enable {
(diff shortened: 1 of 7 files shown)


modules/services/wpaperd.nix
```diff
@@ -0,0 +1,85 @@
...
+    };
+
+    systemd.user.services.wpaperd = {
+      Install = { WantedBy = [ "graphical-session.target" ]; };
+
+      Unit = {
+        ConditionEnvironment = "WAYLAND_DISPLAY";
+        Description = "wpaperd";
+        After = [ "graphical-session-pre.target" ];
+        PartOf = [ "graphical-session.target" ];
```
thiagokokada, line 73: Please use the new `wayland.systemd.target` instead: https://github.com/nix-community/home-manager/pull/6253.

PR #6253, opened 2024-12-31
treewide: standardize wayland graphical services

thiagokokada:
### Description

Second try of PR https://github.com/nix-community/home-manager/pull/6239.

Graphical services that are meant for Wayland environment are all over the place:
- Some set `After = [ "graphical-session-pre.target" ]`, that is almost always a mistake ([see](https://github.com/Vladimir-csp/uwsm/issues/40)), others set `After = [ "graphical-session.target" ]`
- Some have `systemd.target` (or `systemdTarget`) to allow changing the default target, some does not
- Some set `ConditionEnvironment = "WAYLAND_DISPLAY"`, some does not

This PR tries to standardize this by:
- Adding `wayland.systemd.target` that is a new option for the default systemd target for all Wayland services. Defaults to `graphical-session.target`
- Set `After`, `PartOf` and `Wanted` to `cfg.systemd.target` 
- Adding `ConditionEnvironment = "WAYLAND_DISPLAY"` for services missing it

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -417,6 +417,7 @@ let
     ./systemd.nix
     ./targets/darwin
     ./targets/generic-linux.nix
+    ./wayland.nix
     ./xresources.nix
     ./xsession.nix
     ./misc/nix.nix
--- a/modules/programs/waybar.nix
+++ b/modules/programs/waybar.nix
@@ -199,7 +199,8 @@ in {
 
     systemd.target = mkOption {
       type = str;
-      default = "graphical-session.target";
+      default = config.wayland.systemd.target;
+      defaultText = literalExpression "config.wayland.systemd.target";
       example = "sway-session.target";
       description = ''
         The systemd target that will automatically start the Waybar service.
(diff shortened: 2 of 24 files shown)


thiagokokada:
CC @Lillecarl @rycee.

thiagokokada (commented):
I excluded `dunst` from this PR because it supports both X11 and Wayland and it is working fine for me the way it is currently.

teto (approved):
I am not knowledgeable in the area but this looks better than current state so fine by me (not tested).

Lillecarl (approved):
I don't have a machine to test myself, but it's a straightforward change in the right direction. LGTM!

rycee:
Thanks, very nice! Merged to master now 🙂

Lillecarl:
@jmickelin I guess that makes sense, but if we really think about it I think we should just deprecate the unit option from all children since you can use:
```
systemd.service.waybar.unitConfig.RequiredBy = lib.mkForce [ "somethingElse.unit"];
```
(written on phone so read it with imagination :))

PR #6257, opened 2025-01-02
pay-respects: integrate alias configuration and fix nushell integration

jpteb:
### Description

This pr adds the possibility to configure the alias to use pay-respects for all shells. The default is `f`.

This commit also fixes the nushell integration, as pay-respects tries to write the necessary alias command by itself into the config file. As a solution we capture the output of the command and write that output to the config with home-manager, thus resulting in the correct alias and using the "correct" (for now) output of pay-respects. This might break in the future, depending on the stability of the output.

Also for some reason the test does not generate the output from `pay-respects nushell`. I don't know how to debug that unfortunately, but it should work. I would appreciate if nushell users could test this, so it doesn't become a "works-on-my-machine". Maybe @SimonYde could find the time.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@ALameLlama

### DIFF
--- a/tests/modules/programs/pay-respects/integration-enabled.nix
+++ b/tests/modules/programs/pay-respects/integration-enabled.nix
@@ -13,21 +13,23 @@
     assertFileExists home-files/.bashrc
     assertFileContains \
       home-files/.bashrc \
-      'eval "$(@pay-respects@/bin/dummy bash --alias)"'
+      'eval "$(@pay-respects@/bin/dummy bash --alias f)"'
 
     assertFileExists home-files/.zshrc
     assertFileContains \
       home-files/.zshrc \
-      'eval "$(@pay-respects@/bin/dummy zsh --alias)"'
+      'eval "$(@pay-respects@/bin/dummy zsh --alias f)"'
 
     assertFileExists home-files/.config/fish/config.fish
     assertFileContains \
       home-files/.config/fish/config.fish \
-      '@pay-respects@/bin/dummy fish --alias | source'
+      '@pay-respects@/bin/dummy fish --alias f | source'
 
     assertFileExists home-files/.config/nushell/config.nu
     assertFileContains \
       home-files/.config/nushell/config.nu \
-      '@pay-respects@/bin/dummy nushell --alias [<alias>]'
+      'alias f ='
+      # For some reason the test does not generate the pay-respects output
+      # 'alias f = with-env { _PR_LAST_COMMAND : (history | last).command,_PR_ALIAS : "",_PR_SHELL : nu } { @pay-respects@/bin/dummy }'
   '';
 }
--- a/modules/programs/pay-respects.nix
+++ b/modules/programs/pay-respects.nix
@@ -1,9 +1,15 @@
 { config, lib, pkgs, ... }:
 let
-  inherit (lib) mkEnableOption mkPackageOption getExe optionalString mkIf;
+  inherit (lib)
+    mkEnableOption mkPackageOption mkOption getExe optionalString mkIf;
 
   cfg = config.programs.pay-respects;
   payRespectsCmd = getExe cfg.package;
+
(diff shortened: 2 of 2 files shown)


tests/modules/programs/pay-respects/integration-enabled.nix
```diff
@@ -13,21 +13,23 @@
...
-      '@pay-respects@/bin/dummy fish --alias | source'
+      '@pay-respects@/bin/dummy fish --alias f | source'
 
     assertFileExists home-files/.config/nushell/config.nu
     assertFileContains \
       home-files/.config/nushell/config.nu \
-      '@pay-respects@/bin/dummy nushell --alias [<alias>]'
+      'alias f ='
+      # For some reason the test does not generate the pay-respects output
+      # 'alias f = with-env { _PR_LAST_COMMAND : (history | last).command,_PR_ALIAS : "",_PR_SHELL : nu } { @pay-respects@/bin/dummy }'
```
ALameLlama, line 33: I believe this isn't working since the test doesn't actually download the binary.

so trying to pipe this, it doesn't return anything in the test
```
payRespectsNushellOutput = lib.readFile (pkgs.runCommand "pr-output" { }
    "${payRespectsCmd} nushell 1>$out || true");
```

Might be to be better to use something similar to what "thefuck" did and have it hardcoded

```
      nushell.extraConfig = ''
        ${optionalString cfg.enableNushellIntegration ''
          alias ${cfg.alias} = "with-env { _PR_LAST_COMMAND : (history | last).command,_PR_ALIAS : "",_PR_SHELL : nu } { ${payRespectsCmd} }"
        ''}
      '';
```

Or I guess you can have it fall back to the current output, something like this?
```
  payRespectsNushellOutput = lib.readFile (pkgs.runCommand "pr-output" { } ''
    ${payRespectsCmd} nushell 1>$out || echo "with-env { _PR_LAST_COMMAND : (history | last).command, _PR_ALIAS : \"\", _PR_SHELL : nu } { ${payRespectsCmd} }" > $out
  '');
```

ALameLlama (changes requested):

SimonYde:
Using the implemetation results in the line:
```nu
alias f =
```
It would appear that the substitution is not working. Furthermore, what the nushell implementation does when using `pay-respects nushell --alias d` is generate a nushell function that calls pay-respects in the way it expects. It does _not_ define an alias in the nushell sense. 
This is true as off pay-respects version 0.6.7, which is the current latest version. It is however not the way that it works on current nixpkgs-unstable version of pay-respects.

I think perhaps it would be wise to wait for the latest version to be merged into nixpkgs, before making the nushell integration work properly.

instead what we should do is the following:
```nix
    programs.nushell = {
      extraEnv = # nu
        ''
          let pay_respects_cache = "${config.xdg.cacheHome}/pay-respects"
          if not ($pay_respects_cache | path exists) {
            mkdir $pay_respects_cache
          }
          ${getExe cfg.package} nushell --alias ${cfg.alias} | save -f $"($pay_respects_cache)/init.nu"
        '';
      extraConfig = # nu
        ''
          source ${config.xdg.cacheHome}/pay-respects/init.nu
        '';
    };

```

SimonYde:
it would be ideal if we could split [this pull](https://github.com/NixOS/nixpkgs/pull/357425) into one that updates the package separate from the nixos changes.

jpteb:
Okay, it seems pay-respects manages the nushell config differently than I initially thought. I’d agree that we should wait for the update in nixpkgs, and then tackle the problem again.

SigmaSquadron:
> it would be ideal if we could split [this pull](https://github.com/NixOS/nixpkgs/pull/357425) into one that updates the package separate from the nixos changes.

Will do. Gimme a few hours; I'm packaging something else right now.

PR #6267, opened 2025-01-04
easyeffects: add option to import presets

hauskens:
### Description
- added options to import presets
- added tests
- added hausken as maintainer

### Checklist

(tested with `nix-shell --pure tests -A run.easyeffects-service` and `nix-shell --pure tests -A run.easyeffects-example-preset`)

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@fufexan

### DIFF
--- a/modules/services/easyeffects.nix
+++ b/modules/services/easyeffects.nix
@@ -54,7 +54,7 @@ let
     '';
   };
 in {
-  meta.maintainers = [ maintainers.fufexan ];
+  meta.maintainers = with maintainers; [ fufexan hausken ];
 
   options.services.easyeffects = {
     enable = mkEnableOption ''
--- a/modules/services/easyeffects.nix
+++ b/modules/services/easyeffects.nix
@@ -1,19 +1,19 @@
 { config, lib, pkgs, ... }:
-
-with lib;
-
 let
+  inherit (lib) literalExpression mkOption types;
 
   cfg = config.services.easyeffects;
 
-  presetOpts = optionalString (cfg.preset != "") "--load-preset ${cfg.preset}";
+  presetOpts =
+    lib.optionalString (cfg.preset != "") "--load-preset ${cfg.preset}";
 
   jsonFormat = pkgs.formats.json { };
 
   presetType = let baseType = types.attrsOf jsonFormat.type;
   in baseType // {
     check = v:
-      baseType.check v && elem (head (attrNames v)) [ "input" "output" ];
+      baseType.check v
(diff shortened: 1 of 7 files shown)


modules/services/easyeffects.nix
```diff
@@ -8,8 +8,59 @@ let
...
+              "input-gain" = 0.0;
+              "model-path" = "";
+              "output-gain" = 0.0;
+              release = 20.0;
+              "vad-thres" = 50.0;
+              wet = 0.0;
+            };
+          };
+        };
+      }]
```
rycee, line 58: Maybe this would be prettier as an attribute set, where the preset name is the attribute key. I.e., something like

```suggestion
      {
        my-preset = {
          input = {
            blocklist = [

            ];
            "plugins_order" = [
              "rnnoise#0"
            ];
            "rnnoise#0" = {
              bypass = false;
              "enable-vad" = false;
              "input-gain" = 0.0;
              "model-path" = "";
              "output-gain" = 0.0;
              release = 20.0;
              "vad-thres" = 50.0;
              wet = 0.0;
            };
          };
        };
      }
```

modules/services/easyeffects.nix
```diff
@@ -35,6 +86,9 @@ in {
         Will likely need to launch easyeffects to initially create preset.
       '';
     };
+
+    extraInputPresets = presetOptionType;
+    extraOutputPresets = presetOptionType;
```
rycee, line 91: Maybe this could simply be
```suggestion
    extraPresets = presetOptionType;
```
where the "input" or "output" field inside the preset determine whether it is for input or output?

rycee: That is, the type of this option would be something like `attrsOf presetType` where `presetType` would be the regular JSON type with the additional requirement that it is an object containing either an "input" or "output" field. In practice I think this would look something like

``` nix
presetType = let baseType = attrsOf jsonFormat.type; in baseType // {
  check = v: baseType.check v && elem (attrNames v) == [ ["input"] ["output"] ];
  description = "EasyEffects input or output JSON preset";
}
```

Completely untested so I'm not certain it will work very well.

hauskens: Thanks for the input, i really appreciate it! 
I agree, it makes much more sense to have it in one attribute set, so i've modified it based on your suggestions. 
I'm fairly inexperienced with programming in nix, so this feedback helped a lot. 
Let me know if there is anything that comes to mind

PR #6269, opened 2025-01-04
ci: add DeterminateSystems/magic-nix-cache-action

thiagokokada:
### Description

There is zero reason why the current CI for home-manager is that slow. We can cache the CI results and basically get a very fast CI (e.g.: only run tests for the changed code) as long as we store the `/nix/store` somewhere and restore before running the tests.

This PR adds [DeterminateSystems/magic-nix-cache-action](https://github.com/DeterminateSystems/magic-nix-cache) to do so.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,6 +15,7 @@ jobs:
     - uses: cachix/install-nix-action@v30
       with:
         nix_path: nixpkgs=channel:nixos-unstable
+    - uses: DeterminateSystems/magic-nix-cache-action@v8
     - run: |
         if grep -R --exclude stdlib-extended.nix literalExample modules ; then
           echo "Error: literalExample should be replaced by literalExpression" > /dev/stderr


rycee:
We actually used to have caching but I removed it at some point (<https://github.com/nix-community/home-manager/pull/5093>) because it was actually slowing things down.

Maybe it's worth trying to re-enable it. But I suspect it still is not a big win.

Would be great to make the tests have a more reasonable runtime so well worth investigating more deeply. E.g., how to reduce the number of evaluations of Nixpkgs.

thiagokokada:
> Maybe it's worth trying to re-enable it. But I suspect it still is not a big win.

I amended the commit to re-trigger the tests, let's see. I imagine that this new run should be significantly faster than the previous one (that took ~43 min), but if there is no change it wouldn't be worth it.

thiagokokada:
Took basically the same time, yes, it seems it is not worth it at all.

PR #6275, opened 2025-01-06
treewide: stub tests

thiagokokada:
### Description

Stub some tests to make the test suite run faster.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/ranger.nix
+++ b/modules/programs/ranger.nix
@@ -142,7 +142,7 @@ in {
   config = mkIf cfg.enable (mkMerge [
     {
       programs.ranger.finalPackage = cfg.package.overrideAttrs (oldAttrs: {
-        propagatedBuildInputs = oldAttrs.propagatedBuildInputs
+        propagatedBuildInputs = (oldAttrs.propagatedBuildInputs or [ ])
           ++ cfg.extraPackages;
       });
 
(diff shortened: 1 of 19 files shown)


teto:
I am not a fan of the stubs, it adds friction to writing the tests. What I wonder is why do we have to rerun all the tests everytime, with 45mn pipelines on each PR push. Can't the tests be cached ? could we use the nix-community builders/build-bot instead ? HM is the most LGTM candidate for those builders IMO.
I am not willing to do it myself and stubs are an easier short win so we can merge this but it would be nice to have someone motivated to work on a more profound fix.

teto (approved):

thiagokokada:
> What I wonder is why do we have to rerun all the tests everytime, with 45mn pipelines on each PR push. Can't the tests be cached ?

I tried to add a cache in https://github.com/nix-community/home-manager/pull/6269 but the results were that there was no gain. I am not sure why though, it seems what is taking time is not the builds themselves since most tests are stubbed already (and the few that aren't generally only download 1 or 2 small packages anyway), but the evaluation of each test (and evaluation in Nix is single threaded).

I am not sure how evaluation can be fixed though. Maybe https://github.com/nix-community/nix-eval-jobs?

Issue #4198, opened 2023-07-04
bug: Breakage from changed checkout due to flake path: prefix.


galenhuntington:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

The PR #3646 appears to have changed how a repo is checked out.  Formerly, the entire repo in which the flake lived would be copied into `/nix/store/*-source/`.  (This feature, by the way, surprised me at first and induced me to move my Nix config into its own repo so as to not grab other material I was keeping under revision control.)

After the last commit, the checkout only happens relative to the directory the flake is in, rather than the entire repo.  I currently have the flake files and expressions in their own directory, with my `home.nix` having references such as `../programs/foobar` to pull in other system files and programs.  These relative paths now error out with `error: access to absolute path '***' is forbidden in pure eval mode (use '--impure' to override)`.  Note that, contrary to the error, `--impure` won't work either since it points to the `/nix/` directory rather than the files, so the messaging here is also wrong (but perhaps out of scope).

This can be worked around.  The directory tree can be rearranged.  And perhaps it's best practice to have `flake.nix` in the repo root.  But it seems to be a major breaking change which should at least be announced.

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 5.4.0-150-generic, Ubuntu, 18.04.6 LTS (Bionic Beaver), nobuild`
 - multi-user?: `no`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.13.3`
```

dongcarl:
I'm the author of #3646, w/re this I have a few thoughts:

1. The immediate solution is: `home-manager switch --flake git+file:///$HOME/.config/home-manager`
2. The best solution is for upstream (`nix` itself) to adopt the following flake resolution logic:
    1. When a flake URL's `type` field is not specified (e.g. `nix build /path/to/flake`), try to see if it is a `git` repository first (e.g. if `/path/to/flake/.git` exists`), if so, use `git+file:///path/to/flake` as the effective URL
    2. If not, try to see if `/path/to/flake/flake.nix` exists, if so, use `path:/path/to/flake` as the effective URL
    3. If still not (the git repo might be a parent of `/path/to/flake`), use `git+file:///path/to/flake` as the effective URL
    This will allow us to revert #3646 and have all the cases still work fine

NobbZ:
[...]

PR #6277, opened 2025-01-06
Revert "home-manager: Use path: URI type for flake default"

dongcarl:
From the discussions in https://github.com/nix-community/home-manager/issues/4198, it seems like folks prefer the default Nix behaviour, and there's been no progress on https://github.com/NixOS/nix/issues/8710

So let's revert https://github.com/nix-community/home-manager/pull/3646 (I was the original author)


NobbZ (approved):

rycee:

PR #6278, opened 2025-01-06

3ulalia:
### Description

Re: #4538, wpaperd needs something to run it after being enabled. This commit adds a systemd service to run it, and accordingly moves it to services.wpaperd.

In addition, the existing tests have been migrated to services, and an entry in the newslist has been created alerting users to this change.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@Avimitin

### DIFF
--- a/modules/po/de.po
+++ b/modules/po/de.po
@@ -8,8 +8,8 @@ msgstr ""
 "Project-Id-Version: Home Manager Modules\n"
 "Report-Msgid-Bugs-To: https://github.com/nix-community/home-manager/issues\n"
 "POT-Creation-Date: 2025-01-03 09:09+0100\n"
-"PO-Revision-Date: 2024-10-15 21:37+0000\n"
-"Last-Translator: Viktor Illmer <accounts@viktor.im>\n"
+"PO-Revision-Date: 2025-01-05 15:00+0000\n"
+"Last-Translator: programmerlexi <lexi@reyer.name>\n"
 "Language-Team: German <https://hosted.weblate.org/projects/home-manager/"
 "modules/de/>\n"
 "Language: de\n"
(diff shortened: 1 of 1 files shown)


modules/services/wpaperd.nix
```diff
@@ -38,12 +50,35 @@ in {
...
       };
     };
+
+    systemd.user.services.wpaperd = {
+      Install = { WantedBy = [ "wayland.systemd.target" ]; };
+
+      Unit = {
+        ConditionEnvironment = "WAYLAND_DISPLAY";
+        Description = "wpaperd";
+        PartOf = [ "wayland.systemd.target" ];
```
thiagokokada, line 72:
```suggestion
        PartOf = [ config.wayland.systemd.target ];
        After = [ config.wayland.systemd.target ];
```

modules/services/wpaperd.nix
```diff
@@ -38,12 +50,35 @@ in {
...
     home.packages = [ cfg.package ];
 
     xdg.configFile = {
       "wpaperd/wallpaper.toml" = mkIf (cfg.settings != { }) {
         source = tomlFormat.generate "wpaperd-wallpaper" cfg.settings;
       };
     };
+
+    systemd.user.services.wpaperd = {
+      Install = { WantedBy = [ "wayland.systemd.target" ]; };
```
thiagokokada, line 67:
```suggestion
      Install = { WantedBy = [ config.wayland.systemd.target ]; };
```

modules/services/wpaperd.nix
```diff
@@ -3,12 +3,24 @@
 with lib;
```
thiagokokada, line 3: In general a top-level `with lib` is discouraged.

teto:
I dont have a strong opinion on this but it might make sense to have `programs` AND `service` separate in case you want to generate a config without the service. That would be for wpaperd to decide.

Avimitin:
In my daily use case, wpaperd is usually not use as a standalone program. So from my own perspective, for those users who wants to use home-manager to manage their wpaperd config, they might also want the systemd user service automatically installed. So I think having only a single service module is ok.

PR #6288, opened 2025-01-08
helm: init module

folliehiyuki:
### Description

This PR adds a new module `programs.helm` to install Helm package and manage Helm plugins.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/lib/maintainers.nix
+++ b/modules/lib/maintainers.nix
@@ -628,4 +628,10 @@
     github = "ALameLlama";
     githubId = 55490546;
   };
+  folliehiyuki = {
+    name = "Hoang Nguyen";
+    email = "folliekazetani@protonmail.com";
+    github = "folliehiyuki";
+    githubId = 67634026;
+  };
 }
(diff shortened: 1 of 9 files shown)


teto:
before starting a review do you need this even when `wrapHelm` exists ?

folliehiyuki:
I didn't know about `wrapHelm`. The wrapper approach overrides HELM_PLUGINS variable, pointing it the the Nix store, so it doesn't allow users to install a random Helm plugin with an one-off `helm plugin install` command.

With that said, I only use a few Helm plugins and they are all available in Nixpkgs. If no one desires the interactive CLI usage, I'll close this PR.

Thanks for the info btw!

teto:
> I didn't know about wrapHelm. The wrapper approach overrides HELM_PLUGINS variable, pointing it to the Nix store, so it doesn't allow users to install a random Helm plugin with an one-off helm plugin install command.

Thanks that would be a valid usecase for the HM module then. Since you dont need this module after all, I'll close but anyone who needs it feel to reuse and resubmit.

PR #6291, opened 2025-01-08
fish: expose session variables package

ian-h-chamberlain:
### Description

Expose the builtin fish-translated `hm-session-vars.fish` as a package, and include it by default when `programs.fish.enable = true`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/fish.nix
```diff
@@ -369,12 +369,15 @@ let
       passAsFile = [ "text" ];
     } "env HOME=$(mktemp -d) fish_indent < $textPath > $out";
 
-  translatedSessionVariables = pkgs.runCommandLocal "hm-session-vars.fish" { } ''
+  profileDir = "/etc/profile.d";
```
teto: it seems overkill to have 2 variables seldomly used with profileDir and sessionsVarsFile. Maybe just hardcode profileDir at least.

ian-h-chamberlain: Thanks for the review! I consolidated to one variable for `etc/profile.d/hm-session-vars.fish` instead

PR #6292, opened 2025-01-08
thunderbird: add native host support

booxter:
### Description

This change allows to enable native hosts modules for Thunderbird, similar to Firefox. For reference, [this](https://github.com/booxter/nix/blob/85096699428b0c8e849dfa2992b9e02895360012/modules/home-manager/modules/thunderbird.nix#L5) is how it looks like in nix config.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@d-dervishi @jkarlson

### DIFF
--- a/tests/modules/programs/thunderbird/thunderbird.nix
+++ b/tests/modules/programs/thunderbird/thunderbird.nix
@@ -41,6 +41,9 @@
     # Disable warning so that platforms' behavior is the same
     darwinSetupWarning = false;
 
+    # Darwin doesn't support wrapped Thunderbird, using unwrapped instead
+    package = pkgs.thunderbird-unwrapped;
+
     profiles = {
       first = {
         isDefault = true;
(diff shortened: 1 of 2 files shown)


tests/modules/programs/thunderbird/thunderbird.nix
```diff
@@ -65,8 +65,6 @@
     };
   };
 
-  test.stubs.thunderbird = { };
-
   nmt.script = ''
```
teto, line 68: would it be possible to test it with a plugin that requires native messenging ?

booxter: I've added a check using a firefox native host. I don't think it matters to the test if it's not TB since they are packaged the same way. (There are no native hosts in nixpkgs or elsewhere that we could use right now. In the future, when NUR or nixpkgs support some native hosts, we may switch to a "proper" one.)

khaneliman (approved):

PR #6298, opened 2025-01-10
ghostty: don't install package on darwin platforms

### LINKED ISSUE #6295: bug: ghostty package should be nullable for unsupported platforms

### Issue description

I believe that the `package` option for the ghostty module should be "nullable", as ghostty is not supported on macos through nixpkgs, but you can still use the hm module manage your config while having it installed via another method.

Currently, you get an error about macos being broken, as it tries to install it.

### Maintainer CC

@HeitorAugustoLN

### System information

```markdown
- system: `"aarch64-darwin"`
 - host os: `Darwin 24.1.0, macOS 15.1.1`
 - multi-user?: `yes`
 - sandbox: `no`
 - version: `nix-env (Nix) 2.24.10`
 - nixpkgs: `/Users/mitchell/.nix-defexpr/channels/nixpkgs`
```

HeitorAugustoLN:
### Description

Closes #6295

Makes ghostty package option nullable on darwin platform

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/ghostty.nix
+++ b/modules/programs/ghostty.nix
@@ -13,7 +13,7 @@ in {
   options.programs.ghostty = {
     enable = lib.mkEnableOption "Ghostty";
 
-    package = lib.mkPackageOption pkgs "ghostty" { };
+    package = lib.mkPackageOption pkgs "Ghostty" { default = [ "ghostty" ]; };
 
     settings = lib.mkOption {
       inherit (keyValue) type;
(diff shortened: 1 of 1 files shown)


teto:
rather than making it nullable, it could be added to home.packages only for supported platforms ?

HeitorAugustoLN:
Closing in favor of: #6300

PR #6300, opened 2025-01-10
ghostty: allow darwin users to manager their config

### LINKED ISSUE #6295: bug: ghostty package should be nullable for unsupported platforms

anund:
### Description

Currently nixpkgs does not contain a package definition for ghostty compatible with Darwin. Darwin users may still want to use this module to manage their config or share config between systems. This carries over behaviour from the beta period where this same technique was used.

see: https://github.com/clo4/ghostty-hm-module/blob/887e13a6e7acf5ffaab0119d96e476d84db90904/module.nix#L167-L173

Also improves validation to cover theme files.

Requested via ghostty discord. Testing included building against a local copy of nixpkgs with the ghostty package deleted and marked as broken.

### Checklist

- [x] Commit messages are formatted correctly

#### Maintainer CC

cc @HeitorAugustoLN 

closes #6295

### DIFF
--- a/modules/programs/ghostty.nix
+++ b/modules/programs/ghostty.nix
@@ -13,7 +13,11 @@ in {
   options.programs.ghostty = {
     enable = lib.mkEnableOption "Ghostty";
 
-    package = lib.mkPackageOption pkgs "ghostty" { };
+    package = lib.mkPackageOption pkgs "ghostty" {
+      nullable = true;
+      extraDescription =
+        "On darwin set this to null to manage ghostty config while ghostty is not present in nixpkgs.";
+    };
 
     settings = lib.mkOption {
       inherit (keyValue) type;
@@ -78,7 +82,9 @@ in {
     installBatSyntax =
       lib.mkEnableOption "installation of Ghostty configuration syntax for bat"
       // {
-        default = true;
+        default = cfg.package != null;
+        defaultText =
+          "Enabled by default when programs.ghostty.package is not null.";
       };
 
     enableBashIntegration = lib.mkEnableOption ''
@@ -108,7 +114,7 @@ in {
 
   config = lib.mkIf cfg.enable (lib.mkMerge [
     {
-      home.packages = [ cfg.package ];
+      home.packages = lib.mkIf (cfg.package != null) [ cfg.package ];
 
       programs.ghostty.settings = lib.mkIf cfg.clearDefaultKeybinds {
         keybind = lib.mkBefore [ "clear" ];
@@ -116,27 +122,47 @@ in {
 
       # MacOS also supports XDG configuration directory, so we use it for both
       # Linux and macOS to reduce complexity
-      xdg.configFile = lib.mkMerge [
+      xdg.configFile = let
+        validate = file:
+          lib.mkIf (cfg.package != null) "${
+            lib.getExe cfg.package
+          } +validate-config --config-file=${config.xdg.configHome}/ghostty/${file}";
+      in lib.mkMerge [
         {
           "ghostty/config" = lib.mkIf (cfg.settings != { }) {
             source = keyValue.generate "ghostty-config" cfg.settings;
-            onChange = "${lib.getExe cfg.package} +validate-config";
+            onChange = validate "config";
           };
         }
 
         (lib.mkIf (cfg.themes != { }) (lib.mapAttrs' (name: value: {
           name = "ghostty/themes/${name}";
-          value.source = keyValue.generate "ghostty-${name}-theme" value;
+          value = {
+            source = keyValue.generate "ghostty-${name}-theme" value;
+            onChange = validate "themes/${name}";
+          };
         }) cfg.themes))
       ];
     }
 
     (lib.mkIf cfg.installVimSyntax {
-      programs.vim.plugins = [ cfg.package.vim ];
+      assertions = [{
+        assertion = lib.elem "vim" cfg.package.outputs or [ ];
+        message =
+          "programs.ghostty.package does not have a vim output or is null.";
+      }];
+
+      programs.vim.plugins = lib.mkIf (cfg.package != null) [ cfg.package.vim ];
     })
 
     (lib.mkIf cfg.installBatSyntax {
-      programs.bat = {
+      assertions = [{
+        assertion = cfg.package != null;
+        message =
+          "Cannot install bat syntax when programs.ghostty.package is null.";
+      }];
+
+      programs.bat = lib.mkIf (cfg.package != null) {
         syntaxes.ghostty = {
           src = cfg.package;
           file = "share/bat/syntaxes/ghostty.sublime-syntax";
--- a/modules/programs/ghostty.nix
+++ b/modules/programs/ghostty.nix
@@ -13,7 +13,11 @@ in {
   options.programs.ghostty = {
     enable = lib.mkEnableOption "Ghostty";
 
-    package = lib.mkPackageOption pkgs "ghostty" { };
+    package = lib.mkPackageOption pkgs "ghostty" {
+      nullable = true;
+      extraDescription =
+        "Set programs.ghostty.package to null on platfroms where ghostty is not available or marked broken";
+    };
 
     settings = lib.mkOption {
       inherit (keyValue) type;
@@ -78,9 +82,9 @@ in {
     installBatSyntax =
       lib.mkEnableOption "installation of Ghostty configuration syntax for bat"
       // {
-        default = pkgs.stdenv.hostPlatform.isLinux;
+        default = cfg.package != null;
         defaultText =
-          lib.literalMD "`true` if host platform is Linux, if not, `false`";
+          lib.literalMD "`true` if programs.ghostty.package is not null";
       };
 
     enableBashIntegration = lib.mkEnableOption ''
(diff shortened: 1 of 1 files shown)


modules/programs/ghostty.nix
```diff
@@ -78,7 +82,7 @@ in {
     installBatSyntax =
       lib.mkEnableOption "installation of Ghostty configuration syntax for bat"
       // {
-        default = true;
+        default = (cfg.package != null);
```
rycee, line 85: Need to add a `defaultText` attribute to avoid making the documentation depend on the configuration.

modules/programs/ghostty.nix
```diff
@@ -108,35 +112,53 @@ in {
...
+      }];
+
+      programs.vim.plugins = lib.mkIf (cfg.package != null) [ cfg.package.vim ];
     })
 
     (lib.mkIf cfg.installBatSyntax {
-      programs.bat = {
+      assertions = [{
+        assertion = cfg.package != null;
+        message = "cannot install bat syntax while .package is null";
```
rycee, line 158:
```suggestion
        message = "Cannot install bat syntax when programs.ghostty.package is null.";
```

modules/programs/ghostty.nix
```diff
@@ -108,35 +112,53 @@ in {
...
+          };
         }) cfg.themes))
       ];
     }
 
     (lib.mkIf cfg.installVimSyntax {
-      programs.vim.plugins = [ cfg.package.vim ];
+      assertions = [{
+        assertion = lib.elem "vim" cfg.package.outputs or [ ];
+        message = "ghostty.package does not have a vim output or is null";
```
rycee, line 149:
```suggestion
        message = "programs.ghostty.package does not have a vim output or is null.";
```

rycee:
Seems like a duplicate of #6298?

modules/programs/ghostty.nix
HeitorAugustoLN, line 85: redundant parentheses, they can be removed.

anund: `literalExpression "(cfg.package != null)"` or `"enabled by default when package is not null`?

HeitorAugustoLN:
This is more complete than my PR in #6298. But i'd prefer to not make things conditionally if the hostPlaform is darwin.
As @teto commented on https://github.com/nix-community/home-manager/pull/6298#issuecomment-2581577321

anund:
I think `default = null` if the host is darwin might cover the user comment. Nullable was suggested for users managing config across multiple nixos/darwin systems as a ~least worst option.

modules/programs/ghostty.nix
anund: Picking option 2

modules/programs/ghostty.nix
```diff
@@ -13,7 +13,11 @@ in {
   options.programs.ghostty = {
     enable = lib.mkEnableOption "Ghostty";
 
-    package = lib.mkPackageOption pkgs "ghostty" { };
+    package = lib.mkPackageOption pkgs "ghostty" {
+      nullable = true;
+      extraDescription =
+        "On darwin set this to null to manage ghostty config while ghostty is not present in nixpkgs.";
+    };
```
HeitorAugustoLN, line 20: Just to capitalize ghostty in the description
```suggestion
    package = lib.mkPackageOption pkgs "Ghostty" { default = [ "ghostty" ]; };
```

modules/programs/ghostty.nix
```diff
@@ -78,7 +82,9 @@ in {
     installBatSyntax =
       lib.mkEnableOption "installation of Ghostty configuration syntax for bat"
       // {
-        default = true;
+        default = cfg.package != null;
+        defaultText =
+          "Enabled by default when programs.ghostty.package is not null.";
```
HeitorAugustoLN, line 87:
```suggestion
        default = pkgs.stdenv.hostPlatform.isLinux;
        defaultText = lib.literalMD "`true` if host platform is Linux, if not, `false`";
```

modules/programs/ghostty.nix
```diff
@@ -108,35 +114,55 @@ in {
 
   config = lib.mkIf cfg.enable (lib.mkMerge [
     {
-      home.packages = [ cfg.package ];
+      home.packages = lib.mkIf (cfg.package != null) [ cfg.package ];
```
HeitorAugustoLN, line 117:
```suggestion
      home.packages = lib.optionals pkgs.stdenv.hostPlatform.isLinux [ cfg.package ];
```

modules/programs/ghostty.nix
```diff
@@ -108,35 +114,55 @@ in {
...
 
     (lib.mkIf cfg.installVimSyntax {
-      programs.vim.plugins = [ cfg.package.vim ];
+      assertions = [{
+        assertion = lib.elem "vim" cfg.package.outputs or [ ];
+        message =
+          "programs.ghostty.package does not have a vim output or is null.";
+      }];
+
+      programs.vim.plugins = lib.mkIf (cfg.package != null) [ cfg.package.vim ];
```
HeitorAugustoLN, line 155:
```suggestion
      assertions = [{
        assertion = cfg.installVimSyntax -> pkgs.stdenv.hostPlatform.isLinux;
        message =
          "programs.ghostty.installVimSyntax cannot be enabled on darwin as it depends on the package";
      }];

      programs.vim.plugins = lib.optionals pkgs.stdenv.hostPlatform.isLinux [ cfg.package.vim ];
```

modules/programs/ghostty.nix
```diff
@@ -108,35 +114,55 @@ in {
...
 
     (lib.mkIf cfg.installBatSyntax {
-      programs.bat = {
+      assertions = [{
+        assertion = cfg.package != null;
+        message =
+          "Cannot install bat syntax when programs.ghostty.package is null.";
+      }];
+
+      programs.bat = lib.mkIf (cfg.package != null) {
```
HeitorAugustoLN, line 165:
```suggestion
      assertions = [{
        assertion = cfg.installBatSyntax -> pkgs.stdenv.hostPlatform.isLinux;
        message =
          "programs.ghostty.installBatSyntax cannot be enabled on darwin as it depends on the package";
      }];

      programs.bat = lib.mkIf pkgs.stdenv.hostPlatform.isLinux {
```

HeitorAugustoLN (commented):
Here is how I think this should be handled:

modules/programs/ghostty.nix
anund: That doesn't appear to work when nixpkgs is marking the package as [broken](https://github.com/NixOS/nixpkgs/blob/d5d576db277dbfb1463b47d77fddba818eff69f7/pkgs/by-name/gh/ghostty/package.nix#L199). Setting the package to null in config isn't allowed (not nullable) and nixpkgs evaluates the default since it _is_ present. This does work when the package isn't present at all.

HeitorAugustoLN: It won't be present look at: https://github.com/nix-community/home-manager/pull/6300?notification_referrer_id=NT_kwDOAqUkqrQxNDIwOTgxOTU1Mjo0NDM3NzI1OA#discussion_r1910173889

anund: Testing.

anund: Yup, my mistake on where the actual eval was coming from. "Ghostty" isn't needed if we're guarding all the package sites. `mkPackageOption` currently only checks for presence not whether or not the package is broken.

I considered checking if the package is broken instead and looked at [~hostPlatform](https://github.com/NixOS/nixpkgs/blob/c0b4b9614390535f61f103fff1e48255a588fa4b/lib/systems/inspect.nix#). I think Linux is the only supported output from `hostPlatform` so that doesn't seem to buy much generality.

anund: Turns out this trying to avoid installing only on linux brings in different complications. When we just gate on `isLinux` we add a synchronization point with nixpkgs. If nixpkgs gains a darwin version of ghostty (as some are working on) this module will have to change at the ~same as it's introduced into nixpkgs. Addtionally, to a lesser degree, it also interferes with users ability to set their own package if they figure out some work around for themselves. If, say, a *bsd or darwin user gets a package working for themselves before it lands in nixpkgs they would not be able to use the module as is.

There's a bit of weirdness around a transition from broken -> working for macos users too. If a nixpkgs copy exists some users will want that copy and some will want the macos managed version instead. (technically this could also be the case for nix on Arch)

To me this suggests setting `nullable: true` and handling the null case. Trying to default to the 'right' value for darwin users runs into that point above around choosing darwin or nix hosted packages. My user facing suggestion for that is adding an assertion: ~"isBroken, isNull`-> "programs.ghostty.package is marked as broken on your platform, set the value to null to manage your configuration without installing the package"

spikespaz: Handling null for `package` is exactly the path I have chosen for my private Ghostty module. I support that route.

spikespaz:
Perhaps allow setting `ghosttyBinary` with a default of `/run/current-system/sw/bin/ghostty` if `isDarwin`, and on linux use `lib.getExe' "ghostty" cfg.package`?

HeitorAugustoLN (approved):

mhanberg:
Should close #6295 (maybe edit the description with this so it auto closes my issue when this is merged 😃)

Thanks for this!

isabelroses (approved):

khaneliman (approved):
LGTM, resolves issue of broken symlinks on darwin for me and don't have to pass `pkgs.emptyDirectory` for the package option.

PR #6301, opened 2025-01-10
dconf: support configuring specific user databases

DDoSolitary:
### Description

By default, dconf uses `$XDG_CONFIG_HOME/dconf/user` as the user database, but this can be changed by specifying `user-db:<name>` in a profile file and setting the `DCONF_PROFILE` environment variable to that profile. One may want to [use different user databases for different DE/WMs](https://wiki.archlinux.org/title/Hyprland#Separate_dconf_profile) to avoid collision.

Currently the module invokes dconf without touching `DCONF_PROFILE`, which means that 1) it is unable to configure multiple different user databases, and 2) the behavior of activation script will be affected by the `DCONF_PROFILE` environment variable when it is invoked, possibly leading to undesired results.

This PR adds a `dconf.databases` option, so that settings under `dconf.databases.<name>` will be written to `$XDG_CONFIG_HOME/dconf/<name>`. The old `dconf.settings` option is left as-is to avoid breaking compatibility.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/misc/dconf.nix
+++ b/modules/misc/dconf.nix
@@ -13,10 +13,23 @@ let
   # The dconf keys managed by this configuration. We store this as part of the
   # generation state to be able to reset keys that become unmanaged during
   # switch.
-  stateDconfKeys = pkgs.writeText "dconf-keys.json" (builtins.toJSON
-    (concatLists (mapAttrsToList
-      (dir: entries: mapAttrsToList (key: _: "/${dir}/${key}") entries)
-      cfg.settings)));
+  mkStateDconfKeys = nameSuffix: settings:
+    pkgs.writeText "dconf-keys${nameSuffix}.json" (builtins.toJSON (concatLists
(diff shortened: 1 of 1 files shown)


teto:
that looks LGTM. Maybe add a test. What would be nice is  a review from someone using it

DDoSolitary:
Rebased to master.

> that looks LGTM. Maybe add a test. What would be nice is a review from someone using it

@teto The dconf databases are modified on activation so I think it's a bit hard to test them.

DDoSolitary:
> that looks LGTM. Maybe add a test. What would be nice is a review from someone using it

@teto Added an integration test for this.

DDoSolitary:
CI failure seems to be caused by Cloudflare outage. I'll retry later.

teto (approved):

PR #6302, opened 2025-01-10

3ulalia:
### Description

See #6278 for additional conversation regarding usage of `config.wayland.systemd.target`.

Re: #4538, wpaperd needs something to run it after being enabled. This commit adds a systemd service to run it, and accordingly moves it to services.wpaperd.

In addition, the existing tests have been migrated to services, and an entry in the newslist has been created alerting users to this change.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@3ulalia


modules/programs/wpaperd.nix
```diff
@@ -1,49 +0,0 @@
-{ config, lib, pkgs, ... }:
-
-with lib;
-
-let
-  cfg = config.programs.wpaperd;
-  tomlFormat = pkgs.formats.toml { };
-in {
-  meta.maintainers = [ hm.maintainers.Avimitin ];
```
khaneliman, line 9: Should probably check if the previous maintainer is interested in maintaining still.

khaneliman: Sorry, didn't see that. Thanks

PR #6303, opened 2025-01-10
Improve Darwin story for firefox and thunderbird

booxter:
### Description

1. Allow to set null as package for thunderbird. (Common on Darwin when it's pulled from e.g. Homebrew. No need to build a dummy package.)
2. Use -unwrapped packages for Darwin where wrapped versions are not supported in nixpkgs.
3. While at it, fix example for thunderbird.package to list a package that actually exists (thunderbird-91 was removed a while ago).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


booxter:
This may no longer be needed if https://github.com/NixOS/nixpkgs/pull/377863 proves to be working.

booxter:
It won't be needed once https://github.com/NixOS/nixpkgs/pull/377863 is merged. Closing.

PR #6305, opened 2025-01-11
vscode: fix extensions dir path

ameszy:
### Description

On macOS, vscode-insiders (via `pkgs.vscode.override { isInsiders = true; }`) reads extension configuration from (and write to) `.vscode/extensions/extensions.json`, not under the `.vscode-insiders` folder. This commit affects only macOS since the author doesn't know and has no device to test the behaviour on other platform like linux.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/vscode.nix
+++ b/modules/programs/vscode.nix
@@ -20,7 +20,8 @@ let
 
   extensionDir = {
     "vscode" = "vscode";
-    "vscode-insiders" = "vscode-insiders";
+    "vscode-insiders" =
+      if pkgs.stdenv.hostPlatform.isDarwin then "vscode" else "vscode-insiders";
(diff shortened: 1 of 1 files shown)


khaneliman:
I've verified it behaves that way on darwin, but I was testing the linux side of things and noticed that it doesn't even build on linux currently because they don't generate the binary with the suffixed name. Fixing the derivation and then will confirm if this needs to just be changed for everyone.

khaneliman:
I think the derivation was just incorrectly not pulling the insider version. I have https://github.com/NixOS/nixpkgs/pull/377798 for fixing the derivation and building on my macbook I see the binary as `code-insiders` now. Could you try with that build and see if the current HM module works with it?

PR #6309, opened 2025-01-12
bash: Make sure HISTFILE's directory exists

DamienCassou:
-------

### Description

If HISTFILE's directory doesn't exist, bash history doesn't work. This PR fixes this.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/bash.nix
+++ b/modules/programs/bash.nix
@@ -180,7 +180,7 @@ in {
 
     sessionVarsStr = config.lib.shell.exportAll cfg.sessionVariables;
 
-    historyControlStr = concatStringsSep "\n"
+    historyControlStr = (concatStringsSep "\n"
       (mapAttrsToList (n: v: "${n}=${v}") ({
         HISTFILESIZE = toString cfg.historyFileSize;
         HISTSIZE = toString cfg.historySize;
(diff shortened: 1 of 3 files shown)


tests/modules/programs/bash/bash-history-control-with-file.nix
```diff
@@ -0,0 +1,21 @@
...
+{
+  config = {
+    programs.bash = {
+      enable = true;
+      historyControl = [ "erasedups" ];
+      historyFile = "/home/hm-user/foo/bash/history";
+    };
+
+    nmt.script = ''
+        assertFileExists home-files/.bashrc
```
rycee, line 14: A bit too much indentation:
```suggestion
     assertFileExists home-files/.bashrc
```

modules/programs/bash.nix
```diff
@@ -190,7 +190,10 @@ in {
         HISTCONTROL = concatStringsSep ":" cfg.historyControl;
       } // optionalAttrs (cfg.historyIgnore != [ ]) {
         HISTIGNORE = escapeShellArg (concatStringsSep ":" cfg.historyIgnore);
-      }));
+      }))) + (optionalString (cfg.historyFile != null) ''
+
+        mkdir -p "$(dirname "$HISTFILE")"
+      '');
```
rycee, line 196: Maybe sufficient to do

```suggestion
      }) ++ optional (cfg.historyFile != null)
        ''mkdir -p "$(dirname "$HISTFILE")"'');
```

rycee (commented):
Thanks! Added a few minor comments.

tests/modules/programs/bash/bash-history-control-with-file.nix
DamienCassou: fixed, thank you

modules/programs/bash.nix
DamienCassou: much simpler indeed. Thank you.

DamienCassou:
Thank you for the feedback. I applied your changes.

rycee:
Sorry about the delay. Merged to master now.

Issue #5031, opened 2024-02-16
bug: Thunderbird creates new local folders account on every launch


nfelber:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

## Issue

On every launch, Thunderbird adds a new line `user_pref("mail.account.account<N>.server", "server1");` to `prefs.js` of the profile being used, with `<N>` a value incrementing in steps of 1. The parameter `mail.account.lastKey` is also incremented to be equal to the largest `<N>`. It doesn't affect the accounts declared in the home-manager nix config.

## Example

`prefs.js` after launching and closing Thunderbird 5 times:

```js
// ...
// Lines weirdly added on each launch
user_pref("mail.account.account1.server", "server1");
user_pref("mail.account.account2.server", "server1");
user_pref("mail.account.account3.server", "server1");
user_pref("mail.account.account4.server", "server1");
user_pref("mail.account.account5.server", "server1");
// Accounts normally declared in the home-manager nix config
user_pref("mail.account.account_576ba7c2e4abb7184ca409154dbbbd5306c1a80747fbce4148ea6271fd21e776.identities", "id_576ba7c2e4abb7184ca409154dbbbd5306c1a80747fbce4148ea6271fd21e776");
user_pref("mail.account.account_576ba7c2e4abb7184ca409154dbbbd5306c1a80747fbce4148ea6271fd21e776.server", "server_576ba7c2e4abb7184ca409154dbbbd5306c1a80747fbce4148ea6271fd21e776");
user_pref("mail.account.account_6f0a53fb760ead798eff28586351f9cda5211f8d18df529aaca8391c814ba87a.identities", "id_6f0a53fb760ead798eff28586351f9cda5211f8d18df529aaca8391c814ba87a");
user_pref("mail.account.account_6f0a53fb760ead798eff28586351f9cda5211f8d18df529aaca8391c814ba87a.server", "server_6f0a53fb760ead798eff28586351f9cda5211f8d18df529aaca8391c814ba87a");
user_pref("mail.account.account_8758902c10feb2ee51fb2a4a0afcde74fd6b11057e0be55d98d4d85335e8deee.identities", "id_8758902c10feb2ee51fb2a4a0afcde74fd6b11057e0be55d98d4d85335e8deee");
[...]

PR #6310, opened 2025-01-12
thunderbird: add accountsOrder option

nfelber:
### Description

Fixes https://github.com/nix-community/home-manager/issues/5031 and adds `accountsOrder` option for declarative account ordering in Thunderbird's folder pane as discussed in the issue.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@d-dervishi @jKarlson


modules/programs/thunderbird.nix
```diff
@@ -203,6 +203,23 @@ in {
...
               };
 
+              accountsOrder = mkOption {
+                type = types.listOf types.str;
+                default = [ ];
+                description = ''
+                  Custom ordering of accounts in Thunderbird's folder pane.
+                  Enabled accounts that aren't listed here appear in an arbitrary
+                  order after the ordered accounts.
+                '';
```
alois31, line 213: This should explain how one has to specify the accounts for this option to actually work (by primary email address it seems). Adding this to the test would also be nice.

modules/programs/thunderbird.nix
```diff
@@ -203,6 +203,23 @@ in {
...
+                example = ''
+                  [
+                    "account1"
+                    "account2"
+                    /* Other accounts in arbitrary order */
+                  ]
+                '';
```
alois31, line 220: This example is severely misleading since it suggests that the order is specified using the internal account ID. That said, I think it would be nice if something similar actually worked, since it would allow removing the special case for `account1` as well as enabling the user to use multiple imperative accounts.

modules/programs/thunderbird.nix
```diff
@@ -415,14 +432,24 @@ in {
           a.thunderbird.profiles == [ ]
           || any (p: p == name) a.thunderbird.profiles) enabledAccountsWithId;
 
+        orderedAccountsIds = let
+          accountsOrderIds =
+            map (a: builtins.hashString "sha256" a) profile.accountsOrder;
+          enabledAccountsIds = map (a: a.id) accounts;
+        in (lists.intersectLists enabledAccountsIds accountsOrderIds)
+        ++ (lists.subtractLists accountsOrderIds enabledAccountsIds);
+
```
alois31, line 441: This silently discards accounts in the order that are missing. Such accounts should either throw an error, or (to enable the use of multiple imperative accounts as noted above) be included verbatim.

alois31 (changes requested):

modules/programs/thunderbird.nix
nfelber: The accounts must be specified using their key  in `config.accounts.email.accounts`, as for `enabledAccounts`. I'll update the description to make it clearer.

modules/programs/thunderbird.nix
nfelber: I'm not sure what you mean here, although I agree that using the name `"account1"` is misleading. I'll change it. Can you elaborate a little?

modules/programs/thunderbird.nix
nfelber: I agree and I can do either depending on what is better. I would tend towards the second option.

modules/programs/thunderbird.nix
alois31: That's what I meant exactly, since `account1` currently doesn't work (it would work in addition to the email address of configured accounts if you choose to implement the second solution below).

nfelber:
I implemented the discussed changes, let me know if it is better now. About the tests, I'd like to add one for the account order but I'm unsure why there are 3 accounts in `user_pref("mail.accountmanager.accounts", "...")` in the file `tests/modules/programs/thunderbird/thunderbird-expected-first-linux.js` while only 2 are declared in `tests/modules/programs/thunderbird/thunderbird.nix` and I don't want to break the current tests.

Also, I had a bit of an issue when trying to rebase my commit on to of upstream/main and made a bit of a mess (should be solved now), so I apologize for any trouble.

tests/modules/programs/thunderbird/thunderbird-expected-second-linux.js
```diff
@@ -3,7 +3,7 @@
 user_pref("general.useragent.override", "");
 user_pref("mail.account.account_bcd3ace52bed41febb6cdc2fb1303aebaa573e0d993872da503950901bb6c6fc.identities", "id_bcd3ace52bed41febb6cdc2fb1303aebaa573e0d993872da503950901bb6c6fc");
 user_pref("mail.account.account_bcd3ace52bed41febb6cdc2fb1303aebaa573e0d993872da503950901bb6c6fc.server", "server_bcd3ace52bed41febb6cdc2fb1303aebaa573e0d993872da503950901bb6c6fc");
-user_pref("mail.accountmanager.accounts", "account_bcd3ace52bed41febb6cdc2fb1303aebaa573e0d993872da503950901bb6c6fc");
+user_pref("mail.accountmanager.accounts", "account_bcd3ace52bed41febb6cdc2fb1303aebaa573e0d993872da503950901bb6c6fc,account1");
```
khaneliman, line 6: can you make these changes to the darwin test files, too

alois31:
> I implemented the discussed changes, let me know if it is better now. About the tests, I'd like to add one for the account order but I'm unsure why there are 3 accounts in `user_pref("mail.accountmanager.accounts", "...")` in the file `tests/modules/programs/thunderbird/thunderbird-expected-first-linux.js` while only 2 are declared in `tests/modules/programs/thunderbird/thunderbird.nix` and I don't want to break the current tests.

The third account (c6cc42837ed0a8041f93ff12c579a4af0dbe702461c97eef069f9f5f8dc4bfab) is a feed account.

nfelber:
I added the `accountsOrder` option to the tests and mentioned feed accounts in the description of the option.

modules/programs/thunderbird.nix
```diff
@@ -277,6 +277,28 @@ in {
...
+              accountsOrder = mkOption {
+                type = types.listOf types.str;
+                default = [ ];
+                description = ''
+                  Custom ordering of accounts in Thunderbird's folder pane. The
+                  accounts must be specified using the name of their attribute
+                  in `config.accounts.email.accounts` (or
+                  `config.programs.thunderbird.profiles.<name>.feedAccounts` for feed
+                  accounts). Enabled accounts that aren't listed here appear in
+                  an arbitrary order after the ordered accounts.
```
alois31, line 289: … and local folders.

modules/programs/thunderbird.nix
```diff
@@ -573,13 +595,23 @@ in {
           (attrValues profile.feedAccounts);
 
         accounts = emailAccounts ++ feedAccounts;
+
+        orderedAccountsIds = let
+          accountsOrderIds =
+            map (a: builtins.hashString "sha256" a) profile.accountsOrder;
+          enabledAccountsIds = map (a: a.id) accounts;
+        in accountsOrderIds
```
alois31, line 603: If the intent of this is to allow the ordering of imperative accounts, this is both undocumented and does not work as intended, as imperative accounts do not use the SHA256 hash in their name.

alois31: If this is fixed it would also be a good idea to not include the `account1` again below when it is already in the list.

nfelber:
Thanks for the review, good catches. I think I fixed the issues. I updated the tests to account for imperative accounts and the local folders as well. I tried to make the documentation not too confusing, but let me know if it is still not very clear.

khaneliman:
Sorry, could you rebase. Looks like I can't push to your branch.

nfelber:
Rebased

PR #6320, opened 2025-01-15
yazi: improve fish integration

iSma:
### Description

Improve yazi fish integration:

Calls yazi as `command yazi`, allowing to use "yazi" as `shellWrapperName` without infinite recursion. This could be backward-incompatible for users who have an alias named "yazi" and expect this shell wrapper to call that alias instead of the yazi executable directly.

Also defines the wrapper with `programs.fish.functions` instead of `interactiveShellInit`, which seems to be the preferred way.

### Checklist

- [x] Change is backwards compatible. (with one exception, see above)

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@XYenon @eljamm

### DIFF
--- a/tests/modules/programs/yazi/fish-integration-enabled.nix
+++ b/tests/modules/programs/yazi/fish-integration-enabled.nix
@@ -1,27 +1,18 @@
-{ ... }:
+{ config, ... }:
 
-let
-  shellIntegration = ''
-    function yy
-      set tmp (mktemp -t "yazi-cwd.XXXXX")
-      yazi $argv --cwd-file="$tmp"
-      if set cwd (cat -- "$tmp"); and [ -n "$cwd" ]; and [ "$cwd" != "$PWD" ]
-        builtin cd -- "$cwd"
-      end
-      rm -f -- "$tmp"
-    end
-  '';
-in {
+{
   programs.fish.enable = true;
 
   programs.yazi = {
     enable = true;
+    shellWrapperName = "yy";
     enableFishIntegration = true;
   };
 
   test.stubs.yazi = { };
 
   nmt.script = ''
-    assertFileContains home-files/.config/fish/config.fish '${shellIntegration}'
+    assertFileContent home-files/.config/fish/functions/${config.programs.yazi.shellWrapperName}.fish \
+      ${./fish-integration-expected.fish}
   '';
 }
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -1960,6 +1960,19 @@ in {
           as well as wf-shell.
         '';
       }
+
+      {
+        time = "2025-01-21T17:28:13+00:00";
+        condition = with config.programs.yazi; enable && enableFishIntegration;
+        message = ''
+          Yazi's fish shell integration wrapper now calls the 'yazi' executable
+          directly, ignoring any shell aliases with the same name.
+
+          Your configuration may break if you rely on the wrapper calling a
+          'yazi' alias.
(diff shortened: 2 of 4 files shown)


XYenon (approved):

eljamm:
> This could be backward-incompatible for users who have an alias named "yazi" and expect this shell wrapper to call that alias instead of the yazi executable directly.

Could you explain this, perhaps? To me, it seems that we're doing the same thing: creating a function for the integration, but now we're storing it in `functions` instead of including it directly in the `config`.

tests/modules/programs/yazi/fish-integration-enabled.nix
```diff
@@ -1,27 +1,18 @@
...
     enable = true;
+    shellWrapperName = "yy";
     enableFishIntegration = true;
   };
 
   test.stubs.yazi = { };
 
   nmt.script = ''
-    assertFileContains home-files/.config/fish/config.fish '${shellIntegration}'
+    assertFileContent home-files/.config/fish/functions/yy.fish \
```
eljamm, line 15: I'd personally prefer if we don't hardcode the function name, here, since the test can break if it's not changed when the `shellWrapperName` does.

```suggestion
    assertFileContent home-files/.config/fish/functions/${config.programs.yazi.shellWrapperName}.fish \
```

iSma:
> > This could be backward-incompatible for users who have an alias named "yazi" and expect this shell wrapper to call that alias instead of the yazi executable directly.
> 
> Could you explain this, perhaps? To me, it seems that we're doing the same thing: creating a function for the integration, but now we're storing it in `functions` instead of including it directly in the `config`.

Yes, that part doesn't change; the end result is that we create a function named `$shellWrapperName`.

The difference I was pointing to is the way yazi is called from within this function: I've changed it to `command yazi` which calls executable directly and ignores potential shell aliases named "yazi". Previously, a user could potentially define `alias yazi=yazi --some-options` and the wrapper would call this alias.

I don't think the old behavior is particularly useful, since users can (and probably should) set their options directly in the settings. But I still wanted to point it out for the sake of completeness.

tests/modules/programs/yazi/fish-integration-enabled.nix
iSma: Good point, I've amended the commit with this change.

eljamm:
> The difference I was pointing to is the way yazi is called from within this function: I've changed it to command yazi which calls executable directly and ignores potential shell aliases named "yazi".

Ah, I see. I missed that, somehow.

Since this might be a breaking change, do you think it's better if we write something in the news file, like this perhaps:

https://github.com/nix-community/home-manager/blob/fc52a210b60f2f52c74eac41a8647c1573d2071d/modules/misc/news.nix#L1679-L1690

But just for fish integration users.

iSma:
> > The difference I was pointing to is the way yazi is called from within this function: I've changed it to command yazi which calls executable directly and ignores potential shell aliases named "yazi".
> 
> Ah, I see. I missed that, somehow.
> 
> Since this might be a breaking change, do you think it's better if we write something in [the news file](https://nix-community.github.io/home-manager/index.xhtml#sec-news), like this perhaps:
> 
> https://github.com/nix-community/home-manager/blob/fc52a210b60f2f52c74eac41a8647c1573d2071d/modules/misc/news.nix#L1679-L1690
> 
> But just for fish integration users.

Yes to be on the safe side I added a news entry.

rycee:

PR #6325, opened 2025-01-16
nh: fix auto clean

pedorich-n:
### Description

- Fix `PATH` environment variable
- Remove flake path assertion
- Add warning when HM's `nix.gc.automatic` is enabled 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@JohnRTitor

### DIFF
--- a/modules/programs/nh.nix
+++ b/modules/programs/nh.nix
@@ -52,15 +52,11 @@ in {
   };
 
   config = {
-    warnings = lib.optionals (!(cfg.clean.enable -> !osConfig.nix.gc.automatic))
-      [
-        "programs.nh.clean.enable and nix.gc.automatic (system-wide in configuration.nix) are both enabled. Please use one or the other to avoid conflict."
-      ];
-
-    assertions = [{
-      assertion = (cfg.flake != null) -> !(lib.hasSuffix ".nix" cfg.flake);
-      message = "nh.flake must be a directory, not a nix file";
-    }];
+    warnings =
+      (lib.optional (cfg.clean.enable && (osConfig.nix.gc.automatic or false))
+        "programs.nh.clean.enable and nix.gc.automatic (system-wide in configuration.nix) are both enabled. Please use one or the other to avoid conflict.")
+      ++ (lib.optional (cfg.clean.enable && config.nix.gc.automatic)
+        "programs.nh.clean.enable and nix.gc.automatic (Home-Manager) are both enabled. Please use one or the other to avoid conflict.");
 
     home = lib.mkIf cfg.enable {
       packages = [ cfg.package ];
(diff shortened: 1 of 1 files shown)


modules/programs/nh.nix
```diff
@@ -52,15 +52,11 @@ in {
...
   config = {
-    warnings = lib.optionals (!(cfg.clean.enable -> !osConfig.nix.gc.automatic))
-      [
-        "programs.nh.clean.enable and nix.gc.automatic (system-wide in configuration.nix) are both enabled. Please use one or the other to avoid conflict."
-      ];
-
-    assertions = [{
-      assertion = (cfg.flake != null) -> !(lib.hasSuffix ".nix" cfg.flake);
-      message = "nh.flake must be a directory, not a nix file";
-    }];
```
pedorich-n, line 63: I have my HM configuration in a folder called `.config.nix`. So by this assertion, it can't be used. 
IMO home-manager shouldn't try to guess what paths it doesn't have access to in build time are.

modules/programs/nh.nix
```diff
@@ -52,15 +52,11 @@ in {
   };
 
   config = {
-    warnings = lib.optionals (!(cfg.clean.enable -> !osConfig.nix.gc.automatic))
```
pedorich-n, line 55: On standalone install `osConfig` is [null](https://github.com/nix-community/home-manager/blob/0dfec9deb275854a56c97c356c40ef72e3a2e632/modules/misc/submodule-support.nix#L42), so this fails to evaluate

modules/programs/nh.nix
```diff
@@ -52,15 +52,11 @@ in {
...
-      [
-        "programs.nh.clean.enable and nix.gc.automatic (system-wide in configuration.nix) are both enabled. Please use one or the other to avoid conflict."
-      ];
-
-    assertions = [{
-      assertion = (cfg.flake != null) -> !(lib.hasSuffix ".nix" cfg.flake);
-      message = "nh.flake must be a directory, not a nix file";
-    }];
+    warnings =
+      (lib.optional (cfg.clean.enable && (osConfig.nix.gc.automatic or false))
```
pedorich-n, line 56: I might be missing something, but we want to raise a warning if both of these values are `true`, right? Then I think `&&` is easier to read and reason about than `!(cfg.clean.enable -> !osConfig.nix.gc.automatic)`.
Please correct me if there are other cases.

pedorich-n:
@JohnRTitor please have a look 🙇

JohnRTitor:
Changes mostly look good. I am busy this week due to work, I'll try to look at them in the weekend or next week.

pedorich-n:
Looks like @tetov beat me to it with https://github.com/nix-community/home-manager/pull/6350
😅

Weathercold:
Don't close the pr, it has fixes that #6350 doesn't have ;)

tetov:
> Don't close the pr, it has fixes that #6350 doesn't have ;)

What are you missing? :)

Weathercold:
This pr also removes the flake path assertion and adds warning for `nix.gc.automatic`

tetov:
Ah, maybe I only did that in my private config, sorry! I can make a new PR later tonight or tomorrow.

PR #6330, opened 2025-01-17
zed-editor: add installRemoteServer option

GaetanLepage:
### Description

Building on top of https://github.com/NixOS/nixpkgs/pull/370017.
This option automatically symlinks the Zed remote server binary to the right location in the user's home directory.
It allows to connect to this host from a remote Zed instance.

See https://wiki.nixos.org/wiki/Zed#Remote_Server for more details.

cc @niklaskorz

**Note:** Requires https://github.com/NixOS/nixpkgs/pull/370017 to land on the unstable channels.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

cc @libewa

### DIFF
--- a/modules/programs/zed-editor.nix
+++ b/modules/programs/zed-editor.nix
@@ -80,6 +80,21 @@ in {
           Use the name of a repository in the [extension list](https://github.com/zed-industries/extensions/tree/main/extensions).
         '';
       };
+
+      installRemoteServer = mkOption {
+        type = types.bool;
+        default = false;
+        example = true;
+        description = ''
+          Whether to symlink the Zed's remote server binary to the expected
+          location. This allows remotely connecting to this system from a
+          distant Zed client.
+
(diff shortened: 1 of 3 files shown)


libewa:
Looks like the tests didn't run using the latest nixpkgs.

niklaskorz (approved):
Tested and works fine for providing the remote server binary on a NixOS server

teto (commented):
could you test the executable is created at the correct location.
Does it need some special firewall rules ? maybe link towards them directly . Also I would disable it by default. Not everyone wants its machine to be accessible remotely.

I dont think we can setup them from HM alone and a correct approach is out of scope.
Thinking out loud:  maybe HM modules could generate iptables scripts for the user to load.

rycee:
Thanks for the contribution! I made some minor changes and added a basic test. Merged now 🙂

Issue #6329, opened 2025-01-17
bug: `services.kdeconnect.indicator` on wayland fails over missing `tray.target`


KiaraGrouwstra:
### Are you following the right branch?

- [x] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Issue description

since #5299, I've had `services.kdeconnect.indicator` fail stating `Unit tray.target not found`.

it seems `tray.target` [is available using `xsession` tho not by default on Wayland](https://github.com/nix-community/home-manager/issues/2064#issuecomment-853448012).
we may want to instead make this dependency conditional on use of `xsession`.

### Maintainer CC

@onemoresuza

### System information

```markdown
- system: `"x86_64-linux"`
- host os: `Linux 6.12.2-zen1, NixOS, 24.11 (Vicuna), 24.11.20250102.c1df670`
- multi-user?: `yes`
- sandbox: `yes`
- version: `nix-env (Nix) 2.24.11`
- nixpkgs: `/run/current-system/nixpkgs`
```

ada4a:
I think this is a duplicate of #2064?

ada4a:
Oh, no, sorry, this is partially new after all I think. I've been having the issue described in #2064 as well, and therefore had `services.kdeconnect` disabled. But it turns out that I've also had `services.kdeconnect.indicator` enabled all this time, and now _that_ broke with the error you described. So I think it might make sense to keep both of the issues opened.

ada4a:
For #2064, I've also been using [this](https://github.com/nix-community/home-manager/issues/2064#issuecomment-887300055) workaround. So here's a version of it that's adapted to `kdeconnect.indicator` as well:
```nix
# with cfg = config.services.kdeconnect;
systemd.user.targets.tray = lib.mkIf (cfg.enable || cfg.indicator) {
  Unit = {
    Description = "Home Manager System Tray";
    Requires = [ "graphical-session-pre.target" ];
  };
};
```

onemoresuza:
The tray.target problem goes beyond the kdeconnect service, as it is mentioned on #2064: as it was implemented, the target is only created when `services.xsession.enable = true;`.

As a workaround for wayland users, @ada4a [solution](https://github.com/nix-community/home-manager/issues/6329#issuecomment-2598736548) may be used unconditionally, *i. e.*, without the `lib.mkIf` call. That would fix any other service that depends on `tray.target`.

[...]

Issue #2064, opened 2021-06-02
Unit tray.target not found


jlesquembre:
### Issue description

On current master, I get the error `Unit tray.target not found` when I try to start some systemd units (like udiskie or blueman).
I'm using waybar, I think this error is related to #2027 

#### Maintainer CC
@berbiche @NickHu

berbiche:
Unfortunately, the tray.target is only created if xsession management is enabled in HM.

The creation of said systemd target should be moved elsewhere.

In the meantime, you can create the systemd target manually from the xsession.nix? file and add `Install.WantedBy = [ "graphical-session-pre.target" ]` to the target definition.

sumnerevans:
> Unfortunately, the tray.target is only created if xsession management is enabled in HM.

Ah, this explains why I encountered that error on my wayland system. We should definitely fix that.

zeorin:
I have this problem on X11. `flameshot` includes a service that wants the `tray.target`.

vyorkin:
Can confirm the problem on X11 with `flameshot`, subscribed to this issue exactly for this reason

zeorin:
@vyorkin I should probably mention that I worked around this for the time being with the following in my `home.nix`:
```nix
	systemd.user.targets.tray = {
		Unit = {
			Description = "Home Manager System Tray";
			Requires = [ "graphical-session-pre.target" ];
		};
	};
```
which I lifted from `home-manager`'s source somewhere.

DavHau:
Still an issue. The home-manager udiskie service doesn't work on my nixos 21.05.
I'm using i3wm on x11 and there is no tray target

aru-hackZ:
What I noticed is that I can't start any service in `$XDG_CONFIG_HOME/systemd/user`, so services like the ones created from `services.emacs` or `services.gpg-agent` can't be started. Searching a bit I didn't find any issue directly talking about it. `systemd-analyze --user unit-paths` shows `$XDG_CONFIG_HOME/systemd/user`, which makes the fact that it can't find services there even more strange.

AshleyYakeley:
I believe this should be reopened? I'm using marco (rather than sway) without xsession management, and I'm coming across this issue.

ncfavier:
I can see two solutions:
- change `Requires` directives on `tray.target` to `Wants` instead; @NickHu is there a specific reason you used `Requires`?
- define the `tray` target in the systemd module, regardless of whether xsession or sway is enabled

[...]

PR #6332, opened 2025-01-17
wayland: create tray.target if xsession is not enabled

onemoresuza:
### Description

As it current stands, the tray target is only created if xsession is
enabled. This causes services that depend on it to fail on wayland
sessions on configurations that do not have `services.xsession.enable =
true;` or the workaround as the following:

```nix
{config, lib, ...}: {
  systemd.user.targets.tray = lib.mkIf (!config.xsession.enable) {
    Unit = {
      Description = "Home Manager System Tray";
      Requires = [ "graphical-session-pre.target" ];
    };
  };
}
```

This PR addresses [#6329](https://github.com/nix-community/home-manager/issues/6329) and
[#2064](https://github.com/nix-community/home-manager/issues/2064),
albeit I do not think it is the greatest solution, since it allows the
targets defined by the xsession and wayland modules to be kept out of
sync.

I'd recommend defining `tray.target` independently from either one of
those modules, since virtually every graphical session that requires a
system tray would need such target.

### Checklist

- [ ] Change is backwards compatible. May break config of non xsession enabled
users that do not use `mkForce` for the tray target definition.

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/xsession.nix
+++ b/modules/xsession.nix
@@ -13,6 +13,20 @@ in {
     xsession = {
       enable = mkEnableOption "X Session";
 
+      trayTarget = mkOption {
+        readOnly = true;
+        internal = true;
+        visible = false;
+        description = "Common tray.target for both xsession and wayland";
+        type = types.attrs;
+        default = {
+          Unit = {
+            Description = "Home Manager System Tray";
+            Requires = [ "graphical-session-pre.target" ];
+          };
+        };
+      };
+
       scriptPath = mkOption {
         type = types.str;
         default = ".xsession";
@@ -163,12 +177,7 @@ in {
           };
         };
 
-        tray = {
-          Unit = {
-            Description = "Home Manager System Tray";
-            Requires = [ "graphical-session-pre.target" ];
-          };
-        };
+        tray = cfg.trayTarget;
       };
     };
 
--- a/modules/services/window-managers/hyprland.nix
+++ b/modules/services/window-managers/hyprland.nix
@@ -277,12 +277,5 @@ in {
           [ "xdg-desktop-autostart.target" ];
       };
     };
-
-    systemd.user.targets.tray = {
-      Unit = {
-        Description = "Home Manager System Tray";
-        Requires = [ "graphical-session-pre.target" ];
-      };
-    };
   };
 }
--- a/modules/services/window-managers/i3-sway/sway.nix
+++ b/modules/services/window-managers/i3-sway/sway.nix
@@ -553,13 +553,6 @@ in {
             optional cfg.systemd.xdgAutostart "xdg-desktop-autostart.target";
         };
       };
-
-      systemd.user.targets.tray = {
-        Unit = {
-          Description = "Home Manager System Tray";
-          Requires = [ "graphical-session-pre.target" ];
-        };
-      };
     }
   ]);
 }
--- a/modules/services/window-managers/river.nix
+++ b/modules/services/window-managers/river.nix
@@ -201,12 +201,5 @@ in {
         After = [ "graphical-session-pre.target" ];
       };
     };
-
-    systemd.user.targets.tray = {
-      Unit = {
-        Description = "Home Manager System Tray";
-        Requires = [ "graphical-session-pre.target" ];
-      };
-    };
   };
 }
--- a/modules/services/window-managers/wayfire.nix
+++ b/modules/services/window-managers/wayfire.nix
@@ -189,12 +189,5 @@
         After = [ "graphical-session-pre.target" ];
       };
     };
-
-    systemd.user.targets.tray = {
-      Unit = {
-        Description = "Home Manager System Tray";
-        Requires = [ "graphical-session-pre.target" ];
-      };
-    };
   };
 }
--- a/modules/wayland.nix
+++ b/modules/wayland.nix
@@ -1,4 +1,4 @@
-{ lib, ... }:
+{ config, lib, ... }:
 
 {
   meta.maintainers = [ lib.maintainers.thiagokokada ];
@@ -20,5 +20,9 @@
         '';
       };
     };
+  };
+
+  config = lib.mkIf (!config.xsession.enable) {
+    systemd.user.targets.tray = config.xsession.trayTarget;
   };
 }


KiaraGrouwstra (approved):

teto:
the target looks defined in several modules like modules/services/window-managers/wayfire.nix.  / river /hyperland/sway. shouldn't the tray.target be removed from there as well ? 

> I'd recommend defining tray.target independently from either one of
those modules, since virtually every graphical session that requires a
system tray would need such target.

fine by me. We could also merge the current approach as a first step.

onemoresuza:
@teto, you're right. I've added some commits removing them to be squashed if this is to be merged.

Regarding setting `tray.target` independently from either the xsession or a window manager, `systemd.user.targets.tray` could be a read-only/internal option from either xsession or wayland so that it would not, as my current implementation, allow the tray targets to get out of sync, since there'd be only one.

onemoresuza:
Created a `trayTarget` option (interanl/read-only) under `xsession`, which is not only used by it, but also by wayland's config when the former is not enabled.

onemoresuza:
Maintainer CC for xsession and wayland:

@thiagokokada @rycee

teto (approved):

modules/xsession.nix
```diff
@@ -13,6 +13,20 @@ in {
...
+        readOnly = true;
+        internal = true;
+        visible = false;
+        description = "Common tray.target for both xsession and wayland";
+        type = types.attrs;
+        default = {
+          Unit = {
+            Description = "Home Manager System Tray";
+            Requires = [ "graphical-session-pre.target" ];
+          };
```
thiagokokada, line 26: Why is this a `Unit` instead of being a simple string (or a list of strings if we want to set multiple targets for some reason)?

onemoresuza: It is how it is currently implemented in the [xsession module](https://github.com/nix-community/home-manager/blob/1b4f2a48168b3d90e11365552d1e7e601a4be6b6/modules/xsession.nix#L166).

thiagokokada: This is the implementation that is fine, but the option shouldn't implement a whole `Unit` unless there is a good reason to do so.

onemoresuza:
>
> > +          Unit = {
> +            Description = "Home Manager System Tray";
> +            Requires = [ "graphical-session-pre.target" ];
> +          };
>
> This is the implementation that is fine, but the option shouldn't implement a whole `Unit` unless there is a good reason to do so.
>

I've set it just like the implementation because, as of now, it's a
readOnly/internal option, so the user would not be allowed to change it.

Do you see a use case in which the user would have to modify
`systemd.user.target.tray`?

thiagokokada:
> > + Unit = { + Description = "Home Manager System Tray"; + Requires = [ "graphical-session-pre.target" ]; + }; This is the implementation that is fine, but the option shouldn't implement a whole `Unit` unless there is a good reason to do so.
> I've set it just like the implementation because, as of now, it's a readOnly/internal option, so the user would not be allowed to change it. Do you see a use case in which the user would have to modify `systemd.user.target.tray`?

Oh ok, true, didn't realise that this was set to `internal`/`readOnly`.

I can find a few cases where this would be a good idea to change it, for example I know that `grapgical-session-pre.target` is a bad choice for Hyprland in a few other cases at least (not sure about the tray though).

I would recommend at least making the actual target an option just to make sure that we can cover other use cases, but the current implementation looks fine and we can add an option later too if you want.

onemoresuza:
> I can find a few cases where this would be a good idea to change it, for example I know that `grapgical-session-pre.target` is a bad choice for Hyprland in a few other cases at least (not sure about the tray though).
> 
> I would recommend at least making the actual target an option just to make sure that we can cover other use cases, but the current implementation looks fine and we can add an option later too if you want.

I think we could create an option under the wayland module for enabling the systemd tray target, while leaving as is for the xsession one, since, on the latter, the tray target is always created.

I saw that the wayland module has a `systemd.target` option already. For a more smooth transition, I'd suggest something like the following:

```nix
# wayland module

options = {
  wayland.systemd.enableTrayTarget = mkEnableOption "tray.target" // { default = true; };
};

config = mkIf config.wayland.systemd.enableTrayTarget {
  systemd.user.targets.tray = config.xsession.trayTarget;
};
```

That would allow modules, for instance, hyprland's, to not have a tray.target:

```nix
# hyprland module

config = mkIf cfg.enable {
  assertions = [
    {
      assertion = !config.wayland.systemd.enableTrayTarget;
      message = ''
        hyprland cannot work correctly when `wayland.systemd.enableTrayTarget` is enabled.
      '';
    }
  ];

  wayland.systemd.enableTrayTarget = false;
};
```

What do you think?

thiagokokada:
> What do you think?

I think the main issue is that the tray target is hardcoded to `graphical-session-pre.target` that is known to cause some issues in Hyprland and probably other Wayland DMs too, not that we need to enable/disable this target. `graphical-session.target` is more reliable in general since it ensures that everything in the graphical session is initialised (e.g.: `WAYLAND_DISPLAY`).

In my opinion we should:
- Either change the target to `graphical-session.target` (this would change the behavior but should work fine)
- Or allow the target to be configurable

onemoresuza:
It can be made configurable by removing the readOnly/internal attributes; however, as of now, wayland DM modules would have to access an xsession attribute to change their `tray.target`.

This goes back to what I've mention on the description of this PR:

> I'd recommend defining tray.target independently from either one of
those modules, since virtually every graphical session that requires a
system tray would need such target.

We could either create a `graphical-session` module, defining the `trayTarget` on it, or move it to another module like systemd's.

thiagokokada (approved):

onemoresuza:
> It can be made configurable by removing the readOnly/internal attributes; however, as of now, wayland DM modules would have to access an xsession attribute to change their `tray.target`.
> 
> This goes back to what I've mention on the description of this PR:
> 
> > I'd recommend defining tray.target independently from either one of
> > those modules, since virtually every graphical session that requires a
> > system tray would need such target.
> 
> We could either create a `graphical-session` module, defining the `trayTarget` on it, or move it to another module like systemd's.

@thiagokokada, are you approving the current state of the PR or the proposed changes mentioned above?

thiagokokada:
> @thiagokokada, are you approving the current state of the PR or the proposed changes mentioned above?

The current state, it doesn't make sense to approve for something that doesn't exist.

I think we can merge it in the current state and change it to accommodate customization in the future.

onemoresuza:
Got it.

PR #6343, opened 2025-01-20
feat(yazi): Add assertion to prevent conflicting keybinds

lordkekz:
### Description
Right now, you can accidentally define conflicting keybinds and you will not get any warning.
This PR adds an assertion which checks if there are any two keybinds A and B where the `on` value (sequence of keypresses) of A starts with the `on` value of B.

**This is not yet finished but I will pick it up when I have time again.**

The new assertions produce error messages like this:
```
There are conflicting keybinds in: `programs.yazi.keymap.input`
┢ Conflict between:
┃ {"on":["<Esc>"],"run":"escape"}
┃ {"on":["<Esc>"],"run":"backspace"}


--
There are conflicting keybinds in: `programs.yazi.keymap.manager`
┢ Conflict between:
┃ {"on":["c","d","e"],"run":"something"}
┃ {"on":["c","d"],"run":"somethingelse"}
│
┢ Conflict between:
┃ {"on":["q"],"run":"escape"}
┃ {"on":["q"],"run":"quit"}
```

### Caveats
- For now, it only checks for conflicts of keybinds which are defined in the home-manager config. It does _not_ check for conflicts with yazi's default keymap.
  - The latter is theoretically possible e.g. by parsing the default keymap.toml (would require IFD) or by maintining a Nix-language version of it.
  - If checks with default bindings get added, it would be important to indicate them as such (and ignore overriding the defaults).
- I would be nice to be able to print the definition location of each keybind, i.e. the home-manager module that added it to the list, but I'm not sure how to do that easily.

### Checklist

#### Maintainer CC
@XYenon @eljamm 
btw would you be interested in me becoming a maintainer of the yazi module?
This is my second contribution to this module and I am also maintaining the [`nix-yazi-plugins`](https://github.com/lordkekz/nix-yazi-plugins) repo (not afiliated with yazi or home-manager but relevant IMO), so I plan to keep contributing here in the forseeable future.

### DIFF
--- a/modules/programs/yazi.nix
+++ b/modules/programs/yazi.nix
@@ -275,7 +275,12 @@ in {
             '';
           }) cfg.${opt};
 
-        allCustomKeybinds = pipe cfg.keymap [ (collect isList) (concatLists)];
+        mkAssertNoConflictedKeybinds = keymapSection: let
+        allCustomKeybinds = pipe keymapSection [
+          (collect isList)
+          (concatLists)
+          (map (x: setAttr x "remainingSequence" x.on))
+        ];
         seqKeysStructured = remainingKeybinds: with lib; let
           currentKeys = pipe remainingKeybinds [
             (catAttrs "remainingSequence")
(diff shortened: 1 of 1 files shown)


teto:
I dont think it's the role of home-manager or any nix code to reimplement configuration validators. It's a sysiphean task.It's much simpler to run an activation script that launches yazi to validate the configuration.
If it's important to you, do you know you can extend home-manager modules in your own configuration ? If a simple change can help you plug your validation logic, we can merge it but not sure that's even necessary

lordkekz:
Alright, that's fair. I'll close this then.

PR #6350, opened 2025-01-22
nh: fix conditional for warning (when osConfig is null) and remove potential null variable config.nix.package

tetov:
### Description

`nh.clean.enable = true;` on my home-manager standalone installation gave me two evaluation errors.

First because osConfig is `null` which breaks the warning (for when GC is enabled both in nh and in NixOS configuration).

Then because `config.nix.package` can't be added to the nh-clean systemd service if evaluates to `null` (which is the default value).

After testing I found that nix doesn't need to be on path for nh to run so I removed the addition to path. 

I actually found a few problems with the systemd service. 

* `exec` is not found when I run it (bash built-in)
* I don't think `$PATH` would evaluate in a unit file like this

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/programs/nh.nix
+++ b/modules/programs/nh.nix
@@ -74,8 +74,7 @@ in {
         Service = {
           Type = "oneshot";
           ExecStart =
-            "exec ${lib.getExe cfg.package} clean user ${cfg.clean.extraArgs}";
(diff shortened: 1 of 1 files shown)


rycee:

PR #6356, opened 2025-01-24
linux-wallpaperengine: add module

ckgxrg-salt:
### Description
Adds the `linux-wallpaperengine` systemd service.
The original project is here: https://github.com/Almamu/linux-wallpaperengine.
It simply converts options into command-line arguments and provides a systemd service to run `linux-wallpaperengine` in the background.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

### DIFF
--- a/modules/services/linux-wallpaperengine.nix
+++ b/modules/services/linux-wallpaperengine.nix
@@ -1,0 +1,121 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+
+  cfg = config.services.linux-wallpaperengine;
+
+in {
+  meta.maintainers = [ hm.maintainers.ckgxrg ];
+
+  options.services.linux-wallpaperengine = {
+    enable = mkEnableOption
+      "linux-wallpaperengine, an implementation of Wallpaper Engine functionality";
+
+    package = mkPackageOption pkgs "linux-wallpaperengine" { };
+
+    assetsPath = mkOption {
+      type = types.path;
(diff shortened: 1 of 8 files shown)


ckgxrg-salt:
Oh so I have to use hm.maintainers instead of maintainers

modules/services/linux-wallpaperengine.nix
```diff
@@ -0,0 +1,121 @@
...
+              description = "Mutes all sound of the wallpaper.";
+            };
+
+            automute = mkOption {
+              type = types.bool;
+              default = true;
+              description = "Automute when another app is playing sound.";
+            };
+
+            audio-processing = mkOption {
```
rycee, line 75: Perhaps can skip the "audio-" part since it's already inside the audio attribute set?

```suggestion
            processing = mkOption {
```

ckgxrg-salt: I did this according to the original project's cmdline `--no-audio-processing`.
Anyways it looks good to me, I'll change it

rycee:

ccalhoun1999:
How are we supposed to set assetsPath here?
```  services.linux-wallpaperengine.assetsPath = "${config.home.homeDirectory}/.steam/steam/steamapps/common/wallpaper_engine/assets/";
```
causes `error: cannot coerce null to a string: null`
The package's readme also says it should auto-detect common installation locations, but this module forces us to set a value.

PR #6358, opened 2025-01-25
treewide: standardize shell integration options

trueNAHO:
### Description

```
Standardize all 'programs.<PROGRAM>.enable<SHELL>Integration' options
with the following new functions:

- lib.hm.shell.mkBashIntegrationOption
- lib.hm.shell.mkFishIntegrationOption
- lib.hm.shell.mkIonIntegrationOption
- lib.hm.shell.mkNushellIntegrationOption
- lib.hm.shell.mkZshIntegrationOption

These functions should default to their corresponding global option:

- shell.enableBashIntegration
- shell.enableFishIntegration
- shell.enableIonIntegration
- shell.enableNushellIntegration
- shell.enableZshIntegration

All these global options default to the 'shell.enableShellIntegration'
value.

This hierarchy standardizes the shell integration and increases end-user
flexibility.

BREAKING CHANGE: This modifies the following inconsistent default values
from 'false' to 'true':

- programs.zellij.enableBashIntegration
- programs.zellij.enableFishIntegration
- programs.zellij.enableZshIntegration
```

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

### DIFF
--- a/modules/lib/shell.nix
+++ b/modules/lib/shell.nix
@@ -1,6 +1,22 @@
 { lib }:
 
-rec {
+let
+
+  mkShellIntegrationOption = name:
+    { config, baseName ? name, extraDescription ? "" }:
+    let attrName = "enable${baseName}Integration";
+    in lib.mkOption {
+      default = config.home.shell.${attrName};
+      defaultText = lib.literalMD "[](#opt-home.shell.${attrName})";
+      example = false;
+      description = "Whether to enable ${name} integration.${
+          lib.optionalString (extraDescription != "")
+          ("\n\n" + extraDescription)
+        }";
+      type = lib.types.bool;
+    };
+
+in rec {
   # Produces a Bourne shell like variable export statement.
   export = n: v: ''export ${n}="${toString v}"'';
 
@@ -8,4 +24,10 @@ rec {
   # assignment, this function produces a string containing an export
   # statement for each set entry.
   exportAll = vars: lib.concatStringsSep "\n" (lib.mapAttrsToList export vars);
+
+  mkBashIntegrationOption = mkShellIntegrationOption "Bash";
+  mkFishIntegrationOption = mkShellIntegrationOption "Fish";
+  mkIonIntegrationOption = mkShellIntegrationOption "Ion";
+  mkNushellIntegrationOption = mkShellIntegrationOption "Nushell";
+  mkZshIntegrationOption = mkShellIntegrationOption "Zsh";
 }
--- a/modules/misc/shell.nix
+++ b/modules/misc/shell.nix
@@ -1,0 +1,43 @@
+{ config, lib, ... }:
+
+{
+  options.home.shell = {
+    enableShellIntegration = lib.mkOption {
+      type = lib.types.bool;
+      default = true;
+      example = false;
+      description = ''
+        Whether to globally enable shell integration for all supported shells.
+
+        Individual shell integrations can be overridden with their respective
+        `shell.enable<SHELL>Integration` option. For example, the following
+        declaration globally disables shell integration for Bash:
+
+        ```nix
+        home.shell.enableBashIntegration = false;
+        ```
+      '';
+    };
+
+    enableBashIntegration = lib.hm.shell.mkBashIntegrationOption {
+      inherit config;
+      baseName = "Shell";
+    };
+    enableFishIntegration = lib.hm.shell.mkFishIntegrationOption {
+      inherit config;
+      baseName = "Shell";
+    };
+    enableIonIntegration = lib.hm.shell.mkIonIntegrationOption {
+      inherit config;
+      baseName = "Shell";
+    };
+    enableNushellIntegration = lib.hm.shell.mkNushellIntegrationOption {
+      inherit config;
+      baseName = "Shell";
+    };
+    enableZshIntegration = lib.hm.shell.mkZshIntegrationOption {
+      inherit config;
+      baseName = "Shell";
+    };
+  };
+}
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -2018,6 +2018,22 @@ in {
           systems.
         '';
       }
+
+      {
+        time = "2025-02-07T22:31:45+00:00";
+        message = ''
+          All 'programs.<PROGRAM>.enable<SHELL>Integration' values now default
+          to the new 'home.shell.enable<SHELL>Integration' options, which
+          inherit from the new the 'home.shell.enableShellIntegration' option.
+
+          The following inconsistent default values change from 'false' to
+          'true':
+
+          - programs.zellij.enableBashIntegration
+          - programs.zellij.enableFishIntegration
+          - programs.zellij.enableZshIntegration
+        '';
+      }
     ];
   };
 }
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -36,6 +36,7 @@ let
     ./misc/pam.nix
     ./misc/qt.nix
     ./misc/qt/kconfig.nix
+    ./misc/shell.nix
     ./misc/specialisation.nix
     ./misc/submodule-support.nix
     ./misc/tmpfiles.nix
--- a/modules/programs/atuin.nix
+++ b/modules/programs/atuin.nix
@@ -23,33 +23,26 @@ in {
       description = "The package to use for atuin.";
     };
 
-    enableBashIntegration = mkOption {
-      type = types.bool;
-      default = true;
-      description = ''
-        Whether to enable Atuin's Bash integration. This will bind
-        `ctrl-r` to open the Atuin history.
-      '';
+    enableBashIntegration = lib.hm.shell.mkBashIntegrationOption {
+      inherit config;
+      extraDescription =
+        "If enabled, this will bind `ctrl-r` to open the Atuin history.";
     };
 
-    enableZshIntegration = mkOption {
-      type = types.bool;
-      default = true;
-      description = ''
-        Whether to enable Atuin's Zsh integration.
-
-        If enabled, this will bind `ctrl-r` and the up-arrow
-        key to open the Atuin history.
-      '';
+    enableFishIntegration = lib.hm.shell.mkFishIntegrationOption {
+      inherit config;
+      extraDescription =
+        "If enabled, this will bind the up-arrow key to open the Atuin history.";
     };
 
-    enableFishIntegration = mkOption {
-      default = true;
-      type = types.bool;
-      description = ''
-        Whether to enable Atuin's Fish integration.
+    enableNushellIntegration =
+      lib.hm.shell.mkNushellIntegrationOption { inherit config; };
 
-        If enabled, this will bind the up-arrow key to open the Atuin history.
+    enableZshIntegration = lib.hm.shell.mkZshIntegrationOption {
+      inherit config;
+      extraDescription = ''
+        If enabled, this will bind `ctrl-r` and the up-arrow key to open the
+        Atuin history.
       '';
     };
 
@@ -86,14 +79,6 @@ in {
 
         See <https://atuin.sh/docs/config/> for the full list
         of options.
-      '';
-    };
-
-    enableNushellIntegration = mkOption {
-      default = true;
-      type = types.bool;
-      description = ''
-        Whether to enable Nushell integration.
       '';
     };
 
--- a/modules/programs/autojump.nix
+++ b/modules/programs/autojump.nix
@@ -13,29 +13,14 @@ in {
   options.programs.autojump = {
     enable = mkEnableOption "autojump";
 
-    enableBashIntegration = mkOption {
-      default = true;
-      type = types.bool;
-      description = ''
-        Whether to enable Bash integration.
-      '';
-    };
+    enableBashIntegration =
+      lib.hm.shell.mkBashIntegrationOption { inherit config; };
 
-    enableZshIntegration = mkOption {
-      default = true;
-      type = types.bool;
-      description = ''
-        Whether to enable Zsh integration.
-      '';
-    };
+    enableFishIntegration =
+      lib.hm.shell.mkFishIntegrationOption { inherit config; };
 
-    enableFishIntegration = mkOption {
-      default = true;
-      type = types.bool;
-      description = ''
-        Whether to enable Fish integration.
-      '';
-    };
+    enableZshIntegration =
+      lib.hm.shell.mkZshIntegrationOption { inherit config; };
   };
 
   config = mkIf cfg.enable {
--- a/modules/programs/broot.nix
+++ b/modules/programs/broot.nix
@@ -154,37 +154,17 @@ in {
   options.programs.broot = {
     enable = mkEnableOption "Broot, a better way to navigate directories";
 
-    enableBashIntegration = mkOption {
-      default = true;
-      type = types.bool;
-      description = ''
-        Whether to enable Bash integration.
-      '';
-    };
+    enableBashIntegration =
+      lib.hm.shell.mkBashIntegrationOption { inherit config; };
 
-    enableZshIntegration = mkOption {
-      default = true;
-      type = types.bool;
-      description = ''
-        Whether to enable Zsh integration.
-      '';
-    };
+    enableFishIntegration =
+      lib.hm.shell.mkFishIntegrationOption { inherit config; };
 
-    enableFishIntegration = mkOption {
-      default = true;
-      type = types.bool;
-      description = ''
-        Whether to enable Fish integration.
-      '';
-    };
+    enableNushellIntegration =
+      lib.hm.shell.mkNushellIntegrationOption { inherit config; };
 
-    enableNushellIntegration = mkOption {
-      default = true;
-      type = types.bool;
-      description = ''
-        Whether to enable Nushell integration.
-      '';
-    };
+    enableZshIntegration =
+      lib.hm.shell.mkZshIntegrationOption { inherit config; };
 
     package = mkOption {
       type = types.package;
--- a/modules/programs/carapace.nix
+++ b/modules/programs/carapace.nix
@@ -16,21 +16,17 @@ in {
 
     package = mkPackageOption pkgs "carapace" { };
 
-    enableBashIntegration = mkEnableOption "Bash integration" // {
-      default = true;
-    };
+    enableBashIntegration =
+      lib.hm.shell.mkBashIntegrationOption { inherit config; };
 
-    enableZshIntegration = mkEnableOption "Zsh integration" // {
-      default = true;
-    };
+    enableFishIntegration =
+      lib.hm.shell.mkFishIntegrationOption { inherit config; };
 
-    enableFishIntegration = mkEnableOption "Fish integration" // {
-      default = true;
-    };
+    enableNushellIntegration =
+      lib.hm.shell.mkNushellIntegrationOption { inherit config; };
 
-    enableNushellIntegration = mkEnableOption "Nushell integration" // {
-      default = true;
-    };
+    enableZshIntegration =
+      lib.hm.shell.mkZshIntegrationOption { inherit config; };
   };
 
   config = mkIf cfg.enable {
--- a/modules/programs/dircolors.nix
+++ b/modules/programs/dircolors.nix
@@ -19,29 +19,14 @@ in {
       '';
     };
 
-    enableBashIntegration = mkOption {
-      type = types.bool;
-      default = true;
-      description = ''
-        Whether to enable Bash integration.
-      '';
-    };
+    enableBashIntegration =
+      lib.hm.shell.mkBashIntegrationOption { inherit config; };
 
-    enableFishIntegration = mkOption {
-      type = types.bool;
-      default = true;
-      description = ''
-        Whether to enable Fish integration.
-      '';
-    };
+    enableFishIntegration =
+      lib.hm.shell.mkFishIntegrationOption { inherit config; };
 
-    enableZshIntegration = mkOption {
-      type = types.bool;
-      default = true;
-      description = ''
-        Whether to enable Zsh integration.
-      '';
-    };
+    enableZshIntegration =
+      lib.hm.shell.mkZshIntegrationOption { inherit config; };
 
     settings = mkOption {
       type = with types; attrsOf str;
--- a/modules/programs/direnv.nix
+++ b/modules/programs/direnv.nix
@@ -48,44 +48,32 @@ in {
       '';
     };
 
-    enableBashIntegration = mkOption {
-      default = true;
-      type = types.bool;
-      description = ''
-        Whether to enable Bash integration.
-      '';
-    };
+    enableBashIntegration =
+      lib.hm.shell.mkBashIntegrationOption { inherit config; };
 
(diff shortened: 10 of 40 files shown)


trueNAHO:
I assume the following error has nothing to do with my change:

<details open>
<summary><code>nix-shell --show-trace --arg enableBig false --pure tests -A run.all</code></summary>

```console
$ nix-shell --show-trace --arg enableBig false --pure tests -A run.all
swayr-basic-configuration: FAILED
Expected home-files/.config/swayr/config.toml to be same as /nix/store/09w5r31apd793gypn9lmwp91cf63045l-basic-configuration.toml but were different:
--- actual
+++ expected
@@ -13,10 +13,35 @@

 [layout]
 auto_tile = false
-auto_tile_min_window_width_per_output_width = [[800, 400], [1024, 500], [1280, 600], [1400, 680], [1440, 700], [1600, 780], [1680, 780], [1920, 920], [2048, 980], [2560, 1000], [3440, 1200], [3840, 1280], [4096, 1400], [4480, 1600], [7680, 2400]]
+auto_tile_min_window_width_per_output_width = [
+    [800, 400],
+    [1024, 500],
+    [1280, 600],
+    [1400, 680],
+    [1440, 700],
+    [1600, 780],
+    [1680, 780],
+    [1920, 920],
+    [2048, 980],
+    [2560, 1000],
+    [3440, 1200],
+    [3840, 1280],
+    [4096, 1400],
+    [4480, 1600],
+    [7680, 2400],
+]

 [menu]
-args = ["--show=dmenu", "--allow-markup", "--allow-images", "--insensitive", "--cache-file=/dev/null", "--parse-search", "--height=40%", "--prompt={prompt}"]
+args = [
+    "--show=dmenu",
+    "--allow-markup",
+    "--allow-images",
+    "--insensitive",
+    "--cache-file=/dev/null",
+    "--parse-search",
+    "--height=40%",
+    "--prompt={prompt}",
+]
 executable = "wofi"

 [misc]
For further reference please introspect /nix/store/v42kvixkz9k3pf2cl84lpxx0d8w8qazm-nmt-report-swayr-basic-configuration
wezterm-example-setting: FAILED
Expected home-files/.config/wezterm/colors/test.toml to be same as /nix/store/kxlsbx22rwij88ydvgkr22kjcdkdq2ar-test.toml but were different:
--- actual
+++ expected
@@ -1,7 +1,25 @@
 [colors]
-ansi = ["#222222", "#D14949", "#48874F", "#AFA75A", "#599797", "#8F6089", "#5C9FA8", "#8C8C8C"]
+ansi = [
+    "#222222",
+    "#D14949",
+    "#48874F",
+    "#AFA75A",
+    "#599797",
+    "#8F6089",
+    "#5C9FA8",
+    "#8C8C8C",
+]
 background = "#1B1B1B"
-brights = ["#444444", "#FF6D6D", "#89FF95", "#FFF484", "#97DDFF", "#FDAAF2", "#85F5DA", "#E9E9E9"]
+brights = [
+    "#444444",
+    "#FF6D6D",
+    "#89FF95",
+    "#FFF484",
+    "#97DDFF",
+    "#FDAAF2",
+    "#85F5DA",
+    "#E9E9E9",
+]
 cursor_bg = "#BEAF8A"
 cursor_border = "#BEAF8A"
 cursor_fg = "#1B1B1B"
For further reference please introspect /nix/store/c18f1lcmhd2avc1p78icbkbybhmgas27-nmt-report-wezterm-example-setting
```

</details>

modules/lib/shell.nix
```diff
@@ -1,11 +1,23 @@
 { lib }:
 
-rec {
+let
+  mkShellIntegrationOption = name: default:
+    lib.mkEnableOption "${name} integration" // {
+      inherit default;
+      example = !default;
```
rycee, line 7: Unfortunately cannot allow `default` or `example` to be affected by the configuration since it would in turn affect the documentation build. Can resolve it by adding a `defaultText` and make the example hard-coded. Something like this (untested) perhaps

```suggestion
      defaultText = literalExpression "config.shell.enableShellIntegration";
      example = false;
```

modules/misc/shell.nix
```diff
@@ -0,0 +1,30 @@
...
+        `shell.enable<SHELL>Integration` option. For example, the following
+        declaration globally disables shell integration for Bash:
+
+        ```nix
+        config.shell.enableBashIntegration = false;
+        ```
+      '';
+
+      example = false;
+      type = lib.types.bool;
```
rycee, line 25: Can sort the field in the more idiomatic way (also fix the overridden typo 🙂):

```suggestion
      type = lib.types.bool;
      default = true;
      example = false;
      description = ''
        Whether to enable shell integration for all supported shells.

        Individual shell integrations can be overridden with their respective
        `shell.enable<SHELL>Integration` option. For example, the following
        declaration globally disables shell integration for Bash:

        ```nix
        config.shell.enableBashIntegration = false;
        ```
      '';
```

modules/misc/shell.nix
```diff
@@ -0,0 +1,30 @@
...
+    };
+
+    enableZshIntegration = lib.hm.shell.mkZshIntegrationOption enable;
```
rycee, line 28: I think you can put this together with the other `enable<shell>Integration` options above.

rycee (commented):
Very cool, thanks for the contribution! I've added a few minor comments.

trueNAHO:
# Changelog

## v1: https://github.com/nix-community/home-manager/commit/80001f1ef5091b7e7721e58eaf01497a772e9517

- `/modules/misc/shell.nix`: reorder entries and fix typo

## v0: https://github.com/nix-community/home-manager/commit/4534c005eaad3360ef21f39d70dae3d473cfc6be

rycee:
@trueNAHO I pushed a commit with my proposal for how it could be done. To my eyes this is a bit cleaner and as far as I can tell there are no dependencies between the manual build and the user's configuration.

One thing I'm wondering is whether it really is necessary to have an `enableShellIntegration` option at all, I couldn't think of any compelling use-case.

Another thing is whether it perhaps would be good to have these top-level options under `home.` instead of `shell.`. I.e., `home.enableBashIntegration` or even `home.enableBashIntegrationGlobally` 🙂

trueNAHO:
> I pushed a commit with my proposal for how it could be done. To my eyes this is a bit cleaner

LGTM.

> as far as I can tell there are no dependencies between the manual build and the user's configuration.

If `defaultText` is what is used to build the manual, this should be fine.

> One thing I'm wondering is whether it really is necessary to have an `enableShellIntegration` option at all, I couldn't think of any compelling use-case.

Declaring `shell.enableShellIntegration = true` ensures that all my enabled shells receive all shell integrations provided by Home Manager.

> Another thing is whether it perhaps would be good to have these top-level options under `home.` instead of `shell.`. I.e., `home.enableBashIntegration` or even `home.enableBashIntegrationGlobally` 🙂

Since the `home` namespace already seems to be used by some global options (`home.enableDebugInfo` and `home.enableNixpkgsReleaseCheck`), we could move the options from `shell.*` to `home.shell.*`.

rycee:
My idea with the `shell.enableShellIntegration` perhaps not being necessary would be that the `shell.enableBashIntegration`, `shell.enableZshIntegration`, etc. still exist and simply have `true` as default value.

The downside would be that if you really want to disable all shell integration then you would have to do so for each of them. I suspect that is such a rare occurrence, though, that the pedagogical advantage of less indirection would be worth it…

khaneliman:
> My idea with the `shell.enableShellIntegration` perhaps not being necessary would be that the `shell.enableBashIntegration`, `shell.enableZshIntegration`, etc. still exist and simply have `true` as default value.
> 
> The downside would be that if you really want to disable all shell integration then you would have to do so for each of them. I suspect that is such a rare occurrence, though, that the pedagogical advantage of less indirection would be worth it…

I like this approach to standardize it treewide but allow overriding disabling. There are some programs that the "integration" auto launches it ie: zellij and user's would want to disable that integration explicitly.

khaneliman:
Sorry for the noise... was rebasing to resolve conflict and didn't realize you had just pushed something before that.

rycee:
I probably explained poorly, I meant something like <https://github.com/nix-community/home-manager/pull/6358/commits/5bc459a8bea9303a841b837b1c6a56a621ef5a4a>. It would of course still be able to override the shell integration for a specific module, just not change the default value across all shells at once.

trueNAHO:
> I probably explained poorly, I meant something like [5bc459a](https://github.com/nix-community/home-manager/commit/5bc459a8bea9303a841b837b1c6a56a621ef5a4a). It would of course still be able to override the shell integration for a specific module, just not change the default value across all shells at once.

Actually, this is the whole purpose of the `enableShellIntegration` variable, as visualized in the following dependency chain (`ALL_SHELLS -> EACH_SHELL -> EACH_PROGRAM_EACH_SHELL`):

```
                          enableBashIntegration       programs.gpg-agent.enableBashIntegration
                          enableFishIntegration       programs.gpg-agent.enableFishIntegration
enableShellIntegration -> enableIonIntegration     -> programs.gpg-agent.enableIonIntegration
                          enableNushellIntegration    programs.gpg-agent.enableNushellIntegration
                          enableZshIntegration        programs.gpg-agent.enableZshIntegration
```

Removing the `enableShellIntegration` option simply removes the following convenience:

> Declaring `shell.enableShellIntegration = true` ensures that all my enabled shells receive all shell integrations provided by Home Manager.
>
> -- https://github.com/nix-community/home-manager/pull/6358#issuecomment-2625592483

khaneliman:
> > I probably explained poorly, I meant something like [5bc459a](https://github.com/nix-community/home-manager/commit/5bc459a8bea9303a841b837b1c6a56a621ef5a4a). It would of course still be able to override the shell integration for a specific module, just not change the default value across all shells at once.
> 
> Actually, this is the whole purpose of the `enableShellIntegration` variable, as visualized in the following dependency chain (`ALL_SHELLS -> EACH_SHELL -> EACH_PROGRAM_EACH_SHELL`):
> 
> ```
>                           enableBashIntegration       programs.gpg-agent.enableBashIntegration
>                           enableFishIntegration       programs.gpg-agent.enableFishIntegration
> enableShellIntegration -> enableIonIntegration     -> programs.gpg-agent.enableIonIntegration
>                           enableNushellIntegration    programs.gpg-agent.enableNushellIntegration
>                           enableZshIntegration        programs.gpg-agent.enableZshIntegration
> ```
> 
> Removing the `enableShellIntegration` option simply removes the following convenience:
> 
> > Declaring `shell.enableShellIntegration = true` ensures that all my enabled shells receive all shell integrations provided by Home Manager.
> > -- [#6358 (comment)](https://github.com/nix-community/home-manager/pull/6358#issuecomment-2625592483)

I agree, this feels powerful and flexible. If the high level options default to false, a user can choose to keep toggling things per application module or just toggle shell specific integrations or all shell integrations treewide.

rycee:
Fair enough, I removed that commit and instead added two others to slightly improve an option description and also fix the usage of the functions inside the atuin and scmpuff modules. Please have a look and if everybody is happy we can merge it.

trueNAHO:
I made the following changes:

- `programs.direnv.enableFishIntegration` is no longer read-only. This is mentioned in the commit message and in `/modules/misc/news.nix`.
- Leverage `lib.hm.shell.mkFishIntegrationOption` in the `programs.direnv.enableFishIntegration` implementation.
- Update `time` in `/modules/misc/news.nix` entry.
- Prepend `extraDescription` in `mkShellIntegrationOption` from `/modules/lib/shell.nix` with `\n\n` instead of ` ` (a single space).
- Fixup all commits and add `Link:` and `Co-authored-by:` tags to the commit message.
- Double check that all `enable<SHELL>Integration` options are covered by this patch.

I assume the following `programs.fzf.tmux.enableShellIntegration` option should not inherit from `shell.enableShellIntegration` since they represent different things:

https://github.com/nix-community/home-manager/blob/f99c704fe3a4cf8d72b2d568ec80bc38be1a9407/modules/programs/fzf.nix#L145-L147

Maybe we should rename `programs.fzf.tmux.enableShellIntegration`?

> Please have a look and if everybody is happy we can merge it.

The following point is still unaddressed:

> > Another thing is whether it perhaps would be good to have these top-level options under `home.` instead of `shell.`. I.e., `home.enableBashIntegration` or even `home.enableBashIntegrationGlobally` 🙂
>
> Since the `home` namespace already seems to be used by some global options like `home.enableDebugInfo` and `home.enableNixpkgsReleaseCheck`, we could move the options from `shell.*` to `home.shell.*`. To what files should we move the new `/modules/lib/shell.nix` and `/modules/misc/shell.nix` content in that case?
>
> I think `home.shell.enableBashIntegration` is better than `home.shell.enableBashIntegrationGlobally` considering that it should be somewhat obvious and that the `description` already explicitly points this out.
>
> -- https://github.com/nix-community/home-manager/pull/6358#issuecomment-2625592483

Otherwise, this PR LGTM.

rycee:
I added a commit that simple changes the `misc/shell.nix` file to add the `home` attribute. I don't think it'll be necessary to move the file anywhere else.

About the fzf tmux option I think it's safest just to leave it as-is for now. As far as I know there haven't been any complaints about the name and changing it would just cause unnecessary churn for people.

About `programs.direnv.enableFishIntegration` I think perhaps it's best to keep it as read only. We explicitly made it read only at some point due to the confusion caused when people ended up having direnv in their fish shell even when they disabled the option in HM. We kept the option around mostly as a source of documentation, ultimately I think it could be completely removed.

trueNAHO:
> I added a commit that simple changes the `misc/shell.nix` file to add the `home` attribute. I don't think it'll be necessary to move the file anywhere else.

Do you know why the CI is failing?

> About the fzf tmux option I think it's safest just to leave it as-is for now. As far as I know there haven't been any complaints about the name and changing it would just cause unnecessary churn for people.

Agreed.

> About `programs.direnv.enableFishIntegration` I think perhaps it's best to keep it as read only. We explicitly made it read only at some point due to the confusion caused when people ended up having direnv in their fish shell even when they disabled the option in HM. We kept the option around mostly as a source of documentation, ultimately I think it could be completely removed.

I post-merged the option with `{ default = true; readOnly = true; }` and updated the commit message and `/modules/misc/news.nix` entry.

rycee:
The CI failures are due to two bugs in Nixpkgs. One of them is already fixed but the second is still pending: https://github.com/NixOS/nixpkgs/pull/380467

I'll run through the tests locally and if everything passes will merge.

trueNAHO:
> I'll run through the tests locally and if everything passes will merge.

OK.

I squashed all commits and updated the commit message with the new `home` namespace.

rycee:
Thanks for your patience, merged to master now 🙂

s1syph0s:
Hi!

is there a reason why the default value is set to `true`? It was for me an unexpected behavior, that my zellij starts automatically through this commit.. I think that it would be better if the default is set to `false`, and let the user decide if they want to enable shell integration globally or not.

trueNAHO:
> is there a reason why the default value is set to `true`?

The default value is `true` for convenience and because only the Zellij module is affected by this behavior.

> It was for me an unexpected behavior, that my zellij starts automatically through this commit..

Since shell integrations run arbitrary code, it seems impossible to generally define "unexpected behavior" in this case.

> I think that it would be better if the default is set to `false`, and let the user decide if they want to enable shell integration globally or not.

The same argument holds for the opposite argument of enabling it by default.

Either way, the Zellij shell integration should not have any effect unless `ZELLIJ_AUTO_ATTACH = true` or `ZELLIJ_AUTO_EXIT = true` are set, which are rather explicit opt-ins.

timhae:
> > I think that it would be better if the default is set to `false`, and let the user decide if they want to enable shell integration globally or not.
> 
> The same argument holds for the opposite argument of enabling it by default.
> 
> Either way, the Zellij shell integration should not have any effect unless `ZELLIJ_AUTO_ATTACH = true` or `ZELLIJ_AUTO_EXIT = true` are set, which are rather explicit opt-ins.

It does have an effect even without those environment variables, it starts zellij.

ony:
> It was for me an unexpected behavior, that my zellij starts automatically

Same was for me. I just had it in my config to try it out eventually.

> Since shell integrations run arbitrary code, it seems impossible to generally define "unexpected behavior" in this case.

That's true if we try to generalize. But somehow for me auto-start is not really a shell integrations. There is nothing that shell gets zellij or provides to zellij to call it an integration.  
From such perspective any program have shell integration because it can be started from shell.

Examples of shell integrations in my opinion are:
- Recognizing particular terminal and reporting special markers to it like jumping between prompts
- Getting keys or commands added like with broot/navi
- Kinda `zellij setup --generate-completion` , but we'd better make it more efficient through pre-compiled completions
- Adding information to prompt

Maybe those are not really an integrations. Just imagine what experience will happen if more programs will get similar arbitrary code as shell integrations.

trueNAHO:
> > Since shell integrations run arbitrary code, it seems impossible to generally define "unexpected behavior" in this case.
>
> That's true if we try to generalize. But somehow for me auto-start is not really a shell integrations. There is nothing that shell gets from zellij and/or provides to zellij to call it an integration. From such perspective any program have shell integration because it can be started from shell.
>
> Examples of shell integrations in my opinion are:
>
> * Recognizing particular terminal and reporting special markers to it like jumping between prompts
> * Getting keys or commands added like with broot/navi
> * Kinda `zellij setup --generate-completion` , but we'd better make it more efficient through pre-compiled completions
> * Adding information to prompt
>
> Maybe those are not really an integrations. Just imagine what experience will we get if more programs use similar arbitrary code as shell integrations.

Agreed.

In that case, it might be better to rename the `programs.zellij.enable<SHELL>Integration` options to something like `programs.zellij.enable<SHELL>Startup` and have it default to `false`. Feel free to submit a PR.

For reference, the `programs.zellij.enable<SHELL>Integration` defaults changed from `true` to `false` in commit 70c8bd08e6c186e5c628a4e5af6f7ad67cd344b8 ("zellij: disables shell integrations by default (#3981)").

ony:
> Agreed.
> 
> In that case, it might be better to rename the `programs.zellij.enable<SHELL>Integration` options to something like `programs.zellij.enable<SHELL>Startup` and have it default to `false`. Feel free to submit a PR.

I understand frustration of person who fixes/re-factors/unifies something to make better code-base.  
Good intent almost always pulls in more work than should be required.  
And I appreciate work people do on this project to keep things improving.

Lazy users most likely will either will not reach this point, or will just find work-around that will land in their configs.  
Hence, it is naïve to rely on person submitting PR after that. Beside that some people may misread "Feel free to submit PR" as a "polite" way to turn down follow-up changes/comments.

> However, should this behavior not be opt-in:
> 
> > Either way, the Zellij shell integration should not have any effect unless `ZELLIJ_AUTO_ATTACH = true` or `ZELLIJ_AUTO_EXIT = true` are set, which are rather explicit opt-ins.

Sorry. I'm not sure I can understand first phrase before quote.  
But in case if you still assume that statement in quote is correct you can check output of `zellij setup --generate-auto-start zsh` yourself. Version 0.42.1 (currently in `master`) gives:

```zsh
if [[ -z "$ZELLIJ" ]]; then
    if [[ "$ZELLIJ_AUTO_ATTACH" == "true" ]]; then
        zellij attach -c
    else
        zellij
    fi

    if [[ "$ZELLIJ_AUTO_EXIT" == "true" ]]; then
        exit
    fi
fi
```

Only case when it doesn't start is when running from `zellij` which automatically sets `ZELLIJ`.  
- `ZELLIJ_AUTO_ATTACH` only influences if we always start fresh or attempt to attach to existing session.
- `ZELLIJ_AUTO_EXIT` only influences if we should keep normal shell .

I.e. it is *not* an opt-in since `zellij` will be auto-started regardless of environment variables you mentioned. Only difference is with `attach -c` or not.

trueNAHO:
> > However, should this behavior not be opt-in:
>
> > > Either way, the Zellij shell integration should not have any effect unless `ZELLIJ_AUTO_ATTACH = true` or `ZELLIJ_AUTO_EXIT = true` are set, which are rather explicit opt-ins.
>
> [...] if you still assume that statement in quote is correct you can check output of `zellij setup --generate-auto-start zsh` yourself. Version 0.42.1 (currently in `master`) gives:
>
> ```shell
> if [[ -z "$ZELLIJ" ]]; then
>     if [[ "$ZELLIJ_AUTO_ATTACH" == "true" ]]; then
>         zellij attach -c
>     else
>         zellij
>     fi
>
>     if [[ "$ZELLIJ_AUTO_EXIT" == "true" ]]; then
>         exit
>     fi
> fi
> ```
>
> Only case when it doesn't start is when running from `zellij` which automatically sets `ZELLIJ`.
>
> * `ZELLIJ_AUTO_ATTACH` only influences if we always start fresh or attempt to attach to existing session.
> * `ZELLIJ_AUTO_EXIT` only influences if we should keep normal shell after exiting from `zellij`.
>
> I.e. it is _not_ an opt-in since `zellij` will be auto-started regardless of environment variables you mentioned. Only difference is with `attach -c` or not.

Yes, you are right. I must have somehow interpreted `-z "$ZELLIJ"` as `-n "$ZELLIJ"`.

> some people may misread "Feel free to submit PR" as a polite way to turn down follow-up changes/comments.

Sorry if it sounded that way. I am totally fine with reverting to the previous behavior of disabling the Zellij shell integration by default. As I previously mentioned, this change would essentially re-apply the original commit that already changed this default behavior:

> For reference, the `programs.zellij.enable<SHELL>Integration` defaults changed from `true` to `false` in commit 70c8bd08e6c186e5c628a4e5af6f7ad67cd344b8 ("zellij: disables shell integrations by default (#3981)").

If you want, I can submit this PR myself.

khaneliman:
> Lazy users most likely will either not reach this point, or will just find work-around that will land in their configs. Hence, it is naïve to rely on person submitting PR after that. And some people may misread "Feel free to submit PR" as a polite way to turn down follow-up changes/comments.

I wouldn't take it that way, personally. I know a lot of FOSS people that just want to make sure someone gets proper credit / dopamine rush for their contributions for the community and don't want to snipe an idea/PR from someone :P

trueNAHO:
> As a zellij user, I agree it was a minor annoyance to have them default enabled again. But, it was a quick fix in the config (since I was aware of the change) From working with PR's / issues related to zellij users who were more annoyed, I would welcome a PR to revert the default enable again.

Should we rename the option or just change its default value:

> In that case, it might be better to rename the `programs.zellij.enable<SHELL>Integration` options to something like `programs.zellij.enable<SHELL>Startup` and have it default to `false`.

khaneliman:
> > As a zellij user, I agree it was a minor annoyance to have them default enabled again. But, it was a quick fix in the config (since I was aware of the change) From working with PR's / issues related to zellij users who were more annoyed, I would welcome a PR to revert the default enable again.
> 
> Should we rename the option or just change its default value:
> 
> > In that case, it might be better to rename the `programs.zellij.enable<SHELL>Integration` options to something like `programs.zellij.enable<SHELL>Startup` and have it default to `false`.

From what I could see, the integration is auto attaching. They have some other environment variables to control other behavior once it's been auto attached. I'd just change the default for the module.

trueNAHO:
To: @ony, @s1syph0s, @timhae

See: https://github.com/nix-community/home-manager/pull/6695

PR #6369, opened 2025-01-29
imapnotify: add `wait` option to imapnotify-accounts

onemoresuza:
### Description

`wait` is no longer a part of the goimapnotify's config, but one of its
cli options.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@nickhu

### DIFF
--- a/modules/services/imapnotify-accounts.nix
+++ b/modules/services/imapnotify-accounts.nix
@@ -1,10 +1,8 @@
 { pkgs, lib, ... }:
-
-with lib;
-
-{
+let inherit (lib) mkOption types;
+in {
   options.imapnotify = {
-    enable = mkEnableOption "imapnotify";
+    enable = lib.mkEnableOption "imapnotify";
 
     onNotify = mkOption {
       type = with types; either str (attrsOf str);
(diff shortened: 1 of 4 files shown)


teto:
starting from what version ? is that in unstable yet ?

onemoresuza:
I've encountered this problem on goimapnotify 2.4, which is the current version for both nixos-24.11 and nixpkgs-unstable.

teto:
An `extraArgs` like in modules/services/swayidle.nix would be more generic.

onemoresuza:
Good idea. Going to implement it like it is on the swayidle module.

modules/services/imapnotify-accounts.nix
```diff
@@ -30,10 +30,15 @@ with lib;
       description = "IMAP folders to watch.";
     };
 
+    extraArgs = mkOption {
+      type = types.listOf types.str;
+      default = [ "-wait 1" ];
```
teto, line 35: the current module doesn't seem to set `wait` at all. Can we keep this behavior and use "-wait 1" as an example instead ?

PR #6371, opened 2025-01-29
Revert "thunderbird: add native host support (#6292)"

khaneliman:
This reverts commit 608b26d16ee69384580b1b14ac947d81d0240beb.

Forgot to test with my config before merging... looks like we have 2 modules trying to write to the same location with different options. Reverting to fix the conflict, we can re-implement it. 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/thunderbird.nix
+++ b/modules/programs/thunderbird.nix
@@ -136,24 +136,6 @@ let
     '') prefs)}
     ${extraPrefs}
   '';
-
-  nativeMessagingHostsPath = if isDarwin then
-    "${cfg.vendorPath}/NativeMessagingHosts"
-  else
-    "${cfg.vendorPath}/native-messaging-hosts";
-
-  nativeMessagingHostsJoined = pkgs.symlinkJoin {
-    name = "th_native-messaging-hosts";
-    paths = [
-      # Link a .keep file to keep the directory around
-      (pkgs.writeTextDir "lib/mozilla/native-messaging-hosts/.keep" "")
-      # Link package configured native messaging hosts (entire mail app actually)
-      cfg.package
-    ]
-    # Link user configured native messaging hosts
-      ++ cfg.nativeMessagingHosts;
-  };
-
 in {
   meta.maintainers = with hm.maintainers; [ d-dervishi jkarlson ];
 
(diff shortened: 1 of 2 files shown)


booxter:
@khaneliman which modules? firefox and thunderbird? do you hit the issue on linux or darwin?

khaneliman:
> @khaneliman which modules? firefox and thunderbird? do you hit the issue on linux or darwin?

It was on linux

booxter:
@khaneliman thanks. I can also take a look later. I suspect the culprit is in the fact that macos and linux builds expect inconsistent directories on these platforms, as mentioned in

```
        # Thunderbird doesn't look in `Application Support` on macOS for user
        # config (in contrast to global settings that are the same for Firefox
        # and Thunderbird):
        # https://developer.thunderbird.net/add-ons/mailextensions/supported-webextension-api
        default = if isDarwin then "Library/Mozilla" else ".mozilla";
        example = ".mozilla";
        description =
          "Directory containing the native messaging hosts directory.";
      };
```

For Darwin, the path is different so it doesn't overlap so it doesn't overlap.

```
      platforms.linux = rec {
        vendorPath = ".mozilla";
        configPath = "${vendorPath}/firefox";
      };
      platforms.darwin = {
        vendorPath = "Library/Application Support/Mozilla";
        configPath = "Library/Application Support/Firefox";
      };
```

khaneliman:
> @khaneliman thanks. I can also take a look later. I suspect the culprit is in the fact that macos and linux builds expect inconsistent directories on these platforms, as mentioned in
> 
> ```
>         # Thunderbird doesn't look in `Application Support` on macOS for user
>         # config (in contrast to global settings that are the same for Firefox
>         # and Thunderbird):
>         # https://developer.thunderbird.net/add-ons/mailextensions/supported-webextension-api
>         default = if isDarwin then "Library/Mozilla" else ".mozilla";
>         example = ".mozilla";
>         description =
>           "Directory containing the native messaging hosts directory.";
>       };
> ```
> 
> For Darwin, the path is different so it doesn't overlap.
> 
> ```
>       platforms.linux = rec {
>         vendorPath = ".mozilla";
>         configPath = "${vendorPath}/firefox";
>       };
>       platforms.darwin = {
>         vendorPath = "Library/Application Support/Mozilla";
>         configPath = "Library/Application Support/Firefox";
>       };
> ```

Sounds good, that makes sense.

PR #6383, opened 2025-01-30
nixos/common: forward `nix.enable` from the OS configuration

emilazy:
### Description

This prevents problematic references to `nix.package` when Nix management is disabled in the host OS.

Drafted until the corresponding `nix.enable` functionality is merged into nix-darwin, since this is probably more useful there than on NixOS.

---

**Note:** This work was funded by Determinate Systems. The design is my own.

### Checklist

- [x] Change is backwards compatible. *(‐ish; you could theoretically be relying on the host OS’s `nix.package` when `nix.enable` is turned off, but that sounds unwise…)*

- [ ] Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef). *(it doesn’t seem like there are any existing tests for `nix.enable`, so not sure where to add this)*

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/nixos/common.nix
+++ b/nixos/common.nix
@@ -34,6 +34,14 @@ let
           home.username = config.users.users.${name}.name;
           home.homeDirectory = config.users.users.${name}.home;
 
+          # Forward `nix.enable` from the OS configuration. The
+          # conditional is to check whether nix-darwin is new enough
+          # to have the `nix.enable` option; it was previously a
+          # `mkRemovedOptionModule` error, which we can crudely detect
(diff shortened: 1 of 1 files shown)


emilazy:
The corresponding nix-darwin PRs have been merged, so this should now be ready for review.

khaneliman (approved):
LGTM, thanks!

PR #6387, opened 2025-01-31
gpg-agent: no-allow-external-cache option

cafkafk:
### Description

This is a pretty common configuration option, and one that many people
will find useful to discover, specially if they're not using a desktop
environment.

I thought I would add it. It is also useful to have for
`pass-secret-service` in the future, since you'd likely want to avoid DE
keyrings.

Signed-off-by: Christina Sørensen <ces@fem.gg>

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/services/gpg-agent.nix
+++ b/modules/services/gpg-agent.nix
@@ -218,6 +218,21 @@ in {
         '';
       };
 
+      noAllowExternalCache = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          Tell Pinentry not to enable features which use an external cache for
+          passphrases.
+
+          Some desktop environments prefer to unlock all credentials with one
+          master password and may have installed a Pinentry which employs an
+          additional external cache to implement such a policy. By using this
+          option the Pinentry is advised not to make use of such a cache and
+          instead always ask the user for the requested passphrase.
+        '';
+      };
+
       extraConfig = mkOption {
         type = types.lines;
         default = "";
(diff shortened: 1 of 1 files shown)


modules/services/gpg-agent.nix
```diff
@@ -218,6 +218,21 @@ in {
...
+        default = false;
+        description = ''
+          Tell Pinentry not to enable features which use an external cache for
+          passphrases.
+
+          Some desktop environments prefer to unlock all credentials with one
+          master password and may have installed a Pinentry which employs an
+          additional external cache to implement such a policy. By using this
+          option the Pinentry is advised not to make use of such a cache and
+          instead always ask the user for the requested passphrase.
```
teto, line 232: I am not much familiar with gpg.
is `allow-external-cache` allowed along with `no-allow-external-cache` ? 

I like the effort to describe the option but I found it a bit long. Can this be summed up by `ask pin not to cache password`. What if the pin package ignores the request ? does it collide with other options like `defaultCacheTtl` in which case some warnings might be of interest ?

cafkafk: I took the description directly from gpg-agent, I think that without familiarity with gpg-agent and gpg, it wil likely not make much sense. It cannot be summed up as ask pin not to cache password.

It doesn't collide with other options. I am using it with all the options home manager currently provides.

cafkafk: Also, I'm not aware of an `allow-external-cache` option, at least it's not present in any documentation https://www.gnupg.org/documentation/manuals/gnupg/Agent-Options.html.

cafkafk:
ty :3

Issue #4618, opened 2023-10-30
declarative configurations of firefox extensions


teto:
### Description

We support declarative configuration of firefox but what about its addons ? I recently had to setup a new machine and lost access to my firefox sync profile and there are some extensions I would like to keep my settings via nix instead, for instance this one 
https://addons.mozilla.org/en-US/firefox/addon/refined-github-/?utm_source=addons.mozilla.org&utm_medium=referral&utm_content=search which I like but has a few bad defaults.

I have no idea how this works, if it's even possible. I suspect addons can use any mechanism they want unless they care about firefox sync. I found https://discourse.mozilla.org/t/in-which-file-are-stored-an-add-on-settings-options/26064 but I have no storage.js in my ~/.mozilla it looks like

NB: I know about https://github.com/nix-community/home-manager/issues/606 but wanted a distinct more focused issue

GetPsyched:
I wanted this for a long time too!

I did some digging and found:
1. Firefox's own management of extension configuration goes in these 2 files:
   - `~/.mozilla/firefox/<profile>/extension-preferences.json` # couldn't really decipher much of this
   - `~/.mozilla/firefox/<profile>/extension-preferences.json` # this has permissions and origins for both themes and extensions
2. Extensions themselves can have their own files in any location within a Firefox profile for their config (assumption)
   - The [containers extension](https://addons.mozilla.org/en-US/firefox/addon/multi-account-containers/) saves its container config in `~/.mozilla/firefox/<profile>/containers.json`
   - Extension state (not exactly config) are stored in `~/.mozilla/firefox/<profile>/storage/default/moz-extension+++<uuid>`

I've only recently learnt how to create NixOS modules, so, I can't help much in actually making a declarative config module (though I might later down the line) but yeah this definitely seems quite messy and/or volatile in terms of what config goes where. Perhaps someone with hands on experience messing with Firefox config would have a better idea.

DamienCassou:
I don’t think there is a way to configure extensions in general except through policies and most extensions don’t seem to support policy-based configuration.

I think we should close the issue as I don't think there is anything home-manager can do.

khaneliman:
https://github.com/nix-community/home-manager/pull/6389

Issue #3698, opened 2023-02-23
bug: search.json.mozlz4 symlink being replaced by file


OleMussmann:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Firefox causes home-manager to fail restarting.

After restarting Firefox, the `search.json.mozlz4` symlink is replaced by a file. This causes the home-manager to fail with:
```
Existing file '/home/[user]/.mozilla/firefox/default/search.json.mozlz4' is in the way of '/nix/store/[hash]-home-manager-files/.mozilla/firefox/default/search.json.mozlz4'
Please move the above files and try again or use 'home-manager switch -b backup' to back up existing files automatically.
```

I use home-manager as a module in NixOS. The `home-manager` command is thus not available, and `nixos-rebuild switch --flake` does not have a `-b backup` flag. The only way out is indeed to manually remove `search.json.mozlz4`. Then the symlink is re-created.

NixOS 22.11.20230222.c95bf18 (Raccoon)
Home-manager rev 2cb27c79117a2a75ff3416c3199a2dc57af6a527

### Maintainer CC

@rycee @kira-bruneau

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.1.12, NixOS, 22.11 (Raccoon), 22.11.20230222.c95bf18`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.11.1`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`
```

kira-bruneau:
Oh yep, there's a `programs.firefox.profiles.<name>.search.force` option to handle this. Its not enabled by default because it can wipe out an existing search configuration.

The docs mention that this is recommened when customzing search, but do you think theres anything that could help make this more discoverable?

OleMussmann:
Thanks @kira-bruneau , that indeed works nicely. I understand why it's not enabled by default.

I did not study _all_ search related options before playing with it. Maybe the `force` option could be mentioned in any section that would need it?

sg-qwt:
I'm not sure if this happens to anyone. But I do have a single profile setup with search being force enabled. But suddenly today, my default search has been changed to Bing without any notice after restarting firefox.  `search.json.mozlz4` is no longer a symlink as well.

thomaseizinger:
[...]

PR #6389, opened 2025-01-31
firefox: add support for declarative extension configurations

HPsaucii:
### Description

Adds support for declarative extension configuration through JSON by disabling IDB (check commit ca31bdda57eb0373041315d00074581e1215a8b6 for more details)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@rycee @brckd


HPsaucii:
Not backwards compatible due to previous lack of forwards compatibility previously in programs.firefox.profiles.<name>.extensions.

Added myself as module maintainer, but not a new module

HPsaucii:
closes https://github.com/nix-community/home-manager/issues/4618

HPsaucii:
(whoops, tests failed because i forgot to commit the description :p)

brckd:
Thank you for the promising pull request! I didn't even expect #4618 to ever be resolved at this point, but now I'm looking forward to it!

Because this change is not trivial I'll just summarize what I've gathered so far:
- This adds the `profiles.<profile>.extensions.settings.<extension>` option which writes its contents to `browser-extension-data/<extension>/storage.js"`
- Therefore it moves the previous `profiles.<profile>.extensions` option to `profiles.<profile>.extensions.packages`
- In order for the browser to use the `storage.js` file, the `extensions.webextensions.ExtensionStorageIDB.enabled` policy is set to `false`. This might introduce some regression in runtime-performance for extensions.

Overall it looks like a good approach, but I have a couple of questions for completeness.
- Have you tested this with any extension? Also consider adding unit tests somewhere in `tests/modules/programs/firefox/profiles` that test whether the correct `storage.js` is generated.
- Does this in theory work for all extensions, like those using the policy system? That would be important for consistency.

Ideally we would like to keep some backwards compatibility so users have some time to transition. It might be possible for both the old and the new options layout to coexist using `lib.types.either`.

ambroisie:
> Ideally we would like to keep some backwards compatibility so users have some time to transition. It might be possible for both the old and the new options layout to coexist using lib.types.either.

BTW: that's what `lib.types.coercedTo` is for.

brckd:
> > Ideally we would like to keep some backwards compatibility so users have some time to transition. It might be possible for both the old and the new options layout to coexist using lib.types.either.
> 
> BTW: that's what `lib.types.coercedTo` is for.

Yea this looks perfect! It would be nice if this allowed to raise warnings in the event the config was coerced though.

HPsaucii:
Hiya @brckd thanks for the response, I've also been really excited about the potential for having this implemented.

You summary is mostly correct, only nitpick is that `extensions.webextensions.ExtensionStorageIDB.enabled` is a user preference, as in from about:config and modified using the user.js file, and not what would typically be called a policy.

Keeping backwards compatibility would be a plus, even if its only temporary. I did not know that the libs you and @ambroisie suggested were a thing, and i'll get around to implementing it later today (GMT+0)

I've tested this pretty thoroughly, including with the following extensions:
- uBlock Origin
- uMatrix
- Redirector
- Stylus
- Decentraleyes

But it should work with any extension. This is because the `extensions.webextensions.ExtensionStorageIDB.enabled` preference changes how firefox and its derivatives handle extension requests to the `storage` API, but doesn't change the API itself.
I will still implement some sanity checks in the form of tests, again, later today.

Extensions that utilise the policy system (which is barely any, with uBlock Origin being the gold in the rough) will have their own unique behaviour with *both* policies and declarative changes to storage.js enabled simultaneously, and due to no upstream cohesive solution from mozilla, will vary from implementation to implementation. Atleast with uBlock Origin, it appears to prioritise policies over storage, regardless of whether its using storage.js or IDB. Regardless though, there shouldn't be any issues.

As far as I'm aware, it is not possible to manually or declaratively modify the IndexedDB storage. This is, to the best of my memory, because the storage itself has to be signed by mozilla, and cannot be mitigated.

Some extra things I want to clear up:

The performance difference should be minimal, at most a handful of milliseconds, and would only be present if the user set any declarative extension settings in the first place.

Additionally, a similar implementation **might** be possible for chromium based browsers. I have no proof of this and it's just purely speculation, but I hope to investigate it further eventually.

brckd:
Thanks for clarifying!

> i'll get around to implementing it later today (GMT+0)

Take as much time as you need, I'll be asleep and on vacation anyways ;)

> I've tested this pretty thoroughly, including with the following extensions:
> 
> * uBlock Origin
> * uMatrix
> * Redirector
> * Stylus
> * Decentraleyes

This is already a good selection of extensions one would configure declaratively! I would personally also be interested in the Containers, Dark Reader and Libredirect extensions and will test these once we get this PR done.

> I will still implement some sanity checks in the form of tests, again, later today.

Thanks, but again, no pressure.

> Extensions that utilise the policy system (which is barely any, with uBlock Origin being the gold in the rough) will have their own unique behaviour with _both_ policies and declarative changes to storage.js enabled simultaneously, and due to no upstream cohesive solution from mozilla, will vary from implementation to implementation. Atleast with uBlock Origin, it appears to prioritise policies over storage, regardless of whether its using storage.js or IDB. Regardless though, there shouldn't be any issues.

Alright, so this would indeed work consistently across extensions and keeps compatability with the policy-based approach. Perfect!

> As far as I'm aware, it is not possible to manually or declaratively modify the IndexedDB storage. This is, to the best of my memory, because the storage itself has to be signed by mozilla, and cannot be mitigated.

> The performance difference should be minimal, at most a handful of milliseconds, and would only be present if the user set any declarative extension settings in the first place.

In that case the JSON approach is absolutely sufficient.

> Additionally, a similar implementation **might** be possible for chromium based browsers. I have no proof of this and it's just purely speculation, but I hope to investigate it further eventually.

In my experience, Chromium is a bit less customizable e.g. when it comes to theming, but maybe you can surprise me once again!

HPsaucii:
@brckd i've implemented your suggestions, lmk if there's anything additional i should do :+1: 
(yes, i know, forgot to add a description for the second time now :facepalm:)

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -775,47 +840,59 @@ in {
...
+            recursive = true;
+            force = true;
+          };
+      }] ++
+        # Add extension settings as separate attributes
+        optional (profile.extensions.settings != { }) (mkMerge (mapAttrsToList
+          (name: settingConfig: {
+            "${profilesPath}/${profile.path}/browser-extension-data/${name}/storage.js" =
+              {
+                force = true;
```
brckd, line 892: Do we really need to forcefully override this? Consider adding a `profiles.<profile>.extensions.settings.<extension>.force` option similar to the `profiles.<name>.search.force` one.

```suggestion
                force = settingsConfig.force;
```

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -666,37 +662,106 @@ in {
...
+                        "ublock-unbreak"
+                        "ublock-quick-fixes"
+                      ];
+                    };
+                    ```
+
+                    Or in the form of a UUID, wrapped with curly braces
+                    ```nix
+                    # extensionID for Stylus
+                    "{7a7a4a92-a2a0-41d1-9fd7-1e92480d612d}" = {
```
brckd, line 727:
```suggestion
                    "{7a7a4a92-a2a0-41d1-9fd7-1e92480d612d}".settings = {
```

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -666,37 +662,106 @@ in {
...
+                };
 
-              To automatically enable extensions add
-              `"extensions.autoDisableScopes" = 0;`
-              to
-              [{option}`${moduleName}.profiles.<profile>.settings`](#opt-${moduleName}.profiles._name_.settings)
+                settings = mkOption {
+                  default = { };
+                  description = ''
+                    ExtensionID's for addons that you wish to configure.
```
brckd, line 708: "ExtensionID's" sounds like it would just be a list of IDs. It should be more clear that it's an attribute set of options.
```suggestion
                    Attribute set of options for each extension. The keys of the attribute set consist of the ID of the extension or its UUID wrapped in curly braces.
```

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -666,37 +662,106 @@ in {
...
+
+                    This can be in the form of either an ID string
```
brckd, line 710: Consider moving these examples from `description` to `example`.

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -666,37 +662,106 @@ in {
...
+                    to
+                    [{option}`${moduleName}.profiles.<profile>.settings`](#opt-${moduleName}.profiles._name_.settings)
+                  '';
+                };
 
-              To automatically enable extensions add
-              `"extensions.autoDisableScopes" = 0;`
-              to
-              [{option}`${moduleName}.profiles.<profile>.settings`](#opt-${moduleName}.profiles._name_.settings)
+                settings = mkOption {
```
brckd, line 705: It's slightly confusing that this is called settings even though the actual settings option is contained within. We can discuss whether we should change this to something like `options`.

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -666,37 +662,106 @@ in {
...
+                  default = { };
+                  description = ''
+                    ExtensionID's for addons that you wish to configure.
+
+                    This can be in the form of either an ID string
+                    ```nix
+                    # extensionID for uBlock Origin
+                    "uBlock0@raymondhill.net".options = {
```
brckd, line 713: Judging from the test you added I think you renamed that option to `settings`. https://github.com/nix-community/home-manager/blob/678dd3daeb85ee39d37561dc73aff73c1b1cfec4/tests/modules/programs/firefox/profiles/extensions/default.nix#L14
```suggestion
                    "uBlock0@raymondhill.net".settings = {
```

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -666,37 +662,106 @@ in {
...
+            default = { };
+            description = ''
+              Submodule for installing and configuring extensions.
+            '';
+            example = literalExpression ''
+              {
+                packages = with pkgs.nur.repos.rycee.firefox-addons; [
+                  ublock-origin
+                ];
+                "uBlock0@raymondhill.net".options = {
```
brckd, line 753:
```suggestion
                settings."uBlock0@raymondhill.net".settings = {
```

brckd (changes requested):
Just some minor suggestions. I also see a lot of formatting changes. Did you run nixfmt?

HPsaucii:
@brckd yeah i did, i think the issue might be that nixfmt doesn't play super nicely with git diff when moving stuff around, the test should have failed if it wasn't formatted according to ./format -c, just going to review your changes rq (yes, i was tired when i wrote all this, can't you tell :p)

modules/programs/firefox/mkFirefoxModule.nix
HPsaucii: I mean, we can, the issue arises that if firefox with extensions has been opened whatsoever, home-manager will fail to place the file

modules/programs/firefox/mkFirefoxModule.nix
HPsaucii: We could alternatively go do
```nix
extensions = {
  "[someName]" = {
    package = somePackage;
    options = {...};
  };
  extraPackages = [ someOtherPackage ];
};
```

modules/programs/firefox/mkFirefoxModule.nix
HPsaucii: Yeah, i'll do this in a little while

modules/programs/firefox/mkFirefoxModule.nix
brckd: Good idea but I think it's quite inconsistent because `extraPackages` would seem to be another extension.

brckd: We can leave it as is if you want. The description should be sufficient to clarify the option's intention.

modules/programs/firefox/mkFirefoxModule.nix
brckd: Reminds me of #3698.
> Its not enabled by default because it can wipe out an existing search configuration.

We should keep in mind that wiping out configurations is not the usual home manager behavior and should always be an active decision of the consuming user. However we can try to make that option as discoverable as possible.

modules/programs/firefox/mkFirefoxModule.nix
HPsaucii: Probably gonna leave it as-is for now, in lieu of a better solution

HPsaucii:
woops, wrong button, about to commit

modules/programs/firefox/mkFirefoxModule.nix
HPsaucii: fixed by https://github.com/nix-community/home-manager/pull/6389/commits/3243875f23216e646e067197ee3647313e242d85

modules/programs/firefox/mkFirefoxModule.nix
HPsaucii: fixed by https://github.com/nix-community/home-manager/pull/6389/commits/06b77de29a9f47aa9df90b646a19ea4f4ed34793

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -666,37 +662,114 @@ in {
...
+                  '';
+                };
 
-              To automatically enable extensions add
-              `"extensions.autoDisableScopes" = 0;`
-              to
-              [{option}`${moduleName}.profiles.<profile>.settings`](#opt-${moduleName}.profiles._name_.settings)
+                settings = mkOption {
+                  default = { };
+                  example = ''
```
brckd, line 707:
```suggestion
                  example = literalExpression ''
```

brckd (changes requested):
Just this one little fix and the PR should be good to go!

brckd (approved):
Everything is looking good now. @rycee I think you can merge this whenever you're ready.

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -666,37 +662,114 @@ in {
...
-            description = ''
-              List of ${name} add-on packages to install for this profile.
-              Some pre-packaged add-ons are accessible from the
-              [Nix User Repository](https://github.com/nix-community/NUR).
-              Once you have NUR installed run
+            type = types.coercedTo (types.listOf types.package) (packages: {
+              packages = mkIf (builtins.length packages > 0) (warn ''
+                In order to support declarative extension configuration,
+                extension installation has been moved from
+                programs.firefox.profiles.<profile>.extensions
```
kira-bruneau, line 670: Any instances of `programs.firefox` in the documentation should be replaced with `moduleName` since this file meant to be re-usable in firefox forks.

khaneliman (approved):
LGTM, tested locally and it works.

PerchunPak:
Great PR! Is there something like `about:config` where I can export my non-yet-declarative settings?

P.S. https://github.com/nix-community/home-manager/issues/4618 should be closed now

brckd:
Now that I'm thinking about it, wouldn't setting 
```nix
profiles.<name>.settings."extensions.webextensions.ExtensionStorageIDB.enabled" = false;
```
completely destroy all extension data? When I set it, all extension settings were wiped so I'd assume that. I hope there is a backup mechanism.

khaneliman:
> Now that I'm thinking about it, wouldn't setting
> 
> ```nix
> profiles.<name>.settings."extensions.webextensions.ExtensionStorageIDB.enabled" = false;
> ```
> 
> completely destroy all extension data? When I set it, all extension settings were wiped so I'd assume that. I hope there is a backup mechanism.

Yeah, you're switching from the old storage to new storage so it wipes out existing data.

brckd:
Since that option is implicitly set when declaring extension data, shouldn't we prevent this from happening accidentally?

khaneliman:
> Since that option is implicitly set when declaring extension data, shouldn't we prevent this from happening accidentally?

Would be open for a PR to make it more obvious about the repercussions of using it. I'm not sure best route other than an assertion with an option to "accept" changing that flag.

PR #6391, opened 2025-02-01
mpv: support includes directives

teto:
to allow mixing imperative and declarative config

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/mpv.nix
+++ b/modules/programs/mpv.nix
@@ -130,6 +130,19 @@ in {
         '';
       };
 
+      includes = mkOption {
+        type = types.listOf types.str;
+        default = [ ];
+        example = literalExpression ''
+          [
+            "~/path/to/config.inc";
+            "~/path/to/conditional.inc";
+          ]
+        '';
+        description =
+          "List of configuration files to include at the end of mpv.conf.";
+      };
+
       profiles = mkOption {
         description = ''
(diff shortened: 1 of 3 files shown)


ambroisie:
Should it have an ordering guarantee, coming before/after other nix-configured options?

modules/programs/git.nix
```diff
@@ -433,7 +433,7 @@ in {
         "git/config".text = generators.toGitINI cfg.iniContent;
 
         "git/ignore" = mkIf (cfg.ignores != [ ]) {
-          text = concatStringsSep "\n" cfg.ignores + "\n";
+          text = concatMapStringsSep "\n" (x: "include ${x}") cfg.ignores;
```
teto, line 436: @ambroisie do you mean like adding a `mkBefore` ?

ambroisie: Yeah `mkBefore`/`mkAfter` is what I had in mind. Not sure which is better, presumably `mkAfter` for local-only changes (would allow overrides _I assume_).

ambroisie: Also, I'm assuming the `git` change is unintentional?

teto: should all be fixed.

ambroisie (approved):

PR #6392, opened 2025-02-01
flake-module: rename `homeModules` to `homeManagerModules`

HeitorAugustoLN:
### Description

Renamed flake.homeModules to flake.homeManagerModules. It is the most common way to define home-manager modules in a nix flake.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/docs/manual/nix-flakes/flake-parts.md
+++ b/docs/manual/nix-flakes/flake-parts.md
@@ -22,7 +22,7 @@ you may wish to import Home Manager's flake module,
         inputs.home-manager.flakeModules.home-manager
       ];
       flake = {
-        # Define `homeModules`, `homeConfigurations`,
+        # Define `homeManagerModules`, `homeConfigurations`,
         # `nixosConfigurations`, etc here
       };
       # See flake.parts for more features, such as `perSystem`
@@ -30,10 +30,10 @@ you may wish to import Home Manager's flake module,
 }
 ```
 
-The flake module defines the `flake.homeModules` and `flake.homeConfigurations`
+The flake module defines the `flake.homeManagerModules` and `flake.homeConfigurations`
 options, allowing them to be properly merged if they are defined in multiple
 modules.
 
-If you are only defining `homeModules` and/or `homeConfigurations` once in a
+If you are only defining `homeManagerModules` and/or `homeConfigurations` once in a
 single module, flake-parts should work fine without importing
 `flakeModules.home-manager`.
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -14,11 +14,11 @@ in {
           that you can reference them in this or another flake's `homeConfigurations`.
         '';
       };
-      homeModules = mkOption {
+      homeManagerModules = mkOption {
         type = types.lazyAttrsOf types.unspecified;
         default = { };
         apply = mapAttrs (k: v: {
-          _file = "${toString moduleLocation}#homeModules.${k}";
+          _file = "${toString moduleLocation}#homeManagerModules.${k}";
           imports = [ v ];
         });
         description = ''


teto:
the doc references (it's faint I'll admit) `homeModules` though https://nix-community.github.io/home-manager/ so projects using `homeManagerModules` might be in the wrong here.

HeitorAugustoLN:
> the doc references (it's faint I'll admit) `homeModules` though https://nix-community.github.io/home-manager/ so projects using `homeManagerModules` might be in the wrong here.

The documentation was introduced with the flake module, so it just followed what was defined in the flake module.

khaneliman (approved):
Makes sense, looks like every other home manager module I use is this naming. Didn't notice it with original PR, thanks

teto:
I wonder if we should revert that. The burden of proof is on OP and there was none. All modules I use use homeModules.
Github shows I am not the only one https://github.com/search?q=homeModules+language%3ANix&type=code&l=Nix . 
If home-manager gets merged with nixpkgs `homeModules` makes more sense than `homeManagerModules`. We reward here modules that haven't respected the documentation and penalize those who did. @rycee

khaneliman:
> I wonder if we should revert that. The burden of proof is on OP and there was none. All modules I use use homeModules. Github shows I am not the only one https://github.com/search?q=homeModules+language%3ANix&type=code&l=Nix . If home-manager gets merged with nixpkgs `homeModules` makes more sense than `homeManagerModules`. We reward here modules that haven't respected the documentation and penalize those who did. @rycee

I merged the original or just a couple days ago that introduced it ? https://github.com/nix-community/home-manager/commit/a5e196d61f6e564f7fdeeb4a92cef92e1785d3f5

MattSturgeon:
> > If home-manager gets merged with nixpkgs `homeModules` makes more sense than `homeManagerModules`.
> 
> Is that something that will actually happen, you think?

I know there's been some discussion around this, but I haven't seen any conclusions that a) nixpkgs actually _want_ to merge home-manager into their repo or b) home-manager's owners/maintainers want home-manager to be moved into nixpkgs.

There's also been alternatives proposed, including nixpkgs coming up with its own home-configuration story.

But I don't think whether or not this is likely is _too_ relevant to this discussion.

> After thinking about it some more, I do agree that `homeManagerModules` and `hmModules` usage that I have in my config feels less sane than just doing `homeModules` similar to `darwinModules` and `nixosModules` (in line with @MattSturgeon reason for implementing that way).

Personally, I don't care _too_ much what other projects have used in the past, given this is anon-standard flake-output. It shouldn't be too hard for actively maintained flakes to migrate towards a naturally evolving standard.

This isn't the first time flake outputs have been soft-deprecated and migrated away from; e.g. `nixosModule` -> `nixosModules.default`.

But if we do want to look at existing "standards", I'd probably put some weight on the [experimental flake-schemas repo](https://github.com/DeterminateSystems/flake-schemas), which is also using `homeModules` and `homeConfigurations`.

Given `homeModules`, `hmModules`, and `homeManagerModules` are _all_ in use, I think we should just choose whichever we feel is the most appropriate name and hope that others eventually follow suit.

MattSturgeon:
> I would assert that the most logical one would be `homeManagerModules` since the class is called `homeManager`

I might be being slow, but I don't follow why it'd be more logical for the class to be different to the flake-output name?

isabelroses:
> I might be being slow, but I don't follow why it'd be more logical for the class to be different to the flake-output name?

Other nix projects have names based of the class like, nix-darwin's class is `darwin` therefore modules are called `darwinModules`. nixpkgs's nixos class is called `nixos` therefore the modules are called `nixosModules`.

MattSturgeon:
> Other nix projects have names based of the class like, nix-darwin's class is `darwin` therefore modules are called `darwinModules`. nixpkgs's nixos class is called `nixos` therefore the modules are called `nixosModules`.

Thanks for clarifying :slightly_smiling_face: 

Sounds to me like an argument for changing the class to `home` or `home-manager`, not an argument for changing the flake-output name. I see the two as almost entirely unrelated.

isabelroses:
> I see the two as almost entirely unrelated.

I think the two are somewhat related because the class is used to check if the imported modules are correct so there should be a correlation between the two. But as I said before we could totally change the class instead of the output name. They should just be consistent with each other.

teto:
> I merged the [original pr](https://github.com/nix-community/home-manager/commit/a5e196d61f6e564f7fdeeb4a92cef92e1785d3f5) just a couple days ago that introduced the module and documentation.

There was one "hidden" reference to homeModules in the official doc  or a template before even that PR. 
The github fight show both `homeModules` and `homeManagerModules`are used with the same order of magnitude.

@isabelroses thanks for evoking the "class" I had never used. It's one of the few "objective" unbiased info to help us. I find it logical to tie this to the flake output name. Ultimately, since there are not many reasons to go one way or another, I think it should be up to @rycee since it's his baby in the first place :)

rycee:
I personally use `hmModules` in NUR and my nix flake experiments but don't have any strong opinion in general. I guess `homeModules` is nice since it is shorter and `homeManagerModules` is nice because it is more explicit (and we wouldn't have to revert this PR). If I had to choose one of those two I guess it would be `homeManagerModules` with the idea that more explicit is more better and it is not something that you need to type particularly often.

Btw, there is some [work done](https://discourse.nixos.org/t/brainstorm-for-rfc-assimilate-home-manager-into-nixpkgs-monorepo/48230) on including HM in the nixpkgs repo.

MattSturgeon:
IMO it is important that modules and configurations have a consistent output name.

Currently, `homeConfigurations` is well documented and is standardised by the `home-manager` CLI.

Therefore I would argue we should use `homeModules` for modules.

As mentioned above, this is also the name used by the somewhat established [flake schemas repo](https://github.com/DeterminateSystems/flake-schemas), that may eventually get merged into `nix` (https://github.com/NixOS/nix/pull/8892).

teto:
come on @rycee, I call out to you as a figure of authority and you throw yet another crazier option :p 

I vote for `homeModules` because shorter, the historical one, all the reasons mentioned by @MattSturgeon (homeConfigurations, the flake schemas repos). This might imply updating the module `class = "homeManager"; `. Not sure how breaking is that, and how many users it affects.

isabelroses:
> Not sure how breaking is that, and how many users it affects.

Very breaking 😓. Not so sure how many people it will affect though. Its used for preventing bad imports of classes. So if we rename it now to `home` for example it will break every single module that sets `_class = "homeManger"`.

HeitorAugustoLN:
In the flake schema, it was supposed to be `homeManagerModules`, because it was the most common way to define home-manager modules (DeterminateSystems/flake-schemas#30). But it was changed to `homeModules`, because of this [review](https://github.com/DeterminateSystems/flake-schemas/pull/30#pullrequestreview-2200208003), and I don't know how true is that statement, that it is the preferred way.

rafaelrc7:
I am currently refactoring my dotfiles and noticed this discussion because of the flake-parts module. I've always used `homeModules` as I thought it was consistent with `homeConfigurations`. For consistency sake, shouldn't both options use the same prefix? Either `homeModules` and `homeConfigurations` or `homeManagerModules` and `homeManagerConfigurations`? 

From what I've seem `homeModules` vs `homeManagerModules` is quite divided. However, `homeConfigurations` is the de facto standard.

I also reached this PR by the fact that the flakeModule has also inconsistent documentation:

https://github.com/nix-community/home-manager/blob/433799271274c9f2ab520a49527ebfe2992dcfbd/flake-module.nix#L13

was not updated and still points to `homeModules`.

HeitorAugustoLN:
> I vote for `homeModules` because shorter, the historical one, all the reasons mentioned by @MattSturgeon (homeConfigurations, the flake schemas repos). This might imply updating the module `class = "homeManager"; `. Not sure how breaking is that, and how many users it affects.

I think I like `homeModules` over `homeManagerModules` too, at least on [public github repos](https://github.com/search?q=%22_class+%3D+%5C%22homeManager%5C%22%22+language%3ANix&type=code), aren't that many people that set `_class = "homeManager"`.

isabelroses:
> I think I like `homeModules` over `homeManagerModules` too, at least on [public github repos](https://github.com/search?q=%22_class+%3D+%5C%22homeManager%5C%22%22+language%3ANix&type=code&rgh-link-date=2025-02-08T13%3A35%3A39Z), aren't that many people that set `_class = "homeManager"`.

I couldn't say we can take this too much into account since we must also consider `mkModule` or any alternative wrapper. Such as these two examples
- https://github.com/tgirlcloud/beapkgs/blob/6d1ff49ea956b99106a70e1c562b5e92b6dd635a/flake.nix#L32
- https://github.com/catppuccin/nix/pull/465/files

HeitorAugustoLN:
> > I think I like `homeModules` over `homeManagerModules` too, at least on [public github repos](https://github.com/search?q=%22_class+%3D+%5C%22homeManager%5C%22%22+language%3ANix&type=code&rgh-link-date=2025-02-08T13%3A35%3A39Z), aren't that many people that set `_class = "homeManager"`.
> 
> I couldn't say we can take this too much into account since we must also consider `mkModule` or any alternative wrapper. Such as these two examples
> 
>     * https://github.com/tgirlcloud/beapkgs/blob/6d1ff49ea956b99106a70e1c562b5e92b6dd635a/flake.nix#L32
> 
>     * https://github.com/catppuccin/nix/pull/465/files

That's right, haven't thought of these cases.

PR #6401, opened 2025-02-03
tldr-update: init

PerchunPak:
### Description

This PR adds a simple systemd timer to automatically run `tldr --update`

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

CC @pedorich-n (maintainer of tealdeer module)

### DIFF
--- a/modules/programs/tealdeer.nix
+++ b/modules/programs/tealdeer.nix
@@ -80,6 +80,11 @@ in {
         See <https://tealdeer-rs.github.io/tealdeer/config.html> for more information.
       '';
     };
+
+    enableAutoUpdates = mkEnableOption "Auto updates" // {
+      default = true;
+      example = false;
+    };
   };
 
   config = mkIf cfg.enable {
@@ -89,5 +94,10 @@ in {
       mkIf (cfg.settings != null && cfg.settings != { }) {
         source = tomlFormat.generate "tealdeer-config" cfg.settings;
       };
+
+    services.tldr-update = mkIf cfg.enableAutoUpdates {
+      enable = true;
+      package = pkgs.tealdeer;
+    };
   };
 }
--- a/modules/services/tldr-update.nix
+++ b/modules/services/tldr-update.nix
@@ -1,0 +1,50 @@
+{ config, lib, pkgs, ... }:
+let cfg = config.services.tldr-update;
+in {
+  meta.maintainers = [ lib.maintainers.perchun ];
+
+  options.services.tldr-update = {
+    enable = lib.mkEnableOption ''
+      Automatic updates for the tldr CLI
+    '';
+
+    package = lib.mkPackageOption pkgs "tldr" { example = "tlrc"; };
+
+    period = lib.mkOption {
+      type = lib.types.str;
+      default = "weekly";
+      description = ''
+        Systemd timer period to create for scheduled {command}`tldr --update`.
+
+        The format is described in {manpage}`systemd.time(7)`.
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    systemd.user.services.tldr-update = {
+      Unit = {
+        Description = "Update tldr CLI cache";
+        Documentation = "https://tldr.sh/";
+      };
+
+      Service = {
+        Type = "oneshot";
+        ExecStart = ''
+          ${lib.getExe cfg.package} --update
+        '';
+      };
+    };
+
+    systemd.user.timers.tldr-update = {
+      Unit.Description = "Update tldr CLI cache";
+
+      Timer = {
+        OnCalendar = cfg.period;
+        Persistent = true;
+      };
+
+      Install.WantedBy = [ "timers.target" ];
+    };
+  };
+}
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -393,6 +393,7 @@ let
     ./services/taffybar.nix
     ./services/tahoe-lafs.nix
     ./services/taskwarrior-sync.nix
+    ./services/tldr-update.nix
     ./services/trayer.nix
     ./services/trayscale.nix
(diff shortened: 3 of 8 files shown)


modules/programs/tldr-update.nix
pedorich-n, line 1: Wouldn't it make more sense for this to be a part of `tldr`? Like `programs.tldr.auto-update`? You can also add yourself as a maintainer to the `tldr` module if you want.

tests/flake.lock
pedorich-n, line 1: Is this intentional?

modules/programs/tldr-update.nix
```diff
@@ -0,0 +1,50 @@
...
+    systemd.user.services.tldr-update = {
+      Unit = {
+        Description = "Update tldr CLI cache";
+        Documentation = "https://tldr.sh/";
+      };
+
+      Service = {
+        Type = "oneshot";
+        ExecStart = ''
+          "${lib.getExe cfg.package}" --update
```
pedorich-n, line 34: You don't usually need quotes around `lib.getExe`, IIRC 🤔

pedorich-n (commented):
Thanks for the PR!

modules/programs/tldr-update.nix
PerchunPak: There are a lot of [tldr implementations](https://github.com/tldr-pages/tldr/wiki/Clients), I wanted to separate this as a implementation-independent module

pedorich-n: Ah, right. I forgot that `tealdeer` is just one of the possible implementations.

pedorich-n: In that case maybe `services.tldr-update` is better? Since this module doesn't add anything other than a systemd service.

modules/programs/tealdeer.nix
```diff
@@ -89,5 +89,7 @@ in {
       mkIf (cfg.settings != null && cfg.settings != { }) {
         source = tomlFormat.generate "tealdeer-config" cfg.settings;
       };
+
+    programs.tldr-update.package = mkDefault pkgs.tealdeer;
```
pedorich-n, line 93: I think this should be a part of `tldr-update` and not `tealdeer`, if the intention is possibly using it with multiple clients.

Maybe something like `mkDefault (if config.programs.tealdeer.enable then config.programs.tealdeer.package else pkgs.tldr)`
And then if more clients become available, just add them to the list and chose one or something. 🤔

modules/programs/tldr-update.nix
PerchunPak: You're right, moved it to services

modules/programs/tealdeer.nix
PerchunPak: I am unsure if conditional default is possible in nix module system

khaneliman: It also needs to be updated for `service` instead of `programs`

modules/services/tldr-update.nix
```diff
@@ -0,0 +1,50 @@
+{ config, lib, pkgs, ... }:
+let cfg = config.programs.tldr-update;
+in {
+  meta.maintainers = [ lib.maintainers.perchun ];
+
+  options.programs.tldr-update = {
```
khaneliman, line 6: Looks like this was moved to services but is still optioned under programs.

modules/programs/tealdeer.nix
PerchunPak: > This also will be an issue when you have multiple programs that default to enabling this. You' d want to gate it behind a module option for enabling `tldr-update` support.

In that case user should set `tldr-update.package` in their own config

modules/services/tldr-update.nix
PerchunPak: Oh, you're right. Fixed that

modules/programs/tealdeer.nix
khaneliman: > > This also will be an issue when you have multiple programs that default to enabling this. You' d want to gate it behind a module option for enabling `tldr-update` support.
> 
> In that case user should set `tldr-update.package` in their own config

> There are a lot of [tldr implementations](https://github.com/tldr-pages/tldr/wiki/Clients), I wanted to separate this as a implementation-independent module

I'm saying if we implement the integration in more than one module and someone enables multiple providers we have created the conflict and force a user to resolve it. Instead of it being more flexible and opt in. 

```bash
error: The option `home-manager.users.khaneliman.services.tldr-update.package' is defined multiple times while it's expected to be unique.

       Definition values:
       - In `/nix/store/xxj87wb814rg1dyrmhwwzz1kjibrcln4-source/modules/programs/tlrc.nix': <derivation tlrc-1.9.3>
       - In `/nix/store/xxj87wb814rg1dyrmhwwzz1kjibrcln4-source/modules/programs/tealdeer.nix': <derivation tealdeer-1.7.1>
       Use `lib.mkForce value` or `lib.mkDefault value` to change the priority on any of these definitions.
```

But, It looks like not an issue atm if we don't have any other modules for the other providers.

PerchunPak: Added `programs.tealdeer.enableAutoUpdates`

khaneliman (approved):

modules/programs/tealdeer.nix
```diff
@@ -89,5 +94,10 @@ in {
       mkIf (cfg.settings != null && cfg.settings != { }) {
         source = tomlFormat.generate "tealdeer-config" cfg.settings;
       };
+
+    services.tldr-update = mkIf cfg.enableAutoUpdates {
+      enable = true;
+      package = pkgs.tealdeer;
+    };
```
PerchunPak, line 101: BTW tealdeer has auto-enable option in config

PR #6402, opened 2025-02-03
firefox: split bookmarks into separate submodule file

kira-bruneau:
### Description

This splits the bookmarks submodule into a seperate file, to make it easier to maintain (like how the search module was previously split out in #5697).

This also refactors bookmarks to require a new `force` option, to be more explicit about overriding existing bookmarks.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@brckd @rycee

### DIFF
--- a/modules/programs/firefox/mkFirefoxModule.nix
+++ b/modules/programs/firefox/mkFirefoxModule.nix
@@ -56,10 +56,10 @@ let
     else
       builtins.toJSON pref);
 
-  mkUserJs = prePrefs: prefs: extraPrefs: bookmarks: extensions:
+  mkUserJs = prePrefs: prefs: extraPrefs: bookmarksFile: extensions:
     let
-      prefs' = lib.optionalAttrs ([ ] != bookmarks) {
-        "browser.bookmarks.file" = toString (browserBookmarksFile bookmarks);
+      prefs' = lib.optionalAttrs (bookmarksFile != null) {
+        "browser.bookmarks.file" = toString bookmarksFile;
         "browser.places.importBookmarksHTML" = true;
       } // lib.optionalAttrs (extensions != { }) {
         "extensions.webextensions.ExtensionStorageIDB.enabled" = false;
@@ -110,59 +110,6 @@ let
           }
         ];
       }}
-    '';
-
-  browserBookmarksFile = bookmarks:
-    let
-      indent = level:
-        lib.concatStringsSep "" (map (lib.const "  ") (lib.range 1 level));
-
-      bookmarkToHTML = indentLevel: bookmark:
-        ''
-          ${indent indentLevel}<DT><A HREF="${
-            escapeXML bookmark.url
-          }" ADD_DATE="1" LAST_MODIFIED="1"${
-            lib.optionalString (bookmark.keyword != null)
-            " SHORTCUTURL=\"${escapeXML bookmark.keyword}\""
-          }${
-            lib.optionalString (bookmark.tags != [ ])
-            " TAGS=\"${escapeXML (concatStringsSep "," bookmark.tags)}\""
-          }>${escapeXML bookmark.name}</A>'';
-
-      directoryToHTML = indentLevel: directory: ''
-        ${indent indentLevel}<DT>${
-          if directory.toolbar then
-            ''
-              <H3 ADD_DATE="1" LAST_MODIFIED="1" PERSONAL_TOOLBAR_FOLDER="true">Bookmarks Toolbar''
-          else
-            ''<H3 ADD_DATE="1" LAST_MODIFIED="1">${escapeXML directory.name}''
-        }</H3>
-        ${indent indentLevel}<DL><p>
-        ${allItemsToHTML (indentLevel + 1) directory.bookmarks}
-        ${indent indentLevel}</DL><p>'';
-
-      itemToHTMLOrRecurse = indentLevel: item:
-        if item ? "url" then
-          bookmarkToHTML indentLevel item
-        else
-          directoryToHTML indentLevel item;
-
-      allItemsToHTML = indentLevel: bookmarks:
-        lib.concatStringsSep "\n"
-        (map (itemToHTMLOrRecurse indentLevel) bookmarks);
-
-      bookmarkEntries = allItemsToHTML 1 bookmarks;
-    in pkgs.writeText "${packageName}-bookmarks.html" ''
-      <!DOCTYPE NETSCAPE-Bookmark-file-1>
-      <!-- This is an automatically generated file.
-        It will be read and overwritten.
-        DO NOT EDIT! -->
-      <META HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=UTF-8">
-      <TITLE>Bookmarks</TITLE>
-      <H1>Bookmarks Menu</H1>
-      <DL><p>
-      ${bookmarkEntries}
-      </DL>
     '';
 
   mkNoDuplicateAssertion = entities: entityKind:
@@ -334,6 +281,8 @@ in {
     profiles = mkOption {
       inherit visible;
       type = types.attrsOf (types.submodule ({ config, name, ... }: {
+        imports = [ (pkgs.path + "/nixos/modules/misc/assertions.nix") ];
+
         options = {
           name = mkOption {
             type = types.str;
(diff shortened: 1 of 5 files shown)


tests/modules/programs/firefox/profiles/bookmarks/expected-bookmarks-user.js
```diff
@@ -2,7 +2,7 @@
 
 
 
-user_pref("browser.bookmarks.file", "/nix/store/00000000000000000000000000000000-@name@-bookmarks.html");
+user_pref("browser.bookmarks.file", "/nix/store/00000000000000000000000000000000-bookmarks.html");
```
kira-bruneau, line 5: The app name isn't actually necessary for this file, so I just removed it to make things simpler.

brckd:
Thanks for your contribution! I like the direction you're going with splitting everything into selfcontained submodules.

I see that you renamed the `bookmarks` option to `bookmarkSubmodule.bookmarks`. I don't think we should put implementation details ("submodule") in the option name, since that's nothing the user would be concerned about. I suggest calling it `bookmarks.settings` instead, as that would be in line with #6389's `extensions.settings`.

Apart from that it looks good to me!

brckd:
Did you add yourself as maintainer for the module?

kira-bruneau:
> I see that you renamed the `bookmarks` option to `bookmarkSubmodule.bookmarks`. I don't think we should put implementation details ("submodule") in the option name, since that's nothing the user would be concerned about. I suggest calling it `bookmarks.settings` instead, as that would be in line with #6389's `extensions.settings`.

Oh yep, that's why I kept that option as internal so it won't be exposed to the user. The alias I added for `bookmarks` -> `bookmarkSubmodule.bookmarks` should keep the existing structure, to avoid a breaking change.

> Did you add yourself as maintainer for the module?

Oh nope, I'm not too familiar with the bookmarks code myself. I was just going through other options that could be broken out into submodules. I'm not quite sure who'd be best to add here :thinking: - would it be ok to add you and @rycee?

@ethorsoe, would you be interested in being the maintainer since you originally added bookmarks support?

kira-bruneau:
Oh wait, actually I like the idea of the submodule matching https://github.com/NixOS/rfcs/pull/42.

I'll rename `bookmarksSubmodule.bookmarks` -> `bookmarksSubmodule.settings` & `bookmarksSubmodule.file` -> ``bookmarksSubmodule.configFile`, but I'll still keep the alias to avoid a breaking change.

brckd:
> Oh yep, that's why I kept that option as internal so it won't be exposed to the user.

Oh, I didn't catch that! In that case it shouldn't matter what the option containing the submodule is called.

kira-bruneau:
Hmmmm, I'm realizing that like the search submodule, this could potentially override any existing bookmarks.

I was planning on making the `enable` option for the search module act as the new `force` option, so I think it would make sense to do the same here.

That would be a breaking change though :sweat_smile: - I guess there's no point trying to avoid it.

kira-bruneau:
Ok it should be good now!

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -452,104 +399,31 @@ in {
...
-                    description = "Bookmark name.";
-                  };
-
-                  tags = mkOption {
-                    type = types.listOf types.str;
-                    default = [ ];
-                    description = "Bookmark tags.";
-                  };
+            type = (with types;
+              coercedTo (listOf anything) (bookmarks:
```
kira-bruneau, line 403: I used `(listOf anything)` here because I couldn't find a nice way to pull out the original type from the submodule.

I don't believe this is an issue since this will just redirect you to the new submodule format.

kira-bruneau:
Ok, I decided to add myself as a maintainer, since this change ended up becoming more than just a simple refactor, and I feel familiar enough with the code now. 

Ideally it'd be nice to have a maintainer who actually uses this feature though!

_I'm realizing, I might actually reconsider adding myself back as maintainer for all of firefox once we have everything organized out into smaller submodule files._

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -786,9 +660,9 @@ in {
 
       "${profilesPath}/${profile.path}/user.js" = mkIf (profile.preConfig != ""
         || profile.settings != { } || profile.extraConfig != ""
-        || profile.bookmarks != [ ]) {
+        || profile.bookmarks.enable) {
```
ethorsoe, line 663: does this change the compatibility, with existing configs and if so why is it done?

brckd:
> I was planning on making the `enable` option for the search module act as the new `force` option, so I think it would make sense to do the same here.
> 
> That would be a breaking change though :sweat_smile: - I guess there's no point trying to avoid it.

I still prefer `force` since IMO the module should be enabled implicitly and `force` be used only in case bookmarks cause issues. It's also more explicit about overriding something.

brckd:
@kira-bruneau Do you plan to resolve the merge conflict or do you need assistance?

kira-bruneau:
Oh yep sorry! I just noticed I was running out of energy on this and another firefox-related PR and needed to step away from them for a bit.

I'll take a look at this again today though!

brckd:
No worries! Only work on it if you want to.

modules/programs/firefox/mkFirefoxModule.nix
ethorsoe: I did not quite understand though, is this a breaking change or not, do old configs still produce same bookmarks with no changes?

Summary lists the change as backwards compatible, if I am not mistaken?

kira-bruneau: Sorry @ethorsoe for leaving this so long! I wanted to give a well thought out response, but kinda burnt myself out a few weeks ago and just needed to step away for a bit.

Yes this does break backwards compatibly. This PR was originally backwards compatible, but the latest push changed that. I'll update the description - thanks for catching that!

It breaks compatibility for a few reasons:
- Defining bookmarks with the home-manager module will stomp over any existing bookmarks, so enabling this behaviour should be explicit as possible
- Other submodules currently define a "force" option to try to make this stomping behaviour more explicit, but multiple users have run into confusing cases where their config get's overriden by firefox because they accidentally forgot to define `force = true`.
- An explicit enable option for submodules is more consistent how root-level modules are typically enabled

So by using an explicit enable option, we make this "stomping" behaviour more explicit, while avoiding the pitfalls of defining a seperate force option. The goal being that other submodules will also switch to using an explicit enable option (instead of using a force option).

kira-bruneau:
> I still prefer force since IMO the module should be enabled implicitly and force be used only in case something needs to be overridden. It's also more explicit about overriding something.

@brcked I agree that force is more explicit about overriding settings, but I think it's too confusing to have an additional option you need to enable, when it never really makes sense to have it disabled.

I think it's better to fold the force behaviour into an explicit "enable" option, and while "enable" is less explicit than "force", I think it's a better name for the combined behaviour.

kira-bruneau:
Wait - actually thinking about it more now, we could just add an assertion to request the user to add `force` if they haven't already.

brckd:
Sure! There should definitely be the an assertion for when the module doesn't do anything.

kira-bruneau:
Ok! I updated it to match the assertions that were added for the extensions submodule.

brckd (approved):
Looks good!

ethorsoe:
to confirm, this change is indeed not backwards compatible, I have no idea why this was merged without fixing it.

brckd:
> to confirm, this change is indeed not backwards compatible, I have no idea why this was merged without fixing it.

Thank you for reporting this issue. I'm sorry for any troubles caused.
How exactly is this change not backwards compatible? You will need to enable the `force` option now, but that should be pointed out in an assertion. Please let me know if I missed another incompatibility.

modules/programs/firefox/mkFirefoxModule.nix
ethorsoe: Here is an explicit promise `existing configs will continue to work`.

ethorsoe:
I am also unable to make the test you modified in this pass with just inserting `force = true;` I have absolutely no idea, in which world is this backwards compatible. Even just requiring any change is not backwards compatible and you should leave it unticked.

If a change is backwards compatible, you should not have to change the tests to make it pass.

Also old version allows bookmarks to be an attrset, it even explicitly defaults the name from attrset, probably work done by `coercedTo (attrsOf nodeType) attrValues (listOf nodeType);`.

ethorsoe:
name default possibly did not work anymore, but the code was still there.

ethorsoe:
It's actually possible this still works with settings, it's just you never see the assertion, if your bookmarks is an attrset, as it fails with unknown options.

ethorsoe:
ok attrset still works, only assertion does not work with the attrset

brckd:
> Also old version allows bookmarks to be an attrset, it even explicitly defaults the name from attrset, probably work done by `coercedTo (attrsOf nodeType) attrValues (listOf nodeType);`.

Ok, I didn't know that, thank you for pointing it out! I ran the relevant old test and it passes, because the attrset type actually wasn't tested.

```console
> git checkout 9d55428~ -- tests/modules/programs/firefox/profiles/{shared-path.nix,bookmarks/default.nix}
> nix build --reference-lock-file flake.lock ./tests#test-firefox-profiles-bookmarks 
warning: Git tree '/home/bricked/Documents/home-manager' is dirty
evaluation warning: Firefox bookmarks have been refactored into a submodule that now explicitly require a 'force' option to be enabled.
                    Replace:
                    programs.firefox.profiles.bookmarks.bookmarks = [ ... ];
                    With:
                    programs.firefox.profiles.bookmarks.bookmarks = {
                      force = true;
                      settings = [ ... ];
                    };
evaluation warning: Firefox bookmarks have been refactored into a submodule that now explicitly require a 'force' option to be enabled.
                    Replace:
                    programs.firefox.profiles.bookmarks.bookmarks = [ ... ];
                    With:
                    programs.firefox.profiles.bookmarks.bookmarks = {
                      force = true;
                      settings = [ ... ];
                    };
```
Since this seems to be the only incompatibility that went unnoticed, I'll fix it immediately! I'm sorry I didn't check the new assertion thoroughly enough.

brckd:
@ethorsoe see #6670! I am once again sorry for the inconvenience.

kira-bruneau:
Wow I'm sorry @ethorsoe, I was pretty certain I had every case covered, but obviously I was wrong.

Of course it broke the one case that didn't have a test written for it. Thanks for fixing this so quickly @brckd! :heart:

> You will need to enable the force option now, but that should be pointed out in an assertion. Please let me know if I missed another incompatibility.

I just wanted to clarify that configs without a `force` option will still work, there will just be a warning to migrate to the new `bookmarks` submodule. If you're using the new submodule, only then will you get an assertion if you don't set `force = true`.

danth:
Looks like the code constructing the assertion for `force` uses an undefined variable name. See https://github.com/danth/stylix/issues/1049

PR #6404, opened 2025-02-03
tests: rework derivation stubbing

rycee:
### Description

Instead of having to manually stub packages that should not be downloaded we instead automatically stub all packages (except a small list of whitelisted ones). Tests can re-introduce the real package by using the `realPkgs` module argument.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


khaneliman:
Nice, looks like that shaved 12-15 minutes off CI!

rycee:
Yeah, I'll clean this up a little more and hopefully merge soon 🙂 But it's still quite slow and I don't see an obvious reason why, so I think it still needs a bit more investigation.

booxter:
This PR broke darwin tests completely. There are some differences in stdenvs for darwin and linux that affect the evaluation. I've shared this issue and PR with Darwin team on Matrix and we are a bit confused what's going on but seems like the `coreutils`, `jq`, `gnugrep`, `gettext` overrides affect Darwin stdenv stage1 builds for these derivations.

I don't know what the "right" fix here is (a revert or some smart fix for scrubbing logic that I am too ignorant to see), but I think Darwin tests should still work.

PR #5229, opened 2024-04-05
Add flake-parts module

MattSturgeon:
### Description

Add a [flake-parts](https://flake.parts) module, output as `flakeModules.home-manager` and `flakeModules.default`.

The module defines options for `flake.homeModules` and `flake.homeConfigurations`, based on the respective nixos equivalents; [`flake.nixosModules`](https://github.com/hercules-ci/flake-parts/blob/9126214d0a59633752a136528f5f3b9aa8565b7d/modules/nixosModules.nix) and [`flake.nixosConfigurations`](https://github.com/hercules-ci/flake-parts/blob/9126214d0a59633752a136528f5f3b9aa8565b7d/modules/nixosConfigurations.nix).

This is mostly useful for people who want to define modules or configs in multiple flake modules. Without the options, they can still be defined, but the definition must be unique.

I've added some initial documentation, although I don't know how to build/browse the docs to check it renders how I expect.

### Checklist

- [x] Change is backwards compatible.

- [x] Code formatted with `./format`.

- [ ] Code tested through `nix-shell --pure tests -A run.all` or `nix develop --ignore-environment .#all` using Flakes.

- [ ] Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).

- [x] Commit messages are formatted like

    ```
    {component}: {description}

    {long description}
    ```

    See [CONTRIBUTING](https://nix-community.github.io/home-manager/#sec-commit-style) for more information and [recent commit messages](https://github.com/nix-community/home-manager/commits/master) for examples.

- If this PR adds a new module

  - [ ] Added myself as module maintainer. See [example](https://github.com/nix-community/home-manager/blob/068ff76a10e95820f886ac46957edcff4e44621d/modules/programs/lesspipe.nix#L6).

PR #6406, opened 2025-02-04
flake-module: revert rename 

khaneliman:
This reverts commit 066ba0c5cfddbc9e0dddaec73b1561ad38aa8abe.

### Description
Follow up to discussion https://github.com/nix-community/home-manager/pull/6392#issuecomment-2633445057 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@rycee @teto @MattSturgeon

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -14,12 +14,12 @@ in {
           that you can reference them in this or another flake's `homeConfigurations`.
         '';
       };
-      homeManagerModules = mkOption {
+      homeModules = mkOption {
         type = types.lazyAttrsOf types.deferredModule;
         default = { };
         apply = mapAttrs (k: v: {
           _class = "homeManager";
-          _file = "${toString moduleLocation}#homeManagerModules.${k}";
+          _file = "${toString moduleLocation}#homeModules.${k}";
           imports = [ v ];
         });
         description = ''
--- a/docs/manual/nix-flakes/flake-parts.md
+++ b/docs/manual/nix-flakes/flake-parts.md
@@ -22,7 +22,7 @@ you may wish to import Home Manager's flake module,
         inputs.home-manager.flakeModules.home-manager
       ];
       flake = {
-        # Define `homeManagerModules`, `homeConfigurations`,
+        # Define `homeModules`, `homeConfigurations`,
         # `nixosConfigurations`, etc here
       };
       # See flake.parts for more features, such as `perSystem`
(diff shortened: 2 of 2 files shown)


MattSturgeon (approved):

nyabinary:
I think homeManagerModules make more sense since it's more specific personally :3

MattSturgeon:
I'll attempt to summaries my arguments made in https://github.com/nix-community/home-manager/pull/6392:

IMO it is important that modules and configurations have a consistent output name.

Currently, `homeConfigurations` is well documented and is standardized by the `home-manager` CLI.

Additionally, these are also the names used by the somewhat established [flake schemas repo](https://github.com/DeterminateSystems/flake-schemas), that may eventually get merged into `nix` (https://github.com/NixOS/nix/pull/8892).

According to @teto, `homeModules` was also used in templates and/or less visible parts of documentation (prior to https://github.com/nix-community/home-manager/pull/5229).

The only strong argument I've seen against this is the module `class` name is currently `"homeManager"`, and it is often convention for the class-name to correlate with the flake output name. However, surely this argument _also_ applies to the configuration output name, is less user-facing, and something that could be addressed separately if needed.


> I think `homeManagerModules` [is] more specific

IMO, if we go down this route, we'd _also_ want to rename `homneConfigurations` -> `homeManagerConfigurations`, which is a _much_ more significant change.

teto:
the revert is coherent with the other discussion. Something about the `class` thing could be done but I confess I dont have a good grasp on the implications & it could be done later if need be. Can you rebase please ?

nyabinary:
wrong commit name btw :P you put homModules instead of homeModules

flake-module.nix
```diff
@@ -14,12 +14,12 @@ in {
...
         '';
       };
-      homeManagerModules = mkOption {
+      homeModules = mkOption {
         type = types.lazyAttrsOf types.deferredModule;
         default = { };
         apply = mapAttrs (k: v: {
-          _class = "homeManager";
-          _file = "${toString moduleLocation}#homeManagerModules.${k}";
+          _class = "home";
```
teto, line 21: this didn't appear in the originally reverted commit. Is it your change or a revert from another commit. I dont mind either way, as I suspect it affects only a very limited number of people, just wanted to know.

teto (approved):

flake-module.nix
khaneliman: There were a few changes to module since the original

MattSturgeon: I don't think it's correct to set the class to `"home"`

Based on others' comments, the class should still be `"homeManager"`

khaneliman: The only comment, that I can see, referring to the class mentions we could change it https://github.com/nix-community/home-manager/pull/6392#issuecomment-2634595310 if we wanted it to match the output.

PR #6407, opened 2025-02-04
firefox: fix referencing firefox fork name in profile-specific docs

kira-bruneau:
### Description

Within the context of the profiles submodule, `${name}` refers to the profile name, and not the firefox fork name.

This fixes all descriptions to use `${cfg.name}` instead (even where it's not necessary, for consistency).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@brckd @rycee

### DIFF
--- a/modules/programs/firefox/mkFirefoxModule.nix
+++ b/modules/programs/firefox/mkFirefoxModule.nix
@@ -201,7 +201,7 @@ let
     in {
       assertion = duplicates == { };
       message = ''
-        Must not have a ${name} ${entityKind} with an existing ID but
+        Must not have a ${cfg.name} ${entityKind} with an existing ID but
       '' + concatStringsSep "\n" (mapAttrsToList mkMsg duplicates);
     });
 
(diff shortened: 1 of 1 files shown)


brckd (approved):
Thank you for fixing this!

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -236,7 +236,7 @@ in {
       default = false;
       example = true;
       description = ''
-        Whether to enable ${name}.${
+        Whether to enable ${cfg.name}.${
```
rycee, line 239: Unfortunately this is not OK since it makes the manual output depend on your configuration, causing unnecessary rebuilds and makes the pre-built manual inconsistent with the manual an individual user has.

So, regrettably this needs to be reverted. If the output needs to be revised then it will have to be done without any reference to `config`.

kira-bruneau: Oh right sorry about that! I'll submit a fix that just references the attribute.

PR #6411, opened 2025-02-05
docs: reduce dependencies on user's configuration

rycee:
### Description

This fixes a few modules such that their documentation no longer depends on `config`. To reduce the risk of similar problems in the future this PR also introduces a "poison module" when building documentation. The poison module will abort the documentation build if a reference is made to any attribute inside `config` other than `_module`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -76,9 +76,35 @@ let
       ...
     }:
     let
+      # to discourage references from option descriptions and defaults.
+      poisonModule =
+        let
+          poisonAttr = n: {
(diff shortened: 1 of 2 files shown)


khaneliman:
@rycee rebased and pulled out the config specific stuff that already got merged in https://github.com/nix-community/home-manager/pull/7547

rycee:
@khaneliman Great, thanks!

PR #6417, opened 2025-02-08
vscode: Support Cursor AI

HyunggyuJang:
### Description

Support Cursor AI IDE as a VSCode family.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/vscode.nix
+++ b/modules/programs/vscode.nix
@@ -17,6 +17,7 @@ let
     "vscodium" = "VSCodium";
     "openvscode-server" = "OpenVSCode Server";
     "windsurf" = "Windsurf";
+    "cursor" = "Cursor";
   }.${vscodePname};
 
   extensionDir = {
@@ -25,6 +26,7 @@ let
     "vscodium" = "vscode-oss";
     "openvscode-server" = "openvscode-server";
     "windsurf" = "windsurf";
+    "cursor" = "cursor";
   }.${vscodePname};
 
   userDir = if pkgs.stdenv.hostPlatform.isDarwin then


Lalit64 (approved):

Lalit64:
Hey, is there any updates on the state of this?

HyunggyuJang:
@rycee Sorry to mention; can you take a look?

tylergets (approved):

tylergets (approved):

Lalit64:
@teto, Is it possible maybe to add an option to where you can specify the package to `pkgs.emptyDirectory` and way to specify `configDir`.

teto:
@Lalit64 I dunno I dont use vscode

tylergets:
I tested this and I don't believe extensions are working correctly here

HyunggyuJang:
> I tested this and I don't believe extensions are working correctly here because the cursor package is wrapping an app image https://github.com/NixOS/nixpkgs/blob/nixos-24.11/pkgs/by-name/co/code-cursor/package.nix

I’m using it so far (about one month), there were no issues. This changes just setting the user config location appropriately so that the cursor can look up for it. 
You might find my config helpful: https://github.com/HyunggyuJang/nix-config/blob/master/darwin/configuration.nix#L791

Lalit64:
@HyunggyuJang It seems that the extensions.json is not copy over and cursor is not loading the extensions, can you please check it out?

HyunggyuJang:
> @HyunggyuJang It seems that the extensions.json is not copy over and cursor is not loading the extensions, can you please check it out?

I’ve dug into it a bit, and it turns out that the upstream `code-cursor` doesn’t seem to support declarative extension setting at all for now as it is using AppImage: https://forum.cursor.com/t/cursor-is-now-available-on-nixos/16640/14. As a result, `.cursor/extensions` folder seems not recognized by cursor from linux side. I can confirm it works as expected from darwin (which I’m using), but it is quite hard to debug the issue on linux side.

I guess, the only downside from linux side is extension setting, and the other configs, like key bindings and settings, work fine as expected. Is that right?

Lalit64:
@HyunggyuJang I think that it is just the extensions.json file that has to copy over no? It seems that is the problem?

HyunggyuJang:
> @HyunggyuJang I think that it is just the extensions.json file that has to copy over no? It seems that is the problem? It is not working on darwin either which is the platform I am on.

On darwin, I’m using without any problems. Will look into the darwin side once again, though I cannot investigate linux side.

Issue #5190, opened 2024-03-26
Add nativeMessagingHosts for librewolf (for plasma desktop intergration)


SpiderUnderUrBed:
### Description
nativeMessagingHosts options for firefox already exists, while i would like that be be on more browsers, I would particulary like it to be a feature for librewolf as i think that may be the most used firefox fork on nixos, this is so i can enable plasma-desktop intergration, and yes i searched if this feature exists and it doesnt

plokid:
also want support for chromium browser

lolwierd:
+1 for support in chromium.  
[Similar option in nixos programs module](https://github.com/NixOS/nixpkgs/blob/9078780158b1a7364735c133eb8cb5d6276d7ccf/nixos/modules/programs/chromium.nix#L24) 
[Relevant PR](https://github.com/NixOS/nixpkgs/pull/232528)

brckd:
Could be fixed by #5684

johnhamelink:
I can confirm this isn't working yet:

Not working:

```nix
  programs.librewolf = {
    enable = true;

    nativeMessagingHosts = with pkgs; [
      ff2mpv-rust
      tridactyl-native
    ];

};
```

---

Working:

```nix
  programs.librewolf = {
    enable = true;

    package = pkgs.librewolf.override {
      nativeMessagingHosts = with pkgs; [
        # Convert Firefox URLs to MPV runs
        ff2mpv-rust

        # Native messenger for tridactyl to enable all the features.
        # You can confirm it's found by FF by running `:native`.
        tridactyl-native
      ];
    };
};
```

teto:
fixed ? reopen if not

PR #6421, opened 2025-02-09
firefox: correct vendorPath of firefox forks

Mikilio:
### Description

The path for messaging hosts seems to have been
~/.mozilla/native-messaging-hosts/ all along.
This PR fixes that
Solves #5190

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@bricked

### DIFF
--- a/modules/programs/floorp.nix
+++ b/modules/programs/floorp.nix
@@ -21,9 +21,12 @@ in {
 
       platforms.linux = {
         configPath = ".floorp";
-        vendorPath = ".floorp";
+        vendorPath = ".mozilla";
       };
-      platforms.darwin = { configPath = "Library/Application Support/Floorp"; };
+      platforms.darwin = {
+        configPath = "Library/Application Support/Floorp";
+        vendorPash = "Library/Application Support/Mozilla";
+      };
     })
   ];
 }
--- a/modules/programs/floorp.nix
+++ b/modules/programs/floorp.nix
@@ -25,7 +25,7 @@ in {
       };
       platforms.darwin = {
         configPath = "Library/Application Support/Floorp";
-        vendorPash = "Library/Application Support/Mozilla";
+        vendorPath = "Library/Application Support/Mozilla";
       };
     })
   ];
--- a/modules/programs/librewolf.nix
+++ b/modules/programs/librewolf.nix
@@ -30,12 +30,12 @@ in {
       unwrappedPackageName = "librewolf-unwrapped";
 
       platforms.linux = {
-        vendorPath = ".librewolf";
         configPath = ".librewolf";
+        vendorPath = ".mozilla";
       };
       platforms.darwin = {
-        vendorPath = "Library/Application Support/LibreWolf";
         configPath = "Library/Application Support/LibreWolf";
+        vendorPash = "Library/Application Support/Mozilla";
       };
 
       enableBookmarks = false;
--- a/modules/programs/librewolf.nix
+++ b/modules/programs/librewolf.nix
@@ -35,7 +35,7 @@ in {
       };
       platforms.darwin = {
         configPath = "Library/Application Support/LibreWolf";
-        vendorPash = "Library/Application Support/Mozilla";
+        vendorPath = "Library/Application Support/Mozilla";
       };
 
       enableBookmarks = false;


Mikilio:
The failure is completely unrelated to this PR

modules/programs/floorp.nix
```diff
@@ -20,7 +20,7 @@ in {
       visible = true;
 
       platforms.linux = {
-        configPath = ".floorp";
+        configPath = ".mozilla";
         vendorPath = ".floorp";
```
brckd, line 24: I think you meant to replace the vendor path, right?
```suggestion
        configPath = ".floorp";
        vendorPath = ".mozilla";
```

brckd (commented):
It makes sense that the config path and vendor path shouldn't be the same. As you already mentioned, we should check whether this change actually works for both LibreWolf and Floorp browser.

brckd:
I also noticed that LibreWolf's config and vendor paths on Darwin are the same as well. That can't be right. We should test whether we can use Mozila's vendor path as well, or otherwise remove that declaration.
https://github.com/nix-community/home-manager/blob/7850ab82239d763341198595c9f8024291901a01/modules/programs/librewolf.nix#L36-L39

modules/programs/floorp.nix
Mikilio: Yes ...

Mikilio:
Unfortunately I don't have a Mac and quickemu isn't doing it's thing for me.

brckd (approved):
Looks good now! Have you tested this change with any native messaging host on one of the affected browsers?

modules/programs/floorp.nix
```diff
@@ -21,9 +21,12 @@ in {
 
       platforms.linux = {
         configPath = ".floorp";
-        vendorPath = ".floorp";
+        vendorPath = ".mozilla";
+      };
+      platforms.darwin = {
+        configPath = "Library/Application Support/Floorp";
+        vendorPash = "Library/Application Support/Mozilla";
```
khaneliman, line 28:
```suggestion
        vendorPath = "Library/Application Support/Mozilla";
```

modules/programs/librewolf.nix
```diff
@@ -30,12 +30,12 @@ in {
...
 
       platforms.linux = {
-        vendorPath = ".librewolf";
         configPath = ".librewolf";
+        vendorPath = ".mozilla";
       };
       platforms.darwin = {
-        vendorPath = "Library/Application Support/LibreWolf";
         configPath = "Library/Application Support/LibreWolf";
+        vendorPash = "Library/Application Support/Mozilla";
```
khaneliman, line 38:
```suggestion
        vendorPath = "Library/Application Support/Mozilla";
```

Mikilio:
I tested the Linux derivations with the tridactyl native messenger application.

JuneStepp:
This broke native messaging on Librewolf on Linux for me. As you can see from [the documentation](https://librewolf.net/docs/faq/#how-do-i-get-native-messaging-to-work), Librewolf reads from ~/.librewolf.

brckd:
That's weird. Others have reported that it's the other way around.

Mikilio:
@JuneStepp may you share with apps you are trying to use? Usually it's the extensions themselves that choose the path for this. Note that Web compatibility sometimes is like the wild west.

JuneStepp:
> @JuneStepp may you share with apps you are trying to use? Usually it's the extensions themselves that choose the path for this. Note that Web compatibility sometimes is like the wild west.
> 
> Looking at your link, it looks more like librewolf specifically requires this symlink to make sure both paths point to the same directory.

I've disabled any self-managing of manifests in applications. This is what I'm using: (safer-trydactyl-native is just my customized version of trydactyl-native)
```nix
nativeMessagingHosts = [
    inputs.safer-tridactyl-native.packages.${pkgs.system}.safer-tridactyl-native
    pkgs.keepassxc
];
```
Both of them used to work in an earlier version of home-manager using the pkgs.librewolf.override method but now only work by manually symlinking the directories. (neither the override, the home-manger option, or both work now)

Here's an open issue in Librewolf for adding support for Mozilla manifest directories: https://codeberg.org/librewolf/issues/issues/2308
They mention the source code for native manifest paths there. I've confirmed that `XREUserNativeManifests` is ~/.librewolf on my installation.

I believe `Services.dirsvc.get("XREUserNativeManifests", Ci.nsIFile)` uses this code: (from [here](https://searchfox.org/mozilla-central/source/toolkit/xre/nsXREDirProvider.cpp))
```c++
#if defined(XP_UNIX) || defined(XP_MACOSX)
  else if (!strcmp(aProperty, XRE_SYS_NATIVE_MANIFESTS)) {
    rv = ::GetSystemParentDirectory(getter_AddRefs(file));
  } else if (!strcmp(aProperty, XRE_USER_NATIVE_MANIFESTS)) {
    rv = GetUserDataDirectoryHome(getter_AddRefs(file), false);
    NS_ENSURE_SUCCESS(rv, rv);
#  if defined(XP_MACOSX)
    rv = file->AppendNative("Mozilla"_ns);
#  else   // defined(XP_MACOSX)
    rv = file->AppendNative(".mozilla"_ns);
#  endif  // defined(XP_MACOSX)
  }
```

In https://codeberg.org/librewolf/source/src/branch/main/patches/mozilla_dirs.patch, Librewolf patches it with:
```diff
@@ -371,9 +371,9 @@ nsXREDirProvider::GetFile(const char* aProperty, bool* aPersistent,
     rv = GetUserDataDirectoryHome(getter_AddRefs(file), false);
     NS_ENSURE_SUCCESS(rv, rv);
 #  if defined(XP_MACOSX)
-    rv = file->AppendNative("Mozilla"_ns);
+    rv = file->AppendNative("LibreWolf"_ns);
 #  else   // defined(XP_MACOSX)
-    rv = file->AppendNative(".mozilla"_ns);
+    rv = file->AppendNative(".librewolf"_ns);
 #  endif  // defined(XP_MACOSX)
   }
 #endif  // defined(XP_UNIX) || defined(XP_MACOSX)
```

Floorp [has code](https://github.com/Floorp-Projects/Floorp/blob/ESR128/toolkit/components/extensions/NativeManifests.sys.mjs#L152) to specifically ensure `.mozilla` is used instead of `.floorp` (though `XREUserNativeManifests` is already .mozilla). I don't see anything similar in Librewolf.

Large grains of salt should be added to my limited understanding of these browsers, but I'm not sure how Librewolf could be using .mozilla for some people. Would maybe have to ask them.

kira-bruneau:
Oh wait never mind, I missed this on first glance, but it looks like the root cause is from this PR which includes fixes to sync the nix derivation patching with upstream: https://github.com/NixOS/nixpkgs/pull/388153

kira-bruneau:
Ok yep, it looks like what happened is that https://github.com/NixOS/nixpkgs/pull/336304 broke the application of librewolf patches between 129.0.1–136.0-2.

Then it looks like this PR was opened because of the breakage, assuming librewolf was designed to read the native extensions from `~/.mozilla`, when really what happened is that the patches were no longer being applied in the nix derivation.

Now that this was fixed in https://github.com/NixOS/nixpkgs/pull/388153, the home-manager module should reverted back to read from `~/.librewolf` and not `~/.mozilla`.

Mikilio:
> Now that this was fixed in [NixOS/nixpkgs#388153](https://github.com/NixOS/nixpkgs/pull/388153), the home-manager module should reverted back to write to `~/.librewolf` and not `~/.mozilla`. It looks like the vendorPath option should also be un-deprecated to support this.

So far, librewolf is the only browser behaving like this though and on the [link](https://librewolf.net/docs/faq/#how-do-i-get-native-messaging-to-work) above indicates that librewolf themselves recommend a workaround because some known extensions will still use the `~/.mozilla` path. It would be nice if someone could find a source that the standard for nativeMessengerHost is changing, and we should prepare for it. Otherwise, I'd say this is a Librewolf specific issue and can be solved with a single line added.

kira-bruneau:
> It would be nice if someone could find a source that the standard for nativeMessengerHost is changing, and we should prepare for it.

The firefox docs specify [their own location for app manifests](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Native_manifests#manifest_location), but there isn't a formal standard for native messaging. The Browser Extension Community Group previously attempted to define one, but still left native process registration up to the browser:

> Each browser implementation may have different mechanisms for registration, message filtering and lifetime management of the local client native process.

https://browserext.github.io/native-messaging/#local-client-native-process-registration

The group closed in 2023, and now link to https://github.com/w3c/webextensions, but w3c still don't specify a standard location for app manifests.

Since each browser defines their own location, I don't think we can always assume that firefox forks will always expect extensions in `~/.mozilla`:

- Firefox: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Native_manifests#manifest_location
- Chrome: https://developer.chrome.com/docs/extensions/develop/concepts/native-messaging#native-messaging-host-location
- Edge: https://learn.microsoft.com/en-us/microsoft-edge/extensions-chromium/developer-guide/native-messaging?tabs=v3%2Clinux#step-3-copy-the-native-messaging-host-manifest-file-to-your-system

I think it would be better if `programs.librewolf.nativeMessagingHosts` were written directly to `~/.librewolf/native-messaging-hosts` (because that's where librewolf expects them), and then also include a compat option to symlink `~/.mozilla` extensions into `~/.librewolf`, but I could see why it would be easier to just symlink from `~/.mozilla/native-messaging-hosts` -> `~/.librewolf/native-messaging-hosts`.

nakoo:
> I think it would be better if `programs.librewolf.nativeMessagingHosts` were written directly to `~/.librewolf/native-messaging-hosts` (because that's where librewolf expects them), and then also include a toggle-able compat option to symlink `~/.mozilla` extensions into `~/.librewolf`, but I could see why it would be easier to just symlink from `~/.mozilla/native-messaging-hosts` -> `~/.librewolf/native-messaging-hosts`.

I created a PR to address this issue. I would really appreciate it if you could test it.
https://github.com/nix-community/home-manager/pull/7127

PR #6422, opened 2025-02-09
misc/specialisation: escape specialisation name

ToborWinner:
The specialisation name is included in home.extraBuilderCommands without being properly escaped and checked. This commit fixes that.

### Description

The specialisation name is currently not escaped when included in the command to create the symlink in home.extraBuilderCommands. This PR properly escapes it and adds an assertion to ensure it does not contain a forward slash.

This PR, for example, prevents people from running commands through the specialisation name:
```nix
home-manager.users.tobor.specialisation."test && echo 'test' > $out/test.txt".configuration = { };
```
After rebuilding, the test.txt file is present in the built derivation:
```bash
❯ cat .local/state/home-manager/gcroots/current-home/test.txt
test
```

Note: specialisation.nix was probably implemented by copying how it works in nixpkgs, which currently has the same problem. I will probably submit a PR in nixpkgs later too.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/specialisation.nix
+++ b/modules/misc/specialisation.nix
@@ -71,10 +71,16 @@ with lib;
   };
 
   config = mkIf (config.specialisation != { }) {
+    assertions = map (n: {
+      assertion = !lib.hasInfix "/" n;
+      message =
+        "<name> in specialisation.<name> cannot contain a forward slash.";
+    }) (attrNames config.specialisation);
+
     home.extraBuilderCommands = let
       link = n: v:
         let pkg = v.configuration.home.activationPackage;
-        in "ln -s ${pkg} $out/specialisation/${n}";
+        in "ln -s ${pkg} $out/specialisation/${escapeShellArg n}";
     in ''
       mkdir $out/specialisation
       ${concatStringsSep "\n" (mapAttrsToList link config.specialisation)}


ToborWinner:
The check failure seems to have nothing to do with my PR (nushell is failing because of a hash mismatch).

modules/misc/specialisation.nix
```diff
@@ -71,10 +71,16 @@ with lib;
   };
 
   config = mkIf (config.specialisation != { }) {
+    assertions = map (n: {
+      assertion = match ".*/.*" n == null;
```
ambroisie, line 75: Maybe use the library function instead?

```suggestion
      assertion = lib.hasInfix "/" n == null;
```

ToborWinner: Oh yea, thank you! Didn't know that function existed (although you should remove the == null I think).

ToborWinner:
I'm pretty new to PRs, so I don't know if I resolved the conversation properly (I don't think it gave you the contribution? Not sure how I should have handled that), so let me know if I did anything wrong. Thank you for the suggestion.

ambroisie (approved):

ToborWinner:
Sorry, accidentally merged master while trying to make the tests pass (they are not failing because of my PR, they're failing because of nushell). I reverted my merge.

rycee:
Thanks for the contribution! Merged to master now 🙂

PR #6430, opened 2025-02-10
feat(bat): set it as MANPAGER

blaggacao:
On grounds of reason and probability (only), I claim that: this overwhelminly reflects what people who choose to enable bat expect

_Note: People who benefit from this might as well enjoy a supplementary integration with `batman`._

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/bat.nix
+++ b/modules/programs/bat.nix
@@ -136,6 +136,7 @@ in {
     })
     {
       home.packages = [ cfg.package ] ++ cfg.extraPackages;
+      home.sessionVariables = { MANPAGER = "bat -p -lman"; };
 
       xdg.configFile = lib.mkMerge ([{
         "bat/config" =
--- a/modules/programs/bat.nix
+++ b/modules/programs/bat.nix
@@ -136,7 +136,7 @@ in {
     })
     {
       home.packages = [ cfg.package ] ++ cfg.extraPackages;
-      home.sessionVariables = { MANPAGER = "bat -p -lman"; };
+      home.sessionVariables = { MANPAGER = lib.mkDefault "bat -p -lman"; };
 
       xdg.configFile = lib.mkMerge ([{
         "bat/config" =


blaggacao:
CI failure seems unrelated:

```
'/nix/store/j38j51h00hb61ldv5w20qp6hxmzxa1p9-nushell_plugin_formats-0.102.0-vendor-staging.drv':
         specified: sha256-88qcDlIukhpaUJ1vl1v8KMzj4XaYvATxlU+BOMZu6tk=
            got:    sha256-3cmNlCTawMUpr6kSyT/YZzC717FoXkF0uTeE/D8BSFM=
```

modules/programs/bat.nix
```diff
@@ -136,6 +136,7 @@ in {
     })
     {
       home.packages = [ cfg.package ] ++ cfg.extraPackages;
+      home.sessionVariables = { MANPAGER = "bat -p -lman"; };
```
khaneliman, line 139: We should gate this behind an option.

blaggacao: You mean if the variable is already set to something specifically by the user? (we could make it a default for that)

Or what kind of disruption do you have in mind?

ambroisie: The kind of disruption for users like me: I _do not_ want `bat` as my `MANPAGER`. If I wanted it to be so, I'd use `batman` instead.

blaggacao: Thanks! This:
- https://github.com/eth-p/bat-extras/blob/master/doc/batman.md
- https://github.com/eth-p/bat-extras/blob/master/src/batman.sh

I briefly looked at the docs and script and I couldn't comenup with a way this PR breaks the usage &mdash; as a user yourself, can you explain how this breaks your setup? Please anticipate the change to `mkDefault` in your explanation.

blaggacao: There doesn't seem to be built-in support for `batman` in HM, yet. If that were to land, then maybe it could be given priority over this default by introducing different override priorities around the default value (iirc 1000)

blaggacao: > It breaks the usage by setting MANPAGER: I don't want MANPAGER set. I want to have the default man experience by default.

Ah, so you don't agree with:

> On grounds of reason and probability (only), I claim that: this overwhelminly reflects what people who choose to enable bat expect

In line with what you also mention:

> and not the default experience when installing bat on other distros

That's a fair point, ofc. Before, I had read your statement as if you where presenting a concrete example of a breakage based on @khaneliman 's comment.

---

> > There doesn't seem to be built-in support for batman in HM, yet.

> `home.packages = [ pkgs.bat-extra.batman ]` is all that is necessary.

Oh, I should have clarified, that by "_support_" I rather meant "_integration_" that would go beyond that and include setting `MANPAGER`. I had wrongly assumed that was clear from context about how to setup `batman` in different scenarios.

modules/programs/bat.nix
```diff
@@ -136,6 +136,7 @@ in {
     })
     {
       home.packages = [ cfg.package ] ++ cfg.extraPackages;
+      home.sessionVariables = { MANPAGER = lib.mkDefault "bat -p -lman"; };
```
khaneliman, line 139: This isn't what was requested. This is still setting the variable for everyone automatically with no way to disable it. `mkDefault` just sets the option's value at a lower `mkOrder` so that higher priorities can override it, but they would have to be provided through the module system in the same eval.

blaggacao: This PR intends to make things simpler for a majority of users out-of-the-box. Adding an option imposes a cost which does no longer fulfills that intent.

khaneliman: > > with no way to disable it
> 
> That is not correct. People can set it to the default, despite them having previously stated a certain general preference for `bat` (via `bat.enable = true;`).
> 
> ```nix
> { MANPAGER = "less"; }
> ```

Right now `MANPAGER` is unset for me. I have `programs.bat` enabled. I do not want this change to set it for me. 

> This PR intends to make things simpler for a majority of users out-of-the-box. Adding an option imposes a cost which does no longer suficently fulfill that intent.

A lot of people just want a replacement for cat, too. Still should be an option to set an environment variable for an optional behavior. It only makes sense to forcibly enable things that are required for basic functionality.

modules/programs/bat.nix
ambroisie: > as if you where presenting a concrete example of a breakage

It _is_ a breakage, for all users (who suddenly have a different `MANPAGER`) but especially for those of us who _don't_ want `bat` as the `MANPAGER`.

blaggacao:
After this has shown to be controversial, I'll rather close and desist.

PR #6437, opened 2025-02-12
accounts: add auth attrset in order to add more auth methods in the future

blaggacao:
### Description

This PR should lay the foundation for adding new auth mechanisms (such as Gmail OAuth) in the module's current shapeliness (i.e. without adding arbitrary auth-related attributes to the top-level namespace).

This shapeliness will make it obvious and (once the necessary migration of current modules is accomplished) cheaper to implement new auth mechanisms on the generic accounts interface.

It appears that previous efforts were frustrated by overwhelming complexity. This PR tries to split up the problem into smaller chunks, and specifically ensure migration of existing modules is completed by the time it can be undrafted.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


blaggacao:
@rycee I already got some of the necessary refactorings done locally. But before I charge on and involve the maintainers of aerc, neomutt and the likes to help review them: do you agree with this approach in principle?

rycee:
Apologies for the long delay. In principle I see no problem with this but it is not obvious to me how the new authentication mechanisms would be added (such as the Gmail OAuth). Could you give an example? Would the idea be to make the `auth` submodule optional and introduce another optional submodule like `gmailAuth`?

soywod:
On `master`, looks like email has a `authenticationOption`. But PIM domains really lack of proper shared config. `authenticationOption` brings some SASL support, but HTTP-based protocols like JMAP, CardDAV, CalDAV lack its equivalent HTTP Auth. It should at least support Basic and Bearer.

Regarding OAuth there is also room for improvement. I recently published Ortie, a CLI to manage OAuth token https://github.com/nix-community/home-manager/pull/9799 (similar project as pizauth), and I found nothing in hm to integrate with. I don't know which form it should take, integration can go far. But the simplest form would be to just support SASL XOAUTH/OAUTHBEARER (partially the case), and HTTP Bearer Auth. `passwordCommand` can be kept as it is. I will try a PR soon.

Issue #6435, opened 2025-02-11
bug: programs.eww module fails to build due to missing 'isNull' attribute


Kodlak15:
### Are you following the right branch?

- [x] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Issue description

I was unable to rebuild my home-manager configuration today due to a missing attribute inside of the eww home-manager module. This is the error message I received:

```
error: attribute 'isNull' missing
at /nix/store/7xn20glmjzdffpjwjc7rkslvpm3jx4zv-source/modules/programs/eww.nix:49:14:
    48|     xdg.configFile."eww".source =
    49|       mkIf (!types.isNull cfg.configDir) cfg.configDir;
      |              ^
    50|
```

I gather the module is attempting to set the eww configuration directory source using `configDir` if `configDir` has been assigned a value. However, it seems as though `types` is missing the `isNull` attribute. I managed to fix this by simply using `builtins.isNull` rather than `types.isNull` and loading my patched module instead. I am not sure whether there was any particular reason this was the way it was in the first place, so I don't feel comfortable authoritatively declaring this as the solution to this problem. However, if this is acceptable I would be happy to open a pull request. 

### Maintainer CC

@mainrs 

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.12.12, NixOS, 25.05 (Warbler), 25.05.20250208.a79cfe0`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.24.12`
 - channels(root): `"nixos"`
 - nixpkgs: `/nix/store/j33wzkzndh41cyyy7i18bqm1srlv84cq-source`
```

KneeCapStealer:
I just had this same error.
It seems to be this commit: 59fe145f0bb7cfde99f148140a47a54deb8cc23f which added the change

Titaniumtown:
I can reproduce. Please fix!

LamprosPitsillos:
#6282 
PR that introduced this issue

KneeCapStealer:
I can make a pull request pretty quickly

KneeCapStealer:
Just made a pull request. Please yell if it is bad, I haven't contributed before #6438

PR #6438, opened 2025-02-12
eww: fix confDir using missing attribute "types.isNull" and add test

KneeCapStealer:
### Description

The eww module does not evaluate because of a missing attribute `types.isNull` being used for the `confDir` option config.
I have replaced `!types.isNull` with `!= null` to fix this error.
I also added a test, as tests were missing.

Issue: #6435 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@mainrs

### DIFF
--- a/modules/programs/eww.nix
+++ b/modules/programs/eww.nix
@@ -45,8 +45,7 @@ in {
 
   config = mkIf cfg.enable {
     home.packages = [ cfg.package ];
-    xdg.configFile."eww".source =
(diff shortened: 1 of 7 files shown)


KneeCapStealer:
That's odd. All tests passed for me using: `nix build --reference-lock-file flake.lock ./tests#test-all`. I am rerunning them just to make sure...

PR #6447, opened 2025-02-14
podman: link dependencies to containers

n-hass:
### Description

Changes (fixes ;) ) how the podman module resolves dependencies for generated services. 

###### Problem

We are not conforming to the Podman spec for specifying resource dependencies. We are duplicating writing systemd service dependencies in home-manager by manually adding the unit’s After/Requires, but Podman’s systemd generator already supports this if you reference a Quadlet’s full name.

For example, given a container,

```
[Container]
Image = my-img.build
```

and 'build' spec called "my-img" for an image,

```
[Build]
File = /path/to/Containerfile
```

This would currently **fail** to build in home-manager because the Podman systemd generator will look for a file `my-img.build` which it won’t find in the derivation’s source. This is jarring for users with knowledge of how Quadlets are _supposed_ to work.  Currently we require dropping the `.build` (or other resource type) in a config value and then ‘guess’ by name if its a reference to a quadlet (in this case, `services.podman.builds.my-img`), and add the service dependencies manually as mentioned if it matches.

###### Changes

Detect if a user uses a Quadlet reference in a container for Image, Volume, or Network fields, resolve if it is a config they have declared in home-manager, and then link the dependency config file to the derivation source of the container so that it can be built successfully. 

###### Compatability

Referencing a resource by name without its type suffix will not generate the dependency. So users rebuilding existing container configs which reference a network/image/volume by just the name will lose the service dependency without changing their config to be Quadlet spec-compliant. IE, changing `services.podman.containers.my-container.network = [ “my-net” ];` to `[ “my-net.network” ]`. 

However, for all existing configs, the resources _should_ already exist, and so the race is irrelevant because the service will be doing a no-op anyway. Nonetheless, a warning will print so users know to update their configs to be Quadlet spec-compliant if they want the generated container systemd services to have the correct dependencies on the resources it uses.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@bamhm182


bamhm182:
@n-hass, with #6586 and #6137 having been merged, it's time for this one. 🥳 

Just bring it up to date and resolve the merge conflicts and I'll give it another lookover ASAP.

n-hass:
@bamhm182 take a look when you're free

bamhm182:
@n-hass, right. Sorry. Been juggling too many things and have had this in mind, but haven't taken the time. Will take a look at this tonight for sure.

bamhm182:
Great work, overall, @n-hass. Unfortunately I ran into a bit of a bug. It turns out that when you have two different types of quadlets with the same name, some sort conflict occurs resulting in the following message.

```
error: builder for '/nix/store/fw7w8142prlzybs4sdhsqxqchxm2f4g8-home-container-podman-my-container.drv' failed with exit code 1;
       last 2 log lines:
       > Running phase: unpackPhase
       > ln: failed to create symbolic link '/nix/store/mj40l506xgmblb151njyp0f65ranj247-home-container-podman-my-container/quadlets/podman-myapp.volume': File exists
```

The following code causes this:

```
  services.podman = {
    enable = true;
    builds."my-bld" = {
      file = let
        containerFile = pkgs.writeTextFile {
          name = "Containerfile";
          text = ''
            FROM docker.io/alpine:latest
          '';
        };
      in "${containerFile}";
    };
    containers = {
      "my-container" = {
        image = "whatever:latest";
        network = [ "myapp.network" ];
        volumes = [ "myapp.volume:/data" ];
      };
    };
    networks."myapp" = {};
    volumes = {
      "myapp" = {};
    };
```

I'll see if I can fix it.

bamhm182:
Ran out of time for right now, but the problem appears to be that the podman.internal.builtQuadlets don't differentiate between `myapp.network` and `myapp.volume`. They both wind up as `myapp`. Was digging into what could be done to resolve this. Hoping to pick it back up in a few hours, but if you have an easy fix before then, happy to hear it.

bamhm182:
Confirmed good to go. Thanks, @n-hass!

PR #6453, opened 2025-02-14
git: change stateVersion check for compatiblity

khaneliman:
Original change was created before 24.11 release. We are in 25.05 now.

Follow up to https://github.com/nix-community/home-manager/pull/5516

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/git.nix
+++ b/modules/programs/git.nix
@@ -120,8 +120,8 @@ in {
         format = mkOption {
           type = types.enum [ "openpgp" "ssh" "x509" ];
           defaultText = literalExpression ''
-            "openpgp" for state version < 24.11,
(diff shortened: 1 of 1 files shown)


sumnerevans:
Whoops, thanks!

wzxu:
I'm still not quite sure how stateVersion works, but I'm currently using `stateVersion = "25.05"` and I suppose this would fix my failing build due to ```error: The option `home-manager.users.wzxu.programs.git.signing.format' was accessed but has no value defined. Try setting the option.```, but it still fails the same.

Do I have to set this option even if I'm not using signing? Or should I not use `stateVersion = "25.05"` at all? Thanks.

PR #6457, opened 2025-02-15
ntfy-sh: Add client module

matthiasbeyer:
### Description

ntfy-sh user background service.

This is my first time adding a module to home-manager. I've been using this locally for a few days now and it seems to work as intended.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/ntfy-sh.nix
+++ b/modules/programs/ntfy-sh.nix
@@ -1,0 +1,73 @@
+{
+  pkgs,
+  lib,
+  config,
+  ...
+}:
+
+let
+  cfg = config.programs.ntfy-sh;
+  yamlFormat = pkgs.formats.yaml { };
+in
+with lib;
+{
+  options.programs.ntfy-sh = {
+    enable = mkEnableOption "ntfy-sh";
+    enableUserService = mkEnableOption "ntfy-sh user service";
+
+    package = mkPackageOption pkgs "ntfy-sh" { };
+
+    configFilePath = mkOption {
+      type = types.path;
+      default = "${config.xdg.configHome}/ntfy/client.yml";
+      defaultText = literalExpression ''"''${config.xdg.configHome}/ntfy/client.yml"'';
+      description = "Path where the configuration file is placed.";
+    };
+
+    settings = mkOption {
+      type = yamlFormat.type;
+      default = { };
+      description = "Configuration for the ntfy binary";
+    };
+  };
+
+  config = mkIf cfg.enable {
+    assertions = [
+      {
+        assertion = cfg.settings != { };
+        message = ''
+          You must set `programs.ntfy-sh.settings`.
+        '';
+      }
+    ];
+
+    home.packages = [ cfg.package ];
+
+    home.file.ntfy-sh = {
+      enable = true;
+      target = cfg.configFilePath;
+
+      onChange = ''
+        ${pkgs.systemd}/bin/systemctl --user restart ntfy-sh-client
+      '';
+
+      source = mkIf (cfg.settings != { }) (yamlFormat.generate "ntfy-sh-config" cfg.settings);
+    };
+
+    systemd.user.services.ntfy-sh-client = mkIf cfg.enableUserService {
+      Unit.Description = "ntfy-sh client service";
+
+      Install.WantedBy = [ "default.target" ];
+
+      Service = {
+        Type = "simple";
+        Restart = "on-failure";
+        ExecStart = ''
+          ${cfg.package}/bin/ntfy subscribe --config ${cfg.configFilePath} --from-config
+        '';
+      };
+    };
+  };
+
+  meta.maintainers = [ maintainers.matthiasbeyer ];
+}


modules/lib/maintainers.nix
sumnerevans, line 1: No need to add yourself here, since you're already a maintainer in nixpkgs.

modules/programs/ntfy-sh.nix
```diff
@@ -0,0 +1,64 @@
...
+  options.programs.ntfy-sh = {
+    enable = mkEnableOption "ntfy-sh";
+    enableUserService = mkEnableOption "ntfy-sh user service";
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.ntfy-sh;
+      defaultText = literalExpression "pkgs.ntfy-sh";
+      description = "Package providing the {command}`ntfy`.";
+    };
```
sumnerevans, line 16:
```suggestion
    package = mkPackageOption pkgs "ntfy-sh" { };
```

modules/programs/ntfy-sh.nix
```diff
@@ -0,0 +1,64 @@
...
+
+    configFilePath = mkOption {
+      type = types.path;
+      default = "${config.home.homeDirectory}/.config/ntfy/client.yml";
+      defaultText = literalExpression
+        "''${config.home.homeDirectory}/.config/ntfy/client.yml";
```
sumnerevans, line 22:
```suggestion
      default = "${config.xdg.configHome}/ntfy/client.yml";
      defaultText = literalExpression
        "''${config.xdg.configHome}/ntfy/client.yml";
```

modules/programs/ntfy-sh.nix
```diff
@@ -0,0 +1,64 @@
...
+      description = "Path where the configuration file is placed";
```
sumnerevans, line 23: nit:

```suggestion
      description = "Path where the configuration file is placed.";
```

sumnerevans (changes requested):
You need to add `./programs/ntfy-sh.nix` to `modules/modules.nix`.

modules/lib/maintainers.nix
matthiasbeyer: cool!

khaneliman:
Should rebase so CI can pass

N4CH723HR3R:
hi, is still active or should i pick up this PR? @matthiasbeyer

matthiasbeyer:
Yes, rebasing.

modules/programs/ntfy-sh.nix
```diff
@@ -0,0 +1,59 @@
...
+    enable = mkEnableOption "ntfy-sh";
+    enableUserService = mkEnableOption "ntfy-sh user service";
+
+    package = mkPackageOption pkgs "ntfy-sh" { };
+
+    configFilePath = mkOption {
+      type = types.path;
+      default = "${config.xdg.configHome}/ntfy/client.yml";
+      defaultText = literalExpression
+        "''${config.xdg.configHome}/ntfy/client.yml";
```
N4CH723HR3R:
```suggestion
        ''"''${config.xdg.configHome}/ntfy/client.yml"'';
```
an example how it should be done, which i found in the wild
https://github.com/nix-community/home-manager/blob/0b44044694010a77b2ab8effc47d603c9f6ad032/modules/programs/boxxy.nix#L38

N4CH723HR3R:
nix fmt please, then the PR should be ready

modules/programs/ntfy-sh.nix
```diff
@@ -0,0 +1,64 @@
...
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    home.file.ntfy-sh = {
+      enable = true;
+      target = cfg.configFilePath;
+
+      onChange = ''
+        systemctl --user restart ntfy-sh-client
```
octvs: The `home-manager switch` throws

```sh
/tmp/home-manager-build.EOVcy23u5o/generation/activate: line 435: systemctl: command not found
```

```suggestion
        ${pkgs.systemd}/bin/systemctl --user restart ntfy-sh-client
```

modules/programs/ntfy-sh.nix
```diff
@@ -0,0 +1,64 @@
...
+    };
+
+    settings = mkOption {
+      type = yamlFormat.type;
+      default = { };
+      description = "Configuration for the ntfy binary";
+    };
+  };
+
+  config = mkIf cfg.enable {
```
octvs, line 34: At current its current state only setting

```nix
programs.ntfy-sh.enable = true;
```
fails with
```sh
… while calling the 'throw' builtin
 at «github:nixos/nixpkgs/64c08a7ca051951c8eae34e3e3cb1e202fe36786?narHash=sha256-tpyBcxPpcQb8ukyNF7DoCwfSY3VPsxHoYwj00Cayv5o%3D»/lib/modules.nix:1285:9:
 1284|         # handling.  If changed here, please change it there too.)
 1285|         throw
     |         ^
 1286|           "The option `${showOption loc}' was accessed but has no value defined. Try setting the option.";

error: The option `home.file.ntfy-sh.source' was accessed but has no value defined. Try setting the option.
```
Maybe adding an assertion would let it fail in a kinder way

```suggestion
  config = mkIf cfg.enable {
    assertions = [
      {
        assertion = cfg.settings != { };
        message = ''
          You must set `programs.ntfy-sh.settings`.
        '';
      }
    ];
```
which would say

```sh
       error:
       Failed assertions:
       - You must set `programs.ntfy-sh.settings`.
```

modules/programs/ntfy-sh.nix
octvs: This line also causes an activation failure if one didn't set `programs.ntfy-sh.enableUserService = true;` since the service won't exist

```sh
Failed to restart ntfy-sh-client.service: Unit ntfy-sh-client.service not found.
```

Although it isn't clear to me why there is `enableUserService` is there a use-case without this service?

modules/programs/ntfy-sh.nix
```diff
@@ -0,0 +1,64 @@
...
+        systemctl --user restart ntfy-sh-client
+      '';
+
+      source = mkIf (cfg.settings != { }) (yamlFormat.generate "ntfy-sh-config" cfg.settings);
+    };
+
+    systemd.user.services.ntfy-sh-client = mkIf cfg.enableUserService {
+      Unit.Description = "ntfy-sh client service";
+
+      Install.WantedBy = [ "multi-user.target" ];
```
octvs:
```suggestion
      Install.WantedBy = [ "default.target" ];
```

It can be due to my config, but this seems more fitting.

octvs (commented):
Thanks for the module @matthiasbeyer! 

I copied the module over to my config for use, here is my 2 denarii.

modules/programs/ntfy-sh.nix
matthiasbeyer: I am not sure about the `enableUserService` thing... :thinking: I will resolve this bit anyways, but feel free to add more discussion!

octvs: I can't follow "resolve this bit", wdym?

re: enableUserService: I was expecting the reasoning from you.

Beyond adding systemd unit only thing module does is to install pkg and write the config. The former is not a justification for a module. The latter, in its current state, is tightly knit to the unit file per my comment above.

matthiasbeyer: > I can't follow "resolve this bit", wdym?

This discussion is marked as resolved.

> The former is not a justification for a module. 

Is it not? Other modules do that same thing I suppose :thinking:

octvs:
```nix
    programs.ntfy-sh = {
      enable = true;
      settings.default-host = "https://a-host.net";
    };
```
-> 
```sh
Error: 
   0: Activation failed
   1: Activating configuration (exit status ExitStatus(Exited(5)))
      stderr:
      replacing old 'home-manager-path'
      installing 'home-manager-path'
      Failed to restart ntfy-sh-client.service: Unit ntfy-sh-client.service not found.
```
Doesn't look resolved.

octvs (changes requested):
Refer to the discussion https://github.com/nix-community/home-manager/pull/6457#discussion_r3361965011

PR #6459, opened 2025-02-15
home-cursor: add sway support

ambroisie:
### Description

Not sure if it should use `*` or `seat0` for the configuration, as I'm not sure what the difference in semantics would be.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@Scrumplex @alexarice @sumnerevans @oxalica @league

### DIFF
--- a/modules/config/home-cursor.nix
+++ b/modules/config/home-cursor.nix
@@ -56,6 +56,11 @@ let
           description = "The cursor size for hyprcursor.";
         };
       };
+
+      sway = {
+        enable = mkEnableOption
+          "sway config generation for {option}`home.pointerCursor`";
+      };
     };
   };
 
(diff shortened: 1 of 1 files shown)


khaneliman:
> Not sure if it should use * or seat0 for the configuration, as I'm not sure what the difference in semantics would be.

From what I can read in their docs, it's the difference between applying to the first seat and all possible seats.

ambroisie:
Well, yes. I guess I meant that I'm not sure which semantic is usually wanted: it's usually documented with `seat0` (e.g: Arch Wiki) even though to me `*` seems more correct. If one wants to customize the cursor, don't they want it customized everywhere?

khaneliman (approved):
LGTM

PR #6460, opened 2025-02-16
thunderbird: support native messaging hosts (2nd attempt)

booxter:
- **Reapply "thunderbird: add native host support (#6292)" (#6371)**
- **thunderbird, firefox: fix file conflict for native messaging hosts**

### Description

This is the second attempt to implement native messaging hosts for Thunderbird.
The previous attempt was reverted because it broke activation when both Firefox
and Thunderbird were installed. This happened because on Linux, both
applications use the same directory for native messaging hosts, and the module
did not handle this case correctly.

Note: I've added maintainers of the Firefox module to maintainers for the new
helper module since this is a relocated code that is now a dependency for the
main Firefox module.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/mozilla-messaging-hosts.nix
+++ b/modules/misc/mozilla-messaging-hosts.nix
@@ -1,7 +1,5 @@
 { config, lib, pkgs, ... }:
 
-with lib;
-
 let
   inherit (pkgs.stdenv) isDarwin;
 
@@ -22,22 +20,25 @@ let
   else
     ".mozilla/native-messaging-hosts";
 in {
-  meta.maintainers =
-    [ maintainers.booxter maintainers.rycee hm.maintainers.bricked ];
+  meta.maintainers = with lib.maintainers; [
+    booxter
+    rycee
+    lib.hm.maintainers.bricked
+  ];
 
   options.mozilla = {
-    firefoxNativeMessagingHosts = mkOption {
+    firefoxNativeMessagingHosts = lib.mkOption {
       internal = true;
-      type = with types; listOf package;
+      type = with lib.types; listOf package;
       default = [ ];
       description = ''
         List of Firefox native messaging hosts to configure.
       '';
     };
 
-    thunderbirdNativeMessagingHosts = mkOption {
+    thunderbirdNativeMessagingHosts = lib.mkOption {
       internal = true;
-      type = with types; listOf package;
+      type = with lib.types; listOf package;
       default = [ ];
       description = ''
         List of Thunderbird native messaging hosts to configure.
@@ -45,7 +46,7 @@ in {
     };
   };
 
-  config = mkIf (cfg.firefoxNativeMessagingHosts != [ ]
+  config = lib.mkIf (cfg.firefoxNativeMessagingHosts != [ ]
     || cfg.thunderbirdNativeMessagingHosts != [ ]) {
       home.file = if isDarwin then
         let
@@ -59,14 +60,14 @@ in {
           };
         in {
           "${thunderbirdNativeMessagingHostsPath}" =
-            mkIf (cfg.thunderbirdNativeMessagingHosts != [ ]) {
+            lib.mkIf (cfg.thunderbirdNativeMessagingHosts != [ ]) {
               source =
                 "${thunderbirdNativeMessagingHostsJoined}/lib/mozilla/native-messaging-hosts";
               recursive = true;
             };
 
           "${firefoxNativeMessagingHostsPath}" =
-            mkIf (cfg.firefoxNativeMessagingHosts != [ ]) {
+            lib.mkIf (cfg.firefoxNativeMessagingHosts != [ ]) {
               source =
                 "${firefoxNativeMessagingHostsJoined}/lib/mozilla/native-messaging-hosts";
               recursive = true;
--- a/modules/misc/mozilla-messaging-hosts.nix
+++ b/modules/misc/mozilla-messaging-hosts.nix
@@ -1,0 +1,89 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  inherit (pkgs.stdenv) isDarwin;
+
+  cfg = config.mozilla;
+
+  defaultPaths = [
+    # Link a .keep file to keep the directory around
+    (pkgs.writeTextDir "lib/mozilla/native-messaging-hosts/.keep" "")
+  ];
+
+  thunderbirdNativeMessagingHostsPath = if isDarwin then
+    "Library/Mozilla/NativeMessagingHosts"
+  else
+    ".mozilla/native-messaging-hosts";
+
+  firefoxNativeMessagingHostsPath = if isDarwin then
+    "Library/Application Support/Mozilla/NativeMessagingHosts"
+  else
+    ".mozilla/native-messaging-hosts";
+in {
+  meta.maintainers =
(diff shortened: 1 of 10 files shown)


booxter:
@khaneliman FYI

tests/modules/programs/thunderbird/thunderbird.nix
```diff
@@ -35,12 +37,18 @@
     };
   };
 
+  # Confirm that both Firefox and Thunderbird can be configured at the same time.
+  programs.firefox = { enable = true; };
```
khaneliman, line 41: <3

modules/modules.nix
```diff
@@ -30,6 +30,7 @@ let
     ./misc/fontconfig.nix
     ./misc/gtk.nix
     ./misc/lib.nix
+    ./misc/mozilla-messaging-hosts.nix
```
khaneliman, line 33: I think @kira-bruneau was working on breaking up the firefox module into separate files or something. Wondering if we have a better location for this.

modules/misc/mozilla-messaging-hosts.nix
```diff
@@ -0,0 +1,89 @@
...
+
+  config = mkIf (cfg.firefoxNativeMessagingHosts != [ ]
+    || cfg.thunderbirdNativeMessagingHosts != [ ]) {
+      home.file = if isDarwin then
+        let
+          firefoxNativeMessagingHostsJoined = pkgs.symlinkJoin {
+            name = "ff-native-messaging-hosts";
+            paths = defaultPaths ++ cfg.firefoxNativeMessagingHosts;
+          };
+          thunderbirdNativeMessagingHostsJoined = pkgs.symlinkJoin {
```
booxter, line 56: This will create the directory even when Thunderbird - or Firefox - is not enabled (the `||` conditional above should be adjusted).

tests/modules/programs/thunderbird/thunderbird.nix
booxter: :) On second thought, maybe we should have a separate test case for both installed. Otherwise we may miss a bug in the future where Thunderbird would rely on something that Firefox module implements, and would fail otherwise.

booxter:
There are some minor fixes and improvements for the PR that I would like to include (or it can be a follow up). But I will wait for @kira-bruneau to chime in here first.

khaneliman:
> There are some minor fixes and improvements for the PR that I would like to include (or it can be a follow up). But I will wait for @kira-bruneau to chime in here first.

Feel free to keep making changes/refactoring/whatever. Its easy enough to move a file.

booxter:
This is ready FWIW.

tests/modules/programs/thunderbird/thunderbird.nix
```diff
@@ -41,6 +43,10 @@
     # Disable warning so that platforms' behavior is the same
     darwinSetupWarning = false;
 
+    # Darwin doesn't support wrapped Thunderbird, using unwrapped instead;
+    # using -latest- because ESR is currently broken on Darwin
+    package = realPkgs.thunderbird-latest-unwrapped;
+
```
khaneliman, line 49: Is this needed for anything or can we just let it use the stubbed package?

tests/modules/programs/thunderbird/thunderbird.nix
```diff
@@ -1,6 +1,8 @@
-{ pkgs, ... }: {
+{ lib, realPkgs, ... }: {
   imports = [ ../../accounts/email-test-accounts.nix ];
 
+  _module.args.pkgs = lib.mkForce realPkgs;
+
```
khaneliman, line 5: This is removing the refactored testing stubbing, we should just be inheriting from realPkgs where needed.

tests/modules/programs/thunderbird/thunderbird.nix
booxter: `symlinkJoin` that creates native messaging hosts directory assumes it's a directory and fails

```
> @thunderbird@: No such file or directory
```

tests/modules/programs/thunderbird/thunderbird.nix
```diff
@@ -62,16 +66,31 @@
...
     };
 
+    nativeMessagingHosts = with realPkgs;
+      [
+        # NOTE: this is not a real Thunderbird native host module but Firefox; no
+        # native hosts are currently packaged for nixpkgs or elsewhere, so we
+        # have to improvise. Packaging wise, Firefox and Thunderbird native hosts
+        # are identical though. Good news is that the test will still pass as
+        # long as we don't attempt to run the mail client itself with the host.
+        # (Which we don't.)
```
khaneliman, line 76: One thing this just made me curious about because of this note. If we have a shared file for configuring the native messaging hosts and someone wants to install a firefox native messaging host.... are they going to get errors in thunderbird?

booxter: You mean whether Thunderbird will attempt to call to native hosts that belong to Firefox? Only when Thunderbird has an extension installed that would refer to this native host. For example here's how `cb_thunderlink` that I use spawns the corresponding native host: https://github.com/CamielBouchier/cb_thunderlink/blob/089490e44919dca6127f0398db65f05474e12050/mail_extension/cb_background.js#L452

I don't think TB and Firefox in general share native hosts; but if they do, each of them will start their own native process to communicate with.

booxter: If you suggest that the comment shouldn't imply there would be anything wrong if we WOULD run TB in the test suite per se, then yeah maybe we should clean up. I wanted to communicate that home-manager won't care what's in the actual directory. Perhaps I should make it more clear.

khaneliman: Alright, thanks. Sorry, not familiar with native messaging hosts really. Sounds like native messaging hosts just present an interface for communication that a plugin/extension has to try to communicate through itself. So we shouldn't just cause issues by their existence in a file system.

khaneliman:
Testing with my own config:
```bash
 error: A definition for option `home-manager.users.khaneliman.mozilla.firefoxNativeMessagingHosts."[definition 1-entry 1]"' is not of type `package'. Definition values:
       - In `/nix/store/jdpgxxj0kzrjcp1l14p88d3xq9wlamfx-source/modules/programs/firefox.nix': null

tests/modules/programs/thunderbird/thunderbird.nix
booxter: yeah. it's a way for TB/FF to talk to the system. But since it's generally unsafe, plugins can only do that through native hosts that are prepared (in our case by HM) in a well known directory. If a json file that describes the host module is put in the directory, then TB/FF will allow the plugin to execute it.

I will update the comment, I think it's indeed a bit misleading. (Esp. since as you pointed out we should handle the case of null package.)

booxter:
@khaneliman can you please share your config for Firefox? (Do you set `vendorPath`?)

khaneliman:
> @khaneliman can you please share your config for Firefox? (Do you set `vendorPath`?)



> @khaneliman can you please share your config for Firefox? (Do you set `vendorPath`?)

https://github.com/khaneliman/khanelinix/blob/main/modules/home/programs/graphical/browsers/firefox/default.nix

khaneliman:
LGTM overall, can you remove the `tests/flake.lock`

khaneliman (approved):
LGTM, thanks

PR #6466, opened 2025-02-16
nixpkgs-disabled: warn instead of assert

brckd:
Temporarily fixes backwards compatibility as issued in #6172.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@thiagokokada @justDeeevin

### DIFF
--- a/modules/misc/nixpkgs-disabled.nix
+++ b/modules/misc/nixpkgs-disabled.nix
@@ -63,11 +63,19 @@ in {
   };
 
   config = {
-    assertions = [{
-      assertion = cfg.config == null && cfg.overlays == null;
-      message = ''
-        `nixpkgs` options are disabled when `home-manager.useGlobalPkgs` is enabled.
-      '';
-    }];
+    assertions = [
+      # TODO: Re-enable assertion after 25.05 (&&)
+      {
+        assertion = cfg.config == null || cfg.overlays == null;
+        message = ''
+          `nixpkgs` options are disabled when `home-manager.useGlobalPkgs` is enabled.
+        '';
+      }
+    ];
+
+    warnings = optional ((cfg.config != null) || (cfg.overlays != null)) ''
+      You have set either `nixpkgs.config` or `nixpkgs.overlays` while using `home-manager.useGlobalPkgs`.
+      This will soon not be possible. Please remove all `nixpkgs` options when using `home-manager.useGlobalPkgs`.
+    '';
   };
 }


brckd:
I couldn't test this yet because my configuration sets both the `config` and `overrides` option. It can be tested by setting `inputs.home-manager.url = "github:brckd/home-manager/nixpkgs-disabled/warn-instead-of-error` though.

Doosty:
Would work if you added brackets around the two conditions i think
```sh
error: attempt to call something which is not a function but a Boolean: false
     73|     warnings = optional (cfg.config == null) != (cfg.overlays == null) ''
```

brckd:
> Would work if you added brackets around the two conditions i think

Thank you for your quick review! I attempted to fix this issue now, so feel free to test again.

Doosty:
Works
```sh
evaluation warning: myuser profile: You have set either `nixpkgs.config` or `nixpkgs.overlays` while using `home-manager.useGlobalPkgs`.
                    This will soon not be possible. Please remove all `nixpkgs` options when using `home-manager.useGlobalPkgs`.
Done. The new configuration is /nix/store/8vn5qyfljfbl4si268gbk76jk0rk64cz-nixos-system-nixos-steamdeck-01-25.05.20250213.2ff53fe
```

khaneliman (approved):
Thanks, I agree it makes sense to make a warning for a transition period for those unaware.

Devi0usM0nk:
Hi @brckd,

I have gotten this warning for the first time and I don't understand on what to change. I have used nixpkgs.config in flake.nix and nixos configuration as shown below. I haven't used overlays too. I am unable to understand what to do to fix this warning. I am new to nixos and need help. Thanks in Advance. 

flake.nix
```
outputs =
    {
      self,
      nixpkgs,
      stable-channel,
      home-manager,
      stylix,
      ...
    } @ inputs:
    let
      #
      #------ System Configuration ------#
      system = "x86_64-linux";
      lib = nixpkgs.lib;

      stable = import stable-channel {
        system = "${system}"; 
        config= {
          allowUnfree=true;
          allowUnfreePredicate = _: true;
          allowBroken = false;
        };
      };
 
      pkgs = import nixpkgs {
        system = "${system}"; 
        config= {
          allowUnfree=true;
          allowUnfreePredicate = _: true;
          allowBroken = false;
        };
      };

    in{

```

nixosconfigutation
```
  nixpkgs.config = {
    allowUnfree = true;
    allowBroken = false;
  };
```

dvalnn:
@Devi0usM0nk If you use Stylix, then it's most likely the cause. See this issue https://github.com/danth/stylix/issues/865

brckd:
Exactly, Stylix is likely the cause for this warning. Stylix is already working on this problem, so you can just ignore this warning until it's fixed.

Devi0usM0nk:
Thanks for the help, appreciate it.

PR #6479, opened 2025-02-17
zsh: refactor zsh configuration for better order control over `.zshrc`

Emin017:
### Description

- Refactor `zsh.nix` to use `mkMerge` and `mkOrder` for better control over `.zshrc` content ordering
- Users can add content anywhere by using `lib.mkOrder`, `lib.mkBefore` and `lib.mkAfter` custom configurations
- Added new test files (`zshrc-contents-insert.nix`, `zshrc-contents-insert-before.nix`, `zshrc-contents-insert-after.nix` and `zshrc-contents-init-zprof.nix`) to verify the behavior of `initExtra` with `mkBefore`, `mkAfter`, and default insertion.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


teto (changes requested):
this is getting out of hand. We would have
initExtraBeforeCompInit
initExtraLast
initExtra
. I suggest you split all the additions to zshrc with various `mkOrder`  so one can insert his code wherever he wants

modules/programs/zsh.nix
```diff
@@ -616,158 +622,162 @@ in
         ++ optional cfg.enableCompletion pkgs.nix-zsh-completions
         ++ optional cfg.oh-my-zsh.enable cfg.oh-my-zsh.package;
 
-      home.file."${relToDotDir ".zshrc"}".text = concatStringsSep "\n" ([
+      programs.zsh.initContents = mkMerge [
```
teto, line 625: the introduction of `initContents` is something I would like to avoid.
Can we keep `home.file."${relToDotDir ".zshrc"}".text` or use initExtra instead ?

teto:
thanks for taking on this huge work. I am quite looking forward to it.

modules/programs/zsh.nix
Emin017: In the new commit, I've modified the implementation to use `initExtra` instead of introducing `initContents` as you requested. 

My concern is:
To maintain compatibility, we should ensure that:
- The final .zshrc content maintains the same order of components
-  Any user-provided `initExtra` content still appears in the same location (Just as they used before)

I'm having a bit of a headache assigning this content order (I want to the default insertion order of `initExtra` is on the same location like before), do we need to consider the compatibility of the `initExtra` option? Or is this fine for now?

modules/programs/zsh.nix
```diff
@@ -616,158 +616,160 @@ in
         ++ optional cfg.enableCompletion pkgs.nix-zsh-completions
         ++ optional cfg.oh-my-zsh.enable cfg.oh-my-zsh.package;
 
-      home.file."${relToDotDir ".zshrc"}".text = concatStringsSep "\n" ([
+      programs.zsh.initExtra = mkMerge [
         # zprof must be loaded before everything else, since it
         # benchmarks the shell initialization.
-        (optionalString cfg.zprof.enable ''
+        (mkOrder 500 (optionalString cfg.zprof.enable ''
```
teto, line 622: mkBefore does `mkBefore = mkOrder 500;` so having this at 500 I am not sure how the merge happens, if it depends on eval ? lowering it to 400 or 300 would be better IMO since we most likely want for zprof to be loaded first.

teto:
on a quick glance it looks fine by me. I have follow up ideas but merging this first is better, it's the hardest part. diffing my ~/.zshrc with and without this just shows a few different newlines which is fine. I would like to let this cook a few days to give the opportunity for users with different zsh setups test this since it's a "dangerous" change :)

modules/programs/zsh.nix
teto: > The final .zshrc content maintains the same order of components

That would be ideal but it doesn't have to be 100% exact. The improvement you submitted solves several open issues so we can have small changes and the tradeoff would still be good.

> Any user-provided initExtra content still appears in the same location (Just as they used before)

not sure what you mean. Seems like a good thing.

> Or is this fine for now?

I think it is but I am not a specialist of the merging of elements in the module, I might have missed something.

teto (approved):

modules/programs/zsh.nix
Emin017: Changed the initialization order of `zprof` in the new commit.

modules/programs/zsh.nix
```diff
@@ -616,158 +616,160 @@ in
         ++ optional cfg.enableCompletion pkgs.nix-zsh-completions
         ++ optional cfg.oh-my-zsh.enable cfg.oh-my-zsh.package;
 
-      home.file."${relToDotDir ".zshrc"}".text = concatStringsSep "\n" ([
+      programs.zsh.initExtra = mkMerge [
```
teto, line 619: actually this will change the results for someone who set `initExtra = lib.mkForce []`, it must be `home.file."${relToDotDir ".zshrc"}".text` for backwards compatibility.

teto:
sry I intended to merge than noticed about the lib.mkForce issue. You will have to rebase sorry but I will try to merge it fast after that. Could you maybe mention in the description of `initExtra` how to insert zsh code with lib.mkOrder ?

modules/programs/zsh.nix
khaneliman: I believe this is central to the change, though? A user will pass different priorities to `initExtra` to write to the contents where they want.

teto: `home.file."${relToDotDir ".zshrc"}".text = mkBefore ` would still work but yeah that's less visible from the doc. 
Putting all the generated code in `initExtra` feels odd since it's not `extra` anymore ? In a previous version of the PR @Emin017 had added a new `content` option, maybe that's we need after all to keep backwards compatibility.

khaneliman: Yeah, I would feel better with a `content` or something that implies its the final config.

Emin017:
> `home.file."${relToDotDir ".zshrc"}".text = mkBefore ` would still work but yeah that's less visible from the doc. Putting all the generated code in `initExtra` feels odd since it's not `extra` anymore ? In a previous version of the PR @Emin017 had added a new `content` option, maybe that's we need after all to keep backwards compatibility.

Should I need to switch back to the previous version?

> sry I intended to merge than noticed about the lib.mkForce issue. You will have to rebase sorry but I will try to merge it fast after that. Could you maybe mention in the description of `initExtra` how to insert zsh code with lib.mkOrder ?

Sure, I will add examples and descriptions about this.

khaneliman:
Will be nice to help with an issue https://github.com/nix-community/home-manager/pull/6572 Currently, we are inserting into `initExtra` which isn't far enough down in the generated file.

teto:
> Should I need to switch back to the previous version?

@Emin017 I apologize for my bad advice, yes please restore the `content` option you had added.

tests/modules/programs/zsh/default.nix
```diff
@@ -9,4 +9,8 @@
   zsh-prezto = ./prezto.nix;
   zsh-syntax-highlighting = ./syntax-highlighting.nix;
   zsh-abbr = ./zsh-abbr.nix;
+  zshrc-contents-init-zprof = ./zshrc-contents-init-zprof.nix;
+  zshrc-contents-insert = ./zshrc-contents-insert.nix;
+  zshrc-contents-insert-before = ./zshrc-contents-insert-before.nix;
+  zshrc-contents-insert-after = ./zshrc-contents-insert-after.nix;
```
khaneliman, line 15: I don't think it's worth the extra CI time for these tests, in their current state. It would be better to have a single test that adds to the initContent with different priorities and verifying they are generated in the expected order.

Emin017: I merged them into a single priorities test in the new commit.

modules/programs/zsh.nix
```diff
@@ -471,6 +471,15 @@ in
         description = "Environment variables that will be set for zsh session.";
       };
 
+      initContent = mkOption {
+        default = "";
+        type = types.lines;
+        example = lib.mkOrder 1000 ''
```
teto, line 477: I think here it should be a `literalExample`  else it wont appear in doc. I will merge nevertheless because it's a very interesting PR and I dont want you to have to rebase it or take the risk of CI failing because of me pushing this small fix.

teto:
seemed to work fine here, I 've tested `initExtra = mkForce ""` for instance. 

as a follow up I think we could remove `initExtraFirst` for instance and avoid some of the empty newlines

modules/programs/zsh.nix
teto: yeah it appears as 
           ```
           Type: strings concatenated with “\n”

           Default: ""

           Example:

               {
                 _type = "order";
                 content = ''
                   echo "Hello, initContent!"
                 '';
                 priority = 1000;
               }
               ```
               
               in the doc, which naughty

Emin017: I think I can open another pr to fix this.

PR #6480, opened 2025-02-17
git: fix setting format on >=25.05

sumnerevans:
### Description

With the following config:

```
programs.git = {
  enable = true;
  signing = {
    format = "ssh";
    key = "~/.ssh/id_ed25519";
    signByDefault = true;
  };
  userName = fullName;
  userEmail = userEmail;
};
```

on stateVersion 25.05, you get the following error:

```
The option `home-manager.users.timo.programs.git.signing.format` is defined both null and not null, in `/nix/store/ip7y1cxy95vwscjy8ll86m5ls9i5kiad-source/modules/programs/git.nix' and `<unknown-file>'.
```

This PR fixes this by using `setOptionDefault` instead of just setting it to `null`

**Workaround**

Use `lib.mkForce` when setting the format in your config.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/git.nix
+++ b/modules/programs/git.nix
@@ -496,7 +496,7 @@ in {
           format = if (versionOlder config.home.stateVersion "25.05") then
             (mkOptionDefault "openpgp")
           else
-            null;
+            (mkOptionDefault null);
           signer = let
             defaultSigners = {
               openpgp = getExe config.programs.gpg.package;


khaneliman (approved):

khaneliman:
Created https://github.com/nix-community/home-manager/pull/6482 to resolve CI blockage

Issue #6414, opened 2025-02-07
bug: `fish.functions.*.onEvent` should accept a list of strings


mrjones2014:
### Are you following the right branch?

- [x] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Issue description

`programs.fish.functions.*.onEvent` currently only accepts a single event. In Fish shell you can do multiple events by passing multiple instances of `--on-event event`.

```fish
function my_function --on-event event1 --on-event event2
    switch $event
        case event1
            # handle event1
        case event2
            # handle event2
    end
end
```

### Maintainer CC

@r-vdp 

### System information

```markdown
- system: `"aarch64-darwin"`
 - host os: `Darwin 24.3.0, macOS 15.3`
 - multi-user?: `yes`
 - sandbox: `no`
 - version: `nix-env (Lix, like Nix) 2.91.1
System type: aarch64-darwin
Additional system types: x86_64-darwin
Features: gc, signed-caches
System configuration file: /etc/nix/nix.conf
User configuration files: /Users/mat/.config/nix/nix.conf:/etc/xdg/nix/nix.conf
Store directory: /nix/store
State directory: /nix/var/nix
Data directory: /nix/store/1phn7b1smb3g9jqz7v39q13wlyji2msz-lix-2.91.1/share`
 - nixpkgs: `/nix/store/0p3h8sfhb6a7jwdq39w83y4h4p2rqgph-source`
```

PR #6489, opened 2025-02-18
fish: accept multiple events

isabelroses:
### Description

closes https://github.com/nix-community/home-manager/issues/6414

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -66,7 +66,7 @@ let
       };
 
       onEvent = mkOption {
-        type = with types; nullOr str;
+        type = with types; nullOr (either str (listOf str));
         default = null;
         description = ''
           Tells fish to run this function when the specified named event is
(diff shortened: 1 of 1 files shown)


modules/programs/fish.nix
```diff
@@ -511,7 +511,10 @@ in {
 
             mods = with def;
               modifierStr "description" description ++ modifierStr "wraps" wraps
-              ++ modifierStr "on-event" onEvent
+              ++ (if lib.isList onEvent then
+                lib.concatMap (modifierStrs "on-event") onEvent
+              else
+                modifierStr "on-event" onEvent)
```
khaneliman, line 517: I believe we could simplify this by just calling `toList` on `onEvent`

isabelroses: Your right. Thanks. ❤️

khaneliman (approved):

khaneliman:
```bash
Expected home-files/.config/fish/functions/func-event.fish to be same as /nix/store/1fw3734xf1vgcfxpyn53f94x5gg551cg-func-event.fish but were different:
--- actual
+++ expected
@@ -1,3 +1,3 @@
-function func-event --on-event=fish_command_not_found
+function func-event --on-event="fish_command_not_found"
     echo "Not found!"
 end

isabelroses:
> ```shell
> Expected home-files/.config/fish/functions/func-event.fish to be same as /nix/store/1fw3734xf1vgcfxpyn53f94x5gg551cg-func-event.fish but were different:
> --- actual
> +++ expected
> @@ -1,3 +1,3 @@
> -function func-event --on-event=fish_command_not_found
> +function func-event --on-event="fish_command_not_found"
>      echo "Not found!"
>  end
> ```

Oops, small typo.

PR #6490, opened 2025-02-18
firefox: prevent extensions settings override

brckd:
### Description

Prevents extensions settings from accidentally being overriden when using `profiles.<name>.extensions.settings`. Adds the `profiles.<name>.extensions.force` option to acknowledge the risk. See https://github.com/nix-community/home-manager/pull/6389#issuecomment-2666634695.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@khaneliman @rycee @brckd

### DIFF
--- a/modules/programs/firefox/mkFirefoxModule.nix
+++ b/modules/programs/firefox/mkFirefoxModule.nix
@@ -703,6 +703,17 @@ in {
                   '';
                 };
 
+                force = mkOption {
+                  description = ''
+                    Whether to override all previous firefox settings.
+
+                    This is required when using `settings`.
+                  '';
+                  default = false;
+                  example = true;
+                  type = types.bool;
+                };
+
                 settings = mkOption {
                   default = { };
                   example = literalExpression ''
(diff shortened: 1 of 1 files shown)


modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -829,7 +840,22 @@ in {
...
+            lib.showAttrPath
+            (modulePath ++ [ "profiles" profileName "extensions" "settings" ])
+          }'
+          will override all previous extensions settings.
+          '${
+            lib.showAttrPath
+            (modulePath ++ [ "profiles" profileName "extensions" "force" ])
+          }'
+          needs to be enabled to acknowledge this.
+        '';
```
khaneliman, line 857: Formatting of message looks a little weird. 
image

brckd: I wasn't sure how to format it. Should it all just be on one line?

khaneliman: At least each sentence as a line, probably. That second `needs to be enabled to acknowledge this` can wrap with the option

khaneliman (approved):

PR #6491, opened 2025-02-18
files: support overwriting backup files

khaneliman:
Files that have backups already will prevent activation, add option to support overwriting files so users don't have to constantly clear conflicts if they don't care. This can be helpful with configuration files that get overwritten by programs, such as firefox extension configurations.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@teto @rycee

### DIFF
--- a/modules/files/check-link-targets.sh
+++ b/modules/files/check-link-targets.sh
@@ -33,7 +33,7 @@ for sourcePath in "$@" ; do
     elif [[ ! -L "$targetPath" && -n "$HOME_MANAGER_BACKUP_EXT" ]] ; then
       # Next, try to move the file to a backup location if configured and possible
       backup="$targetPath.$HOME_MANAGER_BACKUP_EXT"
-      if [[ -e "$backup" ]]; then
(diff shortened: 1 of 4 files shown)


teto:
I've mentioned this on previous related issues but I dont support overwriting backup files (who overwrites backups ?) and prefer to keep the code in this area limited. Are you aware you can already force the overwrite for specified files see https://github.com/nix-community/home-manager/pull/6243 ? 
Now If you are motivated to deal with this part of HM and potential losses from angry users, go ahead ;)

khaneliman:
> I've mentioned this on previous related issues but I dont support overwriting backup files (who overwrites backups ?) and prefer to keep the code in this area limited. 

It's an opt in feature though, default behavior is to stop activation. 

> Are you aware you can already force the overwrite for specified files see #6243 ? Now If you are motivated to deal with this part of HM and potential losses from angry users, go ahead ;)

Isn't that for your own custom files? The conflicts I have are auto generated from modules

khaneliman:
Actually, I did miss that the option I was having the issue with supports `force`, as well.. So, my one usecase is invalid :P

Issue #6268, opened 2025-01-04
`home.preferXdgDirectories = true;` should be stronger for home-manager / nix itself


fxzzi:
### Description
I like keeping my ~ folder tidy and clean. After using home-manager I noticed that it kept making the `~/.icons` directory when I had icon themes and cursor theme options enabled. However this directory should not be used. home-manager even knows this and puts the same stuff in `~/.local/share/icons` which is the correct place to put it.

With the option `home.preferXdgDirectories = true;`, I was expecting home-manager to stop creating these erroneous folders which don't follow the xdg base spec. However it's still created.

How could I stop the `~/.icons` directory from being created by home-manager? And on that note what about `~/.nix-profile` and `.nix-defexpr`?

thanks!

teto:
the last two depend on your nix config `use-xdg-base-directories` (man nix.conf). 

As for the first point. A quick grep returned `modules/config/home-cursor.nix` if you want to hack on it.

fxzzi:
I would say #6492 addresses this issue and allows me to stop the generation of `~/.icons` which I wanted. I'll close this issue now.

PR #6492, opened 2025-02-18
home-cursor: modernize

isabelroses:
### Description

Added the `.enable` pattern to `home.pointerCursor` for the selfish reason that is https://github.com/catppuccin/nix/blob/24dac16e6babd41961d985eef3dce6a30dab2e91/modules/home-manager/cursors.nix#L26-L27.

And made the `.icons` directory "toggleable" via `home.pointerCursor.doticons.enable`. Which in my opinion is easier than writing:

```nix 
{
  lib,
  config,
  ...
}:
{
  home.file = lib.mkIf (config.home.pointerCursor != null) {
    ".icons/default/index.theme".enable = false;
    ".icons/${config.home.pointerCursor.name}".enable = false;
  };
}
```

partially addresses https://github.com/nix-community/home-manager/issues/6268

And removing the top-level `with lib`

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/config/home-cursor.nix
+++ b/modules/config/home-cursor.nix
@@ -6,15 +6,7 @@ let
     literalExpression escapeShellArg hm getAttrFromPath any optional;
 
   cfg = config.home.pointerCursor;
-
-  enable = if (lib.versionOlder config.home.stateVersion "25.05") then
-  # respect .enable if it is declared
-    if (any (x: x ? enable) options.home.pointerCursor.definitions) then
-      cfg.enable
-    else
-      cfg != null
-  else
-    (cfg ? enable) && cfg.enable;
+  opts = options.home.pointerCursor;
 
   pointerCursorModule = types.submodule {
     options = {
@@ -132,10 +124,7 @@ in {
   options = {
     home.pointerCursor = mkOption {
       type = types.nullOr pointerCursorModule;
-      default = if (lib.versionOlder config.home.stateVersion "25.05") then
-        null
-      else
-        { };
+      default = null;
       description = ''
         Cursor configuration.
 
@@ -156,7 +145,13 @@ in {
     };
   };
 
-  config = mkMerge [
+  config = let
+    # Check if enable option was explicitly defined by the user
+    enableDefined = any (x: x ? enable) opts.definitions;
+
+    # Determine if cursor configuration should be enabled
+    enable = if enableDefined then cfg.enable else cfg != null;
+  in mkMerge [
     (mkIf enable (mkMerge [
       {
         assertions = [
@@ -242,12 +237,12 @@ in {
         ([ "xsession" "pointerCursor" ] ++ [ x ] ++ [ "isDefined" ])
         options) [ "package" "name" "size" "defaultCursor" ]) ''
           The option `xsession.pointerCursor` has been merged into `home.pointerCursor` and will be removed
-          in the future. Please change to set `home.pointerCursor` directly and enable `home.pointerCursor.x10.enable`
-          to generate x10 specific cursor configurations. You can refer to the documentation for more details.
-        '') ++ (optional ((lib.versionAtLeast config.home.stateVersion "25.05")
-          && (cfg == null)) ''
+          in the future. Please change to set `home.pointerCursor` directly and enable `home.pointerCursor.x11.enable`
+          to generate x11 specific cursor configurations. You can refer to the documentation for more details.
+        '') ++ (optional (opts.highestPrio != (lib.mkOptionDefault { }).priority
+          && cfg == null) ''
             Setting home.pointerCursor to null is deprecated.
-            Please update your configuration so that
+            Please update your configuration to explicitly set:
 
               home.pointerCursor.enable = false;
           '');
--- a/modules/config/home-cursor.nix
+++ b/modules/config/home-cursor.nix
@@ -7,8 +7,19 @@ let
 
   cfg = config.home.pointerCursor;
 
+  enable = if (lib.versionOlder config.home.stateVersion "25.05") then
+  # respect .enable if it is declared
+    if (any (x: x ? enable) options.home.pointerCursor.definitions) then
+      cfg.enable
+    else
+      cfg != null
+  else
+    (cfg ? enable) && cfg.enable;
+
   pointerCursorModule = types.submodule {
     options = {
+      enable = mkEnableOption "cursor config generation";
+
       package = mkOption {
         type = types.package;
         example = literalExpression "pkgs.vanilla-dmz";
@@ -65,8 +76,7 @@ let
     };
   };
 
-  cursorPath =
-    "${cfg.package}/share/icons/${escapeShellArg cfg.name}/cursors/${
+  cursorPath = "${cfg.package}/share/icons/${escapeShellArg cfg.name}/cursors/${
       escapeShellArg cfg.x11.defaultCursor
     }";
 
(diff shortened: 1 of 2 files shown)


modules/config/home-cursor.nix
```diff
@@ -139,10 +156,11 @@ in {
     };
   };
 
-  config = mkIf (cfg != null) (mkMerge [
+  config = mkIf cfg.enable (mkMerge [
```
khaneliman, line 159: This is technically breaking existing configurations, I would think. We used to enable configuration soon as someone provided a config. If we gate this behind the enable option users will suddenly not have pointerCursor configurations applied without warning. We might need to change the warning to also apply to `cfg.enable` not being set, not just `cfg == null`.

isabelroses: Perhaps `home.pointerCursor.enable = lib.mkDefault (cfg != null);` above the warnings code. Or is that too cursed?

isabelroses: Also just considered using `apply` but then you wouldn't get the warning.

isabelroses: Okay. Final idea. `apply = conf: if lib.isAttrs conf then conf else { enable = conf != null; _warn_null = true; }`

ambroisie: Why not gate behind the `stateVersion` (enabled under latest released stable, disabled on unstable and up)? Seems like the easiest solution.

isabelroses: Do you mind giving an example. I can only think of using `stateVersion` in ways that are still breaking.

Weathercold:
```nix
home.pointerCursor = mkOption {
  # Not sure if you can do this
  type = if lib.versionAtLeast config.home.stateVersion "25.05" then
    pointerCursorModule
  else
    types.nullOr pointerCursorModule;
...
config = mkIf
  (
    if lib.versionAtLeast config.home.stateVersion "25.05" then
      cfg.enable
    else
      cfg != null
  )
  (mkMerge [
```

isabelroses: That looks like it will work. But won't that still break peoples configs on main (not sure if that's important)?

isabelroses: Seems like that will work ❤️

rycee:
See https://github.com/nix-community/home-manager/commit/797fbbf826f57165df27911d0cd1413bcb801dfc for a prior example of introducing an `enable` option.

modules/config/home-cursor.nix
```diff
@@ -46,6 +50,14 @@ let
         '';
       };
 
+      doticons = {
```
rycee, line 53:
```suggestion
      dotIcons = {
```

modules/config/home-cursor.nix
```diff
@@ -46,6 +50,14 @@ let
...
+        enable = mkEnableOption ''
+          doticons config generation for {option}`home.pointerCursor`
```
rycee, line 55:
```suggestion
          `.icons` config generation for {option}`home.pointerCursor`
```

modules/config/home-cursor.nix
ambroisie: The `type` shouldn't depend on the `stateVersion`, but yeah my point was that it could be used to gently transition from the old module option defaults to the new one.

modules/config/home-cursor.nix
```diff
@@ -104,23 +116,31 @@ in {
...
+      warnings = (optional (any (x:
         getAttrFromPath
         ([ "xsession" "pointerCursor" ] ++ [ x ] ++ [ "isDefined" ])
         options) [ "package" "name" "size" "defaultCursor" ]) ''
           The option `xsession.pointerCursor` has been merged into `home.pointerCursor` and will be removed
           in the future. Please change to set `home.pointerCursor` directly and enable `home.pointerCursor.x11.enable`
           to generate x11 specific cursor configurations. You can refer to the documentation for more details.
-        '';
+        '') ++ (optional (cfg == null) ''
+          Setting home.pointerCursor to null is deprecated.
```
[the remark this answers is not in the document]
ambroisie: Warnings are not breaking changes, the opposite in fact.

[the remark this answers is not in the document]
Weathercold: No, what I mean is that, first of all, every user on `stateVersion < 25.05` who doesn't use `home.pointerCursor` at all would still get the warning; second of all, in order to get rid of the error, they'd need to update `stateVersion`, which is less than desirable

modules/config/home-cursor.nix
Weathercold: > The `type` shouldn't depend on the `stateVersion`

How would you implement the transition?

modules/config/home-cursor.nix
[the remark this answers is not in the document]
isabelroses: Moving the warning down into the `config` section rather then the `imports` and also check `stateVersion` to avoid that.

khaneliman:
Sorry, I've been slow to respond because I've been thinking through the silent `stateVersion` changes you've been doing vs the louder warnings to inform about the deprecations I would have expected. Originally, I expected us to add `test.assert.warnings.expected` to the legacy test and not need them in the modern test case. 

But, maybe this is a good approach to not confuse / burden users with internal details and only new configurations have to deal with the new schema... just worried about tech debt longterm if we have to gate every change and keep it backwards compatible like that.

isabelroses:
>  just worried about tech debt longterm if we have to gate every change and keep it backwards compatible like that.

The tech debt is great in my opinion too. See https://github.com/catppuccin/nix/blob/b1ff2a638afa827f1473498190a2c1cae1cf41cf/modules/home-manager/swaylock.nix#L13-L26. I see the point of `stateVersion` but I am not a fan and would prefer to break configs, but I also see not many people have time to fix these things. I would prefer a middle ground but submodules make it very difficult. Perhaps we can remove the submodule somehow.

ambroisie:
> Originally, I expected us to add test.assert.warnings.expected to the legacy test and not need them in the modern test case

:+1: this would be my expectation, and the easiest way to on-board users onto the new options.

---

> just worried about tech debt longterm if we have to gate every change and keep it backwards compatible like that.

> The tech debt is great in my opinion too

> [I] would prefer to break configs

You can (and you should), much the same NixOS does it: support the legacy option for one release, while deprecating it loudly, and removing it at the next branch-off. Given that home manager follows NixOS' release schedule and `stateVersion`s, it should be very familiar to users.

khaneliman:
Yes, sorry if I wasn't clear. I meant I'd like a loud deprecation with a stable release support. But, not gate everything silently forever with `stateVersion` checks, making the code base full of crazy conditions that we can't properly remove. 

I think the main thing we want here is a warning when we detect an option default priority of null for the `enable` option while they have a configuration that isn't empty. That way, users who have the current behavior of enabling the module via configuration would be notified to either enable or disable explicitly. 

Something like this? Not sure if we need to even check that closely or we can just check for null. 

```nix
options.enable.highestPrio == (lib.mkOptionDefault null).priority
```

isabelroses:
> Something like this? Not sure if we need to even check that closely or we can just check for null.
> 
> ```nix
> options.enable.highestPrio == (lib.mkOptionDefault null).priority
> ```

Since its a submodule you cannot access outside of the submodule the `options.enable` or at least to my knowledge.

tests/modules/config/home-cursor/default.nix
```diff
@@ -1,8 +1,24 @@
 {
   # Ensure backwards compatibility with existing configs
-  home-cursor-legacy = { realPkgs, ... }: {
+  home-cursor-legacy = { ... }: {
```
khaneliman, line 3: this was supposed to be the test that works with setting a config without enable and threw a warning that the behavior was deprecated

khaneliman:
Alright, sorry about delay. Kept forgetting to come back and do the loud deprecation instead of all the state checks. I think this will be easier to transition away from.

PR #6494, opened 2025-02-19
vinegar: add module

HeitorAugustoLN:
### Description

Adds a module for [vinegar](https://vinegarhq.org/Home/index.html)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/vinegar.nix
+++ b/modules/programs/vinegar.nix
@@ -1,0 +1,50 @@
+{ config, lib, pkgs, ... }:
+let toml = pkgs.formats.toml { };
+in {
+  meta.maintainers = with lib.maintainers; [ HeitorAugustoLN ];
+
+  options.programs.vinegar = {
+    enable = lib.mkEnableOption "Vinegar";
+
+    package = lib.mkPackageOption pkgs "vinegar" { };
+
+    settings = lib.mkOption {
+      type = lib.types.attrsOf toml.type;
+      default = { };
+      example = {
+        env.WINEFSYNC = "1";
+
+        studio = {
+          dxvk = false;
+          renderer = "Vulkan";
+
+          fflags.DFIntTaskSchedulerTargetFps = 144;
+
+          env = {
+            DXVK_HUD = "0";
+            MANGOHUD = "1";
+          };
+        };
+      };
+      description = ''
+        Configuration written to {file}`$XDG_CONFIG_HOME/vinegar/config.toml`.
+
+        See <https://vinegarhq.org/Configuration/> for more information.
+      '';
+    };
+  };
+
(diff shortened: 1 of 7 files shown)


sumnerevans (approved):
Should we provide a systemd service along with this module?

HeitorAugustoLN:
> Should we provide a systemd service along with this module?

For what specifically? I don't remember vinegar needing a service

sumnerevans:
I don't know anything about vinegar in particular, which is why I was asking the question. If no, that's fine, just wanted to make sure.

HeitorAugustoLN:
> I don't know anything about vinegar in particular, which is why I was asking the question. If no, that's fine, just wanted to make sure.

Oh, got it. It is just a launcher for Roblox Studio on Linux

modules/programs/vinegar.nix
```diff
@@ -0,0 +1,45 @@
...
+        Configuration written to {file}`$XDG_CONFIG_HOME/vinegar/config.toml`.
+
+        See <https://vinegarhq.org/Configuration/> for more information.
+      '';
+    };
+  };
+
+  config = let cfg = config.programs.vinegar;
+  in lib.mkIf cfg.enable {
+    home.packages = [ cfg.package ];
```
khaneliman, line 39:
```suggestion
    assertions =
      [ (hm.assertions.assertPlatform "programs.vinegar" pkgs platforms.linux) ];

    home.packages = [ cfg.package ];
```

Might as well make it obvious it isn't supported outside of linux.

tests/modules/programs/vinegar/default.nix
```diff
@@ -0,0 +1,4 @@
+{
+  vinegar-empty-settings = ./empty-settings.nix;
```
khaneliman, line 2:
```suggestion
```

I don't know if we really gain anything from an empty test here.

tests/modules/programs/vinegar/empty-settings.nix
```diff
@@ -0,0 +1,7 @@
+{
+  programs.vinegar.enable = true;
+
+  nmt.script = ''
+    assertPathNotExists home-files/.config/vinegar/config.toml
+  '';
+}
```
khaneliman, line 7:
```suggestion
```

tests/modules/programs/vinegar/default.nix
HeitorAugustoLN: It is just to ensure that the `config.toml` is not being created when no settings is provided. Many other modules have this exact same test

khaneliman: Yeah, I understand that. Just trying to keep the CI from ballooning for little gain.

PR #6502, opened 2025-02-19
Fix jetbrains-remote - do not fail if files do not exist yet

truelecter:
### Description

If you use `programs.jetbrains-remote` for the first time, `rm` will fail, as there are nothing to delete yet. I've tested this on my machine and it fixed an issue with the module

### Checklist

<details>
  <summary>There is an issue with other module when running tests the easy way</summary>
image

</details>

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@GenericNerdyUsername

### DIFF
--- a/modules/programs/jetbrains-remote.nix
+++ b/modules/programs/jetbrains-remote.nix
@@ -31,7 +31,7 @@ in {
         "${ide}/bin/${ide.meta.mainProgram}-remote-dev-server registerBackendLocationForGateway || true";
       lines = map mkLine cfg.ides;
       linesStr = ''
-        rm $HOME/.cache/JetBrains/RemoteDev/userProvidedDist/_nix_store*
+        rm $HOME/.cache/JetBrains/RemoteDev/userProvidedDist/_nix_store* || true
       '' + concatStringsSep "\n" lines;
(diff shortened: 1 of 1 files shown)


modules/programs/jetbrains-remote.nix
```diff
@@ -31,7 +31,7 @@ in {
         "${ide}/bin/${ide.meta.mainProgram}-remote-dev-server registerBackendLocationForGateway || true";
       lines = map mkLine cfg.ides;
       linesStr = ''
-        rm $HOME/.cache/JetBrains/RemoteDev/userProvidedDist/_nix_store*
+        rm $HOME/.cache/JetBrains/RemoteDev/userProvidedDist/_nix_store* || true
```
khaneliman, line 34: Wondering if we should do `rm -f` instead, it's specifically for that purpose and might allow other errors to show through.

truelecter: `rm -f` returns non-zero exit code if it fails to delete a file. That was my initial idea. We can use it to silence an error that comes from rm command though, but otherwise activation script will fail without `|| true`

khaneliman: Alright, fair enough. Thanks

Issue #6467, opened 2025-02-16
bug: librewolf/firefox conflict


pshirshov:
### Are you following the right branch?

- [x] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Issue description

I have both firefox and librewolf enabled. Today I've started getting these errors while trying to build:

```
       error: The option `home-manager.users.pavel.home.file.".mozilla/native-messaging-hosts".source' has conflicting definition values:
       - In `/nix/store/q25mjcgjyxi94s00k0ga58kr8mgf7k36-source/modules/programs/firefox.nix': "/nix/store/35690bcv4943cc6x4a179qp2lilfh9ki-ff_native-messaging-hosts/lib/mozilla/native-messaging-hosts"
       - In `/nix/store/q25mjcgjyxi94s00k0ga58kr8mgf7k36-source/modules/programs/librewolf.nix': "/nix/store/my71xvg12m1kqmssh1l6n71jxj0glyq0-ff_native-messaging-hosts/lib/mozilla/native-messaging-hosts"
       Use `lib.mkForce value` or `lib.mkDefault value` to change the priority on any of these definitions.
```

It seems to be a regression, am I correct?

### System information

```markdown
❯ nix-shell -p nix-info --run "nix-info -m"
 - system: `"x86_64-linux"`
 - host os: `Linux 6.12.10, NixOS, 25.05 (Warbler), 25.05.20250216.9c88845`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.24.11`
 - nixpkgs: `/nix/store/yd3115wbicrgm18qw6h6dmmv5zywzyk1-source`
```

booxter:
You are broken by https://github.com/nix-community/home-manager/pull/6421

The fix is directionally correct (both apps use the same directory), but breaks when both apps are configured. This issue is very similar to what I'm trying to fix in https://github.com/nix-community/home-manager/pull/6460 (where I had conflict between thunderbird and firefox), but more work on top of it would be needed to allow multiple firefox derivatives at the same time.

khaneliman:
Ugh... this firefox nightmare. Yeah, we need to overhaul these to properly support multiple usages.

Mikilio:
What if we just merge the configurations in both definitions.

booxter:
[...]

PR #6509, opened 2025-02-22
tests: check that all firefox derivatives can be installed

booxter:
### Description

The issue itself was resolved with:
https://github.com/nix-community/home-manager/pull/6460

This patch only adds a test case to make sure we don't ever regress by
installing firefox and librewolf and floorp at the same time.

Closes: #6467

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/tests/default.nix
+++ b/tests/default.nix
@@ -286,6 +286,7 @@ in import nmtSrc {
     ./modules/programs/boxxy
     ./modules/programs/cavalier
     ./modules/programs/eww
+    ./modules/programs/firefox
     ./modules/programs/firefox/firefox.nix
     ./modules/programs/firefox/floorp.nix
     ./modules/programs/foot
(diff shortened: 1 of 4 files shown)


tests/modules/programs/firefox/floorp-with-firefox.nix
```diff
@@ -0,0 +1,16 @@
+{ config, lib, realPkgs, ... }:
+
+lib.mkIf config.test.enableBig {
+  programs.firefox = {
+    enable = true;
+    package = realPkgs.firefox-unwrapped;
```
khaneliman, line 6:
```suggestion
    package = null;
```
We don't actually need to include the packages to test the evaluation for this scenario, I believe.

khaneliman (commented):
This should be its own test and not tied to the `floorp` one imo.

booxter:
Latest version also includes librewolf testing (same as for other derivatives).

khaneliman (approved):
LGTM, thanks for doing that

Issue #6476, opened 2025-02-17
bug: Darwin tests are completely broken (failing to eval)


booxter:
### Are you following the right branch?

- [x] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Issue description

It can be reproduced by executing on Darwin:

```
$ nix build --reference-lock-file flake.lock ./tests#test-all
```

This can be seen in https://github.com/nix-community/home-manager/actions/runs/13360690687/job/37309875415?pr=6475

```
       error: assertion '(allDeps isBuiltByBootstrapFilesCompiler [ ((stage1Packages prevStage)) ((darwinPackages prevStage)) ((llvmLibrariesDeps prevStage)) ((llvmToolsDeps prevStage)) ((sdkPackages prevStage)) ((sdkDarwinPackages prevStage)) ])' failed
       at /nix/store/ja4c4d5w9r3jb0yawr9q9fk2y0j55fmm-source/pkgs/stdenv/darwin/default.nix:753:5:
          752|
          753|     assert allDeps isBuiltByBootstrapFilesCompiler [
             |     ^
          754|       (stage1Packages prevStage)
```

The tests were broken by https://github.com/nix-community/home-manager/pull/6404

There's some difference between Darwin and Linux stdenvs that was not considered in that PR. Darwin stdenv stage1 includes packages like `jq` and `coreutils` and `gnugrep` that I think are not part of Linux stdenv.

### Maintainer CC

@rycee @khaneliman 

### System information

```markdown
$ nix-shell -p nix-info --run "nix-info -m"
 - system: `"aarch64-darwin"`
 - host os: `Darwin 24.3.0, macOS 15.3`
 - multi-user?: `yes`
 - sandbox: `no`
 - version: `nix-env (Nix) 2.24.11`
 - channels(root): `"nixpkgs"`
 - nixpkgs: `/nix/store/wzhdvy8pvcrfc1flrnhrbrpyb0q38pp8-source`
```

PR #6517, opened 2025-02-23
mpd: Add support for darwin

DamienCassou:
-------

### Description

Make mpd service work on Darwin. This PR requires https://github.com/nix-community/home-manager/issues/6476 to be fixed before moving it forward.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/mpd.nix
+++ b/modules/services/mpd.nix
@@ -14,6 +14,9 @@ let
     ${lib.optionalString (cfg.dbFile != null) ''
       db_file             "${cfg.dbFile}"
     ''}
+    ${lib.optionalString (pkgs.stdenv.hostPlatform.isDarwin) ''
+      log_file             "${config.home.homeDirectory}/Library/Logs/mpd/log.txt"
+    ''}
     state_file          "${cfg.dataDir}/state"
     sticker_file        "${cfg.dataDir}/sticker.sql"
 
@@ -114,7 +117,7 @@ in {
           type = types.bool;
           default = false;
           description = ''
-            Enable systemd socket activation.
+            Enable systemd socket activation. This is only supported on Linux.
           '';
         };
 
@@ -155,10 +158,6 @@ in {
   ###### implementation
 
   config = mkIf cfg.enable {
-    assertions = [
-      (lib.hm.assertions.assertPlatform "services.mpd" pkgs lib.platforms.linux)
-    ];
-
     services.mpd = mkMerge [
       (mkIf (versionAtLeast config.home.stateVersion "22.11"
         && config.xdg.userDirs.enable) {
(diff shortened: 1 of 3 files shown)


modules/services/mpd.nix
```diff
@@ -214,7 +210,18 @@ in {
       Install = { WantedBy = [ "sockets.target" ]; };
     };
 
+    launchd.agents.mpd = {
```
khaneliman, line 213:
```suggestion
    launchd.agents.mpd = lib.mkIf pkgs.stdenv.hostPlatform.isDarwin {
```

khaneliman: Need a similar change for the systemd service

khaneliman:
> Make mpd service work on Darwin. This PR requires #6476 to be fixed before moving it forward.

That only affects running tests locally, we can still implement darwin changes.

modules/services/mpd.nix
DamienCassou: that's actually a question I had. Is it really important? I think a systemd unit in Darwin isn't going to have much impact on the system. The opposite is probably also true. I don't mind adding the `if` condition, I'm just curious.

DamienCassou:
> That only affects running tests locally, we can still implement darwin changes.

true. But the CI is also not running the tests:
image

modules/services/mpd.nix
khaneliman: I guess technically, we don't need to gate them within the module. But, we need to at least handle the test case to properly only expect a launchd config on darwin. 

```bash
building '/nix/store/fn0xc3v8nqwzbxavg844sfd1rxcsn5ql-nmt-report-mpd-launchd.drv'...
/nix/store/aqi31sid1n78h9afwbb6p9c57xpfh9r8-assertions.sh: line 31: /nix/store/c3vd1wdv6g6m95p17gda4byi6lbavd56-home-manager-generation/LaunchAgents/org.nix-community.home.mpd.plist: No such file or directory
tput: unknown terminal "xterm-kitty"
mpd-launchd: FAILED
Expected LaunchAgents/org.nix-community.home.mpd.plist to exist but it was not found.
```

DamienCassou: Thank you for your answer. I added the 2 `mkIf` expressions. Regarding the tests, I will wait to be able to run them either locally or on CI before doing anything.

DamienCassou:
Because no one could execute the tests on darwin, I deleted the new one from the PR. I can add it back when I can execute the tests again either locally or on CI. In the meantime, I don't want to introduce tests that have never been executed: they would just add work to the person fixing the CI.

DamienCassou:
Thank you very much for fixing the linux tests @khaneliman!

PR #6526, opened 2025-02-24
xdg: create '$XDG_STATE_HOME'

ambroisie:
### Description

Some programs (e.g: `less`) will not make use of `$XDG_STATE_HOME` if the directory does not exist (even if the variable is set).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee @teto (no `meta.maintainers`)

### DIFF
--- a/modules/misc/xdg.nix
+++ b/modules/misc/xdg.nix
@@ -143,6 +143,7 @@ in {
         (mapAttrs' (name: file: nameValuePair "${cfg.stateHome}/${name}" file)
           cfg.stateFile)
         { "${cfg.cacheHome}/.keep".text = ""; }
+        { "${cfg.stateHome}/.keep".text = ""; }
       ];
     }
   ];


modules/misc/xdg.nix
```diff
@@ -143,6 +143,7 @@ in {
         (mapAttrs' (name: file: nameValuePair "${cfg.stateHome}/${name}" file)
           cfg.stateFile)
         { "${cfg.cacheHome}/.keep".text = ""; }
+        { "${cfg.stateHome}/.keep".text = ""; }
```
dotlambda, line 146: :-1: `stateHome` doesn't have a default value:
```
error: The option `home-manager.users.rschuetz.xdg.stateHome' was accessed but has no value defined. Try setting the option.
```

dotlambda: @teto Since there seems to be no intention to implement the easy fix, can you please revert this PR?

dotlambda: Never mind, here's the fix: https://github.com/nix-community/home-manager/pull/6555

PR #6528, opened 2025-02-25
distrobox: add module

aguirre-matteo:
### Description
Added programs/distrobox.nix module. It provides the option "programs.distrobox.containers", which makes it possible to declare a list of containers to be created. Since building those containers is not possible at build time (because none container backend is available at that time), I also added a Systemd Unit to build those containers after switching the configuration.

Ready to be merged!

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/lib/maintainers.nix
+++ b/modules/lib/maintainers.nix
@@ -24,6 +24,12 @@
     github = "afresquet";
     githubId = 29437693;
   };
+  aguirre-matteo = {
+    name = "aguirre-matteo";
+    email = "aguirre.matteo.nix@gmail.com";
+    github = "aguirre-matteo";
+    githubId = 158215792;
+  };
   amesgen = {
     name = "amesgen";
     email = "amesgen@amesgen.de";
--- a/modules/lib/maintainers.nix
+++ b/modules/lib/maintainers.nix
@@ -24,12 +24,6 @@
     github = "afresquet";
     githubId = 29437693;
   };
-  aguirre-matteo = {
-    name = "aguirre-matteo";
-    email = "aguirre.matteo.nix@gmail.com";
-    github = "aguirre-matteo";
-    githubId = 158215792;
-  };
   amesgen = {
     name = "amesgen";
     email = "amesgen@amesgen.de";
--- a/modules/programs/distrobox.nix
+++ b/modules/programs/distrobox.nix
@@ -26,6 +26,8 @@ let
   getContainersConfig = set:
     (concatStringsSep "\n\n" (mapAttrsToList setToContainer set)) + "\n";
 in {
+  meta.maintainers = with lib.hm.maintainers; [ aguirre-matteo ];
+
   options.programs.distrobox = {
     enable = mkEnableOption "distrobox";
 
@@ -65,9 +67,9 @@ in {
         }
       '';
       description = ''
-        A set of containers and all its respective configurations. Each option cat be either a 
+        A set of containers and all its respective configurations. Each option cat be either a
         bool, string or a list of strings. If passed a list, the option will be repeated for each element.
-        See common-debian in the example config. All the available options for the containers can be found 
+        See common-debian in the example config. All the available options for the containers can be found
         in the distrobox-assemble documentation at <https://github.com/89luca89/distrobox/blob/main/docs/usage/distrobox-assemble.md>.
       '';
     };
@@ -98,7 +100,7 @@ in {
 
         if [[ -f $prev_hash_file ]]; then
           prev_hash=$(cat $prev_hash_file)
-        else 
+        else
           prev_hash=0
         fi
 
--- a/modules/programs/distrobox.nix
+++ b/modules/programs/distrobox.nix
@@ -1,30 +1,10 @@
 { lib, pkgs, config, ... }:
 let
-  inherit (lib)
-    types isBool isList boolToString concatStringsSep mapAttrsToList removeAttrs
-    mkIf mkEnableOption mkPackageOption mkOption literalExpression;
+  inherit (lib) generators types mkIf mkEnableOption mkPackageOption mkOption;
 
   cfg = config.programs.distrobox;
 
-  attrToString = name: value:
-    let
-      newvalue = if (isBool value) then
-        [ (boolToString value) ]
-      else if (isList value) then
-        value
-      else
-        [ value ];
-    in concatStringsSep "\n" (map (value: "${name}=${value}") newvalue);
-
-  getFlags = set: concatStringsSep "\n" (mapAttrsToList attrToString set);
-
-  setToContainer = name: set:
-    ''
-      [${name}]
-    '' + (getFlags set);
-
-  getContainersConfig = set:
-    (concatStringsSep "\n\n" (mapAttrsToList setToContainer set)) + "\n";
+  formatter = pkgs.formats.ini { listsAsDuplicateKeys = true; };
 in {
   meta.maintainers = with lib.hm.maintainers; [ aguirre-matteo ];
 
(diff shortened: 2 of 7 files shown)


aguirre-matteo:
Support for Fish and Nushell has been added. So the pull request is ready to be merged.

tests/default.nix
```diff
@@ -146,6 +146,7 @@ in import nmtSrc {
     ./modules/programs/darcs
     ./modules/programs/dircolors
     ./modules/programs/direnv
+    ./modules/programs/distrobox
```
khaneliman, line 149: should be moved to linux only section

modules/programs/distrobox.nix
```diff
@@ -0,0 +1,211 @@
...
+      '';
+      description = ''
+        A set of containers to be created.
+        The name of the container will be equal to the name of the set.
+        To see available options see <https://github.com/89luca89/distrobox/blob/main/docs/usage/distrobox-assemble.md>.
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable {
```
khaneliman, line 198:
```suggestion
  config = mkIf cfg.enable {
    assertions = [
      (lib.hm.assertions.assertPlatform "programs.distrobox" pkgs lib.platforms.linux)
    ];

```

LuNeder:
> Since building those containers is not possible at build time (because none container backend is available at that time), I also added the options "enableBashIntegration" and "enableZshIntegration", which adds some code to .bashrc and .zshrc respectively that looks for changes in the file "containers.ini" and prompts the user to build the containers.

Wouldn’t a systemd unit be a better option for that, working regardlessly of the users shell and with no need for interaction?

aguirre-matteo:
@LuNeder yes, didn't thinked about that. I'll soon make a commit changing it.

aguirre-matteo:
@LuNeder It should be a better option, but how could I prompt the user for confirmation? 

By the way, is Bash always available in NixOS? I know about the symlink in /bin/sh. I think it should, given there's no `programs.bash.enable` option.

If that is true, it would be better to write the script just for Bash, and running "{pkgs.bash}/bin/bash /path/to/script.sh" in the rest of shells.

LuNeder:
> It should be a better option, but how could I prompt the user for confirmation? Try to open a terminal periodically doesn't seem like the right solution to this problem.

Why does it need user confirmation? If the user edited the config they probably want it to get applied, right?



> By the way, is Bash always available in NixOS? I know about the symlink in /bin/sh. I think it should, given there's no `programs.bash.enable` option.
> 
> If that is true, it would be better to write the script just for Bash, and running "{pkgs.bash}/bin/bash /path/to/script.sh" in the rest of shells.

Unfortunately there doesn’t seem to be an easy way to uninstall bash, and NixOS uses it on a bunch of scripts anyway. Besides, if you do “${pkgs.bash}” then bash will be installed anyway in order to fulfill that ‘placeholder’ so it shouldn’t be a problem. Still, why’s interactivity necessary here if the user is already editing the config and running nixos-rebuild?

aguirre-matteo:
@LuNeder I thinked about implementing interactivity mainly because to make the user aware what is happening. Downloading the images and/or the packages for the container takes a lot of both time and bandwidth. Moreover, shutting down the device while the containers are being built could lead to some form of corruption.

aguirre-matteo:
@LuNeder I replaced the shell's extra init code with a Systemd Unit. Here's the code:

```
systemd.user.services.distrobox-home-manager = {
      Unit.Description = "Build the containers declared in ~/.config/distrobox/containers.ini";
      Install.WantedBy = [ "default.target" ];

      Service.ExecStart = "${pkgs.writeShellScript "distrobox-home-manager" ''
        PATH=/run/current-system/sw/bin:

        containers_file=${config.xdg.configHome}/distrobox/containers.ini
        prev_hash_file=${config.xdg.configHome}/distrobox/prev_hash
        new_hash=$(sha256sum $containers_file | cut -f 1 -d " ")

        if [[ -f prev_hash_file ]]; then
          prev_hash=$(cat $prev_hash_file)
        else 
          prev_hash=0
        fi

        if [[ ! prev_hash == new_hash ]]; then
          rm -rf /tmp/storage-run-1000/containers   # <--- These two are because in some cases Podman complains 
          rm -rf /tmp/storage-run-1000/libpod/tmp   # and says to the user that those directories should be deleted
          $HOME/.nix-profile/bin/distrobox-assemble create --file $containers_file
          echo $new_hash > $prev_hash_file
        fi
''}";
```

But it builds the containers even if the previous hash is equal to the new one. This should implemented to prevent from rebuilding those containers that have `replace = true;` when switching the config or at boot time, even if no changes were made to the file.

aguirre-matteo:
@LuNeder I've replaced the extra shell init code with a Systemd Unit as you said. I've done some tests and it seems it always activate after switching the configuration of at boot time.

LuNeder:
> @LuNeder I've replaced the extra shell init code with a Systemd Unit as you said. I've done some tests and it seems it always activate after switching the configuration of at boot time.

Cool! 

I’m not sure how oci-containers do this, but I imagine it’s somewhat similar using systemd containers. Not sure if they have the same issues your mentioned on your other comment or nah.

aguirre-matteo:
@khaneliman Ready to be merged!

modules/lib/maintainers.nix
```diff
@@ -24,6 +24,12 @@
     github = "afresquet";
     githubId = 29437693;
   };
+  aguirre-matteo = {
+    name = "aguirre-matteo";
+    email = "aguirre.matteo.nix@gmail.com";
+    github = "aguirre-matteo";
+    githubId = 158215792;
+  };
```
khaneliman, line 32: I think this would need to be removed?

modules/programs/distrobox.nix
```diff
@@ -0,0 +1,116 @@
...
+    assertions = [
+      (lib.hm.assertions.assertPlatform "programs.distrobox" pkgs
+        lib.platforms.linux)
+    ];
+
+    home.packages = [ cfg.package ];
+
+    xdg.configFile."distrobox/containers.ini".source =
+      pkgs.writeText "containers.ini" (getContainersConfig cfg.containers);
```
khaneliman, line 87: Is there a reason you aren't using a `toINI` generator?

modules/lib/maintainers.nix
aguirre-matteo: Yes

modules/programs/distrobox.nix
aguirre-matteo: Being honest, I've searched about how to do it but didn't found an specific example about INI.

aguirre-matteo: Using `toINI` would translate 
```
programs.distrobox.containers.test.init_hooks = [
    "line1"
    "line2"
  ];
```

to ...?
```
''
  init_hooks=line1
  init_hooks=line2
'';
```

aguirre-matteo: By the way, if it's not a big deal, could you please tell me where to look for documentation about formatters and other generators? Both NixOS and Home-Manager

khaneliman: Not at my computer atm but I saw different arguments available to change the behavior being used throughout home manager. Usually I just look at the source code in nixpkgs to see behavior if a simple search on noogle doesn't help

aguirre-matteo: Thanks, I'll soon test that and commit the changes.

aguirre-matteo: Done

khaneliman: Sweet, thanks!

PR #6533, opened 2025-02-26
xidlehook: set Restart=always

evenbrenden:
### Description

Set `Restart=always` for `xidlehook` service.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/xidlehook.nix
+++ b/modules/services/xidlehook.nix
@@ -153,6 +153,6 @@ in {
         ExecStart = "${script}";
       };
       Install.WantedBy = [ "graphical-session.target" ];
-    };
+    } // pkgs.lib.optionalAttrs (!cfg.once) { Restart = "always"; };
   };
 }


modules/services/xidlehook.nix
```diff
@@ -153,6 +153,7 @@ in {
         ExecStart = "${script}";
       };
       Install.WantedBy = [ "graphical-session.target" ];
+      Restart = "always";
```
khaneliman, line 156: Wondering if we should change make this different when `oneshot`

```suggestion
      Restart = if cfg.once then "on-failure" else "always";
```

khaneliman: https://github.com/systemd/systemd/pull/13754 since `always` isn't supported, but they can get a restart benefit from `on-failure`

evenbrenden: Good catch! Perhaps we should keep the existing behaviour for when `cfg.once` is set, though? `Restart=none` is the default. See updated code.

khaneliman: That's probably fine, if someone wants to add the on-failure for `oneshot` they can.

dimkauzh:
This change gives me the following error:
```
error: A definition for option `home-manager.users.dima.systemd.user.services.xidlehook.Restart' is not of type `attribute set of (boolean or signed integer or string or absolute path or list of (boolean or signed integer or string or absolute path))'. Definition values:
       - In `/nix/store/awcivv27l5b749mwr95wln6q8327gnd7-source/modules/services/xidlehook.nix': "always"
```
It worked before, until this change has been made. Any ideas how to fix this?

khaneliman:
> This change gives me the following error:
> 
> ```
> error: A definition for option `home-manager.users.dima.systemd.user.services.xidlehook.Restart' is not of type `attribute set of (boolean or signed integer or string or absolute path or list of (boolean or signed integer or string or absolute path))'. Definition values:
>        - In `/nix/store/awcivv27l5b749mwr95wln6q8327gnd7-source/modules/services/xidlehook.nix': "always"
> ```
> 
> It worked before, until this change has been made. Any ideas how to fix this?

Looks like it was merged into wrong attrset. Created a PR if you want to try it.

dimkauzh:
> > This change gives me the following error:
> > ```
> > error: A definition for option `home-manager.users.dima.systemd.user.services.xidlehook.Restart' is not of type `attribute set of (boolean or signed integer or string or absolute path or list of (boolean or signed integer or string or absolute path))'. Definition values:
> >        - In `/nix/store/awcivv27l5b749mwr95wln6q8327gnd7-source/modules/services/xidlehook.nix': "always"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It worked before, until this change has been made. Any ideas how to fix this?
> 
> Looks like it was merged into wrong attrset. Created a PR if you want to try it.

Sorry for late reply, I just saw you merged it. I'll give it a go then. Thanks!

PR #6535, opened 2025-02-26
godot: init module and lib

brckd:
### Description

This adds the `programs.godot` module with support for editor configurations and text editor themes. It's accompanied by the `godot` library module which exposes constructors for godot data types.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/lib/default.nix
+++ b/modules/lib/default.nix
@@ -15,4 +15,5 @@ rec {
   shell = import ./shell.nix { inherit lib; };
   zsh = import ./zsh.nix { inherit lib; };
   nushell = import ./nushell.nix { inherit lib; };
+  godot = import ./godot.nix { inherit lib; };
 }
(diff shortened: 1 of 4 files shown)


brckd:
Nevermind, Godot seems to override the editor settings file. Declaring text themes work, but they have to be manually selected for that reason.

awwpotato:
> Nevermind, Godot seems to override the editor settings file. Declaring text themes work, but they have to be manually selected for that reason.

Maybe you could try do something similar to https://github.com/nix-community/home-manager/pull/6940?

brckd:
> > Nevermind, Godot seems to override the editor settings file. Declaring text themes work, but they have to be manually selected for that reason.
> 
> Maybe you could try do something similar to https://github.com/nix-community/home-manager/pull/6940?

Sure, that might work, but then all Godot settings have to be set via Home Manager. I thought about creating an overlay that overrides the default settings, but couldn't them.

PR #6539, opened 2025-02-27
test.yml: parallel execution

khaneliman:
Speed up CI by evaluating our conditions in parallel.

Would require updating merge conditions to match whatever new setup is. 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,23 +9,40 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
+        test-type: [lint, docs, install-test, unit-tests]
     runs-on: ${{ matrix.os }}
     steps:
-    - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


teto:
Looks like a good idea. It would make which check failed easier to spot too.

I dont know how github credits work though. Without things like the magic cache etc, there a risk that this ends up downloading more. It's probably orthogonal  to this PR but maybe we could increase speed even further by using nix-community builders since I believe they share a nix store, contrary to github runners

khaneliman:
> Looks like a good idea. It would make which check failed easier to spot too.
> 
> I dont know how github credits work though. Without things like the magic cache etc, there a risk that this ends up downloading more. It's probably orthogonal to this PR but maybe we could increase speed even further by using nix-community builders since I believe they share a nix store, contrary to github runners

I'd also like to take advantage of parallel execution of tests, but not sure how feasible that is without changes to `nmt`

PR #6543, opened 2025-02-27
swww: add swww service module for swww-daemon

hey2022:
### Description

Adds services/swww.nix module to start swww-daemon with a systemd service.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/swww.nix
+++ b/modules/services/swww.nix
@@ -1,0 +1,37 @@
+{ config, lib, pkgs, ... }:
+let cfg = config.services.swww;
+in {
+  meta.maintainers = with lib.hm.maintainers; [ hey2022 ];
+
+  options.services.swww = {
+    enable =
+      lib.mkEnableOption "swww, a Solution to your Wayland Wallpaper Woes";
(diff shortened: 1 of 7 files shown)


modules/services/swww.nix
```diff
@@ -0,0 +1,32 @@
...
+in {
+  meta.maintainers = with lib.hm.maintainers; [ hey2022 ];
+
+  options.services.swww = {
+    enable =
+      lib.mkEnableOption "swww, a Solution to your Wayland Wallpaper Woes";
+    package = lib.mkPackageOption pkgs "swww" { };
+  };
+
+  config = lib.mkIf cfg.enable {
```
khaneliman, line 12:
```suggestion
  config = lib.mkIf cfg.enable {
        assertions = [
      (lib.hm.assertions.assertPlatform "services.swww" pkgs
        lib.platforms.linux)
    ];
```

khaneliman: Might need to run format again after, didn't test formatting.

Issue #6041, opened 2024-11-05
Sway defaults to workspace 10 on nth monitors


philipwilk:
The default workspace in the hm config is 10 - but this can be solved by setting defaultWorkspace to 1. However, this does not fix the default workspace for nth monitors also being 10 - ex, plug in laptop to an external monitor and that monitor defaults to workspace 10 instead of 2 or 3 (whatever your next unused one is)

_Originally posted by @philipwilk in https://github.com/nix-community/home-manager/issues/4636#issuecomment-1953468742_

philipwilk:
Relevant locations:
[sway keybindings option default](https://github.com/nix-community/home-manager/blob/8f6ca7855d409aeebe2a582c6fd6b6a8d0bf5661/modules/services/window-managers/i3-sway/sway.nix#L85-L94)
[sway/i3 module lib: keybindingsStr](https://github.com/nix-community/home-manager/blob/8f6ca7855d409aeebe2a582c6fd6b6a8d0bf5661/modules/services/window-managers/i3-sway/lib/functions.nix#L23-L27)

philipwilk:
> Thank you for your contribution! I marked this issue as stale due to inactivity. Please be considerate of people watching this issue and receiving notifications before commenting 'I have this issue too'. We welcome additional information that will help resolve this issue. Please read the relevant sections below before commenting.
> If you are the original author of the issue
> 
> 
> If you are not the original author of the issue
> 
> 
> Memorandum on closing issues

I am going to explode you

KoviRobi:
If you want, you could help with https://github.com/nix-community/home-manager/pull/6546 -- I have done the i3 side of things, but haven't yet touched sway

philipwilk:
Will have a look later

philipwilk:
haven't looked for what changed, but the behaviour is no opposite if at boot: internal monitor defaults to 10 and external monitor defaults to 1

michaelfranzl:
The problem with this generated configuration is that it is not consistent. Using numbers 1 to 9 switches to workspaces 1 to 9, but number 0 switches to workstation 10. Without studying the Sway configuration, this can be confusing. My first impulse was to try to type "10", but that works only by accident, because Sway will first switch to workspace 1, then to workspace 10.

The upstream Sway configuration is different because the default workspace is 1, not 10. I'd love to get the upstream Sway behavior.

jadewilk:
not stale, literally saw it happen live at nixcon lol

[...]

PR #6546, opened 2025-02-27
i3-sway: Allow providing order of other workspaces

KoviRobi:
I'm using i3, not sway.

For multi-monitor setups, I end up the second screen having workspace 10 (because it is bound to Mod4+0, so ordered before the other numbers).

Needs more testing, especially with sway/tweaking to get it to work with sway. But posting it here because it might be useful.

CC #6041

### Checklist

- [ ] Change is backwards compatible -- I tried to make it so by making `defaultWorkspace` have a priority of 100. But strictly speaking this isn't backwards compatible, because it makes workspace number 1 the default, not number 10. (See tests).

  This currently moves the workspace stuff next to eachother, because I thought it's neater, but it isn't necessary.

  Currently the sway tests are broken, because I haven't fixed sway up

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@sumnerevans

### DIFF
--- a/modules/services/window-managers/i3-sway/i3.nix
+++ b/modules/services/window-managers/i3-sway/i3.nix
@@ -19,7 +19,14 @@ let
         terminal defaultWorkspace workspaceOutputAssign;
 
       keybindings = mkOption {
-        type = types.attrsOf (types.nullOr types.str);
+        type = let
+          withPriority = types.submodule {
+            options = {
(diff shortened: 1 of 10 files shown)


teto:
you can assign workspaces to monitors, can't you ? or prefix them with 1/2/3 and choose to hide the numbers

PR #6547, opened 2025-02-27
nix-darwin: use correct username in activation script

toodeluna:
### Description

This pull request updates the nix-darwin activation script to use the correct username. This was previously not the case if the username didn't match the key in the users attribute set, e.g.:

```nix
{
  users.users.myUser = {
    name = "actual_name";
  };
}
```

In this case home-manager would try to run the activation script for "myUser" when it should run for "actual_name".

There's another pull-request that fixes this issue (#5881) but it got marked as stale.

### Checklist

### DIFF
--- a/nix-darwin/default.nix
+++ b/nix-darwin/default.nix
@@ -14,9 +14,9 @@ in {
     (mkIf (cfg.users != { }) {
       system.activationScripts.postActivation.text = concatStringsSep "\n"
         (mapAttrsToList (username: usercfg: ''
-          echo Activating home-manager configuration for ${username}
-          sudo -u ${username} --set-home ${
-            pkgs.writeShellScript "activation-${username}" ''
+          echo Activating home-manager configuration for ${usercfg.home.username}
+          sudo -u ${usercfg.home.username} --set-home ${
+            pkgs.writeShellScript "activation-${usercfg.home.username}" ''
               ${lib.optionalString (cfg.backupFileExtension != null)
               "export HOME_MANAGER_BACKUP_EXT=${
                 lib.escapeShellArg cfg.backupFileExtension


toodeluna:
Forgot to add this to pull request, so putting it here:

**Maintainer CC**
@toonn @rycee

toonn:
This is the exact same code change? At least cherry-pick their commit so they retain autorship, otherwise it's plagiarism.

There's also no point in opening a PR with the same exact change, that doesn't make it any less stale than commenting on the original PR. You should also address the comments on the original since those not being addressed is why it went stale.

toodeluna:
I only found out there was already a PR for this after I already made these changes myself. While making this PR I thought to check if someone else had already made the change and found the previous PR with a comment from the stale-bot which says that "If you want to pick up the work on this PR, please create a new PR and indicate that it supercedes and closes this PR". Maybe I miss-interpreted this, but this is one of my first contributions to an open source project so I wasn't sure about what the best way to deal with this was which is why I made the PR. I apologize if this wasn't the right way to deal with this situation.

If there's any comments that need to be addressed for this change I'll cherry-pick their commit and address them, but I'm not sure what these would be since the changes on the other PR seem to be approved (in which case I'll delete this PR and the original one can be merged instead).

khaneliman:
> This is the exact same code change? At least cherry-pick their commit so they retain autorship, otherwise it's plagiarism.
> 
> There's also no point in opening a PR with the same exact change, that doesn't make it any less stale than commenting on the original PR. You should also address the comments on the original since those not being addressed is why it went stale.

Yeah, I was close to rebasing that PR, to solve the conflict, a while ago... but still wasn't sure about the consensus on that change. Looked like it was a change to mirror the behavior of the nixos module, but I'm not fully aware of the ramifications of the change.

toodeluna:
> Yeah, I was close to rebasing that PR, to solve the conflict, a while ago... but still wasn't sure about the consensus on that change. Looked like it was a change to mirror the behavior of the nixos module, but I'm not fully aware of the ramifications of the change.

The main reason for this change is indeed to mirror the behavior of the nixos module in regards to what username is used for the users. It's an issue in both this repo and the main nix-darwin repo where I've also made a [PR](https://github.com/LnL7/nix-darwin/pull/1355) to fix this. The main idea is that I should be able to write this:

```nix
{
  users.users.profile = {
    name = "luna";
  };
}
```

And that should give me the same behavior as this:

```nix
{
  users.users.luna = {};
}
```

Which is the way it works in nixos. Both PRs would have to be merged for this issue to be fixed, but neither of them should affect anyone using the second method (which is what I assume most people do, I'm just weird about how I like my config to be structured).

For what it's worth, I'm currently using my own forks of home-manager and nix-darwin with these changes in them and it's working as expected. I don't think any more changes would have to be made to fix this so I'd prefer if the original PR would be merged instead of this one since it's absolutely not my intent to take credit for these changes. I just want the issue to be fixed so I can have a working config using the main home-manager repo 🙂

isabelroses:
should this be closed since it seems to have been completed in https://github.com/nix-community/home-manager/commit/c630dfa8abcc65984cc1e47fb25d4552c81dd37e

toodeluna:
> should this have been closed since it seems to have been completed in [c630dfa](https://github.com/nix-community/home-manager/commit/c630dfa8abcc65984cc1e47fb25d4552c81dd37e)

Yes, it's been completed in #5881 so this PR can be closed.

PR #6548, opened 2025-02-27
xdg: add option 'xdg.cacheFile'

mmlb:
### Description

Add xdg.cacheFile to mimic configFile, dataFile, stateFile.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/misc/xdg.nix
+++ b/modules/misc/xdg.nix
@@ -24,6 +24,15 @@ in {
   options.xdg = {
     enable = mkEnableOption "management of XDG base directories";
 
+    cacheFile = mkOption {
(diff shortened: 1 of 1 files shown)


khaneliman:
Implementation looks fine, just curious what the use case would be for deterministic cache?

mmlb:
I have ssh control master setup to go to `~/.cache/ssh/control-master` and have been using the tmpfiles feature to create the dir for a while now. But new job new machine that requires darwin so need another option. I've been using `home.file.".cache/ssh/control-master/.keep".text = ""` for a little bit but would prefer to avoid hard coding the .cache dir and it seemed like adding `cacheFile` would be the best way, even if it looks a little odd :)

mmlb:
@khaneliman are you able to merge?

PR #6549, opened 2025-02-27
granted: support fish shell

digyx:
### Description

Due to fish being a non-POSIX shell, granted provides a separate wrapper to be used by fish.  The NixOS package for granted allows a user to enable the fish wrapper if they enable it via an override.

https://github.com/NixOS/nixpkgs/blob/3fd3383906eb781676838a3c51c9212a6f803cce/pkgs/by-name/gr/granted/package.nix#L79

This PR adds a fish integration and enables the withFish override if the fish integration is enabled.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@wcarlsen


ambroisie:
I'd be more inclined to do the following:

1. Fix it upstream, by either switching `withFish` to `true` by default or removing/deprecating it entirely.
2. _Not_ merge this PR, _maybe_ add a `package` option if you want to use the newer (fixed) version of `granted` in your configuration.

digyx:
> I'd be more inclined to do the following:
> 
>     1. Fix it upstream, by either switching `withFish` to `true` by default or removing/deprecating it entirely.
> 
>     2. _Not_ merge this PR, _maybe_ add a `package` option if you want to use the newer (fixed) version of `granted` in your configuration.

I'm alright with the issue being fixed upstream instead

If that does happen, then I do believe the Fish integration would still be useful since [granted requires users to create shell aliases](https://docs.commonfate.io/granted/troubleshooting#manually-configuring-your-shell-profile).  If you don't set the alias, then it will attempt to set it for you, and it will fail due to the fish config being read-only

wcarlsen:
Nice catch @digyx and thanks for the contribution. I think @ambroisie is right that it makes sense to see if we can do something with the `withFish` flag upstream first.

wcarlsen:
So @digyx we removed the `withFish` flag upstream see https://github.com/NixOS/nixpkgs/pull/387013. Are you comfortable reflecting those changes on this PR?

wcarlsen (approved):
@digyx thanks for your contribution

utkarshgupta137:
Something about this change breaks `assume` for me. Reverting just home-manager to the parent of this fix the issue for me. For example no matter what command I run, `assume` asks me to select a profile & then nothing happens. I'm on nixdarwin with everything pointed to master.

PR #6558, opened 2025-03-01
thunderbird: add message filters option

347Online:
Add option to declare account-specific message filters.

### Checklist

    ~~Neither of these are working for me currently, even against master, happy to re-run if there's something I'm missing~~

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@d-dervishi @hakan-demirli @ethorsoe @javaes

### DIFF
--- a/modules/programs/thunderbird.nix
+++ b/modules/programs/thunderbird.nix
@@ -155,6 +155,32 @@ let
     '') prefs)}
     ${extraPrefs}
   '';
+
+  mkFilterToIniString = f:
+    if f.text == null then
+      ''
+        name="${f.name}"
+        enabled="${if f.enabled then "yes" else "no"}"
+        type="${f.type}"
+        action="${f.action}"
+      '' + optionalString (f.actionValue != null) ''
+        actionValue="${f.actionValue}"
+      '' + ''
+        condition="${f.condition}"
+      '' + optionalString (f.extraConfig != null) f.extraConfig
+    else
+      f.text;
+
+  mkFilterListToIni = filters:
+    ''
+      version="9"
+      logging="no"
+    '' + concatStrings (map (f: mkFilterToIniString f) filters);
+
+  getEmailAccountsForProfile = profileName: accounts:
+    (filter (a:
+      a.thunderbird.profiles == [ ]
+      || any (p: p == profileName) a.thunderbird.profiles) accounts);
 in {
   meta.maintainers = with hm.maintainers; [ d-dervishi jkarlson ];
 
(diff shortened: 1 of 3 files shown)


347Online:
This would not have been possible without [this PR](https://github.com/nix-community/home-manager/pull/5432) as a reference

That said, I think I've made some improvements giving the option a fleshed out type rather than a freeform attrset.

I also went a different direction on the example/test rule.

tests/modules/programs/thunderbird/thunderbird-expected-msgFilterRules.dat
```diff
@@ -0,0 +1,7 @@
+version="9"
+logging="no"
+name="Mark as Read on Archive"
+enabled="yes"
+type="128"
+action="Mark read"
+condition="ALL"
```
khaneliman, line 7: looks like this needs to be before action

347Online: Thanks for the tip. I wish I could get these to run locally so I don't have to wait around for them to finish in the CI :melting_face:

khaneliman:
> This would not have been possible without [this PR](https://github.com/nix-community/home-manager/pull/5432) as a reference
> 
> That said, I think I've made some improvements giving the option a fleshed out type rather than a freeform attrset.

I would always prefer a submodule with freeform support and use defined known options as a guide, though. Allows us to not restrict user's if we haven't configured something explicitly and reduce maintenance burden.

347Online:
You make a good point. Would extraConfig
be sufficient here?

347Online:
I went ahead and added a text option to provide a completely custom one, or extraConfig to expand on the existing options.

modules/programs/thunderbird.nix
```diff
@@ -514,6 +603,15 @@ in {
...
         };
-    }));
+    }) ++ (mapAttrsToList (name: profile:
+      let
+        emailAccountsWithFilters =
+          (filter (a: a.thunderbird.messageFilters != [ ])
+            (getEmailAccountsForProfile name enabledAccountsWithId));
+      in (builtins.listToAttrs (map (a: {
+        name =
+          "${thunderbirdConfigPath}/${name}/ImapMail/${a.id}/msgFilterRules.dat";
```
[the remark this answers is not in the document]
347Online: I believe this is what the prior PR had used and in my testing it does work
I'm not actually sure to be perfectly honest. Profiles path seems like it should be more right... 🤔

khaneliman (commented):
Curious if you make these changes if it works for you. I've been doing this locally to make it function.

PR #6567, opened 2025-03-02
Firefox: Apply global extension force setting

HPsaucii:
### Description

Upated the mkFirefoxModule to respect user's global extension configuration force option

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee @brckd @HPsaucii

### DIFF
--- a/modules/programs/firefox/mkFirefoxModule.nix
+++ b/modules/programs/firefox/mkFirefoxModule.nix
@@ -911,7 +911,7 @@ in {
           (name: settingConfig: {
             "${profilesPath}/${profile.path}/browser-extension-data/${name}/storage.js" =
               {
-                force = settingConfig.force;
+                force = settingConfig.force || profile.extensions.force;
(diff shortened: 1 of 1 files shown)


brckd:
The `profile.extensions.force` option already resets all extension settings of the previous file format. With this change it would also override extension settings of the new file format. While we could consider making these separate options, I think it makes sense like this.

HPsaucii:
> The `profile.extensions.force` option already resets all extension settings of the previous file format. With this change it would also override extension settings of the new file format. While we could consider making these separate options, I think it makes sense like this.

Well yes, since the previous file format is in an entirely different location, there is no conflict to be had, minus some basic cleaning for redundancy. The issue is that if someone has idb disabled, without this implementation you have to force it individually, which is annoying.

bricked-dev:
With this in place, is the individual `force` setting for each extension even needed anymore? Because with disabling idb, it's either all or none.

PR #6572, opened 2025-03-04
`zoxide`: update `mkOrder` to place bash configuration at end of bashrc

carschandler:
I updated my flake a few weeks back and `zoxide` has been giving me an error recently:

```
zoxide: detected a possible configuration issue.
Please ensure that zoxide is initialized right at the end of your shell configuration file (usually ~/.bashrc).

If the issue persists, consider filing an issue at:
https://github.com/ajeetdsouza/zoxide/issues

Disable this message by setting _ZO_DOCTOR=0.
```

To be fair, I haven't noticed any issues with `zoxide` apart from this error message.

The `zoxide` eval statement in my `.bashrc` is not "right at the end" of the file as mentioned in the message which isn't surprising given that it is given a `mkOrder 150`. In my own config, I increased this to `2000` (to ensure it comes after any `mkAfter`s which are `mkOrder 1500`) to ensure it happens at the end of the script. After doing so, it appears at the end of my `bashrc` and I no longer get the error message.

I'm open to suggestions on the exact value to use as `2000` may be overkill and I'm unaware of other programs which may need to be evaluated last with more urgency than `zoxide`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@marsam

### DIFF
--- a/modules/programs/zoxide.nix
+++ b/modules/programs/zoxide.nix
@@ -48,7 +48,7 @@ in {
   config = mkIf cfg.enable {
     home.packages = [ cfg.package ];
 
-    programs.bash.initExtra = mkIf cfg.enableBashIntegration (mkOrder 150 ''
+    programs.bash.initExtra = mkIf cfg.enableBashIntegration (mkOrder 2000 ''
       eval "$(${cfg.package}/bin/zoxide init bash ${cfgOptions})"
     '');
 


khaneliman:
Will be nice to help with an issue https://github.com/nix-community/home-manager/pull/6479 Currently, we are inserting into `initExtra` which isn't far enough down in the generated file.

PR #6579, opened 2025-03-07
thunderbird: set additional gmail server settings

jonocodes:
### Description

When email account is gmail in thunderbird, set additional server settings.
As of March 2025, gmail only allows ssl/oauth0 authentication, so here I set the authMethod accordingly.

Also setting 'is_gmail' to true in thunderbird settings such that labels and trash work out of the box when the email account flavor is "gmail". This brings in the gmail configs for how it handles folders, labels, and archiving messages.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/thunderbird.nix
+++ b/modules/programs/thunderbird.nix
@@ -358,7 +358,17 @@ in {
 
     accounts.email.accounts = mkOption {
       type = with types;
-        attrsOf (submodule {
+        attrsOf (submodule ({ config, ... }: {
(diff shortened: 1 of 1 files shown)


khaneliman (commented):
Very similar to what I had to do in my configuration for gmail support, will make it easier for new users and less hassle.
We should probably use mkOptionDefault though, to avoid issues if anything changes in future and someone needs to change the value of one of the options.

PR #6582, opened 2025-03-07
treewide: null package support

khaneliman:
Can generate the config without installing application through home-manager. Helpful when a package is broken (or not provided) on a specific platform through nixpkgs and needs to be installed through other means but you still can benefit from the declarative configuration. 

Skipping service only modules. 
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/aerc.nix
+++ b/modules/programs/aerc.nix
@@ -38,7 +38,7 @@ in {
 
     enable = mkEnableOption "aerc";
 
-    package = mkPackageOption pkgs "aerc" { };
+    package = mkPackageOption pkgs "aerc" { nullable = true; };
 
     extraAccounts = mkOption {
       type = sectionsOrLines;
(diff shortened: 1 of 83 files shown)


sedlund:
```
       … while evaluating the option `programs.gh.extensions':

       … while evaluating definitions from `/nix/store/frd8jxkz2mf4g9c1qqhbl7m8qxhkl6hf-source/modules/programs/gh-dash.nix':

       error: attempt to call something which is not a function but a Boolean: true
       at /nix/store/frd8jxkz2mf4g9c1qqhbl7m8qxhkl6hf-source/modules/programs/gh-dash.nix:37:30:
           36|
           37|     programs.gh.extensions = (cfg.package != null) [ cfg.package ];
           38|
🚀 ❌ [deploy] [ERROR] Failed to evaluate deployment data: Evaluation resulted in a bad exit code: Some(1)
```

PR #6592, opened 2025-03-09
screen-locker: add lockCmdEnv option

marijanp:
### Description

I use `xsecurelock` as a screen locker and it's mainly configurable through environment variables.

Instead of having to directly access the systemd service like this:
```nix
systemd.user.services.xss-lock.Service.Environment = "XSECURELOCK_PAM_SERVICE=xsecurelock";
```

I would like to add the environment through an option I called `lockCmdEnv`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@jrobsonchase @rszamszur

### DIFF
--- a/tests/modules/services/screen-locker/basic-configuration.nix
+++ b/tests/modules/services/screen-locker/basic-configuration.nix
@@ -5,6 +5,7 @@
     enable = true;
     inactiveInterval = 5;
     lockCmd = "${pkgs.i3lock}/bin/i3lock -n -c AA0000";
+    lockCmdEnv = [ "DISPLAY=:0" "XAUTHORITY=/custom/path/.Xauthority" ];
     xss-lock = { extraOptions = [ "-test" ]; };
     xautolock = {
       enable = true;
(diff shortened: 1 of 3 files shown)


tests/modules/services/screen-locker/basic-configuration.nix
```diff
@@ -5,6 +5,7 @@
     enable = true;
     inactiveInterval = 5;
     lockCmd = "${pkgs.i3lock}/bin/i3lock -n -c AA0000";
+    lockCmdEnv = [ "DISPLAY=:0" ];
```
khaneliman, line 8: Should be checking if this is generated below, if you're adding it to the test shape.

marijanp: added

khaneliman:
```bash
screen-locker-moved-options: FAILED
Expected home-files/.config/systemd/user/xss-lock.service to match Environment=DISPLAY=:0 but it did not.
For further reference please introspect /nix/store/sfdz7x6zlh4lgi65qdq9mhid6gx71xva-nmt-report-screen-locker-moved-options

**Comment by marijanp** [comment]
@khaneliman weird, running the test locally works:
```
[~/workspace/home-manager]$ nix-shell --pure tests -A run.screen-locker-basic-configuration
building '/nix/store/w2gbbizbjj3c7z3wag554wrh1sw55q6n-home-manager-generation.drv'...
building '/nix/store/kjw441wddmwm3ifhii8qnh2aq81rk1r4-nmt-report-screen-locker-basic-configuration.drv'...
tput: unknown terminal "xterm-kitty"
screen-locker-basic-configuration: OK
``

marijanp:
@khaneliman I didn't realise another test was importing the basic-configuration in that way. Should be fine now.

PR #6593, opened 2025-03-09
home-environment: add home.sessionSearchVariables, 2nd try

thiagokokada:
### Description

This PR introduces `home.sessionSearchVariables` option, that is created to be a "generic" version of `home.sessionPath` for any environment variables that is similar to PATH (e.g.: MANPATH). This allows composition of those variables between multiple modules, avoiding issues like this one:

https://github.com/nix-community/home-manager/pull/4579/files#r1364374048

This PR also reimplements `home.sessionPath` as terms of `home.sessionSearchVariables`, to reduce code duplication and show that the code is correct, and migrates the `im/fcitx5` module to use it for `QT_PLUGIN_PATH`.

Replaces: https://github.com/nix-community/home-manager/pull/4582.

### Checklist

  + It is since this change the behavior from append to prepend in `PATH`, but the impact should be small.

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/home-environment.nix
+++ b/modules/home-environment.nix
@@ -309,12 +309,33 @@ in
         ".git/safe/../../bin"
       ];
       description = ''
-        Extra directories to add to {env}`PATH`.
+        Extra directories to prepend to {env}`PATH`.
 
         These directories are added to the {env}`PATH` variable in a
         double-quoted context, so expressions like `$HOME` are
         expanded by the shell. However, since expressions like `~` or
         `*` are escaped, they will end up in the {env}`PATH`
+        verbatim.
+      '';
+    };
+
+    home.sessionSearchVariables = mkOption {
+      default = { };
+      type = with types; attrsOf (listOf str);
+      example = {
+        MANPATH = [
+          "$HOME/.npm-packages/man"
+          "\${xdg.configHome}/.local/share/man"
+        ];
+      };
+      description = ''
+        Extra directories to prepend to arbitrary PATH-like
+        environment variables (e.g.: {env}`MANPATH`). The values
+        will be concatenated by `:`.
+        These directories are added to the environment variable in a
+        double-quoted context, so expressions like `$HOME` are
+        expanded by the shell. However, since expressions like `~` or
+        `*` are escaped, they will end up in the environment
         verbatim.
       '';
     };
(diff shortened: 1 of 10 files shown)


thiagokokada:
CC @Scrumplex @ambroisie @khaneliman for review.

modules/home-environment.nix
```diff
@@ -309,7 +309,7 @@ in
         ".git/safe/../../bin"
       ];
       description = ''
-        Extra directories to add to {env}`PATH`.
+        Extra directories to be prepend to {env}`PATH`.
```
khaneliman, line 312:
```suggestion
        Extra directories to prepend to {env}`PATH`.
```

modules/home-environment.nix
```diff
@@ -319,6 +319,27 @@ in
...
+      default = { };
+      type = with types; attrsOf (listOf str);
+      example = {
+        MANPATH = [
+          "$HOME/.npm-packages/man"
+          "\${xdg.configHome}/.local/share/man"
+        ];
+      };
+      description = ''
+        Extra directories to be prepend to arbitrary PATH-like
```
khaneliman, line 332:
```suggestion
        Extra directories to prepend to arbitrary PATH-like
```

khaneliman:
Sorry, I had the other PR in a list of tabs of stuff to review and hadn't gotten back to it.

modules/home-environment.nix
thiagokokada: Done.

modules/home-environment.nix
thiagokokada: Done.

modules/misc/qt.nix
```diff
@@ -248,19 +245,14 @@ in {
...
-      '';
+      sessionSearchVariables = envVarsExtra;
     };
 
     # Apply theming also to apps started by systemd.
-    systemd.user.sessionVariables = envVars // envVarsExtra;
+    systemd.user.sessionVariables = envVars // {
+      QT_PLUGIN_PATH = lib.concatStringsSep ":" envVarsExtra.QT_PLUGIN_PATH;
+      QML2_IMPORT_PATH = lib.concatStringsSep ":" envVarsExtra.QML2_IMPORT_PATH;
+    };
```
ambroisie, line 255: Probably should have `systemd.user.sessionSearchVariables` as well.

Issue #3999, opened 2023-05-18
bug: hm-session-vars forces variables to be only set once, forcing log-out or reboot for Gnome users


Bujiraso:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

Within ` ~/nix-profile/etc/profile.d/hm-session-vars.sh` are the lines:

```
# Only source this once.
if [ -n "$__HM_SESS_VARS_SOURCED" ]; then return; fi
```

Let's call this a "reload blocker". Perhaps these were intended for performance boost but they are having a tremendously faulty effect in a Gnome graphical session.

Gnome loads the user's profile and other variables while logging them in, and these variables (including the reload blocker) are loaded. This means that after the user changes hm-session-vars in their configuration, every single terminal they open has the *old* variables. The new variables are blocked from loading by the reload blocker.

This forces the user to log-out (expensive in time and closes all programs) or to reboot entirely in my case. The reboot may be a bug in Gnome, but the fix is simple: *let's add an option to disable these lines within home-manager*.

From my vantage, the tiny amount of variables being reloaded poses absolutely zero performance impact, so I would love for these to be hot-reloaded on every terminal.

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.3.1-zen2-1-zen, Arch Linux, noversion, rolling`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.15.0`
 - nixpkgs: `/home/nicholas/.nix-defexpr/channels/nixpkgs`
```

Bujiraso:
Sure -- that's definitely a use-case worth keeping the lines available as a toggle. I don't care which way the default goes :wink: 

But otherwise, it's this old meme :laughing: 
![image](https://github.com/nix-community/home-manager/assets/1975771/7592a68b-a2ac-409b-ab59-69ea2517d886)

Bujiraso:
Actually... Quick follow-ups, this could work, too:

```
home.sessionVariablesOnce = {
      PATH = "$PATH:/my/dir;
      MANPATH = "$MANPATH:/man/path;
};
home.sessionVariables = { 
      CARGO_HOME = ''''${HOME}/.cargo'';
      EDITOR = "vim";
};
```

Results in 
```
CARGO_HOME = ''''${HOME}/.cargo'';
EDITOR = "vim";
[...]

PR #6594, opened 2025-03-09
Add sessionVariablesOnce to home-env, bash and zsh

stevenroose:
### Description

There are some [issues](https://github.com/nix-community/home-manager/issues/3999) about users that expect their environment variables to change when they change their `sessionVariables` in their config files. I am one of them.

I learned twice the hard way that they are only sourced once per session and for me this is quite unintuitive. I thought sessions always start fresh and all variables are set at the beginning always.

F.e. when opening a new terminal with zsh, the sessionVariables are not set after they are changed, but only after logging out and in entirely.

In fact, bash did not have the same behavior, making me think that maybe zsh shouldn't even have this in the first place. Happy to amend this change to drop the "only once" behavior from zsh program entirely.

As for home session vars, I think it might make sense, and I added a field explicitly for this purpose, `sessionVariablesOnce`. This solution was suggested [here](https://github.com/nix-community/home-manager/issues/3999#issuecomment-1564478072) by user @Bujiraso. 

### Checklist

Unfortunately, people expecting the `Once` behavior for their regular variables might end up with very long variables in the end, but since this is mostly used for PATH-like variables, this will not break any functionality. So IMO this is backwards compatible *enough*.

They both seem broken. They error on stuff totally unrelated to my changes.

#### Maintainer CC

@rycee (no maintainers for `programs.zsh`)

### DIFF
--- a/modules/home-environment.nix
+++ b/modules/home-environment.nix
@@ -253,12 +253,26 @@ in
       '';
     };
 
+    home.sessionVariablesOnce = mkOption {
+      default = {};
+      type = with types; lazyAttrsOf (oneOf [ str path int float ]);
+      example = { EDITOR = "emacs"; GS_OPTIONS = "-sPAPERSIZE=a4"; };
+      description = ''
+        Environment variables to set once at login.
+        See description of `sessionVariables` for more info.
+      '';
+    };
+
     home.sessionVariables = mkOption {
       default = {};
       type = with types; lazyAttrsOf (oneOf [ str path int float ]);
       example = { EDITOR = "emacs"; GS_OPTIONS = "-sPAPERSIZE=a4"; };
       description = ''
         Environment variables to always set at login.
+
+        For variables that should only be set once at the beggining of the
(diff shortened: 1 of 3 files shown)


Bujiraso:
Cool!

ambroisie:
I personally don't really like this change, as I think it should be a user expectation/documentation issue rather than introducing more complexity to the module's API. On the other hand I can understand that restarting one's session can be an annoying friction.

---

One note: this should move all variables that currently use `sessionVariables` to `sessionVariablesOnce` for backwards compatibility (and/or individually review each usage to check whether it should be set once or would be harmless if set on each shell startup).

As an example, `PATH` and other `sessionSearchVariables` should only be set once, or you could quickly accumulate duplicate entires.

teto:
this seems like an awful lot of code. Is it possible to just `unset __HM_BASH_SESS_VARS_SOURCED` via programs.zsh.initContent for instance ?

or maybe just warn the user on switch that (s)he should logout ?

PR #6596, opened 2025-03-09
home-environment: add line-break after sessionSearchVariables

thiagokokada:
### Description

PR #6593 broke activation when `sessionVariablesExtra` is used, e.g.: `services.ssh-agent` because it concatenate the strings without a line break, so the resulting `hm-session-vars.sh` file became:

```bash
export XCURSOR_PATH="/etc/profiles/per-user/thiagoko/share/icons${XCURSOR_PATH:+:}$XCURSOR_PATH"if [ -z "$SSH_AUTH_SOCK" ]; then
  export SSH_AUTH_SOCK=$XDG_RUNTIME_DIR/ssh-agent
fi
```

This commit fixes it by enforcing a line break between `sessionSearchVariables` and `sessionVariablesExtra`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/home-environment.nix
+++ b/modules/home-environment.nix
@@ -601,7 +601,7 @@ in
           (env: values: config.lib.shell.export
             env
             (config.lib.shell.prependToVar ":" env values))
-          cfg.sessionSearchVariables)
+          cfg.sessionSearchVariables) + "\n"
         + cfg.sessionVariablesExtra;
     };
 
--- a/tests/modules/home-environment/session-variables.nix
+++ b/tests/modules/home-environment/session-variables.nix
@@ -1,4 +1,4 @@
-{ config, lib, pkgs, ... }:
+{ config, pkgs, ... }:
 
 let
 
@@ -16,6 +16,7 @@ let
     export XDG_CONFIG_HOME="/home/hm-user/.config"
     export XDG_DATA_HOME="/home/hm-user/.local/share"
     export XDG_STATE_HOME="/home/hm-user/.local/state"
+
   '';
 
   darwinExpected = ''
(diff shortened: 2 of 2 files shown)


thiagokokada:
CC @khaneliman, sorry about the issue.

I tested my configuration with this commit and it is applying cleanly, and also the resulting file looks good.

khaneliman:
```bash
Expected home-path/etc/profile.d/hm-session-vars.sh to be same as /nix/store/qm4qkwr655hkvpcbj0p2m48ibzw0090m-expected but were different:
--- actual
+++ expected
@@ -8,4 +8,3 @@
 export XDG_CONFIG_HOME="/home/hm-user/.config"
 export XDG_DATA_HOME="/home/hm-user/.local/share"
 export XDG_STATE_HOME="/home/hm-user/.local/state"
-
```

Looks like we need to account for new line ending in test or only add the line ending when we see there are values in the list for the following variable.

thiagokokada:
> ```diff
> Expected home-path/etc/profile.d/hm-session-vars.sh to be same as /nix/store/qm4qkwr655hkvpcbj0p2m48ibzw0090m-expected but were different:
> --- actual
> +++ expected
> @@ -8,4 +8,3 @@
>  export XDG_CONFIG_HOME="/home/hm-user/.config"
>  export XDG_DATA_HOME="/home/hm-user/.local/share"
>  export XDG_STATE_HOME="/home/hm-user/.local/state"
> -
> ```
> 
> Looks like we need to account for new line ending in test or only add the line ending when we see there are lines in the following variable.

Fixed, I think it will be less complicated to fix the tests than try to make a conditional check here. It is not like having a blank space after the script contents would be an issue.

khaneliman:
> Fixed, I think it will be less complicated to fix the tests than try to make a conditional check here. It is not like having a blank space after the script contents would be an issue.

Yeah, that's fine. Was just outlining a different route if wanted

PR #6605, opened 2025-03-11
macos-remap-keys: add

WeetHet:
### Description

A module to remap keys using macos native tool (hidutil)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/macos-remap-keys/default.nix
+++ b/modules/services/macos-remap-keys/default.nix
@@ -1,0 +1,72 @@
+{ config, lib, pkgs, ... }:
+let
+  cfg = config.services.macos-remap-keys;
+  keytables = import ./keytables.nix { inherit lib; };
+
+  keyToHIDCode = table: key: keytables.${table}.${key};
+
+  # Note: hidutil requires HIDKeyboardModifierMapping values to be in hexadecimal
+  # format rather than decimal JSON. Using hex strings instead of numbers will
+  # crash macOS.
+  makeMapping = table: from: to:
+    ''
+      { "HIDKeyboardModifierMappingSrc": ${
+        keyToHIDCode table from
+      }, "HIDKeyboardModifierMappingDst": ${keyToHIDCode table to} }'';
+
+  makeMappingsList = table: mappings:
+    lib.mapAttrsToList (from: to: makeMapping table from to) mappings;
+
+  allMappings = (makeMappingsList "keyboard" (cfg.keyboard or { }))
+    ++ (makeMappingsList "keypad" (cfg.keypad or { }));
+
+  allMappingsString = lib.concatStringsSep ", " allMappings;
+  propertyString = ''{ "UserKeyMapping": [ ${allMappingsString} ] }'';
+in {
+  meta.maintainers = [ lib.maintainers.WeetHet ];
+
+  options.services.macos-remap-keys = {
+    enable = lib.mkEnableOption "macOS key remapping service";
+
+    keyboard = lib.mkOption {
+      type = lib.types.attrsOf lib.types.str;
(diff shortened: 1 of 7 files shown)


modules/services/macos-remap-keys/default.nix
```diff
@@ -0,0 +1,71 @@
...
+  };
+
+  config = lib.mkIf cfg.enable {
+    assertions = [{
+      assertion = (cfg.enable -> pkgs.stdenv.hostPlatform.isDarwin);
+      message = "Key remapping only supports macOS";
+    }];
+    home.activation.macosRemapKeys =
+      lib.hm.dag.entryAfter [ "writeBoundary" ] ''
+        $DRY_RUN_CMD /usr/bin/hidutil property --set '${propertyString}' &> /dev/null
```
rycee, line 59: Can use the `run` tool, see description in [`home.activation`](https://nix-community.github.io/home-manager/options.xhtml#opt-home.activation).

```suggestion
        run --silence /usr/bin/hidutil property --set '${propertyString}'
```

modules/services/macos-remap-keys/default.nix
```diff
@@ -0,0 +1,71 @@
...
+      };
+      description = "Mapping of keypad keys to remap";
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    assertions = [{
+      assertion = (cfg.enable -> pkgs.stdenv.hostPlatform.isDarwin);
+      message = "Key remapping only supports macOS";
+    }];
```
rycee, line 56:
```suggestion
    assertions = [
      (lib.hm.assertions.assertPlatform "services.macos-remap-keys" pkgs
        lib.platforms.darwin)
    ];

```

modules/services/macos-remap-keys/default.nix
```diff
@@ -0,0 +1,71 @@
...
+    home.activation.macosRemapKeys =
+      lib.hm.dag.entryAfter [ "writeBoundary" ] ''
+        $DRY_RUN_CMD /usr/bin/hidutil property --set '${propertyString}' &> /dev/null
+      '';
+    launchd.agents.remap-keys = {
```
rycee, line 61:
```suggestion

    launchd.agents.remap-keys = {
```

khaneliman:
Tested migrating from nix-darwin `remapCapsLockToEscape`, it works on home-manager activation but not on reboot for me.

WeetHet:
> Tested migrating from nix-darwin `remapCapsLockToEscape`, it works on home-manager activation but not on reboot for me. I have to re-activate my config to get it to take effect.

Weird, could you share the resulting launchd service?

khaneliman:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>KeepAlive</key>
	<dict>
		<key>SuccessfulExit</key>
		<false/>
	</dict>
	<key>Label</key>
	<string>org.nix-community.home.remap-keys</string>
	<key>ProgramArguments</key>
	<array>
		<string>/usr/bin/hidutil</string>
		<string>property</string>
		<string>--set</string>
		<string>{ "UserKeyMapping": [ { "HIDKeyboardModifierMappingSrc": 0x700000039, "HIDKeyboardModifierMappingDst": 0x700000029 } ] }</string>
	</array>
	<key>RunAtLoad</key>
	<true/>
</dict>
</plist>
```

khaneliman (approved):

PR #6622, opened 2025-03-13
Fix potential recursion between users.users and home-manager.users

Bert-Proesmans:
### Description

Changes the strategy how home-manager integrates with `users.users.<name>.packages` to pull instead of push. This side-steps the potential circular dependency when the attribute sets `home-manager.users` or `users.users` are calculated from each other.

### Checklist

- [x] Change is backwards compatible. (AFAIK, please doublecheck edge cases)

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/docs/default.nix
+++ b/docs/default.nix
@@ -54,17 +54,17 @@ let
 
   hmPath = toString ./..;
 
-  buildOptionsDocs = args@{ modules, includeModuleSystemOptions ? true, ... }:
+  buildOptionsDocs =
+    args@{ modules, includeModuleSystemOptions ? true, isNixos ? false, ... }:
     let
       options = (lib.evalModules {
         inherit modules;
         class = "homeManager";
       }).options;
     in pkgs.buildPackages.nixosOptionsDoc ({
-      options = if includeModuleSystemOptions then
-        options
-      else
-        builtins.removeAttrs options [ "_module" ];
+      options = builtins.removeAttrs options ([ ]
+        ++ (lib.optional (!includeModuleSystemOptions) "_module")
+        ++ (lib.optional (isNixos) "users"));
       transformOptions = opt:
         opt // {
           # Clean up declaration sites to not refer to the Home Manager
@@ -80,7 +80,11 @@ let
             else
               decl) opt.declarations;
         };
-    } // builtins.removeAttrs args [ "modules" "includeModuleSystemOptions" ]);
+    } // builtins.removeAttrs args [
+      "modules"
+      "includeModuleSystemOptions"
+      "isNixos"
+    ]);
 
   hmOptionsDocs = buildOptionsDocs {
     modules = import ../modules/modules.nix {
@@ -93,6 +97,7 @@ let
   nixosOptionsDocs = buildOptionsDocs {
     modules = [ ../nixos scrubbedPkgsModule dontCheckDefinitions ];
     includeModuleSystemOptions = false;
+    isNixos = true;
     variablelistId = "nixos-options";
     optionIdPrefix = "nixos-opt-";
   };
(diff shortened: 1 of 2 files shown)


khaneliman:
```bash
error: attribute 'enable' missing
       at /home/runner/work/home-manager/home-manager/nixos/common.nix:56:10:
           55|       config = mkIf
           56|         (config.enable && cfg.useUserPackages && (lib.hasAttr name cfg.users)) {
             |          ^
           57|           packages = [ cfg.users.${name}.home.path ];
```

Bert-Proesmans:
I based that if-test on the users options on unstable. looks like 24.11 does not have the "enable" option backported.
supposing we want backwards compatibility accross nixpkgs versions, is the preference to remove that check or do i also validate if 'enable' is even an option?

Bert-Proesmans:
I checked the impact of setting on users.\<name\>.packages and it doesn't look like anything could go wrong, so removing the check for backwards compatibility should be ok.
Now, this change manipulates a non-owned option set so it might be a good idea to look into future proofing. Home-manager shouldn't generally do anything to the user if that user is disabled AFAIK, and "enable" is a standardised/intuitive option (so won't be changed in the future). This is an argument to keep the check to prevent future (semi-)related breaking and prevent future unwanted side-effects.

I rebased the commits per checklist instructions. The tests do not all succeed, can't directly say if this error is related to this change (doesn't look so at first sight);

```
error: attribute 'applicationName' missing
       at /nix/store/g4ppspdl4fy7hnp4jgjl4ll03v7i08w3-source/pkgs/applications/networking/browsers/firefox/wrapper.nix:167:23:
          166|         inherit icon;
          167|         desktopName = browser.applicationName;
             |                       ^
          168|         startupNotify = true;
```

Bert-Proesmans:
The failing test has something to do with manual generation. The documentation generator instantiates `users.users.<name>` but didn't instantiate `home-manager.users.<name>`.

```bash
error: attribute '"‹name›"' missing
       at /nix/store/yrr0zqfdp8vygfbink7qrmjp24b4is3d-source/nixos/common.nix:57:24:
           56|         && cfg.useUserPackages && (lib.hasAttr name cfg.users)) {
           57|           packages = [ cfg.users.${name}.home.path ];
             |                        ^
           58|         };
```

Bert-Proesmans:
I think i finally fixed the documentation error. There are still spurious failures with firefox tests.

The fix was tracked down in [a discord thread on unofficial Nix/Nixos server](https://discord.com/channels/568306982717751326/1353109789483339960). Thank you to (discord handles) Gerg and Tobor for their investigation and constructed solution!

ToborWinner (approved):

khaneliman (approved):
Change makes sense and seems to work fine, here.

fpletz:
Fails to evaluate for me:

```
error:
         at /nix/store/zhwwayy4iqxkxk36p0n91nhl83fmz55a-source/lib/attrsets.nix:1574:11:
         1574|           head values
             |           ^
         1575|         else

         at /nix/store/zhwwayy4iqxkxk36p0n91nhl83fmz55a-source/lib/modules.nix:925:9:
          924|     in warnDeprecation opt //
          925|       { value = addErrorContext "while evaluating the option `${showOption loc}':" value;
             |         ^
          926|         inherit (res.defsFinal') highestPrio;

       error: attribute '"‹name›"' missing
       at /nix/store/nd2r72k63l504i6pkmbamvnv0j6vh6f1-source/nixos/common.nix:56:52:
           55|       config.packages = mkIf ((config.enable or true) && cfg.useUserPackages
           56|         && lib.hasAttr name cfg.users) [ cfg.users.${name}.home.path ];
             |                                                    ^
           57|     }));
```

PR #6627, opened 2025-03-14
Kubecolor oc alias and zsh completions

jennydaman:
### Description

- Add a kubecolor alias for `oc`. See https://github.com/kubecolor/kubecolor/blob/main/README.md#whats-this
- Add an option `enableZshIntegration` which fixes ZSH completions for `kubectl` (and `oc`). See https://kubecolor.github.io/setup/shells/zsh/#completions 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@ajgon

### DIFF
--- a/modules/programs/kubecolor.nix
+++ b/modules/programs/kubecolor.nix
@@ -92,6 +92,7 @@ in {
 
     home.shellAliases = lib.mkIf (cfg.enableAlias && (cfg.package != null)) {
       kubectl = lib.getExe cfg.package;
+      oc = lib.mkIf (builtins.elem pkgs.openshift config.home.packages) "env KUBECTL_COMMAND=${lib.getExe pkgs.openshift} ${lib.getExe cfg.package}";
     };
 
     programs.zsh.initContent =
--- a/modules/programs/kubecolor.nix
+++ b/modules/programs/kubecolor.nix
@@ -92,7 +92,10 @@ in {
 
     home.shellAliases = lib.mkIf (cfg.enableAlias && (cfg.package != null)) {
       kubectl = lib.getExe cfg.package;
-      oc = lib.mkIf (builtins.elem pkgs.openshift config.home.packages) "env KUBECTL_COMMAND=${lib.getExe pkgs.openshift} ${lib.getExe cfg.package}";
+      oc = lib.mkIf (builtins.elem pkgs.openshift config.home.packages)
+        "env KUBECTL_COMMAND=${lib.getExe pkgs.openshift} ${
+          lib.getExe cfg.package
+        }";
     };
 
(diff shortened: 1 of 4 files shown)


ajgon:
I think your commits got broken somehow, and you've lost the `oc` alias part :-).

jennydaman:
Whoops, you're right. I made a mistake rewriting history to run `./format`.

jennydaman:
@ajgon ptal?

modules/programs/kubecolor.nix
```diff
@@ -89,6 +92,12 @@ in {
 
     home.shellAliases = lib.mkIf (cfg.enableAlias && (cfg.package != null)) {
       kubectl = lib.getExe cfg.package;
+      oc = "env KUBECTL_COMMAND=${lib.getExe pkgs.openshift} ${
+          lib.getExe cfg.package
+        }";
```
ajgon, line 97:
```suggestion
      oc = lib.mkIf (lib.elem pkgs.openshift config.home-manager.users.ajgon.home.packages || lib.elem pkgs.openshift config.environment.systemPackages) "env KUBECTL_COMMAND=${lib.getExe pkgs.openshift} ${
          lib.getExe cfg.package
        }";
```

I would add this alias only if user has openshift installed.

tests/modules/programs/kubecolor/does-have-openshift.nix
```diff
@@ -0,0 +1,25 @@
...
+    enableAlias = true;
+  };
+  programs.zsh = {
+    enable = true;
+    package = config.lib.test.mkStubPackage {
+      name = "zsh";
+      version = "5.9";
+    };
+  };
+  home.packages = [ pkgs.openshift ];
```
jennydaman, line 19: @ajgon I added some tests for the conditional addition of the `oc` alias.

This is my first PR to nix-community. I'm not sure what the best way to test this would be. Is it bad to have `pkgs.openshift` in the test here, instead of `config.lib.test.mkStubPackage`?

ajgon: I think `mkStubPackage` will work better here, as it won't build whole derivation for openshift, just for testing. You can overload `pkgs` with overlay i.e.

```nix
  nixpkgs.overlays = [
    (self: super: rec {
      openshift = config.lib.test.mkStubPackage {
        name = "openshift";
        version = "4.16.0";
    };
  ];
```

and keep the rest of your test as is. This should use stub instead of a real package, thus avoiding building deriv.  'm not at my laptop right now, so can't guarantee if it works, but in general that's the idea you should follow :-).

ajgon (approved):

Issue #3739, opened 2023-03-04
Polkit authentication agent(s)


cyntheticfox:
### Description

I'm running into needing/wanting a graphical Polkit agent, but there's no current agents present in `home-manager`.

I've seen #2400 and am considering writing something based on that, though with the GNOME polkit as I'm running Sway, and other recommendations I've been given involved xdg-autostart (which is external state). 

Could we re-open the mentioned PR, or should I draft a new one?

I'm currently just running

```nix
systemd.user.services.sway-polkit-authentication-agent = {
  Unit = {
    Description = "Sway Polkit authentication agent";
    Documentation = "https://gitlab.freedesktop.org/polkit/polkit/";
    After = [ "graphical-session-pre.target" ];
    PartOf = [ "graphical-session.target" ];
  };

  Service = {
    ExecStart = "${pkgs.polkit_gnome}/libexec/polkit-gnome-authentication-agent-1";
    Restart = "always";
    BusName = "org.freedesktop.PolicyKit1.Authority";
  };

  Install.WantedBy = [ "graphical-session.target" ];
};
```

Not sure if that `BusName` arg is needed...

zombiehoffa:
did this go anywhere? I see it still open.

mobsenpai:
This should be added.
I am using this currently ->
```nix
{
  config,
  lib,
  pkgs,
  ...
}: {
  options = {
    myNixos.polkit-gnome.enable = lib.mkEnableOption "Enables polkit-gnome";
  };

  config = lib.mkIf config.myNixos.polkit-gnome.enable {
    security.polkit.enable = true;

    systemd.user.services.polkit-gnome-authentication-agent-1 = {
      description = "polkit-gnome-authentication-agent-1";
      wantedBy = ["graphical-session.target"];
      wants = ["graphical-session.target"];
      after = ["graphical-session.target"];
      serviceConfig = {
        Type = "simple";
        ExecStart = "${pkgs.polkit_gnome}/libexec/polkit-gnome-authentication-agent-1";
        Restart = "on-failure"; # TODO: check this
        RestartSec = 1;
        TimeoutStopSec = 10;
      };
    };
  };
}

```

mobsenpai:
> did this go anywhere? I see it still open.

no sadly.

sumnerevans:
@mobsenpai feel free to submit a PR with your changes.

mobsenpai:
[...]

PR #6631, opened 2025-03-14
{hyprpolkitagent, lxqt-policykit-agent, polkit-gnome}: init modules

bobvanderlinden:
### Description

Supersedes https://github.com/nix-community/home-manager/pull/5619
Resolves https://github.com/nix-community/home-manager/issues/3739

This adds polkit agent services for hyprland, gnome and lxqt. I have had the one from lxqt in my own configuration for a while and served me well, but recently wanted to check out other polkits and eventually switched to gnome's one.
I thought I'd add these to home-manager as well this time.

I was contemplating adding `security.polkit.agent = "gnome" or "hyprland" or "lxqt"`, but since there was no `security` directory I wasn't sure that was the right call for now and could be a future step.

All 3 of these tools do not have configuration(-files).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/hyprpolkitagent.nix
+++ b/modules/services/hyprpolkitagent.nix
@@ -1,7 +1,6 @@
 { config, lib, pkgs, ... }:
 let
-  inherit (lib)
-    mkEnableOption mkPackageOption types literalExpression mkIf maintainers;
+  inherit (lib) mkEnableOption mkPackageOption mkIf maintainers;
   cfg = config.services.hyprpolkitagent;
 in {
   meta.maintainers = [ maintainers.bobvanderlinden ];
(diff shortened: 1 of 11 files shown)


modules/services/hyprpolkitagent.nix
```diff
@@ -0,0 +1,29 @@
...
+
+  config = mkIf cfg.enable {
+    systemd.user.services.hyprpolkitagent = {
+      Unit = {
+        Description = "Hyprland PolicyKit Agent";
+        After = [ "graphical-session-pre.target" ];
+        PartOf = [ "graphical-session.target" ];
+      };
+
+      Install = { WantedBy = [ "graphical-session.target" ]; };
```
khaneliman, line 24: These should probably be updated to use the `config.wayland.systemd.target` ?

khaneliman: Yeah, didn't work for me until I made that change.

PR #6632, opened 2025-03-14
nixgl: use original package name

### LINKED ISSUE #6608: bug: config.lib.nixGL.wrap leads to missing Chromium extensions

### Issue description

There seems to have been a similar error in #6016.

Here is an example configuration:
```nix
programs.chromium = {
    enable = true;
    package = config.lib.nixGL.wrap pkgs.brave;
    extensions = [
      "nngceckbapebfimnlniiiahkandclblb" # Bitwarden Passwortmanager
      "eimadpbcbfnmbkopoojfekhnkhdbieeh" # Dark Reader
    ];
  };
```

Workaround: remove config.lib.nixGL.wrap from programs.chromium.package to get back the defined extensions.

### Maintainer CC

@Smona

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.11.0-19-generic, Linux Mint, 22.1 (Xia), nobuild`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.26.2`
 - channels(root): `"nixpkgs"`
 - nixpkgs: `/nix/store/ldwapm29wjwan94c9nqjv57jssqrs0rh-nixpkgs/nixpkgs`
```

Smona:
### Description

Some home manager modules depend on the name, so changing it in the wrapper can break them.

Fixes #6608 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/nixgl.nix
+++ b/modules/misc/nixgl.nix
@@ -217,7 +217,10 @@ in
           # Wrap the package's binaries with nixGL, while preserving the rest of
           # the outputs and derivation attributes.
           (pkg.overrideAttrs (old: {
-            name = "nixGL-${pkg.name}";
+            # Leave the name unchanged and rely on the hash to differentiate
+            # from the original package. Some modules rely on the package name
+            # to e.g. compute config directory paths.
+            name = pkg.name;
 
             # Make sure this is false for the wrapper derivation, so nix doesn't expect
             # a new debug output to be produced. We won't be producing any debug info


teto:
it's helpful to show in store path that it is a nixgl wrapper. I find it odd for a package to change its config based on its name. Maybe something to change more in the upstream package

Smona:
That's a good point, I think this specific issue could possibly be fixed by using `pname` rather than `name` in the part of the chromium module highlighted [here](https://github.com/nix-community/home-manager/issues/6608#issuecomment-2723861297). I only went with this approach first because it would prevent instances of this problem in other modules from generating issues for the nixgl module going forward. I agree though that this change makes the store path less clear, so it's less than ideal.

Smona:
I see @tadfisher authored that line, any thoughts on the alternative mentioned above?

DocBrown101:
How long does it normally take for a changed line to be reviewed?

DocBrown101:
Helllo @teto, can this bugfix be accepted or does it need to be clarified in this bugfix ticket?

rycee:
I rebased this on the latest master and force-pushed. Should be safe to merge?

Smona:
Thanks @rycee! This should still be good to merge, yes. The only disagreement was over whether it was better to keep "nixGL" in the store path and remove the usage of `p.name` for computing paths in downstream packages, or leave the `name` unchanged from the base package, making it harder to identify the nixGL derivation in the store but preventing any further issues like #6608 categorically.

I'm still in favor of fixing the issue categorically, as having the nixGL integration "just work" with existing and new modules is more important than saving a `nix eval` during debugging IMO.

rycee:
I guess it's higher priority for things to work than for easy identification.

It may be possible to collect the paths of wrapped packages into a "nixGL-wrapped-packages.json" file that is added somewhere, then at least it would be possible to indirectly find the wrapped paths? I'm inclined to merge this as-is and let anybody interested in this make a new PR.

@teto Do you have any strong opinions?

rycee:
Let's try it, if some subtle issue arises then we can rethink.

PR #6633, opened 2025-03-14
mergiraf: init module

bobvanderlinden:
### Description

This adds git integration for [mergiraf](https://mergiraf.org/). It sets up git to use mergiraf to automatically resolve merge conflicts. See https://mergiraf.org/usage.html#registration-as-a-git-merge-driver

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/mergiraf.nix
+++ b/modules/programs/mergiraf.nix
@@ -1,0 +1,31 @@
+{ pkgs, config, lib, ... }:
+let
+  inherit (lib)
+    mkEnableOption mkPackageOption types literalExpression mkIf maintainers;
+  cfg = config.programs.mergiraf;
+  mergiraf = "${cfg.package}/bin/mergiraf";
+in {
+  meta.maintainers = [ maintainers.bobvanderlinden ];
+
+  options = {
+    programs.mergiraf = {
+      enable = mkEnableOption "mergiraf";
+      package = mkPackageOption pkgs "mergiraf" { };
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    programs.git = {
+      attributes = [ "* merge=mergiraf" ];
+      extraConfig = {
+        merge.mergiraf = {
+          name = "mergiraf";
+          driver =
+            "${mergiraf} merge --git %O %A %B -s %S -x %X -y %Y -p %P -l %L";
+        };
+      };
+    };
+  };
+}
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -170,6 +170,7 @@ let
     ./programs/mbsync.nix
     ./programs/mcfly.nix
     ./programs/mercurial.nix
+    ./programs/mergiraf.nix
     ./programs/micro.nix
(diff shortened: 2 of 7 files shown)


modules/programs/mergiraf.nix
```diff
@@ -0,0 +1,35 @@
...
+          driver =
+            "${mergiraf} merge --git %O %A %B -s %S -x %X -y %Y -p %P -l %L";
+        };
+      };
+    };
+
+    xdg.configFile."git/attributes".source =
+      pkgs.runCommand "mergiraf-git-attributes" { } ''
+        ${mergiraf} languages --gitattributes > $out
+      '';
```
bobvanderlinden, line 33: This is something I'm unsure of. Right now, it'll conflict if someone has set `programs.git.attributes`. It does update all language settings automatically.

The alternative would be hard-coding the languages, but we need to keep this list up-to-date, which is less than ideal.

Another alternative is allowing `programs.git.attributes` to also include other files and concat it all. I'm unsure whether this is wanted for now.

bobvanderlinden:
Ping @wetneb

bobvanderlinden:
I'm unsure why the test for macos is failing. The package name `@mergiraf@` isn't being used and instead `/nix/store/*` is used somehow :confused:

Any suggestions?

wetneb:
Sounds very useful, thanks for working on that! It reminds me that I would like to investigate if we shouldn't instead recommend to just set `* merge=mergiraf` (without listing the languages), because the performance cost is likely acceptable.

modules/programs/mergiraf.nix
khaneliman: Definitely shouldn't be overriding the source contents of `git/attributes` from another module. We should expose an option in the git module to support appending to the generated config.

khaneliman:
> I'm unsure why the test for macos is failing. The package name `@mergiraf@` isn't being used and instead `/nix/store/*` is used somehow 😕
> 
> Any suggestions?

Just need to add the package name to the list of darwin packages to scrub in `tests/default.nix`, for now.

teto (commented):
pretty cool

modules/programs/mergiraf.nix
bobvanderlinden: I agree, that makes sense. I've now opted to just handle all file extensions. This avoids the need to write the output of mergiraf to `~/.config/git/attributes`, so this is not an issue anymore :+1:

bobvanderlinden:
> Sounds very useful, thanks for working on that! It reminds me that I would like to investigate if we shouldn't instead recommend to just set `* merge=mergiraf` (without listing the languages), because the performance cost is likely acceptable.

I gave this a try. It seems to work fine, so I'll opt for this option in home-manager :+1:

> Just need to add the package name to the list of darwin packages to scrub in `tests/default.nix`, for now.

Thanks! I've moved mergiraf test to the isLinux list.

khaneliman (approved):
LGTM, tested it and it works. Will be curious to try it out.

Issue #5244, opened 2024-04-09
Bug: Using home-manager as a NixOS module with external authentication services (Kanidm, Ldap, Active Directory) fails


TheRealGramdalf:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

I use [Kanidm](https://github.com/kanidm/kanidm) to manage POSIX accounts on my small fleet of NixOS machines. I recently attempted to start using home-manager as a NixOS module, yet quickly ran into issues - the home-manager module expects to handle POSIX accounts itself (via `users.users."therealgramdalf"`), and thus throws an error:
```
       error:
       Failed assertions:
       - Exactly one of users.users.therealgramdalf.isSystemUser and users.users.therealgramdalf.isNormalUser must be set.

       - users.users.therealgramdalf.group is unset. This used to default to
       nogroup, but this is unsafe. For example you can create a group
       for this user with:
       users.users.therealgramdalf.group = "therealgramdalf";
       users.groups.therealgramdalf = {}
```

This is an issue, since the posix accounts don't actually exist locally on the machine - they are loaded dynamically by `kanidm-unixd`. I go more into detail here: https://github.com/kanidm/kanidm/issues/2698

A method to disable this requirement (such that the NixOS module doesn't actually create the users) should solve most of the problems; the UUID problems specified in the issue I linked should be solvable at the configuration level (without altering home-manager itself).

### Maintainer CC

@rycee

### System information

```markdown
N/A
```

TheRealGramdalf:
I should mention that I have `home-manager.useUserPackages` set to true, which may be affecting things due to the usage of `users.users."therealgramdalf".packages`

PR #6640, opened 2025-03-16
Support user accounts managed by external identity management systems

tomeon:
### Description

This changeset introduces support in the NixOS module for entries in `home-manager.users` that do not have corresponding entries in `users.users`.  The goal is to support Home Manager configurations for users provided by external identity management systems, as described in #5244, without resorting to hacks like re-specifying those users' details (UIDs, etc.) in `users.users`.

### Checklist

- [x] Change is backwards compatible.  **NOTE**: this changeset alters the semantics of `home-manager.useUserPackages`: that option now provides the default value of the per-user option `home-manager.users.<name>.home.useUserPackages`, rather than globally instituting (or not) the use of `users.users.<name>.packages`.  I believe this change is backward-compatible, given that no Home Manager NixOS module configuration will have used `home-manager.users.<name>.home.useUserPackages` before now, and so existing configurations will continue to globally use (or not) `users.users.<name>.packages`.

    ```
    {component}: {description}

    {long description}
    ```

- ~If this PR adds a new module~

### DIFF
--- a/docs/release-notes/rl-2505.md
+++ b/docs/release-notes/rl-2505.md
@@ -21,6 +21,16 @@ This release has the following notable changes:
   tests are available through a Nix Flake file inside the `tests`
   directory. See [](#sec-tests) for example commands.
 
+- The Home Manager NixOS module now supports
+  [home-manager.users](#nixos-opt-home-manager.users) entries that do not have
+  corresponding `users.users.<name>` entries, making it easier to provide Home
+  Manager configurations for users managed through external identity management
+  systems.  To take advantage of this, such users must have their
+  [](#opt-home.username) and [](#opt-home.homeDirectory) attributes set to
+  high-precedence values with `lib.mkForce` or similar, overriding the defaults
+  that pull from, respectively, `users.users.<name>.name` and
+  `users.users.<name>.home`.
+
 ## State Version Changes {#sec-release-25.05-state-version-changes}
 
 The state version in this release includes the changes below. These
--- a/docs/release-notes/rl-2505.md
+++ b/docs/release-notes/rl-2505.md
@@ -31,6 +31,13 @@ This release has the following notable changes:
   that pull from, respectively, `users.users.<name>.name` and
   `users.users.<name>.home`.
 
+- The Home Manager NixOS module option
+  [home-manager.useUserPackages](#nixos-opt-home-manager.useUserPackages) now
+  sets the default value of the newly-introduced per-user
+  [](#opt-home.useUserPackages) option, making it possible to define a policy
+  on the use of `users.users.<name>.packages` for package installation, and to
+  override that policy on a user-specific basis.
+
 ## State Version Changes {#sec-release-25.05-state-version-changes}
 
 The state version in this release includes the changes below. These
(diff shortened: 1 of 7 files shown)


jkachmar:
this changeset is pretty much exactly what I’m looking for; is anyone available to review or comment on why this may not be a particularly good approach?

Guvante:
I have been testing this PR locally (after merging it with master to resolve the conflicts) and it seems to be working pretty effectively. The main downside is I didn't realize `useUserPackages` reintroduces a dependency on `users.users` (I thought it was about moving things to /etc/profile.d or something to that effect) but once I disabled that I was able to get it to work.

One issue that I am unsure is related to this or something else in my configuration (I am pretty sure this is related to my identity provider wigging out on me so I think it is unrelated) but I get the following error sometimes

> warning: the following units failed: home-manager-$USER
>× home-manager-$USER.service - Home Manager environment for $USER
>     Loaded: loaded (/etc/systemd/system/home-manager-$USER.service; enabled; preset: ignored)
>     Active: failed (Result: exit-code) since Sun 2025-09-21 09:47:53 PDT; 5s ago
> Invocation: 23b02cd65ce0453196bae3be895c7ced
>    Process: 44914 ExecStart=/nix/store/2bkpj12f380wh0dfci43s6dszikdracn-hm-setup-env /nix/store/7sjslara3vrqpg4h8vicz74xwgg0gzca-home-manager-generation (code=exited, status=217/USER)
>   Main PID: 44914 (code=exited, status=217/USER)
>         IP: 0B in, 0B out
>         IO: 0B read, 0B written
>   Mem peak: 1.5M
>        CPU: 5ms
> 
> $timestamp $host systemd[1]: Starting Home Manager environment for $USER...
> $timestamp $host (etup-env)[44914]: home-manager-$USER.service: Failed to determine user credentials: No such process
> $timestamp $host (etup-env)[44914]: home-manager-$USER.service: Failed at step USER spawning /nix/store/2bkpj12f380wh0dfci43s6dszikdracn-hm-setup-env: No such process
> $timestamp $host systemd[1]: $USER: Main process exited, code=exited, status=217/USER
> $timestamp $host systemd[1]: $USER: Failed with result 'exit-code'.
> $timestamp $host systemd[1]: Failed to start Home Manager environment for $USER.
> warning: error(s) occurred while switching to the new configuration

Oddly enough when it does occur it is transient since manually running the service works

(To be clear not looking for help or anything just figured I would share the weird thing that happened when testing)

tomeon:
@Guvante -- I have hit this same issue.  I tried adding `kanidm-unixd.service` to the `Wants` and `After` lists of `home-manager-<user>.service`, but this did not solve the problem.  I may try updating this PR to add support for delaying `home-manager-<user>.service` until `<user>` logs in.

N3xp7im3:
Any idea on how to automatically create a home directory with your patches?
The usual way is to use `users.users.<name>.createHome`, but this is something I cannot use anymore with your patches. Would be nice to have something similar.

N3xp7im3:
It does not seem to work with #6981 either. The user activation was implemented by deploying a drop-in configuration via `users.user.packages`. I tried to work around this by adding it via `home.packages`, but this causes an infinite recursion as the `home.activationPackage` cannot be used from within the home configuration.

ilyx-me:
This would be great to have merged. What is the current status of the PR, are there any blockers left?

PR #6650, opened 2025-03-17
vscode: fix creation of storage.json file

### LINKED ISSUE #6545: bug: VSCode profiles get overriden on first launch

### Issue description

With the addition of profile support with nix-community/home-manager#5640, I noticed that the first iteration of profile configs gets overriden when I launch VSCodium for the first time.
Subsequently rebuilding the profiles puts them back and VSCode doesn't delete them, leading me to think that HM is missing some marker to show that the files should stay there?

### Maintainer CC

_No response_

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.12.16, NixOS, 25.05 (Warbler), 25.05.20250224.0196c01`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Lix, like Nix) 2.91.1
System type: x86_64-linux
Additional system types: i686-linux, x86_64-v1-linux, x86_64-v2-linux, x86_64-v3-linux, x86_64-v4-linux
Features: gc, signed-caches
User configuration files: /home/toby/.config/nix/nix.conf:/etc/xdg/nix/nix.conf:/home/toby/.nix-profile/etc/xdg/nix/nix.conf:/nix/profile/etc/xdg/nix/nix.conf:/home/toby/.local/state/nix/profile/etc/xdg/nix/nix.conf:/etc/profiles/per-user/toby/etc/xdg/nix/nix.conf:/nix/var/nix/profiles/default/etc/xdg/nix/nix.conf:/run/current-system/sw/etc/xdg/nix/nix.conf
Store directory: /nix/store
State directory: /nix/var/nix
Data directory: /nix/store/zm68nasdmai8qzgzlb55m10mrl93dzqd-lix-2.91.1/share`
 - nixpkgs: `/nix/store/593xvgv994xlkm5mb7w4p1xxnzrs9wv6-source`
```

Reputable2772:
### Description

Resolves #6545 

This PR ensures that directories in the path of `~/.config/<vscodeInstance>/User/globalStorage/storage.json` get created correctly.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/vscode.nix
```diff
@@ -299,14 +299,14 @@ in {
...
+            (flatten (mapAttrsToList (n: v: n) allProfilesExceptDefault))
+          })
 
           if [ -f "$file" ]; then
             existing_profiles=$(jq '.userDataProfiles // [] | map({ (.name): .location }) | add // {}' "$file")
-            file_write=""
-            profiles=(${
-              escapeShellArgs
-              (flatten (mapAttrsToList (n: v: n) allProfilesExceptDefault))
-            })
```
khaneliman, line 309: these look like they were scoped closer to their actual usage, why were they moved up?

Reputable2772: The `else` clause from line 316 also uses those specific variables, hence they were moved up.

khaneliman: Thanks, didn't see that.

Issue #6536, opened 2025-02-26
bug: sops-nix activation in home-manager


gesker:
### Are you following the right branch?

- [x] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Issue description

There appears to be a **plist** file conflict when trying to activate sops-nix on home-manager; nix-darwin module. This issue was was mentioned under a possibly related - or at least similar - issue [#710](https://github.com/Mic92/sops-nix/issues/710) in the [sops-nix](https://github.com/Mic92/sops-nix) repository. 

It was recommended to the OP on that issue that the issue be reported directly to home-manager. Again, not 100% sure that this is the same issue. Also, I did not see the OP of that issue RE-report in this repository.

Upon issuing `nix run nix-darwin -- switch --flake /etc/nix-darwin` an error is returned shortly after reaching Home-Manager activation:

```bash
Starting Home Manager activation
Activating checkFilesChanged
Activating checkKittyTheme
Activating checkLaunchAgents
Existing file '/Users/gesker/Library/LaunchAgents/**org.nix-community.home.sops-nix.plist' is in the way of '/nix/store/x96razajz78mkr4p223bdc91l9yny65y-home-manager-agents/org.nix-community.home.sops-nix.plist'
```

I have tried completely uninstalling Nix and then reinstalling nix running home-manager only (without nix-darwin) and the same error is encountered.

### Maintainer CC

Unsure which maintainer it would be appropriate to tag.

### System information

```markdown
- system: `"aarch64-darwin"`
 - host os: `Darwin 24.3.0, macOS 15.3.1`
 - multi-user?: `yes`
 - sandbox: `no`
 - version: `nix-env (Nix) 2.24.12`
 - channels(root): `"nixpkgs"`
 - nixpkgs: `/nix/store/593xvgv994xlkm5mb7w4p1xxnzrs9wv6-source`
```

khaneliman:
Been meaning to look into it more, but it is inconsistent when it occurs for me.

gesker:
@khaneliman I was grateful to find your original issue. 

I had been trouble shooting on a different/incorrect path as I was not having this same issue on my regular NixOS systems and I'm still new enough to the nix ecosystem that I was not confident in firing off an issue report. 

[...]

PR #6656, opened 2025-03-18
launchd: refactor implementation

khaneliman:
### Description
Closes https://github.com/nix-community/home-manager/issues/6536

Rework to drop the conflict check, these are home-manager generated and controlled plists and not user data. We will drop and redeploy whenever we need. 

Updated implementation for the deploy to unload and deploy without skipping an agent if there are any runtime issues from `launchd` so we don't skip deploying agents because of an agent that might have been deleted improperly. 

Tested scenarios:

- No changes:
  - We don't try to drop/recreate services
- Agent removed
  - We unload and remove plist
- Agent changed
  - We unload, remove, install, and load agent
- Plist deleted
  - We install and load agent
  - When bootstrap fails, we continue and don't block other agent activations

NOTE: 
This should fix issues with deploying the launchd agents, but doesn't resolve another issue discovered that profile generations aren't being properly updated on darwin. So we are usually comparing launchd agents against the wrong generation resulting in incorrect comparison results.

ie: It will say we had a launchd agent that changed since last generation, but it was actually a few generations ago and shouldn't need to be dropped/redeployed. This is the cause for the conflicting file check previously.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/launchd/default.nix
+++ b/modules/launchd/default.nix
@@ -93,81 +93,170 @@ in {
       # NOTE: Launch Agent configurations can't be symlinked from the Nix store
       # because it needs to be owned by the user running it.
       home.activation.setupLaunchAgents =
-        lib.hm.dag.entryAfter [ "writeBoundary" ] ''
-          setupLaunchAgents() {
-            local oldDir newDir dstDir domain err
-            oldDir=""
-            err=0
-            if [[ -n "''${oldGenPath:-}" ]]; then
-              oldDir="$(readlink -m "$oldGenPath/LaunchAgents")" || err=$?
-              if (( err )); then
-                oldDir=""
+        lib.hm.dag.entryAfter [ "writeBoundary" ] # Bash
+        ''
+          # Disable errexit to ensure we process all agents even if some fail
+          set +e
+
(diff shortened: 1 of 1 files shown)


khaneliman:
Hmm after adding some debugging I see it still trying to stop / remove agents that were already removed...

khaneliman:
Dumping out the `oldGenPath` shows the directory we are comparing against when doing collision detection still contains agents that are not actually part of the currently running configuration.

khaneliman:
Looking at the `~/.local/state/nix/profiles` vs the `~/.local/share/home-manager/gcroots/current-home` the current home isn't pointing at the latest profile generation, like in nixos. So we're constantly comparing against an old revision instead of the current revision.

khaneliman:
This should fix issues with deploying the launchd agents, but doesn't resolve another issue discovered that profile generations aren't being properly updated on darwin. So we are usually comparing launchd agents against the wrong generation resulting in incorrect comparison results. 

ie: It will say we had a launchd agent that changed since last generation, but it was actually a few generations ago and shouldn't need to be dropped/redeployed.

PR #6664, opened 2025-03-19
treewide: zsh initExtra -> initContent

khaneliman:
Migrating in tree usages of zsh initExtra to initContent before deprecating. Will maintain similar priorities as previous options, for backwards compatibility. Maintainers can change priority, as they see fit. 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/mise.nix
+++ b/modules/programs/mise.nix
@@ -108,7 +108,7 @@ in {
         eval "$(${getExe cfg.package} activate bash)"
       '';
 
-      zsh.initExtra = mkIf cfg.enableZshIntegration ''
+      zsh.initContent = mkIf cfg.enableZshIntegration ''
         eval "$(${getExe cfg.package} activate zsh)"
       '';
 
--- a/modules/misc/vte.nix
+++ b/modules/misc/vte.nix
@@ -45,7 +45,7 @@
     })
 
     (lib.mkIf config.programs.zsh.enableVteIntegration {
-      programs.zsh.initExtra = ''
+      programs.zsh.initContent = ''
         . ${pkgs.vte}/etc/profile.d/vte.sh
       '';
     })
--- a/modules/programs/antidote.nix
+++ b/modules/programs/antidote.nix
@@ -31,12 +31,12 @@ in {
 
   config = mkIf cfg.enable {
     home.packages = [ cfg.package ];
-    programs.zsh.initExtraBeforeCompInit = let
+    programs.zsh.initContent = let
       configFiles = pkgs.runCommand "hm_antidote-files" { } ''
         echo "${zPluginStr cfg.plugins}" > $out
       '';
       hashId = parseHashId "${configFiles}";
-    in ''
+    in (mkOrder 550 ''
       ## home-manager/antidote begin :
       source ${cfg.package}/share/antidote/antidote.zsh
       ${optionalString cfg.useFriendlyNames
@@ -50,6 +50,6 @@ in {
       antidote load $bundlefile $staticfile
 
       ## home-manager/antidote end
-    '';
+    '');
   };
 }
--- a/modules/programs/atuin.nix
+++ b/modules/programs/atuin.nix
@@ -114,7 +114,7 @@ in {
         fi
       '';
 
-      programs.zsh.initExtra = mkIf cfg.enableZshIntegration ''
+      programs.zsh.initContent = mkIf cfg.enableZshIntegration ''
         if [[ $options[zle] = on ]]; then
           eval "$(${lib.getExe cfg.package} init zsh ${flagsStr})"
         fi
--- a/modules/programs/autojump.nix
+++ b/modules/programs/autojump.nix
@@ -30,7 +30,7 @@ in {
       . ${package}/share/autojump/autojump.bash
     '');
 
-    programs.zsh.initExtra = mkIf cfg.enableZshIntegration ''
+    programs.zsh.initContent = mkIf cfg.enableZshIntegration ''
       . ${package}/share/autojump/autojump.zsh
     '';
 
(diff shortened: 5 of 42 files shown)


khaneliman:
Any of these that we think we might want to proactively change priority on, instead of keeping the same as before?

ambroisie:
More of a question about `initContent` than a review of the migration: should we document which sections use which priority, so that a user can easily slot their configuration where they want it to go?

This would make it easier to migrate from e.g: `initExtraBeforeCompInit` without having to find the implementation details in code.

ambroisie:
It's unfortunate the default priority (`1000`) isn't the one to use for general configuration (i.e: `initExtra`) 🤷.

Otherwise looks great.

khaneliman:
> It's unfortunate the default priority (`1000`) isn't the one to use for general configuration (i.e: `initExtra`) 🤷.
> 
> Otherwise looks great.

Yeah... I thought about shifting that.

modules/programs/mise.nix
```diff
@@ -108,9 +108,9 @@ in {
         eval "$(${getExe cfg.package} activate bash)"
       '';
 
-      zsh.initExtra = mkIf cfg.enableZshIntegration ''
+      zsh.initContent = mkIf cfg.enableZshIntegration (lib.mkOrder 1200 ''
```
teto, line 111: why the change here ?

khaneliman: What do you mean?

teto:
I think it's fine to get ird of initExtraFirst and initExtraBeforeCompInit because these were "niche" options now solved by initContent. Almost everyone is using `initExtra` so the change will impact almost everyone for little gain.

khaneliman:
> I think it's fine to get ird of initExtraFirst and initExtraBeforeCompInit because these were "niche" options now solved by initContent. Almost everyone is using `initExtra` so the change will impact almost everyone for little gain.

The point of the change is that those are pointless options when `initContent` should be used for we're migrating them inside the repo before deprecating them.

modules/programs/mise.nix
teto: has yes I thought it was 1000 `(lib.mkIf (cfg.initExtra != "") (mkOrder 1200 cfg.initExtra))`

khaneliman: > has yes I thought it was 1000 `(lib.mkIf (cfg.initExtra != "") (mkOrder 1200 cfg.initExtra))`

We were saying above it would be nice to if `initExtra` got shifted down to `mkOrder 1000` though. Would make it less cumbersome to use for shifting from initExtra to initContent.

teto:
I find `zsh.initExtra = mkIf cfg.enableZshIntegration '' ` easier to remember than `zsh.initContent = mkIf cfg.enableZshIntegration (lib.mkOrder 1200 ''`. Deprecating initExtra will require lots of users to update their configs with maybe a less readable version while the cost of keeping it for us is almost 0.

khaneliman:
> I find `zsh.initExtra = mkIf cfg.enableZshIntegration '' ` easier to remember than `zsh.initContent = mkIf cfg.enableZshIntegration (lib.mkOrder 1200 ''`. Deprecating initExtra will require lots of users to update their configs with maybe a less readable version while the cost of keeping it for us is almost 0.

Technically, there's no reason they have to do `mkOrder 1200`, they could just do default prio. The migration is just to avoid surprising breakages for a backwards compatible migration.

modules/programs/mise.nix
teto: yes that would make the transition easier/clearer

khaneliman:
Going to hold off on merging this until after migrating the internal priorities so the migration of `initExtra` isn't so obnoxious.

khaneliman:
Alright, removed the commit with all the `mkOrder 1200` changes now that `initExtra` and `initContent` are the same priority.

khaneliman:
Forgot to adjust `initExtraBeforeCompInit` the new priority change in https://github.com/nix-community/home-manager/pull/6676

NireBryce:
Is there any way to use these in the same file?  I just ran into this deprecation warning and the change feels pretty bad.

The way I was using this personally was to use `initExtraFirst` to put things that had to come before everything else (logins, etc)

then use `initExtraBeforeCompInit` to trigger things before home-manager inserted the compinit line

then use `initExtra` for everything else after the auto compinit line.

the distinction matters for some zsh plugins, notably the transient prompt for `p10k` and some others.

(similarly, `initExtraBeforeCompInit` is semantic and `initContents = mkOrder 550` is a manual lookup if I even remember it exists)

khaneliman:
> Is there any way to use these in the same file? I just ran into this deprecation warning and the change feels pretty bad.
> 

The same way we do it in the module. 

```nix
          programs.zsh.initContent = lib.mkMerge [
            (lib.mkIf (cfg.initExtraFirst != "") (lib.mkBefore cfg.initExtraFirst))
            (lib.mkIf (cfg.initExtraBeforeCompInit != "") (mkOrder 550 cfg.initExtraBeforeCompInit))
            (lib.mkIf (cfg.initExtra != "") cfg.initExtra)
      ];
```

> The way I was using this personally was to convert my zshrc to nix using `initExtraFirst` to put things that had to come before everything else (logins, etc)
> 
> then use `initExtraBeforeCompInit` to trigger things before home-manager inserted the compinit line
> 
> then use `initExtra` for everything else after the auto compinit line.
> 
> the distinction matters for some zsh plugins, notably the transient prompt for `p10k` and some others.
> 
> (similarly, `initExtraBeforeCompInit` is semantic and `initContents = mkOrder 550` is a manual lookup if I even remember it exists)
> 

That's why the description of the option lists out these priorities for what the old options were. But, these are just standard `mkOrder` priorities for a `types.lines` option so you can control exactly how you want to configure your shell configuration. Previously, we just kept adding option after option and were limited in how you could generate your file. 

> What I will probably end up doing is to figure out what's automatically enabling `compinit`, explicitly disable it and instead just place the line manually in `initContent`, but this feels like a workaround that misses why things might have been this shape in the first place, even though I know my use is secondary to home-manager being easier to maintain.

PR #6666, opened 2025-03-19
fish: add command option for abbreviations

MrSom3body:
### Description

The fish shell has added a flag to the abbr command which allows one to expand it only if it is typed after a real command e.g.:
git s<Space> -> git status
s<Space> -> s

Also see the last example here: https://fishshell.com/docs/current/cmds/abbr.html#examples

Also every test ran fine, but one firefox related one, which doesn't seem to relate to my change.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@isabelroses

### DIFF
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -167,6 +167,16 @@ let
         '';
       };
 
+      command = mkOption {
+        type = with types; nullOr str;
+        default = null;
+        description = ''
+          Specifies the command(s) for which the abbreviation should expand. If
+          set, the abbreviation will only expand when used as an argument to
+          the given command(s).
+        '';
+      };
+
       setCursor = mkOption {
         type = with types; (either bool str);
         default = false;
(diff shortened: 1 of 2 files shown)


isabelroses:
I don't maintain the fish module. But the code does seem good!

MrSom3body:
Yeahhh there was no maintainer assigned (or I am completely blind) so I just tagged you as you've been the last person to edit it :)

DerRockWolf:
@MrSom3body thanks for implementing this! :slightly_smiling_face: 

The description says `command(s)` but I couldn't find how to specify multiple commands, as the option is just a string.

The fish docs say:
> Multiple --command can be used together, and the abbreviation will expand for each.

MrSom3body:
Oh I must have overseen that. I'll have a look tomorrow or the day after as I don't really have the time for it today

MrSom3body:
Opened a PR (#6851)

PR #6667, opened 2025-03-19
onlyoffice: add module

aguirre-matteo:
### Description

Added programs/onlyoffice.nix module, that provides `enable`, `package` and `settings` options at `programs.onlyoffice`.
All the available settings at `settings` can be deduced by modifing the config in the GUI and observing the changes in ~/.config/onlyoffice/DesktopEditors.conf.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/onlyoffice.nix
+++ b/modules/programs/onlyoffice.nix
@@ -1,0 +1,49 @@
+{ lib, pkgs, config, ... }:
+
+let
+  inherit (lib)
+    types isBool boolToString concatStringsSep mapAttrsToList mkIf
+    mkEnableOption mkPackageOption mkOption;
+
+  cfg = config.programs.onlyoffice;
+
+  attrToString = name: value:
+    let newvalue = if (isBool value) then (boolToString value) else value;
+    in "${name}=${newvalue}";
+
+  getFinalConfig = set:
+    (concatStringsSep "\n" (mapAttrsToList attrToString set)) + "\n";
+in {
+  meta.maintainers = with lib.hm.maintainers; [ aguirre-matteo ];
+
+  options.programs.onlyoffice = {
+    enable = mkEnableOption "distrobox";
+
+    package = mkPackageOption pkgs "onlyoffice-desktopeditors" { };
+
+    settings = mkOption {
+      type = with types; attrsOf (either bool str);
+      default = { };
+      example = ''
+        UITheme = "theme-contrast-dark";
+        editorWindowMode = false;
+        forcedRtl = false;
+        maximized = true;
+        titlebar = "custom";
+      '';
+      description = ''
+        Configuration settings for Onlyoffice.
+
+        All configurable options can be deduced by enabling them through the
+        GUI and observing the changes in ~/.config/onlyoffice/DesktopEditors.conf.
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+
+    xdg.configFile."onlyoffice/DesktopEditors.conf".source =
+      pkgs.writeText "DesktopEditors.conf" (getFinalConfig cfg.settings);
+  };
+}
--- a/modules/programs/onlyoffice.nix
+++ b/modules/programs/onlyoffice.nix
@@ -41,7 +41,7 @@ in {
   };
 
   config = mkIf cfg.enable {
-    home.packages = [ cfg.package ];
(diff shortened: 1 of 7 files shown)


modules/programs/onlyoffice.nix
```diff
@@ -0,0 +1,49 @@
...
+
+  getFinalConfig = set:
+    (concatStringsSep "\n" (mapAttrsToList attrToString set)) + "\n";
+in {
+  meta.maintainers = with lib.hm.maintainers; [ aguirre-matteo ];
+
+  options.programs.onlyoffice = {
+    enable = mkEnableOption "distrobox";
+
+    package = mkPackageOption pkgs "onlyoffice-desktopeditors" { };
```
khaneliman, line 22:
```suggestion
    package = mkPackageOption pkgs "onlyoffice-desktopeditors" { nullable = true; };
```

modules/programs/onlyoffice.nix
```diff
@@ -0,0 +1,49 @@
...
+        Configuration settings for Onlyoffice.
+
+        All configurable options can be deduced by enabling them through the
+        GUI and observing the changes in ~/.config/onlyoffice/DesktopEditors.conf.
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
```
khaneliman, line 44:
```suggestion
    home.packages = lib.mkIf (cfg.package != null) [ cfg.package ];
```

khaneliman (commented):
Can you also add `onlyoffice-desktopeditors` to https://github.com/nix-community/home-manager/blob/97a00e0659b2807454507eb3a593bd09b099bd80/tests/default.nix#L68

aguirre-matteo:
@khaneliman Sure!

khaneliman:
Sorry, the test can stay in shared. Just need to stub the package explicitly on Darwin to avoid the build of it

khaneliman:
I did a treewide recently to support using home manager to allow generating configs without installing the package. This is useful for programs that might not be available for Darwin through Nixpkgs, but are available through other means. So, we can stub the package to verify the module still works without relying on the packaging of upstream.

aguirre-matteo:
@khaneliman Then what should I do? Rollback the last changes, or which of them?

modules/programs/onlyoffice.nix
```diff
@@ -0,0 +1,50 @@
...
+    let newvalue = if (isBool value) then (boolToString value) else value;
+    in "${name}=${newvalue}";
+
+  getFinalConfig = set:
+    (concatStringsSep "\n" (mapAttrsToList attrToString set)) + "\n";
+in {
+  meta.maintainers = with lib.hm.maintainers; [ aguirre-matteo ];
+
+  options.programs.onlyoffice = {
+    enable = mkEnableOption "distrobox";
```
khaneliman, line 20:
```suggestion
    enable = mkEnableOption "onlyoffice";
```

khaneliman (commented):
I would just revert the Linux only change and add the package to the list of Darwin packages to stub

tests/default.nix
```diff
@@ -142,6 +142,7 @@ let
       "nix-index"
       "nix-your-shell"
       "ollama"
+      "onlyoffice"
```
khaneliman, line 145:
```suggestion
      "onlyoffice-desktopeditors"
```

This is the actual package name from nixpkgs.

aguirre-matteo:
@khaneliman Like this?

tests/default.nix
khaneliman:
```bash
nix build .#onlyoffice                                                                              
error: flake 'git+file:///home/khaneliman/Documents/github/nixpkgs' does not provide attribute 'packages.x86_64-linux.onlyoffice', 'legacyPackages.x86_64-linux.onlyoffice' or 'onlyoffice'
```
```bash
nix build .#onlyoffice-desktopeditors
these 3 derivations will be built:
  /nix/store/m7vv4bklz3hc532yv33ykxy68fpjs7g3-onlyoffice-desktopeditors-8.3.1-init.drv
  /nix/store/8bsc57m1n87r85h5dn581k9z5b9gxsbg-onlyoffice-desktopeditors-8.3.1-bwrap.drv
  /nix/store/4v2rn0h9jjvac19mv7m3zzxqn91i5f9d-onlyoffice-desktopeditors-8.3.1.drv

khaneliman:
Just the last fix, that I can see. Everything else looks good.

Issue #6663, opened 2025-03-19
bug: McFly-fzf integration breaks non-interactive shells


mythmon:
### Are you following the right branch?

- [x] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Issue description

I have the following config on one of my systems.

```
mcfly = {
  enable = true;
  enableFishIntegration = true;
  fzf.enable = true;
};
```

When I ssh into that system interactively, everything works fine. If I try to run commands with `ssh` (for example, Zed remote editing), then I get the following message:

```
Mcfly-fzf: Must initialize mcfly before mcfly-fzf
```

This even happens with very simple commands, like this:

```
$ ssh $HOST echo test
Mcfly-fzf: Must initialize mcfly before mcfly-fzf
test
```

This breaks many tools that don't expect the Mcfly message

### Maintainer CC

@trueNAHO @rycee @iofq 

### System information

The system I am sshing *from*:

 - system: `"x86_64-linux"`
 - host os: `Linux 6.12.17, NixOS, 25.05 (Warbler), 25.05.20250301.303bd80`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.24.12`
 - nixpkgs: `/nix/store/7g9h6nlrx5h1lwqy4ghxvbhb7imm3vcb-source`

The system I am sshing *to*:

 - system: `"x86_64-linux"`
 - host os: `Linux 6.12.17, NixOS, 25.05 (Warbler), 25.05.20250301.303bd80`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.24.12`
 - nixpkgs: `/nix/store/7g9h6nlrx5h1lwqy4ghxvbhb7imm3vcb-source`

(Edit: I realize now these are identical. That's not a mistake. Both systems are NixOS built from the same config)

khaneliman:
Kinda weird since the code in the `mcfly` module explicitly adds the lines in order of init `mcfly` and then init `mcfly-fzf`

https://github.com/nix-community/home-manager/blob/94605dcadefeaff6b35c8931c9f38e4f4dc7ad0a/modules/programs/mcfly.nix#L22-L26

mythmon:
[...]

PR #6669, opened 2025-03-20
mcfly: fix mcfly-fzf in non-interactive shells

mythmon:
### Description

mcfly-fzf's initialization steps don't properly check for interactive shells. It does check, but only after checking if mcfly has been initialized. Because mcfly does not initiate itself on non-interactive shells, that causes mcfly-fzf to return an exit code, breaking non-interactive shells.

Closes https://github.com/nix-community/home-manager/issues/6663
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@trueNAHO @rycee @iofq

### DIFF
--- a/modules/programs/mcfly.nix
+++ b/modules/programs/mcfly.nix
@@ -7,22 +7,33 @@ let
 
   tomlFormat = pkgs.formats.toml { };
 
+  # These shell integrations all wrap `mcfly-fzf init` in an interactive shell
+  # check, due to a buggy interactivity check in its initialization.
+  # See: https://github.com/nix-community/home-manager/issues/6663
+  # and https://github.com/bnprks/mcfly-fzf/issues/10
+
   bashIntegration = ''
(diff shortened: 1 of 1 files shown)


modules/programs/mcfly.nix
```diff
@@ -7,22 +7,33 @@ let
...
+    if [[ $- =~ i ]]; then
+      eval "$(${getExe pkgs.mcfly-fzf} init bash)"
+    fi
   '';
 
   fishIntegration = ''
     ${getExe pkgs.mcfly} init fish | source
   '' + optionalString cfg.fzf.enable ''
-    ${getExe pkgs.mcfly-fzf} init fish | source
+    if [[ -o interactive ]]; then
```
khaneliman, line 26: this looks flipped with the zsh check

mythmon: oops, thanks

PR #6670, opened 2025-03-20
firefox: fix bookmarks backwards compatibility

brckd:
### Description

The legacy attrset option type for `firefox.profiles.<name>.bookmarks` was accidentally removed in 9d55428. This adds back support for this type by refactoring the bookmarks submodule. This also adds a new test ensuring this won't happen again.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@kira-bruneau

### DIFF
--- a/modules/programs/firefox/profiles/bookmark-types.nix
+++ b/modules/programs/firefox/profiles/bookmark-types.nix
@@ -1,8 +1,11 @@
 { lib, ... }:
 
-with lib;
+let
 
-rec {
+  inherit (builtins) attrValues;
+  inherit (lib) types mkOption;
+
+in rec {
   settingsType = with types;
     coercedTo (addCheck (attrsOf nodeType) (attrs: !(attrs ? settings)))
     attrValues (listOf nodeType);
--- a/modules/programs/firefox/profiles/bookmark-types.nix
+++ b/modules/programs/firefox/profiles/bookmark-types.nix
@@ -1,0 +1,70 @@
+{ lib, ... }:
+
+with lib;
+
+rec {
+  settingsType = with types;
+    coercedTo (addCheck (attrsOf nodeType) (attrs: !(attrs ? settings)))
+    attrValues (listOf nodeType);
+
+  bookmarkSubmodule = types.submodule ({ name, ... }: {
+    options = {
+      name = mkOption {
+        type = types.str;
+        default = name;
+        description = "Bookmark name.";
+      };
+
+      tags = mkOption {
+        type = types.listOf types.str;
+        default = [ ];
+        description = "Bookmark tags.";
+      };
+
+      keyword = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        description = "Bookmark search keyword.";
+      };
+
+      url = mkOption {
+        type = types.str;
+        description = "Bookmark url, use %s for search terms.";
+      };
+    };
+  }) // {
+    description = "bookmark submodule";
+  };
+
+  bookmarkType = types.addCheck bookmarkSubmodule (x: x ? "url");
+
+  directoryType = types.submodule ({ name, ... }: {
+    options = {
+      name = mkOption {
+        type = types.str;
+        default = name;
+        description = "Directory name.";
+      };
+
+      bookmarks = mkOption {
+        type = types.listOf nodeType;
+        default = [ ];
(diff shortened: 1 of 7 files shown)


modules/programs/firefox/profiles/bookmark-types.nix
```diff
@@ -0,0 +1,70 @@
+{ lib, ... }:
+
+with lib;
```
khaneliman, line 3:
```suggestion
```
I missed this in the last PR. Could you add a commit removing from the other module, as well?

modules/programs/firefox/profiles/bookmark-types.nix
```diff
@@ -0,0 +1,70 @@
...
+
+rec {
+  settingsType = with types;
+    coercedTo (addCheck (attrsOf nodeType) (attrs: !(attrs ? settings)))
```
brckd, line 7: This is a bit hacky as it checks for the existence of a `settings` attribute to determine whether this is the old attribute type or the new submodule type. It's a slight backwards incompatibility, for the unlikely case that someone set a bookmark on the `settings` attribute before.

modules/programs/firefox/profiles/bookmark-types.nix
brckd: Sure! Just out of interest, is `with lib;` now discouraged in general? Many modules still use it.

brckd: I pushed [e878374](https://github.com/nix-community/home-manager/pull/6670/commits/e87837446eeef4c95c1f4240d5c2555acf1641f3), which uses `inherit (lib) ...` instead. Is this good?

khaneliman:
@ethorsoe test looks good, can you verify?

modules/programs/firefox/profiles/bookmark-types.nix
ethorsoe: I have several bookmarks that could be settings, like nixos options search, I might just rather live with the incompatibility, the failure mode actually fails the evaluation, so it's not that bad.

brckd: I have now done this in 5e0785d.

khaneliman:
@ethorsoe sorry to ping again, does this resolve your specific use case?

kira-bruneau:
I found where someone was using an attrs config here: https://github.com/Julow/env/blob/2531ac9bcd3d46b6749ab17c3e92519064a92df0/home/firefox/bookmarks.nix#L5.

After #6402 I was getting:
```
error:
         at /nix/store/qjmsdak7fhqc7pi922i2mgm45f7af9h8-source/lib/attrsets.nix:1574:11:
         1574|           head values
             |           ^
         1575|         else

         at /nix/store/qjmsdak7fhqc7pi922i2mgm45f7af9h8-source/lib/modules.nix:816:9:
          815|     in warnDeprecation opt //
          816|       { value = addErrorContext "while evaluating the option `${showOption loc}':" value;
             |         ^
          817|         inherit (res.defsFinal') highestPrio;

       error: The option `home-manager.users.kira.programs.firefox.profiles.firefox.bookmarks."Android Developers"' does not exist. Definition values:
       - In `/nix/store/ccfjy5ll4di51mx8pkjmp8ckb20vyn7b-source/home/programs/librewolf':
           {
             keyword = "android";
             name = "Android Developers";
             url = "https://developer.android.com/s/results?q=%s&all_languages=true&text";
           }
```

But with this PR the attrs-defined bookmarks are working again.

I did notice another issue though, if you don't a define a name you'll get a generated name. This still happens before any of the bookmarks refactoring though, and it looks like it's been an issue since bookmarks were implemented: 

image

With names explicitly defined I now have:

image

I'll open a PR to fix that too once we have backwards compat issues settled.

kira-bruneau (approved):

khaneliman:
Hmm.. didn't think this would need pulling down to test but see

```bash
evaluation warning: Firefox bookmarks have been refactored into a submodule that now explicitly require a 'force' option to be enabled.

                    Replace:

                    programs.firefox.profiles.khaneliman.bookmarks = [ ... ];

                    With:

                    programs.firefox.profiles.khaneliman.bookmarks = {
                      force = true;
                      settings = [ ... ];
                    };
```

and I don't use this option.

PR #6674, opened 2025-03-21
davmail: init module

bmrips:
### Description

Initialize a module for the [DavMail](https://davmail.sourceforge.net/) service.

The code is mostly copied from the corresponding NixOS module; but slightly refactored. In addition, some restrictions on the systemd service were released to allow DavMail to launch a GUI when necessary.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/davmail.nix
+++ b/modules/services/davmail.nix
@@ -1,0 +1,128 @@
+{ config, lib, pkgs, ... }:
+let
+
+  inherit (lib)
+    mapAttrsRecursive mkDefault mkEnableOption mkIf mkOption optionalAttrs
+    types;
+
+  cfg = config.services.davmail;
+
+  javaProperties = pkgs.formats.javaProperties { };
+
+  settingsFile = javaProperties.generate "davmail.properties" cfg.settings;
+
+in {
+
+  meta.maintainers = [ lib.hm.maintainers.bmrips ];
+
+  options.services.davmail = {
+
+    enable = mkEnableOption "DavMail, an MS Exchange gateway.";
+
+    imitateOutlook = mkOption {
+      type = types.bool;
+      default = false;
+      description = "Whether DavMail pretends to be Outlook.";
+      example = true;
+    };
+
+    settings = mkOption {
+      type = javaProperties.type;
+      default = { };
+      description = ''
+        Davmail configuration. Refer to
+        <http://davmail.sourceforge.net/serversetup.html>
+        and <http://davmail.sourceforge.net/advanced.html>
+        for details on supported values.
+      '';
+      example = {
+        "davmail.url" = "https://outlook.office365.com/EWS/Exchange.asmx";
+        "davmail.allowRemote" = true;
+        "davmail.imapPort" = 55555;
+        "davmail.bindAddress" = "10.0.1.2";
+        "davmail.smtpSaveInSent" = true;
+        "davmail.folderSizeLimit" = 10;
+        "davmail.caldavAutoSchedule" = false;
+        "log4j.logger.rootLogger" = "DEBUG";
+      };
+    };
+
+  };
+
+  config = mkIf cfg.enable {
+
+    assertions = [{
+      assertion = pkgs.stdenv.hostPlatform.isLinux;
+      message = "The DavMail service is only available on Linux.";
+    }];
+
+    services.davmail.settings = mapAttrsRecursive (_: mkDefault) {
+      "davmail.server" = true;
+      "davmail.disableUpdateCheck" = true;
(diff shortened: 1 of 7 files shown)


modules/services/davmail.nix
```diff
@@ -0,0 +1,142 @@
...
+        map (line: name + "." + line) (linesForAttrs value)
+      else
+        [ "${name}=${toStr value}" ]) (attrNames attrs);
+
+  settingsFile = pkgs.writeText "davmail.properties"
+    (concatStringsSep "\n" (linesForAttrs cfg.settings));
+
+in {
+
+  meta.maintainers = [ lib.maintainers.bmrips ];
```
khaneliman, line 32:
```suggestion
  meta.maintainers = [ lib.hm.maintainers.bmrips ];
```

bmrips: Ah, that's why the build fails. Thank you!

modules/services/davmail.nix
```diff
@@ -0,0 +1,142 @@
...
+  linesForAttrs = attrs:
+    concatMap (name:
+      let value = attrs.${name};
+      in if isAttrs value then
+        map (line: name + "." + line) (linesForAttrs value)
+      else
+        [ "${name}=${toStr value}" ]) (attrNames attrs);
+
+  settingsFile = pkgs.writeText "davmail.properties"
+    (concatStringsSep "\n" (linesForAttrs cfg.settings));
```
khaneliman, line 28: I feel like this could all be simplified for generating the config. 

```suggestion
  settingsFile = pkgs.writeText "davmail.properties"
    (lib.generators.toKeyValue { indent = "  "; } cfg.settings));
```

modules/services/davmail.nix
```diff
@@ -0,0 +1,142 @@
...
+
+    imitateOutlook = mkOption {
+      type = types.bool;
+      default = false;
+      description = "Whether DavMail pretends to be Outlook.";
+      example = true;
+    };
+
+    settings = mkOption {
+      type = settingsType;
```
khaneliman, line 46:
```suggestion
      type = types.attrsOf types.anything;
```

modules/services/davmail.nix
```diff
@@ -0,0 +1,142 @@
...
+      '';
+      example = {
+        davmail.url = "https://outlook.office365.com/EWS/Exchange.asmx";
+        davmail.allowRemote = true;
+        davmail.imapPort = 55555;
+        davmail.bindAddress = "10.0.1.2";
+        davmail.smtpSaveInSent = true;
+        davmail.folderSizeLimit = 10;
+        davmail.caldavAutoSchedule = false;
+        log4j.logger.rootLogger = "DEBUG";
```
khaneliman, line 62: The config would change to string keys like 

```nix
"davmail.allowRemote" = true;
```

modules/services/davmail.nix
bmrips: I tried this, but it fails since `lib.generators.toKeyValue` does not handle nested attrs.

modules/services/davmail.nix
bmrips: This might be simpler but enables invalid values. Hence, I prefer to keep the more complex but accurate type.

modules/services/davmail.nix
khaneliman: https://github.com/nix-community/home-manager/pull/6674#discussion_r2008924713

bmrips: I see where you're getting at. Actually, there also is `pkgs.formats.javaProperties` which seems to be even better in this case. And interestingly, a [comment in the implementation of it](https://github.com/NixOS/nixpkgs/blob/f9ec383a105dfa6ea11dc7527563aa6e929503bf/pkgs/pkgs-lib/formats/java-properties/default.nix#L23-L45) discourages the use of nested attribute sets.

Hence, I refactored the module to use `pkgs.formats.javaProperties`.

modules/services/davmail.nix
```diff
@@ -0,0 +1,128 @@
...
+      type = javaProperties.type;
+      default = { };
+      description = ''
+        Davmail configuration. Refer to
+        <http://davmail.sourceforge.net/serversetup.html>
+        and <http://davmail.sourceforge.net/advanced.html>
+        for details on supported values.
+      '';
+      example = {
+        davmail.url = "https://outlook.office365.com/EWS/Exchange.asmx";
```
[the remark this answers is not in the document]
bmrips: Right, I missed that. It's fixed now. Thank you for the thorough review!

khaneliman (approved):
LGTM, I think it's cleaner and easier to maintain.

bmrips:
A news would also be beneficial.

PR #6675, opened 2025-03-21
waybar: integrate with tray.target

alejo7797:
Inspired by [#3432][0], but with a smaller footprint.

This change makes it so that services with `Requires=tray.target` and `After=tray.target`, such as [`blueman-applet`][1] or [`network-manager-applet`][2] get started *after* waybar is ready.

[0]: https://github.com/nix-community/home-manager/pull/3432
[1]: https://github.com/nix-community/home-manager/blob/6c2b79403e9ae852fbf1d55b29f2ea4d2aa43255/modules/services/blueman-applet.nix
[2]: https://github.com/nix-community/home-manager/blob/6c2b79403e9ae852fbf1d55b29f2ea4d2aa43255/modules/services/network-manager-applet.nix

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/waybar.nix
+++ b/modules/programs/waybar.nix
@@ -321,7 +321,7 @@ in {
           Description =
             "Highly customizable Wayland bar for Sway and Wlroots based compositors.";
           Documentation = "https://github.com/Alexays/Waybar/wiki";
-          PartOf = [ cfg.systemd.target ];
+          PartOf = [ cfg.systemd.target "tray.target" ];
           After = [ cfg.systemd.target ];
           ConditionEnvironment = "WAYLAND_DISPLAY";
           X-Restart-Triggers = optional (settings != [ ])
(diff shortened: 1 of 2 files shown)


teto:
the implementation seems to be a perfect duplicate of cfg.systemd.target. Shouldn't it become a default for cfg.systemd.target and/or convert `cfg.systemd.target` into a list ?

alejo7797:
> the implementation seems to be a perfect duplicate of cfg.systemd.target. Shouldn't it become a default for cfg.systemd.target and/or convert `cfg.systemd.target` into a list ?

I think they play different roles, in that `waybar.service` should trigger after `cfg.systemd.target` but *before* `tray.target` (added as a [default dependency](https://www.freedesktop.org/software/systemd/man/latest/systemd.target.html#Default%20Dependencies)) in order to signal readiness to tray applets.

teto (approved):

PR #6680, opened 2025-03-22
vscode: Fix version checks when using Cursor

NeQuissimus:
### Description

The version check for VSCode fails when using Cursor.
Cursor has its own versioning and is 
currently at 0.45.14 (based on VSCode version 1.96.2).
The version checks for generating the `extensions.json`
in the module look at the package version,
which causes them not to be generated for Cursor.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@Reputable2772 @HyunggyuJang

### DIFF
--- a/modules/programs/vscode.nix
+++ b/modules/programs/vscode.nix
@@ -389,25 +389,26 @@ in {
         mkMerge (concatMap toPaths
           (flatten (mapAttrsToList (n: v: v.extensions) cfg.profiles))
           ++ optional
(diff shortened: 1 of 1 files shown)


khaneliman:
Is there perhaps a minimum version of cursor we should check instead of just the pname alone?

NeQuissimus:
VSCode 1.74, which is what we are checking for here, predates Cursor's adoption of VSCode. So all VSCode-based versions of Cursor support this.

Issue #6157, opened 2024-12-01
Stop garbage-collecting and rebuilding fish completions


marienz:
I noticed `nixos-rebuild` (re)building -fish-completion packages more than I'd expect.

What I think is happening is that the `destructiveSymlinkJoin` of all packages' fish-completions ([here](https://github.com/nix-community/home-manager/blob/819f682269f4e002884702b87e445c82840c68f2/modules/programs/fish.nix#L447-L452)) filters out packages without completions, and those then get garbage-collected by `nix-collect-garbage`. That is:

```
> nix derivation show /nix/store/hx407a4i3s4490i05jr3fcapdij1h1xh-marienz-fish-completions
...
      "paths": "/nix/store/gvglak58m5fm0kjzpqr90r9p4mlpq014-man-pages-6.9.1-fish-completions /nix/store/bk4g9wk1p4f0j75hhvbvisgl0k81hpzf-sqlite-3.46.1-fish-completions ...
```
(long list of per-package completions as input)

```
> ls /nix/store/gvglak58m5fm0kjzpqr90r9p4mlpq014-man-pages-6.9.1-fish-completions /nix/store/bk4g9wk1p4f0j75hhvbvisgl0k81hpzf-sqlite-3.46.1-fish-completions
/nix/store/bk4g9wk1p4f0j75hhvbvisgl0k81hpzf-sqlite-3.46.1-fish-completions:

/nix/store/gvglak58m5fm0kjzpqr90r9p4mlpq014-man-pages-6.9.1-fish-completions:
iconvconfig.fish  ldconfig.fish  ld.so.fish  nscd.fish  sched.fish  string_copying.fish  symlink.fish  time_namespaces.fish  zdump.fish
```

(man-pages has completions, sqlite does not, it's an empty directory)

```
> nix-store --query --roots /nix/store/gvglak58m5fm0kjzpqr90r9p4mlpq014-man-pages-6.9.1-fish-completions
/nix/var/nix/profiles/system-574-link -> /nix/store/gdglrxcs5fk3nrd4639giqyfb9fja6ah-nixos-system-jian-24.11.20241128.c71ad5c
/run/booted-system -> /nix/store/gdglrxcs5fk3nrd4639giqyfb9fja6ah-nixos-system-jian-24.11.20241128.c71ad5c
[...]

PR #6686, opened 2025-03-23
Stop garbage-collection empty fish completion packages

marienz:
### Description

- Adds `home.extraDependencies`, which (like `system.extraDependencies` in NixOS) adds a package to the runtime closure without exposing it elsewhere.
- Uses this to keep empty Fish completions around. These are otherwise needlessly garbage-collected: garbage-collecting them saves almost no space (just an empty directory) and they're usually rebuilt soon after. See #6157.

The diff for the Fish change looks large, but all it does is add the inputs to `destructiveSymlinkJoin` to `home.extraDependencies` too: I have to move some code around to make that possible.

### Checklist

(nmt-test-firefox-profiles-search and nmt-test-floorp-profiles-search fail, but they do so on master too... I'll try to find some time to track down what's going wrong there. Might be somehow related to #6668, since they're failing with output differing on `librewolf` versus `firefox` / `floorp` respectively...)

I don't know how to meaningfully unit-test this, but I can probably figure out how to add an integration test if you need one: along the lines of "in a VM test, run `nix-store --query -R ~/.local/state/home-manager/gcroots/current-home` and confirm some package added to `home.extraDependencies` is in there".

I'd appreciate suggestions on how to test this: I'm pretty new to Nix tests, and this is a bit fiddly to test because the whole point is that there should be no observable side effect of what I'm changing other than something not getting GC'd (which I don't know how to test other than running `nix` in an integration test).

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee for home-environment

I don't see a maintainer listed for Fish...

### DIFF
--- a/modules/home-environment.nix
+++ b/modules/home-environment.nix
@@ -383,6 +383,15 @@ in
       '';
     };
 
+    home.extraDependencies = mkOption {
+      type = types.listOf types.pathInStore;
+      default = [ ];
+      description = ''
+        A list of paths that should be included in the home
+        closure but generally not visible.
+      '';
+    };
+
     home.path = mkOption {
       internal = true;
       description = "The derivation installing the user packages.";
@@ -780,6 +789,8 @@ in
       pkgs.runCommand "home-manager-generation"
         {
           preferLocalBuild = true;
+          passAsFile = [ "extraDependencies" ];
+          inherit (config.home) extraDependencies;
         }
         ''
           mkdir -p $out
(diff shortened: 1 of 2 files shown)


marienz:
Rebased to address merge conflicts.

With nixfmt-tree-format, the diff for the Fish change now looks *really* large :( I'm still "only" moving code into a new let-binding, but there's now a lot more re-indenting going on, so a diff not ignoring whitespace is borderline unreadable.

All it does:

- Move `generateCompletions` to a new outer `let` binding
- Move the expression previously passed as `paths` to `destructiveSymlinkJoin` to the same `let` binding as `allCompletions`
- Pass `allCompletions` to `home.extraDependencies` too

rycee:
Thanks for the contribution! I've merged it now 🙂

PR #6689, opened 2025-03-23
firefox: don't show migration warning when bookmarks isn't set

kira-bruneau:
### Description

Continuing from #6670, this stops the migration warning from being shown if you didn't have any bookmarks set.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@khaneliman @brckd

### DIFF
--- a/modules/programs/firefox/mkFirefoxModule.nix
+++ b/modules/programs/firefox/mkFirefoxModule.nix
@@ -383,27 +383,31 @@ in {
           bookmarks = mkOption {
             type = (with types;
               coercedTo bookmarkTypes.settingsType (bookmarks:
-                warn ''
-                  ${cfg.name} bookmarks have been refactored into a submodule that now explicitly require a 'force' option to be enabled.
+                if bookmarks != { } then
+                  warn ''
+                    ${cfg.name} bookmarks have been refactored into a submodule that now explicitly require a 'force' option to be enabled.
 
-                  Replace:
+                    Replace:
 
-                  ${moduleName}.profiles.${name}.bookmarks = [ ... ];
+                    ${moduleName}.profiles.${name}.bookmarks = [ ... ];
 
-                  With:
+                    With:
 
-                  ${moduleName}.profiles.${name}.bookmarks = {
+                    ${moduleName}.profiles.${name}.bookmarks = {
+                      force = true;
+                      settings = [ ... ];
+                    };
+                  '' {
                     force = true;
-                    settings = [ ... ];
-                  };
-                '' {
-                  force = true;
-                  settings = bookmarks;
-                }) (submodule ({ config, ... }:
-                  import ./profiles/bookmarks.nix {
-                    inherit config lib pkgs;
-                    modulePath = modulePath ++ [ "profiles" name "bookmarks" ];
-                  })));
+                    settings = bookmarks;
+                  }
+                else
+                  { }) (submodule ({ config, ... }:
+                    import ./profiles/bookmarks.nix {
+                      inherit config lib pkgs;
+                      modulePath = modulePath
+                        ++ [ "profiles" name "bookmarks" ];
+                    })));
             default = { };
             internal = !enableBookmarks;
             description = "Declarative bookmarks.";


modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -383,27 +383,31 @@ in {
           bookmarks = mkOption {
             type = (with types;
               coercedTo bookmarkTypes.settingsType (bookmarks:
-                warn ''
-                  ${cfg.name} bookmarks have been refactored into a submodule that now explicitly require a 'force' option to be enabled.
+                if bookmarks != { } then
```
khaneliman, line 386:
```suggestion
                if (bookmarks != { } && bookmarks != [ ] && !(bookmarks ? force)) then
```

This is what I got working locally to prevent the warning from showing for users who actually configure the module properly.

kira-bruneau: Hmm, this seems hackier than it needs to be and I think we need more time to think it through. For now I think it'd be better to just revert https://github.com/nix-community/home-manager/pull/6670 until we think of a better solution.

khaneliman: Isn't the point of the warning to alert for the need of `force`? This just makes sure we aren't warning people who have force set.

kira-bruneau: Not exactly, the point of the warning is to inform users to migrate to the new submodule format. There's a seperate assertion to check if `force` isn't set, but only if you're using the submodule.

khaneliman: Right, forgot about that.

khaneliman: coercion is supposed to be oldType -> newType. If we're passing the current type into a coercion as the first argument, I don't believe it'd be behaving the way expected, then. So we have 2 coercions right now, one in the top level option and one in the type defintion.

kira-bruneau: `bookmarksTypes.settingsType` should be the oldType, but with https://github.com/nix-community/home-manager/pull/6670, it also adds a check to try to fix backwards compat.

The behaviour we're getting is different that what I'd expect just through reading the code, which is why I'm erring on reverting it for now.

kira-bruneau: wait i think `nodeType.check attrs.settings` just needs to be `!(nodeType.check attrs.settings)`

kira-bruneau: @khaneliman Wait I think the additional conditions you added was just to work around the warning right? We specifically want to show the warning for:

```nix
bookmarks = { };
```

and the assertion for

```nix
bookmarks = {
  settings = { };
};
```

So this PR should be good as is. I was worried there was something else wrong :melting_face:

khaneliman: > @khaneliman Wait I think the additional conditions you added was just to work around the warning right? We specifically want to show the warning for:
> 
> ```nix
> bookmarks = { };
> ```
> 
> and the assertion for
> 
> ```nix
> bookmarks = {
>   settings = { };
> };
> ```
> 
> So this PR should be good as is. I was worried there was something else wrong 🫠

I get the warning with `bookmarks = { settings = { } ;}; I don't get an assertion.

kira-bruneau: Oh right, because it'd be interpreting empty settings as a bookmark, but something like `bookmarks = { settings = [ ] ; };` should throw an assertion.

khaneliman: Alright, that makes sense... that's part of the coercion to be backwards compatible. That makes me feel better about what I'm observing! Thanks

kira-bruneau:
Thank you so much @khaneliman & @brckd for working through this :heart:. This was a really weird migration to figure out.

khaneliman:
> Thank you so much @khaneliman & @brckd for working through this ❤️. This was a really weird migration to figure out.

Ditto, appreciate the effort you both have been putting in! Sorry for any confusion

PR #6694, opened 2025-03-24
carapace: conditionally disable unnecessary fish completion workaround

feathecutie:
### Description

This disables a workaround used for getting carapace to work correctly with the fish shell.
This workaround stopped being necessary with fish 4.0 (which has been in nixpkgs unstable for a few weeks now), as noted in https://carapace-sh.github.io/carapace-bin/setup.html#fish.

I was debating removing this workaround entirely instead of just adding a version check, but I wasn't sure if it's okay to assume that home-manager `master` users also use nixpkgs unstable.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@Weathercold 
@bobvanderlinden

### DIFF
--- a/modules/programs/carapace.nix
+++ b/modules/programs/carapace.nix
@@ -61,18 +61,23 @@ in {
       };
     };
 
-    xdg.configFile =
-      mkIf (config.programs.fish.enable && cfg.enableFishIntegration) (
+    xdg.configFile = mkIf (config.programs.fish.enable
+      && cfg.enableFishIntegration
+      && lib.versionOlder config.programs.fish.package.version "4.0.0") (
         # Convert the entries from `carapace --list` to empty
         # xdg.configFile."fish/completions/NAME.fish" entries.
         #
         # This is to disable fish builtin completion for each of the
-        # carapace-supported completions It is in line with the instructions from
+        # carapace-supported completions.
+        #
+        # This is necessary for carapace to properly work with fish version < 4.0b1.
+        #
+        # It is in line with the instructions from
         # carapace-bin:
         #
         #   carapace --list | awk '{print $1}' | xargs -I{} touch ~/.config/fish/completions/{}.fish
         #
-        # See https://github.com/rsteube/carapace-bin#getting-started
+        # See https://carapace-sh.github.io/carapace-bin/setup.html#fish
         let
           carapaceListFile = pkgs.runCommandLocal "carapace-list" {
             buildInputs = [ cfg.package ];
--- a/tests/modules/programs/carapace/fish.nix
+++ b/tests/modules/programs/carapace/fish.nix
@@ -8,13 +8,17 @@ lib.mkIf config.test.enableBig {
 
   nixpkgs.overlays = [ (self: super: { inherit (realPkgs) carapace; }) ];
 
-  nmt.script = ''
+  nmt.script = let
+    needsCompletionOverrides = lib.versionOlder realPkgs.fish.version "4.0.0";
+  in ''
     assertFileExists home-files/.config/fish/config.fish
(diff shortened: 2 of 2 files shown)


modules/programs/carapace.nix
```diff
@@ -61,18 +61,23 @@ in {
       };
     };
 
-    xdg.configFile =
-      mkIf (config.programs.fish.enable && cfg.enableFishIntegration) (
+    xdg.configFile = mkIf (config.programs.fish.enable
+      && cfg.enableFishIntegration
+      && lib.versionOlder pkgs.fish.version "4.0.0") (
```
khaneliman, line 66: Technically, you're checking the version of fish available in nixpkgs, not what the user has installed. 
```suggestion
      && lib.versionOlder config.programs.fish.package.version "4.0.0") (
```

tests/modules/programs/carapace/fish.nix
```diff
@@ -12,9 +12,9 @@ lib.mkIf config.test.enableBig {
...
     assertFileRegex home-files/.config/fish/config.fish \
       '/nix/store/.*carapace.*/bin/carapace _carapace fish \| source'
-
-    # Check whether completions are overridden.
+  '' + (lib.optionalString (lib.versionOlder realPkgs.fish.version "4.0.0") ''
+    # Check whether completions are overridden, necessary for fish < 4.0
     assertFileExists home-files/.config/fish/completions/git.fish
     assertFileContent home-files/.config/fish/completions/git.fish /dev/null
-  '';
+  '');
```
bobvanderlinden, line 19: To fully test both cases it might make sense to add:

```
let
  isFish4 = lib.versionOlder realPkgs.fish.version "4.0.0";
in
```

```
  '' + (lib.optionalString isFish4 ''
    # Check whether completions are overridden, necessary for fish < 4.0
    assertFileExists home-files/.config/fish/completions/git.fish
    assertFileContent home-files/.config/fish/completions/git.fish /dev/null
  '' + (lib.optionalString isFish4 ''
    assertPathNotExists home-files/.config/fish/completions
  '';
```

I do have to say I was quite confused about `versionOlder` behaving differently from how I interpret the documentation (https://nixos.org/manual/nixpkgs/stable/#function-library-lib.strings.versionOlder). But your change does match the actual implementation of the function (true when second argument is older than first argument) :+1:

bobvanderlinden (approved):
Aah, good to see the completion overrides aren't needed anymore. Good to see this change. It works locally with fish 4.0.1.

Changes make sense :+1:

Weathercold:
Might be a good idea to warn for deprecation of old fish versions so that we can remove the workaround later

tests/modules/programs/carapace/fish.nix
feathecutie: From the way I understand it, the docs for `versionOlder` should be correct. It checks whether v1 (first argument) is older than v2 (second argument), so in this case, whether `realPkgs.fish.version` is older than `"4.0.0"`.
I did also add the second test

khaneliman:
> Might be a good idea to warn for deprecation of old fish versions so that we can remove the workaround later

Makes me also wonder how necessary it is, tbh. If we're tracking upstream nixpkgs in this branch, we have a release soon and know that the stable branch has the workaround for the possible version mismatch. But, once we're past the release I would say it would be fine to remove it after a short period.

Weathercold (approved):
LGTM, I think the warning is not necessary

khaneliman:
Github refuses to let me do anything with this on my end. Rebase? 

image

feathecutie:
> Github refuses to let me do anything with this on my end. Rebase?

I saw the same issue, just rebased the changes on top of latest master and it seems to work now

PR #6695, opened 2025-03-24
zellij: Add additional options for integrating with shells; default integration disabled again

khaneliman:
### Description
Closes https://github.com/nix-community/home-manager/pull/6037
Follow up to discussion in https://github.com/nix-community/home-manager/pull/6358

Seems like interest was lost in that PR. Decisions made for restructuring from their approach:

- The environment variables mentioned, are only checked inside the `auto-start` scripts. It is also clearly described in their documentation. https://zellij.dev/documentation/integration.html
  - https://github.com/search?q=repo%3Azellij-org%2Fzellij%20ZELLIJ_AUTO_ATTACH&type=code
  - https://github.com/search?q=repo%3Azellij-org%2Fzellij+ZELLIJ_AUTO_EXIT&type=code

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/zellij.nix
+++ b/modules/programs/zellij.nix
@@ -6,7 +6,7 @@ let
   cfg = config.programs.zellij;
   yamlFormat = pkgs.formats.yaml { };
 in {
-  meta.maintainers = [ lib.hm.maintainers.mainrs ];
+  meta.maintainers = [ lib.maintainers.khaneliman lib.hm.maintainers.mainrs ];
 
   options.programs.zellij = {
     enable = lib.mkEnableOption "Zellij";
--- a/modules/programs/zellij.nix
+++ b/modules/programs/zellij.nix
@@ -8,18 +8,39 @@ let
   yamlFormat = pkgs.formats.yaml { };
   zellijCmd = getExe cfg.package;
 
+  autostartOnShellStartModule = types.submodule {
+    options = {
+      enable = mkEnableOption "" // {
+        description = ''
+          Whether to autostart Zellij session on shell creation.
+        '';
+      };
+
+      attachExistingSession = mkEnableOption "" // {
+        description = ''
+          Whether to attach to the default session after being autostarted if a Zellij session already exists.
+        '';
+      };
+
+      exitShellOnExit = mkEnableOption "" // {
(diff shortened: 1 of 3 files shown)


modules/programs/zellij.nix
```diff
@@ -1,32 +1,29 @@
...
-
 in {
-  meta.maintainers = [ hm.maintainers.mainrs ];
+  meta.maintainers = [ lib.maintainers.khaneliman lib.hm.maintainers.mainrs ];
 
   options.programs.zellij = {
-    enable = mkEnableOption "zellij";
+    enable = lib.mkEnableOption "Zellij";
 
     package = mkOption {
```
ambroisie, line 14: Drive-by comment: could be `mkPackageOption "Zellij" { default = "zellij"; }`

ambroisie (commented):
IMO the options should be renamed, rather than be called shell integration (and defaulted to `false`, unlike _all_ other shell integrations). As noted in the linked discussion, it's really an autostart feature rather than an integration.

khaneliman:
> IMO the options should be renamed, rather than be called shell integration (and defaulted to `false`, unlike _all_ other shell integrations). As noted in the linked discussion, it's really an autostart feature rather than an integration.

It is an integration, though. It's listed on their official website as an integration, too.

modules/programs/zellij.nix
khaneliman: Good catch, didn't notice it!

khaneliman:
> Should we add a `/modules/misc/news.nix` entry, since a previous entry mentioned that the default value of these option changed? It would be somewhat inconsistent if the news do not mention reverting this change. However, consider how minor the change is, it might be overkill to do this. I am fine with either approach.

Yeah, I wouldn't have, but forgot that news entry was created for that. Thanks for pointing that out!

trueNAHO (approved):
LGTM.

Adda0 (commented):
Sorry, I have been happily using the previous PR and forgot the PR was still waiting for my review. After our discussion in the previous PR, I checked the proposed changes and agree that your approach is the correct one. LGTM. Thank you for finishing the PR.

PR #6699, opened 2025-03-24
claude-code: add module

malob:
### Description

This PR adds support for Anthropic's Claude Code CLI to Home Manager.

Features:
  - Provides options to disable auto-updates (defaults to `true`)
    - `claude-code` tries to update at each start, which isn't compatible with Nix.
    - There is a setting, [`autoUpdaterStatus`](https://docs.anthropic.com/en/docs/agents-and-tools/claude-code/overview#global-configuration), which can be used to disable this behavior.
    - The natural way to handle this would be to have `home-manager` manage the settings file it uses (`~/.claude.json`). However, this file is updated frequently by the application with project specific settings, so managing it with `home-manager` isn't really an option.
    - As such, this module adds `claude config set -g autoUpdaterStatus disabled` to the activation script when `programs.claude-code.disableAutoUpdate` is set to `true` (which is the default).
  - Provides options to install optional runtime dependencies.

(This is my first PR to `home-manager`, apologies in advance if something about this PR doesn't match conventions etc.)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/claude-code.nix
+++ b/modules/programs/claude-code.nix
@@ -1,0 +1,75 @@
+{ config, lib, pkgs, ... }:
+
+let
+  inherit (lib) mkIf mkOption types;
+  cfg = config.programs.claude-code;
+in {
+  meta.maintainers = with lib.maintainers; [ malo ];
+
+  options.programs.claude-code = {
+    enable = lib.mkEnableOption "Claude Code CLI";
+
+    package = lib.mkPackageOption pkgs "claude-code" { };
+
+    disableAutoUpdate = mkOption {
+      type = types.bool;
+      default = true;
+      description = ''
+        Whether to disable the automatic update check on startup.
+        This is recommended when using home-manager to manage {command}`claude`.
+      '';
+    };
+
+    enableOptionalDependencies = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Whether to install the optional dependencies for {command}`claude`.
+        This includes:
+        - {command}`git` (for git operations)
+        - {command}`rg` (for code search)
+        - {command}`gh` (GitHub CLI, for GitHub operations)
+        - {command}`glab` (GitLab CLI, for GitLab operations)
+      '';
+    };
+
+    withGitHubCLI = mkOption {
+      type = types.bool;
+      default = cfg.enableOptionalDependencies;
+      description = ''
+        Whether to enable GitHub CLI ({command}`gh`) as a dependency.
+        This is useful if you work with GitHub repositories.
+
+        Defaults to the value of {option}`programs.claude-code.enableOptionalDependencies`.
+      '';
+    };
+
+    withGitLabCLI = mkOption {
+      type = types.bool;
+      default = cfg.enableOptionalDependencies;
+      description = ''
+        Whether to install GitLab CLI ({command}`glab`) as a dependency.
+        This is useful if you work with GitLab repositories.
+
+        Defaults to the value of {option}`programs.claude-code.enableOptionalDependencies`.
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ] ++ lib.optional cfg.withGitLabCLI pkgs.glab;
+
+    # Enable integrated modules for dependencies when needed
+    programs.git.enable = mkIf cfg.enableOptionalDependencies true;
+    programs.gh.enable = mkIf cfg.withGitHubCLI true;
+    programs.ripgrep.enable = mkIf cfg.enableOptionalDependencies true;
+
+    # Add activation script to disable auto-updates if the user wants that
+    home.activation.disableClaudeAutoUpdates = lib.mkIf cfg.disableAutoUpdate
+      (lib.hm.dag.entryAfter [ "writeBoundary" ] ''
+        $DRY_RUN_CMD ${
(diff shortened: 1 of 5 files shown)


modules/programs/claude-code.nix
```diff
@@ -0,0 +1,80 @@
...
+
+  options.programs.claude-code = {
+    enable = mkEnableOption "Claude Code CLI";
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.claude-code;
+      defaultText = literalExpression "pkgs.claude-code";
+      description = "Package providing {command}`claude`.";
+    };
```
khaneliman, line 17:
```suggestion
    package = lib.mkPackageOption pkgs { };
```

modules/programs/claude-code.nix
```diff
@@ -0,0 +1,80 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
```
khaneliman, line 3:
```suggestion
inherit (lib) mkIf mkOption types;
```

modules/programs/claude-code.nix
```diff
@@ -0,0 +1,80 @@
...
+
+let cfg = config.programs.claude-code;
+in {
+  meta.maintainers = with lib.maintainers; [ malo ];
+
+  options.programs.claude-code = {
+    enable = mkEnableOption "Claude Code CLI";
```
khaneliman, line 10:
```suggestion
    enable = lib.mkEnableOption "Claude Code CLI";
```

modules/programs/claude-code.nix
```diff
@@ -0,0 +1,80 @@
...
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.claude-code;
+      defaultText = literalExpression "pkgs.claude-code";
```
khaneliman, line 15:
```suggestion
      defaultText = lib.literalExpression "pkgs.claude-code";
```

modules/programs/claude-code.nix
```diff
@@ -0,0 +1,80 @@
...
+    # Enable integrated modules for dependencies when needed
+    programs.git.enable = mkIf cfg.enableOptionalDependencies true;
+    programs.gh.enable = mkIf cfg.withGitHubCLI true;
+    programs.ripgrep.enable = mkIf cfg.enableOptionalDependencies true;
+
+    # Add activation script to disable auto-updates if the user wants that
+    home.activation.disableClaudeAutoUpdates = lib.mkIf cfg.disableAutoUpdate
+      (lib.hm.dag.entryAfter [ "writeBoundary" ] ''
+        $DRY_RUN_CMD ${
+          getExe cfg.package
```
khaneliman, line 76:
```suggestion
          lib.getExe cfg.package
```

khaneliman (commented):
Could you also add `claude-code` to https://github.com/nix-community/home-manager/blob/b9da58d50551ebd74226fb8487b248a5a36c988f/tests/default.nix#L86

modules/programs/claude-code.nix
```diff
@@ -0,0 +1,75 @@
...
+    programs.ripgrep.enable = mkIf cfg.enableOptionalDependencies true;
+
+    # Add activation script to disable auto-updates if the user wants that
+    home.activation.disableClaudeAutoUpdates = lib.mkIf cfg.disableAutoUpdate
+      (lib.hm.dag.entryAfter [ "writeBoundary" ] ''
+        $DRY_RUN_CMD ${
+          lib.getExe cfg.package
+        } config set -g autoUpdaterStatus disabled || true
+      '');
+  };
```
khaneliman, line 74: Feels a little weird as a HM module that basically just auto enables other HM modules. 

The autoUpdaterStatus feels like something that should be patched upstream in nixpkgs, tbh.

malob: (No worries if this module is just too simple or something, such that you don't want to add it. I created it since I'd personally wanted the functionality it provides.)

> Feels a little weird as a HM module that basically just auto enables other HM modules.

Well depending what you mean by "auto enables" this might not quite be accurate. That is, `enableOptionalDependencies` defaults to `false`, so the user has to decide to change this option to `true` to enable those other modules. 

Personally, I like when modules like this give me the option to just install optional dependencies and keep track of what those are, so that I don't have to separately track whether I have them installed or not if I do what them. 

If for some reason it's preferable for this type of functionality to just add the relevant packages to `home.packages` as opposed to enabling the relevant `programs` module when available that seems fine to me. ¯\_(ツ)_/¯

> The autoUpdaterStatus feels like something that should be patched upstream in nixpkgs, tbh

I am the maintainer of the package in `nixpkgs`. Normally I would definitely just do this with a patch. The complication here is that the actual source code isn't available to patch. The "source" which comes for NPM, is basically a single ~6 MB minified JS file. I haven't actually tried, so maybe I should, but my assumption was that finding the relevant code patching it correctly would be annoying and probably brittle, such that updates to the package (which are frequent) will require creating a new patch some reasonable amount of the time.

malob: Just spent some time spelunking through the minified JS and I did find pretty easy fix to disable auto-updating upstream in `nixpkgs`, so maybe this module isn't worth it after all: https://github.com/NixOS/nixpkgs/pull/393488/commits/cb16b779e38e1a8ced2b9fc84aebddee8e48a3b8

PR #6707, opened 2025-03-25
home-manager: use --argstr for news file paths to handle special chars

### LINKED ISSUE #5279: bug: unexpected '@'

### Issue description

Hi,

How can I escape the '@' symbol in the home directory? I've tried single quote and backslash already, is there another character to use?

I'm using home-manager in an environment where the home directories are like this

```bash
/home/<username>@<domain>
```

The error I receive is this

```bash
error: syntax error, unexpected '@', expecting end of file

       at «string»:1:14:

            1| /home/<username>@<domain>
```

The only location I see to have it configured with the '@' is the `homeDirectory` field.

```bash
    configHomeManager = {
      system,
      username,
      ...
    }:
      home-manager.lib.homeManagerConfiguration {
        pkgs = pkgsImportSystem system;

        extraSpecialArgs = {
          inherit inputs;
          inherit system;
        };

        modules = [
          ./home/<username>
          {
            home = {
              homeDirectory = "/home/<username>@<domain>";
              stateVersion = "23.11";
            };
          }
        ];
      };
```

### Maintainer CC

_No response_

### System information

```markdown
  /nix/store/88hr9j7a1r9qrvv7xnfyjmd1ww459krl-nix-info
copying path '/nix/store/88hr9j7a1r9qrvv7xnfyjmd1ww459krl-nix-info' from 'https://cache.nixos.org'...
 - system: `"x86_64-linux"`
 - host os: `Linux 6.5.0-27-generic, Ubuntu, 22.04.4 LTS (Jammy Jellyfish), nobuild`
 - multi-user?: `yes`
 - sandbox: `yes`
 - channels(root): `"nixpkgs"`
 - channels(mduncan): `"nixos-23.05-23.05, nixpkgs"`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixpkgs
```

karypid:
Fix #5279 where paths with special characters such as @ break news handling.

### Description

This simply uses --argstr instead of --arg when invoking build-news.nix so that special characters such as '@' are properly evaluated.

See also discussion in https://discourse.nixos.org/t/62168/11

### Checklist

- [Y ] Change is backwards compatible.

- [ Y] Code formatted with `./format`.

- [ Y] Code tested through `nix-shell --pure tests -A run.all`

- [ N] Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).

- [Y ] Commit messages are formatted like

    ```
    {component}: {description}

    {long description}
    ```

  - [N ] Added myself as module maintainer. See [example](https://github.com/nix-community/home-manager/blob/068ff76a10e95820f886ac46957edcff4e44621d/modules/programs/lesspipe.nix#L6).

### DIFF
--- a/home-manager/home-manager
+++ b/home-manager/home-manager
@@ -791,8 +791,8 @@ function buildNews() {
     nix-instantiate \
         --no-build-output --strict \
         --eval '<home-manager/home-manager/build-news.nix>' \
-        --arg newsJsonFile "\"$(escapeForNix "$newsJsonFile")\"" \
-        --arg newsReadIdsFile "\"$(escapeForNix "$readIdsFile")\"" \
+        --argstr newsJsonFile "\"$(escapeForNix "$newsJsonFile")\"" \
+        --argstr newsReadIdsFile "\"$(escapeForNix "$readIdsFile")\"" \
         "${extraArgs[@]}" \
         > "$newsNixFile"
 }


home-manager/home-manager
```diff
@@ -791,8 +791,8 @@ function buildNews() {
     nix-instantiate \
         --no-build-output --strict \
         --eval '<home-manager/home-manager/build-news.nix>' \
-        --arg newsJsonFile "\"$(escapeForNix "$newsJsonFile")\"" \
-        --arg newsReadIdsFile "\"$(escapeForNix "$readIdsFile")\"" \
+        --argstr newsJsonFile "\"$(escapeForNix "$newsJsonFile")\"" \
+        --argstr newsReadIdsFile "\"$(escapeForNix "$readIdsFile")\"" \
```
jtojnar, line 795: This should be sufficient:

```suggestion
        --argstr newsReadIdsFile "$readIdsFile" \
```

` --argstr` will take the value as a string without any processing so it does not need to be quoted or escaped. Notice that the error shows extra quotes:

       error: string '"/var/folders/bz/h4njq8yd7td8qk42rt_9f0p00000gn/T/home-manager-build.vIGAkEVEQX/news.json"' doesn't represent an absolute path

jtojnar: Actually, https://github.com/nix-community/home-manager/commit/1395379a7a36e40f2a76e7b9936cc52950baa1be already fixed it. I missed it because GitHub search gave me an old revision.

PR #6710, opened 2025-03-26
podman: fix service name in generated manifest

Tarow:
### Description

If a podman container contains the string `podman-`, the activation will fail:
```nix
services.podman.containers."podman-exporter" = {
  image = "quay.io/navidys/prometheus-podman-exporter:latest";
};
```
The generated `~/.config/podman/containers.manifest` will only contain `exporter` and thus the non-existing container with that name is considered orphaned. Attempting to remove that container will fail:
```
Loading manifest from /home/niklas/.config/podman/containers.manifest...
Removing orphaned container: podman-exporter
Running: /nix/store/kixsa8zkgkgdhaaw94w97ibz4ljmajn7-podman-5.4.1/bin/podman container stop podman-exporter/nix/store/kixsa8zkgkgdhaaw94w97ibz4ljmajn7-podman-5.4.1/bin/podman container rm -f podman-exporter
Error: --filter takes no arguments
        Command failed: /nix/store/kixsa8zkgkgdhaaw94w97ibz4ljmajn7-podman-5.4.1/bin/podman container stop podman-exporter/nix/store/kixsa8zkgkgdhaaw94w97ibz4ljmajn7-podman-5.4.1/bin/podman container rm -f podman-exporter
Error: container is an invalid filter
```

The PR should only remove the generated  `podman-` prefix for the manifest file, not other occurrences of the `podman-` string inside the container name.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@n-hass @bamhm182

### DIFF
--- a/modules/services/podman-linux/podman-lib.nix
+++ b/modules/services/podman-linux/podman-lib.nix
@@ -1,18 +1,16 @@
 { pkgs, lib, config, ... }:
-
 with lib;
-
 let
   normalizeKeyValue = k: v:
     let
-      v' = (if builtins.isBool v then
+      v' = if builtins.isBool v then
         (if v then "true" else "false")
       else if builtins.isAttrs v then
         (concatStringsSep ''
 
           ${k}='' (mapAttrsToList normalizeKeyValue v))
       else
-        builtins.toString v);
+        builtins.toString v;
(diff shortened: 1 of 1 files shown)


n-hass (approved):
Thanks. lgtm

Issue #6704, opened 2025-03-25
bug: fzf zsh integration missing keybinds


CLEMENTINATOR:
### Are you following the right branch?

- [x] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Issue description

fzf zsh integration is missing keybinds:  
```
╭─clement@brittle-hollow ~ 
╰─$ bindkey | grep fzf
"^T" fzf-file-widget
"^[c" fzf-cd-widget
╭─clement@brittle-hollow ~ 
╰─$ source <(fzf --zsh)
╭─clement@brittle-hollow ~ 
╰─$ bindkey | grep fzf 
"^I" fzf-completion
"^R" fzf-history-widget
"^T" fzf-file-widget
"^[c" fzf-cd-widget
```
I think its due to this [commit](https://github.com/nix-community/home-manager/commit/10deb9d043e625d8c6a629f5575718e0768dcfb4#diff-931b1bd0225f6a4c6f535651bd7e920217f5f8ba09281d5f1811b30a06020c33) merged 3 days ago 

seems like the added line by enableZshIntegration 
```
if [[ $options[zle] = on ]]; then
  eval "$(/nix/store/8p7nggqkcdz1spffg99hlgdl9pyawcxr-fzf-0.60.3/bin/fzf --zsh)"
fi
```
is setup at the top of .zshrc, and the Ctrl+R keybinding is rewritten afterwards by the standard zsh keybind
```
"^R" history-incremental-search-backward
```

### System information

```markdown
- system: `"x86_64-linux"`
- host os: `Linux 6.13.7, NixOS, 25.05 (Warbler), 25.05.20250322.1e5b653`
- multi-user?: `yes`
- sandbox: `yes`
- version: `nix-env (Nix) 2.24.12`
- nixpkgs: `/nix/store/fnc4632ncr5k5la19gp2gb255dx8pi9x-source`
```

teto:
I dont see the bug here but my setup is non standard. Patch welcome, According to your report. the `mkOrder X` should be changed ?

musjj:
Having the same issue too.

CLEMENTINATOR:
> I dont see the bug here but my setup is non standard. Patch welcome, According to your report. the `mkOrder X` should be changed ?

[...]

PR #6712, opened 2025-03-27
fzf: fix zsh integration (keybinds rewritten by omz)

CLEMENTINATOR:
### Description

fzf history keybind (^R) was being rewritten by omz.
This PR moves the initialization of the fzf zsh integration just after omz integration (but sill before other history managers, eg: atuin)

Fixes https://github.com/nix-community/home-manager/issues/6704

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/fzf.nix
+++ b/modules/programs/fzf.nix
@@ -195,8 +195,10 @@ in {
     # Note, since fzf unconditionally binds C-r we use `mkOrder` to make the
     # initialization show up a bit earlier. This is to make initialization of
     # other history managers, like mcfly or atuin, take precedence.
+    # Still needs to be initialized after oh-my-zsh (order 800), otherwise
+    # omz will take precedence.
     programs.zsh.initContent =
-      mkIf cfg.enableZshIntegration (mkOrder 200 zshIntegration);
+      mkIf cfg.enableZshIntegration (mkOrder 810 zshIntegration);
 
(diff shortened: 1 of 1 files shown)


modules/programs/fzf.nix
```diff
@@ -195,8 +195,10 @@ in {
     # Note, since fzf unconditionally binds C-r we use `mkOrder` to make the
     # initialization show up a bit earlier. This is to make initialization of
     # other history managers, like mcfly or atuin, take precedence.
+    # Still needs to be initialized after oh-my-zsh (order 800), otherwise
+    # omz will take precedence.
     programs.zsh.initContent =
-      mkIf cfg.enableZshIntegration (mkOrder 200 zshIntegration);
+      mkIf cfg.enableZshIntegration (mkOrder 801 zshIntegration);
```
craigfurman, line 201: nit (and also I'm not a maintainer): if OMZ has prio 800, maybe we should pick a slightly lower prio (larger number) here (e.g. 808 or 810), to leave the door open for slotting things in between?

I hope we wouldn't have to actually use that functionality, of course 😅 

Thanks so much for making this PR!

CLEMENTINATOR: good point, ill update to 810

teto:
summoning our zsh expert @khaneliman

PR #6575, opened 2025-03-06
helix: add extraConfig option

nouritsu:
### Description

Add an extraConfig option for Helix. Fixes #6513

### Checklist

- [X] Change is backwards compatible.

- [X] Code formatted with `./format`.

- [X] Code tested through `nix-shell --pure tests -A run.all`
    or `nix build --reference-lock-file flake.lock ./tests#test-all` using Flakes.
**NOTE: I am assuming the tests pass, as I am facing an unrelated error on both of my machines. All the tests in the actions pass.**

- [X] Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).

- [X] Commit messages are formatted like

    ```
    {component}: {description}

    {long description}
    ```

    See [CONTRIBUTING](https://nix-community.github.io/home-manager/#sec-commit-style) for more information and [recent commit messages](https://github.com/nix-community/home-manager/commits/master) for examples.

- If this PR adds a new module

  - [ ] Added myself as module maintainer. See [example](https://github.com/nix-community/home-manager/blob/068ff76a10e95820f886ac46957edcff4e44621d/modules/programs/lesspipe.nix#L6).

PR #6714, opened 2025-03-27
helix: avoid IFD

pineapplehunter:
### Description

Avoids IFD in the helix module which was introduced in #6575.
The helix module fails to build when `--no-allow-import-from-derivation` is enabled.

Command to reproduce the results:

```
nix build --reference-lock-file flake.lock --no-allow-import-from-derivation ./tests#test-helix-example-settings
```

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@Philipp-M

### DIFF
--- a/modules/programs/helix.nix
+++ b/modules/programs/helix.nix
@@ -212,9 +212,13 @@ in {
     xdg.configFile = let
       settings = {
         "helix/config.toml" = mkIf (cfg.settings != { }) {
-          text =
-            builtins.readFile (tomlFormat.generate "helix-config" cfg.settings)
-            + "\n" + cfg.extraConfig;
+          source = let
+            configFile = tomlFormat.generate "config.toml" cfg.settings;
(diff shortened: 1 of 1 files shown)


nouritsu:
Hello! I was not aware that this would be a problem. I did some reading and the option `--no-allow-import-from-derivation` disables realisation of files in the Nix store. This is used to speed up rebuild times. Am I correct or are there any other reasons why people use this option? I will ensure I dont repeat the same mistake in my following contributions.

ambroisie (approved):

pineapplehunter:
@nouritsu I have `allow-import-from-derivation = false` in `nix.conf` because the nixpkgs repo does not allow import from derivation. https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#import-from-derivation
I do not think you need to avoid IDF in personal projects, but it is helpful on nixpkgs.

PR #6715, opened 2025-03-27
feat(bat): interactive init for `batpipe` and `batman`

minhtrancccp:
### Description

let `batpipe` and `batman` set up useful env vars for interactive shell
sessions if declared in `programs.bat.extraPackages`, as demonstrated in
upstream nixpkgs

this commit is adapted from the nixpkgs equivalent as it was seen here:
https://github.com/NixOS/nixpkgs/blob/698214a32beb4f4c8e3942372c694f40848b360d/nixos/modules/programs/bat.nix

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@khaneliman

### DIFF
--- a/modules/programs/bat.nix
+++ b/modules/programs/bat.nix
@@ -1,6 +1,8 @@
 { config, lib, pkgs, ... }:
 let
-  inherit (lib) literalExpression mkEnableOption mkOption mkIf types;
+  inherit (builtins) elem;
+  inherit (lib)
+    getExe literalExpression mkEnableOption mkOption mkIf optionalString types;
 
   cfg = config.programs.bat;
 
@@ -18,6 +20,26 @@ let
         --${k}
       '') (attrNames enabledBoolFlags);
     in keyValuePairs + switches;
+
+  initScript = { program, shell, flags ? [ ], }:
+    if (shell != "fish") then ''
+      eval "$(${getExe program} ${toString flags})"
+    '' else ''
+      ${getExe program} ${toString flags} | source
+    '';
+
+  shellInit = shell:
+    optionalString (elem pkgs.bat-extras.batpipe cfg.extraPackages)
+    (initScript {
+      program = pkgs.bat-extras.batpipe;
+      inherit shell;
+    }) + optionalString (elem pkgs.bat-extras.batman cfg.extraPackages)
+    (initScript {
+      program = pkgs.bat-extras.batman;
+      inherit shell;
+      flags = [ "--export-env" ];
+    });
+
 in {
   meta.maintainers = with lib.maintainers; [ khaneliman ];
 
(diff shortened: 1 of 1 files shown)


modules/programs/bat.nix
```diff
@@ -167,6 +189,19 @@ in {
...
+
+      programs = {
+        bash =
+          mkIf (!config.programs.fish.enable) { initExtra = shellInit "bash"; };
+        fish = mkIf config.programs.fish.enable {
+          interactiveShellInit = shellInit "fish";
+        };
+        zsh =
+          mkIf (!config.programs.fish.enable && config.programs.zsh.enable) {
+            initExtra = shellInit "zsh";
```
khaneliman, line 201:
```suggestion
            initContent = shellInit "zsh";
```

modules/programs/bat.nix
```diff
@@ -167,6 +189,19 @@ in {
...
       '';
+
+      programs = {
+        bash =
+          mkIf (!config.programs.fish.enable) { initExtra = shellInit "bash"; };
+        fish = mkIf config.programs.fish.enable {
+          interactiveShellInit = shellInit "fish";
+        };
+        zsh =
+          mkIf (!config.programs.fish.enable && config.programs.zsh.enable) {
```
khaneliman, line 200: Why does it matter whether fish is enabled?

modules/programs/bat.nix
```diff
@@ -167,6 +189,19 @@ in {
           run ${lib.getExe cfg.package} cache --build
         )
       '';
+
+      programs = {
+        bash =
+          mkIf (!config.programs.fish.enable) { initExtra = shellInit "bash"; };
```
khaneliman, line 195: same question here

modules/programs/bat.nix
minhtrancccp: i don't have the answer unfortunately, since i was adapting from [the module over nixpkgs](https://github.com/NixOS/nixpkgs/pull/350079) as written by @sigmasquadron

SigmaSquadron: IIRC, that was done because if `fish` is enabled, `pay-respects` will generate `fish` code when generating completions for `bash`, which will naturally cause a syntax error. I don't know why this happens, but since I use `fish` myself, I didn't care enough to investigate.

SigmaSquadron: oh, wow. my brain might have died there. I'm usually pinged here because of `pay-respects`, so I completely missed this was about `bat`. 

Anyways, I have no idea why that's on the NixOS bat module. That's my bad.

minhtrancccp:
as i look into the original scripts, it seems like [`batpipe` is shell-sensitive when generating its init](https://github.com/eth-p/bat-extras/blob/36c77c171cc71b2ff3ec4cb781aa16ca3ad258b1/src/batpipe.sh#L56-L110), [but `batman` is not](https://github.com/eth-p/bat-extras/blob/36c77c171cc71b2ff3ec4cb781aa16ca3ad258b1/src/batman.sh#L69-L80). 

should i extract [`batman` from `shellInit`](https://github.com/minhtrancccp/home-manager/blob/84fb0f0e045fea09c46b03f3e1c279e2f0550cc2/modules/programs/bat.nix#L36-L40), and set `home.sessionVariables` directly?

as an example, this is how i would accomplish that:
```nix
home.sessionVariables =
	lib.attrsets.optionalAttrs (builtins.elem pkgs.bat-extras.batman cfg.extraPackages) {
		MANPAGER = "${lib.meta.getExe' pkgs.coreutils "env"} BATMAN_IS_BEING_MANPAGER=yes ${lib.meta.getExe' pkgs.bat-extras.batman ".batman-wrapped"}";
		MANROFFOPT = "-c";
	};
```

PR #6716, opened 2025-03-27
fzf: update zsh integration to be after plugins

CLEMENTINATOR:
### Description

ensure that fzf is initialized after plugins (fixes https://github.com/nix-community/home-manager/issues/6704#issuecomment-2758808689)

we could also add a config to disable plugin lazy loading 🤔 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@khaneliman

### DIFF
--- a/modules/programs/fzf.nix
+++ b/modules/programs/fzf.nix
@@ -198,7 +198,7 @@ in {
     # Still needs to be initialized after oh-my-zsh (order 800), otherwise
     # omz will take precedence.
     programs.zsh.initContent =
-      mkIf cfg.enableZshIntegration (mkOrder 810 zshIntegration);
+      mkIf cfg.enableZshIntegration (mkOrder 910 zshIntegration);
 
     programs.fish.interactiveShellInit =
       mkIf cfg.enableFishIntegration (mkOrder 200 fishIntegration);


khaneliman:
Thanks, looks like this shuffles it back to the prio right before `initExtra` used to be https://github.com/nix-community/home-manager/commit/296ddc64627f4a6a4eb447852d7346b9dd16197d?diff=unified. This will keep it after most of the HM declarations and allow an `initContent` from a user to override it, if someone wants to include something in their config to override the behavior.

modules/programs/fzf.nix
```diff
@@ -198,7 +198,7 @@ in {
     # Still needs to be initialized after oh-my-zsh (order 800), otherwise
     # omz will take precedence.
     programs.zsh.initContent =
-      mkIf cfg.enableZshIntegration (mkOrder 810 zshIntegration);
+      mkIf cfg.enableZshIntegration (mkOrder 910 zshIntegration);
```
ambroisie, line 201: Could maybe use a comment similar to the one about `oh-my-zsh`?

CLEMENTINATOR: yeah i sorta forgot to update the comment :|

PR #6717, opened 2025-03-28
newsboat: add a package option to the module

ipsavitsky:
### Description

The newboat module did not have a package option. 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@sumnerevans

### DIFF
--- a/modules/programs/newsboat.nix
+++ b/modules/programs/newsboat.nix
@@ -39,6 +39,8 @@ in {
     programs.newsboat = {
       enable = mkEnableOption "the Newsboat feed reader";
 
+      package = mkPackageOption pkgs "newsboat" { nullable = true; };
+
       urls = mkOption {
         type = types.listOf (types.submodule {
           options = {
@@ -132,7 +134,7 @@ in {
       '';
     }];
 
-    home.packages = [ pkgs.newsboat ];
+    home.packages = lib.mkIf (cfg.package != null) [ cfg.package ];
 
     # Use ~/.newsboat on stateVersion < 21.05 and use ~/.config/newsboat for
     # stateVersion >= 21.05.


modules/programs/newsboat.nix
```diff
@@ -39,6 +39,13 @@ in {
     programs.newsboat = {
       enable = mkEnableOption "the Newsboat feed reader";
 
+      package = mkOption {
+        type = types.package;
+        default = pkgs.newsboat;
+        defaultText = literalExpression "pkgs.newboat";
+        description = "The newsboat package to install";
+      };
+
```
khaneliman, line 48:
```suggestion
      package = lib.mkPackageOption pkgs "newsboat";

```

modules/programs/newsboat.nix
```diff
@@ -132,7 +139,7 @@ in {
       '';
     }];
 
-    home.packages = [ pkgs.newsboat ];
+    home.packages = [ cfg.package ];
```
khaneliman, line 142:
```suggestion
    home.packages = lib.mkIf (cfg.package != null) [ cfg.package ];
```

modules/programs/newsboat.nix
ipsavitsky: Fixed, but I left the nullable option out (see my next comment).

modules/programs/newsboat.nix
ipsavitsky: I may be understanding wrong, but our intention is to have an option that will automatically set a package passed, but if no package is passed have a default to fall back to. IIUC this will not set a package if no package is passed, which, I think, is not what this does.

khaneliman: The idea is to have an option that allows configuring the package, if desired, to be installed. If not, just apply a configuration.

ipsavitsky: Ah, fair enough, I get it now. Change pushed.

Issue #3127, opened 2022-08-05
bug: home-manager-auto-upgrade.service is failing


imincik:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

`home-manager-auto-upgrade.service` is failing with following error message when executed by `home-manager-auto-upgrade.timer`.

```
$ systemctl --user status home-manager-auto-upgrade.service

home-manager-auto-upgrade.service - Home Manager upgrade
     Loaded: loaded (/home/imincik/.config/systemd/user/home-manager-auto-upgrade.service; linked; vendor preset: enab>
     Active: failed (Result: exit-code) since Thu 2022-08-04 12:25:10 CEST; 1 day 4h ago
TriggeredBy: ○ home-manager-auto-upgrade.timer
   Main PID: 1157762 (code=exited, status=127)
        CPU: 120ms

aug 04 12:25:09 imincik-nx systemd[2344]: Started Home Manager upgrade.
aug 04 12:25:09 imincik-nx 8fgs2az0z3hw1c8j46fkcynkcqb0mhq3-home-manager-auto-upgrade[1157762]: Update Nix's channels
aug 04 12:25:10 imincik-nx 8fgs2az0z3hw1c8j46fkcynkcqb0mhq3-home-manager-auto-upgrade[1157769]: unpacking channels...
aug 04 12:25:10 imincik-nx 8fgs2az0z3hw1c8j46fkcynkcqb0mhq3-home-manager-auto-upgrade[1157762]: Upgrade Home Manager
aug 04 12:25:10 imincik-nx 8fgs2az0z3hw1c8j46fkcynkcqb0mhq3-home-manager-auto-upgrade[1157850]: /nix/store/zjcpc7dp2k1>
aug 04 12:25:10 imincik-nx systemd[2344]: home-manager-auto-upgrade.service: Main process exited, code=exited, status=>
aug 04 12:25:10 imincik-nx systemd[2344]: home-manager-auto-upgrade.service: Failed with result 'exit-code'.
```

BUT executing script called by this service by hand (in my case `/nix/store/8fgs2az0z3hw1c8j46fkcynkcqb0mhq3-home-manager-auto-upgrade`) works as expected - with no error, only reporting that `home-manager-auto-upgrade.service` is failing which is the point of this bug report.

```
$ cat .config/systemd/user/home-manager-auto-upgrade.service
[Service]
[...]

PR #6719, opened 2025-03-28
auto-upgrade: unbreak on unattended, loginctl enable-linger systems

Ekleog:
Fixes https://github.com/nix-community/home-manager/issues/3127

Tested on my servers with the following configuration:
```
  services.home-manager.autoUpgrade = {
    enable = true;
    frequency = "03:40";
  };
  # TODO: upstream to home-manager
  systemd.user.services.home-manager-auto-upgrade.Service.ExecStart = lib.mkForce ("${pkgs.writeShellApplication {
    name = "home-manager-auto-upgrade";
    text = ''
      echo "Update Nix's channels"
      nix-channel --update
      echo "Upgrade Home Manager"
      home-manager switch
    '';
    runtimeInputs = with pkgs; [
      nix
      (callPackage <home-manager/home-manager> { path = config.programs.home-manager.path; })
    ];
  }}/bin/home-manager-auto-upgrade");
```
By running:
```
home-manager switch
systemctl --user start home-manager-auto-upgrade.service
journalctl --user -u home-manager-auto-upgrade.service -f
```
And then confirming that everything seemed to be working fine.
Disclaimer: I haven't waited 24 hours yet so the auto-run didn't happen yet, but I don't see why it would fail. I'll make sure to come report here if it does fail, as I have alerting setup for failing systemd services.

Code tested 

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

cc @pinage404

### DIFF
--- a/modules/services/home-manager-auto-upgrade.nix
+++ b/modules/services/home-manager-auto-upgrade.nix
@@ -8,6 +8,16 @@ let
     path = config.programs.home-manager.path;
   };
 
+  autoUpgradeApp = pkgs.writeShellApplication {
+    name = "home-manager-auto-upgrade";
+    text = ''
+      echo "Update Nix's channels"
+      nix-channel --update
+      echo "Upgrade Home Manager"
+      home-manager switch
+    '';
+    runtimeInputs = with pkgs; [ homeManagerPackage nix ];
+  };
 in {
   meta.maintainers = [ lib.hm.maintainers.pinage404 ];
 
@@ -52,14 +62,7 @@ in {
 
       services.home-manager-auto-upgrade = {
         Unit.Description = "Home Manager upgrade";
-
-        Service.ExecStart = toString
-          (pkgs.writeShellScript "home-manager-auto-upgrade" ''
-            echo "Update Nix's channels"
-            ${pkgs.nix}/bin/nix-channel --update
-            echo "Upgrade Home Manager"
-            ${homeManagerPackage}/bin/home-manager switch
-          '');
+        Service.ExecStart = "${autoUpgradeApp}/bin/home-manager-auto-upgrade";
       };
     };
   };


pinage404 (approved):
I haven't tested

LGTM

I don't use it since i moved to Nix Flake, you can remove me as the maintainer

Ekleog:
Thank you!
Out of curiosity, do you know how the rest of the process for landing looks like? This is my first contribution to home-manager since quite a few years ago and the process seems to have changed 😅

Ekleog:
Thank you!
Out of curiosity, do you know how the rest of the process for landing looks like? This is my first contribution to home-manager since quite a few years ago and the process seems to have changed 😅

pinage404:
I don't know, it took me 6 months to get [my PR merged](https://github.com/nix-community/home-manager/pull/2162), which made me not want to do other PRs

Ekleog:
Oh. Well, thank you 🥲 

cc @rycee for landing maybe if only you still have access to this repo, then?

khaneliman:
Is the point of the change just to add the runtime inputs to the systemd service environment?

Ekleog:
Yes exactly! Without this the upgrade fails with this message:
```
[...]-home-manager/bin/home-manager: ligne 539: nix-build : commande introuvable
```

Ekleog:
Environment could be an alternative; I used `writeShellApplication` because it already has all the magic stuff to make the proper path without having to manually `makeBinPath`, which is fine for now but if it were to grow more complex it could become problematic

khaneliman:
> Environment could be an alternative; I used `writeShellApplication` because it already has all the magic stuff to make the proper path without having to manually `makeBinPath`, which is fine for now but if it were to grow more complex it could become problematic

Fair enough, it can be tweaked later, if desired.

Ekleog:
Awesome, thank you for the quick merge :purple_heart:

Ekleog:
I just rebuilt on top of that and can confirm things work fine now :tada:

PR #6720, opened 2025-03-28
zsh-abbr: Add global abbreviations

kawarimidoll:
### Description

zsh-abbr has a global abbreviations feature. The global abbreviations expand anywhere on the line and are defined by using the `-g` option.

https://zsh-abbr.olets.dev/types.html#global

However, home-manager did not have this setting, so I added it in this PR.
Following the naming of `zsh.shellGlobalAliases`, I named it `zsh-abbr.globalAbbreviations`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/zsh/zsh-abbr.nix
+++ b/modules/programs/zsh/zsh-abbr.nix
@@ -24,9 +24,32 @@ in {
         the longer phrase after they are entered.
       '';
     };
+
+    globalAbbreviations = mkOption {
+      type = types.attrsOf types.str;
+      default = { };
+      example = {
+        G = "| grep";
+        L = "| less -R";
+      };
+      description = ''
+        Similar to [](#opt-programs.zsh.zsh-abbr.abbreviations),
+        but are expanded anywhere on a line.
+      '';
+    };
   };
 
-  config = mkIf cfg.enable {
+  config = let
+    abbreviations =
+      mapAttrsToList (k: v: "abbr ${escapeShellArg k}=${escapeShellArg v}")
+      cfg.abbreviations;
+
+    globalAbbreviations =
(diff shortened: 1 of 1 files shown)


modules/programs/zsh/zsh-abbr.nix
```diff
@@ -36,7 +49,9 @@ in {
     xdg.configFile = {
       "zsh-abbr/user-abbreviations".text = concatStringsSep "\n"
         (mapAttrsToList (k: v: "abbr ${escapeShellArg k}=${escapeShellArg v}")
-          cfg.abbreviations) + "\n";
+          cfg.abbreviations ++ mapAttrsToList
+          (k: v: "abbr -g ${escapeShellArg k}=${escapeShellArg v}")
+          cfg.globalAbbreviations) + "\n";
```
ambroisie, line 54: Could be more readable this way:

```suggestion
      "zsh-abbr/user-abbreviations".text = 
      let
        abbreviations = mapAttrsToList
          (k: v: "abbr ${escapeShellArg k}=${escapeShellArg v}")
          cfg.abbreviations;
        globalAbbreviations = mapAttrsToList
          (k: v: "abbr -g ${escapeShellArg k}=${escapeShellArg v}")
          cfg.globalAbbreviations;
        allAbbreviations = abbreviations ++ globalAbbreviations;
      in
      concatStringsSep "\n" allAbbreviations + "\n";

```

kawarimidoll:
I force-pushed to avoid leaving format changes in the log.

modules/programs/zsh/zsh-abbr.nix
kawarimidoll: I agree. However, when I followed your instructions, the format script broke the structure. So, I moved `let...in` block. How does this look? f8e5f87

ambroisie (approved):

PR #6722, opened 2025-03-28
granted: Add override for package

fzakaria:
I would like to use the unstable version in my home-manager so let's make it possible to override.

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

 
 @wcarlsen

### DIFF
--- a/modules/programs/granted.nix
+++ b/modules/programs/granted.nix
@@ -5,13 +5,19 @@ with lib;
 let
 
   cfg = config.programs.granted;
-  package = pkgs.granted;
 
 in {
   meta.maintainers = [ hm.maintainers.wcarlsen ];
 
   options.programs.granted = {
     enable = mkEnableOption "granted";
+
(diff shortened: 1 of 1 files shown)


modules/programs/granted.nix
```diff
@@ -5,14 +5,20 @@ with lib;
...
   options.programs.granted = {
     enable = mkEnableOption "granted";
 
+    package = mkOption {
+      type = types.package;
+      default = pkgs.granted;
+      defaultText = literalExpression "pkgs.granted";
+      description = "The granted package to install.";
+    };
+
```
khaneliman, line 21:
```suggestion
     package = lib.mkPackageOption pkgs "granted" { };
```

khaneliman:
sorry, had extra space in there. needs a `nix fmt`

fzakaria:
Ty for fixing for me.

PR #6735, opened 2025-03-31
programs: remove with lib 

khaneliman:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```


ambroisie:
It looks like you mixed explicit `lib.` prefixing _and_ `inherit`s. Is there a reason for using one or the other, rather than stick to e.g: `lib.*` everywhere?

khaneliman:
> I believe this change caused this because I tested previous commit ([ccd7df8](https://github.com/nix-community/home-manager/commit/ccd7df836e1f42ea84806760f25b77b586370259)) and there wasn't an error.
> 
> image

Thanks, I’ll fix it.

khaneliman:
Should be fixed with https://github.com/nix-community/home-manager/pull/6744

PR #6736, opened 2025-03-31
mcfly: Fix fzf overriding mcfly's Ctrl+R bind

wyyllou:
### Description
In the fzf module, there is a comment about overriding the order to prevent this from happening:
```nix
    # Note, since fzf unconditionally binds C-r we use `mkOrder` to make the
    # initialization show up a bit earlier. This is to make initialization of
    # other history managers, like mcfly or atuin, take precedence.

    programs.fish.interactiveShellInit =
      mkIf cfg.enableFishIntegration (mkOrder 200 fishIntegration);
```

However it does not work, and mcfly appears in a separate block in config.fish, *before* fzf, since mcfly is not using interactiveShellInit, meaning the ctrl+r bind is the fzf bind when the intended behavior is to prioritise mcfly. Example:
```fish
# config.fish
/nix/store/anzmx9jfbl2gf2m2ivy5fnavl99kf13c-mcfly-0.9.3/bin/mcfly init fish | source
if status is-interactive
    eval "$(/nix/store/80l3p88i1kkbskf6zq14yz4kg520rbv5-mcfly-fzf-0.1.3/bin/mcfly-fzf init fish)"
end
# ...
status is-interactive; and begin

    # Abbreviations

    # Aliases

    # Interactive shell initialisation
    /nix/store/8p7nggqkcdz1spffg99hlgdl9pyawcxr-fzf-0.60.3/bin/fzf --fish | source
end
```
This commit turns that into:
```fish
status is-interactive; and begin
    # ...
    
    # Interactive shell initialisation
    /nix/store/8p7nggqkcdz1spffg99hlgdl9pyawcxr-fzf-0.60.3/bin/fzf --fish | source
   
    # ...

    /nix/store/anzmx9jfbl2gf2m2ivy5fnavl99kf13c-mcfly-0.9.3/bin/mcfly init fish | source
    eval "$(/nix/store/80l3p88i1kkbskf6zq14yz4kg520rbv5-mcfly-fzf-0.1.3/bin/mcfly-fzf init fish)"
end
```

And that displays the intended behavior in shell usage.

This commit moves both mcfly and mcfly-fzf into the main is-interactive block, which is technically unneccesary as the checks are done internally (but mcfly-fzf is broken in this regard https://github.com/bnprks/mcfly-fzf/issues/10), but the only other, slightly less elegant way to do this is adding another is-interactive block at the end of the config using `shellInitLast`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/mcfly.nix
+++ b/modules/programs/mcfly.nix
@@ -23,9 +23,7 @@ let
   fishIntegration = ''
     ${getExe pkgs.mcfly} init fish | source
   '' + optionalString cfg.fzf.enable ''
-    if status is-interactive
-      eval "$(${getExe pkgs.mcfly-fzf} init fish)"
-    end
+    eval "$(${getExe pkgs.mcfly-fzf} init fish)"
   '';
 
   zshIntegration = ''
@@ -143,7 +141,8 @@ in {
 
       programs.zsh.initContent = mkIf cfg.enableZshIntegration zshIntegration;
 
-      programs.fish.shellInit = mkIf cfg.enableFishIntegration fishIntegration;
+      programs.fish.interactiveShellInit =
+        mkIf cfg.enableFishIntegration fishIntegration;
 
       home.sessionVariables.MCFLY_KEY_SCHEME = cfg.keyScheme;
 


khaneliman:
This was changed in https://github.com/nix-community/home-manager/pull/6669 @mythmon

modules/programs/mcfly.nix
```diff
@@ -143,7 +141,8 @@ in {
 
       programs.zsh.initContent = mkIf cfg.enableZshIntegration zshIntegration;
 
-      programs.fish.shellInit = mkIf cfg.enableFishIntegration fishIntegration;
+      programs.fish.interactiveShellInit =
```
khaneliman, line 144: This seems to make sense, though.

wyyllou:
> This was changed in https://github.com/nix-community/home-manager/pull/6669 @mythmon 

Yes, i saw that, but the is-interactive block does not need to be there if it is moved into the interactiveShellInit.

mythmon:
TIL about `interactiveShellInit`, thanks for the fix.

Upstream recently released 1.0.4 that should have a fix for the original issue. https://github.com/bnprks/mcfly-fzf/releases/tag/1.0.4. Once that update is available in nix, would both this PR and the original be obsolete, or should we keep the change in this PR?

wyyllou:
> Would both this PR and the original be obsolete, or should we keep the change in this PR?

After that version lands everything will still work but to avoid extra is-interactive checks, i think the `fishIntegration` could be moved to `shellInitLast` just to make sure it is still after the fzf initialisation.

Issue #6724, opened 2025-03-29
Bug Report: `enableFishIntegration = true;` Causes `File name too long` Error in Yazi Home Manager


phucleeuwu:
## Summary
When enabling Fish shell integration for Yazi in Home Manager with the following configuration:

```nix
programs.yazi = {
  enable = true;
  enableFishIntegration = true;
  shellWrapperName = "y";
};
```

the following error occurs:

```
cd: File name too long
cd: Unknown error trying to locate directory
“───────┬────────────────────────────────────────────────────────────────────────
        │ File: /var/folders/xx/rkm835zx78b4z1n3_wbxv6ww0000gn/T/yazi-cwd.XXX
        │ XX.huRUF8y1j5
        │ Size: 42 B
───────┼────────────────────────────────────────────────────────────────────────
    1   │ /Users/phuc/nixos-config/.github/workflows
───────┴────────────────────────────────────────────────────────────────────────”
```

## Steps to Reproduce
1. Enable Yazi in Home Manager with Fish integration and set `shellWrapperName = "y";`.
2. Open a Fish shell and `y` to use yazi.
3. Try to go into a different directory with Yazi.
4. After exit yazi observe the error message regarding file name length.

## Expected Behavior
- The `cd` command should function correctly without errors.
- The integration should not create an excessively long or malformed path.

## Actual Behavior
- `cd` fails with a `File name too long` error.
- The path appears to be improperly formatted and overly long.

## System Information
- **OS**: macOS (Darwin)
- **Shell**: Fish
- **Home Manager**: Latest with Nix-Darwin modules
- **Yazi Version**: Latest

## Additional Notes
- This issue does not occur when `enableFishIntegration = false;`.
- Other shells (e.g., Zsh, Bash) may not exhibit the same issue.
- The error suggests a problem with how Yazi writes its temporary directory or handles Fish integration.

## Possible Workarounds
- Disabling Fish integration temporarily.
- Using a different `shellWrapperName`.

## Suggestions for Fix
- Investigate how Yazi manages the Fish shell wrapper and its temporary files.
- Check if `yazi-cwd.XXX` is being generated correctly.

---
Any insights or fixes would be appreciated!

khaneliman:
Thanks for a well defined bug report, I don't use fish atm. But, if no one gets to it and I find time. I can try looking into it.

phucleeuwu:
[...]

PR #6739, opened 2025-04-01
Update yazi.nix

phucleeuwu:
### Description
fix https://github.com/nix-community/home-manager/issues/6724


khaneliman (changes requested):
The place this string gets inserted, will automatically wrap it as a function based on the `cfg.shellWrapperName`. The only real change from the documentation, is using `command cat` and possibly `set tmp`

PR #6747, opened 2025-04-02
news: create an individual file for each news entry

thiagokokada:
### Description

The current way to define a news entry in Home-Manager is error prone (since you need to type the date manually) and also it is common cause of conflicts after merges because all entries are defined in the same file.

This commit fixes this: we can now create individual news entries for each new entry. A script `create-news-entry.sh` also helps to create it in the correct format (with the correct filenames and structure).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/news.nix
+++ b/modules/misc/news.nix
@@ -45,6 +45,13 @@ let
       id = lib.mkDefault (builtins.hashString "sha256" config.message);
     };
   });
+
+  isNixFile = n: v: v == "regular" && lib.hasSuffix ".nix" n;
+  # builtins.attrNames return the values in alphabetical order
+  newsFiles =
+    builtins.attrNames (lib.filterAttrs isNixFile (builtins.readDir ./news));
+  newsEntries =
+    builtins.map (newsFile: import (./news + "/${newsFile}")) newsFiles;
 in {
   meta.maintainers = [ lib.maintainers.rycee ];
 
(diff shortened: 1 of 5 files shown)


thiagokokada (commented):
CC @khaneliman for review.

thiagokokada:
Ok, tests are passing now. Not sure what caused the previous issue but I don't need to move the files to add the change.

CC @khaneliman for review.

khaneliman (approved):
LGTM, this will be super helpful and I know @teto might have an interest in it

thiagokokada (commented):
Rebased with the current `master` changes (ironically, there are some news changes in `master` 😅).

Also did some small improvements in the `create-news-entry.sh` script.

modules/misc/news.nix
```diff
@@ -45,6 +45,13 @@ let
       id = lib.mkDefault (builtins.hashString "sha256" config.message);
     };
   });
+
+  isNixFile = n: v: v == "regular" && lib.hasSuffix ".nix" n;
```
thiagokokada, line 49: Changed from `builtins.match` to `lib.hasSuffix` to avoid some of the known issues with the Nix's regex implementation and also because `nix fmt` was removing the escaping (making the regex matching more than it should).

teto:
I am not a fan of the news system, which I would rather ditch completely. This looks better than the current one though so +1 from me. How does it deal with old news ? should they all be migrated ? Has anyone tested this (I wont)

thiagokokada:
> How does it deal with old news ?

It keep them as-is. We could migrate all of them but not sure if it is worth it.

> Has anyone tested this (I wont)

I replaced home-manager as an input for my `flake.nix` and run `nix build .#homeConfigurations.hostname.config.news.json.output` and it seemed to be correct to me. I am not sure if there is a better way to test this to be honest.

khaneliman:
> > How does it deal with old news ? should they all be migrated ?
> 
> It keep them as-is. We could migrate all of them but not sure if it is worth it.
> 

Yeah, I don't know if it's worth migrating over... other than possibility of people still using the old file anyways. 

> > Has anyone tested this (I wont)
> 
> I replaced home-manager as an input for my `flake.nix` and run `nix build .#homeConfigurations.hostname.config.news.json.output` and it seemed to be correct to me. I am not sure if there is a better way to test this to be honest.

Script works fine for generating the news entry, `home-manager news` command has never worked for me, though.

thiagokokada:
Can we get this merged, or is there anything else left?

CC @khaneliman.

khaneliman:
Sorry, can you rebase again... forgot to merge and merged a new module with a news announcement.

thiagokokada:
> Sorry, can you rebase again... forgot to merge and merged a new module with a news announcement.

Done.

PR #6761, opened 2025-04-05
progs: firefox: *.userChrome may be path or drv 

spikespaz:
### Description

This allows `programs.firefox.profiles.*.userChrome` to be set to a:
derivation, path/path-like string to directory or file, or multiline
text to be used as content verbatim.

This allows setting, for example(s):
```nix
programs.firefox.profiles."jacob.default".userChrome = pkgs.wavefox;
programs.firefox.profiles."jacob.default".userChrome =
    "${pkgs.wavefox}/userChrome.css";
```

### Checklist

- [X] (Painstakingly) Change is backwards compatible.

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/firefox/mkFirefoxModule.nix
+++ b/modules/programs/firefox/mkFirefoxModule.nix
@@ -397,12 +397,19 @@ in
               };
 
               userChrome = mkOption {
-                type = types.oneOf [
-                  types.lines
-                  types.path
-                ];
-                default = "";
-                description = "Custom ${appName} user chrome CSS.";
+                type = types.nullOr (
+                  types.oneOf [
+                    types.lines
+                    types.path
+                  ]
+                );
+                default = null;
+                description = ''
+                  Custom ${appName} user chrome CSS.
+
+                  This can be a path to a file or directory in the Nix store,
+                  or a derivation, or a verbatim multi-line string.
+                '';
                 example = ''
                   /* Hide tab bar in FF Quantum */
                   @-moz-document url(chrome://browser/content/browser.xul), url(chrome://browser/content/browser.xhtml) {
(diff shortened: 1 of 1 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+  "nix.serverSettings": {
+    "formatter": {
+      "command": "nixfmt-rfc-style"
+    }
+  }
+}
```
khaneliman, line 7: should be a separate commit

spikespaz:
Removed accidental VS Code workspace change. Should be good now.

khaneliman (approved):
Thanks, this actually allows me to remove some hacky stuff I had to do with my configuration for installing a theme.

antonmosich:
This broke my config, in which I set
```
programs.firefox.profiles.<name>.userChrome = ''
<css>
'';
```
The error says that the string doesn't represent an absolute path.

khaneliman:
> This broke my config, in which I set
> 
> ```
> programs.firefox.profiles.<name>.userChrome = ''
> <css>
> '';
> ```
> 
> The error says that the string doesn't represent an absolute path.
> 
> It seems to fail while executing [`lib.pathIsDirectory`](https://github.com/nix-community/home-manager/pull/6761/files#diff-41488fcc8e42ec317277f7961b22cc4f023f5452d525c02bb41dbaa7ac895910R872).

Oops, thought we had basic tests already.. I'll add some tests and fix it.

spikespaz:
Setting a string for me worked, `pathIsDirectory` should return `null`. It's possible that I misremember and cherry-picked the wrong commit.

jvanbruegge:
The function seems to get tripped up by a URL, it complains about the first line of this:
```
@namespace url("http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul"); /* set default namespace to XUL */

#webrtcIndicator {
  display: none !important;
}
```

khaneliman:
I have a fix, just finishing up unit tests.

Issue #6484, opened 2025-02-17
bug: nushell module w/ `xdg.enable = true;` on macOS has incorrect env and nu variables


rykugur:
### Are you following the right branch?

- [x] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Issue description

I got a new MacBook the other day, but I also experienced this same issue on my work MacBook. I do not experience this issue on nixOS with the same HM module below.

With `xdg.enable = false;` (or commented out), my nushell config files get written to `~/Library/Application Support/nushell`.

With `xdg.enable = true;`', my nushell config files get written to `~/.config/nushell`.

Based on my reading of the HM module nix file, this is working as expected. In the former case above, everything (my config) works just fine. But I want to enable XDG for other apps that utilize it.

In the latter case, while the files show up in the right location:

<img width="542" alt="Image" src="https://github.com/user-attachments/assets/211579ff-525e-432e-bbcb-745989ae1577" />

the only env variable for XDG that seems to get set is `XDG_DATA_DIRS`:

<img width="849" alt="Image" src="https://github.com/user-attachments/assets/c80dbfe9-8aa8-496b-af80-98b9454aab3e" />

in addition, my `$nu.data-dir` and `$nu.config-dir` variables are incorrect:

<img width="620" alt="Image" src="https://github.com/user-attachments/assets/8398837f-8190-4529-96f6-cb2e214d2d60" />

## Config files

Here's my nushell HM module:

```
{ config, lib, pkgs, ... }:
let
  cfg = config.rhx.nushell;
  nu-scripts = pkgs.fetchFromGitHub {
    owner = "nushell";
    repo = "nu_scripts";
    rev = "e380c8a355b4340c26dc51c6be7bed78f87b0c71";
    sha256 = "sha256-b2AeWiHRz1LbiGR1gOJHBV3H56QP7h8oSTzg+X4Shk8=";
  };
  # nupm = pkgs.fetchFromGitHub {
  #   owner = "nushell";
  #   repo = "nupm";
  #   rev = "7e3e5779ff86a1b8dadcf7a90eee2e3dcfe449df";
  #   sha256 = "sha256-BNFBQ9kK2/P7mjdBqMj/8cbBPVogK0n1qcx6dx9mer8=";
  # };
in {
  options.rhx.nushell = {
    enable = lib.mkEnableOption "Enable nushell home-manager module.";
  };

[...]

PR #6763, opened 2025-04-06
nushell: expose config dir location as option

JoaquinTrinanes:
### Description

This PR exposes the nushell configuration directory through the `programs.nushell.configDir` option.

#### Motivation

Nushell has different defaults for the configuration dir depending on the OS: `Library/Application Support/nushell` on MacOS and `.config/nushell` on Linux. However, setting `XDG_CONFIG_HOME` will override the default config unconditionally.

This means that there exists the possibility of the configuration not working in systems where users have `XDG_CONFIG_HOME` managed outside of home-manager (see https://github.com/nix-community/home-manager/issues/6484 for a related example).

This also allows just overriding the default location in general, for whatever the user might want to use it or for setting up integrations (adding files to `$nu.user-autoload-dirs` for example)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/nushell.nix
+++ b/modules/programs/nushell.nix
@@ -9,12 +9,6 @@ let
   inherit (lib.hm.nushell) isNushellInline toNushell;
   cfg = config.programs.nushell;
 
-  configDir =
-    if pkgs.stdenv.isDarwin && !config.xdg.enable then
-      "Library/Application Support/nushell"
-    else
-      "${config.xdg.configHome}/nushell";
-
   linesOrSource =
     name:
     types.submodule (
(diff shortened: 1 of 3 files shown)


JoaquinTrinanes:
I submitted ahead of time without pinging the rest of maintainers 🥲 @Philipp-M @aidalgol

Philipp-M (approved):
Seems reasonable to me. Maybe @rycee has input as well, as this (making the config dir configurable) is *maybe* relevant compared to other modules as well.

jimmyff:
any update on getting this merged in? this would solve my issues

modules/programs/nushell.nix
```diff
@@ -55,6 +49,24 @@ in
 
     package = lib.mkPackageOption pkgs "nushell" { nullable = true; };
 
+    configDir = lib.mkOption {
+      type = types.path;
```
khaneliman:
```suggestion
      type = types.either types.str types.path;
```

Needs to support a basic string, too. 

>error: A definition for option `programs.nushell.configDir' is not of type `absolute path'. Definition values:
       - In `/nix/store/crj3fd4mm48jbl2bb4ml37apsib68ij4-source/tests/modules/programs/nushell/config-dir.nix': "Library/Application Support/nushell"

JoaquinTrinanes: Rebased to `master` and applied that change in my latest commit. Thanks for reporting the error, I don't have a darwin machine :)

This made me wonder if allowing any string is better than setting the default as an absolute path tho:

```nix
      default =
        if pkgs.stdenv.isDarwin && !config.xdg.enable then
          "${config.home.homeDirectory}/Library/Application Support/nushell"
        else
          "${config.xdg.configHome}/nushell";
```

khaneliman: I'd prefer the flexibility since we're using `home.file`

PR #6772, opened 2025-04-07
format: migrate to nixfmt-rfc-style; treewide: nixfmt-rfc-style reformat

khaneliman:
### Description

Would like to standardize on the modern formatting and update our repo to match upstream behavior better. This migrates from `nixfmt` to `nixfmt-tree/nixfmt-rfc-style` with `treefmt-nix`. We can also supply formatters we'd like to use for markdown, and other languages to reduce issues from users having different formatters, as well. 

Updated CI to use the new formatting check. Will do a format before we actually merge. 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


khaneliman:
> @khaneliman is there a way to not include this as a Nix flake input? Your dev tools are now propagated to my `flake.lock` using more bandwidth. Or should I find a different, non-Flake way to fetch home-manager?

Yeah, I think I'm going to switch it to just use the `nixfmt-tree` for now, until I move it all into the dev flake as mentioned in another PR.

PR #6775, opened 2025-04-07
jujutsu: evaluate the settings after merging with other options

teto:
else the file might ignore some options

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/jujutsu.nix
+++ b/modules/programs/jujutsu.nix
@@ -66,10 +66,9 @@ in
   config = mkIf cfg.enable {
     home.packages = mkIf (cfg.package != null) [ cfg.package ];
 
-    home.file."${configDir}/jj/config.toml" = mkIf (cfg.settings != { }) {
-      source = tomlFormat.generate "jujutsu-config" (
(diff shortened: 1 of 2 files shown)


ambroisie (commented):
IMO it would be simpler to have a snippet like this in the module:

```nix
  programs.jujutsu.settings = lib.mkMerge [
    (lib.mkIf cfg.ediff {
      merge-tools.ediff = {
        program = lib.getExe emacsDiffScript;
        merge-args = [ "$left" "$right" "$base" "$output" ];
      };
    })
  ];
```

Using `mkMerge` to easily add more `lib.mkIf` blocks if new "meta-settings" pop-up in the future.

This would also loudly signal to the user if they have a clashing `merge-tools.ediff` option set somewhere in their configuration (needing to `mkForce` or delete it).

khaneliman (approved):

ambroisie (approved):

teto:
@ambroisie thanks for the suggestion !

PR #6776, opened 2025-04-08
astroid: Fix use of `fromJSON`

Lyndeno:
### Description

Module was rewritten to use it from `lib`, when it is actually in `builtins`
### Checklist

  - [x] **Code tested with actual config**

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@khaneliman

### DIFF
--- a/modules/programs/astroid.nix
+++ b/modules/programs/astroid.nix
@@ -36,7 +36,7 @@ let
 
   # See https://github.com/astroidmail/astroid/wiki/Configuration-Reference
   finalConfig = let
-    template = builtins.fromJSON (lib.readFile ./astroid-config-template.json);
+    template = lib.importJSON ./astroid-config-template.json;
     astroidConfig = lib.foldl' lib.recursiveUpdate template [
       {
         astroid.notmuch_config =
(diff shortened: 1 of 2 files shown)


modules/programs/astroid.nix
```diff
@@ -36,7 +36,7 @@ let
 
   # See https://github.com/astroidmail/astroid/wiki/Configuration-Reference
   finalConfig = let
-    template = lib.fromJSON (lib.readFile ./astroid-config-template.json);
+    template = builtins.fromJSON (lib.readFile ./astroid-config-template.json);
```
ambroisie, line 39:
```suggestion
    template = lib.importJSON ./astroid-config-template.json;
```

ambroisie, line 39: Previous comment was deleted as I wasn't sure which `importJSON` is at library root, but it is indeed [`lib.trivial.importJSON`](https://github.com/NixOS/nixpkgs/blob/32a40c2a6bc36c3d6ebd914da27c39e0dba66c67/lib/default.nix#L138) (and not [`lib.modules.importJSON`](https://github.com/NixOS/nixpkgs/blob/32a40c2a6bc36c3d6ebd914da27c39e0dba66c67/lib/modules.nix#L1917)).

```suggestion
    template = lib.importJSON ./astroid-config-template.json;
```

PR #6778, opened 2025-04-08
Aerospace: Flatten on-window-detected.*.if values to fix incorrect TOML

axman6:
### Description

The current aerospace.nix module produces incorrect TOML when for `userSettings.on-window-detected`. Using the following 

```nix
{
  on-window-detected = [
    {
      "if" = { app-id = "com.apple.MobileSMS"; };
      run = [ "move-node-to-workspace 10" ];
    }
    {
      "if" = { app-id = "ru.keepcoder.Telegram"; };
      run = [ "move-node-to-workspace 10" ];
    }
    {
      "if" = { app-id = "org.whispersystems.signal-desktop"; };
      run = [ "move-node-to-workspace 10" ];
    }
  ];
}
```

produces 

```toml
[[on-window-detected]]
run = ["move-node-to-workspace 10"]

[on-window-detected.if]
app-id = "com.apple.MobileSMS"

[[on-window-detected]]
run = ["move-node-to-workspace 10"]

[on-window-detected.if]
app-id = "ru.keepcoder.Telegram"

[[on-window-detected]]
run = ["move-node-to-workspace 10"]

[on-window-detected.if]
app-id = "org.whispersystems.signal-desktop"
```

when it should produce

```toml
[[on-window-detected]]
if.app-id = "com.apple.MobileSMS"
run = ["move-node-to-workspace 10"]

[[on-window-detected]]
if.app-id = "ru.keepcoder.Telegram"
run = ["move-node-to-workspace 10"]

[[on-window-detected]]
if.app-id = "org.whispersystems.signal-desktop"
run = ["move-node-to-workspace 10"]
```

These changes should fix this. Since the current implementation produces the wrong TOML, I believe this should be considered backwards compatible.

I'm not sure how to test this fully with Aerospace itself, the output differs slightly from the example above, it actually produces the following, with quoted keys:

```toml
[[on-window-detected]]
"if.app-id" = "com.apple.MobileSMS"
run = ["move-node-to-workspace 10"]
...
```

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@damidoug

### DIFF
--- a/tests/modules/programs/aerospace/aerospace.nix
+++ b/tests/modules/programs/aerospace/aerospace.nix
@@ -14,6 +14,20 @@
         alt-k = "focus up";
         alt-l = "focus right";
       };
+      on-window-detected = [
+        {
+          "if" = { app-id = "com.apple.MobileSMS"; };
+          run = [ "move-node-to-workspace 10" ];
+        }
+        {
+          "if" = { app-id = "ru.keepcoder.Telegram"; };
+          run = [ "move-node-to-workspace 10" ];
+        }
+        {
+          "if" = { app-id = "org.whispersystems.signal-desktop"; };
+          run = [ "move-node-to-workspace 10" ];
+        }
+      ];
     };
   };
 
--- a/tests/modules/programs/aerospace/settings-expected.toml
+++ b/tests/modules/programs/aerospace/settings-expected.toml
@@ -8,7 +8,6 @@ enable-normalization-opposite-orientation-for-nested-containers = true
 exec-on-workspace-change = []
 on-focus-changed = []
 on-focused-monitor-changed = ["move-mouse monitor-lazy-center"]
-on-window-detected = []
 start-at-login = false
 
 [gaps.outer]
@@ -25,3 +24,15 @@ alt-h = "focus left"
 alt-j = "focus down"
 alt-k = "focus up"
 alt-l = "focus right"
+
+[[on-window-detected]]
+"if.app-id" = "com.apple.MobileSMS"
+run = ["move-node-to-workspace 10"]
+
+[[on-window-detected]]
+"if.app-id" = "ru.keepcoder.Telegram"
+run = ["move-node-to-workspace 10"]
+
+[[on-window-detected]]
+"if.app-id" = "org.whispersystems.signal-desktop"
+run = ["move-node-to-workspace 10"]
--- a/modules/programs/aerospace.nix
+++ b/modules/programs/aerospace.nix
@@ -23,6 +23,22 @@ let
       else
         [ ]) (lib.attrNames set));
   filterNulls = filterListAndAttrsRecursive (v: v != null);
+
+  # Turns
+  #   {if = {foo = "xxx"; bar = "yyy"}}
+  # into
+  #   {"if.foo" = "xxx"; "if.bar" = "yyy"}
+  # so that the correct TOML is generated for the
+  # on-window-detected table.
+  flattenConditions = attrs:
+    let conditions = attrs."if" or { };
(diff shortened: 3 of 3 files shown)


khaneliman:
Can rebase to fix CI

khaneliman:
I don't use aerospace, but I just was reading the docs https://nikitabobko.github.io/AeroSpace/guide#on-window-detected-callback and it sounds like order of rules matters, too. Might want to double check implementation if it will make sure the rules are ordered how you want.

axman6:
The order should be the same as the list from the config. The keys in each `[[on-window-detected]]` section shouldn't matter, it's akin to a structure like thins in JSON

```json
{
"on-window-detected":  [
  { "if.app-I’d": "com.apple.MobileSMS",
     "run": ["move-node-to-workspace 10"]
  },
  { "if.app-I’d": "ru.keepcoder.Telegram",
     "run": ["move-node-to-workspace 10"]
  },
  { "if.app-I’d":"org.whispersystems.signal-desktop",
     "run": ["move-node-to-workspace 10"]
  }
  ]
}
```
at least as far as I understand TOML anyway. If the ordering of the list isn't preserved, I'd say that's a bug in the Nix TOML formatter.

axman6:
@khaneliman  I just tried this out with aerospace proper, and it reports an error, not sure why it's unhappy (though... it's slightly less broken than it was before)

tests/modules/programs/aerospace/settings-expected.toml
```diff
@@ -25,3 +24,15 @@ alt-h = "focus left"
 alt-j = "focus down"
 alt-k = "focus up"
 alt-l = "focus right"
+
+[[on-window-detected]]
+"if.app-id" = "com.apple.MobileSMS"
```
khaneliman, line 29: Oh, might be these are quote enclosed instead of `if.app-id`

tests/modules/programs/aerospace/aerospace.nix
```diff
@@ -14,6 +14,20 @@
         alt-k = "focus up";
         alt-l = "focus right";
       };
+      on-window-detected = [
+        {
+          "if" = { app-id = "com.apple.MobileSMS"; };
```
khaneliman, line 19: Might need this to be `"if.app-id" = "com.apple.MobileSMS"

khaneliman:
Sorry, just saw your edit. Yeah, might need to play around with it a little to get it to generate that properly.

axman6:
I need to head to bed, I'll see if I can make it work tomorrow. Any tips for speeding up testing generating the aerospace.toml file would be greatly appreciated - I ended up changing my nix-darwin flake and using my branch, but that feels like a pretty heavyweight way to test just this (I've been doing a full darwin-rebuild, which seems excessive).

khaneliman:
> I need to head to bed, I'll see if I can make it work tomorrow. Any tips for speeding up testing generating the aerospace.toml file would be greatly appreciated - I ended up changing my nix-darwin flake and using my branch, but that feels like a pretty heavyweight way to test just this (I've been doing a full darwin-rebuild, which seems excessive).

Should be able to change the test output to the proper expected output and run the test itself instead of generating your whole config.

axman6:
> > I need to head to bed, I'll see if I can make it work tomorrow. Any tips for speeding up testing generating the aerospace.toml file would be greatly appreciated - I ended up changing my nix-darwin flake and using my branch, but that feels like a pretty heavyweight way to test just this (I've been doing a full darwin-rebuild, which seems excessive).
> 
> Should be able to change the test output to the proper expected output and run the test itself instead of generating your whole config.

Oh yeah, obviously =) I really do need to get to sleep I guess!

PR #6783, opened 2025-04-08
modules/programs/ssh: support identityAgent

colemickens:
### Description

Allow configuring `identityAgent` in the ssh config.

I manually set up the identity agent/paths to better handle/understand how forwarding/agents work in my various usages (local, remote with gpg forwarded, etc).

This is a pretty straight-forward module-option -> config key change.

### Checklist

- [ ] **(hopefully not needed)** Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/ssh.nix
+++ b/modules/programs/ssh.nix
@@ -167,6 +167,22 @@ let
           '';
         };
 
+        identityAgent = mkOption {
+          type = with types; either (listOf str) (nullOr str);
+          default = [ ];
+          apply =
+            p:
+            if p == null then
+              [ ]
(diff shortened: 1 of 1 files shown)


khaneliman (approved):
Not a fan of the apply function, it could be simplified a lot. But, I see it's matching the existing styling in the module. We can clean that up in a module refactor.

PR #6784, opened 2025-04-08
firefox: add release option

### LINKED ISSUE #6758: bug: `programs.librewolf.languagePacks` doesn't work when `librewolf` version doesn't match `firefox` version

### Issue description

The `programs.librewolf.languagePacks` doesn't work correctly when the version of Librewolf doesn't exactly match the version of firefox it's based on.

For example, the current version of librewolf is: (note the `{...}-1`)

```nix
nix-repl> outputs.homeConfigurations.desktop.pkgs.librewolf.version
"136.0.4-1"
```

Under the hood, the `program.librewolf` module uses `mkFirefoxModule` which uses the version of the package to fetch the language pack extensions from `releases.mozilla.org` assuming that it lines up with a version of firefox:

https://github.com/nix-community/home-manager/blob/b5e29565131802cc8adee7dccede794226da8614/modules/programs/firefox/mkFirefoxModule.nix#L820-L828

When I look at the generated `policies.json` on my system, I see this: (I selected `en-CA`, `en-US` and `fr` as `languagePacks`)

```json5
{
  "policies": {
    "DisableAppUpdate": true,
    "ExtensionSettings": {
      "langpack-en-CA@firefox.mozilla.org": {
        "install_url": "https://releases.mozilla.org/pub/firefox/releases/136.0.4-1/linux-x86_64/xpi/en-CA.xpi",
        "installation_mode": "normal_installed"
      },
      "langpack-en-US@firefox.mozilla.org": {
        "install_url": "https://releases.mozilla.org/pub/firefox/releases/136.0.4-1/linux-x86_64/xpi/en-US.xpi",
        "installation_mode": "normal_installed"
      },
      "langpack-fr@firefox.mozilla.org": {
        "install_url": "https://releases.mozilla.org/pub/firefox/releases/136.0.4-1/linux-x86_64/xpi/fr.xpi",
        "installation_mode": "normal_installed"
      }
    },
    // ...
  }
}
```

If you try to open the `install_url`s listed here you'll get a `404`. If you try to get the same URLs but you remove the `-1` suffix from the version you get the right language pack.

### Maintainer CC

Librewolf maintainers: @chayleaf @onny
mkFirefoxModule author: @brckd 

### System information

```markdown
 - system: `"x86_64-linux"`
 - host os: `Linux 6.13.8-arch1-1, Arch Linux, noversion, rolling`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.27.1`
 - channels(root): `""`
 - channels(dotboris): `"nixpkgs"`
 - nixpkgs: `/nix/store/ypwkfln5h5hv5bb5a16pipb6w6bb8nb7-nixpkgs/nixpkgs`
```

brckd:
Adds the `release` option to declare what upstream release version is used to fetch rom `releases.mozilla.org`. Used to fetch language packs. See https://github.com/nix-community/home-manager/issues/6758#issuecomment-2781465236. 

Closes #6758

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@getchoo @brckd @HPsaucii

### DIFF
--- a/modules/programs/firefox/mkFirefoxModule.nix
+++ b/modules/programs/firefox/mkFirefoxModule.nix
@@ -24,6 +24,7 @@ let
     mkIf
     mkMerge
     mkOption
+    mkOptionDefault
     optionalString
     optional
     setAttrByPath
@@ -235,6 +236,12 @@ in
       '';
     };
 
+    release = mkOption {
+      internal = true;
+      type = types.str;
+      description = "Upstream release version used to fetch from `releases.mozilla.org`.";
+    };
+
     languagePacks = mkOption {
       type = types.listOf types.str;
       default = [ ];
@@ -242,7 +249,9 @@ in
         The language packs to install. Available language codes can be found
         on the releases page:
         `https://releases.mozilla.org/pub/firefox/releases/''${version}/linux-x86_64/xpi/`,
-        replacing `''${version}` with the version of ${appName} you have.
+        replacing `''${version}` with the version of ${appName} you have. If
+        the version string of your Firefox derivative diverts from the upstream
+        version, try setting the `release` option.
       '';
       example = [
         "en-GB"
@@ -952,6 +961,7 @@ in
     }
     // setAttrByPath modulePath {
       finalPackage = wrapPackage cfg.package;
+      release = mkOptionDefault (builtins.head (lib.splitString "-" cfg.package.version));
 
       policies = {
         ExtensionSettings = lib.mkIf (cfg.languagePacks != [ ]) (
@@ -960,7 +970,7 @@ in
               lang:
               lib.nameValuePair "langpack-${lang}@firefox.mozilla.org" {
                 installation_mode = "normal_installed";
-                install_url = "https://releases.mozilla.org/pub/firefox/releases/${cfg.package.version}/linux-x86_64/xpi/${lang}.xpi";
+                install_url = "https://releases.mozilla.org/pub/firefox/releases/${cfg.release}/linux-x86_64/xpi/${lang}.xpi";
               }
             ) cfg.languagePacks
           )


brckd:
Drafting this, because the formatter mysteriously stopped working and I haven't tested this yet. I'm also quite busy this time around.

brckd:
Good to know! It's just that `./format` raises an obscure error.

khaneliman:
> Good to know! It's just that `./format` raises an obscure error.

Are you using jujutsu instead of git?

brckd:
No, and it was something along the lines of *couldn't write to ~/.cache/formatter*. Seems like a non reproducible problem, but I'll check again when I'm back from vacation. I'll open a separate issue if it persists.

bricked-dev:
I made it so the `release` option defaults to the package's version string minus anything following a dash. So `140.0.2-1` becomes `140.0.2` by default.

bricked-dev:
Seems to work for me on Librewolf `140.0.2-1`!
Screenshot From 2025-07-05 23-26-00

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -960,7 +970,7 @@ in
               lang:
               lib.nameValuePair "langpack-${lang}@firefox.mozilla.org" {
                 installation_mode = "normal_installed";
-                install_url = "https://releases.mozilla.org/pub/firefox/releases/${cfg.package.version}/linux-x86_64/xpi/${lang}.xpi";
+                install_url = "https://releases.mozilla.org/pub/firefox/releases/${cfg.release}/linux-x86_64/xpi/${lang}.xpi";
```
khaneliman, line 973: Hmm interesting, we have the `linux-x86_64` hardcoded... are language packs actually platform dependent or will they work just fine on others?

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -952,6 +961,7 @@ in
     }
     // setAttrByPath modulePath {
       finalPackage = wrapPackage cfg.package;
+      release = mkOptionDefault (builtins.head (lib.splitString "-" cfg.package.version));
```
khaneliman, line 964: Doesn't look like this will work for [floorp](https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/networking/browsers/floorp/default.nix#L20), at least. They have weird versioning. 

EDIT: Might need to add a condition if there are 3 sections its an ESR release? Using their [128.12.0](https://releases.mozilla.org/pub/firefox/releases/128.12.0/linux-x86_64/) didn't work, but [128.12.0esr](https://releases.mozilla.org/pub/firefox/releases/128.12.0esr/linux-x86_64/) returned a language pack.

modules/programs/firefox/mkFirefoxModule.nix
bricked-dev: According to https://support.mozilla.org/en-US/kb/deploying-firefox-language-packs:

> Even though the XPI files are duplicated in the different platforms, there is no difference between them. You can download the XPI from any platform directory.

So I think no action needs to be taken until this changes.

modules/programs/firefox/mkFirefoxModule.nix
bricked-dev: According to https://github.com/Floorp-Projects/Floorp/discussions/620#discussioncomment-7799590, Floorp is based on Firefox ESR. I don't think that the default LibreWolf or Firefox packages are ESRs and they use the same versioning scheme. So I'd add an `isEsr` option instead.

bricked-dev: Ok I added a `mkRelease` parameter instead which maps a version to a release. This can be passed to `mkFirefoxModule` for each Firefox derivative. I checked the latest versions of LibreWolf and Floorp to confirm that the `mkRelease` methods should work.

khaneliman: That looks nice. Each derivative owns their versioning instead of the helper function.

bricked-dev: Tested this with

```nix
{
  programs.librewolf.enable = true;
  programs.floorp.enable = true;
}
```

```console
nix-repl> legacyPackages.x86_64-linux.homeConfigurations.bricked.config.programs.librewolf.package.version
"140.0.2-1"

nix-repl> legacyPackages.x86_64-linux.homeConfigurations.bricked.config.programs.librewolf.release
"140.0.2"

nix-repl> legacyPackages.x86_64-linux.homeConfigurations.bricked.config.programs.floorp.package.version
"11.28.0"

nix-repl> legacyPackages.x86_64-linux.homeConfigurations.bricked.config.programs.floorp.release
"11.28.0esr"
```
Oops, seems like I used the wrong Floorp version!

bricked-dev: I undid that commit because it didn't fix the issue with Floorp.

bricked-dev:
Since the Floorp package only exposes the Floorp version which cannot directly be mapped to a Firefox version, the Floorp release will have to be set manually. We can tackle after the fact, when https://github.com/NixOS/nixpkgs/pull/397228 is merged.

modules/programs/firefox/mkFirefoxModule.nix
khaneliman: Oh, yeah.. didn't notice it pulled the floorp version instead of the firefox version set to `version` in the `buildMozillaMach`

bricked-dev:
Depending on how long the Nixpkgs PR takes to be merged and backported, we might want to merge this PR first. This way, LibreWolf users would already have a working fix and Floorp users can set the release manually :thinking:

khaneliman:
> Depending on how long the Nixpkgs PR takes to be merged and backported, we might want to merge this PR first. This way, LibreWolf users would already have a working fix and Floorp users can set the release manually 🤔

Yeah, we can at least fix `librewolf` users, first.

PR #6789, opened 2025-04-09
hyprland: load plugins using exec-once

amadejkastelic:
### Description

Change to match the behavior of hyprland flake - https://github.com/hyprwm/Hyprland/pull/9836

Should eliminate crashes / plugin version mismatches when updating hyprland.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@fufexan

### DIFF
--- a/modules/services/window-managers/hyprland.nix
+++ b/modules/services/window-managers/hyprland.nix
@@ -287,12 +287,13 @@ in
           plugins:
           lib.hm.generators.toHyprconf {
             attrs = {
-              plugin =
+              "exec-once" =
                 let
                   mkEntry =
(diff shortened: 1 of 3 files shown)


fufexan (approved):

amadejkastelic:
@horriblename They should not be unloaded. Can you provide some more info?

horriblename:
sorry, false alarm. I was updating from an old commit (before this change), where the `plugin` keyword is still used to load plugins. So after updating, the old `plugin` keyword was removed, hence plugin was unloaded, and `exec-once` doesn't trigger in that situation

so it's a one-time problem that won't happen again

after restarting hyprland everything is working

PR #6791, opened 2025-04-10
way-displays: init module

jolars:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/services/way-displays.nix
```diff
@@ -0,0 +1,115 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+with lib;
```
khaneliman, line 8:
```suggestion
```

modules/lib/maintainers.nix
```diff
@@ -181,6 +181,14 @@
...
     name = "Emil Karlson";
   };
+  jolars = {
+    email = "jolars@posteo.com";
+    matrix = "@jola:mozilla.org";
+    github = "jolars";
+    githubId = 13087841;
+    name = "Johan Larsson";
+    keys = [ { fingerprint = "F0D6 BDE7 C7D1 6B3F 7883  73E7 2A41 C0FE DD6F F540"; } ];
+  };
```
khaneliman, line 191:
```suggestion
```

No need when you're in the nixpkgs list.

modules/services/way-displays.nix
```diff
@@ -0,0 +1,115 @@
...
+in
+{
+  meta.maintainers = [ maintainers.jolars ];
+
+  options.services.way-displays = {
+    enable = mkEnableOption "" // {
+      description = ''
+        Whether to enable way-displays.
+      '';
+    };
```
khaneliman, line 23:
```suggestion
    enable = mkEnableOption "way-displays";
```

modules/services/way-displays.nix
```diff
@@ -0,0 +1,115 @@
...
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.way-displays;
+      defaultText = literalExpression "pkgs.way-displays";
+      description = ''
+        way-displays derivation to use.
+      '';
+    };
```
khaneliman, line 32:
```suggestion
    package = lib.mkPackageOption pkgs "way-displays" {};
```

khaneliman (changes requested):

modules/services/way-displays.nix
```diff
@@ -0,0 +1,112 @@
...
+    assertions = [
+      (lib.hm.assertions.assertPlatform "services.way-displays" pkgs lib.platforms.linux)
+    ];
+
+    xdg.configFile."way-displays/cfg.yaml".source =
+      yaml.generate "way-displays-config.yaml"
+        (mergeSets [
+          {
+            CALLBACK_CMD = "${pkgs.libnotify}/bin/notify-send \"way-displays \${CALLBACK_LEVEL}\" \"\${CALLBACK_MSG}\"";
+          }
```
khaneliman, line 87: Looks like you're always including this regardless of the user configuration, this feels like it should be optional since it includes another dependency, as well.

tests/modules/services/way-displays/configuration.nix
```diff
@@ -0,0 +1,43 @@
+{ config, ... }:
+{
+  config = {
+    services.way-displays = {
+      enable = true;
+      package = config.lib.test.mkStubPackage { };
+      settings = {
+        CALLBACK_CMD = "notify-send \"way-displays \${CALLBACK_LEVEL}\" \"\${CALLBACK_MSG}\"";
```
khaneliman, line 8: Here you show it being passed as a settings config, which makes more sense since it would be opt in

jolars: The reason I included this is that this is the default setting, but it is not guaranteed to work out-of-the-box if notify-send isn't in the path. I guess we could patch the package in nixpkgs too, but this seemed like an easier solution (albeit with the alternative maintenance load of having to keep the default up to date).

modules/services/way-displays.nix
khaneliman: > The reason I included this is that this is the default setting, but it is not guaranteed to work out-of-the-box if notify-send isn't in the path. I guess we could patch the package in nixpkgs too, but this seemed like an easier solution (albeit with the alternative maintenance load of having to keep the default up to date).

I see what you mean with this being a default upstream https://github.com/alex-courtis/way-displays/blob/master/examples/cfg.yaml#L78 If this will just be used to `fix` the derivation not wrapping `libnotify` upstream, then it makes more sense.

khaneliman: Might make sense then, to make this a `mkDefault` though so user's wanting to set their own CALLBACK_CMD don't have to mkForce it.

PR #6792, opened 2025-04-10
formatter: use a different treefmt root

teto:
Problem: i use jujutsu as a VCS these days https://jj-vcs.github.io/jj/ and when using jujutsu workspaces, they dont contain the `.git` file/folder that git worktrees have. `nix fmt` thus fails `with Error: failed to load config: failed to find tree-root based on tree-root-file: could not find [.git/index] in /home/teto/hm2`

solution: use a different (unique) root file. I first tried with flake.nix for instnace but we have several in the tree while there is only one "release.json"

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -50,11 +50,11 @@
       {
         formatter = forAllSystems (
           system:
-          nixpkgs.legacyPackages.${system}.nixfmt-tree.override ({
+          nixpkgs.legacyPackages.${system}.nixfmt-tree.override {
             settings = {
               tree-root-file = "release.json";
             };
-          })
+          }
         );
 
         packages = forAllSystems (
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -48,7 +48,14 @@
...
       {
-        formatter = forAllSystems (system: nixpkgs.legacyPackages.${system}.nixfmt-tree);
+        formatter = forAllSystems (
+          system:
+          nixpkgs.legacyPackages.${system}.nixfmt-tree.override ({
+            settings = {
+              tree-root-file = "release.json";
+            };
+          })
+        );
```
khaneliman, line 58:
```suggestion
        formatter = forAllSystems (
          system:
          nixpkgs.legacyPackages.${system}.nixfmt-tree.override {
            settings = {
              tree-root-file = "release.json";
            };
          }
        );
```

Just minor nit for redundant parens

teto:
./format seems to work differently than `nix fmt` as it uses git ls-files. Would be nice to have both in sync, maybe with https://github.com/edolstra/flake-compat

PR #6795, opened 2025-04-10
xmonad: fix binary name lookup on armv7l when cross-compiled

sorki:
### Description

This patch along with https://github.com/NixOS/nixpkgs/pull/397767 allows pre-built xmonad running on `armv7l` target.

GHC converts `armv7l` and similar to just `arm` https://git.science.uu.nl/f100183/ghc/-/blob/8fa4bf9ab3f4ea4b208f4a43cc90857987e6d497/aclocal.m4#L1874 making `hostPlatform.system` point to `xmonad-arm7l-linux` instead of correct `xmonad-arm-linux`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/window-managers/xmonad.nix
+++ b/modules/services/window-managers/xmonad.nix
@@ -111,6 +111,12 @@ in
 
   config =
     let
+      # GHC sets OS_ARCH to arm in case of armv7l, need to coerce
+      executableSystemSuffix =
+        if pkgs.stdenv.hostPlatform.isArmv7 then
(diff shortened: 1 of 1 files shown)


khaneliman:
Does this need to wait for that? Or can this go in before it hits unstable?

sorki:
> Does this need to wait for that? Or can this go in before it hits unstable?

No need, the change is fairly unobtrusive. Without the nixpkgs patch you'll simply fail earlier at `xmonad --recompile` step.

PR #6802, opened 2025-04-11
nh: remove incorrect warning on gc conflicts

fxzzi:
### Description

From what I understand, the warning relating to the system GC conflicting with the nh user GC (added in #6470)  was incorrect. There shouldn't be any conflict between osConfig.nix.gc and programs.nh.clean as our home-manager service only cleans user profiles (which nix-collect-garbage does not do).

In this scenario they should both work at the same time as they are cleaning two different things.

### Checklist

~~- [ ] Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).~~

    ```
    {component}: {description}

    {long description}
    ```

  ~~- [ ] Added myself as module maintainer. See [example](https://github.com/nix-community/home-manager/blob/068ff76a10e95820f886ac46957edcff4e44621d/modules/programs/lesspipe.nix#L6).~~

#### Maintainer CC

@JohnRTitor

### DIFF
--- a/modules/programs/nh.nix
+++ b/modules/programs/nh.nix
@@ -1,15 +1,11 @@
 {
   config,
-  osConfig,
   lib,
   pkgs,
   ...
 }:
-
 let
-
   cfg = config.programs.nh;
-
 in
 {
   meta.maintainers = with lib.maintainers; [ johnrtitor ];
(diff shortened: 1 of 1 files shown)


JohnRTitor:
CC @viperML

fxzzi:
Also had a small conversation with NobbZ about the issue: https://canary.discord.com/channels/568306982717751326/1283818433049530380/1360219733114355866

Of course we should still wait for viperML's insight. I have been running this PR fine locally for a week or so now without issues.

viperML:
I don't use HM so `nh hone` is maintained by the community.

fxzzi:
considering that home-manager nix.gc doesn't conflict with system nix.gc it does seem like unintended behaviour for these to conflict.

JohnRTitor (approved):
ok, on to @khaneliman for merge.

PR #6807, opened 2025-04-12
labwc: Add module for Labwc

LesVu:
### Description

Add module for Labwc

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/services/window-managers/labwc/function.nix
```diff
@@ -0,0 +1,146 @@
...
+    # If the value is a list, emit the same tag name for each item
+    else if builtins.isList value then
+      # Reuse the same tag name for each list item
+      lib.concatStringsSep "\n" (map (v: generateRc name v) value)
+
+    # All other primitive values: wrap in start/end tag
+    else
+      "<${name}>${escape (builtins.toString value)}</${name}>";
+
+  generateXML = name: config: extraConfig: ''
```
khaneliman, line 122: Haven't looked too closely into the details, but is there a need for all this custom xml generation over utilizing the https://noogle.dev/f/builtins/toXML builtin toXML generator?

LesVu: Hi @khaneliman, from my research xml generated from builtins.toXML is hard to convert to xml use by labwc and need to use external package to parse whereas custom xml generation is easier to use as it is pure nix. Labwc also need to have certain attribute order that effect the behavior.

LesVu: There is pkgs.formats.xml that has generate function but it generate the xml attribute in alphabetical order which will not apply the correct behavior that the user want in labwc.

modules/services/window-managers/labwc/labwc.nix
```diff
@@ -0,0 +1,272 @@
...
+      cfg.environment ++ (lib.optionals (!cfg.xwayland.enable) [ "WLR_XWAYLAND=" ])
+    );
+
+    systemd.user.targets.labwc-session = lib.mkIf cfg.systemd.enable {
+      Unit = {
+        Description = "labwc compositor session";
+        Documentation = [ "man:systemd.special(7)" ];
+        BindsTo = [ "graphical-session.target" ];
+        Wants = [ "graphical-session-pre.target" ];
+        After = [ "graphical-session-pre.target" ];
```
khaneliman, line 268: Can you change these to follow the other wayland services that use `config.wayland.systemd.target`

rostby:
You can't use the icon option for menu.xml & the menu.xml file generates even if you don't specify the option

PR #6821, opened 2025-04-14
direnv: fix silent option after update to direnv v2.36.0

m4r1vs:
see https://github.com/direnv/direnv/issues/68 (bottom of thread)

### Description
Closes https://github.com/nix-community/home-manager/issues/6822

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee @khaneliman @shikanime

### DIFF
--- a/modules/programs/direnv.nix
+++ b/modules/programs/direnv.nix
@@ -118,73 +118,88 @@ in
     silent = mkEnableOption "silent mode, that is, disabling direnv logging";
   };
 
-  config = mkIf cfg.enable {
-    home.packages = [ cfg.package ];
+  config =
+    let
+      packageVersion = lib.getVersion cfg.package;
+      isVersion236orHigher = lib.versionAtLeast packageVersion "2.36.0";
+    in
+    mkIf cfg.enable {
+      home.packages = [ cfg.package ];
 
-    xdg.configFile."direnv/direnv.toml" = mkIf (cfg.config != { }) {
-      source = tomlFormat.generate "direnv-config" cfg.config;
-    };
-
-    xdg.configFile."direnv/lib/hm-nix-direnv.sh" = mkIf cfg.nix-direnv.enable {
-      source = "${cfg.nix-direnv.package}/share/nix-direnv/direnvrc";
-    };
-
-    xdg.configFile."direnv/direnvrc" = lib.mkIf (cfg.stdlib != "") { text = cfg.stdlib; };
-
-    xdg.configFile."direnv/lib/hm-mise.sh" = mkIf cfg.mise.enable {
-      text = ''
-        eval "$(${getExe cfg.mise.package} direnv activate)"
-      '';
-    };
-
-    programs.bash.initExtra = mkIf cfg.enableBashIntegration (
-      # Using mkAfter to make it more likely to appear after other
-      # manipulations of the prompt.
-      mkAfter ''
-        eval "$(${getExe cfg.package} hook bash)"
-      ''
-    );
-
-    programs.zsh.initContent = mkIf cfg.enableZshIntegration ''
-      eval "$(${getExe cfg.package} hook zsh)"
-    '';
-
-    programs.fish.interactiveShellInit = mkIf cfg.enableFishIntegration (
-      # Using mkAfter to make it more likely to appear after other
-      # manipulations of the prompt.
-      mkAfter ''
-        ${getExe cfg.package} hook fish | source
-      ''
-    );
-
-    # Using mkAfter to make it more likely to appear after other
-    # manipulations of the prompt.
-    programs.nushell.extraConfig = mkIf cfg.enableNushellIntegration (mkAfter ''
-      $env.config = ($env.config? | default {})
-      $env.config.hooks = ($env.config.hooks? | default {})
-      $env.config.hooks.pre_prompt = (
-          $env.config.hooks.pre_prompt?
-          | default []
-          | append {||
-              ${getExe cfg.package} export json
-              | from json --strict
-              | default {}
-              | items {|key, value|
-                  let value = do (
-                      $env.ENV_CONVERSIONS?
-                      | default {}
-                      | get -i $key
-                      | get -i from_string
-                      | default {|x| $x}
-                  ) $value
-                  return [ $key $value ]
+      xdg.configFile."direnv/direnv.toml" =
+        mkIf (cfg.config != { } || (cfg.silent && isVersion236orHigher))
+          {
+            source = tomlFormat.generate "direnv-config" (
+              cfg.config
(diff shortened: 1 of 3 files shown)


khaneliman:
Thanks, was waiting to hear back on proper solution since this didn't work for me on nixos https://github.com/direnv/direnv/issues/1418

modules/programs/direnv.nix
```diff
@@ -121,8 +121,12 @@ in
   config = mkIf cfg.enable {
     home.packages = [ cfg.package ];
 
-    xdg.configFile."direnv/direnv.toml" = mkIf (cfg.config != { }) {
-      source = tomlFormat.generate "direnv-config" cfg.config;
+    xdg.configFile."direnv/direnv.toml" = mkIf (cfg.config != {} || cfg.silent) {
+      source = tomlFormat.generate "direnv-config" cfg.config
+        // (lib.mkIf cfg.silent {
+          log_format = "-";
+          log_filter = "^$";
```
khaneliman, line 128: These are supposed to be under the global attribute.

m4r1vs: oh your're right, event though it works, I'll change it accordingly..

modules/programs/direnv.nix
```diff
@@ -121,8 +121,14 @@ in
   config = mkIf cfg.enable {
     home.packages = [ cfg.package ];
 
-    xdg.configFile."direnv/direnv.toml" = mkIf (cfg.config != { }) {
-      source = tomlFormat.generate "direnv-config" cfg.config;
+    xdg.configFile."direnv/direnv.toml" = mkIf (cfg.config != {} || cfg.silent) {
+      source = tomlFormat.generate "direnv-config" cfg.config
+        // (lib.mkIf cfg.silent {
```
khaneliman, line 126: Should probably be checking the package version, since this was a recent change.

m4r1vs: Also I don't know how to check the package version here, is there a "version" property on cfg.package? Could you provide an example please?

m4r1vs: > > I just downgraded to direnv 2.35 and the "new" way of silencing it does still work, would you still prefer to check the direnv version?
> 
> That doesn't really make sense... they only added these config values in this release [direnv/direnv@v2.35.0...v2.36.0#diff-a59528653ec4d512d46856d5f6f241559e6839b4ca63d60e427d298b189003e9R63-R64](https://github.com/direnv/direnv/compare/v2.35.0...v2.36.0#diff-a59528653ec4d512d46856d5f6f241559e6839b4ca63d60e427d298b189003e9R63-R64)

Yeah, my bad, I didn't properly clear my environment variables..

modules/programs/direnv.nix
```diff
@@ -184,7 +190,5 @@ in
           }
       )
     '');
-
-    home.sessionVariables = lib.mkIf cfg.silent { DIRENV_LOG_FORMAT = ""; };
```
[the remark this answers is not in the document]
m4r1vs: Cleared my environment variables and now the logs appear again. Seems like DIRENV_LOG_FORMAT still needs to be set to an empty string on v2.36.0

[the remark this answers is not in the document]
khaneliman: Yeah, looks like its simply just adding this to config, as well. Not a replacement (atm)

khaneliman:
Just needs the formatting, I believe

m4r1vs:
The format script somehow fails for me with the following error:
```
unpacking 'https://github.com/NixOS/nixpkgs/archive/42a1c966be226125b48c384171c44c651c236c22.tar.gz' into the Git cache...
Error: failed to open cache: could not resolve local path for the cache: could not create any of the following paths: [/.cache/treefmt/eval-cache]
```
I just ran nixfmt-tree without any args (I don't think the script does anything else?) but it changed the indentation of the entire config block so I pushed it as a seperate commit for now. If you approve, I'll fix it up into a single commit

m4r1vs:
> > The format script somehow fails for me with the following error:
> > ```
> > unpacking 'https://github.com/NixOS/nixpkgs/archive/42a1c966be226125b48c384171c44c651c236c22.tar.gz' into the Git cache...
> > Error: failed to open cache: could not resolve local path for the cache: could not create any of the following paths: [/.cache/treefmt/eval-cache]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I just ran nixfmt-tree without any args (I don't think the script does anything else?) but it changed the indentation of the entire config block so I pushed it as a seperate commit for now. If you approve, I'll fix it up into a single commit
> 
> That's weird, but yeah, it's basically just a wrapper for `nixmft-tree`. It looks like it's passing CI and looks similar to what I get locally. I think it just shifted a bunch of stuff with the change.
> 
> Change is still easy to see when you hide whitespace in diff image

Oh that's a nice one, thanks for the tip!

modules/programs/direnv.nix
```diff
@@ -118,73 +118,88 @@ in
     silent = mkEnableOption "silent mode, that is, disabling direnv logging";
   };
 
-  config = mkIf cfg.enable {
-    home.packages = [ cfg.package ];
+  config =
+    let
+      packageVersion = lib.getVersion cfg.package;
+      isVersion236rHigher = lib.versionAtLeast packageVersion "2.3.6";
```
khaneliman, line 124:
```suggestion
      isVersion236rHigher = lib.versionAtLeast packageVersion "2.36.0";
```

m4r1vs: oh good catch 😅

khaneliman:
Just noticed generated config is a little weird... seems to be including the mkIf properties

```toml
_type = "if"
condition = true

[content.global]
log_filter = "^$"
log_format = "-"
```

m4r1vs:
Ad

> Just noticed generated config is a little weird... seems to be including the mkIf properties
> 
> ```toml
> _type = "if"
> condition = true
> 
> [content.global]
> log_filter = "^$"
> log_format = "-"
> ```
> 
> Should add a silent test and play around with it.

Added a simple test. Actually parsing the toml in case a default config is provided by some future direnv version is not necessary or is it?

modules/programs/direnv.nix
khaneliman: Looks like this got reverted during the tweaking

khaneliman:
> > Should add a silent test and play around with it.
> 
> Added a simple test. Actually parsing the toml in case a default config is provided by some future direnv version is not necessary or is it?

Nah, the test looks good, to me.

khaneliman (approved):
LGTM, thanks for working through that

m4r1vs:
Thanks for reviewing!

PR #6823, opened 2025-04-14
khard: add option to set mutiple subdirs

octvs:
Add new option `accounts.contact.accounts.<name>.khard.addressbooks`.

Remove the previous soln,
`accounts.contact.accounts.<name>.khard.defaultCollection`, which is superseded with the new option.

Previous commit a38f88 allowed a hardcoded path to be set for khard if the path set for its local storage is not the actual `vdir`. This was accomplished via adding the `defaultCollection` option. However this accepted only a single sub-directory, and when one has more than a single collection on the same dir this would require repetition on configuration to set [1].

This is a continuation of the soln given to
nix-community/home-manager#4531, refer to there and the previous PR [2] for reference.

[1]: https://github.com/nix-community/home-manager/issues/4531#issuecomment-2701156246
[2]: https://github.com/nix-community/home-manager/pull/5220

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


octvs:
This is currently on draft status since I need to update the test cases, and
think about edge cases. 

But I wanted to open a PR so that I can receive some feedback already.i

modules/programs/khard.nix
```diff
@@ -103,21 +112,30 @@ in
     xdg.configFile."khard/khard.conf".text =
       let
         makePath =
-          anAccount:
+          x: y:
```
antonmosich, line 115: I'd appreciate to have names here which tell a reader more about the function. E.g. something like `localPath` and `addressbookName`, but you will probably be able to come up with something better than that.

modules/programs/khard.nix
```diff
@@ -88,10 +88,19 @@ in
...
           options.khard.enable = lib.mkEnableOption "khard access";
-          options.khard.defaultCollection = lib.mkOption {
-            type = types.str;
-            default = "";
-            description = "VCARD collection to be searched by khard.";
+          options.khard.addressbooks = lib.mkOption {
+            type = types.listOf types.str;
+            default = [ "" ];
+            description = ''
+              If provided; each item on this list will generate an
```
antonmosich, line 95: I believe this should be a comma instead of a semicolon, shouldn't it?

antonmosich (commented):
Thanks for tackling this! I added a few small notes, but in principle the PR looks good to me!

modules/programs/khard.nix
```diff
@@ -103,21 +112,30 @@ in
...
+        makeName =
+          x: y:
+          lib.concatStringsSep "-" [
+            x
+            y
+          ];
+        expandAccount =
+          acc:
+          lib.concatMapStringsSep "\n" (book: ''
+            [[${makeName acc.name book}]]
```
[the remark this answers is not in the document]
octvs: The reason, I opted for the `concatMapStringsSep` to let the default entry to
be `foo` not `foo-`. The names attributed here are used in cli to specify
address books. These are listed visible via `khard addressbooks`, would be used
as `khard show -a foo-`. Where a trailing `-` would be annoying, imo.

We can still get rid of superfluous functions, which was actually the initial
implementation. However `./format` makes it look much less readable to me. 

```nix
xdg.configFile."khard/khard.conf".text =
  let
    expandAccount =
      account:
      lib.concatMapStringsSep "\n" (book: ''
        [[${
          lib.concatStringsSep "-" [
            account.name
            book
          ]
        }]]
        path = ${
          builtins.toString (
            /.
            + lib.concatStringsSep "/" [
              account.local.path
              book
            ]
          )
        }
      '') account.khard.addressbooks;
  in
```

But both serve as the same. Let me know which one do you prefer.

modules/programs/khard.nix
octvs: Depending on the result of the above thread, I would do that, for sure!

modules/programs/khard.nix
octvs: Yes, thanks!

octvs:
> Thanks for tackling this! I added a few small notes, but in principle the PR looks good to me!

Thanks! I will include the changes here as soon as I'm back to my personal computer.

modules/programs/khard.nix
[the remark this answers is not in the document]
antonmosich: I missed that detail with the trailing `-`, thanks for pointing it out. In that case it makes sense to rely on `lib.concatStringsSep'.
Having the functions in a separate `let in` block is better in my eyes as well.

modules/programs/khard.nix
octvs: The latest version should be more readable

modules/programs/khard.nix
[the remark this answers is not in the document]
octvs: I need to correct myself, while I was adding the tests, I figured the trailing
"-" appeared nevertheless. Therefore I tweaked the logic a bit. Please refer to
the last version.

octvs:
I included the suggestions, although some parts had to be changed after that so
it might be a good idea to skim over again.

At first I thought I should add a news entry but then I didn't see a past entry
for the `defaultCollection` (the option we remove now), therefore I was unsure.
  Let me know if I should write one.

modules/programs/khard.nix
```diff
@@ -88,10 +88,19 @@ in
         with lib.types;
         attrsOf (submodule {
           options.khard.enable = lib.mkEnableOption "khard access";
-          options.khard.defaultCollection = lib.mkOption {
-            type = types.str;
-            default = "";
-            description = "VCARD collection to be searched by khard.";
```
khaneliman, line 94: If we're going to remove an option, please add a `mkRemovedOptionModule` if it's going to be a breaking change. Otherwise, it might be nice to have a `mkRenamedOptionModule` with an apply function if they are supposed to be similar functionality.

octvs: > ... Otherwise, it might be nice to have a `mkRenamedOptionModule` with an apply function if they are supposed to be similar functionality. 

And how can I achieve that?

I tried to add this to the submodule 

```diff
diff --git a/modules/programs/khard.nix b/modules/programs/khard.nix
index 23d806d6..7067334c 100644
--- a/modules/programs/khard.nix
+++ b/modules/programs/khard.nix
@@ -87,6 +87,14 @@ in
       type =
         with lib.types;
         attrsOf (submodule {
+          imports = [
+            (lib.mkChangedOptionModule
+              [ "khard" "defaultCollection" ]
+              [ "khard" "addressbooks" ]
+              (x: [ x.khard.defaultCollection ])
+            )
+          ];
+
           options.khard.enable = lib.mkEnableOption "khard access";
           options.khard.addressbooks = lib.mkOption {
             type = types.listOf types.str;
``` 

To test one can apply the following

```diff
diff --git a/tests/modules/programs/khard/dirty_path.nix b/tests/modules/programs/khard/dirty_path.nix
index 941dbacf..838f1fb5 100644
--- a/tests/modules/programs/khard/dirty_path.nix
+++ b/tests/modules/programs/khard/dirty_path.nix
@@ -5,7 +5,8 @@
       local.type = "filesystem";
       khard = {
         enable = true;
-        addressbooks = [ "named-abook" ];
+        # addressbooks = [ "named-abook" ];
+        defaultCollection = "named-abook";
       };
     };
   };
```

And run 

```sh
$ nix-shell --pure --show-trace tests -A run.khard_dirty_path
...<redacted stack>
       error: The option `accounts.contact.accounts.kebap-case.warnings' does not exist. Definition values:
       - In `/home/oct/git/home-manager/modules/programs/khard.nix':
           [
             "The option `khard.defaultCollection' defined in `/home/oct/git/home-manager/tests/modules/programs/khard/dirty_path.nix' has been changed to `khard.addressbooks' that has a different type. Please read `khard.addressbooks' documentation and update your configuration accordingly."
           ]
```

Which doesn't look alright. 

An alternative idea is to have this in the top level `imports` 

```nix
diff --git a/modules/programs/khard.nix b/modules/programs/khard.nix
index 23d806d6..542ee356 100644
--- a/modules/programs/khard.nix
+++ b/modules/programs/khard.nix
@@ -106,6 +106,15 @@ in
     };
   };
 
+  imports = map
+    (lib.mkChangedOptionModule
+      [ "khard" "defaultCollection" ]
+      [ "khard" "addressbooks" ]
+      (x: [ x.khard.defaultCollection ])
+    )
+    (lib.attrValues accounts)
+    ;
+
 
```
Which throws an infinite recursion.

khaneliman: Is the idea just that the `defaultCollection` is a single item in the `addressbooks` option?

octvs:
> Is the idea just that the `defaultCollection` is a single item in the `addressbooks` option? 

Yeah `defaultCollection` is a string, `addressbooks` accepts a list of them.

modules/programs/khard.nix
octvs: > Is the idea just that the `defaultCollection` is a single item in the `addressbooks` option? 

Yeah `defaultCollection` is a string, `addressbooks` accepts a list of them.

khaneliman:
```diff
diff --git a/modules/programs/khard.nix b/modules/programs/khard.nix
index 23d806d6f..80e224751 100644
--- a/modules/programs/khard.nix
+++ b/modules/programs/khard.nix
@@ -4,8 +4,10 @@
   pkgs,
   ...
 }:
 let
+  inherit (lib) types;
+
   cfg = config.programs.khard;
 
   accounts = lib.filterAttrs (_: acc: acc.khard.enable) config.accounts.contact.accounts;
 
@@ -24,8 +26,9 @@ let
       } "=";
     };
 in
 {
+
   meta.maintainers = [
     lib.hm.maintainers.olmokramer
     lib.maintainers.antonmosich
   ];
@@ -83,14 +86,16 @@ in
       };
     };
 
     accounts.contact.accounts = lib.mkOption {
-      type =
-        with lib.types;
-        attrsOf (submodule {
+      type = types.attrsOf (
+        types.submodule {
+          imports = [
+            (lib.mkRenamedOptionModule [ "khard" "defaultCollection" ] [ "khard" "addressbooks" ])
+          ];
           options.khard.enable = lib.mkEnableOption "khard access";
           options.khard.addressbooks = lib.mkOption {
-            type = types.listOf types.str;
+            type = types.coercedTo types.str lib.toList (types.listOf types.str);
             default = [ "" ];
             description = ''
               If provided, each item on this list will generate an
               entry on khard configuration file as a separate addressbook
@@ -101,9 +106,10 @@ in
               (accounts.contact.accounts.<name>.local.path) for khard. The
               default value will set the aforementioned path as a single vdir.
             '';
           };
-        });
+        }
+      );
     };
   };
 
   config = lib.mkIf cfg.enable {
```

octvs:
Thank you!

I didn't know about `types.coercedTo` before.

PR #6837, opened 2025-04-17
atuin: add support for themes

isabelroses:
### Description

Adds theme support for atuin.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/atuin.nix
```diff
@@ -89,6 +89,26 @@ in
       '';
     };
 
+    themes = mkOption {
+      type = types.attrsOf tomlFormat.type;
```
khaneliman, line 93: Would be nice to support a theme pointing to a file, too. But, that can be done later if someone has interest.

isabelroses: I do belive `pkgs.format.toml` allows for paths: https://github.com/NixOS/nixpkgs/blob/19286d440966ed00115b69c91002932f3e349ee1/pkgs/pkgs-lib/formats.nix#L401

khaneliman: Oh, neat, didn't know that!

khaneliman:
LGTM, but test is failing

modules/programs/atuin.nix
awwpotato: > I do belive `pkgs.format.toml` allows for paths: https://github.com/NixOS/nixpkgs/blob/19286d440966ed00115b69c91002932f3e349ee1/pkgs/pkgs-lib/formats.nix#L401

when passing a path to a toml file it's fails with `Error: Top-level value of type 'str' cannot be encoded as TOML`

isabelroses: Ahh, right thats probably because of the https://github.com/NixOS/nixpkgs/blob/19286d440966ed00115b69c91002932f3e349ee1/pkgs/pkgs-lib/formats.nix#L418. unfortunate

Issue #6834, opened 2025-04-17
bug: Firefox userchrome string cannot start with a multiline comment


flyingpeakock:
### Are you following the right branch?

- [x] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Issue description

If the type of programs.firefox.profiles.<name>.userChrome is a string but it starts with a multiline comment (/*) It will be interpreted as a path, and since that path does not exist the build will fail.

### Maintainer CC

@khaneliman 
@spikespaz 

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.12.23, NixOS, 25.05 (Warbler), 25.05.20250412.2631b0b`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.28.1`
 - channels(root): `"nixos"`
 - nixpkgs: `/nix/store/5ds20jm3x2s4z7wn3581r6lc9ybmh45b-source`
```

flyingpeakock:
The workaround described here https://github.com/nix-community/home-manager/commit/7ede02c32a729db0d6340bdb41d10e73ec511ca0#commitcomment-155487782 also works for this.

spikespaz:
Interesting. I wonder, what is the reason the value string starts with `/*`? Is this a multi-line string or did you call `builtins.readFile`? If it is a multi-line string, I presume you're generating CSS from Nix and desire the comments to be output to the `userChrome.css`?

The first thing that comes to mind is perhaps you used `readFile`, and in that event, now you don't have to? 

Still should probably add special behavior for `/*`, but I'm curious your use-case.

johnhamelink:
This issue seems to have arisen within the last few days, as I've been using a userChrome.css with a comment in the first line for many months without issue. Thanks to @flyingpeakock for flagging a workaround in the meantime

khaneliman:
Well, that's interesting... didn't think the path function would just trip up over the contents of lines...

flyingpeakock:
> Interesting. I wonder, what is the reason the value string starts with `/*`? Is this a multi-line string or did you call `builtins.readFile`? If it is a multi-line string, I presume you're generating CSS from Nix and desire the comments to be output to the `userChrome.css`?
> 
> The first thing that comes to mind is perhaps you used `readFile`, and in that event, now you don't have to?
> 
> Still should probably add special behavior for `/*`, but I'm curious your use-case.

[...]

PR #6844, opened 2025-04-18
mkFirefoxModule: userchrome support derivations

khaneliman:
### Description
@spikespaz 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


khaneliman:
@spikespaz does this work for you?

spikespaz:
Replying to <https://github.com/nix-community/home-manager/pull/6836#issuecomment-2816327088>
This is what I would have contributed to resolve <https://github.com/nix-community/home-manager/issues/6834>, but I waited for your solution hoping it would be a better idea.
```nix
            # Handle the edge-case where a CSS string is treated as a file path
            # where the content begins with a comment.
            startsWithCssComment = v: lib.isString v && lib.hasPrefix "/*" v;

            chromePath =
              if ((v: !(startsWithCssComment v) && lib.pathIsDirectory v) profile.userChrome) then
                "chrome"
              else
                "chrome/userChrome.css";
            sourcePath =
              if ((v: !(startsWithCssComment v) && lib.types.path.check v) profile.userChrome) then
                profile.userChrome
              else
                null;
```

Indeed, I'm not sure if including an `isDerivation` check also would fix all corner cases. I put a lot of thought into the first PR but didn't consider the actual CSS syntax. I can't recall all of that consideration to adjust it.

spikespaz:
I seem to remember settling on `lib.types.path.check` *instead of* `lib.isPath` because they have different behavior. At least, I think we *don't* want `isPath` in at least `sourcePath`, but fail to remember why I discovered this.

spikespaz:
Ideally the proper solution would be to make the `userChrome` option have a type `types.nullOr (types.oneOf [ types.lines configFile ])` and introduce a deprecation notice that `text` should be specified for the `configFile` submodule instead of using `types.lines` directly.

khaneliman:
> I seem to remember settling on `lib.types.path.check` _instead of_ `lib.isPath` because they have different behavior. At least, I think we _don't_ want `isPath` in at least `sourcePath`, but fail to remember why I discovered this.

I had to check isPath or else the path.check throws an error. That’s why I created the test cases.

spikespaz:
> > I seem to remember settling on `lib.types.path.check` _instead of_ `lib.isPath` because they have different behavior. At least, I think we _don't_ want `isPath` in at least `sourcePath`, but fail to remember why I discovered this.
> 
> I had to check isPath or else the path.check throws an error. That’s why I created the test cases.

Right, and I think that's what is convincing me now that my first PR is a hack and a bad idea, if the option types check functions aren't sufficient validation.

[This](https://github.com/nix-community/home-manager/pull/6844#issuecomment-2816394479) is what I should have done. What are your thoughts?

khaneliman:
> [This](https://github.com/nix-community/home-manager/pull/6844#issuecomment-2816394479) is what I should have done. What are your thoughts?

Not sure, would have to see/think about the implementation a little more. I don't really like the idea of discouraging passing lines to the option directly. I'll merge this to unbreak supporting derivations and we can rework it more if we want to change the accepted types.

spikespaz:
sgtm

PR #6847, opened 2025-04-19
mkFirefoxModule: userChrome: now of type sourceFileOrLines

spikespaz:
### Description

The user can now provide `types.lines` or a submodule with options `text`, `source` and `recursive`. If multiple definitions of differing types are evaluated, they will be merged as if they were simply `types.lines`.

### Checklist

- [NO] Change is backwards compatible.

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@khaneliman


khaneliman:
> This does resolve the issue to some extent but:
> 
>     * It is backwards incompatible (requires changing `userChrome =` to e.g. `userChrome.source = ` in case the value was a file drv)
> 
>     * If `userChrome` is set to a dir, and `userContent` is also set, then cryptic error `Error installing file '.librewolf/default/chrome/userContent.css' outside $HOME` is given
> 
>     * As stated in [bug: firefox userChrome with a derivation that is a file fails #6882](https://github.com/nix-community/home-manager/issues/6882), I think it would be quite a bit simpler to restore `userContent` to the previous state and add a separate option e.g. `chromeDir` and assertions that `chromeDir` is mutually exclusive with `user{Content,Chrome}`. This should preserve backwards compatibility and remove the need for the complex new submodule options (aside from tests it should only be about a dozen lines I think?).

Yeah, I'm leaning more towards `userChrome` and `chromeDir`... to reduce complexity and keep it obvious what the responsibility of each option is.

spikespaz:
With [#6761], I introduced the ability to use either a derivation or a file path in a way that preserves backward compatibility and requires minimal future maintenance. My goal was simply to remove a hack from my dotfiles--not to shoot for perfection, but to be pragmatic.

I always intended to handle `userContent.css` in a follow-up, but I wanted to finalize the design for `userChrome` before adjusting the corresponding option for the former.

This change was poorly tested and ultimately broken; it should have been reverted at the first sign of trouble rather than patched over with [#6833], [#6836], and [#6844]. Now that we’re here bikeshedding, any further adjustments can’t realistically be considered backward-compatible.

I chose not to introduce `chromeDir` ([#6884]) because it would require checks and assertions to ensure mutual exclusivity between `chromeDir` and `userChrome`. I find mutually exclusive options unintuitive—it fragments the documentation and forces users to think harder. That said, the assertion is there to catch the lazy, so perhaps this isn't a big deal.

Instead, I believe that [#6847] is more viable as a generalized solution, offering several advantages:
- It follows a familiar pattern (similar to `home.file` and `xdg.configFile`).
- It’s portable: I can reuse the `sourceFile` and `sourceFileOrLines` factories in other modules without reinventing the wheel or enforcing mutual exclusivity.
- It’s arguably more intuitive than `chromeDir`, once you adjust your mental model to accept `sourceFileOrLines` with a custom `merge`.

Regarding [https://github.com/nix-community/home-manager/pull/6847#issuecomment-2822046668](https://github.com/nix-community/home-manager/pull/6847#issuecomment-2822046668) in relation to [#6847]:
- This change is backward compatible--if you accept that the post-[#6761] patches are not part of the stable API history.
- I plan to address `userContent` in a follow-up, but I can tackle it now if preferred.
- It restores `userChrome` to its previous behavior and is purely additive. I find this approach clearer (the submodule and `sourceFileOrLines` are self-documenting) compared to the mutual-exclusivity logic in [#6884]. Plus, it establishes a pattern I’d like to reuse.

Regarding [https://github.com/nix-community/home-manager/pull/6847#issuecomment-2822063767](https://github.com/nix-community/home-manager/pull/6847#issuecomment-2822063767):
- The original module had a broken edge case: passing a path to `userChrome` would insert that path into the generated `userChrome.css` file because the type-check would accept any string (see the `isString` branch: [https://github.com/nix-community/home-manager/blob/cba2f9ce95c8d10b66cacf05a275e3ad71959638/modules/programs/firefox/mkFirefoxModule.nix#L872](https://github.com/nix-community/home-manager/blob/cba2f9ce95c8d10b66cacf05a275e3ad71959638/modules/programs/firefox/mkFirefoxModule.nix#L872)). I'd lived with a workaround in my dotfiles for two years due to this issue.

[#6761]: https://github.com/nix-community/home-manager/pull/6761
[#6833]: https://github.com/nix-community/home-manager/pull/6833
[#6836]: https://github.com/nix-community/home-manager/pull/6836
[#6844]: https://github.com/nix-community/home-manager/pull/6844
[#6847]: https://github.com/nix-community/home-manager/pull/6847
[#6884]: https://github.com/nix-community/home-manager/pull/6884

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -202,6 +203,71 @@ let
       (pkgs.wrapFirefox.override { config = bcfg; }) package { };
 
   bookmarkTypes = import ./profiles/bookmark-types.nix { inherit lib; };
+
+  sourceFile =
```
khaneliman, line 207: I like the concept of this type, but agree it should be pulled out into the `lib` so other modules can take advantage of it.

tests/modules/programs/firefox/profiles/userchrome/default.nix
```diff
@@ -21,15 +18,39 @@ in
...
           id = 1;
-          userChrome = builtins.readFile ./chrome/userChrome.css;
+          userChrome = dummyUserChromeText;
         };
-        file = {
+        file-text = {
           id = 2;
-          userChrome = ./chrome/userChrome.css;
+          userChrome.text = dummyUserChromeText;
+        };
```
khaneliman, line 26: It would be nice to see a backwards compatible test case of passing the file, still.

khaneliman (commented):
When building against my current config:

```bash
 error: syntax error, unexpected ':'
┃        at /nix/store/bh64wbqs6g6fw42hc035i5jbwigaizv6-vrislri6h9fas9i9z0nj0j3hr6pr1p1w-source/modules/home/programs/graphical/browsers/firefox/chrome/userChrome.css:1:1:
┃             1| :root {
┃              | ^
┃             2|   --base: #24273a;
```

modules/programs/firefox/mkFirefoxModule.nix
spikespaz: Do you think we should observe it in-practice for a while first? I promise, I won't forget about it.

tests/modules/programs/firefox/profiles/userchrome/default.nix
spikespaz: What name do you think is missing?

khaneliman: I meant, this no longer has the `file` test which had the path being passed in. There are 2 tests for the lines which are read from the file itself, one with `userChrome` and one with `userChrome.text`. `userChrome = file.css` throws an error, for me because it tries to read the contents of the file always.

spikespaz:
> When building against my current config:
> 
> ```shell
>  error: syntax error, unexpected ':'
> ┃        at /nix/store/bh64wbqs6g6fw42hc035i5jbwigaizv6-vrislri6h9fas9i9z0nj0j3hr6pr1p1w-source/modules/home/programs/graphical/browsers/firefox/chrome/userChrome.css:1:1:
> ┃             1| :root {
> ┃              | ^
> ┃             2|   --base: #24273a;
> ```

<https://github.com/khaneliman/khanelinix/blob/main/modules/home/programs/graphical/browsers/firefox/default.nix#L175>

This PR does not retain compatibility with #6761, please use `.source`.

tests/modules/programs/firefox/profiles/userchrome/default.nix
spikespaz: The `file` text has been replaced by `source-file` because these changes are additive to cba2f9ce95c8d10b66cacf05a275e3ad71959638 only.

khaneliman:
> This PR does not retain compatibility with #6761, please use `.source`.

That is not when it was introduced.  https://github.com/nix-community/home-manager/pull/3856 That has been in the wild for a long time.

khaneliman:
If we introduce a breaking change, we would at least need an assertion of some kind with migration instructions. I agree we bring back heuristics to try and create a type coercion for just a warning.

spikespaz:
The error is occurring before the option values are merged. With my limited knowledge, I found a compromise. If you're aware of a proper way to handle this, please suggest.

This PR now to be reviewed per-commit.

tests/modules/programs/firefox/profiles/userchrome/default.nix
```diff
@@ -21,15 +18,39 @@ in
...
+        source-file = {
+          id = 3;
+          userChrome.source = ./chrome/userChrome.css;
         };
-        derivation-file = {
+        source-dir = {
           id = 4;
-          userChrome = pkgs.writeText "userChrome.css" (builtins.readFile ./chrome/userChrome.css);
+          userChrome.source = ./chrome;
+        };
```
khaneliman, line 34: This still feels wrong, though, in the context of `userChrome` overwriting the contents of the entire `chrome` directory. Seems like it would make more sense to have a `chrome` option that has a type supporting a folder or a submodule with the `userChrome` and `userContent` options that support lines or paths.

spikespaz: I guess I'm not really sure what you don't like about it or why. Despite my lack of understanding, here's why I think it doesn't matter:

If someone wants to customize it themselves, they will use one of:
- a. `userChrome.source = ./chrome;` (implies the directory contains `userChrome.css`)
- b. `userChrome.source = ./userChrome.css;`
- c. `userChrome = ''#custom { style: 1; }'';`

If they want to use a pre-made theme, they will probably use:
- d. `userChrome.source = "${someDerivation}/extras/firefox/userChrome.css";`
- e. `userChrome.source = someDerivation;`

I think I understand that to you, A and E don't make sense, because they are not files, but `userChrome` (the option) is the name of a file in `chrome`, thus, the option should not correspond with a directory.

My argument would be, and I'm not confident, that more people use one of the predictable above options, and don't need, want, or care about the mutability of the linked `chrome` directory. If they're making their own theme, and it lives in the same repository, it could be a directory or a file. The goal of that directory or file is to produce `userChrome.css`, regardless of how many CSS files the theme itself depends on.

In other words, whether file or directory, the *goal* is *named* "user chrome".

Now, `userContent` lives in there too, unfortunately, but that's what `recursive` is for. Setting `userContent` necessarily requires `recursive`, and thus, when I get around to it, `recursive = lib.mkDefault (userContent != null);`.

Say there's another scenario where they use a pre-packaged theme, option E above, and want to make adjustments. Now they're in warranty-void territory. They have many options, among them: overriding the derivation, passing their own text that imports the store-bought `userChrome.css`, and so-on.

Did I understand your grievance? I sort of pulled out a shotgun of thoughts, but maybe I hit something.

spikespaz: In fewer words, my favorite theme is WaveFox. WaveFox *is* `userChrome`, but just happens to have multiple files. It *is not* `userContent`. Having `chrome.source = pkgs.wavefox`, for example, would mean you're still not able to set `chrome.userContent` without `recursive`. WaveFox is also not `chrome`, its scope is isolated to `userChrome`, exclusive of `userContent`.

I don't think there's a perfect design here.

PR #6860, opened 2025-04-20
vesktop: add support for multiple themes

awwpotato:
### Description
adds to #5587

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@Flameopathic @LilleAila

### DIFF
--- a/modules/programs/vesktop.nix
+++ b/modules/programs/vesktop.nix
@@ -50,12 +50,15 @@ in
 
     vencord = {
       useSystem = lib.mkEnableOption "Vencord package from Nixpkgs";
-      theme = lib.mkOption {
-        description = "The theme to use for Vencord";
-        default = null;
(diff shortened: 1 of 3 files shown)


modules/programs/vesktop.nix
```diff
@@ -50,12 +50,12 @@ in
 
     vencord = {
       useSystem = lib.mkEnableOption "Vencord package from Nixpkgs";
-      theme = lib.mkOption {
-        description = "The theme to use for Vencord";
-        default = null;
+      themes = lib.mkOption {
+        description = "Themes to add for Vencord";
```
Flameopathic, line 54: Because the user will have to specify these theme files as active themes, consider mentioning in the description exactly how the theme files are named.

Flameopathic (approved):

PR #6861, opened 2025-04-21
atuin: Fix deprecated string type warning

JoaquinTrinanes:
### Description

Right now, building the configuration reports a warning:

> trace: evaluation warning: The type `types.string` is deprecated. See https://github.com/NixOS/nixpkgs/pull/66346 for better alternative types.

Technically the code is not equivalent because `str` and `string` behave differently, but given that the change was introduced ~4h ago I wouldn't consider it _that_ breaking.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@hawkw @water-sucks

### DIFF
--- a/modules/programs/atuin.nix
+++ b/modules/programs/atuin.nix
@@ -94,7 +94,7 @@ in
         types.oneOf [
           tomlFormat.type
           types.path
-          types.string
+          types.lines
         ]
       );
       description = ''


modules/programs/atuin.nix
```diff
@@ -94,7 +94,7 @@ in
         types.oneOf [
           tomlFormat.type
           types.path
-          types.string
+          types.str
```
awwpotato, line 97:
```suggestion
          types.lines
```

JoaquinTrinanes: Sensible, just updated it

PR #6868, opened 2025-04-21
git: support maintenance on darwin

joshheinrichs-shopify:
### Description

This doesn't yet work for folks using git-credential-osxkeychain as it isn't placed on git's exec path[1]. I mostly attempted to mimic the agents that git maintenance generates for itself. The settings differ slightly because git configures calendar intervals incorrectly[2], and git started injecting some additional config settings recently[3] which would be good to mimic in another PR.

[1] https://github.com/NixOS/nixpkgs/pull/347515
[2] https://lore.kernel.org/git/20250421054633.231069-1-joshiheinrichs@gmail.com
[3] https://github.com/git/git/commit/4f5551957ddd6c679fc8499d4e1e110d43a49071

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@khaneliman @rycee


modules/programs/git.nix
```diff
@@ -753,6 +753,54 @@ in
               };
           in
           lib.attrsets.mapAttrs' toSystemdTimer cfg.maintenance.timers;
+
+        launchd.agents =
+          let
+            baseArguments = [
+              "${cfg.package}/bin/git"
```
khaneliman, line 760:
```suggestion
              "${lib.getExe cfg.package}"
```

modules/programs/git.nix
```diff
@@ -753,6 +753,54 @@ in
...
+            ];
+          in
+          {
+            "git-maintenance-hourly" = {
+              enable = true;
+              config = {
+                ProgramArguments = baseArguments ++ [ "--schedule=hourly" ];
+                StartCalendarInterval = map (hour: {
+                  Hour = hour;
+                  Minute = 53;
```
khaneliman, line 775: What's the significance of the `Minute = 53` in all of these units?

khaneliman (commented):
LGTM overall, thanks

modules/programs/git.nix
joshheinrichs-shopify: i was just matching the times above

  https://github.com/nix-community/home-manager/blob/ddc65de72b8d46bdd63bb2da28697dfe9e34c1eb/modules/programs/git.nix#L282

[git picks a random minute when generating maintenance jobs to avoid everyone fetching at the same time](https://github.com/git/git/blob/master/builtin/gc.c#L1990). i don't think there's a super easy way to do that from nix? i could be wrong.

khaneliman (approved):

PR #6870, opened 2025-04-21
nushell: throw instead of abort

JoaquinTrinanes:
### Description

This PR prevents `lib.hm.nushell.toNushell` from unconditionally aborting the evaluation if an error is detected, allowing consumers to `tryEval` the result and provide a fallback value.

The `abort` was there originally because it's what the `toLua` function it's based on does, but I think it makes sense to `throw` in most cases. I'd even argue this should be a treewide change :)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@Philipp-M @aidalgol

### DIFF
--- a/modules/lib/nushell.nix
+++ b/modules/lib/nushell.nix
@@ -81,5 +81,5 @@ rec {
           }${outroSpace}}"
       )
     else
-      abort "nushell.toNushell: type ${lib.typeOf v} is unsupported";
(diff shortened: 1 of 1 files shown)


aidalgol:
Seems reasonable to me.  @Philipp-M is the main maintainer of the home-manager Nushell module, so I'll defer to him on this one.

PR #6872, opened 2025-04-21
home-environment: add option for relocatable configurations

emilazy:
### Description

A relocatable configuration doesn’t know its username or home directory at evaluation time, and can therefore be evaluated and built once and applied to many users. It has inherent limitations, such as the inability to produce configuration files that require reference to the username or the absolute home directory path, but can support most Home Manager functionality out of the box.

This is accomplished by adding a new `homePath` type to represent either a known absolute path, or a path relative to a potentially‐unknown home directory.

This should generally be backwards‐compatible with existing modules and configurations, as string interpolation produces an absolute path as before, although there are edge cases (e.g. direct comparison of such paths to strings). Modules that are adapted to relocatable configurations use more specific attributes of the type as appropriate, like `shell` to render a shell script fragment.

Less churn across the tree could be achieved by picking different representations at the cost of backwards compatibility. For instance, `xdg.configHome` is seemingly invariably used to construct a `home.files` path rather than to represent the absolute path it previously was; it could instead interpolate as a relative path to make that work with relocatable configurations without having to use `xdg.configHome.join`. I chose to go with the most conservative option in terms of compatibility: anything that was previously guaranteed to be an absolute path is still an absolute path when coerced to a string, even if that means throwing an error for relocatable configurations. In any case, the required adaptations for any given module are usually minimal.

At present, integration into a NixOS or nix-darwin configuration is not supported, although it would be feasible, and would be useful for integration testing the feature and speeding up evaluation and builds of system configurations with large numbers of similarly‐configured users. `useUserPackages` would need some adjustments to the design to handle late‐bound variables in `home.profileDirectory` beyond a home directory prefix in order to encode `/etc/profiles/per-user/$USER`, or else a different path structure for relocatable configurations. I have prototyped the former solution, but not implemented it here, as it is slightly more complex and it is unclear if there is interest in it.

This does cement the assumption that `home.homeDirectory` matches the runtime `$HOME` value. This is already assumed elsewhere in the codebase and explicitly checked unless `SKIP_SANITY_CHECKS` is set, so hopefully that is acceptable.

---

I’m putting this draft PR up as an RFC on the basic idea behind the functionality before pursuing this further. It needs more work – there’s no documentation, the current testing is fairly rudimentary and could use a proper integration test, and of course there’s a long tail of in‐tree modules that would need conversion along the lines of the example commits here to enable them to work with relocatable configurations.

It does add some unavoidable complexity to support evaluating and building configurations that are agnostic to the username and home directory, so I can understand if there are reservations about supporting this, but I have tried to take an approach that keeps existing modules and configurations working as much as possible, makes the required modifications to enable modules to work with relocatable configurations as simple as possible, and puts most of the complexity in the implementation of `homePath`.

If there’s agreement that this would be reasonable to support, I can finish polishing this off, add proper tests and documentation, and convert more of the in‐tree modules as desired.

Open questions and potential alternative paths that could be taken:

* Naming of the feature – I went with “relocatable”, but “portable” also seems plausible. Happy to hear suggestions here.

* Perhaps `.toShell`, `.toSystemd`, and so on would be better than the current `.shell` and `.systemd`.

* It might be nice to do more normalization in the `homePath` type. `~/foo` and `/home/hm-user/foo` aren’t treated as equal even when `home.homeDirectory == "/home/hm-user"`, and so `config.home.homeDirectory.join "foo"` can be subtly different to `config.lib.homePath.parse "~/foo"`.

* The type is injected directly as `config.lib.homePath.type`. The implementation strategy requires it to depend on `config.home.homeDirectory`, so it can’t go directly into `lib`. `fileType` has the same issue for the same reason, and is currently imported directly across the tree. That seemed inconvenient here, but I’m happy to adjust it to work like `fileType`, or move `fileType` over to the same scheme, as desired.

* As outlined above, if we wanted to handle username placeholders (required for things like the `useUserPkgs` profile path), we would need a more general “templated string” type, like JavaScript’s [tagged templates](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates) or Python’s upcoming [template strings](https://davepeck.org/2025/04/11/pythons-new-t-strings/). I’ve implemented most of the work for this and it seemed to work fine, including handling `%u` specifiers in systemd units and so on, but I stuck with the simpler `homePath` approach for now, as few things require the username and only handling the “absolute or home‐relative path” case simplifies the implementation. Still, supporting external `/etc/profiles/per-user` package installation may be nice in the kind of managed environment this feature is suitable for, so I’m happy to take that path instead if desired.

* We could just make every path that starts with the home directory relative instead, make `home.homeDirectory` solely for getting the absolute path, and avoid having a separate `homePath` type altogether. However, that would break anything that assumes `xdg.configHome` and the like are absolute paths, as they are currently. It seems like nothing in‐tree depends on that assumption, so it would be fairly easy to make the change, but it’s hard to measure the effect on external configurations. Implementation‐wise, this would be much simpler, but it wouldn’t handle things like paths in environment variables and systemd services which are handled fairly elegantly with the `homePath` solution.

* We could set `home.username` and `home.homeDirectory` to some unique placeholder string and then replace everything; this would mean that the types could remain the same, all the complexity would be in one place, and it would even support templating absolute paths and username references into files. But this would be pretty complicated, hard‐to‐impossible to get right across multiple formats (shell quoting etc.), and require rewriting entire sets of derivations and references between them, so it seems too gnarly to be worth it to me.

---

**Note:** This work was funded by Determinate Systems.

### Checklist

- [x] Change is backwards compatible. *(mostly; see caveats above)*

    or `nix build --reference-lock-file flake.lock ./tests#test-all` using Flakes. *(on both `x86_64-darwin` and `x86_64-linux`)*

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

cc @rycee @khaneliman


emilazy:
Rebased for conflicts.

khaneliman:
Apologies, forgot about this one.. I'll try and set some time aside to check this out.

PR #6873, opened 2025-04-21
process-compose: init module

bbigras:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/process-compose.nix
+++ b/modules/programs/process-compose.nix
@@ -1,0 +1,33 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  inherit (lib) mkIf;
+  cfg = config.programs.process-compose;
+  yamlFormat = pkgs.formats.yaml { };
+in
+{
+  meta.maintainers = [ lib.hm.maintainers.bbigras ];
+  options.programs.process-compose = {
+    enable = lib.mkEnableOption "the process-compose orchestrator";
+    package = lib.mkPackageOption pkgs "process-compose" { nullable = true; };
+    settings = lib.mkOption {
+      type = yamlFormat.type;
+      default = { };
+      defaultText = lib.literalExpression "{}";
+      description = "The process-compose configuration.";
+    };
+  };
+  config = mkIf cfg.enable {
+    home.packages = mkIf (cfg.package != null) [ cfg.package ];
+
+    xdg.configFile."process-compose/settings.yaml" =
+      mkIf (cfg.settings != null && cfg.settings != { })
+        {
+          source = yamlFormat.generate "settings" cfg.settings;
+        };
+  };
+}
--- a/modules/lib/maintainers.nix
+++ b/modules/lib/maintainers.nix
@@ -54,6 +54,12 @@
     github = "bamhm182";
     githubId = 920269;
   };
+  bbigras = {
+    name = "bbigras";
+    email = "24027+bbigras@users.noreply.github.com";
+    github = "bbigras";
+    githubId = 24027;
+  };
   blmhemu = {
     name = "blmhemu";
     email = "19410501+blmhemu@users.noreply.github.com";
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -223,6 +223,7 @@ let
       ./programs/poetry.nix
       ./programs/powerline-go.nix
       ./programs/pqiv.nix
+      ./programs/process-compose.nix
       ./programs/pubs.nix
       ./programs/pyenv.nix
       ./programs/pylint.nix
--- a/tests/default.nix
+++ b/tests/default.nix
@@ -432,6 +432,7 @@ import nmtSrc {
       ./modules/programs/pls
       ./modules/programs/poetry
       ./modules/programs/powerline-go
+      ./modules/programs/process-compose
       ./modules/programs/pubs
       ./modules/programs/pyenv
       ./modules/programs/qcal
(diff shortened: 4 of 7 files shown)


bbigras:
Oh there is also #5788

modules/programs/process-compose.nix
```diff
@@ -0,0 +1,33 @@
...
+      type = yamlFormat.type;
+      default = { };
+      defaultText = lib.literalExpression "{}";
+      description = "The process-compose configuration.";
+    };
+  };
+  config = mkIf cfg.enable {
+    home.packages = mkIf (cfg.package != null) [ cfg.package ];
+
+    xdg.configFile."process-compose/settings.yaml" =
```
khaneliman, line 27: Is the path the same for darwin (with or without xdg enabled) and linux? 

I see the other PR doing this https://github.com/nix-community/home-manager/pull/5788/files#diff-0250cdf0ec8bf9d6c96a28941631125be687497d95ea33f037397f1d20abd5f0R48-R51

modules/programs/process-compose.nix
```diff
@@ -0,0 +1,33 @@
...
+in
+{
+  meta.maintainers = [ lib.hm.maintainers.bbigras ];
+  options.programs.process-compose = {
+    enable = lib.mkEnableOption "the process-compose orchestrator";
+    package = lib.mkPackageOption pkgs "process-compose" { nullable = true; };
+    settings = lib.mkOption {
+      type = yamlFormat.type;
+      default = { };
+      defaultText = lib.literalExpression "{}";
```
awwpotato, line 20: this seems unnecessary

bbigras: What do you mean?

I would like to do something like:
```nix
settings = {
  theme = "Custom Style";
};
```

Would you prefer that `theme` would be typed with a `theme` attr?

awwpotato: what I mean is `defaultText` is only necessary when default is a dynamic value, when default is `{ }` there's no need to set `defaultText`

modules/programs/process-compose.nix
bbigras: Thanks for the comment. I have no idea how darwin works, so I just took the same code as the other PR.

modules/programs/process-compose.nix
bbigras: Thanks, TIL. I removed it.

modules/programs/process-compose.nix
```diff
@@ -0,0 +1,42 @@
...
+          "${config.xdg.configHome}/process-compose";
+    in
+    mkIf cfg.enable {
+      home = {
+        packages = mkIf (cfg.package != null) [ cfg.package ];
+
+        file.process-compose-settings = {
+          enable = cfg.settings != null && cfg.settings != { };
+          target = "${configDir}/settings.yaml";
+          source = settingsFormat.generate "settings.yaml" cfg.settings;
```
awwpotato, line 38:
```suggestion
          source = settingsFormat.generate "process-compose-settings" cfg.settings;
```

modules/programs/process-compose.nix
```diff
@@ -0,0 +1,42 @@
...
+          "Library/Application Support/process-compose"
+        else
+          "${config.xdg.configHome}/process-compose";
+    in
+    mkIf cfg.enable {
+      home = {
+        packages = mkIf (cfg.package != null) [ cfg.package ];
+
+        file.process-compose-settings = {
+          enable = cfg.settings != null && cfg.settings != { };
```
awwpotato, line 36: settings can't be null here
```suggestion
          enable = cfg.settings != { };
```

modules/programs/process-compose.nix
```diff
@@ -0,0 +1,42 @@
...
+in
+{
+  meta.maintainers = [ lib.hm.maintainers.bbigras ];
+  options.programs.process-compose = {
+    enable = lib.mkEnableOption "the process-compose orchestrator";
+    package = lib.mkPackageOption pkgs "process-compose" { nullable = true; };
+    settings = lib.mkOption {
+      type = settingsFormat.type;
+      default = { };
+      description = "The process-compose configuration.";
```
0xda157: linking to document here would be nice, also adding a `example`

bbigras: Done.

modules/programs/process-compose.nix
```diff
@@ -0,0 +1,51 @@
...
+  meta.maintainers = [ lib.hm.maintainers.bbigras ];
+  options.programs.process-compose = {
+    enable = lib.mkEnableOption "the process-compose orchestrator";
+    package = lib.mkPackageOption pkgs "process-compose" { nullable = true; };
+    settings = lib.mkOption {
+      type = settingsFormat.type;
+      default = { };
+      description = ''
+        The process-compose configuration.
+        See [docs](`https://f1bonacc1.github.io/process-compose/tui/?h=settings.yaml#tui-state-settings`).
```
0xda157, line 22: shouldn't it be this?
```suggestion
        See [docs](`https://f1bonacc1.github.io/process-compose/configuration/`).
```

bbigras: I'm interested in generating the `settings.yaml` file, which https://f1bonacc1.github.io/process-compose/configuration/ doesn't seem to mention.

0xda157: oh I see, why don't you add support for `shortcuts.yaml` and `process-compose.yml`?

PR #6874, opened 2025-04-21
gpg-agent: support pinentryBinary

colemickens:
### Description

This adds a new option, `pinentryBinary` to the `gpg-agent` service definition. And and assertion that it is not set at the same time as `pinentryPackage`.

This is to support prompts like `wayprompt` that have than one exe output, or don't follow the naming assumption that `pinentryPackage` relies on.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

~~- If this PR adds a new module~~

### DIFF
--- a/modules/services/gpg-agent.nix
+++ b/modules/services/gpg-agent.nix
@@ -296,6 +296,7 @@ in
           configuration file.
         '';
       };
+
       pinentryPackage = mkOption {
         type = types.nullOr types.package;
         example = lib.literalExpression "pkgs.pinentry-gnome3";
@@ -308,6 +309,15 @@ in
           ```nix
           home.packages = [ pkgs.gcr ];
           ```
+        '';
+      };
+
+      pinentryBinary = mkOption {
+        type = types.nullOr (types.path);
+        example = "\${pkgs.wayprompt}/bin/wayprompt-pinentry";
+        default = null;
+        description = ''
+          Which binary to use for `pinentry-program`.
         '';
       };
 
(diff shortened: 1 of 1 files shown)


modules/services/gpg-agent.nix
```diff
@@ -336,6 +354,7 @@ in
           ++ optional (cfg.maxCacheTtl != null) "max-cache-ttl ${toString cfg.maxCacheTtl}"
           ++ optional (cfg.maxCacheTtlSsh != null) "max-cache-ttl-ssh ${toString cfg.maxCacheTtlSsh}"
           ++ optional (cfg.pinentryPackage != null) "pinentry-program ${lib.getExe cfg.pinentryPackage}"
+          ++ optional (cfg.pinentryBinary != null) "pinentry-program ${cfg.pinentryBinary}"
```
khaneliman, line 357: Wondering if, instead of supporting 2 different configurations for the same option, we just deprecate the package option in favor of supplying the program itself. 

Another idea, was that maybe we change it to be a `pinentry` option with sub options for `package` and `program` and `program` accepts a string to forward to `lib.getExe' cfg.pinentryPackage`.

colemickens: Thanks for the review -- No opinion here. If you tell me which one you like better, I'll do it.

khaneliman: Created https://github.com/nix-community/home-manager/pull/6900, does that look like what you were hoping for in functionality?

colemickens: Nice. Let me pull that and test it out in a bit. Thanks!

PR #6879, opened 2025-04-22
keepassxc: register as native messaging host

bmrips:
### Description

Register KeepassXC as a native messaging host in supported browsers to integrate with KeepassXC's browser extension.  See [this Discourse thread](https://discourse.nixos.org/t/keepassxc-browser-extension-integration/39553) for more information.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@d-brasher

### DIFF
--- a/modules/programs/keepassxc.nix
+++ b/modules/programs/keepassxc.nix
@@ -45,12 +45,26 @@ in
     };
   };
 
-  config = lib.mkIf cfg.enable {
-    home.packages = lib.mkIf (cfg.package != null) [ cfg.package ];
-    xdg.configFile = {
-      "keepassxc/keepassxc.ini" = lib.mkIf (cfg.settings != { }) {
-        source = iniFormat.generate "keepassxc-settings" cfg.settings;
-      };
-    };
-  };
+  config = lib.mkIf cfg.enable (
+    lib.mkMerge [
+
+      {
+        xdg.configFile = {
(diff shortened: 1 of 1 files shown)


modules/programs/keepassxc.nix
```diff
@@ -45,12 +45,26 @@ in
...
+        };
+      }
+
+      (lib.mkIf (cfg.package != null) {
+        home.packages = [ cfg.package ];
+        programs.brave.nativeMessagingHosts = [ cfg.package ];
+        programs.chromium.nativeMessagingHosts = [ cfg.package ];
+        programs.firefox.nativeMessagingHosts = [ cfg.package ];
+        programs.floorp.nativeMessagingHosts = [ cfg.package ];
+        programs.vivaldi.nativeMessagingHosts = [ cfg.package ];
```
khaneliman, line 65: Just wondering if we want it to default to installing it as a native messaging host or if it should be gated behind a `enableBrowserIntegration` option.

bmrips: AFAIK, there is not downside to enabling the integration.

khaneliman: Mostly just worry about people who don't like unexpected installations/configurations or can't disable stuff. But, I agree it is probably more confusing for the application to just not work with your browser...

khaneliman:
Merging, as-is, for now. Could see a bigger effort towards standardizing an approach for this if it becomes more common, though.

PR #6885, opened 2025-04-23
i18n.inputMethod: align enable option with nixos

awwpotato:
confusing for home-manager to use the deprecated style from nixos 
see: https://github.com/NixOS/nixpkgs/pull/310708

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@Kranzes

### DIFF
--- a/modules/i18n/input-method/default.nix
+++ b/modules/i18n/input-method/default.nix
@@ -133,5 +133,8 @@ in
     ];
   };
 
-  meta.maintainers = [ lib.hm.maintainers.kranzes ];
+  meta.maintainers = [
+    lib.hm.maintainers.kranzes
+    lib.maintainers.awwpotato
+  ];
 }
--- a/modules/i18n/input-method/default.nix
+++ b/modules/i18n/input-method/default.nix
@@ -4,10 +4,17 @@
   lib,
   ...
 }:
-
 let
-
   cfg = config.i18n.inputMethod;
+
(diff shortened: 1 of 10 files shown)


khaneliman (approved):

khaneliman:
any objections @Kranzes

awwpotato:
> any objections @Kranzes

they seem to not be very active and [don't have inputMethod in their config](https://github.com/search?q=repo%3AKranzes%2Fnix-config%20inputMethod&type=code). Might be best for me to take over (the root of) this module.

sifmelcara:
Hi @awwpotato , thanks for cleaning this up

However migrating from `enabled` to `enable` breaks input method on my machine

I didn't test to confirm but at the first glance you forgot to update `config = lib.mkIf (cfg.enabled != null)` to `config = lib.mkIf cfg.enable`  in `modules/i18n/input-method/default.nix` (not sure if there are other places also need change)

awwpotato:
> Hi @awwpotato , thanks for cleaning this up
> 
> However migrating from `enabled` to `enable` breaks input method on my machine
> 
> I didn't test to confirm but at the first glance you forgot to update `config = lib.mkIf (cfg.enabled != null)` to `config = lib.mkIf cfg.enable` in `modules/i18n/input-method/default.nix` (not sure if there are other places also need change)

should be fixed when #6896 is merged

PR #6892, opened 2025-04-23
fcitx5: add upstream options

awwpotato:
### Description
Adds options that are nixos but are missing here

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


awwpotato:
also @Flameopathic please be more familiar with the tool your writing a module for. I have to deal with deprecating an option now because of #6876.

modules/i18n/input-method/fcitx5.nix
```diff
@@ -38,12 +40,74 @@ in
...
+        '';
+      };
+
       classicUiConfig = lib.mkOption {
         type = with lib.types; either path lines;
         default = "";
-        description = ''
-          Configuration to be written to {file}`$XDG_DATA_HOME/fcitx5/conf/classicui.conf`
-        '';
+        description = "Deprecated, use `settings.addons.classicui` instead";
```
Flameopathic, line 110: Should this be marked as `internal` now so that it is no longer shown as an available option? I understand why you are not using `pkgs.mkRenamedOptionModule`, but I am not sure of the precedent when needing to manually handle this.

modules/i18n/input-method/fcitx5.nix
```diff
@@ -83,7 +153,34 @@ in
   };
 
   config = lib.mkIf (im.enable && im.type == "fcitx5") {
-    i18n.inputMethod.package = fcitx5Package;
+    assertions = [
+      {
+        assertion = cfg.classicUiConfig != null;
+        message = "classicUiConfig has been removed, use `settings.addons.classicui` instead";
+      }
+    ];
```
Flameopathic, line 161: Should this be a warning instead of an assertion, as `pkgs.mkRenamedOptionModule` does?

If it were, we would also have to set `i18n.inputMethod.fcitx5.settings.addons.classicui` based on what the user set `i18n.inputMethod.fcitx5.classicUiConfig` to.

Flameopathic (commented):
> also @Flameopathic please be more familiar with the tool your writing a module for. I have to deal with deprecating an option now because of #6876.

I'm sorry. Thank you for cleaning up my mess.

modules/i18n/input-method/fcitx5.nix
khaneliman: Don't even think the option was around long enough to need to deprecate imo

PR #6893, opened 2025-04-23
workflows: only run conflicts and update flake on main repo

awwpotato:
### Description
github keeps sending me emails b/c they are failing on my fork (e.g. https://github.com/awwpotato/home-manager/actions/runs/14610278483/job/40986902179, https://github.com/awwpotato/home-manager/actions/runs/14609635006/job/40985133944)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/.github/workflows/conflicts.yml
+++ b/.github/workflows/conflicts.yml
@@ -5,6 +5,7 @@ on:
 jobs:
   triage:
     runs-on: ubuntu-latest
+    if: github.repository_owner == 'nix-community'
     steps:
       - uses: mschilde/auto-label-merge-conflicts@master
         with:
--- a/.github/workflows/update-flake.yml
+++ b/.github/workflows/update-flake.yml
@@ -8,6 +8,7 @@ on:
 jobs:
   update:
     runs-on: ubuntu-latest
+    if: github.event_name != 'schedule' || github.repository_owner == 'nix-community'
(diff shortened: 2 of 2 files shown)


khaneliman:
Thanks, I meant to tweak this a long time ago but forgot. I would like it though to also include manual runs

.github/workflows/conflicts.yml
```diff
@@ -2,9 +2,12 @@ on:
   push:
     branches:
       - master
+  workflow_dispatch:
+
```
khaneliman, line 6: Sorry, didn't mean adding this functionality. Meant we needed to change it on the `if` condition for the `update-flake`

.github/workflows/update-flake.yml
```diff
@@ -8,6 +8,7 @@ on:
 jobs:
   update:
     runs-on: ubuntu-latest
+    if: github.repository_owner == 'nix-community'
```
khaneliman, line 11:
```suggestion
    if: github.event_name != 'schedule' || github.repository_owner == 'nix-community' 
```

khaneliman (approved):

.github/workflows/conflicts.yml
awwpotato: oops

PR #6905, opened 2025-04-24
treewide: Prevent IFD by default

amarshall:
### Description

See individual commits for details.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,6 +26,11 @@ jobs:
     - run: ./format --ci
     - run: nix-shell --show-trace . -A install
     - run: yes | home-manager -I home-manager=. uninstall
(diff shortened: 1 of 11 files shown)


khaneliman:
I really like the overall idea of preventing IFD. But, definitely would miss the `nmt` reporting in the test results (just as someone constantly parsing them and fixing based on the reporting).

spikespaz:
Can you clarify the term IFD for me? I thought it stood for "import from derivation" and in that case, this is something different, it is not using `import`.

`lib.pathIsDirectory` and `builtins.readDir` where the argument is a package is not always "import from derivation" if I understand correctly.

khaneliman:
Haven't forgotten about this one. Just have been pretty busy and wanted to test a few things with CI before migrating to this approach.

PR #6907, opened 2025-04-25
onedrive: add module

### LINKED ISSUE #6578: Onedrive for Linux module
Hi!

Would it be possible to include onedrive for Linux module into home-manager?

https://github.com/abraunegg/onedrive

Best,
Miro

aguirre-matteo:
### Description

This PR provides a module for configurating the [OneDrive Client for Linux](https://github.com/abraunegg/onedrive). Closes #6578.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/onedrive.nix
```diff
@@ -0,0 +1,63 @@
...
+      '';
+    };
+  };
+
+  config = mkIf cfg.enable {
+    assertions = [
+      (lib.hm.assertions.assertPlatform "programs.onedrive" pkgs lib.platforms.linux)
+    ];
+
+    home.packages = [ cfg.package ];
```
khaneliman, line 57:
```suggestion
    home.packages = mkIf (cfg.package != null) [ cfg.package ];
```

modules/programs/onedrive.nix
```diff
@@ -0,0 +1,63 @@
...
+    mkOption
+    types
+    concatStringsSep
+    mapAttrsToList
+    ;
+
+  cfg = config.programs.onedrive;
+
+  attrToString = name: value: ''${name} = "${value}"'';
+  generateConfig = settings: concatStringsSep "\n" (mapAttrsToList attrToString settings) + "\n";
```
khaneliman, line 21: if we're just generating key value pairs here, we could use that generator instead

aguirre-matteo: I took a look at the generators but didn't found a way to put the quotes around the string.

khaneliman:
```nix
  lib.generators.toKeyValue {
    mkKeyValue = name: value: ''${name} = "${value}"'';
  };
```
Does this work?

PR #6909, opened 2025-04-25
thefuck: Add alias option

Cool-Game-Dev:
Added an alias option supported by bash, zsh, fish, and nushell integration

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@ilaumjd

### DIFF
--- a/modules/programs/thefuck.nix
+++ b/modules/programs/thefuck.nix
@@ -24,6 +24,15 @@ in
     enableNushellIntegration = lib.hm.shell.mkNushellIntegrationOption { inherit config; };
 
     enableZshIntegration = lib.hm.shell.mkZshIntegrationOption { inherit config; };
+
+    alias = lib.mkOption {
+      type = lib.types.str;
+      default = "fuck";
+      description = ''
+        `thefuck` needs an alias to be configured.
+        The default value is `fuck`, but you can use anything else as well.";
+      '';
+    };
   };
 
   config =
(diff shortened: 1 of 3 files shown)


modules/programs/thefuck.nix
```diff
@@ -24,14 +24,23 @@ in
...
 
     enableZshIntegration = lib.hm.shell.mkZshIntegrationOption { inherit config; };
+
+    alias = lib.mkOption {
+      type = lib.types.str;
+      default = "fuck";
+      description = ''
+        `thefuck` needs an alias to be configured.
+        The default value is `fuck`, but you can use anything else as well.";
+      '';
```
khaneliman, line 34:
```suggestion
      description = "Alias used to invoke `thefuck`.";
```

Default will already be documented by the option's default value.

modules/programs/thefuck.nix
```diff
@@ -24,14 +24,23 @@ in
...
+    };
   };
 
   config =
     let
       cfg = config.programs.thefuck;
 
       cliArgs = lib.cli.toGNUCommandLineShell { } {
-        alias = true;
+        alias = cfg.alias;
```
khaneliman, line 43: Doesn't seem like this needs to change?

Cool-Game-Dev: The `--alias` flag sets the alias, so when `--alias FUCK` is added, it changes the alias to FUCK, which is used on bash and zsh integration.

khaneliman: Ah, I see. Default behavior is to just set to `fuck` but actually passing a value lets you change it.

PR #6910, opened 2025-04-26
rmpc: add module

### LINKED ISSUE #6611: Program Request: rmpc
### Description

Hello,

I would like to request the addition of the  [rmpc](https://github.com/mierak/rmpc) to Home-Manager, along with support for configuring its settings. Since rmpc is already available in the Nix package repository, integrating it into Home-Manager with configurable options would improve its usability.

Detailed information on rmpc's configuration options can be found in the official documentation: [rmpc Configuration](https://mierak.github.io/rmpc/next/configuration/).

Thank you for your time and consideration!

aguirre-matteo:
### Description

This PR adds a module for configuring [RMPC](https://github.com/mierak/rmpc), a terminal based Media Player.
Since the config format of RMPC is the Rusty Object Notation (RON), and Nixpkgs doesn't have a generator for that,
I added the `extraConfig` options instead of `settings`. Closes #6611.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/rmpc.nix
+++ b/modules/programs/rmpc.nix
@@ -1,0 +1,59 @@
+{
+  lib,
+  pkgs,
+  config,
+  ...
+}:
+let
+  inherit (lib)
+    mkIf
+    mkEnableOption
+    mkPackageOption
+    mkOption
+    types
+    ;
+
+  cfg = config.programs.rmpc;
+in
+{
+  meta.maintainers = with lib.hm.maintainers; [ aguirre-matteo ];
+
+  options.programs.rmpc = {
+    enable = mkEnableOption "rmpc";
+    package = mkPackageOption pkgs "rmpc" { nullable = true; };
+    extraConfig = mkOption {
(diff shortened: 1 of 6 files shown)


modules/programs/rmpc.nix
```diff
@@ -0,0 +1,59 @@
...
+
+  cfg = config.programs.rmpc;
+in
+{
+  meta.maintainers = with lib.hm.maintainers; [ aguirre-matteo ];
+
+  options.programs.rmpc = {
+    enable = mkEnableOption "rmpc";
+    package = mkPackageOption pkgs "rmpc" { nullable = true; };
+    extraConfig = mkOption {
```
khaneliman, line 24:
```suggestion
    config = mkOption {
```

If we're using this as the actual configuration, `extra` doesn't make sense.

aguirre-matteo: Ok

PR #6921, opened 2025-04-28
tag-maintainers.yml: init 

khaneliman:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```


khaneliman:
@rycee I think our github token needs 
>"Pull requests" repository permissions (write)

.github/workflows/tag-maintainers.yml
```diff
@@ -0,0 +1,141 @@
...
+            for PATH_FORMAT in \
+              "$FILE" \
+              "/Users/khaneliman/github/home-manager/$FILE" \
+              "/$PWD/$FILE" \
+              "$PWD/$FILE"
+            do
+              echo "Trying path format: $PATH_FORMAT"
+
+              # Get maintainers for this file using the current path format
+              MAINTAINERS=$(jq -r ".[\"$PATH_FORMAT\"] | if . != null then .[].github else empty end" <<< "$MODULE_MAINTAINERS" 2>/dev/null || true)
```
rycee, line 64: Wouldn't it be possible to make jq read the file directly?

```suggestion
              MAINTAINERS=$(jq -r ".[\"$PATH_FORMAT\"] | if . != null then .[].github else empty end" ./result 2>/dev/null || true)
```

rycee:
Looks nice! About the permissions I'm not certain. From what I could see in the [GitHub documentation](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) it should be possible to add a `permissions` field in the yaml file. Something like
``` yaml
permissions:
  contents: read
  pull-requests: write
```
perhaps. This seems to be used by the eval flow in Nixpkgs: https://github.com/NixOS/nixpkgs/blob/d66115b18ccf2fbb03da4f2ea8a41499eb8d3136/.github/workflows/eval.yml#L241-L243

khaneliman:
> Looks nice! About the permissions I'm not certain. From what I could see in the [GitHub documentation](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) it should be possible to add a `permissions` field in the yaml file. Something like
> 
> ```yaml
> permissions:
>   contents: read
>   pull-requests: write
> ```
> 
> perhaps. This seems to be used by the eval flow in Nixpkgs: https://github.com/NixOS/nixpkgs/blob/d66115b18ccf2fbb03da4f2ea8a41499eb8d3136/.github/workflows/eval.yml#L241-L243

I'm wondering if this is related to something I've seen previously, github wont let us request reviews from those who aren't collaborators. In nixpkgs, we add everyone to the Nixos organization. We might not be able to do this here, then.... 

https://github.com/NixOS/nixpkgs/blob/d66115b18ccf2fbb03da4f2ea8a41499eb8d3136/ci/request-reviews/request-reviewers.sh#L61-L69

Tested with this block in there and that seems to be the case

khaneliman:
Using PR target I need to check this in to really test it. But from a test in my repo with https://github.com/khaneliman/home-manager/actions/runs/16031756101/job/45233835846?pr=567 it appeared to make it down to requesting a review but failed because it was looking against my fork and not this repo. 

EDIT: Still have some cleanup I want to do

khaneliman:
https://github.com/khaneliman/home-manager/actions/runs/16033202876/job/45238595516?pr=568 looks even better. Was able to test simplifying some logic and handling people who have already been reviewing so they dont get re-requested.

khaneliman:
Would like to test against someone invited to my fork as a collaborate to make sure that flow works to avoid a bunch of fixes PRs.

khaneliman:
Sweet, got it working https://github.com/khaneliman/home-manager/pull/568

PR #6936, opened 2025-04-30
carapace: fix nushell PATH env var

brynedwards:
### Description

`carapace _carapace nushell` adds `/homeless-shelter/.config/carapace/bin` to PATH, this changes it to ` $HOME/.config/carapace/bin`. Other shells don't have this problem as they run the relevant command and source on startup whereas Nix writes the command output to a file as part of the build.

### Checklist

- [ ] Change is backwards compatible. (Technically this would break someone's setup if they actually decided to use `/homeless-shelter/.config/carapace/bin` but that seems like a very silly thing to do.)

- [ ] Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef). (I couldn't figure out how to refer to the generated `carapace-nushell-config.nu` file from the test file.)

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@Weathercold
@bobvanderlinden

### DIFF
--- a/modules/programs/carapace.nix
+++ b/modules/programs/carapace.nix
@@ -49,7 +49,7 @@ in
         extraConfig = ''
           source ${
             pkgs.runCommand "carapace-nushell-config.nu" { } ''
-              ${bin} _carapace nushell >> "$out"
+              ${bin} _carapace nushell | sed 's|"/homeless-shelter|$"($env.HOME)|g' >> "$out"
             ''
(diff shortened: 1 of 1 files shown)


Weathercold (approved):
LGTM

Issue #4313, opened 2023-08-05
Make `home.sessionVariables` work with nushell


Kreyren:
### Description

Currently i have to set the following:

```nix
  home.sessionVariables = {
    SHELL = "nu";
    TERM = "alacritty";
    EDITOR = "vim";
  };

  programs.nushell.extraEnv = ''
    # FIXME(Krey): This should be handled in `home.sessionVariables`
    $env.SHELL = "nu";
    $env.TERM = "alacritty"
    $env.EDITOR = "vim";

    # Include local binaries
    $env.PATH = ($env.PATH | split row (char esep) | append ($env.HOME | path join '.local/bin'));
  ''; 
```

Because `home.sessionVariables` doesn't export the variables for `nushell` -> Make it work with nushell too?

Expected workflow:

```diff
  home.sessionVariables = {
    SHELL = "nu";
    TERM = "alacritty";
    EDITOR = "vim";
  };

  programs.nushell.extraEnv = ''
-    # FIXME(Krey): This should be handled in `home.sessionVariables`
-    $env.SHELL = "nu";
-    $env.TERM = "alacritty"
-    $env.EDITOR = "vim";
-
    # Include local binaries
    $env.PATH = ($env.PATH | split row (char esep) | append ($env.HOME | path join '.local/bin'));
  ''; 
```

Suyashtnt:
It's been a few months now. Are there any plans for this?

bluskript:
quick fix that I haven't had time to cook into a PR / probably has compat problems:
```nix
    environmentVariables = builtins.mapAttrs (name: value: "\"${builtins.toString value}\"") config.home.sessionVariables;
```

heisfer:
Maybe we could make use of https://github.com/tesujimath/nu_plugin_bash_env this plugin?

fehnomenal:
This is still relevant.

After https://github.com/NixOS/nixpkgs/pull/343036 is merged, I can create a PR because I have this setup working.

hey2022:
> quick fix that I haven't had time to cook into a PR / probably has compat problems:
> 
> ```nix
>     environmentVariables = builtins.mapAttrs (name: value: "\"${builtins.toString value}\"") config.home.sessionVariables;
> ```

Buidling upon @bluskript quick fix, this changes references to environment variables such as $PATH to $env.PATH

```nix
environmentVariables = builtins.mapAttrs (
[...]

PR #6941, opened 2025-04-30
nushell: implement build-time hm-session-vars.sh translation to NUON

jaredmontoya:
### Description
- Automatically translates hm-session-vars.sh at build time into a NUON (NUshell Object Notation) and loads it using `load-env` built in command.

One caveat that I noticed is that my gpg-agent does not set the correct `SSH_AUTH_SOCK` even though `gpg-agent.enableSshSupport` is true because this variable is already set in the `hm-session-vars.sh` to an incorrect value due to being evaluated in the nix sandbox.
That's why in this pull request SSH_AUTH_SOCK is being filtered out before passing the environment NUON to `load-env`

This kind of issue can be mitigated at the user side by adding `hide-env -i INCORRECTLY_SET_VAR` at the beginning of their `envFile.text` but ideally they should report the variable to be added to the `reject` filter in the nushell module like I did with SSH_AUTH_SOCK.

- Adds support for nushell to the `dircolors` module (uses the same mechanism under the hood)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


jaredmontoya:
@khaneliman

JoaquinTrinanes (commented):
(I won't have access to my laptop until a couple of days from now so this is written without actually testing the changes)

What is the motivation for this change? If it's to add dircolors integration perhaps a less intrusive approach could look like (rough untested code):

```nix
use ${pkgs.runCommand "dircolors.nu" ''
    eval $(dircolors ...)
    echo "export-env { $env.LS_COLORS = \"$LS_COLORS\" }" > $out
''}
```

I'm worried that sourcing hm-session-vars is more trouble than it's worth. SSH_AUTH_SOCK already breaks, but what about HOME, or USER? This file will always assume that it's running in the host machine, not in the sandbox. Any unrelated integration that ends up in home-manager would be able to break user nu configurations (and because of that I also feel like this is a breaking change)

jaredmontoya:
Motivation: #4313 #4556

I don't see USER or HOME variables defined in my `/etc/profiles/per-user/username/etc/profile.d/hm-session-vars.sh` so I don't know how they are related.

It has this SSH_AUTH_SOCK definition though:
```bash
if [ -z "$SSH_AUTH_SOCK" ]; then
  export SSH_AUTH_SOCK="$(/nix/store/hrgmxpl5hfikkv2201mml82c7xyd6ld3-gnupg-2.4.7/bin/gpgconf --list-dirs agent-ssh-socket)"
fi
```
it comes from `gpg-agent.nix` in home-manager.
```nix
        home.sessionVariablesExtra = optionalString cfg.enableSshSupport ''
          if [ -z "$SSH_AUTH_SOCK" ]; then
            export SSH_AUTH_SOCK="$(${gpgPkg}/bin/gpgconf --list-dirs agent-ssh-socket)"
          fi
        '';
```

`gpg-agent.nix` also provides this nushell integration separately:
```nix
gpgNushellInitStr =
    ''
      $env.GPG_TTY = (tty)
    ''
    + optionalString cfg.enableSshSupport ''
      ${gpgPkg}/bin/gpg-connect-agent updatestartuptty /bye | ignore

      $env.SSH_AUTH_SOCK = ($env.SSH_AUTH_SOCK? | default (${gpgPkg}/bin/gpgconf --list-dirs agent-ssh-socket))
    '';

programs.nushell.extraEnv = mkIf cfg.enableNushellIntegration gpgNushellInitStr;
```

This nushell integration normally sets the actual correct value for nushell.
After my change this variable is already set incorrectly in `hm-session-vars.sh` so the nushell integration does nothing.

Therefore it makes sense to change `gpg-agent.nix` so that if the nushell integration is enabled, it does not set SSH_AUTH_SOCK in `hm-session-vars.sh` and therefore filtering out SSH_AUTH_SOCK is not necessary anymore.

This means that as long as all home-manager integrations affected by this are changed accordingly this won't be a problem.

JoaquinTrinanes:
Re: #4313, users can now do `programs.nushell.environmentVariables = config.home.sessionVariables;` if they wish to.

#4556 tries to achieve the same thing but was closed due to nushell not being compatible with `pam_env.conf`.

> I don't see USER or HOME variables defined in my /etc/profiles/per-user/username/etc/profile.d/hm-session-vars.sh so I don't know how they are related.

Exactly, they are not set there. Which means they will be different depending on the user/host running the file. So if you have (for example)

```sh
export HOME=/home/$USER # this is more unlikely but just an example
export XDG_CONFIG_HOME=$HOME/.config
```

With the new changes, this would result in the first line setting `HOME` to `/home/nixbld` or something similar. Without that first line, `XDG_CONFIG_HOME` would be set to `/homeless-shelter/.config`. This just can't be done inside the sandbox.

On top of that, there's [a warning in nushell's docs about setting it as the default shell](https://www.nushell.sh/book/default_shell.html#setting-nu-as-login-shell-linux-bsd-macos). IMO, the less error-prone way of setting it up is to invoke nu from the .profile or .bashrc or whatever POSIX shell config file is in use. Then nushell will inherit the environment already set by that shell.

jaredmontoya:
@JoaquinTrinanes 
Re: Re: #4313, why should they need to do that? 
(Aren't options like home.sessionVariables supposed to work for all supported shells out of the box? That's the goal in my opinion.)

> Exactly, they are not set there. Which means they will be different depending on the user/host running the file. So if you have (for example)

I still can't understand where exactly the problem with that is.
As long as nobody purposefully sets
`home.sessionVariables.HOME = ''/home/$USER''` (which is asking for trouble)
and it ends up becoming
`HOME=/home/nixbld`
everything will be fine.

If XDG_CONFIG_HOME was set like you described(using bash variables) it would probably be like asking for trouble too, but it is in fact set in `hm-session-vars.sh` and it's absolutely fine for me because the path is determined not by bash at package build time but by home manager at evaluation time and uses whatever `home.homeDirectory` is.

we don't need to worry about multiple users using the same `hm-session-vars.sh` because they don't, home manager builds a slightly different version that fits each user.

I used this kind of configuration for a very long time with nushell as a user login shell.

jaredmontoya:
But what is true is that it's not a perfect solution, A perfect solution would be to forbid the use of plain text conditional code like this:
```nix
        home.sessionVariablesExtra =
          optionalString (cfg.enableSshSupport && !cfg.enableNushellIntegration)
            ''
              if [ -z "$SSH_AUTH_SOCK" ]; then
                export SSH_AUTH_SOCK="$(${gpgPkg}/bin/gpgconf --list-dirs agent-ssh-socket)"
              fi
            '';
```

And create a Nix interface that when variables are added to they will be automatically added to the necessary config file, be it `hm-session-vars.sh` or `hm-session-vars.nu` with correct logic that ensures the variable is sourced only once and not overwritten and this logic is written in the correct language for the config file.

basically this means supporting nushell as much as bash.
Generating a so called `hm-session-vars.nu` that produces the same environment as `hm-session-vars.sh`
But I don't know how realistic that is.

jaredmontoya:
I reverted to filtering out SSH_AUTH_SOCK after it was added to `hm-session-vars.sh` instead of not adding it to `hm-session-vars.sh` because not adding it means it's not possible to use more than one shell at once with ssh gpg-agent if nushell integration is enabled.

Not adding it in the first place would require changing other shell integrations to not rely on `hm-session-vars.sh`

jaredmontoya:
Extracted dircolors integration into a separate PR: #6963

jaredmontoya:
> Since it seems relevant, I'm going to link a comment I made about removing `sessionVariablesExtra`: [#2659 (comment)](https://github.com/nix-community/home-manager/issues/2659#issuecomment-1976950376). It's the root cause of numerous issues.

I also think removing `sessionVariablesExtra` is a good idea.

iniw:
Anything stopping this from getting merged?

dtgagnon:
Discussion seems to have stalled.

What, if anything, remains to be addressed before merging this?

jaredmontoya:
In the meantime, if anyone wants to use this and the dircolors integration without them being merged into home-manager then just do something like this in your config:

```nushell
{
  lib,
  pkgs,
  ...
}:

let
  hmSessionVarsNu = pkgs.runCommand "hm-session-vars.nu" { } ''
    echo "if (\$env.__HM_SESS_VARS_SOURCED? | is-empty) {''$(${lib.getExe config.programs.nushell.package} -c "
      use ${pkgs.nu_scripts}/share/nu_scripts/modules/capture-foreign-env
      open ${config.home.sessionVariablesPackage}/etc/profile.d/hm-session-vars.sh | capture-foreign-env | to nuon
    ") | load-env}" >> "$out"
  '';

  dircolorsIntegration = pkgs.runCommand "dircolors.nu" { } ''
    eval "$(${lib.getExe' pkgs.coreutils "dircolors"} -b)"
    echo "export-env { \$env.LS_COLORS = \"$LS_COLORS\" }" >> $out
  '';
in
{
  programs = {
    nushell = {
      enable = true;

      envFile.text = # nu
        ''
          source ${hmSessionVarsNu}
          source ${dircolorsIntegration}
          hide-env -i SSH_AUTH_SOCK
        '';
    };
  };
}
```

iniw:
With that snippet I was finally able to directly launch nushell from ghostty without having to go through zsh/bash first. It took some surgery but here's what I came up with:

```nix
{
  config,
  osConfig,
  pkgs,
  lib,
  ...
}:
{
  programs.nushell = {
    enable = true;
    envFile.text =
      let
        environment = pkgs.runCommand "setup-env-vars.nu" { } ''
            echo "if ((\$env.__HM_SESS_VARS_SOURCED? | is-empty) or ((\$env.__NIXOS_SET_ENVIRONMENT_DONE? | is-empty) and ((\$env.__NIX_DARWIN_SET_ENVIRONMENT_DONE? | is-empty)))) {''$(${lib.getExe pkgs.nushell} -c "
            use ${pkgs.nu_scripts}/share/nu_scripts/modules/capture-foreign-env
            with-env {
              HOME: ${config.home.path}
              USER: ${config.home.username}
            } {
              open ${osConfig.system.build.setEnvironment} ${config.home.sessionVariablesPackage}/etc/profile.d/hm-session-vars.sh | str join "\n" | capture-foreign-env | to nuon
            }
          ") | load-env}" >> "$out"
        '';
      in
      # nu
      ''
        source ${environment}
        hide-env -i SSH_AUTH_SOCK
      '';
  };
  programs.ghostty.settings.command =
    with config.home.sessionVariables;
    "env XDG_CONFIG_HOME=${XDG_CONFIG_HOME} XDG_DATA_HOME=${XDG_DATA_HOME} ${lib.getExe pkgs.nushell}";
};
```

It's pretty hacky but it works.

The `with-env` scope on the foreing-env capture is required because `setEnvironment` uses `$HOME` and `$USER`. Not sure how to avoid that.

The `env` hack on the last line is required to make nushell look for it's config in the correct xdg directories instead of `~/Library/Application Support`

JoaquinTrinanes:
I have some issues with the current approach:

- Building the variables file in the sandbox breaks, due to the sandbox environment variables being different. This is reinforced by the fact that some variables have to be manually rejected.
  So far we have `SSH_AUTH_SOCK`, but I don't think we can assume users won't have anything else breaking in their configs. Normal home-manager usage will cause breakage:

  ```nix
    # modules/services/ssh-tpm-agent.nix
    home.sessionVariables = {
      # Override ssh-agent's $SSH_AUTH_SOCK definition since ssh-tpm-agent is a
      # proxy to it.
      SSH_AUTH_SOCK = lib.mkOverride 90 "$XDG_RUNTIME_DIR/ssh-tpm-agent.sock";
      SSH_TPM_AUTH_SOCK = "$XDG_RUNTIME_DIR/ssh-tpm-agent.sock";
    };
  ```

  ```nix
    # modules/misc/xdg-system-dirs.nix
    (lib.mkIf (cfg.config != [ ]) {
      home.sessionVariables.XDG_CONFIG_DIRS = "${configDirs}\${XDG_CONFIG_DIRS:+:$XDG_CONFIG_DIRS}";

      systemd.user.sessionVariables.XDG_CONFIG_DIRS = "${configDirs}\${XDG_CONFIG_DIRS:+:$XDG_CONFIG_DIRS}";
    })
  ```

  ```nix
    # modules/programs/tmux.nix
    (lib.mkIf cfg.secureSocket {
    home.sessionVariables = {
        TMUX_TMPDIR = ''''${XDG_RUNTIME_DIR:-"/run/user/$(id -u)"}'';
    };
    })
  ```

  In a previous comment I wrote an example setting `HOME=/home/$USER` and `XDG_CONFIG_HOME=$HOME/.config`, which _shouldn't_ be done. But it's pretty common to use code like `config_dir=${XDG_CONFIG_HOME:-$HOME/.config}` in shell scripts and that will always have the wrong value (and there's an infinite amount of variables that can be referenced that are not set correctly in the sandbox, especially if they are not managed by nix).

  <details><summary>Optional read: another specific, niche example that I found in my config</summary>
    

  > I have this in my bash config:
  >     
  >     ```nix
  >       programs.bash.initExtra = ''
  >         if [ -f "$HOME/.bashrc.local" ]; then
  >           source "$HOME/.bashrc.local"
  >         fi
  >       '';
  >     ```
  >     
  >     It's handy because it lets me modify my shell environment without rebuilding my config,
  >     making testing things easier. Note that anything that could modify the environment outside
  >     of the sandbox won't do it inside of it, so the values will be different at runtime!
  >     (I think it's not that big of a deal because we're talking about configuring a different shell)

</details>


- home-manager's own `sessionVariables` option documents that POSIX syntax is allowed, but we can't guarantee any referenced variable will be valid in the sandbox:

  ```
    The values may refer to other environment variables using
    POSIX.2 style variable references. For example, a variable
    {var}`parameter` may be referenced as
    `$parameter` or `''${parameter}`. A
    default value `foo` may be given as per
    `''${parameter:-foo}` and, similarly, an alternate
    value `bar` can be given as per
    `''${parameter:+bar}`.
  ```

However, I agree that `sessionVariables` should be shell-agnostic. I agree with @jaredmontoya in that `sessionVariablesExtra` should go away. That would let us use `sessionVariables` in whatever shell is being configured without worrying about the syntax (assuming we officially drop support for the POSIX syntax, which would be a big-ish breaking change). Instead of `sessionVariablesExtra` we could enforce setting this conditional logic per-shell. So instead of

```nix
home.sessionVariablesExtra = ''
    if [ -z "$SSH_AUTH_SOCK" ]; then
      export SSH_AUTH_SOCK=$XDG_RUNTIME_DIR/${cfg.socket}
    fi
'';
```

We could do:

```nix
programs.bash.profileExtra = ''...'';
programs.zsh.profileExtra = ''...'';
programs.nushell.extraConfig = ''
if "SSH_AUTH_SOCK" not-in $env {
    $env.SSH_AUTH_SOCK = (...)
}
'';
```

(we could rename `sessionVariablesExtra` to `posixShellExtra` or similar (awful name, we can bikeshed later), if we want to avoid setting the same value multiple times for many shells).

This has the disadvantage of having to manually support every shell, but it's better than not supporting it at all I guess. And we already have to do it with all the `enable<shell>Integration` options.

I personally like the approach of launching nu from the login shell so all variables are already set for its simplicity but it's not as "native" as actually supporting things.

Note that getting rid of `sessionVariablesExtra` doesn't fix the issues I have with using `capture-foreign-env` in the sandbox due to potantially having broken values, but unifying the config so `capture-foreign-env` isn't even needed should be the endgoal IMO :)

iniw:
> I personally like the approach of launching nu from the login shell so all variables are already set for its simplicity but it's not as "native" as actually supporting things.

I used this approach for a long time and it's "fine". The only real issue I have with it is that it makes shell startup times much much longer, which is particularly annoying when using splits/tabs in terminal emulators that support them (which I happen to use all the time). Launching nushell through `$SHELL` -> `exec nu` easily takes 2-3x longer than just launching it directly on my setup, which is extremely noticeable.

IanHollow:
@jaredmontoya I thought it might be useful in some way to share the way I convert env vars for the PATH to my nushell config. Hopefully this is some what useful. https://github.com/IanHollow/nix-conf/blob/main/configs/home/shells/nushell/extra-config-after.nix

This what i do for env vars in general https://github.com/IanHollow/nix-conf/blob/main/configs/home/shells/nushell/env.nix

PR #6946, opened 2025-04-30
clipcat: add module

### LINKED ISSUE #6297: home-manager module for clipcat
### Description

Would be nice, if there would be a home-manager-module to just enable [clipcat]

[clipcat]: https://github.com/xrelkd/clipcat

aguirre-matteo:
### Description

This PR adds a module for configuring Clipcat (clipcatctl, clipcatd, clipcat-menu) and its respective Systemd Unit. Closes #6297.

NOTE: I didn't added `{ nullable = true; }` to the package option since the package is needed in the Systemd Unit (`"${cfg.package}/bin/clipcatd"`). Feel free to show me an alternative.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -332,6 +332,7 @@ let
       ./services/cachix-agent.nix
       ./services/caffeine.nix
       ./services/cbatticon.nix
+      ./services/clipcat.nix
       ./services/cliphist.nix
       ./services/clipman.nix
       ./services/clipmenu.nix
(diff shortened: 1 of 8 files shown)


khaneliman:
>NOTE: I didn't added { nullable = true; } to the package option since the package is needed in the Systemd Unit ("${cfg.package}/bin/clipcatd"). Feel free to show me an alternative.

Could make the package nullable if you added an assertion when the user has the package set to null and the systemd service enabled, but that's up to you.

PR #6957, opened 2025-05-02
gimp: add module

### LINKED ISSUE #6178: Program request: gimp
### Description

GIMP supports plugins, and there's a separate package, `gimp-with-plugins`, and nixpkgs also has many gimp plugin packages like `gimp-plugin-gap` and `gimp-plugin-fourier`.

aguirre-matteo:
### Description

This PR adds a module for enabling Gimp and its plugins. It seems Gimp uses a large amount of config files at ~/.config/GIMP/, some of them are updated every time Gimp starts/closes, some of them not, but they are a lot and a have the package version as a prefix (see `man gimp`, FILES section), so it will require constant maintenance and could easily break. Closes #6178.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -117,6 +117,7 @@ let
       ./programs/gh-dash.nix
       ./programs/ghostty.nix
       ./programs/git-cliff.nix
+      ./programs/gimp.nix
       ./programs/git-credential-oauth.nix
       ./programs/git-worktree-switcher.nix
       ./programs/git.nix
--- a/modules/programs/gimp.nix
+++ b/modules/programs/gimp.nix
@@ -14,13 +14,22 @@ let
     ;
 
   cfg = config.programs.gimp;
+  defaultPackage = if cfg.withPlugins then "gimp-with-plugins" else "gimp";
 in
 {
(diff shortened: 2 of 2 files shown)


khaneliman:
Not really a fan of modules that only purpose are to add a package to `home.packages`.

aguirre-matteo:
Should I directly close the PR? I don't think GIMP will change its config format ever

khaneliman:
> Should I directly close the PR? I don't think GIMP will change its config structure ever

Sounded like @rycee liked the idea of the module in the original issue. Maybe he has a different opinion.

rycee:
This PR doesn't actually seem to make use of `gimp-with-plugins`? I.e., it doesn't override `plugins`. As-is I don't think the module adds anything, might as well simply add the packages to `home.packages`.

philocalyst:
Interested in doing something like this, I don't know if GIMP is in the habit of changing their configuration locations, although they certainly like adding them! So maintenance might just be adding new ones, or signaling deprecations. Is that something that's okay with the team?

Past that, it just simply might be a large module.

@rycee @khaneliman

PR #6959, opened 2025-05-03
thunderbird: fix accounts being removed

### LINKED ISSUE #6800: bug: thunderbird's all accounts disappear except the first one.

### Issue description

I suspect `user_pref("mail.accountmanager.accounts", "account1");` line in user.js.

Added by:
https://github.com/nix-community/home-manager/blob/f1ffd097e717a8d1b441577b8d23f9d2c96e0657/modules/programs/thunderbird.nix#L701-L705

### Maintainer CC

@d-dervishi @jKarlson

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.14.0-zen1, NixOS, 25.05 (Warbler), 25.05.20250408.c8cd814`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Lix, like Nix) 2.91.1
System type: x86_64-linux
Features: gc, signed-caches
User configuration files: /home/honnip/.config/nix/nix.conf:/etc/xdg/nix/nix.conf:/home/honnip/.nix-profile/etc/xdg/nix/nix.conf:/nix/profile/etc/xdg/nix/nix.conf:/home/honnip/.local/state/nix/profile/etc/xdg/nix/nix.conf:/etc/profiles/per-user/honnip/etc/xdg/nix/nix.conf:/nix/var/nix/profiles/default/etc/xdg/nix/nix.conf:/run/current-system/sw/etc/xdg/nix/nix.conf:/nix/store/433ipd9axfqx86mm6s9wn6ix4gpd91si-gnome-settings-daemon-47.2/etc/xdg/nix/nix.conf
Store directory: /nix/store
State directory: /nix/var/nix
Data directory: /nix/store/b9b8mvcrbq5dy1xb4n5y59hc8axj0yzf-lix-2.91.1/share`
 - channels(root): `""`
 - nixpkgs: `/nix/store/m7y9g8z15va6sd0h12zc3iw0acxn7463-source`
```

MithicSpirit:
### Description

Even if no accounts were configured, it was defaulting to just account1, which meant all other accounts were deleted every time thunderbird was restarted. This fixes that by defaulting to the empty list (so it remains unset in the user.js).

Introduced #6310 (probably)

Fixes #6800

Btw, I'm not even sure whether `account1` always corresponds to the local folders account; on my machine, it instead corresponds to the first email account I added. I feel like trying to handle thunderbird account management through home-manager is inherently fragile.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@d-dervishi @jKarlson @nfelber

### DIFF
--- a/modules/programs/thunderbird.nix
+++ b/modules/programs/thunderbird.nix
@@ -704,7 +704,9 @@ in
                     "account1"
                   ];
                 in
-                accountsOrderIds ++ (lib.lists.subtractLists accountsOrderIds enabledAccountsIds);
+                lib.optionals (accounts != [ ]) (
+                  accountsOrderIds ++ (lib.lists.subtractLists accountsOrderIds enabledAccountsIds)
+                );
             in
             {
               text = mkUserJs (builtins.foldl' (a: b: a // b) { } (


MithicSpirit:
There should probably be a testcase for this (and some more manual testing of this change would also be helpful), but I don't have the time to do so atm, and this feels like quite an important issue to fix.

khaneliman:
Change makes sense to me, we shouldn't be automatically injecting configuration when the user hasn't chosen to use that customization.

PR #6970, opened 2025-05-04
ghostty: fix config syntax file location on darwin

### LINKED ISSUE #6961: bug: `programs.ghostty.installBatSyntax` option is broken on macOS

### Issue description

When both Ghostty and Bat enabled I'm getting this error during system rebuild:

```
[bat error]: error finding all the files in a directory: IO error for operation on ~/.config/bat/syntaxes/ghostty.sublime-syntax: No such file or directory (os error 2)
```

The error occurs, because `programs.ghostty.installBatSyntax` option (which is enabled by default) tries to link this file: https://github.com/nix-community/home-manager/blob/50eee705bbdbac942074a8c120e8194185633675/modules/programs/ghostty.nix#L142 which does not exist in macOS version of Ghostty, yet the link is created without any error or warning. On macOS the file can be found at `${pkgs.ghostty}/Applications/Ghostty.app/Contents/Resources/bat/syntaxes/ghostty.sublime-syntax`

My current workaround is to re-implement the option myself. I have created a module with following code:

```nix
{
  lib,
  pkgs,
  config,
  ...
}:
lib.mkIf pkgs.stdenv.isDarwin {
  programs = {
    ghostty.installBatSyntax = false;
    bat = {
      config.map-syntax = [ "${config.xdg.configHome}/ghostty/config:Ghostty Config" ];

      syntaxes.ghostty = {
        src = config.programs.ghostty.package;
        file = "Applications/Ghostty.app/Contents/Resources/bat/syntaxes/ghostty.sublime-syntax";
      };
    };
  };
}
```

On macOS it disables the `installBatSyntax` option and links correct file.

I could've send a PR with the fix, but since there's [no official support for Ghostty on macOS in nixpkgs](https://github.com/NixOS/nixpkgs/issues/388984), I wasn't sure if this is the right decision, so I open the issue instead.

Also, the problem affects both `master` and `release-24.11` branches.

I'm using `nixos-24.11` branch of `nixpkgs` and `release-24.11` branch of `home-manager` on my system + Ghostty from `nur.repos.DimitarNestorov.ghostty` added to `nixpkgs` via overlay.

Possible solutions: Either implement fix for macOS, or disable this option by default, and add warnings about macOS support for Ghostty somewhere (in the docs maybe?), or recommendations on how it can be installed from NUR or other source, if any.

### Maintainer CC

_No response_

### System information

```markdown
- system: `"aarch64-darwin"`
- host os: `Darwin 24.4.0, macOS 15.4.1`
- multi-user?: `yes`
- sandbox: `no`
- version: `nix-env (Nix) 2.24.14`
- channels(root): `"nixpkgs"`
- nixpkgs: `/nix/store/9s6w9w0qkgrn6g9hxj6q3isbwbhnibxm-source`
```

octet-stream:
### Description

Home Manager creates broken link to the Ghostty config syntax highlighting definition file, because it has different location on Darwin. This PR updates path to the config for Darwin users.

Fixes #6961

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@HeitorAugustoLN
@khaneliman

### DIFF
--- a/modules/programs/ghostty.nix
+++ b/modules/programs/ghostty.nix
@@ -186,7 +186,7 @@ in
           syntaxes.ghostty = {
             src = cfg.package;
             file =
-              if pkgs.stdenv.isDarwin then
+              if pkgs.stdenv.hostPlatform.isDarwin then
                 "Applications/Ghostty.app/Contents/Resources/bat/syntaxes/ghostty.sublime-syntax"
               else
                 "share/bat/syntaxes/ghostty.sublime-syntax";
--- a/modules/programs/ghostty.nix
+++ b/modules/programs/ghostty.nix
@@ -185,7 +185,11 @@ in
         programs.bat = lib.mkIf (cfg.package != null) {
           syntaxes.ghostty = {
             src = cfg.package;
-            file = "share/bat/syntaxes/ghostty.sublime-syntax";
+            file =
+              if pkgs.stdenv.isDarwin then
+                "Applications/Ghostty.app/Contents/Resources/bat/syntaxes/ghostty.sublime-syntax"
+              else
+                "share/bat/syntaxes/ghostty.sublime-syntax";
           };
           config.map-syntax = [ "${config.xdg.configHome}/ghostty/config:Ghostty Config" ];
         };


modules/programs/ghostty.nix
```diff
@@ -185,7 +185,11 @@ in
         programs.bat = lib.mkIf (cfg.package != null) {
           syntaxes.ghostty = {
             src = cfg.package;
-            file = "share/bat/syntaxes/ghostty.sublime-syntax";
+            file =
+              if pkgs.stdenv.isDarwin then
```
khaneliman, line 189:
```suggestion
              if pkgs.stdenv.hostPlatform.isDarwin then
```

khaneliman:
Thanks, I was wondering why I couldn't recreate the error and realized I disabled installing syntaxes on darwin because of the error :P Path LGTM though from looking at the ghostty .App folder locally.

PR #6978, opened 2025-05-05
git: configure patdiff through [patdiff] git section

leana8959:
### Description

This adds the ability to enable the `patdiff` program as git differ.
Website of the project: https://opensource.janestreet.com/patdiff/

I have tested it via a test and not building it myself, since I'm still on 24.11 and it doesn't evaluate.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@rycee

### DIFF
--- a/tests/darwinScrublist.nix
+++ b/tests/darwinScrublist.nix
@@ -45,6 +45,7 @@ let
     "gh-dash"
     "ghostty"
     "git"
+    "patdiff"
     "gitMinimal"
     "git-cliff"
     "git-credential-oauth"
--- a/modules/programs/git.nix
+++ b/modules/programs/git.nix
@@ -496,6 +496,17 @@ in
           '';
         };
       };
+
+      patdiff = {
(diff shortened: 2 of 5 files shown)


leana8959:
Hmm it seems like the `@@` templating syntax is failing, could someone guide me a bit? I can't figure out why it fails in the way it did and only for macOS.

Thanks :)

tests/darwinScrublist.nix
```diff
@@ -45,6 +45,7 @@ let
     "gh-dash"
     "ghostty"
     "git"
+    "git-patdiff"
```
khaneliman, line 48:
```suggestion
    "patdiff"
```

This is the name of the package from nixpkgs

leana8959: Ah ha I see, thanks :)

khaneliman:
Sorry, need to rebase, we had a CI failure that needed to be fixed.

PR #6981, opened 2025-05-05
nixos: optionally use user service for activation (updates #2548)

zackw:
This is an updated version of PR #2548.  I am filing it like this because I was not able to figure out how to file it any other way; it is not my intention to stage a hostile takeover of @pasqui23's work, only to address one (fairly serious) problem and to make the PR mergeable as quickly as possible.  I require this feature to be able to use home-manager at all and I'd really like to see it squeak under the bar for 25.05.

### Description

Add an option, `home-manager.useUserService`, to use systemd user services for per-user activation, rather than activating all users' environments on boot using systemd *system* services.  This option enables use of home-manager in configurations where users' home directories are not available until they log in (for example, when using pam_mount).

The documentation for the new option is:

> Activate each user's environment only when they log in, using a systemd user service, instead of activating all configured users' environments proactively during boot-up.
>
> This mode is necessary if users' home directories are not available until they log in (for example, when using pam_mount).
>
> Other usage scenarios are still experimental.  It may speed up boot when there are many users; this has not yet been confirmed. It could break configurations where the configured users do not  (or do not always) run their processes within a complete systemd-managed user context.

### Changes from PR #2548

1. Remove the special treatment of `.ssh/config`.  As far as I can tell from OpenSSH's documentation, this was never necessary; sshd does not look at `.ssh/config` *at all*, so it does not need to be available pre-login.  The file that *does* need to be available pre-login, on systems that run sshd, is `authorized_keys`, and the NixOS ssh service module already handles that.

   This is the piece that addresses a serious problem; the special treatment of `.ssh/config` was buggy and was causing any attempt (by me) to build a NixOS configuration from @pasqui23's branch to crash.

2. Expand the documentation of the new option.

3. Remove the duplicate copy of the installation section of the manual (this appears to have been added by accident).

3. Squash all patches on the branch, for ease of review, and rebase against current top of trunk (8a31864).

### Checklist

  (Test run in progress, will update if there are any problems.  I see a lot of "trace: evaluation warning:" messages but so far they look to be unrelated to this change.)

   (I will happily add some test cases but I could use some advice first as to what sort of testing would be appropriate for a new option with global effects.)

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee @khaneliman @ncfavier


pasqui23:
Don't worry, clearly I didn't have the time to properly develop it

LordGrimmauld:
I relied on my own adaptation of @pasqui23's work - this is especially useful when using PAM-mounts, as with PAM-mounts the user home is not even mountable to root before user login. I love to see some progress on this!

mikunimaru:
Home-manager startup at login worked fine with patch #2548. But with this new patch, home-manager doesn't start correctly at login for me.

My setup mounts the home directory at login. Patch #2548 fixed a home-manager startup issue with this setup.

Question: Did this new patch change anything related to timing (e.g., service startup order) that could affect the fix from #2548?

If this new patch breaks the #2548 fix, the old problem (where home-manager doesn't start correctly) will return, and this is a big issue for my setup.

zackw:
@mikunimaru I want this patch for the same reason you do -- mounting the home directory at login -- and I did not intentionally change anything that should have broken this usage.  I'm guessing one of the changes from home-manager master branch, that weren't included in #2548, is at fault.  Can you share more details on the behavior change you are observing?  What errors did you get, which operations changed their timing, etc.

mikunimaru:
@zackw Thank you for your kind reply.
I investigated according to your advice and found that there was a problem with my local settings.
This patch works fine and doesn't break #2548. 
Sorry for the noise. Looking forward to seeing this merged.

cassandracomar:
applied this PR as a patch and it works perfectly to allow `security.pam.zfs` to mount/unmount encrypted home dirs. prior to this PR, the home-manager service would attempt to load the home mounts on boot and after updating the system generation, and it would hold the systemd mount even after the user logged out.

so thanks for this and I hope this gets merged soon!

zackw:
Stalebots are bad practice. Please turn it off for this repository.

Korvox:
There hasn't been movement on this in months and myself and several others in here are stuck maintaining this out of tree to use home manager with a very common usage paradigm (encrypted home). Is there a reason its dangling?

sini:
I've been watching this PR for a while and I don't see any review feedback that would push it towards merging or reason for rejection. Anything that can be done? CC: @rycee @khaneliman @ncfavier

nixos/default.nix
```diff
@@ -7,19 +7,54 @@
...
 
+  # we use a service separated from nixos-activation
+  # to keep the logs separate
+  hmDropIn = "/share/systemd/user/home-manager.service.d";
 in
 {
   imports = [ ./common.nix ];
 
+  options.home-manager.useUserService = lib.mkEnableOption ''
+    Activate each user's environment only when they log in, using a
```
teto: I believe the generated doc prepends "Enable" so that wont render great. maybe use `mkOption` instead.

nixos/default.nix
```diff
@@ -97,10 +117,53 @@ in
...
+        )
+      ) cfg.users;
+    })
+
+    (lib.mkIf (cfg.users != { } && cfg.useUserService) {
+      systemd.user.services.home-manager = (baseUnit "%u") // {
+        # user units cannot depend on system units
+        # TODO: Insert in the script logic for waiting on the nix socket
+        # via dbus like https://github.com/mogorman/systemd-lock-handler
+        #wants = [ "nix-daemon.socket" ];
```
teto: are these comments meant to stay ?

teto:
I understand how it could be useful to some. Willing to merge.

mio-19:
> I understand how it could be useful to some. Willing to merge.

It is very useful to me! I am using seperated encrypted user home with zfs native encryption. system level service could not do activation correctly

khaneliman:
Looks overall fine to me, would just need a rebase and clean up some comments.

ncfavier:
This has been around for a while, many people seem to find it useful, it works as far as I could tell and is not enabled by default. I don't see any objections to merging this. I've rebased on `master`. @zackw could you take care of @teto's remaining comments (remove the TODO stuff if you don't plan on doing it in this PR)?

P.S.:

> Stalebots are bad practice. Please turn it off for this repository.

Agreed, opened https://github.com/nix-community/home-manager/pull/8553.

nixos/default.nix
teto: also I suppose you reeused the `use` from `useUserPacakges` in `useUserService` but I believe it would be more self-explanatory as `startAsUserService` ?

teto:
I've tested it and it seemed to work so willing to merge. Ideally we would have @zackw answer the comments but in absence of it:
- we could merge a PR based on this one with a commit addressing comment (keeping attribution)
- merge this if someone agrees to do a quick follow up ?

khaneliman:
> I've tested it and it seemed to work so willing to merge. Ideally we would have @zackw answer the comments but in absence of it:
> 
>     * we could merge a PR based on this one with a commit addressing comment (keeping attribution)
> 
>     * merge this if someone agrees to do a quick follow up ?

There are still a few things that should be cleaned up, would recommend a commit in same PR

nixos/default.nix
zackw: Made both these changes.

nixos/default.nix
zackw: I have updated this to state the systemd limitation but not describe a todo item, and I will file a follow-up issue for the todo item.  I do not plan to work on the todo item myself.

zackw:
I've made the requested changes (as a separate commit for ease of review, but it should probably be squashed before merging).

Thank you @ncfavier for dealing with the merge conflicts, I tried to do that yesterday and did not have enough brain.

nixos/default.nix
```diff
@@ -46,40 +88,27 @@ in
         ];
       };
     }
-    (mkIf (cfg.users != { }) {
+
+    (mkIf (cfg.users != { } && !cfg.useUserService) {
```
khaneliman:
```suggestion
    (mkIf (cfg.users != { } && !cfg.startAsUserService) {
```

nixos/default.nix
```diff
@@ -103,10 +132,51 @@ in
...
               in
-              "${setupEnv} ${usercfg.home.activationPackage}";
-          };
+              "${setupEnv} ${activationPackage}";
+          }
+        )
+      ) cfg.users;
+    })
+
+    (mkIf (cfg.users != { } && cfg.useUserService) {
```
khaneliman:
```suggestion
    (mkIf (cfg.users != { } && cfg.startAsUserService) {
```

nixos/default.nix
zackw: You didn't see that. ^_^

khaneliman:
@zackw if you're comfortable, could you run the script to create a news entry and add a blurb about the new option `nix run .#create-news-entry`. otherwise, we can do it after.

zackw:
Will do shortly. Do you want all the patches on this branch squashed?

khaneliman:
> Will do shortly. Do you want all the patches on this branch squashed?

yes please, news entry can be a separate commit

PR #6989, opened 2025-05-06
alot: allow commas in maildir path

SiriusStarr:
### Description

Previously, maildir path was not escaped in any way, so using a path with commas lead to a TypeError:

```
TypeError: expected string or bytes-like object, got 'list'
```

This change wraps the maildir path in quotes to allow commas and other special characters to be tolerated.

### Checklist

- [X] Code formatted with `nix fmt` or `./format`.

    ```
    {component}: {description}

    {long description}
    ```


khaneliman:
need to rebase to resolve CI issue

SiriusStarr:
Rebased to fix CI failure.

PR #6991, opened 2025-05-07
lib/strings: add {toKebabCase,toSnakeCase,toCaseWithSeparator}; lib/deprecations: add mkSettingsRenamedOptionModules with transform

khaneliman:
### Description
Follow up to https://github.com/nix-community/home-manager/pull/6987#issuecomment-2856395024

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/lib/default.nix
+++ b/modules/lib/default.nix
@@ -6,6 +6,7 @@ rec {
   assertions = import ./assertions.nix { inherit lib; };
 
   booleans = import ./booleans.nix { inherit lib; };
+  deprecations = import ./deprecations.nix { inherit lib; };
   generators = import ./generators.nix { inherit lib; };
   gvariant = import ./gvariant.nix { inherit lib; };
   maintainers = import ./maintainers.nix;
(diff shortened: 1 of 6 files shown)


khaneliman:
@aguirre-matteo This will resolve the rename issue

different-name:
Hi, I am having issues with this
My config keys don't seem to be getting converted from camelCase to kebab-case for Mako
The following does not work
```nix
services.mako = {
  enable = true;
  settings = {
    borderColor = "#ed507c";
  };
};
```
```shell
$ mako
[config:1] Failed to parse option 'borderColor=#ed507c'
Failed to parse config
```
However changing `borderColor` to `border-color` works perfectly

I feel like I am doing *something* wrong here as it appears to be working for the people over at catppuccin/nix who are using camelCase https://github.com/catppuccin/nix/pull/553

I am using the home-manager flake at commit e95a7c5b6fa93304cd2fd78cf676c4f6d23c422c

khaneliman:
You’re not supposed to use camel case in settings. It’s only for migrating from the old options to the new Freeform type settings. Settings should match upstream

Issue #6835, opened 2025-04-17
zed-editor: coexist with preexisting settings and shortcuts


yajo:
### Description

HM replaces any user-chosen settings and keymaps when using one of these options:
- [programs.zed-editor.extensions](https://nix-community.github.io/home-manager/options.xhtml#opt-programs.zed-editor.extensions)
- [programs.zed-editor.userKeymaps](https://nix-community.github.io/home-manager/options.xhtml#opt-programs.zed-editor.userKeymaps)
- [programs.zed-editor.userSettings](https://nix-community.github.io/home-manager/options.xhtml#opt-programs.zed-editor.userSettings)

This is problematic in zed as explained in [the wiki](https://wiki.nixos.org/wiki/Zed#Home_manager_support):

> note that home-manager configuration produces a read only `settings.json` this means some features which requires to write this file might not work. For example change model at runtime. Changing AI model at runtime tries to write in the configuration.

This makes zed less user-friendly when configured with HM. Instead, it would be better if it were able to merge values with those found within those files.

The joplin-desktop module for example already does that:
https://github.com/nix-community/home-manager/blob/c6b75d69b6994ba68ec281bd36faebcc56097800/modules/programs/joplin-desktop.nix#L140-L147

This enables the user change settings dynamically within the app, while still making sure you get the ones you hardcoded in HM when reloading it.

andresilva:
```
Jun 06 14:45:22 zetta hm-activate-andre[725847]: Activating zedKeymapActivation
Jun 06 14:45:22 zetta hm-activate-andre[727899]: jq: error (at /nix/store/mfpsmq04immjqivrl6kzixxhi72vfjbg-zed-user-keymaps:1): array ([]) and object ({}) cannot be added
```

I'm not using zed actively, but my current `keymap.json` is an empty array. I think the default [here](https://github.com/nix-community/home-manager/blob/0ee810c839ee73b2d3973f5683f9a3bf8d430e8a/modules/programs/zed-editor.nix#L76) should be `[ ]`.

khaneliman:
> ```
> Jun 06 14:45:22 zetta hm-activate-andre[725847]: Activating zedKeymapActivation
> Jun 06 14:45:22 zetta hm-activate-andre[727899]: jq: error (at /nix/store/mfpsmq04immjqivrl6kzixxhi72vfjbg-zed-user-keymaps:1): array ([]) and object ({}) cannot be added
> ```
> 
[...]

PR #6993, opened 2025-05-07
zed-editor: respect user interactivity with settings and keymaps

yajo:
### Description
Merges frozen config from HM into dirty config.

Fixes https://github.com/nix-community/home-manager/issues/6835

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@libewa

### DIFF
--- a/tests/modules/programs/zed-editor/extensions.nix
+++ b/tests/modules/programs/zed-editor/extensions.nix
@@ -1,4 +1,9 @@
-{ config, ... }:
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
 
 {
   programs.zed-editor = {
@@ -11,20 +16,54 @@
     ];
   };
 
+  home.homeDirectory = lib.mkForce "/@TMPDIR@/hm-user";
+
   nmt.script =
     let
+      preexistingSettings = builtins.toFile "preexisting.json" ''
+        {
+          "auto_install_extensions": {
+            "python": true,
+            "javascript": true
+          }
+        }
+      '';
+
       expectedContent = builtins.toFile "expected.json" ''
         {
           "auto_install_extensions": {
+            "python": true,
+            "javascript": true,
             "html": true,
             "swift": true,
             "xy-zed": true
           }
         }
       '';
+
+      settingsPath = ".config/zed/settings.json";
+
+      activationScript = pkgs.writeScript "activation" config.home.activation.zedSettingsActivation.data;
     in
     ''
-      assertFileExists "home-files/.config/zed/settings.json"
-      assertFileContent "home-files/.config/zed/settings.json" "${expectedContent}"
+      export HOME=$TMPDIR/hm-user
+
+      # Simulate preexisting settings
+      mkdir -p $HOME/.config/zed
+      cat ${preexistingSettings} > $HOME/${settingsPath}
+
+      # Run the activation script
+      substitute ${activationScript} $TMPDIR/activate --subst-var TMPDIR
+      chmod +x $TMPDIR/activate
+      $TMPDIR/activate
+
+      # Validate the merged settings
+      assertFileExists "$HOME/${settingsPath}"
+      assertFileContent "$HOME/${settingsPath}" "${expectedContent}"
+
+      # Test idempotency
+      $TMPDIR/activate
+      assertFileExists "$HOME/${settingsPath}"
+      assertFileContent "$HOME/${settingsPath}" "${expectedContent}"
     '';
 }
--- a/modules/programs/zed-editor.nix
+++ b/modules/programs/zed-editor.nix
@@ -14,6 +14,16 @@ let
 
   cfg = config.programs.zed-editor;
   jsonFormat = pkgs.formats.json { };
+  impureConfigMerger = empty: jqOperation: path: staticSettings: ''
+    mkdir -p $(dirname ${lib.escapeShellArg path})
+    if [ ! -e ${lib.escapeShellArg path} ]; then
+      # No file? Create it
+      echo ${lib.escapeShellArg empty} > ${lib.escapeShellArg path}
+    fi
+    config="$(${pkgs.jq}/bin/jq -s ${lib.escapeShellArg jqOperation} ${lib.escapeShellArg path} ${lib.escapeShellArg staticSettings})"
+    printf '%s\n' "$config" > ${lib.escapeShellArg path}
+    unset config
+  '';
 
   mergedSettings =
     cfg.userSettings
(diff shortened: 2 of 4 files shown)


khaneliman:
rebase to resolve CI

yajo:
Done. I also modified the script a little bit to ease automatic migrations from older versions of the module.

khaneliman:
```bash
mkdir: cannot create directory '/build': Read-only file system
building '/nix/store/jwhk0sd3ny8dl0dm29yjmmia96ny88sw-nmt-test-zed-extensions.drv'...
error: builder for '/nix/store/arl6x2ra1fqwvbjjpnsrs8yiprsqcrjz-nmt-test-zed-keymap.drv' failed with exit code 1;
       last 1 log lines:
       > mkdir: cannot create directory '/build': Read-only file system
       For full logs, run:
         nix log /nix/store/arl6x2ra1fqwvbjjpnsrs8yiprsqcrjz-nmt-test-zed-keymap.drv
```

tests/modules/programs/zed-editor/extensions.nix
```diff
@@ -11,20 +11,50 @@
     ];
   };
 
+  home.homeDirectory = lib.mkForce "/build/hm-user";
```
yajo, line 14: It looks like MacOS doesn't use `/build/` as the directory where builds are performed. I only have Linux machines. Do you have any idea on what's the correct value here for Mac so I can fix the test?

khaneliman: I have a macbook. I can take a look at this later, when I get time.

nickdichev:
I was able to build this on aarch64-darwin with your latest commit

khaneliman:
Looks like formatting failed.

tests/modules/programs/zed-editor/extensions.nix
khaneliman: Sorry, i forgot about this...

yajo:
Thanks, I reformatted.

yajo:
So, could this one be merged?

libewa:
Other editors (code) also have something like `mutableSettings`

khaneliman:
> Other editors (code) also have something like `mutableSettings`

I see `mutableExtensionsDir` but not `mutableSettings`. Right now, the home manager module, from my understanding, has a similar limitation to this zed-editor one. Where we write a read only file and vscode tries to update it and fails.

yajo:
Immutable settings in zed are extremely uncomfortable. Many of its buttons only work when writing to its settings file. Zed is not designed to work without write access to its settings: https://github.com/zed-industries/zed/issues/24537#issuecomment-2784496147

I understand your concerns, but I'm not sure implementing immutable settings has any use for Zed.

khaneliman:
@libewa any opinion on implementation?

yajo:
After trying it locally, I have changed a bit the algorithm.

I removed the part that was checking for a read-only file. Because of how the DAG is set, those symlinks will be removed automatically when we get to this part of the script. Thus, there's no need to check that.

I also added an argument to set the default value for an empty file. `jq` wasn't able to multiply an empty file with an object.

nyabinary:
Can we get this merged?

khaneliman:
> Can we get this merged?

Failing CI

nyabinary:
```error: 'thefuck' has been removed due to lack of maintenance upstream and incompatible with python 3.12+. Consider using 'pay-respects' instead```

khaneliman:
Sorry, can you rebase. CI was blocked.

nyabinary:
@khaneliman should be good now with that recent push

Masrepus:
I just noticed a small issue this PR introduced: In case you don't have an existing keymap.json file and you also didn't set `userKeymaps`, the default value provided by the Nix module is going to be `{}` instead of `[]`. This leads to some jq issues while activating the config and ultimately to Zed complaining about an invalid keymap file. Manually setting it to `[]` resolves it, so I believe the default should be updated

andresilva:
https://github.com/nix-community/home-manager/pull/7222

CodedNil:
Is there any way to opt out of this change? I have noticed that if I change settings for example adding experimental theme overrides, any time i remove an option in the declarative it will just linger and i have to delete the settings file. I personally would never use any of the zed settings ui imperatively.

yajo:
Currently there's no opt-out option because of https://github.com/nix-community/home-manager/pull/6993#issuecomment-2898561400.

However if you still have use for it, you can open a PR 😊. A very simple implementation would be to delete the preexisting files before running current script. Since it would merge with nothing, it would just restore your nix stuff always.

musjj:
I'd also appreciate an option to turn this off. This seems to be completely against the spirit of reproducibility and immutability of Nix. When I configure something with Nix, I expect it to produce the same results in every machine that I deploy to, and this completely breaks that expectation.

There are some downsides with Zed, but the AI model switcher thing is basically the *only* real deal-breaker here. And if you don't use it there's basically no downsides to immutability. In any case, that issue is something that should be fixed upstream, instead of contorting the entire module just to please this one use-case.

For example, VS Code has option called `programs.vscode.mutableExtensionsDir`. Maybe we can have a similar option for Zed.

woile:
This feature has helped me with the migration of the settings, as the editor was reporting deprecated settings and thanks to this I was able to go back to nix and update them easily. It would be nice to have a setting to be able to switch between mutable and immutable, but if there's no work on it, I'd rather have it like it's now.

yajo:
Zed is designed to have writable settings files. Like it or not, that's how it's done. [Upstream has confirmed that's their vision on this subject](https://github.com/zed-industries/zed/issues/24537#issuecomment-2784496147).

Those files [can even contain secrets](https://github.com/akbxr/gitlab-mcp-zed?tab=readme-ov-file#required-settings), which you don't want in your Nix store.

Just the work of adding a simple hotkey was a nightmare with the previous implementation. You had to:
1. Copy the `keymaps.json` file contents.
2. Delete it because it was a symlink to the nix store.
3. Create a new one.
4. Paste the contents.
6. Change the keymap.
8. Test it works.
9. Copy what you've done into your nix configuration, converted into Nix.
11. Remove your `keymaps.json` file again.
12. Re-apply your home-manager configuration.

Now it's simpler:
1. Edit the `keymaps.json` file.
2. Test it works.
3. If you want it to become part of your deterministic configuration, copy what you've done into your nix configuration, converted into Nix.

Changing the AI model was a nightmare too, as you said.

Also changing icons or colors, or testing themes.

If at any point you want to reset your settings, all you have to do is delete the files (at least after #7351 is merged).

I understand complete determinism has its benefits, but it's not practical for Zed.

PR #6994, opened 2025-05-07
jujutsu: store configuration in $XDG_CONFIG_HOME for all platforms

bnjmnt4n:
Following https://github.com/jj-vcs/jj/pull/6300, Jujutsu has deprecated support for configuration files in `~/Library/Application Support` for darwin. The XDG-standard configuration location can be used instead, for all platforms.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/jujutsu.nix
+++ b/modules/programs/jujutsu.nix
@@ -9,9 +9,14 @@ let
 
   cfg = config.programs.jujutsu;
   tomlFormat = pkgs.formats.toml { };
+  packageVersion = lib.getVersion cfg.package;
 
-  configDir = if pkgs.stdenv.isDarwin then "Library/Application Support" else config.xdg.configHome;
-
+  # jj v0.29+ deprecated support for "~/Library/Application Support" on Darwin.
+  configDir =
+    if pkgs.stdenv.isDarwin && !(lib.versionAtLeast packageVersion "0.29.0") then
+      "Library/Application Support"
+    else
+      config.xdg.configHome;
 in
 {
   meta.maintainers = [ lib.maintainers.shikanime ];
--- a/tests/modules/programs/jujutsu/empty-config.nix
+++ b/tests/modules/programs/jujutsu/empty-config.nix
@@ -1,7 +1,19 @@
-{ pkgs, ... }:
-
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
 let
-  configDir = if pkgs.stdenv.isDarwin then "Library/Application Support" else ".config";
+  cfg = config.programs.jujutsu;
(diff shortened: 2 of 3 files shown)


bnjmnt4n:
> Change is backwards compatible.

This change definitely isn't backward compatible for Darwin users on older versions of Jujutsu (before the upcoming v0.29.0 release). However, from that release onwards, there will be a noisy warning on every invocation of that function. I'm not sure what the appropriate solution is here. Should I try using `lib.versionAtLeast` to check for a minimum version of Jujutsu?

khaneliman:
> > Change is backwards compatible.
> 
> This change definitely isn't backward compatible for Darwin users on older versions of Jujutsu (before the upcoming v0.29.0 release). However, from that release onwards, there will be a noisy warning on every invocation of that function. I'm not sure what the appropriate solution is here. Should I try using `lib.versionAtLeast` to check for a minimum version of Jujutsu?

Yeah, we can check if the `cfg.package` version is recent enough for that change.

khaneliman:
```bash
error: builder for '/nix/store/82g1w8h51zhzzs09svlqzm0mis5axrpm-nmt-test-jujutsu-example-config.drv' failed with exit code 1;
       last 1 log lines:
       > Expected home-files//home/hm-user/.config/jj/config.toml to exist but it was not found.
       For full logs, run:
         nix log /nix/store/82g1w8h51zhzzs09svlqzm0mis5axrpm-nmt-test-jujutsu-example-config.drv
```

bnjmnt4n:
I realized this doesn't work for users of the Jujutsu flake because the version there is set to [`unstable-[shortRev]`](https://github.com/jj-vcs/jj/blob/4711b787120ebd7ad95a399988b4c803c6445b4e/flake.nix#L93). Does anyone have any recommendations on how this can be handled?

LucioFranco:
@bnjmnt4n thanks for getting this change in! For the moment, I just overrode the version locally. I wonder if upstream we could make them change the unstable version to be something like `unstable-{shortRev}-{cargoVersion}` and then I think `getVersions` will resolve correctly when using the flake.

```
jujutsu = (
  prev.jujutsu.overrideAttrs {
    version = "unstable-0.29.0";
  }
);
```

khaneliman:
The nixpkgs convention is to use `0.29.0-unstable-YYYY-MM-DD`

bnjmnt4n:
Thanks for the suggestion—I've created the following PR: https://github.com/jj-vcs/jj/pull/6499.

evantravers:
Will this be backported to home-manager 24.05 or 24.11, or is this the kind of thing that is saved for next release?

PR #6999, opened 2025-05-08
direnv: update nushell env conversion logic

koffydrop:
### Description
Direnv exports PATH as a string, nu expects it to be a list and breaks some functionality like external command completion. This requires nushell 0.104.0, which is merged into nixpkgs-unstable.

Change loosely based on [this](https://github.com/nushell/nushell.github.io/pull/1878) while trying to not change the logic too much

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@khaneliman
@rycee
@shikanime

### DIFF
--- a/modules/programs/direnv.nix
+++ b/modules/programs/direnv.nix
@@ -175,6 +175,7 @@ in
       # Using mkAfter to make it more likely to appear after other
       # manipulations of the prompt.
       programs.nushell.extraConfig = mkIf cfg.enableNushellIntegration (mkAfter ''
+        use std/config env-conversions
         $env.config = ($env.config? | default {})
         $env.config.hooks = ($env.config.hooks? | default {})
         $env.config.hooks.pre_prompt = (
@@ -186,8 +187,8 @@ in
                 | default {}
                 | items {|key, value|
                     let value = do (
-                        $env.ENV_CONVERSIONS?
-                        | default {}
+                        { PATH: (env-conversions).path }
+                        | merge ($env.ENV_CONVERSIONS? | default {})
                         | get -i $key
                         | get -i from_string
                         | default {|x| $x}
@@ -198,6 +199,7 @@ in
                 | load-env
             }
         )
+        hide env-conversions
       '');
 
       home.sessionVariables = lib.mkIf (cfg.silent && !isVersion236orHigher) { DIRENV_LOG_FORMAT = ""; };


jaredmontoya:
I guess this is supposed to be a fix for this:
image

koffydrop:
Pretty much, yeah. I'm not sure if it breaks anything else, but, tab completing external commands (anything in PATH that's not a shell command) does.

I could go into a bit more detail of what the code does if you two want.

jaredmontoya:
I read more into the code and it looks to me like it won't break anything.

modules/programs/direnv.nix
```diff
@@ -186,8 +187,8 @@ in
                 | default {}
                 | items {|key, value|
                     let value = do (
-                        $env.ENV_CONVERSIONS?
-                        | default {}
+                        { PATH: (env-conversions).path }
```
JoaquinTrinanes, line 190: I think we can skip importing and hiding `env-conversions` by scoping it:

```nu
{ PATH: (do {
    use std/config env-conversions
    (env-conversions).path
    }) }
```

JoaquinTrinanes (changes requested):
Looks good!

I wonder if there's value in doing it like this:

```nu
let value = do (
        env-conversions
        | merge ($env.ENV_CONVERSIONS? | default {})
        | get -i $key
        | get -i from_string
        | default {|x| $x}
    ) $value
```

AFAIK `ENV_CONVERSIONS` (and env vars in general) aren't case sensitive, so this would use all conversions if more get added to the std (the current approach is also fine :slightly_smiling_face:)

---

I was about to approve the PR but just realized that this is a breaking change for users that use the `--no-std-lib` flag. The only way I found of detecting if it's available is checking the output of `view files`:

`view files | find std`:

|index|filename|start|end|size|
|-|-|-|-|-|
|0|std/mod.nu|17031|17765|734|
|1|std/prelude/mod.nu|17765|20375|2610|
|2|std/assert/mod.nu|20375|28294|7919|
|3|std/bench/mod.nu|28294|30037|1743|
|4|std/dirs/mod.nu|30037|34875|4838|
|5|std/dt/mod.nu|34875|42409|7534|
|6|std/formats/mod.nu|42409|43567|1158|
|7|std/help/mod.nu|43567|77496|33929|
|8|std/input/mod.nu|77496|79218|1722|
|9|std/iter/mod.nu|79218|83859|4641|
|10|std/log/mod.nu|83859|92410|8551|
|11|std/math/mod.nu|92410|92595|185|
|12|std/util/mod.nu|92595|95641|3046|
|13|std/xml/mod.nu|95641|103707|8066|
|14|std/config/mod.nu|103707|108603|4896|
|15|std/testing/mod.nu|108603|109071|468|
|16|std-rfc/mod.nu|109071|109362|291|
|17|std-rfc/clip/mod.nu|109362|111162|1800|
|18|std-rfc/conversions/mod.nu|111162|113604|2442|
|19|std-rfc/kv/mod.nu|113604|119471|5867|
|20|std-rfc/path/mod.nu|119471|121207|1736|
|21|std-rfc/str/mod.nu|121207|124853|3646|
|22|std-rfc/tables/mod.nu|124853|130698|5845|
|23|loading stdlib prelude|130698|130727|29|

khaneliman:
Oops saw last message and then merged, but read again `This requires nushell 0.104.0, which is merged into nixpkgs-unstable.` Which looks like a fairly recent change we might want to gate behind.

modules/programs/direnv.nix
koffydrop: I opted to import it earlier and hide after to avoid having to import it every time the closure runs, I'm not sure how impactful on performance one would be over the other.

koffydrop:
https://github.com/nix-community/home-manager/pull/6999#pullrequestreview-2828241172
> I wonder if there's value in doing it like this:
> 
> ```nu
> let value = do (
>         env-conversions
>         | merge ($env.ENV_CONVERSIONS? | default {})
>         | get -i $key
>         | get -i from_string
>         | default {|x| $x}
>     ) $value
> ```
That makes more sense! It did not occur to me, oops

PR #7015, opened 2025-05-09
direnv: update nushell env conversion logic

jaredmontoya:
### Description
Follow up to #6999

Includes suggestions and version gating.

My first time using `lib.versionAtLeast` and I only tested the case of when nushell is equal to 0.104.0 so can't say for sure it uses the old implementation when nushell is older than 0.104.0
If I used this function in a wrong way, tell me.

There was also a concern about users who use `--no-std-lib` flag being affected. I really don't know what to do with that, runtime gating with `view files | find std` feels wrong.

Why are people disabling stdlib anyway?

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@khaneliman @rycee @shikanime @JoaquinTrinanes


koffydrop:
With regards to breaking `--no-std-lib` we could just duplicate the code from it since it's very little:
https://github.com/nushell/nushell/blob/0.104.0/crates/nu-std/std/config/mod.nu#L138
```nu
# Returns helper closures that can be used for ENV_CONVERSIONS and other purposes
export def env-conversions [] {
    {
        "path": {
            from_string: {|s| $s | split row (char esep) | path expand --no-symlink }
            to_string: {|v| $v | path expand --no-symlink | str join (char esep) }
        }
    }
}
```
and instead of a version check, just merge that record into $env.ENV_CONVERSIONS.
```nu
$env.config = ($env.config? | default {})
$env.config.hooks = ($env.config.hooks? | default {})
$env.config.hooks.pre_prompt = (
$env.config.hooks.pre_prompt?
	| default []
	| append {||
		${getExe cfg.package} export json
		| from json --strict
		| default {}
		| items {|key, value|
		| value = do (
			{
				"path": {
					from_string: {|s| $s | split row (char esep) | path expand --no-symlink }
					# probably omitting this since it doesn't affect this particular module
					to_string: {|v| $v | path expand --no-symlink | str join (char esep) }
				}
			}
        	| merge ($env.ENV_CONVERSIONS? | default {})
        	| get -i $key
        	| get -i from_string
        	| default {|x| $x}
		) $value
    	[ $key $value ]
	}
	| into record
	| load-env
})

**Comment by jaredmontoya** [comment]
This certainly is the easiest option to get this merged with no consequences.

But what about the long term implications of refusing to use nushell stdlib in home-manager  because we support the use case of running nushell with --no-std-lib?
What if it ends up being used more and more, are we just going to copy code snippets from stdlib each time we need them?

**Comment by koffydrop** [comment]
https://github.com/nix-community/home-manager/pull/7015#issuecomment-2867720376
> I don't think we should skip `to_string`

It doesn't make a difference since internally nushell already has and applies those conversions for PATH, they're just not exposed to the user or in `$env.ENV_CONVERSIONS`.

https://github.com/nix-community/home-manager/pull/7015#issuecomment-2867732695
For those cases maybe extracting the stdlib functions at build time with a mix of `readFile` and `runCommand` is solution?
From quick testing in a nix repl:
```nix
builtins.readFile (pkgs.runCommandLocal "test.nu" {} ''
	${lib.getExe pkgs.nushell} -c "use std/config env-conversions; view source env-conversions | save -f $out"
'')
```
`$out` will then be the nu stdlib definition for `env-conversions`, which we could add to the extraConfig string without relying on actual std.

koffydrop:
Something like this (tested locally, doesn't break with `--no-std-lib`)
```nix
# modules/programs/direnv.nix
programs.nushell.extraConfig = mkIf cfg.enableNushellIntegration (mkAfter ''
  source ${pkgs.runCommandLocal "direnv-nushell-helper.nu" { } ''
    ${lib.getExe pkgs.nushell} -c "use std/config env-conversions; view source env-conversions | save -f $out"
  ''}
  $env.config = ($env.config? | default {})
  $env.config.hooks = ($env.config.hooks? | default {})
  $env.config.hooks.pre_prompt = (
      $env.config.hooks.pre_prompt?
      | default []
      | append {||
          ${getExe cfg.package} export json
          | from json --strict
          | default {}
          | items {|key, value|
              let value = do (
                  env-conversions
                  | merge ($env.ENV_CONVERSIONS? | default {})
                  | get -i $key
                  | get -i from_string
                  | default {|x| $x}
              ) $value
              [ $key $value ]
          }
          | into record
          | load-env
      }
  )
  hide env-conversions
'');
```

jaredmontoya:
@koffydrop nice idea, I implemented it.
@JoaquinTrinanes what do you think about this?

modules/programs/direnv.nix
```diff
@@ -186,9 +186,25 @@ in
                 | default {}
                 | items {|key, value|
                     let value = do (
-                        $env.ENV_CONVERSIONS?
-                        | default {}
-                        | get -i $key
+                        ${
+                          if (lib.versionAtLeast (lib.getVersion cfg.package) "0.104") then
```
koffydrop, line 190: Wouldn't `cfg.package` be the direnv package? This check depends on nushell version 0.104

modules/programs/direnv.nix
```diff
@@ -186,9 +186,25 @@ in
...
+                            ''
+                              (do {
+                                source ${
+                                  pkgs.runCommand "nushell-stdlib-env-conversions.nu" { } ''
+                                    ${lib.getExe config.programs.nushell.package} -c "use std/config env-conversions; view source env-conversions | save -f $out"
+                                  ''
+                                }
+                                env-conversions
+                              })
```
koffydrop, line 199: I still think sourcing the file should go outside any closures and env-conversions hidden after to avoid having to source it *every time the prompt renders*.

modules/programs/direnv.nix
```diff
@@ -186,9 +186,25 @@ in
...
+                              | merge ($env.ENV_CONVERSIONS? | default {})
+                            ''
+                          else
+                            ''
+                              $env.ENV_CONVERSIONS?
+                              | default {}
+                            ''
+                        }| get -i $key
```
koffydrop, line 207: Should this pipe be on a new line? Not sure how the interpolation in nix works out here.

koffydrop (commented):
A few notes, and this is my first time doing a review here, bear with me

modules/programs/direnv.nix
jaredmontoya: It doesn't really matter because of Nu's syntax, it will work anyway.
I just did this so that in the generated config there won't be an empty line between `| default {}` and `| get -i $key`. This doesn't mean however that the indentation of the part that is conditional depending on the version will not be scuffed.
That part looks ugly in the generated config, but I prioritized readability in the home-manager codebase.

JoaquinTrinanes:
@jaredmontoya 

> But what about the long term implications of refusing to use nushell stdlib in home-manager because we support the use case of running nushell with --no-std-lib?

I think it's more likely that the stdlib breaks (the `env-conversion` helper was added ~3 weeks ago). I think it makes more sense to support `--no-std-lib` (because otherwise nushell would break _now_ for some users) than to worry about future breakage that we can fix.

>  What if the nushell stdlib ends up being developed and used more in the future, are we just going to copy code snippets from stdlib each time we need them? What if the function we need is 50 lines long?

In a perfect workd we won't need to use stdlib at all. I guess that if in the future an integration requires stdlib to work then it's on the user to not use `--no-std-lib` or to disable the integration. But right now it is not expected by any integration that I know of.

> Why are people disabling stdlib anyway?

AFAIK the main reason is startup time. That got way better in recent nushell releases tho (I just tested it and go from ~42ms -> ~38ms on average, so it's not that big of a deal).

> This is certainly something to think about, but I will just go with your suggestion if you read this and still say that we should just copy the env-conversions implementation into home-manager and be done with it (for now).

I am 100% NOT willing to die on this hill. I support using `ENV_CONVERSIONS` for `direnv` (hell, I wrote that code and the original version in the cookbook because it was so annoying to _not_ have it). At that time `PATH` was present in `ENV_CONVERSIONS` (I think it should still be there but I'm digressing). But I don't like the idea of breaking the configs of users because we make assumptions about the shell being used the "normal, expected way". If nushell supports `--no-std-lib`, it's the module that breaks, not the user that uses a non-deprecated option. So I'd go with just copying it personally (it's not just my decision tho :slightly_smiling_face:) 

----

@koffydrop 

> >  I don't think we should skip to_string
>
> It doesn't make a difference since internally nushell already has and applies those conversions for PATH, they're just not exposed to the user or in $env.ENV_CONVERSIONS.

You're right, I did the same tests but with the actual PATH variable and it works fine (I assumed that explicitly adding it to `ENV_CONVERSIONS` would overwrite both internal conversions even if one is missing). But I still wouldn't skip it because [missing one of the items is not officially supported](https://www.nushell.sh/book/configuration.html#env-conversions) (but we're probably bikeshedding at this point).

>  For those cases maybe extracting the stdlib functions at build time with a mix of readFile and runCommand is solution?

That would probably work, but sadly it causes [IFD](https://nix.dev/manual/nix/2.28/language/import-from-derivation). We already cause IFD in some specific cases (when setting the source of the config/env file) but I'd like to fix that at some point and avoid new cases of it.

-----

I guess this is a design choice on my side, but I like to offer the bare minimum functionality in home-manager, but powerful enough for users to support any usecase (this usually includes avoiding deviating for defaults and trying to do fancy stuff that can be easily added by them). Happy to hear your thoughts ~about my rambling~ :grin:

modules/programs/direnv.nix
jaredmontoya: nice catch, fixed

JoaquinTrinanes:
Random thought: if we want to support `PATH`, we can technically add it to `ENV_CONVERSIONS` in the normal config for everyone by default and have it work here automatically? 

```nix
programs.nushell.extraConfig = ''
$env.ENV_CONVERSIONS = { 
    path: {...}
    ...($env.ENV_CONVERSIONS? | default {}) 
}
'';
```

koffydrop:
> That would probably work, but sadly it causes IFD. We already cause IFD in some specific cases (when setting the source of the config/env file) but I'd like to fix that at some point and avoid new cases of it.

Do you know of a way that could be evaluated at build time without causing IFD that could be applied here?

> I guess this is a design choice on my side, but I like to offer the bare minimum functionality in home-manager, but powerful enough for users to support any usecase (this usually includes avoiding deviating for defaults and trying to do fancy stuff that can be easily added by them). Happy to hear your thoughts about my rambling 😁

I do generally agree, but, I feel like not breaking an important variable like PATH should be seen as part of the bare minimum functionality for a shell integration, not a deviation for a default or other fancy stuff, right?

modules/programs/direnv.nix
jaredmontoya: does hiding `env-conversions` and sourcing it again not cause it to be sourced again? I assume that's why nushell calls it `hide` probably but wouldn't nushell need to read the file from disk and parse it anyway?

koffydrop: I'm not exactly sure how that works internally. I *assume* it doesn't get sourced again since it's only calling the function that got defined, not explicitly sourcing the file. env-conversions itself is still in scope for closure before hiding it at the end of extaConfig.
Hiding it afterwards is so env-conversions isn't left in scope for the user session as a side effect of having this module.

modules/programs/direnv.nix
koffydrop: Ahh, gotcha thanks!

jaredmontoya:
Considering everything we have discussed so far, I think the best way to go forward with this PR is to just duplicate the full implementation. This way will bring the most benefit in the short term without breaking anything at all (even if nushell starts requiring both `to` and `from` implementations in `ENV_CONVERSIONS`).

modules/programs/direnv.nix
jaredmontoya: You are right.

koffydrop:
Makes perfect sense to me, I'll try to think of a simpler, more future-proof way of handling this.

Well i thought of something while typing this. Since nushell applies and does the conversion internally, we could make it do the conversion work for us. Something like this:
```nu
$env.config = ($env.config? | default {})
$env.config.hooks = ($env.config.hooks? | default {})
$env.config.hooks.pre_prompt = (
   	$env.config.hooks.pre_prompt?
   	| default []
   	| append {||
       	direnv export json
       	| from json --strict
       	| default {}
       	| items {|key, value|
           	let value = do (
               	$env.ENV_CONVERSIONS?
               	| default {}
               	| get -i $key
               	| get -i from_string
               	| default {|x| $x}
           	) $value
           	# if the value of PATH is still a string, get nu to convert it to nuon for us make it the new value
           	let value = if $key == PATH and ($value | describe) == string {
           		with-env {$key: $value} { nu -c "$env.PATH | to nuon" | from nuon }
           	} else {
           		$value
           	}
           	return [ $key $value ]
       	}
       	| into record
       	| load-env
   	}
)
```
Also just tested locally, no apparent issues or weird behavior

khaneliman:
Gonna unsubscribe from thread, for now, if you all want to ping me when you are satisfied with the final decision/implementation.

jaredmontoya:
https://github.com/nix-community/home-manager/pull/7015#issuecomment-2869266843
I am afraid this might not be sufficient in some edge cases. I don't see how it deals with the `path expand --no-symlink` part.
I also don't see any reason to do this because it's not going to reduce the amount of nushell code in home manager by much.

@khaneliman I think it's ready to merge as it is now.

khaneliman:
So, in the end this doesn't need a version check? From the sounds of it, we are writing the stdlib method into the logic ourselves for people who opt out of stdlib?

jaredmontoya:
> So, in the end this doesn't need a version check? From the sounds of it, we are writing the stdlib method into the logic ourselves for people who opt out of stdlib?

There is no stdlib being used now.

PR #7021, opened 2025-05-10
dconf: Provide

Rosuavio:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/dconf.nix
+++ b/modules/misc/dconf.nix
@@ -43,14 +43,7 @@ in
         # re-enabled, unclear whether there's actual value in it though.
         default = !pkgs.stdenv.hostPlatform.isDarwin;
         visible = false;
-        description = ''
-          Whether to enable dconf settings.
-
-          Note, if you use NixOS then you must add
-          `programs.dconf.enable = true`
-          to your system configuration. Otherwise you will see a systemd error
-          message when your configuration is activated.
-        '';
+        description = "Whether to enable dconf.";
       };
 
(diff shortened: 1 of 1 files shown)


Rosuavio:
When writing https://github.com/nix-community/home-manager/pull/7045 I found that certain paths are always linked from packages in `home.packages` I wonder is a similar solution could be achieved here?

- Do we want to ubiquitously link `/share/systemd/user/` paths from packages in `home.packages`?
- Do we want to ubiquitously link `/share/dbus-1/services` paths from packages in `home.packages`?
- Do we want be able to be able to selectively link files from packages?
- Do we want these file to end up in the `home-path` and not the users `$HOME`?

Rosuavio:
> Do we want to ubiquitously link `/share/systemd/user/` paths from packages in `home.packages`? 

Did not need to link those files to home.confFile."systemd/user" for this to work. Only had to add `dconf` to `home.packages`.
  
> Do we want to ubiquitously link `/share/dbus-1/services` paths from packages in `home.packages`?

No, this would mean enabling all the dbus services for all packages in `home.packages` that have dbus service files.

> Do we want these file to end up in the `home-path` and not the users `$HOME`?

For the dbus services files, they were already in `home-path` and they were not being picked up. They need to be linked to `home.dataFile."dbus-1/services/*" for them to be enabled.

This investigation yielded https://github.com/nix-community/home-manager/pull/7064

PR #7028, opened 2025-05-11
programs.numbat: add module

Aehmlo:
### Description

This PR adds a module for configuring [Numbat](https://numbat.dev/), a high-precision scientific calculator with full support for physical units. It uses [RFC 42](https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md)-style `settings` and provides `enable` and `package` options.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/numbat.nix
+++ b/modules/programs/numbat.nix
@@ -1,0 +1,50 @@
+{
+  lib,
+  config,
+  pkgs,
+  ...
+}:
+
+let
+  inherit (lib) mkIf;
+  cfg = config.programs.numbat;
+  tomlFormat = pkgs.formats.toml { };
+  configDir =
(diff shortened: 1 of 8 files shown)


modules/programs/numbat.nix
```diff
@@ -0,0 +1,50 @@
...
+}:
+
+let
+  inherit (lib) mkIf;
+  cfg = config.programs.numbat;
+  tomlFormat = pkgs.formats.toml { };
+  configDir =
+    if pkgs.stdenv.hostPlatform.isDarwin then
+      "Library/Application Support/numbat"
+    else
```
khaneliman, line 15: Just double checking, are you sure it doesn't respect xdg when its enabled on darwin?

PR #7039, opened 2025-05-12
Aerospace: auto reload config

phucleeuwu:
This tells Home Manager to run aerospace reload-config every time the config file is regenerated

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


rycee:
Thanks for the contribution! But have you actually tried using this code? You need to place the `onChange` inside the file's attribute set, see the CI error.

Also please fix the commit message (see the [contributing guidelines](https://nix-community.github.io/home-manager/#sec-commit-style)).

PR #7040, opened 2025-05-12
wayprompt: tweak example

panchoh:
### Description

Show that the hex values can be entered without the '0x' prefix.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@khaneliman @awwpotato

### DIFF
--- a/modules/programs/wayprompt.nix
+++ b/modules/programs/wayprompt.nix
@@ -42,7 +42,7 @@ in
             pin-square-amount = 32;
           };
           colours = {
-            background = "0xffffffaa";
+            background = "ffffffaa";
           };
         }
       '';


awwpotato (approved):
LGTM

khaneliman:
Can be or have to be? If it's a different option, I'd rather add a second example that change the only one.

panchoh:
> Can be or have to be? If it's a second option, I'd rather add a second example than change the only one.

Well, in fact, has to be. If the string has the `0x` prefix, it will be rendered as `"0xDEADBEEF";` (as a string within double quotes), which is not what the documentation (wayprompt(5)) says it accepts, which is `0xDEADBEEF;` (no quotes).

The fact that the string version works is just because the wayprompt parser is lax, but I’d rather not show it as an example.

So, the way I devised it, it should be used *without* the `0x` prefix, no question about that. Which is also the format that Stylix uses to represent colours, so, bonus! :sweat_smile: 

(note that the config will produce double quoted strings for font specs, such as `"sans:size=14"`.

PR #7045, opened 2025-05-12
fontconfig: Provide default fontconfig config files

Rosuavio:
### Description

Make home-manager provide fontconfig's default config files in `home-path/etc/font`, this fixes and issue where fonts would render incorrectly if the system was not already providing the fontconfig default config files.

#### Context

Using standalone hm, with fontconfig disabled in my NixOS config and enabled in hm, I have been experiencing font rendering issues in several programs, particularly in mako emojis would be oversized. This was due to fontconfig not being able to find its default config files that come with the package. This fixes that issue.

This change can be a starting point to then add all of the fontconfig options that NixOS provides in home-manager as well (as the implementation is similar so far).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/misc/fontconfig.nix
+++ b/modules/misc/fontconfig.nix
@@ -96,6 +96,19 @@ in
       # trying to write to a read-only location.
       (pkgs.runCommandLocal "dummy-fc-dir1" { } "mkdir -p $out/lib/fontconfig")
       (pkgs.runCommandLocal "dummy-fc-dir2" { } "mkdir -p $out/lib/fontconfig")
+      # Provide fontconfig default files from /etc/fonts/
+      (pkgs.runCommand "fontconfig-conf" { } ''
+        dst=$out/etc/fonts/conf.d
+        mkdir -p $dst
+
+        # fonts.conf
+        ln -s ${pkgs.fontconfig.out}/etc/fonts/fonts.conf \
+              $dst/../fonts.conf
+
+        # fontconfig default config files
+        ln -s ${pkgs.fontconfig.out}/etc/fonts/conf.d/*.conf \
+              $dst/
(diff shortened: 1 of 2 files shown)


Rosuavio:
@rycee, let me know what you think.

modules/misc/fontconfig.nix
```diff
@@ -96,6 +96,24 @@ in
       # trying to write to a read-only location.
       (pkgs.runCommandLocal "dummy-fc-dir1" { } "mkdir -p $out/lib/fontconfig")
       (pkgs.runCommandLocal "dummy-fc-dir2" { } "mkdir -p $out/lib/fontconfig")
+      # Provide fontconfig defualt files from /etc/fonts/
```
rycee, line 99:
```suggestion
      # Provide fontconfig default files from /etc/fonts/
```

modules/misc/fontconfig.nix
```diff
@@ -96,6 +96,24 @@ in
...
+      (pkgs.runCommand "fontconfig-conf"
+        {
+          preferLocalBuild = true;
+        }
```
rycee, line 103:
```suggestion
      (pkgs.runCommandLocal "fontconfig-conf" { }
```

rycee (commented):
Looks ok to me. Just some minor comments.

modules/misc/fontconfig.nix
Rosuavio: Done

modules/misc/fontconfig.nix
Rosuavio: Done

PR #7050, opened 2025-05-12
sketchybar: add module

### LINKED ISSUE #7017: Feature Request: Add module for Sketchybar configuration
I would like to request adding a module to manage [Sketchybar](https://felixkratz.github.io/sketchybar/) as part of Home Manager.

`Sketchybar` is the only customizable status bar that macOS user have and it would be great to configure it declaratively through `Home Manager`. At the moment, I need to manually install `Sketchybar` and maintain its config outside of Home Manager, typically under `~/.config/sketchybar` which is not declarative and sad 😞🫠.

Anyway, thank you for considering this

aguirre-matteo:
### Description

This PR adds a module for configuring Sketchybar. Closes #7017.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/sketchybar.nix
+++ b/modules/programs/sketchybar.nix
@@ -1,0 +1,86 @@
+{
+  lib,
+  pkgs,
+  config,
+  ...
+}:
+let
+  inherit (lib)
+    types
+    mkIf
+    mkEnableOption
+    mkPackageOption
+    mkOption
+    ;
+
+  cfg = config.programs.sketchybar;
+
+  getSource = { name, arg }: if lib.isString arg then pkgs.writeText name arg else arg;
+
+  getSketchybarrc =
+    setting:
+    if setting != "" then
+      {
+        "sketchybar/sketchybarrc".source = getSource {
+          name = "sketchybarrc";
+          arg = setting;
+        };
+      }
+    else
+      { };
+
+  getPlugins =
+    set:
+    lib.mapAttrs' (
+      name: value:
+      lib.nameValuePair "sketchybar/plugins/${name}.sh" {
+        source = getSource {
+          name = "sketchybar-plugin-${name}";
+          arg = value;
+        };
+      }
+    ) set;
+in
+{
+  meta.maintainers = with lib.hm.maintainers; [ aguirre-matteo ];
+
+  options.programs.sketchybar = {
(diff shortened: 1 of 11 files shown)


modules/programs/sketchybar.nix
```diff
@@ -0,0 +1,86 @@
...
+    };
+  };
+
+  config = mkIf cfg.enable {
+    assertions = [
+      (lib.hm.assertions.assertPlatform "programs.sketchybar" pkgs lib.platforms.darwin)
+    ];
+
+    home.packages = mkIf (cfg.package != null) [ cfg.package ];
+    xdg.configFile = (getSketchybarrc cfg.sketchybarrc) // (getPlugins cfg.plugins);
```
awwpotato, line 84: 1. this should be using mkMerge to ensure there aren't collisions
2. both of these functions are only used once, imo it would be more readable to inline instead of having their definition and usage on opposite ends of the file

modules/programs/sketchybar.nix
```diff
@@ -0,0 +1,86 @@
...
+    setting:
+    if setting != "" then
+      {
+        "sketchybar/sketchybarrc".source = getSource {
+          name = "sketchybarrc";
+          arg = setting;
+        };
+      }
+    else
+      { };
```
awwpotato, line 30:
```suggestion
    mkIf setting != "" {
      "sketchybar/sketchybarrc".source = getSource 
        name = "sketchybarrc"
        arg = setting;
      };
    };
```

modules/programs/sketchybar.nix
```diff
@@ -0,0 +1,86 @@
...
+
+        sketchybar --bar position=top height=40 blur_radius=30 color=0x40000000
+      '';
+      description = ''
+        Script to be written to the sketchybarrc. All the details about its contents
+        can be found here: <https://felixkratz.github.io/SketchyBar/config/bar>.
+      '';
+    };
+    plugins = mkOption {
+      type = with types; attrsOf (either lines path);
```
awwpotato, line 64: merging shell script is probably a bad idea
```suggestion
      type = with types; attrsOf (either str path);
```

khaneliman:
I have a branch that i've been working on this for a bit, just haven't gotten back to it. I wanted to make it flexibly support bash or lua configs.

PR #7051, opened 2025-05-13
nix-darwin: use `launchctl asuser` to get correct Mach bootstrap context

Enzime:
### Description

This is important on multi-user systems for example:

```
{
  home-manager.users.user1.home.activation.setDefaultBrowser = lib.hm.dag.entryAfter [ "writeBoundary" ] ''
    ${lib.getExe pkgs.defaultbrowser} firefox
  '';

  home-manager.users.user2.home.activation.setDefaultBrowser = lib.hm.dag.entryAfter [ "writeBoundary" ] ''
    ${lib.getExe pkgs.defaultbrowser} chrome
  '';
}
```

If I run activation as `user1` then `user2`'s `defaultbrowser chrome` command will fail to work without the `launchctl asuser`.

See: https://github.com/nix-darwin/nix-darwin/pull/1341#discussion_r1989887502

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@khaneliman

### DIFF
--- a/nix-darwin/default.nix
+++ b/nix-darwin/default.nix
@@ -19,7 +19,7 @@ in
       system.activationScripts.postActivation.text = lib.concatStringsSep "\n" (
         lib.mapAttrsToList (username: usercfg: ''
           echo Activating home-manager configuration for ${usercfg.home.username}
-          sudo -u ${usercfg.home.username} --set-home ${pkgs.writeShellScript "activation-${usercfg.home.username}" ''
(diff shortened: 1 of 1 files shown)


Enzime:
After some testing I've determined this PR doesn't actually depend on https://github.com/nix-darwin/nix-darwin/pull/1341 however it would be good to land it before that PR gets merged as any users who deploy using `ssh root@host darwin-rebuild switch` will not have the correct Mach bootstrap context when running their home-manager activation scripts

PR #7055, opened 2025-05-13
vscode: allow specifying paths for VSCode JSON settings

Drakota:
### Description

Enables users to provide paths to JSON files for VS Code settings, tasks, and keybindings. This allows for more flexible configuration management and reuse of existing configuration files instead of using inline configurations.

This allows for something like this:
```
let
  settingsPath = "${config.home.homeDirectory}/dotfiles/.vscode/settings.json";
  keybindingsPath = "${config.home.homeDirectory}/dotfiles/.vscode/keybindings.json";
  userTasksFile = "${config.home.homeDirectory}/dotfiles/.vscode/tasks.json";
in
{
  programs.vscode = {
    profiles = {
      default = {
        userSettings = config.lib.file.mkOutOfStoreSymlink settingsPath;
        keybindings = config.lib.file.mkOutOfStoreSymlink keybindingsPath;
        userTasks = config.lib.file.mkOutOfStoreSymlink userTasksFile;
      };
    };
  };
}
```

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/vscode.nix
```diff
@@ -81,10 +81,15 @@ let
       "extensions.autoCheckUpdates" = false;
     };
 
+  # This is a hack to determine if a variable is a path by coercing it to a path.
+  # It works for both derivation and normal paths. The `isList` check is because
+  # `tryEval` doesn't catch type errors.
+  isPath = p: if lib.isList p then false else (builtins.tryEval (builtins.toPath p)).success;
```
awwpotato, line 87: why not use `builtins.isPath p || lib.isStorePath p` like the other modules in home-manager? This current implementation seems to be causing the tests to fail.

Drakota: Thank you! I missed `isStorePath`, just pushed the changes

modules/programs/vscode.nix
```diff
@@ -95,11 +97,12 @@ let
         description = ''
           Configuration written to Visual Studio Code's
           {file}`settings.json`.
+          This can be a JSON object or a path to a custom JSON file.
```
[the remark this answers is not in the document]
Drakota: Hey, yeah I see your point, it's a bit redundant for basically the same info as the types. I thought it would warrant some mention in the description. I can remove it also if it's clear enough with the type hint.

PR #7056, opened 2025-05-13
formatter: remove script, add treefmt.toml + keep-sorted

awwpotato:
### Description
currently `./format` does a bunch of stuff with git that was only necessary for nixfmt but not nixfmt-tree, so I ~simplified it to just call treefmt directly~ deleted it (just use treefmt command directly instead). I also added keep sorted to keep the really long module lists sorted (already used in nixpkgs).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


awwpotato:
not sure why the tests are failing, it claims a file changed but when I run `nix fmt` or `./format` neither of them change anything. Maybe it's pulling the format stuff that is currently on master?

awwpotato:
so I decided to just delete `./format` because I changed it to just call treefmt which seemed kinda pointless.

modules/modules.nix
```diff
@@ -15,9 +15,10 @@ let
 
   modules =
     [
-      ./accounts/email.nix
+      # keep-sorted start case=no numeric=yes
```
khaneliman, line 18: Ooh I really like this!

format
```diff
@@ -1,35 +0,0 @@
-#! /usr/bin/env nix-shell
```
khaneliman, line 1: The only reason I left this around was because it seemed to be implemented for people who refused to use the flake formatter and might not have the package installed or know/want to do a dev shell. @teto @rycee any opinions ?

awwpotato: I don't think we really need to accommodate the use case of "I hate flake, dev shells and globally installing formatters", I don't know of any other nix projects that have a format script like this. From what I've seen they all just tell you to use `nix fmt`, their dev shell or just run whatever command yourself.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -21,7 +21,7 @@ Also make sure to read the guidelines found at
 
 - [ ] Change is backwards compatible.
 
-- [ ] Code formatted with `nix fmt` or `./format`.
+- [ ] Code formatted with `nix fmt` or `treefmt`.
```
khaneliman, line 24: Maybe give a more explicit example of the command to run if you don't have installed? Since `./format` was kinda a nice helper to just call something to do the job.

awwpotato: yeah we could but it is quite a lengthy command to be included in the pr template, I don't really have any strong feelings about.

khaneliman: Might not even need the `--config-file` part, i believe that's default behavior of the cli.

awwpotato: by default it will search up until finding a treefmt.toml, so it's not necessary.

octvs:
I bumped into this while working on another PR, wouldn't an update to [relevant docs](https://github.com/nix-community/home-manager/blob/8b55a6ac58b678199e5bba701aaff69e2b3281c0/docs/manual/contributing/guidelines.md#L237-L240) be necessary?

PR #7057, opened 2025-05-13
zsh: consider zsh.{profile,login,logout,env}Extra in prezto

thiagokokada:
### Description
Right now the `zsh.prezto` module ignores the contents of the `zsh.{profile,login,logout,env}Extra` options, so it means that if you try to set, e.g., `zsh.profileExtra = "something";` this option will be (silently) ignored.

This commit fixes another issue: since the main `zsh` module sets `home.file."${relToDotDir".zshenv"}".text` while `zsh.prezto` set the same file using `.source`, `zsh.prezto` would have priority so the environment variables from Home-Manager would not be loaded in non-NixOS systems. Now that we are using `.text` for both, the issue is fixed.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/zsh/prezto.nix
```diff
@@ -423,12 +424,30 @@ in
...
-        home.file."${relToDotDir ".zprofile"}".source = "${cfg.package}/share/zsh-prezto/runcoms/zprofile";
-        home.file."${relToDotDir ".zlogin"}".source = "${cfg.package}/share/zsh-prezto/runcoms/zlogin";
-        home.file."${relToDotDir ".zlogout"}".source = "${cfg.package}/share/zsh-prezto/runcoms/zlogout";
         home.packages = [ cfg.package ];
 
-        home.file."${relToDotDir ".zshenv"}".source = "${cfg.package}/share/zsh-prezto/runcoms/zshenv";
+        home.file."${relToDotDir ".zprofile"}".text = ''
+          # Generated by Nix
+          . ${cfg.package}/share/zsh-prezto/runcoms/zprofile
+          ${cfgZsh.profileExtra}
```
thiagokokada, line 432: Changing from `.source` to `.text` and using source to load the `zsh-prezto` files so we can avoid doing a IFD: https://github.com/nix-community/home-manager/pull/6905.

thiagokokada: BTW, https://github.com/nix-community/home-manager/pull/6905 is very likely the reason we had the second issue I described in the PR (the environment variables not loading anymore).

thiagokokada (commented):
CC @khaneliman for review.

thiagokokada:
Ok, should be good now. Tests are passing and there are no more IFDs in the code.

PR #7061, opened 2025-05-14
zsh: do not duplicate zsh.{profile,login,logout,env}Extra in prezto

thiagokokada:
### Description
In #7057 changed the definition of how the prezto files where loaded (from using `.source` to `.text`) and also manually add the text from `zsh.{profile,login,logout,env}Extra` files to the prezto definitions, but the last part was unnecessary and now we have duplicates inside the file.

This commit fixes it.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


thiagokokada:
CC @khaneliman, sorry :sweat_smile:.

Enzime (approved):
Tested this and it works 👍

Only other thing worth noting from my testing is that all the `*Extra` contents will be above the Prezto sourcing, which we might want the other way around

`{ programs.zsh.profileExtra = "profileExtra"; }`:

```
profileExtra
# Generated by Nix
source /nix/store/fbxjsrc5p85sh7la957sgp1z1q5sz9qx-zsh-prezto-0-unstable-2025-01-10/share/zsh-prezto/runcoms/zprofile
```

PR #7063, opened 2025-05-14
emacs: respect `defaultEditor` on Darwin

andre4ik3:
### Description

Currently on Darwin, `services.emacs.defaultEditor = true` isn't respected, as the variable is only configured on Linux.

This PR simply moves the variable out to a common block, so it's applied on all platforms.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@tadfisher


awwpotato (approved):
LGTM

PR #7066, opened 2025-05-15
Add initial support for borgmatic on darwin

DamienCassou:
### Description

On darwin, services.borgmatic should create a launchd agent.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/lib/default.nix
+++ b/modules/lib/default.nix
@@ -6,6 +6,7 @@ rec {
   assertions = import ./assertions.nix { inherit lib; };
 
   booleans = import ./booleans.nix { inherit lib; };
+  darwin = import ./darwin.nix { inherit lib; };
   deprecations = import ./deprecations.nix { inherit lib; };
   generators = import ./generators.nix { inherit lib; };
   gvariant = import ./gvariant.nix { inherit lib; };
--- a/modules/services/nix-gc.nix
+++ b/modules/services/nix-gc.nix
@@ -66,7 +66,7 @@ let
         ];
       };
     in
-    freq.${frequency};
+    freq.${frequency} or null;
 
   nixPackage =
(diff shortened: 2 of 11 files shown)


modules/lib/default.nix
```diff
@@ -6,6 +6,7 @@ rec {
   assertions = import ./assertions.nix { inherit lib; };
 
   booleans = import ./booleans.nix { inherit lib; };
+  darwin = import ./darwin.nix { inherit lib; };
```
khaneliman, line 9: Was wondering if this should be a top level uniqueness or if it makes sense to be an interval module with a `mkDarwin*` function instead

modules/services/nix-gc.nix
```diff
@@ -95,8 +36,7 @@ in
 
           On Linux this is a string as defined by {manpage}`systemd.time(7)`.
 
-          On Darwin it must be one of: ${toString darwinIntervals}, which are
-          implemented as defined in the manual page above.
+          $lib.hm.darwin.{intervalDocumentation}
```
ambroisie, line 39: Unless I'm missing something, this should be:
```suggestion
          ${lib.hm.darwin.intervalDocumentation}
```

khaneliman: Yeah, that's my bad for missing that...

DamienCassou: sorry for this. I will fix it.

DamienCassou: Fixed by https://github.com/nix-community/home-manager/pull/7071.

modules/lib/default.nix
DamienCassou: I don't understand what you mean here.

khaneliman: `lib.hm.interval` vs `lib.hm.darwin`

PR #7069, opened 2025-05-15
modules: auto import services and programs

awwpotato:
### Description
having this huge list is problematic for many reasons, it would be best to automatically import them. It might be best to put everything in it's own dir and auto import `default.nix` so modules that use multiple files don't need special treatment. further adopting the structure of `/modules/<category>/<name>/` for all modules would allow for auto importing without hard coding the "categories" into the import function.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


khaneliman:
thanks, i kept forgetting to come back and do this

awwpotato:
> thanks, i kept forgetting to come back and do this

Not sure how far I'm going to get on this, the merge conflicts are already piling up :sob:

awwpotato:
so I'm releasing implementing in my prefered method of searching each dir recursively until finding a default.nix would require restructuring the entire module dir. We should probably decide on a structure before I go implementing it. I would propose that the structure would reflect the modules prefix e.g. everything update `options.programs` would go in `/modules/programs`, everything under `options.xdg` would go in `/modules/xdg` etc.

MattSturgeon:
> searching each dir recursively until finding a default.nix

I agree, this feels a little _too_ automated. There are scenarios where a specific directory may wish to do things its own way.

I'd recommend having a list of "root" directories, each of which gets scanned for `«root»/«subdir»/default.nix` modules to import. This forms a nice middle ground.

Ideally this behaviour (and where the list of root dirs is defined) should be clearly documented somewhere, so the automatic behaviour isn't surprising.

awwpotato:
I'm currently busy with stuff in stylix, I should be able to work on this more after I've finished up some of the things there.

PR #7074, opened 2025-05-16
New option `home.file.<name>.mode` copies instead of symlinks

### LINKED ISSUE #3090: feature: add an option `home.file.<name>.mode` that allows copy instead of linking and permission setting.
### Description

This option should default to `"symlink"`, which instruct home-manager to create a symbolic link at the specified location to be compatible to the current behavior. When set to an octal-mode, e. g. `"0600"`, the content will be copied from the source to the specified location and `chmod` to the value given.

This issue generalize the goal of #1145, while adding the ability to set permissions and to align with the NixOS option `environment.etc.<name>.mode`.

Inspired by https://github.com/nix-community/home-manager/pull/1145#issuecomment-1175795593

zeorin:
### Description

Add an option `home.file.<name>.mode` that allows copy instead of linking and permission setting.

Fixes #3090.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee @khaneliman

### DIFF
--- a/modules/files.nix
+++ b/modules/files.nix
@@ -138,68 +138,16 @@ in
     # source and target generation.
     home.activation.linkGeneration = lib.hm.dag.entryAfter [ "writeBoundary" ] (
       let
-        link = pkgs.writeShellScript "link" ''
-          ${config.lib.bash.initHomeManagerLib}
+        link = pkgs.replaceVars ./files/link.sh {
+          inherit (config.lib.bash) initHomeManagerLib;
+        };
 
-          newGenFiles="$1"
-          shift
-          for sourcePath in "$@" ; do
-            relativePath="''${sourcePath#$newGenFiles/}"
-            targetPath="$HOME/$relativePath"
-            if [[ -e "$targetPath" && ! -L "$targetPath" && -n "$HOME_MANAGER_BACKUP_EXT" ]] ; then
-              # The target exists, back it up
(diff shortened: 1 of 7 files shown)


teto:
I wonder if it would be possible to achieve the same with systemd.tmpfiles ?

auscyber:
> I wonder if it would be possible to achieve the same with systemd.tmpfiles ?

it would still require something on darwin

mio-19:
why was modules/programs/browserpass.nix changed

mio-19:
It would be great if there is an option to set global default

AkechiShiro:
Is there still any way to see if this PR could be merged.

@zeorin : I still don't understand as well why `browserpass` has been modified, if you have any idea why it is changed in this PR?
There is also a merge conflict with this PR, could you please solve it ?

@rycee : I don't know if you have time to review this PR.

I believe the tool [putter]( https://git.sr.ht/~rycee/putter/) is still too experimental to use/aka replace this PR nowadays.

kr-nn:
I've been watching this for sometime. I have in the meantime built my own solution but it's not perfect. I recently had the pleasure of testing out [hjem](https://github.com/feel-co/hjem) which uses their own custom linker [smfh](https://github.com/feel-co/smfh)

hjem is a nixos module only but I do think we can use smfh as a better linker for home.file

PR #7078, opened 2025-05-17
qutebrowser: add support for per domain settings

### LINKED ISSUE #7077: feat: qutebrowser: support per domain settings
### Description

Allow users to set per-domain options for qutebrowser, as described in [1].

[1]: https://github.com/qutebrowser/qutebrowser/blob/bb7bbb6ead5b7412fa5828249d56e51c1b35cb5a/doc/help/configuring.asciidoc#per-domain-settings

octvs:
Add `programs.qutebrowser.perDomainSettings` which let's one to set configuration options for specific URLs [1].

It option doesn't check if the options passed to it are valid, it translates the config to python code to be written on the file as is. Mimicking the behaviour of `programs.qutebrowser.settings`.

[1]: https://github.com/qutebrowser/qutebrowser/blob/bb7bbb6ead5b7412fa5828249d56e51c1b35cb5a/doc/help/configuring.asciidoc#per-domain-settings

### Description

Resolves #7077.

I will update the test cases.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

  - ~~[ ] Added myself as module maintainer. See [example](https://github.com/nix-community/home-manager/blob/a51598236f23c89e59ee77eb8e0614358b0e896c/modules/programs/lesspipe.nix#L11).~~

#### Maintainer CC

_no maintainer is listed, pinging the author of the last non treeewide change_
@somasis


octvs:
Refactored it to merge previous and the new, per-domain, config setting method.

octvs:
Looks like the test failure is linked to #7079.

somasis:
> Notice, the feature uses new helper functions not the previous config writer formatLine, which writes options as c.<config> = <value>, whereas domain settings need config.set(<config>, <value>, <url>).

This actually sounds like an improvement anyway. I feel like I've ran into weird edge cases when using the `c.` variables in the past; at some point I just started using `config.set` in my `extraConfig` out of habit.

I was thinking maybe `perDomainSettings` might be a worse name compared to the `:set` equivalent `-u` / `--pattern` option names (`patternSettings` or something) but apparently that's what qutebrowser's own documentation uses for the feature name anyway.

looks good to me :)

somasis (approved):

somasis:
I like it! Plus, the change in dict handling I think should solve an issue I noticed some time ago with qutebrowser's config infrastructure that I was working around in my config.

Currently, doing this works
```nix
programs.qutebrowser.extraConfig = ''
  config.set("hints.padding", {"top":2,"bottom":2,"left":2,"right":2})
  config.set("statusbar.padding", {"top":14,"bottom":0,"left":2,"right":8})
  config.set("tabs.padding", {"top":6,"bottom":6,"left":6,"right":6})
'';
```

but

```nix
programs.qutebrowser.settings = {
  hints.padding = { top = 2; bottom = 2; left = 2; right = 2; };
  statusbar.padding = { top = 14; bottom = 0; left = 2; right = 8; };
  tabs.padding = { top = 6; bottom = 6; left = 6; right = 6; };
};
```

(which generates `c.hints.padding.top = 2` instead...) does not!

octvs:
> Currently, doing this works ...

Thanks for providing a solid use-case for what was itching my mind to refactor
the module.

> The change in dict handling contained in
> [13220f6](https://github.com/nix-community/home-manager/commit/13220f686c111a56c20b5b24d7a8c267b9587ea6)
> I think would solve an issue I noticed some time ago with qutebrowser's
> config infrastructure that I was working around in my config (I don't think
> it's currently reported though).

This becomes solve-able, but not as is. With the change we can set dictionaries
to options without "flattening"
(`hints.padding = { ... } ->
hints.padding.bottom = 2`) them. However this
behaviour is only controlled by the way they are called when we merge config strings
[1]. If with `flattenSettings` it would be the same outcome, just as in
options under `settings.*` for this PR.

Since some qutebrowser configuration options set as dicts (i.e. hints.padding
[2]) and some want to be "flattened" (i.e. colors.hints [3]), home-manager
module has to embody the knowledge of this distinction.

While I was refactoring I thought why some options, `aliases`, would be set as
a top level option. It could easily be set under `settings.aliases` given we
allow passing through attrsets to python dicts. Now it dawns to me, we can use
the `settings.*` attrset to separate what needs to be straightened or not.

> I like it! Maybe it could be split into another PR, but I think an additional
> commit sounds fine to me.

At this point it definitely makes sense to address this on a separate PR.

[1]: https://github.com/nix-community/home-manager/blob/13220f686c111a56c20b5b24d7a8c267b9587ea6/modules/programs/qutebrowser.nix#L319-L343
[2]: https://github.com/qutebrowser/qutebrowser/blob/main/doc/help/settings.asciidoc#hintspadding
[3]: https://github.com/qutebrowser/qutebrowser/blob/main/doc/help/settings.asciidoc#colorshintsbg

khaneliman:
needs rebase to fix CI

PR #7086, opened 2025-05-18
aerospace: auto-reload config on update

phucleeuwu:
### Description

Configure Aerospace to automatically reload its configuration when the TOML file changes.

This uses the `onChange` option in `home.file` to run `aerospace reload-config` via the executable path from `cfg.package`. Ensures that any changes to user settings in `aerospace.toml` are immediately applied without requiring manual reload.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/aerospace.nix
+++ b/modules/programs/aerospace.nix
@@ -269,9 +269,10 @@ in
 
     home = {
       packages = lib.mkIf (cfg.package != null) [ cfg.package ];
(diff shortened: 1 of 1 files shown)


leon-erd:
@damidoug :)

damidoug:
I never had this problem when launchd and KeepAlive are enabled, if we try to run `reload-config` without aerospace is running we got a error, i recommend add a way to valid if aerospace is running before run that command.

PR #7087, opened 2025-05-18
blesh: add module

ajhalili2006:
### Description

Another attempt at https://github.com/nix-community/home-manager/pull/3238, although I will probably need help implementing `programs.blesh.attach = "manual` through conditional statement trickery due to being inexperienced with the rest of Nix language over than the basics (like `let...in` keywords and imports).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


ajhalili2006:
I haven't added this patch to the news entry yet until https://github.com/nix-community/home-manager/pull/3238#issuecomment-1946053896 (technically https://github.com/nix-community/home-manager/pull/3238#issuecomment-1946005014) is implemented for obvious reasons, alongside some manual testing on my side.

modules/programs/blesh.nix
```diff
@@ -0,0 +1,159 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+with lib;
```
awwpotato, line 8: use of `with lib` is bad pratice, instead use inherit .e.g
```nix
let
  inherit (lib)
    mkEnableOption
    mkOption
    types
    ;
in
```

ajhalili2006: Thanks for the note, I have updated that (also rebased against the master branch to keep things in sync).

PR #7107, opened 2025-05-21
shell: separate posix-shells from bash

### LINKED ISSUE #5474: bug: `home.sessionVariables` and `bash.sessionVariables` end up serving the same purpose: `.profile`

### LINKED ISSUE #6412: bug: `home.sessionVariables` is missing in any shell other than activated ones.

### Issue description

`~/.profile` is only generated when `programs.bash.enable` is true. If this is not true, then it's not genreated. This isn't always a problem, but it is when other programs use `sh -c` to run their commands, since these variables are not set.

This may seem like an intended outcome, but there's programs such as `bemenu` which require these variables to be set in order to apply any sort of theming.

Possible solution:
When `programs.bash.enable` is not enabled, generate a simple `.profile` which only imports `hm-session-vars.sh`.

Maybe aliases too??

### Maintainer CC

_No response_

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.13.1, NixOS, 25.05 (Warbler), 25.05.20250204.799ba5b`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.24.12`
 - channels(root): `"nixos-23.05"`
 - nixpkgs: `/nix/store/k8nkf470zpidpa5nh76lh2x6rxfzpwa4-source`
```

Noodlez1232:
### Description

Creates a posix-compliant `.profile` and `.shinit` usable from any posix-compliant shell. This is great for runtime scripts to have their variables set without having to import `programs.enable.bash`

For backwards compatibility, it is enabled by default only if `programs.enable.bash` is enabled.

This commit also changes bash's `.profile` to be `.bash_profile`, once again, to keep them separate, since bash is not posix compliant, its profile should not be included in the normal `.profile`.

fixes: https://github.com/nix-community/home-manager/issues/5474
fixes: https://github.com/nix-community/home-manager/issues/6412

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/misc/news/2026/05/2026-05-16_12-54-11.nix
+++ b/modules/misc/news/2026/05/2026-05-16_12-54-11.nix
@@ -1,0 +1,13 @@
+{
+  time = "2026-05-16T19:54:11+00:00";
+  condition = true;
+  message = ''
+    A new module is available: 'programs.sh'. 'programs.sh' will control all POSIX compatible shells (e.g. dash, bash, zsh, etc.).
+
+    For now, 'programs.sh' is enabled by default if 'programs.bash' is enabled for backwards compatibilty.
+
+    This also means a breaking change: bash's configuration is now stored in '~/.bash-profile' and '~/.bashrc' *only*. Since 'programs.sh' is enabled by default if bash is enabled, this should not be a problem, but is something to be aware of.
+
+    If you were using 'programs.bash' to configure the POSIX shells, this you should move that configuration into 'programs.sh'.
+  '';
+}
--- a/modules/misc/ssh-auth-sock.nix
+++ b/modules/misc/ssh-auth-sock.nix
@@ -11,16 +11,16 @@ let
     };
 
   initSubmodule = {
-    options.bash = mkShellInitOption "bash";
+    options.sh = mkShellInitOption "sh";
     options.fish = mkShellInitOption "fish";
     options.nushell = mkShellInitOption "nushell";
   };
 
   # Preserve $SSH_AUTH_SOCK only if it stems from a forwarded agent which
   # is the case if both $SSH_AUTH_SOCK and $SSH_CONNECTION are set.
-  bashIntegration = ''
+  shIntegration = ''
     if [ -z "$SSH_AUTH_SOCK" -o -z "$SSH_CONNECTION" ]; then
-      ${cfg.initialization.bash}
+      ${cfg.initialization.sh}
     fi
   '';
   fishIntegration = ''
@@ -54,7 +54,7 @@ in
       proper {option}`programs.(bash|fish|nushell|zsh).*` options.
     '';
     example = {
-      bash = "export SSH_AUTH_SOCK=$HOME/.ssh/agent.sock";
+      sh = "export SSH_AUTH_SOCK=$HOME/.ssh/agent.sock";
       fish = "set -x SSH_AUTH_SOCK $HOME/.ssh/agent.sock";
       nushell = "$env.SSH_AUTH_SOCK = $HOME/.ssh/agent.sock";
     };
@@ -65,9 +65,10 @@ in
 
   config = lib.mkIf (cfg.initialization != null) {
     # $SSH_AUTH_SOCK has to be set early since other tools rely on it
-    programs.bash.profileExtra = lib.mkOrder 900 bashIntegration;
+    programs.sh.profileExtra = lib.mkOrder 900 shIntegration;
+    programs.bash.profileExtra = lib.mkOrder 900 shIntegration;
     programs.fish.shellInit = lib.mkOrder 900 fishIntegration;
     programs.nushell.extraConfig = lib.mkOrder 900 nushellIntegration;
-    programs.zsh.envExtra = lib.mkOrder 900 bashIntegration;
+    programs.zsh.envExtra = lib.mkOrder 900 shIntegration;
   };
 }
--- a/modules/programs/bash.nix
+++ b/modules/programs/bash.nix
@@ -249,8 +249,11 @@ in
       home.packages = lib.mkIf (cfg.package != null) [ cfg.package ];
 
       home.file.".bash_profile".source = writeBashScript "bash_profile" ''
-        # include .profile if it exists
-        [[ -f ~/.profile ]] && . ~/.profile
+        . "${config.home.sessionVariablesPackage}/etc/profile.d/hm-session-vars.sh"
+
+        ${sessionVarsStr}
+
+        ${cfg.profileExtra}
 
         # include .bashrc if it exists
         [[ -f ~/.bashrc ]] && . ~/.bashrc
@@ -266,15 +269,6 @@ in
           fi
         ''
       );
-
-      home.file.".profile".source = writeBashScript "profile" ''
-        . "${config.home.sessionVariablesPackage}/etc/profile.d/hm-session-vars.sh"
-
-        ${sessionVarsStr}
-
-        ${cfg.profileExtra}
-      '';
-
       home.file.".bashrc".source = writeBashScript "bashrc" ''
         ${cfg.bashrcExtra}
 
--- a/modules/programs/sh.nix
+++ b/modules/programs/sh.nix
@@ -1,0 +1,140 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  inherit (lib)
+    mkIf
+    mkOption
+    optionalAttrs
+    types
+    ;
+
+  cfg = config.programs.sh;
+
+  writePosixScript =
+    name: text:
+    pkgs.writeTextFile {
+      inherit name text;
+      checkPhase = ''
+        ${pkgs.stdenv.shellDryRun} "$target"
+      '';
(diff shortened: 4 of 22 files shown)


Noodlez1232:
This *should* be ready to go, but there is a couple questions I have about it:

1. Is the import of `.bashrc` required in `.bash_profile`
2. Should this be default enabled for stateVersion 25.11? Right now it's only enabled by default if bash is enabled (for backwards compatibility)

Noodlez1232:
Oh and also I did test the POSIX side of it with two different shells:
1. Bash in posix mode (invoked using `sh`)
2. Dash, since it itself is a posix compliant shell.
And it worked perfectly for me without breaking actual bash support.

modules/modules.nix
```diff
@@ -266,6 +266,7 @@ let
       ./programs/senpai.nix
       ./programs/sesh.nix
       ./programs/sftpman.nix
+      ./programs/sh.nix
```
khaneliman: can be removed in rebase

tests/default.nix
```diff
@@ -290,6 +290,7 @@ import nmtSrc {
       ./modules/programs/senpai
       ./modules/programs/sesh
       ./modules/programs/sftpman
+      ./modules/programs/sh
```
khaneliman: can be removed in rebase

khaneliman:
Hmm aliases I've set don't seem to work in `sh` for me. If I manually source `.shinit` i get them to work.

Noodlez1232:
The merge conflicts are now fixed. ~~I haven't been able to replicate the alias problem at all either. It seems that maybe it's due to it not being a login shell?~~ The alias problem is due to the shell not being a login shell. This is intended POSIX shell behavior as `.shinit` does not run unless the shell is a login shell (much like .bashrc vs .bash_profile). If you want, I can change that behavior.

modules/misc/ssh-auth-sock.nix
```diff
@@ -11,13 +11,19 @@ let
     };
 
   initSubmodule = {
+    options.sh = mkShellInitOption "sh";
```
bmrips, line 14: Given that on every occurence, `sshAuthSock.initialization.sh` is the same as `sshAuthSock.initialization.bash`, I suggest to remove the Bash option entirely to reduce code duplication.

bmrips (changes requested):

modules/misc/ssh-auth-sock.nix
Noodlez1232: bash and sh create different `.profile` options (`.profile` and `.bash_profile`). Also they aren't necessarily compatible shells which is why I left it that way. It's why I split it in every occurance (imagine somebody doing `programs.bash.enable = true` but `programs.sh.enable = lib.mkForce false`).

If you still think it's necessary, however, I'll do so.

bmrips: I meant to remove the Bash option while still setting `programs.bash.profileExtra` to the sh initialization code. Subsequently, `sshAuthSock.initialization.sh`  would influence sh, Bash, and Zsh which is fine in my opinion since these shells do not need different code up until now.

modules/programs/sh.nix
```diff
@@ -0,0 +1,140 @@
...
+              int
+              float
+              bool
+            ])
+          );
+        example = {
+          MAILCHECK = 30;
+        };
+        description = ''
+          Environment variables that will be set for the Bash session.
```
khaneliman:
```suggestion
          Environment variables that will be set for the POSIX shell session.
```

khaneliman:
Could you add a news entry for the new `programs.sh` module?

modules/misc/ssh-auth-sock.nix
khaneliman: Yeah, that makes sense for this module. Since the current snippets are identical across sh/bash/zsh, using one POSIX-compatible initialization field for those shells seems fine and keeps `sshAuthSock` simpler.

Noodlez1232:
> Could you add a news entry for the new `programs.sh` module?

I went ahead and did that but I'd like to request that you look at that a bit extra. I'm bad at release notes and news stuff, so it might not be up to snuff.

Noodlez1232:
I'm not quite sure why the format step is failing. I ran `nix fmt` locally, and no changes occur, even after rebasing on master.

bmrips:
> I'm not quite sure why the format step is failing. I ran `nix fmt` locally, and no changes occur, even after rebasing on master.

It is failing because the formatter modifies the news entry.

Btw, this is just a nit, but I suggest making the news entry more concise by removing the unused lambda argument (`_:`) and the example comments.

bmrips (approved):
LGTM for the `sshAuthSock`, `services.ssh-agent`, and `ssh-tpm-agent` modules.

Noodlez1232:
Went ahead and changed the news file as well.

modules/misc/news/2026/05/2026-05-16_12-54-11.nix
```diff
@@ -0,0 +1,13 @@
+{
+  time = "2026-05-16T19:54:11+00:00";
+  condition = true;
+  message = ''
+    A new module is available: 'programs.sh'. 'programs.sh' will control all POSIX compatible shells (e.g. dash, bash, zsh, etc.).
+
+    For now, 'programs.sh' is enabled by default if 'programs.bash' is enabled for backwards compatibilty.
+
+    This also means a breaking change: bash's configuration is now stored in '~/.bash-profile' and '~/.bashrc' *only*. Since 'programs.sh' is enabled by default if bash is enabled, this should not be a problem, but is something to be aware of.
```
khaneliman, line 9: `~/.bash-profile` should be `~/.bash_profile` to match the generated file.

```suggestion
    This also means a breaking change: bash's configuration is now stored in '~/.bash_profile' and '~/.bashrc' *only*. Since 'programs.sh' is enabled by default if bash is enabled, this should not be a problem, but is something to be aware of.
```

modules/programs/bash.nix
```diff
@@ -249,8 +249,11 @@ in
       home.packages = lib.mkIf (cfg.package != null) [ cfg.package ];
 
       home.file.".bash_profile".source = writeBashScript "bash_profile" ''
-        # include .profile if it exists
-        [[ -f ~/.profile ]] && . ~/.profile
+        . "${config.home.sessionVariablesPackage}/etc/profile.d/hm-session-vars.sh"
+
```
khaneliman, line 253: This likely explains the alias issue from earlier. POSIX `sh` read `$ENV` for interactive shells, so `sh -i` does not need to be a login shell, but `$ENV` must already be exported by the parent login environment.

With this PR, `programs.sh.enable` defaults to `programs.bash.enable`, but bash login shells read `.bash_profile`, not `.profile`. Since `programs.sh` only exports `ENV` from `.profile`, interactive `sh` sessions started from a bash login never source `.shinit`.

I tested this with a NixOS VM integration test that logs in as a bash user, runs `sh -ic hm-sh-env-check`, and waits for an alias from `programs.sh.initExtra` to create a marker file. It fails on this branch because `.shinit` is not loaded. Exporting `ENV` from `.bash_profile` when `programs.sh.enable` is true makes the VM test pass.

This feels like the least confusing/compliant behavior to me: bash does not source POSIX `.profile`, but the real login environment still exports the standard POSIX `ENV` variable so interactive `sh` can source `.shinit`. Interested in hearing other opinions on that, though.

```suggestion
        ${lib.optionalString config.programs.sh.enable ''
          ENV="$HOME/.shinit"; export ENV
        ''}

```

Noodlez1232:
Hi. I'm working a lot these days. This is still on my mind, just been hard to find time to work on. Not dead, just dormant.

PR #7110, opened 2025-05-22
podman: set switch method to "restart"

averyanalex:
### Description

Restart quadlet containers on switch instead of stop-start.

Reason:

I have containers A, B and container B has

```nix
Unit = rec {
  Requires = [
    "podman-A.service"
  ];
  After = Requires;
}
```

in it's config.

When I update A configuration, switch stops A, systemd automatically stops B, switch starts A but not B. B is stopped now.

Setting `X-SwitchMethod = "restart"` fixes it.

Also it may worth it to set SwitchMethod in other units like networks.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/podman-linux/containers.nix
+++ b/modules/services/podman-linux/containers.nix
@@ -136,6 +136,7 @@ let
               else
                 "Service for container ${name}"
             );
+            X-SwitchMethod = "restart";
           };
         } containerDef.extraConfig
       );


khaneliman:
@bamhm182 @n-hass

n-hass:
Excuse my ignorance, but won't this cause unnecessary restarts for all containers on every switch? That would be an undesirable default. However I'm not 100% confident on when `X-SwitchMethod` applies.

AveryanAlex:
> Excuse my ignorance, but won't this cause unnecessary restarts for all containers on every switch? That would be an undesirable default. However I'm not 100% confident on when `X-SwitchMethod` applies. 

Oh, I forgot to answer. No, it just changes default "stop-start" method to "restart" on container config changes.

AveryanAlex:
> In addition to that, I believe that the following home manager config resolves the issue as well.
> 
> ```
> config.systemd.user.startServices = "sd-switch";
> ```

It's already default value

PR #7111, opened 2025-05-22
meli: adding the meli email client

MunsMan:
Adding [Meli email client](https://meli-email.org/) with integrated into the `accounts.email.accounts` configuration. 

### Description

As this begin my first contribution, I added myself as a maintainer.
Furthermore, I added a program option for `meli` with settings. 
Missing is currently the raw `extraConfig`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/meli.nix
+++ b/modules/programs/meli.nix
@@ -1,0 +1,222 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+let
+  inherit (lib)
+    mkEnableOption
+    mkOption
+    types
+    mkIf
+    ;
+
+  enabledAccounts = lib.attrsets.filterAttrs (
+    name: value: value.meli.enable or false
+  ) config.accounts.email.accounts;
+
+  meliAccounts = (lib.attrsets.mapAttrs (name: value: (mkMeliAccounts name value)) enabledAccounts);
+
+  mkMeliAccounts = (
+    name: account: {
+      root_mailbox = "${config.accounts.email.maildirBasePath}/${account.maildir.path}";
+      format = "Maildir";
+      identity = "${account.address}";
+      display_name = "${account.realName}";
+      subscribed_mailboxes = account.meli.mailboxes;
+      send_mail = mkSmtp account;
+      mailboxes = account.meli.mailboxAliases;
+    }
+  );
+
+  mkSmtp = account: {
+    hostname = account.smtp.host;
+    port = account.smtp.port;
+    auth = {
+      type = "auto";
+      username = account.userName;
+      password = {
+        type = "command_eval";
+        value = lib.strings.concatStringsSep " " account.passwordCommand;
+      };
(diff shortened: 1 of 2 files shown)


modules/programs/meli.nix
```diff
@@ -0,0 +1,222 @@
...
+        name: account:
+        lib.optional (account.meli.enable or false) {
+          assertion = account ? smtp;
+          message = "The email account '${name}' must have an 'smtp' attribute defined when enabling meli.";
+        }
+      ) config.accounts.email.accounts
+    );
+
+    # Generate meli configuration from email accounts
+    home.file.".config/meli/config.toml".source = (pkgs.formats.toml { }).generate "meli-config" (
```
cardboardcpu, line 204: Shouldn't `xdg.configFile` be used here instead? Feels more HM idiomatic and does not assume `.config` as the `$XDG_CONFIG_HOME`.

cardboardcpu:
Nitpicking a bit. (:
I am not an authoritative source here tho.

modules/programs/meli.nix
MunsMan: Yes, you are right. Thank for the feedback, I will fix that.

MunsMan:
Thank, I'm looking for feedback as it is my first home-manager module.

cardboardcpu (approved):
Looks good. Can't wait to try it once I get home.

modules/programs/meli.nix
```diff
@@ -0,0 +1,222 @@
...
+      };
+
+      extraConfig = mkOption {
+        type = types.lines;
+        default = "";
+        description = "Additional lines to add to meli config.toml";
+      };
+      settings = mkOption {
+        type = types.submodule {
+          options = {
```
khaneliman, line 82:
```suggestion
          freeformType = (pkgs.formats.toml { }).type;
          options = {
```
We're using a toml generator so we should be able to just accept freeform entry for the user.

modules/programs/meli.nix
```diff
@@ -0,0 +1,222 @@
...
+          shortcuts = {
+            general = config.programs.meli.settings.shortcuts.general;
+            composing = config.programs.meli.settings.shortcuts.composing;
+            contact-list = config.programs.meli.settings.shortcuts.contact-list;
+            listing = config.programs.meli.settings.shortcuts.listing;
+            pager = config.programs.meli.settings.shortcuts.pager;
+          };
+        }
+    );
+  };
```
khaneliman, line 221:
```suggestion
    xdg.configFile."meli/config.toml".source = (pkgs.formats.toml { }).generate "meli-config" (
      {
        accounts = meliAccounts;
      }
      // config.programs.meli.settings
    );
```

Something like this would allow more freeform support.

khaneliman (commented):
Could also benefit from a unit test to prove we're generating what we expect.

PR #7114, opened 2025-05-23
(darwin) fix app linking

lylythechosenone:
### Description

Link the `Contents` subfolder of each individual app rather than a folder containing all apps, thereby enabling the apps to appear in Spotlight and Launchpad.

I'm not a nix master, so I fully expect at least some of this to be the wrong way to do it. Please correct me if it is.

### Checklist

- [x] Change is backwards compatible. — all options are the same. The file structure is the same, but some files have become links, and some links have become files.

    or `nix build --reference-lock-file flake.lock ./tests#test-all` using Flakes. — tested to work for its intended purpose. I don't currently have the capacity to run all tests.

- [ ] Commit messages are formatted like — I got close; I'm not sure how to change the message from gitweb, sadly.

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/targets/darwin/linkapps.nix
+++ b/modules/targets/darwin/linkapps.nix
@@ -23,14 +23,26 @@ in
 
   config = lib.mkIf (pkgs.stdenv.hostPlatform.isDarwin && cfg.linkApps.enable) {
     # Install MacOS applications to the user environment.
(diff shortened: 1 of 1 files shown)


lylythechosenone:
Actually, I'm rescinding this for now, as it seems to have broken after a reboot/update (not sure why). I will investigate.

Issue #5997, opened 2024-10-23
bug: services.gpg-agent.enableSshSupport does not work on macOS


shivangswain:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Issue description

I believe `SSH_AUTH_SOCK` is only set by home-manager if it is empty, which is not the case in macOS by default. On my machine, it was set to:
```
> echo $SSH_AUTH_SOCK
/private/tmp/com.apple.launchd.ksw57i6cJR/Listeners
```

This causes:
```
> ssh-add -L
The agent has no identities.
```

To fix this, I had to add the following to my `home.nix`:
```
programs.zsh = {
  enable = true;
  initExtra = ''
    gpgconf --launch gpg-agent
  '';
  sessionVariables = {
    SSH_AUTH_SOCK = "$(gpgconf --list-dirs agent-ssh-socket)";
  };
};
```
<sub>Semi-linked to #3263 but created a new issue because I am not sure if it is an issue in Linux anymore. Also, that had gone stale.</sub>

<sub>Kudos to https://code.dumpstack.io/infra/localhost, specifically https://code.dumpstack.io/infra/localhost/commit/f44a829461b772654841e10203d79320c62635a6. Helped me fix this.</sub>

### Maintainer CC

@rycee (maybe?)

### System information

```markdown
- system: `"aarch64-darwin"`
 - host os: `Darwin 24.0.0, macOS 15.0.1`
 - multi-user?: `yes`
 - sandbox: `no`
 - version: `nix-env (Nix) 2.24.9`
 - channels(root): `"nixpkgs"`
 - nixpkgs: `/nix/store/bc6afipyc33jj5iwp10acc848jj3fswr-source`
```

rycee:
Sorry, don't think I can help much. I did merge https://github.com/nix-community/home-manager/pull/6001 a little while ago and perhaps that may help?

shivangswain:
Thank you for your insight. I hope somebody else is able to figure this out.

folliehiyuki:
I believe https://github.com/nix-community/home-manager/pull/5901 should fix your issue.

khaneliman:
Should be resolved by https://github.com/nix-community/home-manager/pull/7117 reopen if not

PR #7117, opened 2025-05-23
fix(service/gpg-agent): ensure SSH_AUTH_SOCK is set on Darwin

squat:
This commit addresses https://github.com/nix-community/home-manager/issues/5997
At the root, the gpg-agent SSH integration is not working on Darwin
because shells are started with `SSH_AUTH_SOCK` already set to something
like `/private/tmp/com.apple.launchd.ozLmoURHDC/Listeners`, which
prevents the hook from setting the variable to the gpg-agent's socket.

This commit fixes the issue using a different mechanism to detect if the
current shell has already had the `SSH_AUTH_SOCK` variable set to the
gpg-agent's socket, namely by checking for `gnupg_SSH_AUTH_SOCK_by` as
documented in
https://www.gnupg.org/documentation/manuals/gnupg/Agent-Examples.html#Agent-Examples.

This is essentially a simplified version of
https://github.com/nix-community/home-manager/pull/5901 that does not
attempt to migrate gpg-agent over to launchd.

Signed-off-by: squat <lserven@gmail.com>

### DIFF
--- a/modules/services/gpg-agent.nix
+++ b/modules/services/gpg-agent.nix
@@ -371,7 +371,8 @@ in
         );
 
         home.sessionVariablesExtra = optionalString cfg.enableSshSupport ''
-          if [ -z "$SSH_AUTH_SOCK" ]; then
+          unset SSH_AGENT_PID
+          if [ "''${gnupg_SSH_AUTH_SOCK_by:-0}" -ne $$ ]; then
             export SSH_AUTH_SOCK="$(${gpgPkg}/bin/gpgconf --list-dirs agent-ssh-socket)"
           fi
         '';


squat:
cc @rycee since you're listed as a maintainer for this module. PTAL if you can <3

squat:
cc @cmacrae @folliehiyuki since you reviewed #5901

squat:
PTAL @khaneliman
I'm looping in more maintainers since this PR has been waiting for review for a bit now.

khaneliman:
Change makes sense to me, but I don't use to confirm it doesn't break anything. We can try it out.

euank:
Hey, just wanted to give a heads up that this broke my ssh setup on linux.

I have a laptop and desktop, both of which have yubikeys in them and `services.gpg-agent.enableSshSupport = true`.

For my desktop, when I sit in front of it I want `SSH_AUTH_SOCK` to be `/run/user/1000/gnupg/S.gpg-agent.ssh` so I can use the yubikey in my desktop computer.

However, when I ssh into that from my laptop with the ssh `ForwardAgent` option set, I want `SSH_AUTH_SOCK` to be `/tmp/ssh-XXXX/agent.$number` (the forwarded agent from my laptop).

If I'm at my laptop, I obviously can't touch the yubikey inserted in my desktop, so I really want it to respect the `SSH_AUTH_SOCK` variable that `ssh` has set.

Before this PR, the above setup worked fine. When `ssh` set the auth sock to a forwarded agent, zsh left it alone due to the `-z` check... but after this PR, when I ssh in, the `SSH_AUTH_SOCK` values gets clobbered and I can't then use ssh (like `git pull` or whatever) without manually finding the socket in `/tmp/` and re-exporting the variable. Or disabling `enableSshSupport` when I want to use the machine remotely, and then re-enabling it when I'm sitting in front of it.

I feel like my use-case isn't all that unusual, so I kinda expect this'll break some other people too.
Maybe we could do a more targeted thing, and make the check something like `if [[ "$SSH_AUTH_SOCK" =~ *com.apple.launchd* ]]`?
Or otherwise figure out that it was set by launchd?

squat:
@euank that's a good point. Using GPG's suggested/example shell integration effectively breaks the default behavior of `ssh -A`.
In order for ssh-agent forwarding to work when setting `SSH_AUTH_SOCK` in this way, you now have to do something like:
```shell
ssh -R /run/user/1000/gnupg/S.gpg-agent.ssh:/run/user/1000/gnupg/S.gpg-agent.ssh <host>
```
This works well enough, it's just: more cumbersome, more brittle, and less documented. This configuration is mentioned, for example, in the [Arch Wiki](https://wiki.archlinux.org/title/GnuPG#Forwarding_gpg-agent_and_ssh-agent_to_remote) and tellingly, there is an open discussion that asks:
> What about setting `ForwardAgent yes`

I'm not against swapping out the change for something targeted like what you propose. That might be a "good enough" solution for now.

squat:
Another way to create a targeted fix for this could be to add a test for when both `SSH_AUTH_SOCK` and `SSH_CONNECTION` are not null, in which case we skip setting `SSH_AUTH_SOCK`. I'll draft a quick PR for that.

benjamb:
Would these fixes be eligible for backporting to 25.05?

PR #7126, opened 2025-05-25
mergiraf: use mergiraf cli to generate gitattributes

drawbu:
Use `mergiraf` official supported languages list to generate git attributes instead of using a wildcard, as advised in the [official documentation](https://mergiraf.org/usage.html). 
We generate the git attributes, so we can't rely on testing, or they will break every so often

I'd like your opinion @bobvanderlinden 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/mergiraf.nix
+++ b/modules/programs/mergiraf.nix
@@ -13,6 +13,10 @@ let
     ;
   cfg = config.programs.mergiraf;
   mergiraf = "${cfg.package}/bin/mergiraf";
+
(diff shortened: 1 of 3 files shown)


drawbu:
Yeah, I don't know how I missed that from the original change you requested on the PR... I did not use a case where `*` did not suffice, but figured that since it was an explicit requirement in the doc, it was for a good reason. Since it is a change that already has been discussed in the prev PR, I should close it.

Thanks for the message, I guess I'll have to read more carefully next time!

PR #7127, opened 2025-05-25
misc: add librewolf native messaging hosts

nakoo:
### Description

Current native messaging hosts implementation do not support LibreWolf.

mkFirefoxModule links to `.mozilla/native-messaging-hosts` by default, but LibreWolf use their own paths for native messaging hosts, requiring users to create the symlink manually.[1] 

I tested this on Firefox, LibreWolf, and Floorp on my nixos machine but couldn't test darwin. Additionally, I did a bit of refactoring to reduce code duplication.

[1]: https://librewolf.net/docs/faq/#how-do-i-get-native-messaging-to-work

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@booxter
@rycee
@bmrips

### DIFF
--- a/modules/misc/mozilla-messaging-hosts.nix
+++ b/modules/misc/mozilla-messaging-hosts.nix
@@ -23,6 +23,12 @@ let
       "Library/Application Support/Mozilla/NativeMessagingHosts"
     else
       ".mozilla/native-messaging-hosts";
+
+  librewolfNativeMessagingHostsPath =
+    if isDarwin then
+      "Library/Application Support/LibreWolf/NativeMessagingHosts"
+    else
+      ".librewolf/native-messaging-hosts";
 in
 {
   meta.maintainers = with lib.maintainers; [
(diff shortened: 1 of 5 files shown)


bmrips:
LGTM. But let's wait for the reviews of the native messaging hosts infrastructure maintainers.

khaneliman:
Would be nice to see the new option separate from the refactor commit. Will wait for @booxter feedback on actual functionality.

kira-bruneau:
Thanks @nakoo! I tested out the changes and they work as expected!

Looking through the diff though I'm not sure if makes sense to add a special case to `mozilla-messaging-hosts.nix`. I wouldn't want to block merging if everyone else is ok with it (since it does fix the issue), but it seems to me there should be a better way to untangle this.

khaneliman:
> Thanks @nakoo! I tested out the changes and they work as expected!
> 
> Looking through the diff though I'm not sure if makes sense to add a special case to `mozilla-messaging-hosts.nix`. I wouldn't want to block merging if everyone else is ok with it (since it does fix the issue), but it seems to me there should be a way to untangle this.

I agree I don't like having to include a special case in there, but at least it keeps the ownership of the uniqueness to the module that deals explicitly with all the native messaging host logic.

booxter (approved):
It looks fine. I've checked that it doesn't break NMH for Firefox on Darwin. (Checked with `browserpass` extension.)

PR #7129, opened 2025-05-25
jellyfin-mpv-shim: add module

repparw:
### Description

Init of service jellyfin-mpv-shim, emulating plex-mpv-shim already provided.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/tests/modules/services/jellyfin-mpv-shim/default.nix
+++ b/tests/modules/services/jellyfin-mpv-shim/default.nix
@@ -1,0 +1,3 @@
+{
+  jellyfin-mpv-shim-example-settings = ./jellyfin-mpv-shim-example-settings.nix;
+}
--- a/tests/modules/services/jellyfin-mpv-shim/jellyfin-mpv-shim-example-settings.nix
+++ b/tests/modules/services/jellyfin-mpv-shim/jellyfin-mpv-shim-example-settings.nix
@@ -1,0 +1,39 @@
+{ ... }:
+{
+  services.jellyfin-mpv-shim = {
+    enable = true;
+
+    settings = {
+      allow_transcode_to_h265 = false;
+      always_transcode = false;
+      audio_output = "hdmi";
+      auto_play = true;
+      fullscreen = true;
+      player_name = "mpv-shim";
+    };
+
+    mpvBindings = {
+      WHEEL_UP = "seek 10";
+      WHEEL_DOWN = "seek -10";
+      "Alt+0" = "set window-scale 0.5";
+    };
+
+    mpvConfig = {
+      force-window = true;
+      ytdl-format = "bestvideo+bestaudio";
+      cache-default = 4000000;
+    };
+  };
+
+  nmt.script = ''
+    assertFileContent \
+       home-files/.config/jellyfin-mpv-shim/conf.json \
+       ${./jellyfin-mpv-shim-example-settings-expected-settings}
+    assertFileContent \
+       home-files/.config/jellyfin-mpv-shim/mpv.conf \
+       ${./jellyfin-mpv-shim-example-settings-expected-config}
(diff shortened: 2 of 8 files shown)


repparw:
maybe should use another name for mpv options? maybe mpvConfig mpvBindings? (analogous to mpv.config, mpv.bindings), also, I was using in my own config symlinks to the mpv config, maybe this should be same type as mpv.config/bindings?
So we can set them manually if they need to be different or set them like cfg.mpvConfig = programs.mpv.config ?

khaneliman:
Would be nice to see unit tests to verify generators are working, as expected. 

> maybe should use another name for mpv options? maybe mpvConfig mpvBindings? (analogous to mpv.config, mpv.bindings),

I was trying to think if there was a better naming scheme, too. Was thinking `<filename>Settings` or something...

repparw:
just took a quick look before bed, dunno if I should add the test somewhere or anything else. I'll see results tomorrow haha

tests/modules/services/jellyfin-mpv-shim/default.nix
```diff
@@ -0,0 +1,5 @@
+{
+  # Temporarily commented until fixed for recent changes in Nixpkgs.
+  jellyfin-mpv-shim-example-settings = ./jellyfin-mpv-shim-example-settings.nix;
+  jellyfin-mpv-shim-invalid-settings = ./jellyfin-mpv-shim-invalid-settings.nix;
```
khaneliman, line 4: Don't see a test for this
```suggestion
```

tests/modules/services/jellyfin-mpv-shim/default.nix
```diff
@@ -0,0 +1,5 @@
+{
+  # Temporarily commented until fixed for recent changes in Nixpkgs.
```
khaneliman, line 2:
```suggestion
```

tests/modules/services/jellyfin-mpv-shim/jellyfin-mpv-shim-example-settings.nix
```diff
@@ -0,0 +1,27 @@
+{ config, ... }:
+{
+  services.jellyfin-jellyfin-mpv-shim-shim = {
+    enable = true;
+
```
khaneliman, line 5: Would be nice to have something here for the conf.json too

YTG1234:
I don't know if anyone else has encountered this, but for me jellyfin-mpv-shim expects the config file to be writeable, which is really weird; the source code [opens it in read mode](https://github.com/jellyfin/jellyfin-mpv-shim/blob/8bee80d88ceb05975ac5d0500c277998d36fc020/jellyfin_mpv_shim/conf.py#L160).

So linking it to the read-only Nix store causes the config read to fail. In my own version of the module (before this one was upstream) I worked around it by copying the file:
```nix
    # jellyfin-mpv-shim expects, for some reason, conf.json to be writable.
    xdg.configFile."jellyfin-mpv-shim/conf.json" = mkIf (cfg.settings != { }) {
      source = jsonFormat.generate "conf.json" cfg.settings;
      target = "jellyfin-mpv-shim/.conf.json";
      onChange =
        let
          dir = "${config.xdg.configHome}/jellyfin-mpv-shim";
        in
        ''
          cp --dereference ${dir}/.conf.json ${dir}/conf.json
          chmod u+w ${dir}/conf.json
        '';
    };
```

repparw:
huh. didn't notice this, what happens? does the shim not launch, does it give a warning? so I can track this better. I don't have anything against just making a copy instead of symlink I guess, but would like to know what's happening

repparw:
oh, just seen the logs, yeah, i'm getting the same. would you PR the workaround?

PR #7138, opened 2025-05-27
way-displays: support stateful configuration file

mightyiam:
### Description

This is good for when displays are not considered part of the home configuration and are instead considered ephemeral.
Great for laptops but I use it for my desktop as well.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@jolars

### DIFF
--- a/modules/services/way-displays.nix
+++ b/modules/services/way-displays.nix
@@ -1,4 +1,5 @@
 {
+  options,
   config,
   lib,
   pkgs,
@@ -21,7 +22,10 @@ let
   yaml = pkgs.formats.yaml { };
 in
 {
-  meta.maintainers = [ maintainers.jolars ];
+  meta.maintainers = [
+    maintainers.jolars
+    maintainers.mightyiam
+  ];
 
   options.services.way-displays = {
     enable = mkEnableOption "way-displays";
@@ -29,7 +33,7 @@ in
     package = lib.mkPackageOption pkgs "way-displays" { };
 
     settings = mkOption {
-      type = yaml.type;
+      type = types.nullOr yaml.type;
       default = { };
       example = literalExpression ''
         {
@@ -61,6 +65,9 @@ in
         {file}`$XDG_CONFIG_HOME/way-displays/cfg.yml`. See
         <https://github.com/alex-courtis/way-displays/wiki/Configuration> for a
         description of available options.
+
+        When `null` a configuration file is not generated,
+        which allows way-displays to write its own configuration.
       '';
     };
 
@@ -79,14 +86,20 @@ in
       (lib.hm.assertions.assertPlatform "services.way-displays" pkgs lib.platforms.linux)
     ];
 
-    xdg.configFile."way-displays/cfg.yaml".source =
-      yaml.generate "way-displays-config.yaml"
-        (mergeSets [
-          {
-            CALLBACK_CMD = "${pkgs.libnotify}/bin/notify-send \"way-displays \${CALLBACK_LEVEL}\" \"\${CALLBACK_MSG}\"";
-          }
-          cfg.settings
-        ]);
+    home.packages = [ cfg.package ];
+
+    wayland.windowManager = lib.mapAttrs (name: _: {
+      systemd.variables = [ "XDG_VTNR" ];
+    }) options.wayland.windowManager;
+
+    xdg.configFile."way-displays/cfg.yaml" = mkIf (cfg.settings != null) {
+      source = yaml.generate "way-displays-config.yaml" (mergeSets [
+        {
+          CALLBACK_CMD = "${pkgs.libnotify}/bin/notify-send \"way-displays \${CALLBACK_LEVEL}\" \"\${CALLBACK_MSG}\"";
+        }
+        cfg.settings
+      ]);
+    };
 
     systemd.user.services.way-displays = {
       Unit = {


modules/services/way-displays.nix
```diff
@@ -14,22 +14,41 @@ let
     mkEnableOption
     mkIf
     mkOption
+    optionalAttrs
+    optionalString
     types
     ;
   mergeSets = sets: lists.fold attrsets.recursiveUpdate { } sets;
   cfg = config.services.way-displays;
+  serviceName = "way-displays" + optionalString cfg.statefulConfiguration "@";
```
mightyiam, line 23: Friendly tip: a `@` suffix makes this a unit template :shrug:

modules/services/way-displays.nix
```diff
@@ -77,35 +95,49 @@ in
...
 
-    xdg.configFile."way-displays/cfg.yaml".source =
-      yaml.generate "way-displays-config.yaml"
-        (mergeSets [
-          {
-            CALLBACK_CMD = "${pkgs.libnotify}/bin/notify-send \"way-displays \${CALLBACK_LEVEL}\" \"\${CALLBACK_MSG}\"";
-          }
-          cfg.settings
-        ]);
+    home.packages = [ cfg.package ];
```
mightyiam, line 108: The package being available in the environment is aligned with the stateful workflow but I think it can be useful for the static case as well so I did not make it conditional.

modules/services/way-displays.nix
```diff
@@ -77,35 +95,49 @@ in
...
+      ]);
+    };
+
+    systemd.user.services.${serviceName} = {
+      Unit =
+        {
+          Description = "Display configuration service";
+          Documentation = "man:way-displays(1)";
+          ConditionEnvironment = "WAYLAND_DISPLAY";
+          Requisite = cfg.systemdTarget;
```
mightyiam, line 125: `Requires` -> `Requisite`

I think that's more appropriate anyway because with `Requires` starting way-displays would trigger the dependency to start. That doesn't seem expected to me.

mightyiam (commented):
Alright. Tested the statefulConfiguration use case locally!

modules/services/way-displays.nix
```diff
@@ -14,22 +14,40 @@ let
...
 
+    statefulConfiguration = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        A configuration file is not generated.
+        Allows way-displays to write its own configuration.
+
+        The service is _not_ started automatically.
+        It can be started by running `systemctl --user start "way-displays@$XDG_VTNR".service`.
```
jolars, line 45: I'm not sure I understand. Why don't we want the service started?

jolars (commented):
I guess this could be useful, but see my question about auto-starting the service. 

Also, I'm not sure I understand what upside this has compared to just not creating a config file? Wouldn't that have much the same effect? (Except for not starting the service, which I don't understand why you would not want to do.)

modules/services/way-displays.nix
mightyiam: Outdated

modules/services/way-displays.nix
mightyiam: Now out of scope.

modules/services/way-displays.nix
mightyiam: Outdated

modules/services/way-displays.nix
```diff
@@ -79,14 +86,20 @@ in
...
-          {
-            CALLBACK_CMD = "${pkgs.libnotify}/bin/notify-send \"way-displays \${CALLBACK_LEVEL}\" \"\${CALLBACK_MSG}\"";
-          }
-          cfg.settings
-        ]);
+    home.packages = [ cfg.package ];
+
+    wayland.windowManager = lib.mapAttrs (name: _: {
+      systemd.variables = [ "XDG_VTNR" ];
+    }) options.wayland.windowManager;
```
mightyiam, line 93: I don't know if we're allowed to do such a thing but I'm willing to be so bold.

modules/services/way-displays.nix
```diff
@@ -79,14 +86,20 @@ in
...
+
+    xdg.configFile."way-displays/cfg.yaml" = mkIf (cfg.settings != null) {
+      source = yaml.generate "way-displays-config.yaml" (mergeSets [
+        {
+          CALLBACK_CMD = "${pkgs.libnotify}/bin/notify-send \"way-displays \${CALLBACK_LEVEL}\" \"\${CALLBACK_MSG}\"";
```
mightyiam, line 98: Losing this sucks. It's like... there should be both a configuration file and a state file.

modules/services/way-displays.nix
teto: what is XDG_VTNR ?

modules/services/way-displays.nix
teto: that's a request for way-displays I guess

modules/services/way-displays.nix
mightyiam: It can be found in `pam_systemd(8)`. [way-displays uses it](https://github.com/search?q=repo%3Aalex-courtis%2Fway-displays%20xdg_vtnr&type=code). The daemon seems to work without it defined, but then the client fails. So for the static configuration workflow it does not seem necessary but for the stateful one it is.

jolars (approved):
Okay, I've tried this out locally now and it seems to work as expected.

mightyiam:
Thank you!

terlar:
This change forces any `options.wayland.windowManager` to have a `systemd.variables` option. Perhaps it makes sense, but perhaps there are cases when that is not so? Perhaps we should check whether `systemd.variables` exists within the windowManager options before trying to set it.

mightyiam:
You made your own `windowManager` options set which doesn't have that and got an error? :sweat:

I wouldn't be opposed to such a filter.

terlar:
Yes, it was for Niri, the plan is to upstream it eventually. I will look into it a bit. Might be that the systemd stuff should be added there as well.

modules/services/way-displays.nix
```diff
@@ -79,14 +86,20 @@ in
...
-        (mergeSets [
-          {
-            CALLBACK_CMD = "${pkgs.libnotify}/bin/notify-send \"way-displays \${CALLBACK_LEVEL}\" \"\${CALLBACK_MSG}\"";
-          }
-          cfg.settings
-        ]);
+    home.packages = [ cfg.package ];
+
+    wayland.windowManager = lib.mapAttrs (name: _: {
+      systemd.variables = [ "XDG_VTNR" ];
```
teto, line 92: this overrides the default exported by sway which can cause problems

Issue #6424, opened 2025-02-09
Optional user defined script for error handling in case of home-manager backup file already existing


Doc-Steve:
### Description

There have been several proposals for **handling failures for backup of files stopping the home-manager activation**, because there is a need in special cases which can not be solved with e.g. forcing overwrite for a single specified file. Unfortunately it seems that there is no consensus how to deal with that error case: setting a global force-overwrite flag was considered too risky because something important could be deleted (by intended or even unintended behaviour).

My proposal is to give that decision to a **user defined optional error handling script** with an option like **home-manager.backupFileExistsHook**. This bash script should be executed in case of "backup file already exists" and two environment variables (e.g. ORIGINALFILE, BACKUPFILE) should be set. Then the user can define any strategy wanted (from renaming the backups with dates in the filname to moving with overwriting). If the script returns no error and made space for writing a new file in the place of "ORIGINALFILE" the home-manager activation can continue. If no script is set at all, same backward compatible behaviour of immediate failure.

cdo256:
I have created [a pull request](#7153) for this. I've been using this branch for 3 weeks setting `backupCommand` to `${pkgs.trash-cli}/bin/trash` with good success.

old4ever:
If you need the changes from [cdo256’s](https://github.com/cdo256) PR #7153 on the **`release‑25.05`** branch, I have created a fork of home-manager and merged that PR directly into the `release‑25.05` branch (instead of `master`).  
You can view the updated branch here:  

https://github.com/old4ever/home-manager-25.05-backupCommand/tree/release-25.05  

Thanks to [cdo256](https://github.com/cdo256) for the original work.

PR #7153, opened 2025-05-28
home-manager: add support for custom backup command (#6424)

cdo256:
### Description

This PR adds support for a custom backup command in Home Manager.  
A new `backupCommand` option is introduced, allowing users to specify a shell command to run on existing files that would otherwise block activation (e.g., moving files to trash, archiving, or custom backup logic). If set, this command is used in preference to the existing `backupFileExtension` mechanism.

- Adds `backupCommand` option to NixOS and nix-darwin modules.
- Exports `HOME_MANAGER_BACKUP_COMMAND` environment variable when set.
- Updates activation logic to invoke the custom backup command if provided.
- Updates collision checking and user-facing instructions to document the new option.

This enables more flexible and user-defined backup workflows during Home Manager activations.

This resolves [Optional user defined script for error handling in case of home-manager backup file already existing #6424](https://github.com/nix-community/home-manager/issues/6424).

---

### Checklist

  - Behaviour is only changed if the new option `backupCommand`, or the new environment variable `HOME_MANAGER_BACKUP_COMMAND` is used.
- [x] Code formatted.
- [x] Code tested.
  - Requires additional integration test for backing up files generally.
- [x] Commit messages formatted.

- If this PR adds a new module...
  . N/A

#### Maintainer CC

N/A

### DIFF
--- a/modules/files.nix
+++ b/modules/files.nix
@@ -146,10 +146,15 @@ in
           for sourcePath in "$@" ; do
             relativePath="''${sourcePath#$newGenFiles/}"
             targetPath="$HOME/$relativePath"
-            if [[ -e "$targetPath" && ! -L "$targetPath" && -n "$HOME_MANAGER_BACKUP_EXT" ]] ; then
-              # The target exists, back it up
-              backup="$targetPath.$HOME_MANAGER_BACKUP_EXT"
-              run mv $VERBOSE_ARG "$targetPath" "$backup" || errorEcho "Moving '$targetPath' failed!"
+            if [[ -e "$targetPath" && ! -L "$targetPath" ]] ; then
+              if [[ -n "$HOME_MANAGER_BACKUP_COMMAND" ]] ; then
+                # TODO: Bail out early?
+                run $HOME_MANAGER_BACKUP_COMMAND "$targetPath" || errorEcho "Running `$HOME_MANAGER_BACKUP_COMMAND` on '$targetPath' failed."
+              elif [[ -n "$HOME_MANAGER_BACKUP_EXT" ]] ; then
+                # The target exists, back it up
+                backup="$targetPath.$HOME_MANAGER_BACKUP_EXT"
+                run mv $VERBOSE_ARG "$targetPath" "$backup" || errorEcho "Moving '$targetPath' failed!"
+              fi
             fi
 
             if [[ -e "$targetPath" && ! -L "$targetPath" ]] && cmp -s "$sourcePath" "$targetPath" ; then
--- a/modules/files/check-link-targets.sh
+++ b/modules/files/check-link-targets.sh
@@ -30,6 +30,9 @@ for sourcePath in "$@" ; do
     if cmp -s "$sourcePath" "$targetPath"; then
       # First compare the files' content. If they're equal, we're fine.
       warnEcho "Existing file '$targetPath' is in the way of '$sourcePath', will be skipped since they are the same"
+    elif [[ ! -L "$targetPath" && -n "$HOME_MANAGER_BACKUP_COMMAND" ]] ; then
+      # Next, try to run the custom backup command. Assume this always succeeds.
+      :
     elif [[ ! -L "$targetPath" && -n "$HOME_MANAGER_BACKUP_EXT" ]] ; then
       # Next, try to move the file to a backup location if configured and possible
       backup="$targetPath.$HOME_MANAGER_BACKUP_EXT"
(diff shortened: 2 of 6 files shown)


aguirre-matteo:
That looks very good! I think this should have been done a long time ago.

modules/files.nix
```diff
@@ -146,10 +146,15 @@ in
           for sourcePath in "$@" ; do
             relativePath="''${sourcePath#$newGenFiles/}"
             targetPath="$HOME/$relativePath"
-            if [[ -e "$targetPath" && ! -L "$targetPath" && -n "$HOME_MANAGER_BACKUP_EXT" ]] ; then
-              # The target exists, back it up
-              backup="$targetPath.$HOME_MANAGER_BACKUP_EXT"
-              run mv $VERBOSE_ARG "$targetPath" "$backup" || errorEcho "Moving '$targetPath' failed!"
+            if [[ -e "$targetPath" && ! -L "$targetPath" ]] ; then
+              if [[ -n "$HOME_MANAGER_BACKUP_COMMAND" ]] ; then
+                # TODO: Bail out early?
```
khaneliman: Curious what you mean by bail out early? We should continue on to the link creation.

cdo256: I was thinking about the case in which the backup command failed, for whatever reason, but still removed the original file. However, upon second reading, it appears that's not necessary. Suppose the file is deleted without being backed up. In that case, it is lost anyway, and I suppose it is the user's responsibility to ensure that they select a reliable backup command that will fail gracefully. So I think this comment can be removed.

cdo256:
I've rebased my commit onto master. Please let me know if there are any blockers to merging this.

aguirre-matteo:
To me, it seems like you merged the master instead of rebasing, since the commit history says there are two commits.

EDIT: What I mean is, that in a rebase you HAVE TO force push and the commit history would show only one commit.

EDIT2: Maybe just squash the commits last two commits so they are seen as a single one.

nixos/common.nix
```diff
@@ -81,6 +81,16 @@ in
       argument in Home Manager. This disables the Home Manager
       options {option}`nixpkgs.*`'';
 
+    backupCommand = mkOption {
+      type = types.nullOr types.str;
```
khaneliman:
```suggestion
      type = types.nullOr (types.either types.str types.path);
```

Can support a command that's a path, too.

modules/files/check-link-targets.sh
```diff
@@ -30,6 +30,9 @@ for sourcePath in "$@" ; do
     if cmp -s "$sourcePath" "$targetPath"; then
       # First compare the files' content. If they're equal, we're fine.
       warnEcho "Existing file '$targetPath' is in the way of '$sourcePath', will be skipped since they are the same"
+    elif [[ ! -L "$targetPath" && -n "$HOME_MANAGER_BACKUP_COMMAND" ]] ; then
+      # Next, try to run the custom backup command. Assume this always succeeds.
```
khaneliman, line 34: maybe add a `verboseEcho` to add some logging that we've hit this block if we're creating a condition for it.

modules/files.nix
khaneliman: Would recommend removing the TODO, as well.

modules/files.nix
```diff
@@ -146,10 +146,15 @@ in
...
+                run $HOME_MANAGER_BACKUP_COMMAND "$targetPath" || errorEcho "Running `$HOME_MANAGER_BACKUP_COMMAND` on '$targetPath' failed."
+              elif [[ -n "$HOME_MANAGER_BACKUP_EXT" ]] ; then
```
khaneliman, line 153: I think we should add an assertion that `backupFileExtension` and `backupCommand` both can't be set, or the very least a warning when they are both set. 

Otherwise, we're just silently discarding an option.

cdo256:
Apoligies for the long delay, I've been focussing on my dissertation and neglected to check my GitHub notifications. I've pushed a new version that should address all of those points.

modules/files.nix
aguirre-matteo: IMO we shouldn't add a warning or assertion here, since the custom backup command might use the `HOME_MANAGER_BACKUP_EXT` env var during its process. Most users will likely use a shell script rather than a standalone program, so there's the case they might want to reference  that env var in the script. Of course they could hardcode the file extension, but I'm thinking of the case of an script that defaults to some backup extension, but uses Home-Manager's env var if set.

cdo256:
I've resolved the most recent conflicts. Is there anything left to do before this can be merged?

modules/files.nix
khaneliman: That's a good point, thanks. I didn't think about that use case.

modules/files.nix
```diff
@@ -143,16 +143,25 @@ in
 
           newGenFiles="$1"
           shift
+          if [[ -n "$HOME_MANAGER_BACKUP_COMMAND" && -n "$HOME_MANAGER_BACKUP_EXT" ]]; then
+            errorEcho "Conflicting environment variables HOME_MANAGER_BACKUP_COMMAND and HOME_MANAGER_BACKUP_EXT both set."
+            exit 1
+          fi
```
khaneliman:
```suggestion
```

@aguirre-matteo provided a value use case for when someone would configure both, so we can actually remove this.

khaneliman:
Sorry, forgot to ask if you could add a news entry for the new option.

cdo256:
Thank you very much for merging this!

PR #7158, opened 2025-05-29
hwatch: add module

Aehmlo:
### Description

This PR adds support for configuring [`hwatch`](https://github.com/blacknon/hwatch), a modern alternative to the `watch` command. `hwatch` is [available in Nixpkgs](https://search.nixos.org/packages?channel=24.11&type=packages&query=hwatch) and is configured [using the `HWATCH` environment variable](https://github.com/blacknon/hwatch?tab=readme-ov-file#configuration).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/hwatch.nix
+++ b/modules/programs/hwatch.nix
@@ -1,0 +1,50 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+let
+  inherit (lib)
+    mkIf
+    mkOption
+    types
+    ;
+  cfg = config.programs.hwatch;
+in
+{
+  meta.maintainers = with lib.hm.maintainers; [
+    Aehmlo
+  ];
+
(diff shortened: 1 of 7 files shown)


modules/programs/hwatch.nix
```diff
@@ -0,0 +1,50 @@
...
+  ];
+
+  options.programs.hwatch = {
+    enable = lib.mkEnableOption ''
+      hwatch, a modern alternative to the {command}`watch` command
+    '';
+
+    package = lib.mkPackageOption pkgs "hwatch" { nullable = true; };
+
+    extraOptions = mkOption {
```
khaneliman, line 28:
```suggestion
    extraArgs  = mkOption {
```
Or something like this?

Aehmlo: Changed to `extraArgs` for consistency.

PR #7159, opened 2025-05-30
numbat: add initFile

AyushmaanAggarwal:
This adds the option initFile for initialization of numbat shell with custom units, functions and constants

### Description

Numbat has an optional ~/.config/numbat/init.nbt to define custom functions, constants, and units. This PR adds a string option to insert text into that file. 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@Aehmlo


Aehmlo (changes requested):

Aehmlo:
Thanks for the PR! Looks pretty good to me overall; with the suggested change(s), I think we should be good to merge.

modules/programs/numbat.nix
```diff
@@ -38,6 +38,17 @@ in
         <https://numbat.dev/doc/cli-customization.html#configuration> for options.
       '';
     };
+
+    initFile = lib.mkOption {
+      type = lib.types.str;
```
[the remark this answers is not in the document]
AyushmaanAggarwal: Just added `types.nullOr` and `types.sourceFileOrLines` with some additional tests

modules/programs/numbat.nix
```diff
@@ -46,5 +63,16 @@ in
...
+    home.file."${configDir}/init.nbt" = mkIf (cfg.initFile != null) (
+      if (cfg.initFile.source != null) then
+        {
+          source = cfg.initFile.source;
+        }
+      else
+        {
+          text = cfg.initFile.text;
+        }
+    );
```
0xda157, line 76:
```suggestion
    home.file."${configDir}/init.nbt" = mkIf (cfg.initFile != null) cfg.initFile;
```

AyushmaanAggarwal: This is a really nice simplication, but it doesn't seem to work as [sourceFileOrLines sets source to null](https://github.com/nix-community/home-manager/blob/c7fdb7e90bff1a51b79c1eed458fb39e6649a82a/modules/lib/types.nix#L231). Since null is not a valid type for an absolute path, this causes a couple tests to fail [here](https://github.com/AyushmaanAggarwal/home-manager/blob/bcea8b77b6deaaf4c1420ae6079a4a5c1f5d2257/tests/modules/programs/numbat/example-init-text.nix)

AyushmaanAggarwal: I'm not really sure if it makes sense to change soureceFileOrLines (to not set source to null) as it might be used in [custom file creation like sketchybar](https://github.com/nix-community/home-manager/blob/c7fdb7e90bff1a51b79c1eed458fb39e6649a82a/modules/programs/sketchybar.nix#L268) 

And allowing `home.file.<name>.source` to null seems like it would introduce a whole host of issues if a files source is set to null in runtime. This would also define source twice as text is later used to set source by `home.file.<name>`.

If this gets fixed by upstream, I would be happy to incorporate this change in the future, but in the meantime, I will revert it and try to get this merged into home-manager as is

teto (changes requested):
This adds 100 lines of code for something any user can do with `xdg.configFile."numbat/init.nbt".text = ...`. 
The option is not used in the module so I would say it's an unnecessary PR.

AyushmaanAggarwal:
I am relatively new to home manager and didn't realize that was an option, closing this PR now

PR #7161, opened 2025-05-30
linux-wallpaperengine: Fix evaluation error when passing null as options

ckgxrg-salt:
### Description

This fixes evaluation error "cannot coerce null to a string" when setting either `services.linux-wallpaperengine.clamping` or `assetsPath` to null or leave as default.   

I received an email notification from @ccalhoun1999 from my mailbox, but I didn't see their comment on the original pull request, so I'll mention them here for reference.   

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

myself

### DIFF
--- a/modules/services/linux-wallpaperengine.nix
+++ b/modules/services/linux-wallpaperengine.nix
@@ -19,7 +19,8 @@ in
     package = lib.mkPackageOption pkgs "linux-wallpaperengine" { };
 
     assetsPath = mkOption {
-      type = types.path;
+      type = types.nullOr types.path;
+      default = null;
       description = "Path to the assets directory.";
     };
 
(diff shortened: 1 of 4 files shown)


ckgxrg-salt:
> Thank you for making this! I felt my comment was better posted in the nix discourse or flagged as a new issue so I deleted it and never got around to re-posting. Would it be possible to also post an example of proper formatting for the value to docs? I tried numerous ways to reference the home directory but they all resulted in the `cannot coerce null to a string` error.

The null problem is probably due to `service.linux-wallpaperengine.clamping` or `assetsPath` being null, and I didn't check for that...
So this pull request will ignore these options if they are set to null instead of trying to "coerce null to string"
`home.homeDirectory` is never null, it's not your fault, I think this pull request should be able to solve this.
Anyways, thanks for your support, please wait a few days before this get merged ;)

Issue #7143, opened 2025-05-27
bug: dconf.enable breaks gvfs (and probably more) on NixOS


keysmashes:
### Are you following the right branch?

- [x] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Issue description

Since #7021, if Home Manager's `dconf.enable` is set (i.e. on all non-Darwin platforms), then `$GIO_EXTRA_MODULES` is set via `home.sessionVariables`. However, NixOS also likes to set `$GIO_EXTRA_MODULES`, for example when NixOS `services.gvfs.enable` is set.

This means that Home Manager overrides the environment variable set by NixOS (rather than extending it), which breaks things like "mounting remote filesystems in thunar".

### Maintainer CC

@Rosuavio @rycee 

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.12.30, NixOS, 25.11 (Xantusia), 25.11.20250523.063f43f`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Lix, like Nix) 2.93.0
System type: x86_64-linux
Additional system types: i686-linux, x86_64-v1-linux, x86_64-v2-linux, x86_64-v3-linux
Features: gc, signed-caches
System configuration file: /etc/nix/nix.conf
User configuration files: /home/ash/.config/nix/nix.conf:/etc/xdg/nix/nix.conf:/home/ash/.nix-profile/etc/xdg/nix/nix.conf:/nix/profile/etc/xdg/nix/nix.conf:/home/ash/.local/state/nix/profile/etc/xdg/nix/nix.conf:/etc/profiles/per-user/ash/etc/xdg/nix/nix.conf:/nix/var/nix/profiles/default/etc/xdg/nix/nix.conf:/run/current-system/sw/etc/xdg/nix/nix.conf
Store directory: /nix/store
State directory: /nix/var/nix
Data directory: /nix/store/7wv7x01fjp05n40yb263wlrm0hvhzbyn-lix-2.93.0/share`
 - nixpkgs: `/etc/nixpkgs/channels/nixpkgs`
```

PR #7163, opened 2025-05-30
dconf: Fix Gio module variable breakage (#7146)

khaneliman:
In some setups, this would cause missing Gio modules that cause e.g. Nautilus to not be able to view the XDG trash, and potentially other issues.

Fixes: ec8205c3 ("dconf: set env var")
Fixes: #7143

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


LeCalicot:
I hope it's the right place, but I still encounter the issue.

I have a standalone HM installation on top of lubuntu 22.04 and neither nautilus nor pcmanfm-qt are able to display drives and the trash, etc.

From what I saw in https://github.com/NixOS/nixpkgs/issues/412131 there might be some workaround for NixOS, but I haven't found anything for standalone HM.

I do have the environment variable that's necessary, but I have the problem nonetheless.
```
$ env | grep GIO_EXTRA_MODULES
GIO_EXTRA_MODULES=/nix/store/0rq9g4rvssd2ldp9x854h49fn0gb2l43-dconf-0.40.0-lib/lib/gio/modules
```

I'm on HM 25.05 and the matching nix package release.

Should I open a dedicated issue? @khaneliman

khaneliman:
Your output looks like you aren't using the updated environment variable. Are you on latest 25.05 and have rebooted?

LeCalicot:
to make sure I put myself on the unstable version and the latest home-manager (it was not working with the 25.05):
```
    nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
```
I have rebuilt/switched HM and rebooted just now, as a result I still have the issue and the output is:
```
$ env | grep GIO_EXTRA_MODULES
GIO_EXTRA_MODULES=/nix/store/0rq9g4rvssd2ldp9x854h49fn0gb2l43-dconf-0.40.0-lib/lib/gio/modules
```

the output of nix-channel:
```
$ nix-channel --list
home-manager https://github.com/nix-community/home-manager/archive/master.tar.gz
nixgl https://github.com/nix-community/nixGL/archive/main.tar.gz
nixpkgs https://nixos.org/channels/nixpkgs-unstable
```

The issue is still the same on pcmanfm-qt and nautilus. That's the same for the 25.05 and the unstable, so there might be some other underlying issue...?

I would like to have this issue solve, let me know if I shall open a new issue :)

PS: I'm rather new with HM

PR #7165, opened 2025-05-30
ci: add backport workflow

MattSturgeon:
### Description

As I was recently adding this workflow to Nixvim, I figured home-manager would benefit too.

Adds a GitHub workflow to create backport PRs when merged PRs have a "backport" label.

This uses the very flexible https://github.com/korthout/backport-action and is based on the [workflow used by nixpkgs](https://github.com/NixOS/nixpkgs/blob/2566f9dc/.github/workflows/backport.yml), which is also [used by Stylix](https://github.com/nix-community/stylix/blob/0512b0f/.github/workflows/backport.yml) and [most recently, Nixvim](https://github.com/nix-community/nixvim/pull/3411). See [how it works](https://github.com/korthout/backport-action#how-it-works).

Currently, the workflow [will assume](https://github.com/korthout/backport-action#label_pattern) that backport labels follow the regex `backport ([^ ]+)$` (e.g. `backport nixos-25.05`), however it could be customised by assigning a different regex to `with: label_pattern`.

In order to trigger CI workflows to run on PRs created by the backport workflow, I've configured it to use `GH_TOKEN_FOR_UPDATES`. Ideally both the backport workflow and the update workflow would migrate to using a GitHub App with the [create-github-app-token](https://github.com/actions/create-github-app-token) action. I've added some TODO comments, however I don't want to distract from the crux of this PR by tackling such a migration now.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@khaneliman

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,0 +1,40 @@
+# Based on
+# https://github.com/NixOS/nixpkgs/blob/2566f9dc/.github/workflows/backport.yml
+#
+# NOTE: this uses the GH_TOKEN_FOR_UPDATES because pushing a backport PR using
+# GITHUB_TOKEN does not trigger CI.
+# TODO: consider switching to a GitHub App
+name: Backport
+
+on:
+  pull_request_target:
+    types:
+      - closed
+      - labeled
+
+jobs:
+  backport:
+    name: Backport Pull Request
+    if: >
+      secrets.GH_TOKEN_FOR_UPDATES
+      && github.event.pull_request.merged == true
+      && (
+        github.event.action != 'labeled'
+        || startsWith(github.event.label.name, 'backport')
+      )
+
+    runs-on: ubuntu-24.04-arm
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.GH_TOKEN_FOR_UPDATES }}
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Create backport PRs
+        id: backport
+        uses: korthout/backport-action@v3
+        with:
+          # See https://github.com/korthout/backport-action#inputs
+          github_token: ${{ secrets.GH_TOKEN_FOR_UPDATES }}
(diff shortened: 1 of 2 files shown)


.github/workflows/update-flake.yml
```diff
@@ -17,5 +17,8 @@ jobs:
       - name: Update flake.lock
         uses: DeterminateSystems/update-flake-lock@v25
         with:
+          # NOTE: this uses the GH_TOKEN_FOR_UPDATES because pushing a backport PR using
+          # GITHUB_TOKEN does not trigger CI.
```
khaneliman, line 21:
```suggestion
```
Copy paste error?

.github/workflows/backport.yml
```diff
@@ -0,0 +1,40 @@
...
+          token: ${{ secrets.GH_TOKEN_FOR_UPDATES }}
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Create backport PRs
+        id: backport
+        uses: korthout/backport-action@v3
+        with:
+          # See https://github.com/korthout/backport-action#inputs
+          github_token: ${{ secrets.GH_TOKEN_FOR_UPDATES }}
+          branch_name: backport/${target_branch}/${pull_number}
```
MattSturgeon, line 39: @khaneliman I forgot to ask what branch name style you prefer for hm backports.

Personally, I'm a fan of scoped path-like branch names for this kinda thing, but nixpkgs uses [the default](https://github.com/korthout/backport-action#branch_name) `backport-${pull_number}-to-${target_branch}` pattern.

MattSturgeon:
FYI, I believe this workflow will only work on PRs that already have it on their base branch.

So new PRs and rebased PRs should work, but PRs still based on an older revision will not trigger this workflow.

khaneliman:
Cool, needed to fix condition in https://github.com/nix-community/home-manager/pull/7167 but it worked backporting https://github.com/nix-community/home-manager/pull/7157 without updating that branch.

PR #7167, opened 2025-05-30
ci: fix backport if condition

khaneliman:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -16,15 +16,19 @@ jobs:
   backport:
     name: Backport Pull Request
     if: >
-      secrets.GH_TOKEN_FOR_UPDATES
-      && github.event.pull_request.merged == true
-      && (
-        github.event.action != 'labeled'
+      github.event.pull_request.merged == true &&
+      ( github.event.action != 'labeled'
         || startsWith(github.event.label.name, 'backport')
       )
 
     runs-on: ubuntu-24.04-arm
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/backport.yml
```diff
@@ -16,15 +16,19 @@ jobs:
   backport:
     name: Backport Pull Request
     if: >
-      secrets.GH_TOKEN_FOR_UPDATES
```
MattSturgeon, line 19: I was half wondering if `secrets` would work in an `if` condition. Elsewhere, in other repos we can check a `var`.

.github/workflows/backport.yml
```diff
@@ -16,15 +16,19 @@ jobs:
...
         || startsWith(github.event.label.name, 'backport')
       )
 
     runs-on: ubuntu-24.04-arm
     steps:
+      - name: Check token
+        run: |
+          if [ -z "${{ secrets.GH_TOKEN_FOR_UPDATES }}" ]; then
+            echo "GH_TOKEN_FOR_UPDATES secret is not set"
+            exit 1
```
MattSturgeon, line 30: This doesn't gain much; half of the reason to check whether the secret is present, is to cancel the workflow silently without error if it is accidentally run in a fork.

I'd recommend just using a `github.owner == 'nix-comnunity'` check in the original `if` condition if we can't read the secret.

PR #7175, opened 2025-05-31
ci: switch to a GitHub App

MattSturgeon:
### Description

Followup to #7165.

Workflows will now push to PRs as a "bot user" (GitHub App), which will still trigger CI. This is more idiomatic than a "machine" user which is in reality a normal user account.

Its tokens are short lived, so more secure than a PAT that doesn't expire.

This PR is **blocked** by a GitHub App being created for Home Manager and the relevant APP ID and PRIVATE KEY being added to the repo's vars and secrets respectively. A nix-community admin will be able to do this, see [previous matrix thread](https://matrix.to/#/!PbtOpdWBSRFbEZRLIf:numtide.com/$copbS2S4-Jk8Cqvp-DFkZtxfdwmUUPWV1edITMvSnIA?via=matrix.org&via=envs.net&via=nixos.dev) for an example.

Additionally, I've improved the PR body for the update workflow.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@khaneliman

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,9 +1,5 @@
 # Based on
 # https://github.com/NixOS/nixpkgs/blob/2566f9dc/.github/workflows/backport.yml
-#
-# NOTE: this uses the GH_TOKEN_FOR_UPDATES because pushing a backport PR using
-# GITHUB_TOKEN does not trigger CI.
-# TODO: consider switching to a GitHub App
 name: Backport
 
 on:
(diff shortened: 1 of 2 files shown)


khaneliman:
Sounds like it got set up per [zowoq on matrix](https://matrix.to/#/!PbtOpdWBSRFbEZRLIf:numtide.com/$7dEBwfBFANT8haCTeFbpROYHny4jw-WjsYWWAGQvkLw?via=blad.is&via=matrix.org&via=envs.net )

khaneliman:
Will test it, if it doesn't work we can tweak/revert

EDIT: looks to have worked, thanks!

PR #7176, opened 2025-05-31
home-cursor: sway depend on cfg.size instead of gtk.cursorTheme.size

eyJhb:
### Description
It seems like `home.pointerCursor.sway.enable` wrongly depends on `gtk.cursorTheme.size`. instead of just `cfg.size`. If we want to depend on `gtk` instead, then the name should also depend on `gtk`, at least from my logic. 

It looks like it might have been an oversight from the initial PR here https://github.com/nix-community/home-manager/pull/6459

Maybe @ambroisie can clarify if this is intended? :) 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/config/home-cursor.nix
+++ b/modules/config/home-cursor.nix
@@ -249,7 +249,7 @@ in
             config = {
               seat = {
                 "*" = {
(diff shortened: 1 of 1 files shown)


ambroisie:
Yup, copy-paste error when using the `hyprcursor` one as a base.

PR #7178, opened 2025-05-31
yazi: change the default `shellWrapperName` `yy` → `y`

phucisstupid:
### Description

base on yazi docs: https://yazi-rs.github.io/docs/quick-start/#shell-wrapper:
We suggest using this `y` shell wrapper that provides the ability to change the current working directory when exiting Yazi.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/docs/release-notes/rl-2605.md
+++ b/docs/release-notes/rl-2605.md
@@ -24,6 +24,10 @@ changes are only active if the `home.stateVersion` option is set to
   configuration directory (usually `~/.config/zsh`) when
   [](#opt-xdg.enable) is true.
 
+- The [](#opt-programs.yazi.shellWrapperName) option now defaults to
+  `y` instead of `yy`. For users with older `home.stateVersion` values,
+  the legacy default `yy` is retained.
+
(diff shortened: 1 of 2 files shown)


Weathercold:
Technically a breaking change since users expect the old alias to be still available
I do think we should use the recommended alias, though.

phucisstupid:
`y` is faster than `yy` right?. There is no programs except `yazi` use `y` as an alias

Weathercold:
yes. But still a breaking change

teto:
yes, add a news entry via modules/misc/news/create-news-entry.sh.

khaneliman:
Debating whether a simple alias launcher change is worth a `stateVersion` `lib.warn`...

teto:
As a datapoint to help in decision, I have several monoletter aliases, including `alias y=yazi`. I should probably use this wrapper instead since I 've wanted its functionality. Isn't the `stateVersion`straightforward like

        `default = if lib.versionAtLeast config.home.stateVersion "26.05" then "y" else "yy";`

considering it's trivial let's do it ?

PR #7179, opened 2025-05-31
jellyfin-mpv-shim: make sure the config file is read properly

YTG1234:
### Description

jellyfin-mpv-shim can't read its configuration file if it's not writeable (see https://github.com/nix-community/home-manager/pull/7129#issuecomment-2925542595).

### Checklist

    I can't run the tests on my platform (`aarch64-linux`).

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@repparw

### DIFF
--- a/modules/services/jellyfin-mpv-shim.nix
+++ b/modules/services/jellyfin-mpv-shim.nix
@@ -107,22 +107,6 @@ in
     ];
 
     xdg.configFile = {
-      # jellyfin-mpv-shim errors if the config file isn't writeable
-      # so, as a temporary workaround, we generate a hidden file (.conf.json)
-      # and make a writeable copy.
-      "jellyfin-mpv-shim/conf.json" = lib.mkIf (cfg.settings != { }) {
-        source = jsonFormat.generate "jellyfin-mpv-shim-conf" cfg.settings;
-        target = "jellyfin-mpv-shim/.conf.json";
-        onChange =
-          let
-            dir = "${config.xdg.configHome}/jellyfin-mpv-shim";
-          in
-          ''
-            cp --dereference ${dir}/.conf.json ${dir}/conf.json
-            chmod u+w ${dir}/conf.json
-          '';
-      };
-
       "jellyfin-mpv-shim/mpv.conf" = lib.mkIf (cfg.mpvConfig != null) {
         text = renderOptions cfg.mpvConfig;
       };
@@ -148,5 +132,35 @@ in
         WantedBy = [ "graphical-session.target" ];
       };
     };
+
+    # Yoinked from programs/zed-editor.nix
+    # jellyfin-mpv-shim can't load the configuration file if it's not
+    # writeable. So we merge the settings defined here in Nix with the existing
+    # configuration, if any.
+    home.activation.jellyfinMpvShimSettingsActivation =
+      let
+        path = lib.escapeShellArg "${config.xdg.configHome}/jellyfin-mpv-shim/conf.json";
+        staticSettings = lib.escapeShellArg (jsonFormat.generate "jellyfin-mpv-shim-conf" cfg.settings);
+        cmd = "${lib.getExe pkgs.jq} -s '.[0] * .[1]' ${path} ${staticSettings}";
+      in
+      lib.mkIf (cfg.settings != { }) (
+        lib.hm.dag.entryAfter [ "linkGeneration" ] ''
+          run mkdir -p "$(dirname ${path})"
+          if [ ! -e ${path} ]; then
+            # Create the file
+            if [[ -v DRY_RUN ]]; then
+              run echo '{}' '>' ${path}
+            else
+              echo '{}' > ${path}
+            fi
+          fi
+          if [[ -v DRY_RUN ]]; then
+            run ${cmd} '>' ${path}
+          else
+            config="$(${cmd})"
+            printf '%s\n' "$config" > ${path}
+          fi
+        ''
+      );
   };
 }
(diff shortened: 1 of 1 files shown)


repparw (approved):

repparw:
I don't know if there's another more standard way of making a copy instead of the in-store symlink, either in home-manager libs or upstream in nix/nixpkgs. Apart from that, LGTM

@khaneliman PTAL

khaneliman:
> I don't know if there's another more standard way of making a copy instead of the in-store symlink, either in home-manager libs or upstream in nix/nixpkgs. Apart from that, LGTM
> 
> @khaneliman PTAL

I am not familiar with a good standard approach, it is rarely done. I saw https://github.com/nix-community/home-manager/pull/6993 does something with creating the file during home-manager activation when we generate links.

repparw:
The only other program I use that does something like that I believe is https://github.com/nix-community/home-manager/blob/5675a9686851d9626560052a032c4e14e533c1fa/modules/programs/rclone.nix, which has a writeAfter thrown in there

repparw:
probably good enough to be merged, if a std approach is defined it'll probably get a treewide batch PR treatment

YTG1234:
I guess it should be decided whether the file is copied in a dedicated activation script or in `onChange`. Maybe the latter is more efficient?

khaneliman:
> I guess it should be decided whether the file is copied in a dedicated activation script or in `onChange`. Maybe the latter is more efficient?

Personally, I think it would make more sense to deploy the config during the activation script instead of relying on a change event outside of our control.

YTG1234:
> > I don't know if there's another more standard way of making a copy instead of the in-store symlink, either in home-manager libs or upstream in nix/nixpkgs. Apart from that, LGTM
> > @khaneliman PTAL
> 
> I am not familiar with a good standard approach, it is rarely done. I saw #6993 does something with creating the file during home-manager activation when we generate links.

Then would it make sense to copy this approach exactly? As this is also JSON.

khaneliman:
> > > I don't know if there's another more standard way of making a copy instead of the in-store symlink, either in home-manager libs or upstream in nix/nixpkgs. Apart from that, LGTM
> > > @khaneliman PTAL
> > 
> > 
> > I am not familiar with a good standard approach, it is rarely done. I saw #6993 does something with creating the file during home-manager activation when we generate links.
> 
> Then would it make sense to copy this approach exactly? As this is also JSON.

Probably, try and see if it works for your use case. If so, we could move that function outside to the lib so each module doesn't need to redeclare it.

YTG1234:
Making shell redirection work with `DRY_RUN` is annoying (`run` is only a shell function and can't handle redirection)… is there a way to do that without duplicating code? As in
```sh
        if [[ -v DRY_RUN ]]; then
          echo echo "'{}'" '>' ${path}
        else
          echo '{}' > ${path}
        fi
```

khaneliman:
> Making shell redirection work with `DRY_RUN` is annoying (`run` is only a shell function and can't handle redirection)… is there a way to do that without duplicating code? As in
> 
> ```shell
>         if [[ -v DRY_RUN ]]; then
>           echo echo "'{}'" '>' ${path}
>         else
>           echo '{}' > ${path}
>         fi
> ```

Could just create a nix variable that gets inserted for the actual command and the DRY_RUN block just prepends with a echo. 

I saw taskwarrior has its own custom activation that was similar, here's what I would refactor it to. 

```diff
-      home.activation.regenDotTaskRc = lib.hm.dag.entryAfter [ "writeBoundary" ] ''
-        verboseEcho "Ensuring generated taskwarrior config included in taskrc"
-
-        if [[ ! -s "${userConf}" ]]; then
-          # Ensure file's existence
-          if [[ -v DRY_RUN ]]; then
-            run echo "include ${homeConf}" ">" "${userConf}"
-          else
-            echo "include ${homeConf}" > "${userConf}"
+      home.activation.regenDotTaskRc =
+        let
+          cmd = ''echo "include ${homeConf}" ">" "${userConf}"'';
+        in
+        lib.hm.dag.entryAfter [ "writeBoundary" ] ''
+          verboseEcho "Ensuring generated taskwarrior config included in taskrc"
+
+          if [[ ! -s "${userConf}" ]]; then
+            # Ensure file's existence
+            if [[ -v DRY_RUN ]]; then
+              run ${cmd}
+            else
+              ${cmd}
+            fi
+          elif ! grep -qF "include ${homeConf}" ${lib.escapeShellArg userConf}; then
+            # Add include statement for Home Manager generated config.
+            run sed -i '1i include ${homeConf}' ${lib.escapeShellArg userConf}
           fi
-        elif ! grep -qF "include ${homeConf}" ${lib.escapeShellArg userConf}; then
-          # Add include statement for Home Manager generated config.
-          run sed -i '1i include ${homeConf}' ${lib.escapeShellArg userConf}
-        fi
-      '';
+        '';
     };
 }

```

YTG1234:
I've done something similar.

The method used is specific to JSON (relies on `jq`), so factoring it out to the lib isn't trivial. Could still be worth doing, but I think this is somewhat outside the scope of this workaround-y PR.

khaneliman:
LGTM if it still works

repparw (approved):

repparw:
I don't have write perms. @khaneliman

khaneliman:
> I don't have write perms. @khaneliman

Not sure what you mean by that? I was just waiting to hear back if the change was still good before merging.

repparw:
oh didn't see your comment, was asked to review and then tagged you haha, sorry

PR #7182, opened 2025-06-01
timewarrior: init

Prince213:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/timewarrior.nix
+++ b/modules/programs/timewarrior.nix
@@ -1,0 +1,36 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+let
+  cfg = config.programs.timewarrior;
+  taskCfg = config.programs.taskwarrior;
+in
+{
+  options.programs.timewarrior = {
+    enable = lib.mkEnableOption "Timewarrior";
+    package = lib.mkPackageOption pkgs "timewarrior" { };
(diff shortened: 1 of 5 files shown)


modules/programs/timewarrior.nix
```diff
@@ -0,0 +1,36 @@
...
+
+let
+  cfg = config.programs.timewarrior;
+  taskCfg = config.programs.taskwarrior;
+in
+{
+  options.programs.timewarrior = {
+    enable = lib.mkEnableOption "Timewarrior";
+    package = lib.mkPackageOption pkgs "timewarrior" { };
+    taskwarrior.enable = lib.mkOption {
```
cardboardcpu, line 16: Nit: maybe naming the option something like `taskwarriorHook` would be oportune?

Prince213: I do have a few names in mind like this one (as it's similar to aerc.enable) and `enableTaskwarriorIntegration` but I'm not really sure yet.

Prince213:
@cardboardcpu Can you help fix & run the tests? I'm not sure why it fails that way.

```
$ nix-build --pure --option allow-import-from-derivation false tests -A build.timewarrior
Expected symlink home-files/.local/share/task/hooks/on-modify.timewarrior to exist but it was not found.
```

PR #7195, opened 2025-06-03
oh-my-posh: added configFile option

dusktreader:
### Description

This option allows you to provide a custom config file for oh-my-posh as you would with the `--config` option. If set, `useTheme` will be ignored. If `settings` is provided, then this option will be ignored.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@arjan-s 
@emilazy 
@khaneliman

### DIFF
--- a/modules/programs/oh-my-posh.nix
+++ b/modules/programs/oh-my-posh.nix
@@ -14,6 +14,8 @@ let
   configArgument =
     if cfg.settings != { } then
       "--config ${config.xdg.configHome}/oh-my-posh/config.json"
+    else if cfg.configFile != null then
+      "--config ${cfg.configFile}"
     else if cfg.useTheme != null then
       "--config ${cfg.package}/share/oh-my-posh/themes/${cfg.useTheme}.omp.json"
     else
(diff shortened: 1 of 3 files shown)


modules/programs/oh-my-posh.nix
```diff
@@ -41,6 +43,18 @@ in
       '';
     };
 
+    configFile = lib.mkOption {
+      type = lib.types.nullOr lib.types.str;
```
khaneliman, line 47: Should support a path itself for an option that says its for configuring a file.

khaneliman (commented):
> This option allows you to provide a custom config file for oh-my-posh as you would with the `--config` option. If set, `useTheme` will be ignored. If `settings` is provided, then this option will be ignored.

I don't like the idea of just silently ignoring configurations. You should be providing some kind of warning or assertion that a config is incorrect or will have certain side effects.

modules/programs/oh-my-posh.nix
dusktreader: I'll work on making this support an explicit path.

dusktreader:
> > This option allows you to provide a custom config file for oh-my-posh as you would with the `--config` option. If set, `useTheme` will be ignored. If `settings` is provided, then this option will be ignored.
> 
> I don't like the idea of just silently ignoring configurations. You should be providing some kind of warning or assertion that a config is incorrect or will have certain side effects.

The code was already silently ignoring `useTheme` when `settings` is provided. I was just following that pattern. Additionally, you'll notice that the `useTheme` documentation mentions a `configFile` option already (though one didn't exist) and explains that it will not be used if such an option is specified.

If you are sure that we should break away from the existing behavior and show some kind of warning, I can work on adding that logic in.

PR #7196, opened 2025-06-03
keychain: warn about deprecated options

pedorich-n:
### Description

With the Keychain release [2.9.0](https://github.com/funtoo/keychain/releases/tag/2.9.0) options `--agents` and `--inherit` have been deprecated.

The message looks like this:
```
trace: warning: Option `programs.keychain.agents` is deprecated and will be removed in the future.
Please avoid using it.
See https://github.com/funtoo/keychain/releases/tag/2.9.0 for more information
```

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/keychain.nix
+++ b/modules/programs/keychain.nix
@@ -82,25 +82,47 @@ in
     };
   };
 
-  config = mkIf cfg.enable {
-    home.packages = [ cfg.package ];
-    programs.bash.initExtra = mkIf cfg.enableBashIntegration ''
-      eval "$(SHELL=bash ${shellCommand})"
-    '';
-    programs.fish.interactiveShellInit = mkIf cfg.enableFishIntegration ''
-      SHELL=fish eval (${shellCommand})
-    '';
-    programs.zsh.initContent = mkIf cfg.enableZshIntegration ''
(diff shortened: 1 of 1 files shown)


khaneliman:
The idea behind deprecation is giving users a warning that they should make changes to their config, when they have time, so that when it is actually removed it doesn't surprise them. This jumps straight to that surprise removal for HM users. Why not give a similar deprecation warning when the option is used ?

pedorich-n:
> The idea behind deprecation is giving users a warning that they should make changes to their config, when they have time, so that when it is actually removed it doesn't surprise them. This jumps straight to that surprise removal for HM users. Why not give a similar deprecation warning when the option is used ?

That's fair. I guess I didn't find a way as simple as `mkRemovedOptionModule` for warnings about deprecations and went with the nuclear option 😅 
I found some examples of using `warnings` to warn about the deprecated options and followed them.
Please have another look, @khaneliman

kpbaks:
Should these warnings not be guarded by a condition like
```nix
lib.versionAtLeast cfg.package.version "2.9.0"
```

In case a user want to use an older version of `pkgs.keychain`?

PR #5458, opened 2024-05-27
zsh: add plugins.*.completions paths to fpath

zimeg:
### Description

👋 This PR adds a `programs.zsh.plugins.*.completions` options that accepts paths to plugin directories containing additional scripts, such as completion scripts. This adds these paths to `fpath` before calling `zsh.completionInit` - the point that completions are set.

Similar fixes found in #2562 are applied here with the option to extend the files that are being sourced to whichever directories are needed.

### Reviewers

These changes can be tested with [the `wd` package](https://github.com/mfaerevaag/wd). I find this plugin showcases a bit of everything well!

```nix
{
  programs.zsh = {
    enable = true;
    plugins = [
      {
        name = "wd";
        src = pkgs.zsh-wd;
        file = "share/wd/wd.plugin.zsh";
        completions = [ "share/zsh/site-functions" ];
      }
    ];
  };
}
```

After switching to a setup with `wd` the following commands and completions should work:

```sh
# Confirm installation
$ wd --version
$ man wd

# Set a warp point
$ cd /tmp
$ wd add tmp
$ wd list

# Confirm warps work
$ cd
$ pwd
$ wd tmp
$ pwd
```

Also check that the `.zshrc` file includes this new option:

```sh
$ cat ~/.zshrc
...
path+="$HOME/.zsh/plugins/wd"
fpath+="$HOME/.zsh/plugins/wd"
fpath+=("$HOME/.zsh/plugins/wd/share/zsh/site-functions")
...
```

### Notes

For a bit more context, the `result` of a built `zsh-wd` has this structure:

```txt
result
| share/
  | man/
    | man1/
      - wd.1.gz
  | wd/
    - wd.plugin.zsh
    - wd.sh
  | zsh/
    | site-functions/
      - _wd.sh
```

And before this change, this snippet was needed to add completion scripts:
 
```nix
programs.zsh.initExtraBeforeCompInit = '' fpath+=("${pkgs.zsh-wd}"/share/zsh/site-functions) '';
```

Here's a bit of reference I found useful too: https://thevaluable.dev/zsh-completion-guide-examples/

### Checklist

- [x] Change is backwards compatible.

- [x] Code formatted with `./format`.

- [x] Code tested through `nix-shell --pure tests -A run.all` or `nix develop --ignore-environment .#all` using Flakes.

  <details>
  <summary>
  Changes to this module pass tests but a failure happened when running all tests and seems unrelated
  </summary>

  ```
  error:
       … while calling the 'derivationStrict' builtin

[...]

PR #7197, opened 2025-06-03
zsh: add plugins.*.completions paths to fpath

khaneliman:
Autocompletion scripts and additional plugin functions are located in specific directories that might not match the plugin source script but need to be included in fpath before calling compinit.

An option to provide a path to these scripts is added to add the paths to fpath before calling completionInit.

### Description
Closes https://github.com/nix-community/home-manager/pull/5458

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/zsh.nix
+++ b/modules/programs/zsh.nix
@@ -183,15 +183,22 @@ in
               type = types.str;
               description = ''
                 The name of the plugin.
-
-                Don't forget to add {option}`file`
-                if the script name does not follow convention.
               '';
             };
 
             file = mkOption {
               type = types.str;
-              description = "The plugin script to source.";
+              description = ''
+                The plugin script to source.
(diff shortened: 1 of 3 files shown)


modules/programs/zsh.nix
```diff
@@ -183,15 +183,22 @@ in
...
               type = types.str;
-              description = "The plugin script to source.";
+              description = ''
+                The plugin script to source.
+                Required if the script name does not match {file}`name.plugin.zsh`
+                using the plugin {option}`name` from the plugin {option}`src`.
+              '';
+            };
+
+            completions = mkOption {
```
khaneliman, line 198: should this just be a generic option instead of naming after completions?

khaneliman: We can rename later, I suppose. Will just keep this as originally implemented, for now.

PR #7204, opened 2025-06-04
papis: allow libraries to not to be defined

octvs:
papis allows setting configuration via local [1] or via a python file [2]. A user can decide to define libraries via these methods, which home-manager module won't allow. Shift where the check is done to allow a config file without library definitions.

[1]: https://papis.readthedocs.io/en/latest/configuration.html#local-configuration-files
[2]: https://papis.readthedocs.io/en/latest/configuration.html#python-configuration-file

### Description

papis doesn't enforce library definitions, contrary to what is done in this module. This struck me when I wanted to set libraries on local papis config files, and only use the global config file for settings. 

This change works around the issue. Since I don't understand the reasoning of this assertion [1] fully, I'm unsure whether or not to remove it. Maybe this PR can start the discussion.

[1]: https://github.com/nix-community/home-manager/pull/3626

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@marsam @rycee

### DIFF
--- a/modules/programs/papis.nix
+++ b/modules/programs/papis.nix
@@ -14,9 +14,11 @@ let
   );
 
   settingsIni = (lib.mapAttrs (n: v: v.settings) cfg.libraries) // {
-    settings = cfg.settings // {
-      "default-library" = lib.head defaultLibraries;
-    };
+    settings =
+      cfg.settings
+      // lib.optionalAttrs (cfg.libraries != { }) {
+        "default-library" = lib.head defaultLibraries;
+      };
   };
 
 in
@@ -97,6 +99,7 @@ in
           }
         )
       );
+      default = { };
       description = "Attribute set of papis libraries.";
     };
   };
@@ -116,8 +119,6 @@ in
 
     home.packages = lib.mkIf (cfg.package != null) [ cfg.package ];
 
-    xdg.configFile."papis/config" = lib.mkIf (cfg.libraries != { }) {
-      text = lib.generators.toINI { } settingsIni;
-    };
+    xdg.configFile."papis/config".text = lib.generators.toINI { } settingsIni;
   };
 }


khaneliman:
Not familiar with program, but reading the linked documentation it makes complete sense.

octvs:
Most users probably use it via the global config file via ini format. I was
also curious [1] about the other options. Turns out, it is possible but maybe
not perfectly documented yet [2].

[1]: https://github.com/papis/papis/discussions/851
[2]: https://github.com/papis/papis/issues/858

teto:
seems like the assertion could go away then. Maybe improve the doc regarding libraries with your findings ?

octvs:
> seems like the assertion could go away then. Maybe improve the doc regarding libraries with your findings ?

Yes, just elaborated on the option description.

I was unsure at first, there are couple of scenarios where an assertion might
be done but I would argue `papis` gives sufficient warning to the user.

Example: empty config file (`default-library` defaults to `papers`)

```sh
$ papis open
[ERROR] commands.default: Library 'papers' does not have any folders attached to it. Please create and add the required paths to the configuration file.
[WARNING] commands.open: No documents retrieved
```

Hence, the home-manager module should not enforce a premature assertion, and
let `papis` handle how it parses the config [^1].

At the end, I see the whole library submodule as unneeded complexity, and would
remove it on the next refactor. The config file is in ini format, why not let
it be generated as is from the attrset.

[^1]: There are some cases where errors are not as pretty, but this should be an
upstream concern. 

    Example: `settings.default-library` and `papers` library are undefined but
    other libraries exist

    ```sh
    $ papis open
    <longer_trace>
      File "/nix/store/yfccgjhafbjbjxyxvpihfy2mncpm7f7w-python3.12-papis-b116820/lib/python3.12/site-packages/papis/config.py", line 629, in set_lib_from_name
        set_lib(get_lib_from_name(libname))
                ^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/nix/store/yfccgjhafbjbjxyxvpihfy2mncpm7f7w-python3.12-papis-b116820/lib/python3.12/site-packages/papis/config.py", line 650, in get_lib_from_name
        raise RuntimeError(
    RuntimeError: Library 'papers' does not seem to exist. To add a library simply write the following in your configuration file located at '<redacted>/.config/papis/config'

      [papers]
      dir = path/to/your/papers/folder
    ```

khaneliman (approved):

teto:
I agree with you. Home-manager is not intended to duplicate the program checks. Worse if like in this case the added abstraction makes some scenarii impossible.

PR #7221, opened 2025-06-06
lib/default.nix: remove inefficient second copy

grahamc:
### Description

Using ./. forces Nix to copy the flake to the store a second time. Using builtins.path like this, forcing the name to be "source", reduces the extraneous copy.

ref:

```
warning: Copying '/nix/store/b3sj22x9pqf3dyw0409dhpzwi9vswwx7-source' to the store again
You can make Nix evaluate faster and copy fewer files by replacing `./.` with the `self` flake input, or `builtins.path { path = ./.; name = "source"; }`

Location: /nix/store/b3sj22x9pqf3dyw0409dhpzwi9vswwx7-source/lib/default.nix:23:68
```

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

cc @khaneliman

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -20,7 +20,14 @@
       configuration =
         { ... }:
         {
-          imports = modules ++ [ { programs.home-manager.path = "${../.}"; } ];
+          imports = modules ++ [
+            {
+              programs.home-manager.path = builtins.path {
+                path = ../.;
+                name = "source";
+              };
+            }
+          ];
 
           nixpkgs = {
             config = lib.mkDefault pkgs.config;


lib/default.nix
```diff
@@ -20,7 +20,14 @@
       configuration =
         { ... }:
         {
-          imports = modules ++ [ { programs.home-manager.path = "${../.}"; } ];
+          imports = modules ++ [
+            {
+              programs.home-manager.path = builtins.path {
+                path = ./.;
```
gustavderdrache:
```suggestion
                path = ../.;
```

grahamc: wow I'm awesome, thanks

PR #7225, opened 2025-06-06
mc: add midnight commander module with tests

egorkonovalov:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1748693115,
-        "narHash": "sha256-StSrWhklmDuXT93yc3GrTlb0cKSS0agTAxMGjLKAsY8=",
+        "lastModified": 1749143949,
+        "narHash": "sha256-QuUtALJpVrPnPeozlUG/y+oIMSLdptHxb3GK6cpSVhA=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "910796cabe436259a29a72e8d3f5e180fc6dfacc",
+        "rev": "d3d2d80a2191a73d1e86456a751b83aa13085d7d",
         "type": "github"
       },
       "original": {
--- a/modules/modules.nix
+++ b/modules/modules.nix
@@ -174,6 +174,7 @@ let
       ./programs/lapce.nix
       ./programs/lazydocker.nix
       ./programs/lazygit.nix
+      ./programs/lazysql.nix
       ./programs/ledger.nix
       ./programs/less.nix
       ./programs/lesspipe.nix
@@ -219,6 +220,7 @@ let
       ./programs/notmuch.nix
       ./programs/numbat.nix
       ./programs/nushell.nix
+      ./programs/nyxt.nix
       ./programs/obs-studio.nix
       ./programs/obsidian.nix
       ./programs/octant.nix
--- a/modules/programs/fish.nix
+++ b/modules/programs/fish.nix
@@ -212,56 +212,68 @@ let
     };
   };
 
-  bindModule = types.submodule {
-    options = {
-      enable = mkEnableOption "enable the bind" // {
-        default = true;
+  bindModule = types.submodule (
+    { config, ... }:
+    {
+      options = {
+        enable = mkEnableOption "enable the bind. Set false if you want to ignore the bind" // {
+          default = true;
+        };
+        mode = mkOption {
+          description = "Specify the bind mode that the bind is used in";
+          type =
+            with types;
+            nullOr (enum [
+              "default"
+              "insert"
+              "paste"
+            ]);
+          default = null;
+        };
+        command = mkOption {
+          description = "command that will be execute";
+          type =
+            let
+              origin =
+                with types;
+                nullOr (oneOf [
+                  str
+                  (listOf str)
+                ]);
+            in
+            origin
+            // {
+              description = "string or list of string (optional when erase is set to true)";
+              check = x: if !config.erase && isNull x then false else origin.check x;
+            };
+          default = null;
+        };
+        setsMode = mkOption {
+          description = "Change current mode after bind is executed";
+          type =
+            with types;
+            nullOr (enum [
+              "default"
+              "insert"
+              "paste"
+            ]);
+          default = null;
+        };
+        erase = mkEnableOption "remove bind";
+        silent = mkEnableOption "Operate silently";
+        operate = mkOption {
+          description = "Operate on preset bindings or user bindings";
+          type =
+            with types;
+            nullOr (enum [
+              "preset"
+              "user"
+            ]);
+          default = null;
+        };
       };
-      mode = mkOption {
-        description = "Specify the bind mode that the bind is used in";
-        type =
-          with types;
-          nullOr (enum [
-            "default"
-            "insert"
-            "paste"
-          ]);
-        default = null;
-      };
-      command = mkOption {
-        description = "command that will be execute";
-        type =
-          with types;
-          oneOf [
-            str
-            (listOf str)
-          ];
-      };
-      setsMode = mkOption {
-        description = "Change current mode after bind is executed";
-        type =
-          with types;
-          nullOr (enum [
-            "default"
-            "insert"
-            "paste"
-          ]);
-        default = null;
-      };
-      erase = mkEnableOption "remove bind";
-      silent = mkEnableOption "Operate silently";
-      operate = mkOption {
-        description = "Operate on preset bindings or user bindings";
-        type =
-          with types;
-          nullOr (enum [
-            "preset"
-            "user"
-          ]);
-        default = null;
-      };
-    };
-  };
+    }
+  );
 
   abbrsStr = lib.concatStringsSep "\n" (
     lib.mapAttrsToList (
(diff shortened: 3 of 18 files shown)


modules/programs/mc.nix
```diff
@@ -0,0 +1,45 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+with lib;
```
khaneliman:
```suggestion
```

modules/programs/mc.nix
```diff
@@ -0,0 +1,45 @@
...
+{
+  options.programs.mc = {
+    enable = mkEnableOption "Midnight Commander";
+
+    package = mkOption {
+      type = types.package;
+      default = pkgs.mc;
+      defaultText = literalExpression "pkgs.mc";
+      description = "The mc package to install.";
+    };
```
khaneliman:
```suggestion
    package = lib.mkPackageOption pkgs "mc" { nullable = true; }};
```

modules/programs/mc.nix
```diff
@@ -0,0 +1,45 @@
...
+
+    keymapConfig = mkOption {
+      type = types.lines;
+      default = "";
+      description = "Extra contents for ~/.config/mc/mc.keymap";
+    };
+  };
+
+  config = mkIf cfg.enable {
+    home.packages = [ cfg.package ];
```
khaneliman:
```suggestion
    home.packages = lib.mkIf (cfg.package != null) [ cfg.package ];
```

modules/programs/mc.nix
```diff
@@ -0,0 +1,45 @@
...
+
+    xdg.configFile = {
+      "mc/ini".text = cfg.extraConfig;
```
khaneliman:
```suggestion
      "mc/ini" = lib.mkIf (cfg.settings != {}) {
        text = lib.generators.toINI { } cfg.settings;
      };
```

modules/programs/mc.nix
```diff
@@ -0,0 +1,45 @@
...
+      "mc/mc.keymap".text = cfg.keymapConfig;
```
khaneliman:
```suggestion
      "mc/mc.keymap" = lib.mkIf (cfg.keymapSettings != {}) {
        text = lib.generators.toINI { } cfg.keymapSettings;
      };
```

modules/programs/mc.nix
```diff
@@ -0,0 +1,45 @@
...
+      default = pkgs.mc;
+      defaultText = literalExpression "pkgs.mc";
+      description = "The mc package to install.";
+    };
+
+    extraConfig = mkOption {
+      type = types.lines;
+      default = "";
+      description = "Extra contents for ~/.config/mc/ini";
+    };
```
khaneliman:
```suggestion
    settings = lib.mkOption {
      type = (pkgs.formats.ini { }).type;
      default = { };
      description = "Settings for `Midnight Commander`.";
    };
```

modules/programs/mc.nix
```diff
@@ -0,0 +1,45 @@
...
+
+    keymapConfig = mkOption {
+      type = types.lines;
+      default = "";
+      description = "Extra contents for ~/.config/mc/mc.keymap";
+    };
+  };
```
khaneliman:
```suggestion
    keymapSettings = lib.mkOption {
      type = (pkgs.formats.ini { }).type;
      default = { };
      description = "Settings for ~/.config/mc/mc.keymap";
    };
  };
```

tests/modules/programs/mc/basic-configuration.nix
```diff
@@ -0,0 +1,17 @@
...
+      skin=nicedark
+      show_hidden=true
+      auto_save_setup=true
+    '';
+
+    keymapConfig = ''
+      [panel]
+      Enter = Select
+    '';
+  };
```
khaneliman:
```suggestion
    settings = {
      "Midnight-Commander" = {
         skin = "nicedark";
         show_hidden = true;
         auto_save_setup = true;
      };
    };

    keymapSettings = {
      panel = {
        Enter = "Select";
      };
    };
  };
```

khaneliman (commented):
Initial thoughts, looks like they have a basic `ini` style configuration and we could just leverage the generators for it.

egorkonovalov (commented):
changes accepted

tests/modules/programs/mc/basic-configuration.nix
```diff
@@ -0,0 +1,19 @@
...
+      };
+    };
+
+    keymapSettings = {
+      panel = {
+        Enter = "Select";
+      };
+    };
+  };
+}
```
khaneliman, line 66: This test doesn't appear to be checking our generators are working, as expected. Need to add some kind of assertions to verify behavior.

egorkonovalov: Hi! How do I do that? Are there documentation/examples I can lean on?

egorkonovalov: Assertion tests added.

aguirre-matteo:
You could also add support for the others files I've mentioned in the issue, so users can declare their entire config using the module.

egorkonovalov:
@khaneliman

@aguirre-matteo suggested adding support for the remaining Midnight Commander config files.

Unlike `~/.config/mc/ini` and `~/.config/mc/mc.keymap`, the file `~/.config/mc/mc.ext.ini` does not fallback to system defaults if it’s incomplete. This means one needs to copy the default version from `/etc/mc/mc.ext.ini`, and then modify custom settings, otherwise the default won't work. Appending custom settings to the file won't work either because the GLib library used by mc merges two conflicting sections which may break the program's behavior. I'm not sure how to resolve this rn.

khaneliman:
> @khaneliman
> 
> @aguirre-matteo suggested adding support for the remaining Midnight Commander config files.
> 
> Unlike `~/.config/mc/ini` and `~/.config/mc/mc.keymap`, the file `~/.config/mc/mc.ext.ini` does not fallback to system defaults if it’s incomplete. This means one needs to copy the default version from `/etc/mc/mc.ext.ini`, and then modify custom settings, otherwise the default won't work. Appending custom settings to the file won't work either because the GLib library used by mc merges two conflicting sections which may break the program's behavior. I'm not sure how to resolve this rn.

Sounds similar to what `yazi` used to struggle with in its beginning. My opinion would be to add the options that control the files and give a description that explains that current behavior. If someone wants to use the module for those settings, they would just need to make sure it contains everything they want for configuration. I used to have to copy and maintain the upstream defaults in my yazi config until they allowed fallback to default / overriding behavior.

tests/modules/programs/mc/mc.ext.ini
```diff
@@ -0,0 +1,3 @@
+[EPUB]
+Shell=.epub
+Open=fbreader %f &
```
khaneliman: attributes will get generated in alphabetical order

egorkonovalov: That's surprising, thanks

modules/programs/mc.nix
```diff
@@ -0,0 +1,103 @@
...
+{
+  options.programs.mc = {
+    enable = lib.mkEnableOption "Midnight Commander";
+
+    package = lib.mkPackageOption pkgs "mc" { nullable = true; };
+
+    settings = lib.mkOption {
+      inherit type;
+      default = { };
+      description = "Settings for `mc/ini` file. Any missing settings will fall back to the system default.";
```
khaneliman:
```suggestion
      description = ''
        Settings for `mc/ini` file.
      
        Any missing settings will fall back to the system default.
       '';
```

Appreciate your patience. It would be nicer for legibility in docs if we used lines and used some newlines to separate thoughts for all these options.

egorkonovalov: Done

Dinth:
Hi @egorkonovalov. 
Maybe im doing something wrong (im new to nix), but it looks like mc.extensionSettings is generating mc.ext.ini file with  incorrect format
image

Maybe this example can help:
https://github.com/rossnomann/nixos/blob/467c5514f8ffc93a224d0bb5ed67c800ee8c863f/src/nih/lib/gen/mc.nix#L239

[edit]
It looks like just adding:
```
    extensionSettings = {
      "mc.ext.ini" = {
          Version = 4.0;
        };
```
before other rules fixes the problem. But i think it would be easy to automatically add it via flake if the mc version is 4.0+ , or at least document it in the flake.

egorkonovalov:
> Hi @egorkonovalov. Maybe im doing something wrong (im new to nix), but it looks like mc.extensionSettings is generating mc.ext.ini file with incorrect format image
> 
> Maybe this example can help: https://github.com/rossnomann/nixos/blob/467c5514f8ffc93a224d0bb5ed67c800ee8c863f/src/nih/lib/gen/mc.nix#L239
> 
> [edit] It looks like just adding:
> 
> ```
>     extensionSettings = {
>       "mc.ext.ini" = {
>           Version = 4.0;
>         };
> ```
> 
> before other rules fixes the problem. But i think it would be easy to automatically add it via flake if the mc version is 4.0+ , or at least document it in the flake.

Hi @Dinth,
`mc.extenionSettings` generates a `.ini` file with only the values you provide. Apparently, `Version` is required by mc and can't be omitted. However, I don't think implicitly setting the version number is a good idea because an empty `mc.ext.ini` would remove all default settings. That being said, I don't mind leaving a note about this in the documentation.

PR #7233, opened 2025-06-07
Cursor: darwin extensions dirPath

aspauldingcode:
### Description

I am trying to add programs.vscode.extensions setting to work on darwin when using code-cursor package.

So I'm updating home-manager vscode.nix to add code-cursor's darwin extensions path.
on darwin, cursor doesn't look at `~.config/cursor/extensions/`..
it also doesn't look at `~/Application\ Support/Cursor/extensions/`.
Cursor only looks for extensions in the `~/.cursor/` directory.

I have not tested this change! I don't know *how*

EDIT/UPDATE: I was on 24.11. Now on nixpkgs 25.05. And, the home-manager option needs `programs.vscode.profiles.<profilename>.extensions` which appears to work out of the box.
I don't know why, but I guess so. meaning my PR seems pointless at the moment...?

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/vscode.nix
+++ b/modules/programs/vscode.nix
@@ -59,8 +59,10 @@ let
 
   snippetDir = name: "${userDir}/${optionalString (name != "default") "profiles/${name}/"}snippets";
 
-  # TODO: On Darwin where are the extensions?
-  extensionPath = ".${extensionDir}/extensions";
+  extensionPath = if pkgs.stdenv.hostPlatform.isDarwin && vscodePname == "cursor" then
+    ".cursor/extensions"
+  else
+    ".${extensionDir}/extensions";
 
   extensionJson = ext: pkgs.vscode-utils.toExtensionJson ext;
   extensionJsonFile =


modules/programs/vscode.nix
```diff
@@ -59,8 +59,10 @@ let
 
   snippetDir = name: "${userDir}/${optionalString (name != "default") "profiles/${name}/"}snippets";
 
-  # TODO: On Darwin where are the extensions?
-  extensionPath = ".${extensionDir}/extensions";
+  extensionPath = if pkgs.stdenv.hostPlatform.isDarwin && vscodePname == "cursor" then
+    ".cursor/extensions"
+  else
+    ".${extensionDir}/extensions";
```
khaneliman, line 65: This isn't changing anything from what I can see. We determine the path above which has a mapping from `cursor` to `cursor`. 

`".${extensionDir}/extensions"` should equal `".cursor/extensions"` just like the condition you added

aspauldingcode: Oh, right. Whoops

aspauldingcode: Okay, I've tested my changes.

This does actually properly install the cursor extensions on a darwin system.
Screenshot 2025-06-08 at 2 40 11 PM

All I did was update to nixpkgs 25.05.
I don't know what changed. 
It seems to work without my changes too.

khaneliman: Is the file not already there without the changes ?

aspauldingcode:
okay I guess this was not needed then???

PR #7248, opened 2025-06-09
meli: support "include" statement

teto:
so one can mix nix generated and manual configs

### Description

This is one thing I missed from my own module. @MunsMan 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/meli.nix
+++ b/modules/programs/meli.nix
@@ -13,6 +13,10 @@ let
     mkIf
     ;
 
+  cfg = config.programs.meli;
+
+  tomlFormat = pkgs.formats.toml { };
+
   enabledAccounts = lib.attrsets.filterAttrs (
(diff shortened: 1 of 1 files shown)


teto:
due to the recent IFD interdiction (which is cool), my initial attempt using readFile on the generated file wouldn't work. 
So instead I included the generated config. Because it's less readable I disabled it when not using `includes`. This feels a bit hackish but this is the best way I could find to keep the best of both worlds.

teto:
I will abuse my maintainer hat and merge this. Maybe I should add myself as maintainer of the module, anyway I intend to deal with the fallout

MunsMan:
Works for me, feel free to add yourself as a maintainer.

I'm currently working on more improvements on the options regarding the setup of the mail backend. I will create a PR if ready.

PR #7252, opened 2025-06-11
atuin: fix docs url

punsii:
### Description

This change fixes the URLs for the atuin docs since the old ones either redirect or land on a 404 page.
It only touches option descriptions and does not include any functional changes.

### Maintainers

@hawkw
@water-sucks

### DIFF
--- a/modules/programs/atuin.nix
+++ b/modules/programs/atuin.nix
@@ -84,7 +84,7 @@ in
         Configuration written to
         {file}`$XDG_CONFIG_HOME/atuin/config.toml`.
 
-        See <https://atuin.sh/docs/config/> for the full list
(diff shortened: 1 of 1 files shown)


modules/programs/atuin.nix
```diff
@@ -84,7 +84,7 @@ in
         Configuration written to
         {file}`$XDG_CONFIG_HOME/atuin/config.toml`.
 
-        See <https://atuin.sh/docs/config/> for the full list
+        See <https://docs.atuin.sh/configuration/config//> for the full list
```
khaneliman:
```suggestion
        See <https://docs.atuin.sh/configuration/config/> for the full list
```

PR #7255, opened 2025-06-11
nushell: add plugins as home packages

### LINKED ISSUE #7070: bug: nominated nushell plugins' packages get collected as garbage

### Issue description

In my Nix flake, I configure Home Manager to install several _nushell_ plugins but setting the "programs.nushell.plugins" list to include, say, `pkg.nushellPlugins.polars`. When I do so, Home Manager takes care of building that plugin, and when I run the _plugin list_ within _nushell_, I can see that it knows about this plugin, and cites a path to it such as _/nix/store/749jj54r9h4rlvxg5d008vvlb84rgwkl-nushell_plugin_polars-0.104.0/bin/nu_plugin_polars_.

However, if I try to have _nushell_ actually use the plugin, it fails to load it:
```
Error:   × Unable to spawn plugin for `polars into-df`
   ╭─[entry #1:1:6]
 1 │ ls | polars into-df | describe
   ·      ───────┬──────
   ·             ╰── Plugin failed to load: Can't find /nix/store/749jj54r9h4rlvxg5d008vvlb84rgwkl-nushell_plugin_polars-0.104.0/bin/nu_plugin_polars, please make sure that /nix/store/749jj54r9h4rlvxg5d008vvlb84rgwkl-nushell_plugin_polars-0.104.0/bin/nu_plugin_polars is in PATH.
   ╰────
```
At that point, I can confirm that the _/nix/store/749jj54r9h4rlvxg5d008vvlb84rgwkl-nushell_plugin_polars-0.104.0_ directory does not exist.

I run the _nix-collect-garbage_ command regularly, and even though Home Manager ensures that a referenced package such as "nushellPlugins.polars" gets installed, it doesn't "pin it" in place, leaving the garbage collector to find it as an orphan, eligible for collection.

_Should_ Home Manager take care of preserving such referenced plugin modules? I expect that I could request that these packages get installed explicitly by nominating them in the "home.packages" attribute, but that leads to unfortunate repetition, leading to potential inconsistency.

### Maintainer CC

@Philipp-M
@aidalgol
@joaquintrinanes

### System information

```markdown
- system: `"x86_64-darwin"`
- host os: `Darwin 24.5.0, macOS 15.5`
- multi-user?: `yes`
- sandbox: `no`
- version: `nix-env (Nix) 2.28.3`
- channels(root): `"nixpkgs"`
- nixpkgs: `/nix/store/a4lkx9hmv2099n6p1rkzxm481hfw7pwr-source`
```

ealap:
### Description

- Fix #7070 by adding plugins to home.packages
- Reorganize tests for nushell
- Add new test for scenario without a plugin: check if default plugin config file does not exist
- Add new test for scenario with at least one plugin: check if it has 'loaded' status and filename is same as home.package install path which ensures us that it is a valid path.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@Philipp-M
@aidalgol
@JoaquinTrinanes


aidalgol (approved):
Looks good to me, although I admit I still don't fully understand why the derivations are garbage-collected without this.

PR #7257, opened 2025-06-12
element-desktop: combine configs

nim65s:
### Description

This allow users to define element-desktop config as an override of the default one

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

cc @aguirre-matteo

### DIFF
--- a/modules/programs/element-desktop.nix
+++ b/modules/programs/element-desktop.nix
@@ -26,6 +26,20 @@ in
   options.programs.element-desktop = {
     enable = mkEnableOption "element-desktop";
     package = mkPackageOption pkgs "element-desktop" { nullable = true; };
+    combineDefaultSettings = mkOption {
+      type = lib.types.bool;
+      default = false; # true would be better, but a breaking change
+      description = ''
+        Use element-web default configuration as a basis for settings and profiles.
+      '';
+    };
+    combineSettingsProfiles = mkOption {
+      type = lib.types.bool;
+      default = false; # true would be better, but a breaking change
+      description = ''
+        Use settings as a basis for profiles.
+      '';
+    };
     settings = mkOption {
       type = formatter.type;
       default = { };
@@ -52,9 +66,7 @@ in
       '';
       description = ''
         Configuration settings for Element's default profiles.
-        WARNING: Element doesn't combine this config with the defaults,
-        so make sure to configure most options. For details about this
-        behavior and available configuration options see:
+        For details about this behavior and available configuration options see:
         <https://github.com/element-hq/element-web/blob/develop/docs/config.md>
       '';
     };
@@ -90,8 +102,7 @@ in
       '';
       description = ''
         Extra profiles for Element. Those can be accessed using the
-        "--profile $NAME" flag. The same warning and options apply
-        here.
+        "--profile $NAME" flag. The same options apply here.
       '';
     };
   };
@@ -100,12 +111,22 @@ in
     home.packages = mkIf (cfg.package != null) [ cfg.package ];
     home.file =
       let
+        settings =
+          if cfg.combineDefaultSettings then
+            (pkgs.runCommandLocal "element-desktop-default"
+              {
+                nativeBuildInputs = [ pkgs.jq ];
+              }
+              ''
+                jq -s '.[0] * $conf' "${cfg.package}/share/element/webapp/config.json" --argjson "conf" ${lib.escapeShellArg (builtins.toJSON cfg.settings)} > $out
+              ''
+            )
+          else
+            (formatter.generate "element-desktop-default" cfg.settings);
         defaultConfig =
-          if (cfg.settings != { }) then
+          if (settings != { }) then
             {
-              "${prefix}/Element/config.json".source = (
-                formatter.generate "element-desktop-default" cfg.settings
-              );
+              "${prefix}/Element/config.json".source = settings;
             }
           else
             { };
@@ -113,8 +134,24 @@ in
       defaultConfig
       // (lib.mapAttrs' (
         name: value:
+        let
+          profile =
+            if cfg.combineSettingsProfiles then
+              (pkgs.runCommandLocal "element-desktop-${name}"
+                {
+                  nativeBuildInputs = [ pkgs.jq ];
+                }
+                ''
+                  jq -s '.[0] * $conf' "${settings}" --argjson "conf" ${
+                    lib.escapeShellArg (builtins.toJSON cfg.profiles."${name}")
+                  } > $out
+                ''
+              )
+            else
+              (formatter.generate "element-desktop-${name}" cfg.profiles."${name}");
+        in
         lib.nameValuePair "${prefix}/Element-${name}/config.json" {
-          source = (formatter.generate "element-desktop-${name}" cfg.profiles."${name}");
+          source = profile;
         }
       ) cfg.profiles);
   };
--- a/modules/programs/element-desktop.nix
+++ b/modules/programs/element-desktop.nix
@@ -124,7 +124,7 @@ in
           else
             (formatter.generate "element-desktop-default.json" cfg.settings);
         defaultConfig =
-          if (settings != { }) then
+          if (cfg.settings != { } || cfg.combineDefaultSettings) then
             {
               "${prefix}/Element/config.json".source = settings;
             }
--- a/modules/programs/element-desktop.nix
+++ b/modules/programs/element-desktop.nix
@@ -113,7 +113,7 @@ in
       let
         settings =
           if cfg.combineDefaultSettings then
-            (pkgs.runCommandLocal "element-desktop-default"
+            (pkgs.runCommandLocal "element-desktop-default.json"
               {
                 nativeBuildInputs = [ pkgs.jq ];
               }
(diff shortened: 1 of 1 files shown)


modules/programs/element-desktop.nix
```diff
@@ -100,12 +111,22 @@ in
...
+              }
+              ''
+                jq -s '.[0] * $conf' "${cfg.package}/share/element/webapp/config.json" --argjson "conf" ${lib.escapeShellArg (builtins.toJSON cfg.settings)}
+              ''
+            )
+          else
+            (formatter.generate "element-desktop-default" cfg.settings);
         defaultConfig =
-          if (cfg.settings != { }) then
+          if (settings != { }) then
```
aguirre-matteo: Won't this make HM create the config file even if the user didn't manually set `programs.element-desktop.settings`? I don't think we should create the file if the user didn't said to do so.

aguirre-matteo:
I'll check it out when I go home.

modules/programs/element-desktop.nix
```diff
@@ -26,6 +26,20 @@ in
   options.programs.element-desktop = {
     enable = mkEnableOption "element-desktop";
     package = mkPackageOption pkgs "element-desktop" { nullable = true; };
+    combineDefaultSettings = mkOption {
+      type = lib.types.bool;
+      default = false; # true would be better, but a breaking change
```
aguirre-matteo, line 31: Why do you think it could be a breaking change?

modules/programs/element-desktop.nix
```diff
@@ -100,12 +111,22 @@ in
...
+            (pkgs.runCommandLocal "element-desktop-default"
+              {
+                nativeBuildInputs = [ pkgs.jq ];
+              }
+              ''
+                jq -s '.[0] * $conf' "${cfg.package}/share/element/webapp/config.json" --argjson "conf" ${lib.escapeShellArg (builtins.toJSON cfg.settings)} > $out
+              ''
+            )
+          else
+            (formatter.generate "element-desktop-default" cfg.settings);
```
aguirre-matteo: The first case returns a string I guess, while the second returns a path. This will result in HM creating a file with only one line which is a path to the true config file (you can see so in the error message in the failed tests). Since `cfg.combineDefaultSettings` is set to false by default, this would break existing configs. We should instead use a generator, since they return an string instead of a path.

modules/programs/element-desktop.nix
```diff
@@ -114,7 +135,9 @@ in
       // (lib.mapAttrs' (
         name: value:
         lib.nameValuePair "${prefix}/Element-${name}/config.json" {
-          source = (formatter.generate "element-desktop-${name}" cfg.profiles."${name}");
+          source = formatter.generate "element-desktop-${name}" (
+            if cfg.combineSettingsProfiles then (settings // cfg.profiles."${name}") else cfg.profiles."${name}"
+          );
```
aguirre-matteo: Assuming `settings` is either a path or a string, this would fail since the `//` operator expects two attribute sets.

modules/programs/element-desktop.nix
nim65s: Because it will change the default behavior. This is not super breaking, but I guess there might exist an user who would rely on the fact that the generated config did not include any default.

I would be happy to be wrong here :)

aguirre-matteo: Sorry, you're right. I didn't thought about such scenario.

modules/programs/element-desktop.nix
nim65s: I'm sorry, this was a first draft and not tested.

Now, after a few force push and before I saw your (very fast, thanks !) review, I have a correct thing, ie:
```
$ readlink -f ~/.config/Element*/config.json
/nix/store/lkqapk1yjhx3kgrfwsxg3w8mmnyn7wdb-element-desktop-ttnn
/nix/store/bzmg3wa14vkg9964wmfmcykc1l57amx2-element-desktop-default
```

and those contain a correct configuration.

I tested that with and without the 2 new booleans, and with settings and a secondary profile.

So I am happy with this, and will mark the PR at ready to review, but we can continue discussing returning strings instead of paths if you wish. I just don't know how to combine things with a generator.

modules/programs/element-desktop.nix
nim65s: Yes, thanks !

modules/programs/element-desktop.nix
aguirre-matteo: If you can please update the tests for Element. In any case, I'll try to update them later because rn I have some things to do.

modules/programs/element-desktop.nix
aguirre-matteo: Feel free to mark this as resolved.

nim65s:
Hum, I tried to add a test, but my `jq … "${cfg.package}/…/config.json"` is turned into litteral `jq … @element-desktop@/…/config.json` and:
```
element-desktop-default.json> jq: error: Could not open file @element-desktop@/share/element/webapp/config.json: No such file or directory
```

I'll give this another try later, and I'll look deeper into nmt. But if anybody has a better understanding of this and can easily provide a quick clue, that will help :)

aguirre-matteo:
> Hum, I tried to add a test, but my `jq \u2026 "${cfg.package}/\u2026/config.json"` is turned into litteral `jq \u2026 @element-desktop@/\u2026/config.json` and:
> 
> ```
> element-desktop-default.json> jq: error: Could not open file @element-desktop@/share/element/webapp/config.json: No such file or directory
> ```
> 
> I'll give this another try later, and I'll look deeper into nmt. But if anybody has a better understanding of this and can easily provide a quick clue, that will help :)

We could try using `builtins.fromJSON` instead of relying on `jq` to do the job. You could do something like this:

```nix
        settings =
          if cfg.combineDefaultSettings then
          let
            pathElementDefaultCfg = /. + "${cfg.package}/share/element/webapp/config.json";
            elementDefaultSettings = builtins.fromJSON (builtins.readFile pathElementDefaultCfg);
            finalSettings = elementDefaultSettings // cfg.settings;
          in (formatter.generate "element-desktop-config" finalSettings)
          else
            (formatter.generate "element-desktop-config" cfg.settings);
```

Notice that we should also add an assertion to ensure `cfg.package` is not null when either `cfg.combineDefaultSettings` or `cfg.combineSettingsProfile` is set to true, since HM won't be able to get the path to the package.

aguirre-matteo:
By the way, I think `combineSettingsProfiles` should be renamed to `combineProfilesSettings`

aguirre-matteo:
I'm having the same issue you mentioned. `cfg.package` is replaced to `@element-desktop@` which is strange. I'll tag you if I find a solution.

aguirre-matteo:
It seems the issue occurs because Nix doesn't automatically resolve the placeholder `@element-desktop@` to a Nix store path. I'll try to fix that, but the module could become a bit chunky.

modules/programs/element-desktop.nix
```diff
@@ -26,6 +26,20 @@ in
   options.programs.element-desktop = {
     enable = mkEnableOption "element-desktop";
     package = mkPackageOption pkgs "element-desktop" { nullable = true; };
+    combineDefaultSettings = mkOption {
+      type = lib.types.bool;
+      default = false; # true would be better, but a breaking change
+      description = ''
+        Use element-web default configuration as a basis for settings and profiles.
```
teto, line 33: do you mean nix defaults or the program's ?

teto (approved):
I am not a fan of the added complexity here. If upstream is not suitable for declarative configurations, then let's modify upstream or make that an out of tree home-manager module. The bandwidth for HM reviewers is small (that's one issue) and dealing with jq calls that dont fit the latest element config update is not what I want to spend time with.
Maybe expose these options into your flake first to see how it behaves ?

aguirre-matteo:
> I am not a fan of the added complexity here. If upstream is not suitable for declarative configurations, then let's modify upstream or make that an out of tree home-manager module. The bandwidth for HM reviewers is small (that's one issue) and dealing with jq calls that dont fit the latest element config update is not what I want to spend time with.
> Maybe expose these options into your flake first to see how it behaves ?

Yeah, the problem comes from upstream, so it would be better to open a PR on element-desktop's repo to make the app use the defaults if they're not present in the config file.
Personally, I don't care about it too much, but I'm sure we shouldn't rely on jq to do the job.

nim65s:
I stole that `jq` use from [nixpkgs](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/by-name/el/element-web-unwrapped/package.nix#L60)

It was added in https://github.com/NixOS/nixpkgs/pull/80364, and I've been using that to configure my element-desktop for a while, before element-desktop module was added to HM. It didn't break, but that forced me to rebuild element-desktop every time :/

That said, I agree a solution without jq would be better.

>  Maybe expose these options into your flake first to see how it behaves ?

I'm already doing this with success :)
(more precisely, my HM flake input is patched with this PR through https://github.com/jfly/flake-input-patcher))

aguirre-matteo:
I'll still keep looking for a way to resolve the path these days, without having to write a derivation and use `substituteAll` since it would be ugly to read. Do you know a way to resolve the path in a single line? @teto

aguirre-matteo:
For now, @nim65s, mark this PR as draft.

nim65s:
can't do it right now, feel free to pick this up as needed

PR #7259, opened 2025-06-12
meli: move freeformat settings to example

teto:
these hardcoded freeformat settings dont add anything over the default, worse they generate empty sections which can then clash with user included config (my usecase). It's best to move those to example.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/meli.nix
+++ b/modules/programs/meli.nix
@@ -84,65 +84,36 @@ in
 
       settings = mkOption {
         type = types.submodule {
-          freeformType = (pkgs.formats.toml { }).type;
-          options = {
-            shortcuts = mkOption {
-              type = types.submodule {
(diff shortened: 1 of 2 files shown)


teto:
@khaneliman you've added meli to `dev-tools` in commit `ci: labels dedupe and reorganize`, which strikes me as an odd-choice. I would rather have it labeld with "email" ?

khaneliman:
> @khaneliman you've added meli to `dev-tools` in commit `ci: labels dedupe and reorganize`, which strikes me as an odd-choice. I would rather have it labeld with "email" ?

Can push  commit to change it. There was a lot of shuffling around in there

khaneliman (approved):

PR #7262, opened 2025-06-13
k9s: fix config files not generated correctly & improve options

ojsef39:
### Description

i noticed that the config files aren't like k9s expects them when you follow the description:
```yaml
alias:
  cr: clusterroles
```
instead of the correct:
```yaml
aliases:
  cr: clusterroles
```
this made me change this:
```nix
    aliases = {
      alias = {
        pp = "v1/pods";
      };
    };
```
to just this, to cause less confusion about the aliases.alias (which is not working anyway):
```nix
    aliases = {
      pp = "v1/pods";
    };
```

Thank you very much for taking this PR into consideration, i know it's not really backwards compatible, but i think this change makes sense nonetheless.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@liyangau 
@LucasWagler


modules/programs/k9s.nix
```diff
@@ -86,14 +89,14 @@ in
         <https://k9scli.io/topics/aliases/> for supported values.
       '';
       example = literalExpression ''
-        alias = {
+        aliases = {
```
teto: in general literalExpression represents just the value so in this case it would be:

```
{
          # Use pp as an alias for Pod
          pp = "v1/pods";
        };
```

same remark for other examples

ojsef39: Hey, thank you very much for your review, hope i didn't miss anything :)

modules/programs/k9s.nix
```diff
@@ -86,14 +89,12 @@ in
         <https://k9scli.io/topics/aliases/> for supported values.
       '';
       example = literalExpression ''
-        alias = {
-          # Use pp as an alias for Pod
-          pp = "v1/pods";
-        };
+        # Use pp as an alias for Pod
```
teto: you are missing the braces. Some would set "hotkeys = VAL" with VAL be the value shown in example aka:
```
{
  pp = ...
}
```

ojsef39: Ups, my bad, thanks, fixed it :)

teto (approved):

modules/programs/k9s.nix
ojsef39: Thanks for reviewing @teto

modules/programs/k9s.nix
```diff
@@ -214,19 +222,19 @@ in
           };
 
           "k9s/aliases.yaml" = mkIf (cfg.aliases != { }) {
-            source = yamlFormat.generate "k9s-aliases" cfg.aliases;
+            source = yamlFormat.generate "k9s-aliases" { aliases = cfg.aliases; };
```
khaneliman:
```suggestion
            source = yamlFormat.generate "k9s-aliases" { inherit (cfg) aliases; };
```

modules/programs/k9s.nix
```diff
@@ -214,19 +222,19 @@ in
...
           };
 
-          "k9s/hotkeys.yaml" = mkIf (cfg.hotkey != { }) {
-            source = yamlFormat.generate "k9s-hotkey" cfg.hotkey;
+          "k9s/hotkeys.yaml" = mkIf (cfg.hotkeys != { }) {
+            source = yamlFormat.generate "k9s-hotkeys" { hotKeys = cfg.hotkeys; };
```
khaneliman:
```suggestion
            source = yamlFormat.generate "k9s-hotkeys" { inherit (cfg) hotkeys; };
```

modules/programs/k9s.nix
```diff
@@ -214,19 +222,19 @@ in
...
           };
 
-          "k9s/plugins.yaml" = mkIf (cfg.plugin != { }) {
-            source = yamlFormat.generate "k9s-plugin" cfg.plugin;
+          "k9s/plugins.yaml" = mkIf (cfg.plugins != { }) {
+            source = yamlFormat.generate "k9s-plugins" { plugins = cfg.plugins; };
```
khaneliman:
```suggestion
            source = yamlFormat.generate "k9s-plugins" { inherit (cfg) plugins; };
```

modules/programs/k9s.nix
```diff
@@ -214,19 +222,19 @@ in
...
           };
 
           "k9s/views.yaml" = mkIf (cfg.views != { }) {
-            source = yamlFormat.generate "k9s-views" cfg.views;
+            source = yamlFormat.generate "k9s-views" { views = cfg.views; };
```
khaneliman:
```suggestion
            source = yamlFormat.generate "k9s-views" { inherit (cfg) views; };
```

modules/programs/k9s.nix
```diff
@@ -239,19 +247,19 @@ in
           };
 
           "Library/Application Support/k9s/aliases.yaml" = mkIf (cfg.aliases != { }) {
-            source = yamlFormat.generate "k9s-aliases" cfg.aliases;
+            source = yamlFormat.generate "k9s-aliases" { aliases = cfg.aliases; };
```
khaneliman:
```suggestion
            source = yamlFormat.generate "k9s-aliases" { inherit (cfg) aliases; };
```

modules/programs/k9s.nix
```diff
@@ -239,19 +247,19 @@ in
...
           };
 
-          "Library/Application Support/k9s/hotkeys.yaml" = mkIf (cfg.hotkey != { }) {
-            source = yamlFormat.generate "k9s-hotkey" cfg.hotkey;
+          "Library/Application Support/k9s/hotkeys.yaml" = mkIf (cfg.hotkeys != { }) {
+            source = yamlFormat.generate "k9s-hotkeys" { hotKeys = cfg.hotkeys; };
```
khaneliman:
```suggestion
            source = yamlFormat.generate "k9s-hotkeys" { inherit (cfg) hotkeys; };
```

modules/programs/k9s.nix
```diff
@@ -239,19 +247,19 @@ in
...
           };
 
-          "Library/Application Support/k9s/plugins.yaml" = mkIf (cfg.plugin != { }) {
-            source = yamlFormat.generate "k9s-plugin" cfg.plugin;
+          "Library/Application Support/k9s/plugins.yaml" = mkIf (cfg.plugins != { }) {
+            source = yamlFormat.generate "k9s-plugins" { plugins = cfg.plugins; };
```
khaneliman:
```suggestion
            source = yamlFormat.generate "k9s-plugins" { inherit (cfg) plugins; };
```

modules/programs/k9s.nix
```diff
@@ -239,19 +247,19 @@ in
...
           };
 
           "Library/Application Support/k9s/views.yaml" = mkIf (cfg.views != { }) {
-            source = yamlFormat.generate "k9s-views" cfg.views;
+            source = yamlFormat.generate "k9s-views" { views = cfg.views; };
```
khaneliman:
```suggestion
            source = yamlFormat.generate "k9s-views" { inherit (cfg) views; };
```

modules/programs/k9s.nix
```diff
@@ -33,6 +33,9 @@ in
         "skin"
       ]
     )
+    (lib.mkRenamedOptionModule [ "programs" "k9s" "hotkey" ] [ "programs" "k9s" "hotkeys" ])
+    (lib.mkRenamedOptionModule [ "programs" "k9s" "plugin" ] [ "programs" "k9s" "plugins" ])
+    (lib.mkRenamedOptionModule [ "programs" "k9s" "view" ] [ "programs" "k9s" "views" ])
```
khaneliman:
```suggestion
```
This option doesn't appear to have existed?

tests/modules/programs/k9s/deprecated-options.nix
```diff
@@ -16,11 +16,69 @@
...
+      };
+    };
+    view = {
+      "v1/services" = {
+        columns = [
+          "NAME"
+          "TYPE"
+        ];
+      };
+    };
```
khaneliman:
```suggestion
```
Looks like this is only created because of the `mkRenamedOptionModule` but I didn't see a rename in the main module, looks like it was already `views`

ojsef39: Thank you very much for your review and the suggestions!

I guess i confused myself there, and added view as well :)

ojsef39:
Thank you for reviewing and merging :)

PR #7263, opened 2025-06-13
kubeswitch: add module

M0NsTeRRR:
### Description

Add [kubeswitch](https://github.com/danielfoehrKn/kubeswitch) module.
Based on https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/programs/kubeswitch.nix

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/lib/maintainers.nix
```diff
@@ -515,6 +515,12 @@
     github = "miku4k";
     githubId = 89653242;
   };
+  m0nsterrr = {
+    name = "Ludovic Ortega";
+    email = "nix@mail.adminafk.fr";
+    github = "M0NsTeRRR";
+    githubId = 37785089;
+  };
```
khaneliman, line 531: separate commit, please.

M0NsTeRRR: Done :)

modules/programs/kubeswitch.nix
```diff
@@ -0,0 +1,120 @@
...
+        };
+      }
+
+      (lib.mkIf cfg.enableBashIntegration (
+        let
+          shell_files = pkgs.runCommand "kubeswitch-shell-files" { buildInputs = [ cfg.package ]; } ''
+            mkdir -p $out/share
+            switcher init bash | sed 's/switch(/${cfg.commandName}(/' > $out/share/${cfg.commandName}_init.bash
+            switcher --cmd ${cfg.commandName} completion bash > $out/share/${cfg.commandName}_completion.bash
+          '';
```
khaneliman: Sorry, should have commented on this in the first review pass. Would be nice to make all these repeated blocks a single function that generates from a `shell` input. 

```nix
generateKubeswitchShellFiles = shell:
    pkgs.runCommand "kubeswitch-${cfg.commandName}-shell-files-for-${shell}" {
      nativeBuildInputs = [ cfg.package ];
    } ''
      mkdir -p $out/share
      switcher init "${shell}" | sed "s/switch(/${cfg.commandName}(/" > "$out/share/${cfg.commandName}_init.$shell"
      switcher --cmd "${cfg.commandName}" completion "$shell" > "$out/share/${cfg.commandName}_completion.${shell}"
    '';
```

Then you could just have a single variable in each shell 

```nix
  kubeswitchBashFiles = generateKubeswitchShellFiles "bash";
```

M0NsTeRRR: No problem :)
Thanks for your help with my first PR related to Nix*.
I didn’t factor out the Fish snippet as it’s a little bit different from Bash and Zsh

modules/programs/kubeswitch.nix
```diff
@@ -0,0 +1,121 @@
...
+        {
+          programs.zsh.initContent = ''
+            source ${kubeswitchZshFiles}/share/${cfg.commandName}_init.zsh
+            source ${kubeswitchZshFiles}/share/${cfg.commandName}_completion.zsh
+          '';
+        }
+      ))
+    ]
+  );
+}
```
khaneliman:
```suggestion
   config = lib.mkIf cfg.enable {
    home = {
      packages = [ cfg.package ];

      file.".kube/switch-config.yaml" = lib.mkIf (cfg.settings != { }) {
        source = yamlFormat.generate "kubeswitch-settings" cfg.settings;
      };
    };

    programs =
      let
        generateKubeswitchShellFiles =
          shell:
          pkgs.runCommand "kubeswitch-${cfg.commandName}-shell-files-for-${shell}"
            {
              nativeBuildInputs = [ cfg.package ];
            }
            ''
              mkdir -p $out/share
              switcher init "${shell}" | sed "s/switch(/${cfg.commandName}(/" > "$out/share/${cfg.commandName}_init.${shell}"
              switcher --cmd "${cfg.commandName}" completion "${shell}" > "$out/share/${cfg.commandName}_completion.${shell}"
            '';
      in
      {
        bash.initExtra =
          let
            kubeswitchBashFiles = generateKubeswitchShellFiles "bash";
          in
          lib.mkIf cfg.enableBashIntegration ''
            source ${kubeswitchBashFiles}/share/${cfg.commandName}_init.bash
            source ${kubeswitchBashFiles}/share/${cfg.commandName}_completion.bash
          '';

        fish.interactiveShellInit =
          let
            kubeswitchFishFiles = generateKubeswitchShellFiles "fish";
          in
          lib.mkIf cfg.enableFishIntegration ''
            source ${kubeswitchFishFiles}/share/${cfg.commandName}_init.fish
            source ${kubeswitchFishFiles}/share/${cfg.commandName}_completion.fish
          '';

        zsh.initContent =
          let
            kubeswitchZshFiles = generateKubeswitchShellFiles "zsh";
          in
          lib.mkIf cfg.enableZshIntegration ''
            source ${kubeswitchZshFiles}/share/${cfg.commandName}_init.zsh
            source ${kubeswitchZshFiles}/share/${cfg.commandName}_completion.zsh
          '';

      };
  };

```

Would be cleaner to consolidate it a bit more out of a `mkMerge` since we're not really needing that here.

modules/programs/kubeswitch.nix
khaneliman: What's different about fish? Looking at it it appears to be the same code

modules/programs/kubeswitch.nix
M0NsTeRRR: I committed your change and regrouped these two into one, as I don't want to take ownership of your rewrite suggestion :)

modules/programs/kubeswitch.nix
```diff
@@ -0,0 +1,112 @@
...
+            source ${kubeswitchBashFiles}/share/${cfg.commandName}_init.bash
+            source ${kubeswitchBashFiles}/share/${cfg.commandName}_completion.bash
+          '';
+
+        fish.interactiveShellInit =
+          let
+            shell_files = pkgs.runCommand "kubeswitch-shell-files" { buildInputs = [ cfg.package ]; } ''
+              mkdir -p $out/share
+              switcher init fish | sed 's/switch(/${cfg.commandName}(/' > $out/share/${cfg.commandName}_init.fish
+              switcher --cmd ${cfg.commandName} completion fish > $out/share/${cfg.commandName}_completion.fish
```
khaneliman: Is this supposed to be different from https://github.com/nix-community/home-manager/pull/7263/files#diff-f3541d73f88e4348975ba3162fac769a8a9524422c3577b6d4d4fdd2f0de4b64R74-R76 somehow?

M0NsTeRRR: I fixed it yes sorry, the sed is a bit different as the documentation of kubeswitch mention it.
https://github.com/danielfoehrKn/kubeswitch/blob/master/docs/installation.md#fish
> Fish shell have a built-in switch function. Hence, differently from zsh shells, the kubeswitch function is called kubeswitch.

khaneliman: Oh okay thanks. Makes more sense. Probably could have a second argument for what we're replacing to make it work with function. But don't need to change again.

M0NsTeRRR: Okay, thanks for your kind review ! :)

PR #7264, opened 2025-06-13
ncspot: adopt, make package nullable

awwpotato:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/ncspot.nix
+++ b/modules/programs/ncspot.nix
@@ -10,7 +10,7 @@ let
   tomlFormat = pkgs.formats.toml { };
 in
 {
-  meta.maintainers = [ ];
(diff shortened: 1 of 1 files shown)


modules/programs/ncspot.nix
```diff
@@ -5,17 +5,19 @@
   ...
 }:
 let
+  inherit (lib) mkIf;
```
teto, line 8: while I dont feel strongly about this, I wonder if @khaneliman is ok with that ? since he removed the `with lib`.

khaneliman: Inherit is the static analysis supported method of doing it.

PR #7274, opened 2025-06-15
anki: add module

JuneStepp:
### Description

Anki is a popular spaced repetition flashcards learning program. This PR introduces a module for configuring Anki settings and add-ons. It is the counterpart to my [nixpkgs PR](https://github.com/NixOS/nixpkgs/pull/404045) that has now reached unstable. ~~Home Manager's flake.lock is currently too out of date for this. #7273~~

Because of how Anki's configuration is stored, I was not able to add anything like `extraConfig`. That's why there are so many options. I've used official Anki APIs wherever possible to prevent breakage; it has worked for many months for me without issue.

Addons added with `programs.anki.addons` don't seem to show up when Anki is built via tests but do in normal usage. I'm not sure why.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/anki/helper.nix
+++ b/modules/programs/anki/helper.nix
@@ -1,0 +1,261 @@
+{
+  lib,
+  config,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.programs.anki;
+  # This script generates the Anki SQLite settings DB using the Anki Python API.
+  # The configuration options in the SQLite database take the form of Python
+  # Pickle data.
+  # A simple "gldriver6" file is also generated for the `videoDriver` option.
+  buildAnkiConfig = pkgs.writers.writeText "buildAnkiConfig" ''
+    import sys
+
+    from aqt.profiles import ProfileManager, VideoDriver
+    from aqt.theme import Theme, WidgetStyle, theme_manager
+    from aqt.toolbar import HideMode
+
+    profile_manager = ProfileManager(
+      ProfileManager.get_created_base_folder(sys.argv[1])
+    )
+    _ = profile_manager.setupMeta()
+    profile_manager.meta["firstRun"] = False
+
+    # Video driver. Option is stored in a separate file from other options.
+    video_driver_str: str = "${toString cfg.videoDriver}"
+    if video_driver_str:
+      # The enum value for OpenGL isn't "opengl"
+      if video_driver_str == "opengl":
+        video_driver = VideoDriver.OpenGL
+      else:
+        video_driver = VideoDriver(video_driver_str)
+      profile_manager.set_video_driver(video_driver)
+
+
+    # Shared options
+
+    lang_tag: str = "${toString cfg.language}"
+    if lang_tag:
+      profile_manager.setLang(lang_tag)
+
+    theme_str: str = "${toString cfg.theme}"
+    if theme_str:
+      theme: Theme = {
+        "followSystem": Theme.FOLLOW_SYSTEM,
+        "light": Theme.LIGHT,
+        "dark": Theme.DARK
+      }[theme_str]
+      profile_manager.set_theme(theme)
+
+    style_str: str = "${toString cfg.style}"
+    if style_str:
+      style: WidgetStyle = {
+        "anki": WidgetStyle.ANKI, "native": WidgetStyle.NATIVE
+      }[style_str]
+      # Fix error from there being no main window to update the style of
+      theme_manager.apply_style = lambda: None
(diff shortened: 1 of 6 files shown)


khaneliman:
Can rebase now that flake update was merged.

JuneStepp:
> Can rebase now that flake update was merged.

Done. 

@khaneliman Anki is marked as broken on Darwin in nixpkgs.  How should I deal with that here? I tried adding it to the scrub list, but the package used is what's returned by `anki.withAddons` rather than Anki itself.

JuneStepp:
@khaneliman Anything I can do to help stop this from getting lost in the PR sea?

khaneliman:
Sorry, I keep meaning to come look at it again... just seems like a lot of custom python code for a home manager module..

JuneStepp:
> Sorry, I keep meaning to come look at it again... just seems like a lot of custom python code for a home manager module..

While there's a fair amount of it, it's very simple and half just passing Nix values to Python functions. I thought it would be better to go for a verbose unlikely to need much maintenance solution over the alternative of translating options to the DB directly.

modules/programs/anki.nix
```diff
@@ -0,0 +1,567 @@
...
+    if pkgs.stdenv.hostPlatform.isDarwin then
+      "Library/Application Support/Anki2"
+    else
+      "${config.xdg.dataHome}/Anki2";
+  cfg = config.programs.anki;
+  # This script generates the Anki SQLite settings DB using the Anki Python API.
+  # The configuration options in the SQLite database take the form of Python
+  # Pickle data.
+  # A simple "gldriver6" file is also generated for the `videoDriver` option.
+  buildAnkiConfig = pkgs.writers.writeText "buildAnkiConfig" ''
```
khaneliman: Might feel less bloated if we move some this module to a directory structure and put these into a separate helper/lib file.

- anki/
  - default.nix
  - lib/helper.nix

modules/programs/anki.nix
```diff
@@ -0,0 +1,567 @@
...
+          the `usernameFile` option.
+        '';
+      }
+      {
+        assertion = cfg.package ? withAddons;
+        message = ''
+          The value of `programs.anki.package` doesn't support declaratively managing
+          add-ons. Make sure you are using `pkgs.anki`.
+        '';
+      }
```
khaneliman: Why not support a nullable package and change the logic in `home.packages` to either

- Install no package
- Install package without addons
- Install package with addons

modules/programs/anki.nix
```diff
@@ -0,0 +1,567 @@
...
+      ankiPackage = if cfgAnkiPython == null then pkgs.anki else cfg.package;
+      ankiPython = if cfgAnkiPython == null then pkgs.python3 else cfgAnkiPython;
+    in
+    pkgs.runCommand "ankiConfig"
+      {
+        nativeBuildInputs = [ ankiPackage ];
+      }
+      ''
+        ${ankiPython.interpreter} ${buildAnkiConfig} $out
+      '';
```
khaneliman: Or doe the module always require a package since we need the anki package to generate the db file?

khaneliman (commented):
Just some questions about layout / design decisions.

modules/programs/anki.nix
JuneStepp: The builtin addons are needed to support declarative configs, and the package is needed to install the addons.

modules/programs/anki.nix
JuneStepp: See the other answer, but also yes to this.

modules/programs/anki.nix
JuneStepp: Done.

modules/programs/anki/helper.nix
```diff
@@ -0,0 +1,261 @@
...
+    '';
+  };
+
+  # Make Anki work better with declarative settings. See script for specific changes.
+  homeManagerAnkiAddon = pkgs.anki-utils.buildAnkiAddon {
+    pname = "home-manager";
+    version = "1.0";
+    src = pkgs.writeTextDir "__init__.py" ''
+      import aqt
+      from aqt.qt import QWidget, QMessageBox
```
khaneliman, line 193: Assuming the QT dependencies are handled by the package already?

JuneStepp: Ya. aqt is Anki's Python interface.

JuneStepp:
@khaneliman Thanks for reviewing and merging this!

PR #7275, opened 2025-06-15
kakoune: implement a final package option

MaeIsBad:
### Description

This adds a "final-package" option to kakoune options, similar to `programs.neovim.finalPackage`.

This is useful for writing scrips, creating aliases etc.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

The module doesn't have a maintainers, @philipwilk you were the last person to have edited it, would you mind leaving a review?

### DIFF
--- a/modules/programs/kakoune.nix
+++ b/modules/programs/kakoune.nix
@@ -688,6 +688,12 @@ in
 
       package = lib.mkPackageOption pkgs "kakoune-unwrapped" { nullable = true; };
 
+      finalPackage = mkOption {
+        type = types.nullOr types.package;
+        readOnly = true;
+        description = "Resulting customized kakoune package.";
+      };
+
       config = mkOption {
         type = types.nullOr configModule;
         default = { };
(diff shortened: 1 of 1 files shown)


jadewilk:
I don't personally have a use case where I would need this option, but I would just scrap the null checks against cfg.package because they're redundant; you've already enabled the option?

I'd do it like https://github.com/nix-community/home-manager/compare/master...philipwilk:home-manager:kak-finalPackage which I think works, but you're welcome to try it in your use case.
image

khaneliman:
> I don't personally have a use case where I would need this option, but I would just scrap the null checks against cfg.package because they're redundant; you've already enabled the option?
> 
> I'd do it like [master...philipwilk:home-manager:kak-finalPackage](https://github.com/nix-community/home-manager/compare/master...philipwilk:home-manager:kak-finalPackage) which I think works, but you're welcome to try it in your use case. image

The null checks prevent the expression from failing to evaluate when someone wanting to disable installing the package from HM.

Issue #6062, opened 2024-11-09
hyprland: submaps cannot be configured with nix syntax


0x57e11a:
`hyprland` does an odd thing in its usage of `hyprlang` with `submap = ...`, which switches into different submaps when binding keys (ideally this would've been implemented as different `submap:xyz { ... }` sections, but they Decided To Do It That Way for some reason)

```
# other keybinds

submap = something

bindm = , mouse:272, movewindow
bind = , escape, submap, reset

submap = reset

# other keybinds
```

because this specific section of keybinds needs to be surrounded by `submap` boundaries, it's not possible to do this with `wayland.windowManagers.hyprland.settings`

<details>
<summary>solution</summary>

the proposal is to add `wayland.windowManagers.hyprland.submaps`, like so:

```
wayland.windowManager.hyprland.submaps = {
    something = {
        bindm = [ ", mouse: 272, movewindow" ];
        bind = [ ", escape, submap, reset" ];
    };
};
```
</details>

<details>
<summary>workaround</summary>

`wayland.windowManagers.hyprland.extraConfig` exists, but having part of the config in a string and the rest in nix options is frustrating
</details>

Acisama:
Actually you can definie the name of the standard keymap by putting "submap" = "reset"; abouve binds =

0x57e11a:
1. that's... not how nix works
2. not trying to rename the standard submap, trying to specifically define binds in a different submap, which is *not possible to do with the current nix syntax*

de-abreu:
I've suggested the change be done [at the Hyprland side](https://github.com/hyprwm/Hyprland/issues/9296). No luck.

0x57e11a:
a potential workaround that wouldn't be too difficult to turn into a PR (by someone with the time):
```rs
extraConfig = let
	submaps = {
		chord = {
			settings = {
				bind = [
					", escape, submap, reset"

					", q, exit,"

					",,exec,"
				];

				bindm = [
					", mouse:272, movewindow"
				];
			};

			extraConfig = "";
		};
	};
in lib.concatStringsSep "\n" (
	lib.mapAttrsToList
		(name: submap: (
			"submap = ${name}\n"
			+ lib.optionalString (submap.settings != { }) (
				inputs.home-manager.lib.hm.generators.toHyprconf {
					attrs = submap.settings;
				}
			)
			+ lib.optionalString (submap.extraConfig != "") submap.extraConfig
			+ "submap = reset"
		))
		submaps
);
```

[...]

PR #7277, opened 2025-06-15
hyprland: add support for submaps (#6062)

C0Florent:
### Description

This PR adds support for [hyprland submaps](https://wiki.hypr.land/Configuring/Binds/#submaps) in the Hyprland HM module.

This is my first PR here, and this implementation is aimed to be basic, submap support could be extended in the future. The only "flashy" things in this PR are an assertion for submap name `reset` and warnings for non-binds in submaps.

Still slightly WIP : the added features seem to function all as expected (as far as my testing went), but there are still some things to do (I'm open for advice on how to get these done):
- [ ] Writing tests for the submap support
- [ ] Documentation
  - [x] Double-checking/improving `description`s and `example`s
  - [ ] Checking the rendered docs look right

I'm also all open for discussion on the implementation details and option structure, formatting or any comment really

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@fufexan

### DIFF
--- a/modules/services/window-managers/hyprland.nix
+++ b/modules/services/window-managers/hyprland.nix
@@ -216,6 +216,36 @@ in
       '';
     };
 
+    submaps = lib.mkOption {
+      description = "Attribute set of Hyprland submaps";
+      default = { };
+      type = lib.types.attrsOf (
+        lib.types.submodule (
+          { name, config, ... }:
+          {
+            options = {
+              settings = lib.mkOption {
+                type = (with lib.types; attrsOf (listOf str)) // {
+                  description = "Hyprland binds";
+                };
+                default = { };
+                description = ''
+                  Hyprland binds to be put in the submap
+                '';
+                example = lib.literalExpression ''
+                  {
+                    bind = [
+                      ", q, exec $terminal"
+                    ];
+                  }
+                '';
+              };
+            };
+          }
+        )
+      );
+    };
+
     extraConfig = lib.mkOption {
       type = lib.types.lines;
       default = "";
@@ -264,6 +294,10 @@ in
   config = lib.mkIf cfg.enable {
     assertions = [
       (lib.hm.assertions.assertPlatform "wayland.windowManager.hyprland" pkgs lib.platforms.linux)
+      {
+        assertion = !builtins.hasAttr "reset" cfg.submaps;
+        message = "Submaps can't be named 'reset'. The name 'reset' is reserved in order to have a way to switch to the default submap; as if 'reset' was its name.";
+      }
     ];
 
     warnings =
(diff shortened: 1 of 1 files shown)


modules/services/window-managers/hyprland.nix
```diff
@@ -264,15 +294,32 @@ in
...
+          attrs:
+          builtins.filter (n: builtins.match ''bind[[:lower:]]*'' n == null) (builtins.attrNames attrs);
+
+        # attrset of { <submap name> = <list of non bind* keys>; } for all submaps
+        submapWarningsAttrset = builtins.mapAttrs (
+          name: submap: filterNonBinds submap.settings
+        ) cfg.submaps;
+
+        submapWarnings = lib.mapAttrsToList (submapName: nonBinds: ''
+          wayland.windowManager.hyprland.submaps."${submapName}".settings: found non-bind entries: [${builtins.toString nonBinds}], which probably will have no effect in a submap
```
fufexan:
```suggestion
          wayland.windowManager.hyprland.submaps."${submapName}".settings: found non-bind entries: [${builtins.toString nonBinds}], which will have no effect in a submap
```

fufexan (commented):
Other than this, looks good

modules/services/window-managers/hyprland.nix
C0Florent: Fixed

C0Florent:
I also improved the submap examples to give users a better idea of how the option works

I'm thinking of removing the submap example in `extraConfig`, since we now have a proper submap option. No idea what to put in the `extraConfig` example then.

I also added a link to the submap wiki page in the option description. In doing that I noticed the domain name changed from `hyprland.org` to `hypr.land`, I guess we'll change all occurences of the former to the latter in a separate PR (now open at #7293).

khaneliman:
> I'm thinking of removing the submap example in `extraConfig`, since we now have a proper submap option. No idea what to put in the `extraConfig` example then.

I'd love to get rid of `extraConfig` if we think the module can handle everything properly. I know @fufexan said he was working on a new generator a while back, not sure what the state of that is. But, we can just remove the example, for now, and eventually deprecate/remove the `extraConfig` if the options and generator work for an entire configuration.

fufexan:
It's close to finished, though only works properly for Hyprland. Also doesn't handle submaps (yet). We also need to add array indexing support in hyprlang, so that hyprlock/hypridle can also work properly.

Here's the implementation https://github.com/hyprwm/Hyprland/blob/main/nix/lib.nix and the usage in the NixOS module https://github.com/hyprwm/Hyprland/blob/main/nix/module.nix.

Discussion about hyprlang starts around here https://discord.com/channels/961691461554950145/1233827370855694426/1367831435679105026.

PR #7294, opened 2025-06-19
obsidian: add extraSettings

mctrxw:
### Description

added extraSettings to obsidian module

### Example

```
programs.obsidian.extraSettings = {
  frame = "native";
};
```

### Checklist

- [x] Code formatted with nix fmt or
    nix-shell -p treefmt nixfmt-rfc-style keep-sorted --run treefmt.

- [x] Code tested through nix-shell --pure tests -A run.all
    or nix build --reference-lock-file flake.lock ./tests#test-all using Flakes.

    {component}: {description}

    {long description}
    

#### Maintainer CC

@rycee @karaolidis

### DIFF
--- a/modules/programs/obsidian.nix
+++ b/modules/programs/obsidian.nix
@@ -385,6 +385,12 @@ in
       );
       default = { };
     };
+
+    extraSettings = mkOption {
+      description = "Additional settings to include in obsidian.json.";
+      type = types.attrsOf types.anything;
+      default = { };
+    };
   };
 
   config =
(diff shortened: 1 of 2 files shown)


karaolidis (approved):

modules/programs/obsidian.nix
```diff
@@ -385,6 +385,12 @@ in
       );
       default = { };
     };
+
+    extraSettings = mkOption {
+      description = "Additional settings to include in obsidian.json.";
```
0xda157, line 390: "include" would suggest this doesn't override the other options

unazikx: yes, cause it merged via //, or what are u talking about?

0xda157: `//` does override, for example
```
nix-repl> { vaults = { foo = "x"; }; } // { vaults = { foo = "y"; }; }
{
  vaults = {
    foo = "y";
  };
}
```
I'm suggesting either changing the description (probably the better option) or using mkMerge in the application. Usually extraSettings are just dumped at the end of the file, but that obviously doesn't work for json.

unazikx: it overrides if you merging same name attribute, but in obsidian you can't set anything via built-in settings, only nix settings because obsidian.json read only, understood?

musjj: I think what they're saying is that `//` is a shallow merge, which means that it'd be impossible to append a deeply-nested key. For example, let's say that you have:

```nix
{ a.b.c = 1; }
```

and you want to *add* a new attribute: `{ a.b.x = 2; }`. With `//`, the previous value in `a.b.c` will not be preserved:

```nix
nix-repl> :p { a.b.c = 1; } // { a.b.x = 2; }
{
  a = {
    b = { x = 2; };
  };
}
```

This is surprising compared to the `extra{Settings,Config}` options in other modules. By using `mkMerge`, `{ a.b.c = 1; }` and `{ a.b.x = 2; }` will be merged so that:

```nix
{
  a = {
    b = { c = 1; x = 2; };
  };
}
```

Issue #6725, opened 2025-03-29
bug: aerospace doesn't start at login


Eveeifyeve:
### Are you following the right branch?

- [x] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Issue description

The issue is when using aerospace with home manager there is no service to start aerospace. The problem is if you use nix darwin aerospace service it doesn't inherit your config which renders aerospace useless if you can't configure it. So I went to home manager and yet home manager doesn't have a service been added or the ability to add one. So manually I have to specify in the macos to start the application. 

### Maintainer CC

@damidoug 

### System information

```markdown
- system: `"aarch64-darwin"`
 - host os: `Darwin 24.2.0, macOS 15.2`
 - multi-user?: `yes`
 - sandbox: `no`
 - version: `nix-env (Nix) 2.27.0`
 - channels(root): `"nixpkgs"`
 - nixpkgs: `/nix/store/0yw40qnrar3lvc5hax5n49abl57apjbn-source`
```

damidoug:
```
programs.aerospace.userSettings.start-at-login = true;
```

Have you tested this? If so, try logging out and back in to see if the changes take effect.

baptman21:
I had the same issue. From what I understand the home-manager config does not add any entry to the LaunchAgents. However if you enable the `start-at-login` command, then run the Aerospace App (from the LaunchPad), aerospace itself adds the LaunchAgent file itself.

Technically I would say that adding an option to enable the service from home manager would be better, but I don't know if we have the feature to control the user's LaunchAgents from home manager.

If not maybe a precision in the doc could help.

damidoug:
> I had the same issue. From what I understand the home-manager config does not add any entry to the LaunchAgents. However if you enable the `start-at-login` command, then run the Aerospace App (from the LaunchPad), aerospace itself adds the LaunchAgent file itself.
> 
> Technically I would say that adding an option to enable the service from home manager would be better, but I don't know if we have the feature to control the user's LaunchAgents from home manager.
> 
> If not maybe a precision in the doc could help.

Startup at login is now managed via Home Manager's launchd integration. Please refer to the new PR: [#7299](https://github.com/nix-community/home-manager/pull/7299) for details.

eveeifyeve:
If there is still more issues please reopen an issue. but please try to use home manager launchd integration first instead of builtin start-at-login

PR #7299, opened 2025-06-21
aerospace: Add launchd agent and enhance configuration

damidoug:
### Description
This commit introduces a `launchd` agent to automatically start AeroSpace at login, significantly improving its integration.

Solve issue #6725 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/aerospace.nix
+++ b/modules/programs/aerospace.nix
@@ -44,31 +44,63 @@ in
 
     package = lib.mkPackageOption pkgs "aerospace" { nullable = true; };
 
+    launchd = {
+      enable = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          Configure the launchd agent to manage the AeroSpace process.
+
+          The first time this is enabled, macOS will prompt you to allow this background
+          item in System Settings.
+
+          You can verify the service is running correctly from your terminal.
+          Run: `launchctl list | grep aerospace`
+
+          - A running process will show a Process ID (PID) and a status of 0, for example:
+            `12345	0	org.nix-community.home.aerospace`
+
+          - If the service has crashed or failed to start, the PID will be a dash and the
+            status will be a non-zero number, for example:
+            `-	1	org.nix-community.home.aerospace`
+
+          In case of failure, check the logs with `cat /tmp/aerospace.err.log`.
+
+          For more detailed service status, run `launchctl print gui/$(id -u)/org.nix-community.home.aerospace`.
+
+          NOTE: Enabling this option will configure AeroSpace to **not** manage its own
+          launchd agent. Specifically, it will set `start-at-login = false` and
+          `after-login-command = []` in the configuration file, as those are now handled
+          by Home Manager and launchd instead.
+        '';
+      };
+      keepAlive = mkOption {
+        type = types.bool;
+        default = true;
+        description = "Whether the launchd service should be kept alive.";
+      };
+    };
+
     userSettings = mkOption {
       type = types.submodule {
         freeformType = tomlFormat.type;
         options = {
-          start-at-login = lib.mkOption {
-            type = types.bool;
-            default = false;
-            description = "Start AeroSpace at login.";
-          };
-          after-login-command = mkOption {
-            type = with types; listOf str;
-            default = [ ];
-            description = ''
-              You can use it to add commands that run after login to macOS user session.
-              'start-at-login' needs to be 'true' for 'after-login-command' to work.
-            '';
-          };
           after-startup-command = mkOption {
             type = with types; listOf str;
             default = [ ];
             description = ''
-              You can use it to add commands that run after AeroSpace startup.
-              'after-startup-command' is run after 'after-login-command'
+              A list of AeroSpace commands to execute immediately after the AeroSpace application starts.
+              These commands are written to your `aerospace.toml` config file and are run after the `after-login-command` sequence.
+
+              A list of all available commands can be found at <https://nikitabobko.github.io/AeroSpace/commands>.
+
+              While this module checks for valid command names, using incorrect *arguments* can still cause issues.
+              If AeroSpace is not behaving correctly after startup, check the logs for errors with `cat /tmp/aerospace.err.log`.
             '';
-            example = [ "layout tiles" ];
+            example = [
+              "exec-and-forget open -n /System/Applications/Utilities/Terminal.app"
+              "layout tiles accordion"
+            ];
           };
           enable-normalization-flatten-containers = mkOption {
             type = types.bool;
@@ -269,9 +301,28 @@ in
 
     home = {
       packages = lib.mkIf (cfg.package != null) [ cfg.package ];
+
       file.".config/aerospace/aerospace.toml".source = tomlFormat.generate "aerospace" (
-        filterNulls cfg.userSettings
+        filterNulls (
+          cfg.userSettings
+          // lib.optionalAttrs cfg.launchd.enable {
+            # Override these to avoid launchd conflicts
+            start-at-login = false;
+            after-login-command = [ ];
+          }
+        )
       );
+    };
+
+    launchd.agents.aerospace = {
+      enable = cfg.launchd.enable;
+      config = {
+        Program = "${cfg.package}/Applications/AeroSpace.app/Contents/MacOS/AeroSpace";
+        KeepAlive = cfg.launchd.keepAlive;
+        RunAtLoad = true;
+        StandardOutPath = "/tmp/aerospace.log";
+        StandardErrorPath = "/tmp/aerospace.err.log";
+      };
     };
   };
 }
--- a/tests/modules/programs/aerospace/aerospace-service-expected.plist
+++ b/tests/modules/programs/aerospace/aerospace-service-expected.plist
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>KeepAlive</key>
+	<true/>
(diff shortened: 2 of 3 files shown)


modules/programs/aerospace.nix
```diff
@@ -48,27 +48,87 @@ in
...
+
+              Please use `programs.aerospace.userSettings.after-startup-command` directly for all startup commands.
+
+              This option is kept for backward compatibility. It no longer has any effect on the launchd service itself.
+              All commands provided here will be executed by AeroSpace via its configuration file.
+
+              A list of all available commands can be found at <https://nikitabobko.github.io/AeroSpace/commands>.
+
+              While this module checks for valid command names, using incorrect *arguments* can still cause issues.
+              If AeroSpace is not behaving correctly after startup, check the logs for errors with `cat /tmp/aerospace.err.log`.
```
khaneliman: Not sure I understand this deprecation. It looks like an option still available upstream and not deprecated.

damidoug: I just keep this option for backwards compatible, but behind the commands added are going to be available on after-startup-command.

khaneliman: Sorry, asking more of a reason why we are combining them into a `after-startup-command` in the first place instead of using the setting in the correctly named option so that we don't surprise users who want to add it to the setting they are declaring. This isn't a home-manager option really, it is a passthru of the upstream options to the generator. We are kinda defeating the purpose of the freeform config by doing this.

damidoug: I tried to add these commands to `launchd.agents.aerospace.config.ProgramArguments` but didn't work.

Any ideia how fix it? fell free to change if needed.

khaneliman: Well, my suggestion would be to let the config pass through, like it used to. 

If you wanted to add a `hm` option outside of `settings` that overrides a user's `settings` config then that can be merged into the `settings` attr that way. I'm not sure how these options relate to the `launchd` agent program arguments though. From what I can see, they are 2 separate things.

1. Use configures window manager in `settings`
2. We create `launchd` unit for auto launching the program

If we want to not have a surprising or confusing change, I'd say we should create new options outside of `settings` for merging in configurations required for the Home Manager managed `launchd` unit. IE: `programs.aerospace.launchd.enable` or `programs.aerospace.launchd.keepAlive` to control the actual `launchd` configuration. If enabling the `launchd` agent requires a `aerospace.toml` change then merge them in with those options being defined/enabled instead of blurring the lines between `settings` generator and Home Manager configurations.

modules/programs/aerospace.nix
```diff
@@ -48,27 +48,87 @@ in
...
+                      type = bool;
+                      default = true;
+                      description = ''
+                        If set to true, launchd will automatically restart AeroSpace
+                        if it crashes or is quit for any reason.
+                      '';
+                    };
+                  };
+                })
+              ];
```
khaneliman: `settings` options should be generator pass through options. They should not be used for other configurations outside of the configuration file generated.

modules/programs/aerospace.nix
```diff
@@ -48,27 +48,87 @@ in
...
             default = [ ];
             description = ''
-              You can use it to add commands that run after login to macOS user session.
-              'start-at-login' needs to be 'true' for 'after-login-command' to work.
+              DEPRECATED: Commands from this option are now automatically merged into `after-startup-command`.
+
+              Please use `programs.aerospace.userSettings.after-startup-command` directly for all startup commands.
+
+              This option is kept for backward compatibility. It no longer has any effect on the launchd service itself.
+              All commands provided here will be executed by AeroSpace via its configuration file.
```
khaneliman:
```suggestion
```

The option wasn't deprecated upstream, we shouldn't say it is a deprecated option.

modules/programs/aerospace.nix
```diff
@@ -269,9 +329,44 @@ in
...
+          );
+        in
+        tomlFormat.generate "aerospace" (
+          filterNulls (
+            cfg.userSettings
+            // {
+              after-startup-command = combinedStartupCommands;
+              start-at-login = false;
+              after-login-command = [ ];
+            }
```
khaneliman: This should be guarded behind a Home Manager option outside of `settings` so that it's an opt in behavior to overwrite `settings` configurations.

modules/programs/aerospace.nix
```diff
@@ -269,9 +329,44 @@ in
...
+        );
     };
+
+    launchd.agents.aerospace =
+      let
+        salCfg = cfg.userSettings.start-at-login;
+        isSimpleBool = lib.isBool salCfg;
+
+        enableService = if isSimpleBool then salCfg else salCfg.enable;
+        keepAliveService = if isSimpleBool then true else salCfg.keepAlive;
```
khaneliman: We should not determine the launchd configuration from a users `settings` it should be Home Manager options for opting in/configuring.

modules/programs/aerospace.nix
damidoug: I got your suggestion

modules/programs/aerospace.nix
damidoug: I left that option but I add assertion like we have in the [nix-darwin package](https://github.com/nix-darwin/nix-darwin/blob/master/modules/services/aerospace/default.nix) so we avoid to enable it 'Cause it's causing issues.

modules/programs/aerospace.nix
```diff
@@ -44,31 +44,68 @@ in
...
           };
-          after-login-command = mkOption {
+          after-login-command = lib.mkOption {
             type = with types; listOf str;
             default = [ ];
-            description = ''
-              You can use it to add commands that run after login to macOS user session.
-              'start-at-login' needs to be 'true' for 'after-login-command' to work.
-            '';
+            description = "Use programs.aerospace.userSettings.after-startup-command instead. This option is ignored when using launchd.";
```
khaneliman: If we don't want to document these options as something to customize, i'd probably just remove the options from here. This is a freeform submodule so they don't need to be declared for a user to use them, you also already have an assertion in case someone tries.

khaneliman:
Don't worry about the conflict, for now. I am going to push a PR that should resolve it. https://github.com/nix-community/home-manager/pull/7312

modules/programs/aerospace.nix
damidoug: Okay, since you're going to push a PR fell free to change it.

Thank you.

khaneliman: Sorry, meant that I have another PR I'm going to merge that should remove the `conflict` message when we get ready to try merging.

khaneliman: I can remove the options, if you want in here, though.

damidoug: I'm gonna remove no worries.

modules/programs/aerospace.nix
```diff
@@ -269,9 +296,28 @@ in
...
       file.".config/aerospace/aerospace.toml".source = tomlFormat.generate "aerospace" (
-        filterNulls cfg.userSettings
+        filterNulls (
+          cfg.userSettings
+          // {
+            # Override these to avoid launchd conflicts
+            start-at-login = false;
+            after-login-command = [ ];
+          }
+        )
```
khaneliman:
```suggestion
          // lib.optionalAttrs cfg.launchd.enable {
            # Override these to avoid launchd conflicts
            start-at-login = false;
            after-login-command = [ ];
          }
        )
```

I think we should add a note about this behavior to the `launchd` option as well. Just to avoid surprising users, if you don't have a warning/assertion in place.

damidoug: how can I add this note?

khaneliman: Just to the `launchd.enable` description somewhere mentioning this particular side effect. 

Something similar to "NOTE: This will disable `aerospace` managing its own launchd agent, by configuring `start-at-login` and `after-login-command`."

damidoug: done

khaneliman:
Appreciate your patience, would you be interested in adding an assertion to one of the tests that the launchd unit gets created when you enable that option?

damidoug:
> Appreciate your patience, would you be interested in adding an assertion to one of the tests that the launchd unit gets created when you enable that option?

Yes, can you do it? so I can learn for the next modules.

Really thanks for the feedbacks.

modules/programs/aerospace.nix
```diff
@@ -269,9 +301,28 @@ in
...
     };
+
+    launchd.agents.aerospace = {
+      enable = cfg.launchd.enable;
+      config = {
+        Program = "${cfg.package}/Applications/AeroSpace.app/Contents/MacOS/AeroSpace";
+        KeepAlive = cfg.launchd.keepAlive;
+        RunAtLoad = true;
+        StandardOutPath = "/tmp/aerospace.log";
+        StandardErrorPath = "/tmp/aerospace.err.log";
```
khaneliman, line 324: Something I noticed while copying my sketchybar implementation of a test. 

```nix
      errorLogFile = mkOption {
        type = with lib.types; nullOr (either path str);
        default = "${config.home.homeDirectory}/Library/Logs/sketchybar/sketchybar.err.log";
        defaultText = lib.literalExpression "\${config.home.homeDirectory}/Library/Logs/sketchybar/sketchybar.err.log";
        example = "/Users/khaneliman/Library/Logs/sketchybar.log";
        description = "Absolute path to log all stderr output.";
      };

      outLogFile = mkOption {
        type = with lib.types; nullOr (either path str);
        default = "${config.home.homeDirectory}/Library/Logs/sketchybar/sketchybar.out.log";
        defaultText = lib.literalExpression "\${config.home.homeDirectory}/Library/Logs/sketchybar/sketchybar.out.log";
        example = "/Users/khaneliman/Library/Logs/sketchybar.log";
        description = "Absolute path to log all stdout output.";
      };
```

Is how I defined the ability for a user to control this and default to the user directory. Not required, but just a thought.

PR #7300, opened 2025-06-21
bash: change sessionVariables to attrsOf ...

teto:
I had programs.bash.sessionVariables.CDPATH = ... defined in the different nix files. One was taken into account and the other one ignored.

Turns out plain "types.attrs" has the following merging strategy:

===
merge = loc: foldl' (res: def: res // def.value) { }; ===

Changing it to "attrsOf", the module system warned me of the conflicting values.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/bash.nix
+++ b/modules/programs/bash.nix
@@ -137,7 +137,7 @@ in
 
       sessionVariables = mkOption {
         default = { };
-        type = types.attrs;
+        type = with types; attrsOf (either str int);
         example = {
           MAILCHECK = 30;
         };


teto:
Limiting the types to `either str int` should not be breaking i think ? because there were already the 2 being stringified ?

It's breaking in the sense that this might generate eval errors that did not exist before in case of conflicting values. You can bypass a proper fix by resorting to `lib.mkForce`.
If other maintainers are fine with it, i will do the same for zsh.

khaneliman (approved):
LGTM

teto:
Let's see if there is some backlash for bash first before doing the same for zsh xD

teatwig:
@teto it did break my build slightly because I was doing things like `VAR = true` and `VAR = somePackage` (so it automatically used the store path), but I probably shouldn't have done that in the first place xD

NixOS additionally allows paths (and lists of the the three types) in `environment.variables` and `environment.sessionVariables`, so I'd be in favor of at least supporting paths too: https://github.com/NixOS/nixpkgs/blob/nixos-25.05/nixos/modules/config/shells-environment.nix#L56-L69

khaneliman:
> @teto it did break my build slightly because I was doing things like `VAR = true` and `VAR = somePackage` (so it automatically used the store path), but I probably shouldn't have done that in the first place xD
> 
> NixOS additionally allows paths (and lists of the the three types) in `environment.variables` and `environment.sessionVariables`, so I'd be in favor of at least supporting paths too: https://github.com/NixOS/nixpkgs/blob/nixos-25.05/nixos/modules/config/shells-environment.nix#L56-L69

I did consider paths as a possible usage, should have recommended it. At least its a quick fix to add

teto:
@teatwig thanks for the insightful comment. I confess I did not think to check the nixos module (how stupid). path is ok to add. I dont agree with the list support of nixpkgs which sounds a bit magical. `LUA_PATH` separator for instance is `;` and so the module behavior would be incorrect. With khaneli's fix I think we are covered and we should do the same for zsh.

PR #7303, opened 2025-06-22
programs/hyprpanel: init

PerchunPak:
### Description

https://github.com/Jas-SinghFSU/HyprPanel: Bar/Panel for Hyprland with extensive customizability (https://github.com/NixOS/nixpkgs/pull/379994)

I have been daily driving everything from this PR since the 2025 new year, so everything is pretty well tested

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

CC @JohnRTitor

### DIFF
--- a/modules/programs/hyprpanel/default.nix
+++ b/modules/programs/hyprpanel/default.nix
@@ -1,0 +1,121 @@
+{
+  pkgs,
+  lib,
+  config,
+  ...
+}:
+let
+  cfg = config.programs.hyprpanel;
+  jsonFormat = pkgs.formats.json { };
+in
+{
+  meta.maintainers = [ lib.maintainers.perchun ];
+
+  options.programs.hyprpanel = {
+    enable = lib.mkEnableOption "HyprPanel";
+
+    package = lib.mkPackageOption pkgs "hyprpanel" { };
+
+    settings = lib.mkOption {
+      type = jsonFormat.type;
+      default = { };
+      example = lib.literalExpression ''
+        bar.battery.label = true;
+        bar.bluetooth.label = false;
+        bar.clock.format = "%H:%M:%S";
+        bar.layouts = {
+          "*" = {
+            left = [
+              "dashboard"
+              "workspaces"
+              "media"
+            ];
+            middle = [ "windowtitle" ];
+            right = [
+              "volume"
+              "network"
+              "bluetooth"
+              "notifications"
+            ];
+          };
+        };
+      '';
+      description = ''
+        Configuration written to
+        {file}`$XDG_CONFIG_HOME/hyprpanel/config.json`.
+
+        See <https://hyprpanel.com/configuration/settings.html#home-manager-module>
+        for the full list of options.
+      '';
+    };
+
+    systemd.enable = lib.mkEnableOption "HyprPanel systemd integration" // {
+      default = true;
+    };
+
+    dontAssertNotificationDaemons = lib.mkOption {
+      default = true;
+      example = false;
+      description = ''
+        Whether to check for other notification daemons.
+
+        You might want to set this to false, because hyprpanel's notification
+        daemon is buggy and you may prefer something else.
+      '';
+      type = lib.types.bool;
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    assertions = [
+      {
+        assertion = cfg.dontAssertNotificationDaemons && !config.services.swaync.enable;
+        message = ''
+          Only one notification daemon can be enabled at once. You have enabled
+          swaync and hyprpanel at once.
+
+          If you dont want to use hyprpanel's notification daemon, set
+          `programs.hyprpanel.dontAssertNotificationDaemons` to true.
+        '';
+      }
+    ];
+
+    home.packages = [ cfg.package ];
+
+    programs.hyprpanel.settings = lib.mkIf config.services.hypridle.enable {
+      # fix hypridle module if user uses systemd service
+      bar.customModules.hypridle.startCommand = lib.mkDefault "systemctl --user start hypridle.service";
+      bar.customModules.hypridle.stopCommand = lib.mkDefault "systemctl --user stop hypridle.service";
+      bar.customModules.hypridle.isActiveCommand = lib.mkDefault "systemctl --user status hypridle.service | grep -q 'Active: active (running)' && echo 'yes' || echo 'no'";
+    };
+
+    xdg.configFile.hyprpanel = lib.mkIf (cfg.settings != { }) {
+      target = "hyprpanel/config.json";
+      source = jsonFormat.generate "hyprpanel-config" cfg.settings;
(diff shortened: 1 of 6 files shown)


modules/programs/hyprpanel.nix
```diff
@@ -0,0 +1,121 @@
...
+      description = ''
+        Configuration written to
+        {file}`$XDG_CONFIG_HOME/hyprpanel/config.json`.
+
+        See <https://hyprpanel.com/configuration/settings.html#home-manager-module>
+        for the full list of options.
+      '';
+    };
+
+    systemd.enable = (lib.mkEnableOption "HyprPanel systemd integration") // {
```
JumpIn-Git: The brackets here are not needed, so `lib.mkEnableOption "HyprPanel systemd integration" //`

tests/default.nix
```diff
@@ -150,78 +150,381 @@ let
...
+      ./modules/services/wlsunset
+      ./modules/services/wob
+      ./modules/services/wpaperd
+      ./modules/services/xsettingsd
+      ./modules/services/yubikey-agent
+      ./modules/systemd
+      ./modules/targets-linux
+      # keep-sorted end
+    ]
+  );
```
khaneliman: this should be removed

tests/modules/programs/hyprpanel/with-hypridle.nix
```diff
@@ -0,0 +1,35 @@
+{ config, ... }:
+{
+  services.hypridle.enable = true;
+  programs.hyprpanel = {
+    enable = true;
+    package = config.lib.test.mkStubPackage { name = "tldr"; };
```
khaneliman:
```suggestion
```

Shouldn't need this either on linux. We automatically stub packages for linux.

tests/modules/programs/hyprpanel/basic-config.nix
```diff
@@ -0,0 +1,34 @@
+{ config, ... }:
+{
+  programs.hyprpanel = {
+    enable = true;
+    package = config.lib.test.mkStubPackage { name = "tldr"; };
```
khaneliman:
```suggestion
```

Shouldn't need this either on linux. We automatically stub packages for linux.

tests/modules/programs/hyprpanel/with-hypridle.nix
PerchunPak: No, it crashes, but i will try again

khaneliman: Oh, the package hasn't hit the channel, yet. Then yeah the stub works, for now, since the automatic stubbing depends on the package existing. Just update the name

PerchunPak: Done

modules/programs/hyprpanel/default.nix
```diff
@@ -0,0 +1,121 @@
...
+      # hyprpanel replaces it with the same file, but without new line in the end
+      force = true;
+    };
+
+    systemd.user.services.hyprpanel = lib.mkIf cfg.systemd.enable {
+      Unit = {
+        Description = "Bar/Panel for Hyprland with extensive customizability";
+        Documentation = "https://hyprpanel.com/getting_started/hyprpanel.html";
+        PartOf = [ "graphical-session.target" ];
+        After = [ "graphical-session.target" ];
```
khaneliman:
```suggestion
        PartOf = [ config.wayland.systemd.target ];
        After = [ config.wayland.systemd.target ];
```
Sorry, missed this originally. Same for Install.WantedBy

khaneliman:
LGTM, thanks. Will merge after CI

PerchunPak:
> after CI

:pray:

khaneliman:
Ah, the json generator will go alphabetically for attrsets. 

       > @@ -11,9 +11,9 @@
       >      },
       >      "customModules": {
       >        "hypridle": {
       > -        "isActiveCommand": "systemctl --user status hypridle.service | grep -q 'Active: active (running)' && echo 'yes' || echo 'no'",
       >          "startCommand": "systemctl --user start hypridle.service",
       > -        "stopCommand": "systemctl --user stop hypridle.service"
       > +        "stopCommand": "systemctl --user stop hypridle.service",
       > +        "isActiveCommand": "systemctl --user status hypridle.service | grep -q 'Active: active (running)' && echo 'yes' || echo 'no'"
       >        }
       >      },

PerchunPak:
Oh, I found `nix build --reference-lock-file ./tests#test-hyprpanel-with-hypridle`

It should finally pass :pray:

ppenguin:
@PerchunPak Thanks for doing this. I've been through many issues with the nix HP config, because the attribute config/override behaviour was inconsistent.

Before spending time again, I'd like to ask whether this is completely solved now?
And can I now use a *real* JSON hierarchy instead of the non-idiomatic one at https://hyprpanel.com/configuration/settings.html

I.e.:

```nix
bar = {
  battery = {
    hideLabelWhenFull = mkBoolOption false;
    label = mkBoolOption true;
    middleClick = mkStrOption "";
    rightClick = mkStrOption "";
  };
};
```

instead of 

```nix
bar.battery.hideLabelWhenFull = mkBoolOption false;
bar.battery.label = mkBoolOption true;
bar.battery.middleClick = mkStrOption "";
bar.battery.rightClick = mkStrOption "";
```

PerchunPak:
Hey @ppenguin,

Yes, you can. In Nix, dots are actually a syntax sugar (e.g. `a.b.c` is equal to `a = {b={c={};};}`), but what are you refering to is https://github.com/Jas-SinghFSU/HyprPanel/pull/788, and it was merged a while ago.

Also, you probably shouldn't use `mk*Option` functions as an end-user. You should instead use values directly
```nix
bar = {
  battery = {
    hideLabelWhenFull = false;
    label = true;
    middleClick = "";
    rightClick = "";
  };
};
```
Unless I am misunderstanding this, and you are trying to contribute to https://github.com/Jas-SinghFSU/HyprPanel/blob/2c0c66a8ddcdff69b2ff694fefbcd8df4387a9e7/nix/module.nix. If that's the case, don't bother updating that file. In a few days ([once hyprpanel hits nixos-unstable channel](https://nixpk.gs/pr-tracker.html?pr=379994)) I will replace it completely by the system in nixpkgs & home-manager.

ppenguin:
@PerchunPak 
Thanks for the quick reply. 

Yes, I was (among others) referring to the issues that were not working properly in the original module and the "incorrect" documentation on the website.  I had contributed some config workarounds in issue threads as a quick hack because I had no time to do it "correctly", so I'm happy that you did that ;)

(Indeed that `mkOption` stuff should not be used as an end user, I wasn't, it was an incorrect copy from the doc page).

Thanks again, now I can cleanup my config some time...

PerchunPak:
@khaneliman can we backport this to 25.05?

https://github.com/NixOS/nixpkgs/pull/419606

PR #7304, opened 2025-06-22
treewide: implement auto importing for programs/service modules

khaneliman:
### Description
Continuation of work started in https://github.com/nix-community/home-manager/pull/7069

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


khaneliman:
Ugh latest nixos channel has broken nix for darwin.

teto:
I wish we would avoid merging massive disruptive PRs like this one without at least an ACK or some discussion. 
I dont see an improvement between `(import ../programs/vdirsyncer/accounts.nix)` and `(import ../programs/vdirsyncer-accounts.nix)` while this breaks PRs and forks.

teto:
maybe I miss something but why must it be `./toto/default.nix` instead of `./toto.nix` ? I think we could support both, nix can distinguish folders from files. This would remove the merge conflicts on all the current PRs.

khaneliman:
>I wish we would avoid merging massive disruptive PRs like this one without at least an ACK or some discussion.
I dont see an improvement between (import ../programs/vdirsyncer/accounts.nix) and (import ../programs/vdirsyncer-accounts.nix) while this breaks PRs and forks.

We discussed this in the original PR for a bit a month ago. We kept running into merge conflicts and I was just finishing the work that was gonna be merged then. 

> maybe I miss something but why must it be `./toto/default.nix` instead of `./toto.nix` ? I think we could support both, nix can distinguish folders from files. This would remove the merge conflicts on all the current PRs.

We can import those files just fine, but we can't automatically distinguish when we should import them properly. This is to follow a `by-name` import strategy like we do in nixpkgs. Technically, you can still do as you said with manual work. I just migrated things to the new scheme since it's cleaner and follows a convention to show others creating new modules. Any merge conflicts should be a simple rebase since they are just file renames.

modules/modules.nix
```diff
@@ -525,6 +74,20 @@ let
         KeePassX is no longer maintained.
       '')
     ]
+    ++ (lib.concatMap
+      (
+        dir:
+        lib.pipe dir [
+          builtins.readDir
+          (lib.filterAttrs (_path: kind: kind == "directory"))
```
teto, line 82: that's my main complaint: why filter directories ? we could have autoloaded `./services/xsuspender.nix` just fine with a blacklist instead and this would have been a lot less disruptive.

teto:
> We can import those files just fine, but we can't automatically distinguish when we should import them properly.

Maybe I am missing something here, I feel like we could have autoloaded the current files just fine, with possibly a blacklist. 
We dont use any of the by-name code so we are free to do different. I dont think the new scheme is cleaner but let's say it's subjective. In absence of clear "it's better", minimizing changes is friendlier to contributors.

modules/modules.nix
khaneliman: Created https://github.com/nix-community/home-manager/pull/7312

spikespaz:
This appears to break `disabledModules`.

```nix
disabledModules = [ "services/window-managers/hyprland.nix" ];
```

Since I author my own HM module for Hyprland, I'm now greeted with the following error after updating the `home-manager` flake:

```
error: The option `wayland.windowManager.hyprland.settings' in module `/nix/store/a898j3d6v9adsgd7a5fdpl75jx4wy3hg-source/modules/services/window-managers/hyprland' would be a parent of the following options, but its type `Hyprland configuration value' does not support nested options.
 - option(s) with prefix `wayland.windowManager.hyprland.settings.source' in module `<unknown-file>'
```

khaneliman:
> This appears to break `disabledModules`.
> 
> ```nix
> disabledModules = [ "services/window-managers/hyprland.nix" ];
> ```
> 
> Since I author my own HM module for Hyprland, I'm now greeted with the following error after updating the `home-manager` flake:
> 
> ```
> error: The option `wayland.windowManager.hyprland.settings' in module `/nix/store/a898j3d6v9adsgd7a5fdpl75jx4wy3hg-source/modules/services/window-managers/hyprland' would be a parent of the following options, but its type `Hyprland configuration value' does not support nested options.
>  - option(s) with prefix `wayland.windowManager.hyprland.settings.source' in module `<unknown-file>'
> ```

That looks expected, imo. It would just need to be pointed at the new path. But, it's going to be moved back in https://github.com/nix-community/home-manager/pull/7312

spikespaz:
@khaneliman If that's expected behavior, it's a major breaking change that prevents third-party modules from functioning correctly. That big list was providing the path for each module (which, for `hyprland.nix`, hasn't changed). As far as I'm aware, this sort of thing should have been gated by `stateVersion`, but I don't even see how that's possible without some convoluted logic.

MattSturgeon:
Frankly, no project can be expected to consider its module keys[^1] stable.

That would effectively mean no module could ever be moved or renamed; or alternatively all modules that ever get moved should set the `key` attribute to their original location.

Neither of these sound practical to me.

IMO `disabledModules`, while useful, relies on implementation details (the module `key`). Therefore, anyone using it should be ready to investigate themselves when things break.

In my opinion, this is a design flaw in a fairly niche part of the module system, not a maintenance issue in home-manager.

[^1]: usually the same as the module file location

spikespaz:
@MattSturgeon That's very reasonable. Do you think it's worth considering having an explicit top-level `moduleKey` attribute for modules?

MattSturgeon:
> Do you think it's worth considering having an explicit top-level `moduleKey` attribute for modules? 

That's already built into the module system (the `key` attribute).

Maybe this could be discussed in an RFC-style issue, as others may have a different opinion to me...

Personally I think that would add an unreasonable maintenance cost. I expect it could also lead to strange filename/key pairs, as modules' filenames gradually diverge from their original keys. Maybe even weird cases where files have swapped location with each other, without changing their keys 🫨

Additionally there are situations where there should not be a key at all, such as when the module is partially dynamic (e.g. a file that is a function _returning_ a module; the same reason `lib.modules.importApply` sets `_file` but not `key`).

But purely from a maintenance cost perspective, remembering to set a key in all modules would be a challenge.

0xda157:
> maybe I miss something but why must it be `./toto/default.nix` instead of `./toto.nix` ? I think we could support both, nix can distinguish folders from files. This would remove the merge conflicts on all the current PRs.

imo each module being it's directory is nicer, but it's not really practical to make such a large change to entire tree on a project like hm (especially for something entirely vibes based).

PR #7312, opened 2025-06-23
treewide: flatten single file modules

khaneliman:
### Description
Follow up to discussion about trying to minimize conflicts. 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


tests/default.nix
```diff
@@ -214,8 +214,14 @@ import nmtSrc {
...
-              (lib.filterAttrs (_path: kind: kind == "directory"))
-              (lib.mapAttrsToList (path: _kind: lib.path.append dir path))
+              (lib.mapAttrsToList (
+                path: kind:
+                if kind == "directory" || (kind == "regular" && lib.hasSuffix ".nix" path) then
+                  lib.path.append dir path
+                else
+                  null
+              ))
+              (lib.filter (x: x != null && !lib.hasPrefix "_" (baseNameOf (toString x))))
```
khaneliman: Do we really care to support it here? We've always done folders for tests.

teto: If we dont need to, let's not do it. Simpler is better and tests are less visible to the users.
The extra complexity is worth for autoimporting module because it is more backwards compatible.

teto:
haven't tested but that seems a lot better thank you ! good  for me.

tests/default.nix
khaneliman: I'll pull it out when I rebase after merging https://github.com/nix-community/home-manager/pull/7303 since she already rebased on the current way.

modules/modules.nix
```diff
@@ -79,8 +79,14 @@ let
...
-          (lib.filterAttrs (_path: kind: kind == "directory"))
-          (lib.mapAttrsToList (path: _kind: lib.path.append dir path))
+          (lib.mapAttrsToList (
+            path: kind:
+            if kind == "directory" || (kind == "regular" && lib.hasSuffix ".nix" path) then
+              lib.path.append dir path
+            else
+              null
+          ))
+          (lib.filter (x: x != null && !lib.hasPrefix "_" (baseNameOf (toString x))))
```
MattSturgeon: I think this filter should be earlier; we should be checking if the file/directory being scanned starts with `_`, not whether the final basename does (often the basename here is `default.nix`).

E.g.

```nix
(lib.filterAttrs (path: kind: !lib.hasPrefix "_" path && (kind == "directory" || (kind == "regular" && lib.hasSuffix ".nix" path)))) 
(lib.mapAttrsToList (path: _kind: lib.path.append dir path))
```

At that point I'd probably want to split up the filter, either finding a clearer formatting, or just succumb to multiple filter calls:

```nix
(lib.filterAttrs (path: _kind: !lib.hasPrefix "_" path))
(lib.filterAttrs (_path: kind: kind != "directory" -> (kind == "regular" && lib.hasSuffix ".nix" path)))
(lib.mapAttrsToList (path: _kind: lib.path.append dir path))
```

---

Also, can we get this documented somewhere? E.g. a local in-directory README that states nix files and directories are auto-imported unless they start with `_`? As a follow-up PR, perhaps?

khaneliman: >I think this filter should be earlier; we should be checking if the file/directory being scanned starts with _, not whether the final basename does (often the basename here is default.nix).

Good catch, I think I had it earlier and ended up moving it by accident when trying to simplify the logic... 

> Also, can we get this documented somewhere? E.g. a local in-directory README that states nix files and directories are auto-imported unless they start with _? As a follow-up PR, perhaps?

Yeah, I think some of the contributing docs need to be updated a bit. I can do that.

MattSturgeon (approved):

PR #7316, opened 2025-06-24
quickshell: add module

justDeeevin:
### Description

Adds module for quickshell

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/quickshell.nix
+++ b/modules/programs/quickshell.nix
@@ -1,0 +1,90 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.programs.quickshell;
+in
+{
+  meta.maintainers = [ lib.hm.maintainers.justdeeevin ];
+
+  options.programs.quickshell = {
+    enable = lib.mkEnableOption "quickshell, a flexbile QtQuick-based desktop shell toolkit.";
+    package = lib.mkPackageOption pkgs "quickshell" { nullable = true; };
+    configs = lib.mkOption {
+      type = lib.types.attrsOf lib.types.path;
+      default = { };
+      description = ''
+        A set of configs to include in the quickshell config directory. The key is the name of the config.
+
+        The configuration that quickshell should use can be specified with the `activeConfig` option.
+      '';
+    };
+    activeConfig = lib.mkOption {
+      type = lib.types.nullOr lib.types.str;
+      default = null;
+      description = ''
+        The name of the config to use.
+
+        If `null`, quickshell will attempt to use a config located in `$XDG_CONFIG_HOME/quickshell` instead of one of the named sub-directories.
+      '';
+    };
+
+    systemd = {
+      enable = lib.mkEnableOption "quickshell systemd service";
+      target = lib.mkOption {
+        type = lib.types.str;
+        default = config.wayland.systemd.target;
+        defaultText = lib.literalExpression "config.wayland.systemd.target";
+        example = "hyprland-session.target";
+        description = ''
+          The systemd target that will automatically start quickshell.
+
+          If you set this to a WM-specific target, make sure that systemd integration for that WM is enabled (e.g. `wayland.windowManager.hyprland.systemd.enable`). **This is typically true by default**.
+        '';
+      };
+    };
+  };
+
+  config = lib.mkIf cfg.enable (
+    lib.mkMerge [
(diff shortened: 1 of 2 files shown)


modules/programs/quickshell.nix
```diff
@@ -0,0 +1,69 @@
...
+  meta.maintainers = [ lib.hm.maintainers.justdeeevin ];
+
+  options.programs.quickshell = {
+    enable = lib.mkEnableOption "quickshell, a flexbile QtQuick-based desktop shell toolkit.";
+    package = lib.mkPackageOption pkgs "quickshell" { nullable = true; };
+    configDir = lib.mkOption {
+      type = lib.types.nullOr lib.types.path;
+      default = null;
+      description = "Directory containing the configuration files for quickshell.";
+    };
```
khaneliman: Reading the upstream docs this seems like it would make more sense to be an option that accepts an attrset of paths so you can import/configure different configs. https://quickshell.outfoxxed.me/docs/guide/distribution/ 

Otherwise, it wouldn't be composable with different modules providing different configurations.

justDeeevin: I had no idea about the named config feature! Thanks for letting me know

modules/programs/quickshell.nix
```diff
@@ -0,0 +1,90 @@
...
+  options.programs.quickshell = {
+    enable = lib.mkEnableOption "quickshell, a flexbile QtQuick-based desktop shell toolkit.";
+    package = lib.mkPackageOption pkgs "quickshell" { nullable = true; };
+    configs = lib.mkOption {
+      type = lib.types.attrsOf lib.types.path;
+      default = { };
+      description = ''
+        A set of configs to include in the quickshell config directory. The key is the name of the config.
+
+        The config quickshell should use can be specified with the `config` option.
```
khaneliman:
```suggestion
        The configuration that quickshell service will use can be set with the `activeConfig` option.
```

modules/programs/quickshell.nix
```diff
@@ -0,0 +1,90 @@
...
+      '';
+    };
+    activeConfig = lib.mkOption {
+      type = lib.types.nullOr lib.types.str;
+      default = null;
+      description = ''
+        The name of the config to use.
+
+        If null, quickshell will attempt to use a config located in `$XDG_CONFIG_HOME/quickshell` instead of one of the named sub-directories.
```
khaneliman:
```suggestion
        If `null`, quickshell will attempt to use a config located in `$XDG_CONFIG_HOME/quickshell` instead of one of the named sub-directories.
```

modules/programs/quickshell.nix
```diff
@@ -0,0 +1,90 @@
...
+  cfg = config.programs.quickshell;
+in
+{
+  meta.maintainers = [ lib.hm.maintainers.justdeeevin ];
+
+  options.programs.quickshell = {
+    enable = lib.mkEnableOption "quickshell, a flexbile QtQuick-based desktop shell toolkit.";
+    package = lib.mkPackageOption pkgs "quickshell" { nullable = true; };
+    configs = lib.mkOption {
+      type = lib.types.attrsOf lib.types.path;
```
khaneliman, line 17: Might be nice to support `lib.types.lines` in the future with `lib.hm.types.sourceFileOrLines`. I don't know if it is setup to support this usecase of a dynamic location, atm.

modules/programs/quickshell.nix
```diff
@@ -0,0 +1,90 @@
...
+      default = { };
+      description = ''
+        A set of configs to include in the quickshell config directory. The key is the name of the config.
+
+        The config quickshell should use can be specified with the `config` option.
+      '';
+    };
+    activeConfig = lib.mkOption {
```
khaneliman, line 25: What is the behavior if this is unset? Does it just randomly choose a config? Alphabetical?

khaneliman (commented):
Looks a lot better, just a couple nits on verbiage for more readability.

modules/programs/quickshell.nix
justDeeevin: quickshell will look for `~/.config/quickshell/shell.nix` if no config name is specified with `--config`

modules/programs/quickshell.nix
justDeeevin: thanks for catching that typo! (`config` -> `activeConfig`).

PR #7328, opened 2025-06-25
hyprpanel: deprecate `theme.name` option

PerchunPak:
### Description

This option was provided in the original module, but I didn't implement it because it would require IFD. This just adds a warning so the user won't be confused if it doesn't work.

Pointed out in https://github.com/Jas-SinghFSU/HyprPanel/issues/1023#issuecomment-3000694765

This would also need a backport

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/hyprpanel/default.nix
```diff
@@ -91,7 +91,26 @@ in
 
     xdg.configFile.hyprpanel = lib.mkIf (cfg.settings != { }) {
       target = "hyprpanel/config.json";
-      source = jsonFormat.generate "hyprpanel-config" cfg.settings;
+      source = jsonFormat.generate "hyprpanel-config" (
+        if cfg.settings ? theme && cfg.settings.theme ? name then
```
khaneliman, line 95: This is a freeform type using a generator. It doesn't make sense to implement a warning on a deprecation, that never actually was implemented here. It is a warning / documentation that should be provided upstream, for users of their flake and migrating to home-manager.

PerchunPak: Users who used upstream's module were automatically migrated to this module, so this is a continuation of the old module. It is not backward compatible, but we should at least print a warning

khaneliman: Oh that's true. Forgot about the same module option name causing the conflict/migration.

PR #7333, opened 2025-06-27
zsh: add options zsh.history.{appendInc,appendIncTime}

phip1611:
### Description

This adds the ZSH options `INC_APPEND_HISTORY` and
`INC_APPEND_HISTORY_TIME`.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/zsh/history.nix
+++ b/modules/programs/zsh/history.nix
@@ -33,6 +33,36 @@ in
               '';
             };
 
+            appendInc = mkOption {
+              type = types.bool;
+              default = false;
+              description = ''
+                This option works like `programs.zsh.history.append` except that
+                new history lines are added to the $HISTFILE incrementally (as
+                soon as they are entered), rather than waiting until the shell
+                exits. The file will still be periodically re-written to trim it
+                when the number of lines grows 20% beyond the value specified by
+                `programs.zsh.history.save`.
+              '';
+            };
+
+            appendIncTime = mkOption {
+              type = types.bool;
+              default = false;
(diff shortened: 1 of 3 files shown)


phip1611:
Rebased. Are we good to go now, @teto @khaneliman?

khaneliman:
Just not a fan of this option set in the current form. It is too verbose and maintenance heavy to keep adding a new option for each possible `setopt` and `unsetopt`. We currently have `setOptions ` available to do the same thing. 

Seems more like we should deprecate these options and encourage usage of that `setOptions` option? Or is this something that needs to be set in a specific location in the file?

phip1611:
I agree that this is verbose. I wasn't aware of `setOptions`. Should be sufficient!

Personally, my favorite form would be something like:

```nix
setOptions = 
          lib.mapAttrsToList (name: enabled: if enabled then name else "NO_${name}") {
            APPEND_HISTORY = false;
            EXTENDED_HISTORY = true;
            HIST_EXPIRE_DUPS_FIRST = false;
            HIST_FCNTL_LOCK = true;
            HIST_FIND_NO_DUPS = false;
            HIST_IGNORE_ALL_DUPS = false;
            HIST_IGNORE_DUPS = true;
            HIST_IGNORE_SPACE = true;
            HIST_SAVE_NO_DUPS = false;
            INC_APPEND_HISTORY = true;
            SHARE_HISTORY = false;
          };
```

This helps to keep alphabetical sorting. But the current setOptions is also fine.

PR #7334, opened 2025-06-27
ashell: Update to 0.5.0 config

CodedNil:
### Description

Update the ashell module to use the new config file location and toml instead of yaml as per https://github.com/MalpenZibo/ashell/releases/tag/0.5.0

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@justdeeevin


justDeeevin:
This will silently break any existing configs using this module given the many changes to option names with 0.5.0. This should probably warrant a news entry or (if possible) warning

justDeeevin:
Good work otherwise :>

CodedNil:
Created the news entry (hopefully did that right?). Should warn if the user is using the ashell module.

modules/misc/news/2025/06/2025-06-27_18-53-10.nix
```diff
@@ -0,0 +1,13 @@
+{ config, ... }:
+
+{
+  time = "2025-06-27T18-53-10+00:00";
+  condition = config.programs.ashell.enable;
```
justDeeevin:
```suggestion
  condition = config.programs.ashell.enable && (config.programs.ashell.settings != { });
```

justDeeevin (approved):

khaneliman:
Would be more user friendly to use a `lib.versionAtLeast` check to determine which config location and formatter to use based on the module's configured package version.

CodedNil:
> Would be more user friendly to use a `lib.versionAtLeast` check to determine which config location and formatter to use based on the module's configured package version.

I'm not sure how to implement this sorry, happy to have a go but does anyone have any tips?
I'm guessing start with a variable for formatVersion which uses that lib.versionAtLeast to check the package version but not sure how to start with that?
Then use if else on the settingsFile and config location based on that?

justDeeevin:
> and formatter

So it would automatically change the case of the option names?

khaneliman:
Pushed an idea for supporting both. See if it still works for you.

modules/programs/ashell.nix
```diff
@@ -6,7 +6,13 @@
...
 let
   cfg = config.programs.ashell;
-  settingsFormat = pkgs.formats.toml { };
+  tomlFormat = pkgs.formats.toml { };
+  yamlFormat = pkgs.formats.yaml { };
+
+  packageVersion = if cfg.package != null then lib.getVersion cfg.package else "0.5.0";
+  isTomlConfig = lib.versionAtLeast packageVersion "0.5.0";
+  settingsFormat = if isTomlConfig then tomlFormat else yamlFormat;
+  configFileName = if isTomlConfig then "config.toml" else "config.yaml";
```
CodedNil: This won't work with older version I believe as the config.yaml file used to be located directly in .config not in the ashell dir
This should fix it I believe
  configFileName = if isTomlConfig then "ashell/config.toml" else "config.yaml";

modules/programs/ashell.nix
```diff
@@ -69,7 +76,7 @@ in
         ];
 
         home.packages = lib.mkIf (cfg.package != null) [ cfg.package ];
-        xdg.configFile."ashell/config.toml" = lib.mkIf (cfg.settings != { }) {
+        xdg.configFile."ashell/${configFileName}" = lib.mkIf (cfg.settings != { }) {
```
CodedNil: Fix config file location with
xdg.configFile."${configFileName}" = lib.mkIf (cfg.settings != { }) {

modules/programs/ashell.nix
khaneliman: Thanks for catching! Updated

khaneliman:
@CodedNil @justDeeevin pushed that fix, if you want to look again

khaneliman:
> This will silently break any existing configs using this module given the many changes to option names with 0.5.0. This should probably warrant a news entry or (if possible) warning

I did wonder about that myself was reviewing documentation, as well.. I did add some lib helper functions for converting `camelCase` to `snake_case` we could leverage on the `settings` passed in when we detect its a version `0.5.0`

khaneliman:
Just curious if there are any runtime notifications from the application when a config uses old key names with `camelCase` or if we need to add the apply function that converts `camelCase` to `snake_case` for the users, to avoid surprising breakage.

justDeeevin:
I'm not aware of any such help from the app

khaneliman:
Pushed another idea for helping with migration while giving some warnings.

khaneliman:
Alright, wanted to try this application out a bit since I've been using `waybar` for years and wanted to try something new. In testing the migration logic, I noticed they still kept `CustomModule` name casing and I wasn't handling that properly. 

EDIT: Sorry, realized I just kinda took over your PR... hope you don't mind. Found it an interesting problem to solve.

PR #7337, opened 2025-06-28
feat(tmux): move config file `~/.config/tmux.conf` -> `~/.tmux.conf`

phucisstupid:
## Why this change?
- Plugin auto-loading issue: On macOS, tmux does not automatically source ~/.config/tmux/tmux.conf. As a result, plugin don’t work unless the config is manually sourced with:
```
tmux source ~/.config/tmux/tmux.conf
```

- Platform consistency: While ~/.config/tmux/tmux.conf is XDG-compliant, tmux on macOS does not follow this behavior reliably. The move aligns with tmux’s default expectations and common practice.

## ✅ Result
- Tmux plugins now load correctly on startup.
- Manual sourcing is no longer needed.
- Configuration becomes more portable and idiomatic, especially for macOS users.

### DIFF
--- a/modules/programs/tmux.nix
+++ b/modules/programs/tmux.nix
@@ -132,7 +132,7 @@ let
       )
     ];
 
-    xdg.configFile."tmux/tmux.conf".text = ''
+    home.file.".tmux.conf".text = ''
       # ============================================= #
       # Load plugins with Home Manager                #
       # --------------------------------------------- #
@@ -209,7 +209,7 @@ in
         default = "";
         description = ''
           Additional configuration to add to
-          {file}`tmux.conf`.
+          {file}`.tmux.conf`.
         '';
       };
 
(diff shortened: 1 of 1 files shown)


ambroisie:
I find it surprising that `XDG_CONFIG_HOME` doesn't work on MacOS: [`TMUX_CONF`](https://github.com/tmux/tmux/blob/e7f15d09be61fd0b9d54f6fec2f47ace4008b4dc/Makefile.am#L14) is set to the same value all OSes.

kpbaks:
If this is implemented it should be predicated on the host platform, to keep existing behavior on Linux systems

aguirre-matteo:
As @kpbaks said, it would be better to place the config file at `~/.tmux.conf` only if the host platform is Darwin, while on Linux we keep the file at `$XDG_CONFIG_HOME/tmux.com`. By the way, you should update the tests to point to the new config location.

PR #7339, opened 2025-06-29
direnv: fix nushell cell-path handling

koffydrop:
### Description
Since version 0.105.0, cell-paths are case-sensitive by default, since the record we are using is lowercase it doesn't convert `$env.PATH`.

Alternatively, instead of making it uppercase, we could build an actual `cell-path` value instead of relying on the string value of `$key`, which would comply with `$env` remaining a case-insensitive record (this also seems backwards compatible).
```nu
let key_path = [ # becomes $.<key>!
	[value, optional, insensitive];
	[$key, false, true]
]
get -i $key_path
```

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@khaneliman, @rycee, @shikanime

### DIFF
--- a/modules/programs/direnv.nix
+++ b/modules/programs/direnv.nix
@@ -187,13 +187,13 @@ in
                 | items {|key, value|
                     let value = do (
                         {
-                          "path": {
+                          "PATH": {
                             from_string: {|s| $s | split row (char esep) | path expand --no-symlink }
                             to_string: {|v| $v | path expand --no-symlink | str join (char esep) }
                           }
                         }
                         | merge ($env.ENV_CONVERSIONS? | default {})
-                        | get -i $key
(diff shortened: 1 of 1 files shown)


khaneliman:
>Since version 0.105.0, cell-paths are case-sensitive by default, since the record we are using is lowercase it doesn't convert $env.PATH.
>
>Alternatively, instead of making it uppercase, we could build an actual cell-path value instead of relying on the string value of $key, which would comply with $env remaining a case-insensitive record (this also seems backwards compatible).

I'm for whatever method has backwards compatibility or doesn't unsuspectingly break a user's config. If that's using a version check for which string to use or a method that it doesn't matter.

koffydrop:
There's a very small chance the first solution isn't backwards compatible, since the behavior is technically different. I'll change it to the other solution.

PR #7340, opened 2025-06-29
Numbat: add initFile

Aehmlo:
### Description

This PR picks up where #7159 left off, providing an `initFile` option for `programs.numbat` to allow the user to supply custom units, functions, and constants for Numbat to use.

The first commit in this PR implements this using just (`lib.types.nullOr`) `lib.types.lines`; the second additionally allows specifying a path to the `init.nbt` file to use as a source. Though I was drawn to `lib.hm.types.sourceFileOrLines` initially, it has some extra logic around directories that is unnecessary (and perhaps confusing) for this.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/numbat.nix
+++ b/modules/programs/numbat.nix
@@ -40,13 +40,14 @@ in
     };
 
     initFile = lib.mkOption {
-      type = lib.types.nullOr lib.types.lines;
+      type = lib.types.nullOr (lib.types.either lib.types.lines lib.types.path);
       default = null;
       example = ''
         unit kohm: ElectricResistance = kV/A
       '';
       description = ''
-        User initialization file ({file}`init.nbt`) contents.
+        User initialization file ({file}`init.nbt`) contents. May be specified
+        inline or as a path to a source file.
       '';
     };
   };
@@ -55,7 +56,7 @@ in
     home.packages = mkIf (cfg.package != null) [ cfg.package ];
 
     home.file."${configDir}/init.nbt" = mkIf (cfg.initFile != null) {
-      text = cfg.initFile;
+      source = if lib.isString cfg.initFile then pkgs.writeText "init.nbt" cfg.initFile else cfg.initFile;
     };
 
     home.file."${configDir}/config.toml" = mkIf (cfg.settings != { }) {
--- a/modules/programs/numbat.nix
+++ b/modules/programs/numbat.nix
@@ -38,10 +38,25 @@ in
         <https://numbat.dev/doc/cli-customization.html#configuration> for options.
       '';
     };
+
+    initFile = lib.mkOption {
+      type = lib.types.nullOr lib.types.lines;
+      default = null;
+      example = ''
+        unit kohm: ElectricResistance = kV/A
+      '';
+      description = ''
+        User initialization file ({file}`init.nbt`) contents.
+      '';
+    };
   };
 
   config = mkIf cfg.enable {
     home.packages = mkIf (cfg.package != null) [ cfg.package ];
+
+    home.file."${configDir}/init.nbt" = mkIf (cfg.initFile != null) {
+      text = cfg.initFile;
+    };
 
     home.file."${configDir}/config.toml" = mkIf (cfg.settings != { }) {
       source = tomlFormat.generate "numbat-config" cfg.settings;
--- a/tests/modules/programs/numbat/default.nix
+++ b/tests/modules/programs/numbat/default.nix
@@ -1,4 +1,5 @@
 {
   numbat-example-config = ./example-config.nix;
   numbat-empty-config = ./empty-config.nix;
+  numbat-external-initfile = ./external-initfile.nix;
 }
--- a/tests/modules/programs/numbat/empty-config.nix
+++ b/tests/modules/programs/numbat/empty-config.nix
@@ -14,5 +14,6 @@ in
 
   nmt.script = ''
     assertPathNotExists 'home-files/${configDir}/config.toml'
+    assertPathNotExists 'home-files/${configDir}/init.nbt'
   '';
 }
--- a/tests/modules/programs/numbat/example-config.nix
+++ b/tests/modules/programs/numbat/example-config.nix
@@ -19,6 +19,9 @@ in
       prompt = "> ";
       exchange-rates.fetching-policy = "on-first-use";
     };
+    initFile = ''
+      unit kohm: ElectricResistance = kV/A
+    '';
   };
 
   nmt.script = ''
@@ -30,6 +33,11 @@ in
 
         [exchange-rates]
         fetching-policy = "on-first-use"
+      ''}
+    assertFileExists 'home-files/${configDir}/init.nbt'
+    assertFileContent $(normalizeStorePaths 'home-files/${configDir}/init.nbt') \
+      ${builtins.toFile "expected-init.nbt" ''
+        unit kohm: ElectricResistance = kV/A
       ''}
   '';
 }
--- a/tests/modules/programs/numbat/external-initfile.nix
+++ b/tests/modules/programs/numbat/external-initfile.nix
@@ -1,0 +1,27 @@
+{
+  config,
+  pkgs,
+  ...
+}:
+let
+  configDir =
+    if pkgs.stdenv.hostPlatform.isDarwin then
+      "Library/Application Support/numbat"
+    else
+      ".config/numbat";
+in
+{
+  programs.numbat = {
+    enable = true;
+    package = config.lib.test.mkStubPackage { };
+    initFile = ./init.nbt;
+  };
+
+  nmt.script = ''
+    assertFileExists 'home-files/${configDir}/init.nbt'
+    assertFileContent $(normalizeStorePaths 'home-files/${configDir}/init.nbt') \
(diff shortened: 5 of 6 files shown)


teto (changes requested):
same comment as in the linked PR, this does not add any value over `xdg.configFile."numbat/init.nbt"`, just more maintainance overhead

Aehmlo:
> this does not add any value over xdg.configFile."numbat/init.nbt", just more maintainance overhead

While I understand and sympathize with the sentiment, this isn't strictly true, since Numbat uses different configuration locations across Darwin and Linux, which this option abstracts away (this actually is why I initially contributed the Numbat module).

If the verdict ends up being that this just isn't worth the extra code, I can always just publish this as part of my own configuration repo or something, but I'm happy to continue maintaining it for the foreseeable future.

modules/programs/numbat.nix
```diff
@@ -38,11 +38,27 @@ in
         <https://numbat.dev/doc/cli-customization.html#configuration> for options.
       '';
     };
+
+    initFile = lib.mkOption {
+      type = lib.types.nullOr (lib.types.either lib.types.lines lib.types.path);
```
teto, line 43: I find it weird (it might exist elsewhere) to mix path and content though not blocking.

teto:
> While I understand and sympathize with the sentiment, this isn't strictly true, since Numbat uses different configuration locations across Darwin and Linux, which this option abstracts away (this actually is why I initially contributed the Numbat module).

that would be a valid reason to have this in HM imo but if the module was using `xdg.configFile` instead of `home.file` wouldn't that solve it on mac (ie., does xdg.configFile redirect to Library/Application support... on darwin ?). I think the module should rely on xdg.configFile anyway since the `home.file.".config/numbat"` wont work for someone changing its XDG_CONFIG_HOME via HM.

Aehmlo:
> ie., does xdg.configFile redirect to Library/Application support... on darwin ?

Short answer: [no](https://github.com/nix-community/home-manager/blob/7d9e3c35f0d46f82bac791d76260f15f53d83529/modules/misc/xdg.nix#L25).

<details>
<summary>Longer answer</summary>

There's a bit of a…nonuniform ecosystem on Darwin. The typically-cited [guidance from Apple](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/MacOSXDirectories/MacOSXDirectories.html#//apple_ref/doc/uid/TP40010672-CH10-SW1) is to use subdirectories of `~/Library/Application Support` to store "all app-specific data and support files." This leads to a bit of a question as to what an "app" is; some people interpret this to include CLI programs, while others interpret it as solely encompassing GUI programs. Under the hood, Numbat [uses the `dirs` crate](https://github.com/sharkdp/numbat/blob/3fd7464d2020e4a82df9eca7b7f07ed9229c4b2b/numbat-cli/Cargo.toml#L18) to source configuration. Somewhat controversially, the authors of this crate have [made it clear](https://github.com/dirs-dev/directories-rs/issues/47#issuecomment-478337412) that they consider `~/Library` to be the correct location for all application-related files for programs running on Darwin, including CLI tools. Since `dirs` is something of an unofficial standard within the Rust ecosystem, this leads to many Rust programs, including Numbat, unconditionally storing their configuration files within `~/Library/Application Support`.

In my opinion, the current behavior of `home-manager`'s XDG options (namely, forgoing `~/Library/Application Support` in favor of true Linux-style XDG) is sensible and should not be changed. [Several](https://github.com/nix-community/home-manager/blob/7d9e3c35f0d46f82bac791d76260f15f53d83529/modules/programs/tiny.nix#L10-L14) [existing](https://github.com/nix-community/home-manager/blob/7d9e3c35f0d46f82bac791d76260f15f53d83529/modules/programs/poetry.nix#L19) [in-tree](https://github.com/nix-community/home-manager/blob/7d9e3c35f0d46f82bac791d76260f15f53d83529/modules/programs/clock-rs.nix#L51) [modules](https://github.com/nix-community/home-manager/blob/7d9e3c35f0d46f82bac791d76260f15f53d83529/modules/programs/tex-fmt.nix#L17) work around the `Application Support` split in an ad-hoc fashion, much like the style used here. While I personally would love to see an abstraction developed that reduces the need for ad-hoc specifications like these (and the one in the Numbat module), ultimately there's no way around the underlying split between application behavior, and so we'll need some kind of special logic for these programs on Darwin.

I may try to write an abstraction around the XDG/`Application Support` split in a separate PR in the future, but to summarize: the `xdg.configFile` options do *not* store files in `Application Support` (making them unsuitable for configuring Numbat on Darwin), and I do not believe that they should (though it may make sense for another wrapper to in the future in an opt-in fashion).
</details>

> I think the module should rely on xdg.configFile anyway since the home.file.".config/numbat" wont work for someone changing its XDG_CONFIG_HOME via HM.

As implemented, the module uses `config.xdg.configHome` to identify where to store the config file; `.config/numbat` is only hardcoded in tests (but if desired, I can change this to reference `config.xdg.configHome` too).

modules/programs/numbat.nix
Aehmlo: This pattern is present in-tree in a few other places; I believe I took inspiration largely from [the waybar module](https://github.com/nix-community/home-manager/blob/7d9e3c35f0d46f82bac791d76260f15f53d83529/modules/programs/waybar.nix#L253) for this PR. If there's a different type or structure that makes more sense, though, I'm open to suggestions.

khaneliman:
> There's a bit of a…nonuniform ecosystem on Darwin. The typically-cited [guidance from Apple](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/MacOSXDirectories/MacOSXDirectories.html#//apple_ref/doc/uid/TP40010672-CH10-SW1) is to use subdirectories of ~/Library/Application Support to store "all app-specific data and support files." This leads to a bit of a question as to what an "app" is; some people interpret this to include CLI programs, while others interpret it as solely encompassing GUI programs. Under the hood, Numbat [uses the dirs crate](https://github.com/sharkdp/numbat/blob/3fd7464d2020e4a82df9eca7b7f07ed9229c4b2b/numbat-cli/Cargo.toml#L18) to source configuration. Somewhat controversially, the authors of this crate have https://github.com/dirs-dev/directories-rs/issues/47#issuecomment-478337412 that they consider ~/Library to be the correct location for all application-related for programs running on Darwin, including CLI tools. Since dirs is something of an unofficial standard within the Rust ecosystem, this leads to many Rust programs, including Numbat, unconditionally storing their configuration files within ~/Library/Application Support.

Wow, interesting to see such a strong opinion on what seems like an obviously bad take. 

> In my opinion, the current behavior of home-manager's XDG options (namely, forgoing ~/Library/Application Support in favor of true Linux-style XDG) is sensible and should not be changed. [Several](https://github.com/nix-community/home-manager/blob/7d9e3c35f0d46f82bac791d76260f15f53d83529/modules/programs/tiny.nix#L10-L14) [existing](https://github.com/nix-community/home-manager/blob/7d9e3c35f0d46f82bac791d76260f15f53d83529/modules/programs/poetry.nix#L19) [in-tree](https://github.com/nix-community/home-manager/blob/7d9e3c35f0d46f82bac791d76260f15f53d83529/modules/programs/clock-rs.nix#L51) [modules](https://github.com/nix-community/home-manager/blob/7d9e3c35f0d46f82bac791d76260f15f53d83529/modules/programs/tex-fmt.nix#L17) work around the Application Support split in an ad-hoc fashion, much like the style used here. While I personally would love to see an abstraction developed that reduces the need for ad-hoc specifications like these (and the one in the Numbat module), ultimately there's no way around the underlying split between application behavior, and so we'll need some kind of special logic for these programs on Darwin.
>
> I may try to write an abstraction around the XDG/Application Support split in a separate PR in the future, but to summarize: the xdg.configFile options do not store files in Application Support (making them unsuitable for configuring Numbat on Darwin), and I do not believe that they should (though it may make sense for another wrapper to in the future in an opt-in fashion).

Yeah, I have thought about this numerous times and never took the time to do it. I've started implementing more and more lib functions to try and reduce different modules having to rewrite the same thing over and over. This is an area that could benefit from it. Would be along the lines of a function that returns the path for darwin and an argument of whether its XDG compliant (seemingly will never be a Rust program).

modules/programs/numbat.nix
khaneliman: Personally, I like the flexibility of options that support both. I've tried to implement it wherever I can. We also have a special lib type for supporting directories (for more complex configurations) that I use in sketchybar: https://github.com/nix-community/home-manager/blob/master/modules/programs/sketchybar.nix#L48-L48 I agree that it doesn't make sense in this use case because you aren't needing the directory support. I might tweak the function name and make it more obvious about use case.

Aehmlo:
> Yeah, I have thought about this numerous times and never took the time to do it. I've started implementing more and more lib functions to try and reduce different modules having to rewrite the same thing over and over. This is an area that could benefit from it. Would be along the lines of a function that returns the path for darwin and an argument of whether its XDG compliant (seemingly will never be a Rust program).

Glad to hear there's interest in this! I anticipate having some spare time in the next couple of weeks; I'll try to put together a PR in this direction to make things a little DRY-er.

PR #7341, opened 2025-06-29
Nh: fix clean option for darwin

A-jay98:
### Description
This PR fixes the `nh.clean` option's behaviour in Darwin systems. Previously, this option failed silently for Darwin systems as it only configured a systemd unit. I've added launchd configuration if the system is Darwin. 

P.S: This is my first PR in this project. I've tried to read all the contribution guides carefully and do my best to follow them. Sorry if I've missed anything 😅 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@JohnRTitor


modules/programs/nh.nix
```diff
@@ -8,7 +8,10 @@ let
   cfg = config.programs.nh;
 in
 {
-  meta.maintainers = with lib.maintainers; [ johnrtitor ];
+  meta.maintainers = with lib.maintainers; [
+    johnrtitor
+    a-jay98
+  ];
```
khaneliman, line 14: separate commit, please

modules/programs/nh.nix
```diff
@@ -90,6 +99,21 @@ in
...
+
+    launchd.agents.nh-clean = lib.mkIf (cfg.clean.enable && pkgs.stdenv.isDarwin) {
+      enable = true;
+      config = {
+        ProgramArguments = [
+          "${lib.getExe cfg.package}"
+          "clean"
+          "user"
+        ] ++ lib.optional (cfg.clean.extraArgs != "") cfg.clean.extraArgs;
+
```
khaneliman, line 113:
```suggestion
        StandardErrorPath = "${config.home.homeDirectory}/Library/Logs/nh/nh.err.log";
        StandardOutPath = "${config.home.homeDirectory}/Library/Logs/nh/nh.out.log";
```

Would be nice to see whats going on

khaneliman (commented):
image

With logging changes I can see that it just exits after the first failure it receives but looks like an upstream issue.

modules/programs/nh.nix
A-jay98: Thank you for the review!
I'm hesitant to add this. Logs stored in these files aren't automatically rotated or cleaned up, so I'd rather have enabling this be an explicit action instead of something that's on by default. To me, enabling garbage collection is more of a "set it and forget it" type of thing. I'd prefer to see the logs only if something isn’t working properly—and even then, I'm okay with not having historical logs. I’d just enable it at that point or use the log show command to search for them.

Do you think this is a must-have?

modules/programs/nh.nix
A-jay98: Done.

modules/programs/nh.nix
khaneliman: Nah, that's fair. I can add it later as an opt in behavior.

Issue #7336, opened 2025-06-28
bug: oh-my-posh nushell integration failed


tsssni:
### Are you following the right branch?

- [x] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Issue description

The generated `extraConfig` of nushell says:
```nu
echo "Failed to write init script: exec: "nu": executable file not found in $PATH"
```

### Maintainer CC

@jaredmontoya 

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.12.34, NixOS, 25.11 (Xantusia), 25.11.20250624.30a61f0`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.28.3`
 - nixpkgs: `/nix/store/1l4nm55xcq55cdp1xz3x5mfgr4c24058-source`
```

mhemeryck:
I have the same issue; I did notice that from oh-my-posh version 26 there's a breaking change mentioned in the release notes: https://github.com/JanDeDobbeleer/oh-my-posh/releases/tag/v26.0.0

I'm not 100% though what the actual issue is though.

jaredmontoya:
@mhemeryck thank you for pinpointing the version where the change happened.

oh-my-posh decided to completely change the way you install it for nushell.
after version 26 `oh-my-posh init nu --print` no longer prints anything but creates a file in `~/.local/share/nushell/vendor/autoload/` that contains the configuration and is loaded by nushell.

Naturally such a fundamental change would break the current way home-manager handles oh-my-posh.

jaredmontoya:
@mhemeryck or @tsssni please test if PR #7342 fixes your issue
I use starship instead of oh-my-posh so it's better if an oh-my-posh user tests the changes.

https://github.com/jaredmontoya/home-manager/tree/fix-oh-my-posh-update
```nix
inputs.home-manager.url = "github:jaredmontoya/home-manager/fix-oh-my-posh-update";
```

Comment on #7342 if the issues are gone.

jaredmontoya:
@tsssni if the issue is solved please close it.

PR #7342, opened 2025-06-29
oh-my-posh: fix nushell with v26.0.0

jaredmontoya:
### Description
See #7336

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/tests/modules/programs/oh-my-posh/nushell.nix
+++ b/tests/modules/programs/oh-my-posh/nushell.nix
@@ -28,8 +28,18 @@
     in
     ''
       assertFileExists "${configFile}"
-      assertFileRegex \
-        "${configFile}" \
-        'source /nix/store/[^/]*-oh-my-posh-nushell-config.nu'
+      ${
+        if (lib.versionAtLeast (lib.versions.major config.programs.oh-my-posh.package.version) "26") then
+          ''
+            assertFileContains \
+                    "${configFile}" \
+                    "/bin/oh-my-posh init nu --config"''
+        else
+
+          ''
+            assertFileRegex \
+                    "${configFile}" \
+                    "source /nix/store/[^/]*-oh-my-posh-nushell-config.nu"''
+      }
     '';
 }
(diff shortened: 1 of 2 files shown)


tsssni:
It works in my machine :)

mhemeryck:
Same here, also fixes the issue for me:

```
$ oh-my-posh --version
26.8.0
$ nu --version
0.105.1
```

jaredmontoya:
@khaneliman

khaneliman:
Is this compatible with previous versions or do we need a `lib.versionAtleast` check?

jaredmontoya:
@khaneliman added a version check.

tests/modules/programs/oh-my-posh/nushell.nix
```diff
@@ -28,8 +28,18 @@
...
+            assertFileContains \
+                    "${configFile}" \
+                    "/bin/oh-my-posh init nu --config"''
+        else
+
+          ''
+            assertFileRegex \
+                    "${configFile}" \
+                    "source /nix/store/[^/]*-oh-my-posh-nushell-config.nu"''
+      }
```
khaneliman, line 43: Would have preferred to stub instead of using the realPkg. But, won't block on that.

jaredmontoya:
@tsssni @mhemeryck I also have to warn you that because of that change there is now no way for home manager to uninstall oh-my-posh from nushell because it has no idea about the file that oh-my-posh creates in `~/.local/share/nushell/vendor/autoload` and users will have to manually remove that file after they remove oh-my-posh from their home-maanger config.

mhemeryck:
> @tsssni @mhemeryck I also have to warn you that because of that change there is now no way for home manager to uninstall oh-my-posh from nushell because it has no idea about the file that oh-my-posh creates in `~/.local/share/nushell/vendor/autoload` and users will have to manually remove that file after they remove oh-my-posh from their home-maanger config.

Right; that's because the config script now gets generated instead of just sourced. Good to keep in mind.

Thanks anyway for the quick fix!

PR #7349, opened 2025-06-30
gtk: allow more granular customization

khaneliman:
### Checklist
Closes https://github.com/nix-community/home-manager/issues/5133

    ```
    {component}: {description}

    {long description}
    ```


khaneliman:
@zowoq Do you know if the Github App needs different permissions to read the team/member list ? https://github.com/nix-community/home-manager/actions/runs/16034340501/job/45242369177?pr=7349 I was able to get it to work without the app in my fork https://github.com/nix-community/home-manager/pull/6921

From what I have read, it looks like the github app just needs its GraphQL permissions updated in the organization.

MattSturgeon:
> error fetching organization teams: GraphQL: Resource not accessible by integration (organization.teams)

Sounds like the app needs "Organization permissions" -> "Members" -> "Read-only"

Which I _believe_ is configured from: https://github.com/settings/apps/home-manager-ci/permissions

_(as always with app perm changes, they also need "accepting" from the repo's installation, but I assume org-owners are well aware of this :sweat_smile:)_

zowoq:
> Sounds like the app needs "Organization permissions" -> "Members" -> "Read-only"
> 
> Which I _believe_ is configured from: https://github.com/settings/apps/home-manager-ci/permissions
> 
> _(as always with app perm changes, they also need "accepting" from the repo's installation, but I assume org-owners are well aware of this 😅)_

Done.

khaneliman:
@zowoq sweet, it worked thanks. we will need the same thing done for the nixvim and stylix bots too

zowoq:
> we will need the same thing done for the nixvim and stylix bots too

I'll wait until we're adding the rfc39 service for these repos to update the app permissions.

fmbearmf:
```
home-manager.users.<user>.gtk.gtk2.force' does not exist. Definition values:
```

This appears to have removed the gtk.gtk2.force option. Is there a new way of forcefully overwriting?
Sorry if this is a dumb question.

khaneliman:
> ```
> home-manager.users.<user>.gtk.gtk2.force' does not exist. Definition values:
> ```
> 
> This appears to have removed the gtk.gtk2.force option. Is there a new way of forcefully overwriting? Sorry if this is a dumb question.

That would have been an accident, it should be backwards compatible. I'll fix ASAP.

khaneliman:
Created https://github.com/nix-community/home-manager/pull/7437 that should fix it @fmbearmf

PR #7356, opened 2025-07-01
launchd+targets/darwin: Escape XML in plists

linnnus:
### Description

https://github.com/NixOS/nixpkgs/pull/356502 introduced an argument escape to lib.generators.toPlist which makes it escape special characters in the XML. This PR makes nix-darwin use it everywhere.

This is a breaking change, as https://github.com/nixos/nixpkgs/pull/356502#issuecomment-2480637115.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@midchildan
@khaneliman

### DIFF
--- a/modules/launchd/default.nix
+++ b/modules/launchd/default.nix
@@ -43,7 +43,7 @@ let
       };
     };
 
-  toAgent = config: pkgs.writeText "${config.Label}.plist" (toPlist { } config);
+  toAgent = config: pkgs.writeText "${config.Label}.plist" (toPlist { escape = true; } config);
 
   agentPlists = lib.mapAttrs' (n: v: lib.nameValuePair "${v.config.Label}.plist" (toAgent v.config)) (
     lib.filterAttrs (n: v: v.enable) cfg.agents
--- a/modules/launchd/launchd.nix
+++ b/modules/launchd/launchd.nix
@@ -155,6 +155,18 @@ in
         This key maps to the second argument of `execvp(3)`.  This key is required in the absence of the Program
         key. Please note: many people are confused by this key. Please read `execvp(3)` very carefully!
       '';
+      # TODO: Remove this some time after 25.01.
+      apply =
+        value:
+        if value != null then
+          map (
+            item:
+            lib.warnIf (lib.hasInfix "&amp;" item)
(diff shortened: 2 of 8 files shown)


khaneliman:
So the "breaking" change is that someone might have manually escaped themselves before the escaping fix was introduced? Not sure how to protect against that. So, I think it LGTM.

midchildan (approved):

linnnus:
Actually we've been discussing [a similar change](https://github.com/nix-darwin/nix-darwin/pull/1529) in [the nix-darwin development Matrix server](https://app.element.io/#/room/#nix-darwin-dev:nixos.org) and it's clear that there *are* actually at least [some](https://github.com/gustavomedeiross/nix-machina/blob/d159c11a8ccd5db43be3d8471abea6768aa44482/macos/home-manager.nix#L50) [people](https://github.com/search?q=language%3Anix%20%26amp%3B%26amp%3B&type=code) relying on the previous behavior.

We're still discussing solutions in there, so maybe let's just hold off on merging just jet :^)

khaneliman:
starting to see the eval warnings, was there any progress made on this in nixpkgs/nix-darwin ?

linnnus:
We [ended up deciding](https://matrix.to/#/!MMDPZwBJsBXZCsLkuj:matrix.org/$iGsiAQpTDgfewpFSB7En60gwnchSRV3pQHW_4cnCKz0?via=nixos.org&via=matrix.org&via=catgirl.cloud) that a complicated solution wasn't worth it compared to the relatively minor breakage. The most common breakage was people doing this:

```nix
ProgramArguments = [
  "/bin/sh"
  "-c"
  "wait4path /nix/store &amp;&amp; /nix/store/..."
];
```

So I added a warning to that option specifically. I have updated this PR to do the same.

Apologies for the late answer!

khaneliman:
can you rebase and squash

khaneliman:
looks like it needs a `nix fmt`

linnnus:
Should be done now.

khaneliman:
Looks like an eval error: https://buildbot.nix-community.org/#/builders/6063/builds/179/steps/1/logs/nix_error

khaneliman:
Buildbot darwin runners seem stuck today, gonna run test suite locally on my macbook before merging.

PR #7373, opened 2025-07-02
Fix firefox extensions.settings not creating user.js

hey2022:
Creates the user.js if profile.extensions.settings is set. This fixes the case when profile.extensions.settings is set but every other option in the mkIf is false, resulting in no user.js creation.

I also added a user.js test for the firefox-profiles-extensions test.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@rycee @brckd @HPsaucii


bricked-dev:
Thanks for the fix! This seems to have been overlooked in https://github.com/nix-community/home-manager/commit/27ffa35178bc6f359293a8809d32c5da23aaabc7.

brckd (approved):

HPsaucii (approved):
Looks good to me, thanks 👍

PR #7374, opened 2025-07-02
ci: use GITHUB_TOKEN when app config missing

khaneliman:
Allow testing workflows more on forks when app configuration missing.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


.github/workflows/backport.yml
```diff
@@ -23,21 +22,20 @@ jobs:
...
+        if: vars.CI_APP_ID
         id: app-token
         with:
           app-id: ${{ vars.CI_APP_ID }}
           private-key: ${{ secrets.CI_APP_PRIVATE_KEY }}
-
       - uses: actions/checkout@v4
         with:
-          token: ${{ steps.app-token.outputs.token }}
+          token: ${{ steps.app-token.outputs.token || secrets.GITHUB_TOKEN }}
```
MattSturgeon, line 31: Maybe this is just me being weird, but I usually prefer using `github.token` over `secrets.GITHUB_TOKEN`

.github/workflows/backport.yml
```diff
@@ -1,20 +1,19 @@
...
   backport:
     name: Backport Pull Request
     if: >
-      vars.CI_APP_ID
-      && github.event.pull_request.merged == true
-      && (
+      vars.CI_APP_ID && github.event.pull_request.merged == true && (
+
+
+
```
MattSturgeon: This seems to be accidentally formatted?

a) the conditions were joined into a single line
b) there are now _three_ blank lines :eyes:

.github/workflows/backport.yml
khaneliman: Ah, just have always done `secrets.GITHUB_TOKEN` in all my stuff.

.github/workflows/backport.yml
khaneliman: Yeah, my format on save accidentally.. keeps adding blank lines on save... lol

.github/workflows/backport.yml
MattSturgeon: I honestly don't know why both exist; AFAICT they both do the exact same thing. I don't think there's any objective difference between them.

I like the `github.token` name because it demonstrates this is a "special" token generated from the workflow run context, rather than defined within the repo's secrets.

khaneliman: Looking at their documentation, they are functionally identical except `secrets` context limits scope of when it can be accessed to avoid leaking the token when using the `github` context. 

https://docs.github.com/en/actions/how-tos/security-for-github-actions/security-guides/automatic-token-authentication
>An action can access the GITHUB_TOKEN through the github.token context even if the workflow does not explicitly pass the GITHUB_TOKEN to the action. As a good security practice, you should always make sure that actions only have the minimum access they require by limiting the permissions granted to the GITHUB_TOKEN. For more information, see [Permissions for the GITHUB_TOKEN](https://docs.github.com/en/actions/how-tos/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token).

MattSturgeon (approved):

.github/workflows/backport.yml
MattSturgeon: Ah, that's another reason to prefer `github.token` then; it'll also work when used from reusable actions/workflows.

khaneliman: I think that same logic applies to `github.token`

>A token to authenticate on behalf of the GitHub App installed on your repository. This is functionally equivalent to the GITHUB_TOKEN secret. For more information, see Automatic token authentication.
Note: This context property is set by the Actions runner, and is only available within the execution steps of a job. Otherwise, the value of this property will be null.

MattSturgeon: > An action can access the GITHUB_TOKEN through the github.token context even if the workflow does not explicitly pass the GITHUB_TOKEN to the action

I.e. an action can access `github.token` but not `secrets.GITHUB_TOKEN`.

khaneliman: Oh, I guess it's how you look at it. I thought of that as a security mechanism that means you dont accidentally provide a token, but maybe its just more of an inconvenience.

MattSturgeon: From a security perspective the note is saying "even though normally secrets must be passed to actions explicitly, this does not really apply to `GITHUB_TOKEN` because even though `secrets.GITHUB_TOKEN` is not accessible, the same token _is_ accessible at `github.token`"

It's just worded really badly!

PR #7379, opened 2025-07-03
ci: generate-all-maintainers use nix eval update; update list

khaneliman:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```


zowoq:
Looks like this dropped a few people as maintainers as the bot started removing people after it was merged, e.g. `h7x4`.

khaneliman:
> Looks like this dropped a few people as maintainers as the bot started removing people after it was merged, e.g. `h7x4`.

Hmm thought it added more people. I’ll take a look again. Thanks for letting me know

EDIT: looks like some modules had `meta.maintainers` in a conditional `config` section instead of at the top level of the module definition... lol 

I believe the other issue, was some that were in our list but not assigned to modules. The new approach was only adding users who were assigned to a module (since they would need to be pinged). I might just include the entire `HM` list though again.

zowoq:
I went back and checked who was added/removed to make sure the rfc39 tool was behaving, they all seem to match up correctly except for one that was dropped and not added back (`kira-bruneau`) but that looks like that might be due to how the firefox modules are imported.

PR #7385, opened 2025-07-04
programs.msmtp: merge extraConfig and extraAccount into configContent

teto:
We can simplify maintainance of HM while providing more consistency across modules by relying on the nixpkgs module primitive mkBefore / mkOrder as it was done in the Zsh module.

Deprecates:
- programs.msmtp.extraConfig
- programs.msmtp.extraAccounts

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/msmtp/default.nix
```diff
@@ -66,9 +55,36 @@ in
 
       package = lib.mkPackageOption pkgs "msmtp" { };
 
+      configContent = mkOption {
```
teto, line 58: open to suggestions. `settings`  is already used for structured config, `extraConfig` for text.

Would be nice to be able to have a consistent name across modules like we have for `extraConfig`.
So maybe use `initContent` like in the zshModule

modules/programs/msmtp/default.nix
```diff
@@ -98,7 +122,12 @@ in
   config = lib.mkIf cfg.enable {
     home.packages = [ cfg.package ];
 
-    xdg.configFile."msmtp/config".text = configFile msmtpAccounts;
+    xdg.configFile."msmtp/config".text = lib.mkMerge [
+      "# Generated by Home Manager."
+      (lib.mkIf (cfg.extraConfig != "") cfg.extraConfig)
+      (lib.concatStringsSep "\n\n" (map accountStr msmtpAccounts))
+      (lib.mkIf (cfg.extraAccounts != "") cfg.extraAccounts)
+    ];
```
khaneliman: This should move the `lib.mkMerge` to be applied to the `configContent` and then set the `xdg.configFile."msmtp/config".text` = cfg.configContent` similar to zsh otherwise the new option isn't being used. 


```suggestion
    xdg.configFile."msmtp/config".text = cfg.configContent;
 
    programs.msmtp.cfgContent = lib.mkMerge [
      "# Generated by Home Manager."
      (lib.mkIf (cfg.extraConfig != "") cfg.extraConfig)
      (lib.concatStringsSep "\n\n" (map accountStr msmtpAccounts))
      (lib.mkIf (cfg.extraAccounts != "") cfg.extraAccounts)
    ];
```

modules/programs/msmtp/default.nix
```diff
@@ -66,9 +55,36 @@ in
...
+      };
+
       extraConfig = mkOption {
         type = types.lines;
         default = "";
+        visible = false;
+        apply =
+          x:
+          lib.warnIfNot (x == "") ''
+            `programs.msmtp.extraAccounts` is deprecated, use `programs.msmtp.configContent` instead.
```
khaneliman:
```suggestion
            `programs.msmtp.extraConfig` is deprecated, use `programs.msmtp.configContent` instead.
```

modules/programs/msmtp/default.nix
khaneliman: I'm having trouble thinking of a better name, atm. `configContent` makes sense to me. `initContent` feels like it would imply you will have multiple `*Content` options for different phases.

modules/programs/msmtp/default.nix
teto: you are like a hawk, nothing escapes you ;) Thanks a lot for the good reviews !

modules/programs/msmtp/default.nix
teto: I had initally not seen your suggestion and ended up writing it sry  and ty. Should be better now.

teto:
if we are ok with `configContent`, it should be good.

khaneliman (approved):

PR #7388, opened 2025-07-05
codex: starting with v0.2.0 codex uses a TOML configuration file

delafthi:
Updated related settings and test files to reflect the new format.

Depends on https://github.com/NixOS/nixpkgs/pull/418238.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


khaneliman:
Would benefit from an approach similar to https://github.com/nix-community/home-manager/pull/7334 with config generation gating on package version. Then, it wouldn't need to wait for the package to be merged into our flake lock and is backwards compatible.

delafthi:
> Would benefit from an approach similar to #7334 with config generation gating on package version. Then, it wouldn't need to wait for the package to be merged into our flake lock and is backwards compatible.

Good idea. I'll do that.

delafthi:
In the latest force-push, I have implemented backward compatibility as discussed in issue #7334. Additionally, I introduced guards to ensure configuration files are generated only when the configuration is not empty. Corresponding tests have also been added to verify this behavior.

Regarding the configuration, I noticed that `provider` was renamed to `model_provider` and `providers` to `model_providers`. I did not implement a conversion function like #7334, as I believe it is easier for users to manually migrate their configuration to the new version.

delafthi:
https://github.com/NixOS/nixpkgs/pull/418238 has been merged.

Issue #6852, opened 2025-04-20
bug: bookmarks don't work in LibreWolf


victorbnl:
### Are you following the right branch?

- [x] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Issue description

LibreWolf ignores bookmarks set through the `programs.librewolf.profiles.<name>.bookmarks` option. This has been discussed in https://github.com/nix-community/home-manager/pull/5684, and the option has been hidden in https://github.com/nix-community/home-manager/pull/5684/commits/0bd5e9c76c9f2f486f4bec0524e16026547df414 to discourage people from using it.

I would like to try to identify what causes that issue, and see if we can work on fixing it instead of hiding the option.

### Maintainer CC

@chayleaf 
@onny 

### System information

```markdown
Irrelevant
```

victorbnl:
As mentioned in [codeberg:librewolf/librewolf#2378](https://codeberg.org/librewolf/issues/issues/2378), setting the `NoDefaultBookmarks` policy to `false` might solve the issue. However this is currently impossible due to #6853.

JeppeX:
Now that the policies can be modified, should setting some bookmarks implicitly set `NoDefaultBookmarks` to false?

marie-bnl:
I think it should, if you set bookmarks you most likely want them to be set in the browser and I can't see a case where you'd not want to set the policy, despite having set bookmarks.

Issue #5821, opened 2024-09-10
bug: firefox bookmarks are not added when policies.NoDefaultBookmarks is true


s1n7ax:
### Are you following the right branch?

- [X] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Issue description

When using following config, bookmarks are added but when `NoDefaultBookmarks = true` `homepage` bookmark is not added.

```nix
{ ... }:
{
  programs.firefox = {
    enable = true;

    profiles.s1n7ax = {
      isDefault = true;
      bookmarks = [
        {
          name = "Nix sites";
          toolbar = true;
          bookmarks = [
            {
              name = "homepage";
              url = "https://nixos.org/";
            }
          ];
        }
      ];
    };

    policies = {
      NoDefaultBookmarks = false;
    };
  };
}
```

### Maintainer CC

@rycee @kira-bruneau @bricked

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.6.49, NixOS, 24.11 (Vicuna), 24.11.20240906.574d1ea`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.18.5`
 - channels(s1n7ax): `"home-manager, nixos-23.11"`
 - channels(root): `"nixos-23.11"`
 - nixpkgs: `/nix/store/5w3dp0m37794iffsbm9vd9f1xmmhda6i-source`
```

brckd:
Good point! This policy should be set to true if bookmarks are added.

jaredmontoya:
not stale

brckd:
I was unable to override this policy with Home Manager on LibreWolf.

PR #7399, opened 2025-07-06
mkFirefoxModule: set NoDefaultBookmarks when bookmarks are enabled; librewolf: allow bookmark configuration

khaneliman:
### Description
Closes https://github.com/nix-community/home-manager/issues/5821
Closes https://github.com/nix-community/home-manager/issues/6852

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/firefox/mkFirefoxModule.nix
+++ b/modules/programs/firefox/mkFirefoxModule.nix
@@ -964,6 +964,9 @@ in
       release = mkOptionDefault (builtins.head (lib.splitString "-" cfg.package.version));
 
       policies = {
+        NoDefaultBookmarks = lib.mkIf (builtins.any (profile: profile.bookmarks.enable) (
+          builtins.attrValues cfg.profiles
+        )) false;
         ExtensionSettings = lib.mkIf (cfg.languagePacks != [ ]) (
           lib.listToAttrs (
             map (
--- a/modules/programs/librewolf.nix
+++ b/modules/programs/librewolf.nix
@@ -42,8 +42,6 @@ in
       platforms.darwin = {
         configPath = "Library/Application Support/LibreWolf";
       };
-
-      enableBookmarks = false;
     })
   ];
 
--- a/tests/modules/programs/firefox/profiles/bookmarks/default.nix
+++ b/tests/modules/programs/firefox/profiles/bookmarks/default.nix
@@ -2,6 +2,7 @@ modulePath:
 {
   config,
   lib,
+  pkgs,
   ...
 }:
 
@@ -11,12 +12,17 @@ let
 
   firefoxMockOverlay = import ../../setup-firefox-mock-overlay.nix modulePath;
 
+  darwinPath = "Applications/${cfg.darwinAppName}.app/Contents/Resources";
 in
 {
   imports = [ firefoxMockOverlay ];
 
   config = lib.mkIf config.test.enableBig (
-    lib.setAttrByPath modulePath {
+    {
+      # Required for bookmark policy to get set
+      home.stateVersion = "19.09";
+    }
+    // lib.setAttrByPath modulePath {
       enable = true;
       profiles.bookmarks = {
         settings = {
@@ -80,22 +86,39 @@ in
       };
     }
     // {
-      nmt.script = ''
-        bookmarksUserJs=$(normalizeStorePaths \
-          "home-files/${cfg.profilesPath}/bookmarks/user.js")
+      nmt.script =
+        let
+          libDir =
+            if pkgs.stdenv.hostPlatform.isDarwin then
+              "${cfg.finalPackage}/${darwinPath}"
+            else
+              "${cfg.finalPackage}/lib/${cfg.wrappedPackageName}";
+          config_file = "${libDir}/distribution/policies.json";
+        in
+        ''
+          assertFileExists "${config_file}"
 
-        assertFileContent \
-          $bookmarksUserJs \
-          ${./expected-bookmarks-user.js}
+          noDefaultBookmarks_actual_value="$(${lib.getExe pkgs.jq} ".policies.NoDefaultBookmarks" ${config_file})"
 
-        bookmarksFile="$(sed -n \
-          '/browser.bookmarks.file/ {s|^.*\(/nix/store[^"]*\).*|\1|;p}' \
-          "$TESTED/home-files/${cfg.profilesPath}/bookmarks/user.js")"
+          if [[ $noDefaultBookmarks_actual_value != "false" ]]; then
+            fail "Expected '${config_file}' to set 'policies.NoDefaultBookmarks' to false"
+          fi
 
-        assertFileContent \
-          $bookmarksFile \
-          ${./expected-bookmarks-list.html}
-      '';
+          bookmarksUserJs=$(normalizeStorePaths \
+            "home-files/${cfg.profilesPath}/bookmarks/user.js")
+
+          assertFileContent \
+            $bookmarksUserJs \
+            ${./expected-bookmarks-user.js}
+
+          bookmarksFile="$(sed -n \
+            '/browser.bookmarks.file/ {s|^.*\(/nix/store[^"]*\).*|\1|;p}' \
+            "$TESTED/home-files/${cfg.profilesPath}/bookmarks/user.js")"
+
+          assertFileContent \
+            $bookmarksFile \
+            ${./expected-bookmarks-list.html}
+        '';
     }
   );
 }


brckd:
```
┃        error: attribute 'bookmarks' missing
┃        at /nix/store/lnsjlpj4zyzdmwdmkiyy8w2blh45a1my-source/modules/programs/firefox/mkFirefoxModule.nix:967:39:
┃           966|       policies = {
┃           967|         NoDefaultBookmarks = lib.mkIf cfg.bookmarks.enable false;
```

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -964,6 +964,7 @@ in
       release = mkOptionDefault (builtins.head (lib.splitString "-" cfg.package.version));
 
       policies = {
+        NoDefaultBookmarks = lib.mkIf cfg.bookmarks.enable false;
```
bricked-dev: This results in the following error:
```
┃        error: attribute 'bookmarks' missing
┃        at /nix/store/lnsjlpj4zyzdmwdmkiyy8w2blh45a1my-source/modules/programs/firefox/mkFirefoxModule.nix:967:39:
┃           966|       policies = {
┃           967|         NoDefaultBookmarks = lib.mkIf cfg.bookmarks.enable false;
┃              |                                       ^
┃           968|         ExtensionSettings = lib.mkIf (cfg.languagePacks != [ ]) (
```

brckd (changes requested):

modules/programs/firefox/mkFirefoxModule.nix
bricked-dev: `bookmarks` is defined in `profiles.<profile>.bookmarks`

khaneliman: Oops it's under the profiles, isn't it.

bricked-dev: Try something like this:
```suggestion
        NoDefaultBookmarks = lib.mkIf (builtins.any (profile: profile.bookmarks.enable) cfg.profiles) false;
```

khaneliman: So just check if any profile has bookmarks enabled ?

brckd (changes requested):

modules/programs/firefox/mkFirefoxModule.nix
khaneliman: lol you're fast. by the time i responded you had already posted :P

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -964,6 +964,7 @@ in
       release = mkOptionDefault (builtins.head (lib.splitString "-" cfg.package.version));
 
       policies = {
+        NoDefaultBookmarks = lib.mkIf (builtins.any (profile: profile.bookmarks.enable) cfg.profiles) false;
```
bricked-dev: Sorry, it's an attribute set
```suggestion
        NoDefaultBookmarks = lib.mkIf (builtins.any (profile: profile.bookmarks.enable) (builtins.attrValues cfg.profiles)) false;
```

khaneliman:
Trying to add to the bookmarks test to verify it adds it to `policies.json` but not seeing it get populated for some reason.

bricked-dev:
Can confirm this works!
image

bricked-dev:
> Trying to add to the bookmarks test to verify it adds it to `policies.json` but not seeing it get populated for some reason.

How did you try to add the test?

khaneliman:
> > Trying to add to the bookmarks test to verify it adds it to `policies.json` but not seeing it get populated for some reason.
> 
> How did you try to add the test?

diff --git a/tests/modules/programs/firefox/profiles/bookmarks/default.nix b/tests/modules/programs/firefox/profiles/bookmarks/default.nix
index 4fd5f88cc..628178717 100644
--- a/tests/modules/programs/firefox/profiles/bookmarks/default.nix
+++ b/tests/modules/programs/firefox/profiles/bookmarks/default.nix
@@ -2,6 +2,7 @@ modulePath:
 {
   config,
   lib,
+  pkgs,
   ...
 }:
 
@@ -11,6 +12,7 @@ let
 
   firefoxMockOverlay = import ../../setup-firefox-mock-overlay.nix modulePath;
 
+  darwinPath = "Applications/${cfg.darwinAppName}.app/Contents/Resources";
 in
 {
   imports = [ firefoxMockOverlay ];
@@ -80,22 +82,41 @@ in
       };
     }
     // {
-      nmt.script = ''
-        bookmarksUserJs=$(normalizeStorePaths \
-          "home-files/${cfg.profilesPath}/bookmarks/user.js")
+      nmt.script =
+        let
+          libDir =
+            if pkgs.stdenv.hostPlatform.isDarwin then
+              "${cfg.finalPackage}/${darwinPath}"
+            else
+              "${cfg.finalPackage}/lib/${cfg.wrappedPackageName}";
+          config_file = "${libDir}/distribution/policies.json";
+        in
+        ''
+          jq=${lib.getExe pkgs.jq}
 
-        assertFileContent \
-          $bookmarksUserJs \
-          ${./expected-bookmarks-user.js}
+          assertFileExists "${config_file}"
 
-        bookmarksFile="$(sed -n \
-          '/browser.bookmarks.file/ {s|^.*\(/nix/store[^"]*\).*|\1|;p}' \
-          "$TESTED/home-files/${cfg.profilesPath}/bookmarks/user.js")"
+          noDefaultBookmarks_actual_value="$($jq ".policies.NoDefaultBookmarks" ${config_file})"
 
-        assertFileContent \
-          $bookmarksFile \
-          ${./expected-bookmarks-list.html}
-      '';
+          # if [[ $noDefaultBookmarks_actual_value != "false" ]]; then
+          #   fail "Expected '${config_file}' to set 'policies.NoDefaultBookmarks' to false"
+          # fi
+
+          bookmarksUserJs=$(normalizeStorePaths \
+            "home-files/${cfg.profilesPath}/bookmarks/user.js")
+
+          assertFileContent \
+            $bookmarksUserJs \
+            ${./expected-bookmarks-user.js}
+
+          bookmarksFile="$(sed -n \
+            '/browser.bookmarks.file/ {s|^.*\(/nix/store[^"]*\).*|\1|;p}' \
+            "$TESTED/home-files/${cfg.profilesPath}/bookmarks/user.js")"
+
+          assertFileContent \
+            $bookmarksFile \
+            ${./expected-bookmarks-list.html}
+        '';
     }
   );
 }

khaneliman:
> > Trying to add to the bookmarks test to verify it adds it to `policies.json` but not seeing it get populated for some reason.
> 
> How did you try to add the test?

```diff
diff --git a/tests/modules/programs/firefox/profiles/bookmarks/default.nix b/tests/modules/programs/firefox/profiles/bookmarks/default.nix
index 4fd5f88cc..628178717 100644
--- a/tests/modules/programs/firefox/profiles/bookmarks/default.nix
+++ b/tests/modules/programs/firefox/profiles/bookmarks/default.nix
@@ -2,6 +2,7 @@ modulePath:
 {
   config,
   lib,
+  pkgs,
   ...
 }:
 
@@ -11,6 +12,7 @@ let
 
   firefoxMockOverlay = import ../../setup-firefox-mock-overlay.nix modulePath;
 
+  darwinPath = "Applications/${cfg.darwinAppName}.app/Contents/Resources";
 in
 {
   imports = [ firefoxMockOverlay ];
@@ -80,22 +82,41 @@ in
       };
     }
     // {
-      nmt.script = ''
-        bookmarksUserJs=$(normalizeStorePaths \
-          "home-files/${cfg.profilesPath}/bookmarks/user.js")
+      nmt.script =
+        let
+          libDir =
+            if pkgs.stdenv.hostPlatform.isDarwin then
+              "${cfg.finalPackage}/${darwinPath}"
+            else
+              "${cfg.finalPackage}/lib/${cfg.wrappedPackageName}";
+          config_file = "${libDir}/distribution/policies.json";
+        in
+        ''
+          jq=${lib.getExe pkgs.jq}
 
-        assertFileContent \
-          $bookmarksUserJs \
-          ${./expected-bookmarks-user.js}
+          assertFileExists "${config_file}"
 
-        bookmarksFile="$(sed -n \
-          '/browser.bookmarks.file/ {s|^.*\(/nix/store[^"]*\).*|\1|;p}' \
-          "$TESTED/home-files/${cfg.profilesPath}/bookmarks/user.js")"
+          noDefaultBookmarks_actual_value="$($jq ".policies.NoDefaultBookmarks" ${config_file})"
 
-        assertFileContent \
-          $bookmarksFile \
-          ${./expected-bookmarks-list.html}
-      '';
+          # if [[ $noDefaultBookmarks_actual_value != "false" ]]; then
+          #   fail "Expected '${config_file}' to set 'policies.NoDefaultBookmarks' to false"
+          # fi
+
+          bookmarksUserJs=$(normalizeStorePaths \
+            "home-files/${cfg.profilesPath}/bookmarks/user.js")
+
+          assertFileContent \
+            $bookmarksUserJs \
+            ${./expected-bookmarks-user.js}
+
+          bookmarksFile="$(sed -n \
+            '/browser.bookmarks.file/ {s|^.*\(/nix/store[^"]*\).*|\1|;p}' \
+            "$TESTED/home-files/${cfg.profilesPath}/bookmarks/user.js")"
+
+          assertFileContent \
+            $bookmarksFile \
+            ${./expected-bookmarks-list.html}
+        '';
     }
   );
 }
```

bricked-dev:
Looks good to me, but make sure to uncomment these lines and only add them for Librewolf. What do you mean by it didn't get populated?

khaneliman:
> Looks good to me, but make sure to uncomment these lines and only add them for Librewolf. 

The change doesn't distinguish which module is calling it. So the test wouldn't either. Is this only supposed to be applied for `librewolf` in general because https://github.com/nix-community/home-manager/issues/5821 seemed to imply firefox. 

>What do you mean by it didn't get populated?

The test will fail because `policies.json` doesn't contain the `NoDefaultBookmarks` configuration.

bricked-dev:
> > Looks good to me, but make sure to uncomment these lines and only add them for Librewolf.
> 
> The change doesn't distinguish which module is calling it. So the test wouldn't either. Is this only supposed to be applied for `librewolf` in general because #5821 seemed to imply firefox.

You're right, this isn't just for Librewolf.
> 
> > What do you mean by it didn't get populated?
> 
> The test will fail because `policies.json` doesn't contain the `NoDefaultBookmarks` configuration.

Ok thanks for clarifying!

```bash
noDefaultBookmarks_actual_value="$($jq ".policies.NoDefaultBookmarks" ${config_file})"
```
$jq isn't a command

khaneliman:
I found the issue with this test... we aren't explicitly setting `home.stateVersion = "23.05";` like the policies test, when I do the policy gets added... 

EDIT: With that said, I'm not sure why that makes a difference, yet. :P

khaneliman:
> ```shell
> noDefaultBookmarks_actual_value="$($jq ".policies.NoDefaultBookmarks" ${config_file})"
> ```
> 
> $jq isn't a command

It's set above it in the script.

khaneliman:
Ah, I see why. Our default in the test suite is [home.stateVersion = "18.09"](https://github.com/nix-community/home-manager/blob/b4486ff44addd453a64fd8c176ab2fd7ad3f6eb3/tests/default.nix#L132) and the `policies` don't actually support anything before [19.09](https://github.com/nix-community/home-manager/blob/b4486ff44addd453a64fd8c176ab2fd7ad3f6eb3/modules/programs/firefox/mkFirefoxModule.nix#L45)

aguirre-matteo:
For anyone wondering how to put the bookmarks in the toolbar.  TLDR: you can't.

Firefox doesn't allow you to import bookmarks directly to the toolbar, so the only thing you can do is to go to "Manage Bookmarks" (Ctrl+Shift+O) and move the bookmarks from "Bookmarks Menu" to "Bookmarks Toolbar". It's also important to note that, every time you close the browser, the bookmarks from the toolbar will get deleted. To be honest, this is a bit sad, but there's nothing we can do since this problem should be addressed upstream (i.e. add an option to import the bookmarks to the toolbar, or to add a directive in the HTML file that let's you set where bookmarks should go).

EDIT: A workaround could be to make an extension that copies/moves the contents of "Bookmarks Menu" to "Bookmarks Toolbar", but I haven't found an extension that does so.

PR #7402, opened 2025-07-06
firefox: add extension permissions

### LINKED ISSUE #7001: firefox: Restrict extension permissions
### Description

[rycee](https://gitlab.com/rycee/nur-expressions#permission-allow-lists) has a snippet that allows you to restrict an extension's allowed permissions:

```nix
let extensions = {
  "{45d4d1a3-4faa-42b7-9747-bcf2153310cd}" = {
    name = "boring-rss";
    permissions = [
      "activeTab"
    ];
  };
} in {
  # truncated other config

  assertions = lib.mapAttrsToList (k: v: let
    unaccepted = lib.subtractLists
        v.permissions
        config.nur.repos.rycee.firefox-addons.${v.name}.meta.mozPermissions;
    in {
      assertion = unaccepted == [];
      message = ''
        Extension ${v.name} has unaccepted permissions: ${builtins.toJSON unaccepted}'';
    }) extensions;
}
```

We should have this logic built-in into home-manager to provide a safer way to manage extensions.

brckd:
Adds extension permissions as suggested in #7001. Adds the 'profiles.<name>.extensions.settings.<name>.permissions' to Firefox derivatives. If set, this option adds an assertion that fails if an extension package requests permissions that weren't added to the permissions option. In order to not require 'profiles.<name>.extensions.force' to be set when only permissions, but no extension settings were defined, the relevant assertions were changed. They now check whether any 'extensions.settings.<name>.settings' was set instead of checking whether 'extensions.settings' was set. Closes #7001.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@bricked @rycee @HPsaucii

### DIFF
--- a/modules/programs/firefox/mkFirefoxModule.nix
+++ b/modules/programs/firefox/mkFirefoxModule.nix
@@ -729,7 +729,11 @@ in
                                     default = null;
                                     example = [ "activeTab" ];
                                     defaultText = "Any permissions";
-                                    description = "Allowed permissions for this extension.";
+                                    description = ''
+                                      Allowed permissions for this extension. See
+                                      https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/permissions
+                                      for a list of relevant permissions.
+                                    '';
                                   };
                                   force = mkOption {
                                     type = types.bool;
--- a/modules/programs/firefox/mkFirefoxModule.nix
+++ b/modules/programs/firefox/mkFirefoxModule.nix
@@ -731,7 +731,7 @@ in
                                     defaultText = "Any permissions";
                                     description = ''
                                       Allowed permissions for this extension. See
-                                      https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/permissions
+                                      <https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/permissions>
                                       for a list of relevant permissions.
                                     '';
                                   };
--- a/modules/programs/firefox/mkFirefoxModule.nix
+++ b/modules/programs/firefox/mkFirefoxModule.nix
@@ -356,6 +356,12 @@ in
       type = types.attrsOf (
         types.submodule (
           { config, name, ... }:
+          let
+            profilePath = modulePath ++ [
+              "profiles"
+              name
+            ];
+          in
           {
             imports = [ (pkgs.path + "/nixos/modules/misc/assertions.nix") ];
 
@@ -783,28 +789,9 @@ in
                       builtins.all (ext: ext.settings == { }) (attrValues config.extensions.settings)
                       || config.extensions.force;
                     message = ''
-                      Using '${
-                        lib.showAttrPath (
-                          modulePath
-                          ++ [
-                            "profiles"
-                            config.name
-                            "extensions"
-                            "settings"
-                          ]
-                        )
-                      }' will override all previous extensions settings.
-                      Enable '${
-                        lib.showAttrPath (
-                          modulePath
-                          ++ [
-                            "profiles"
-                            config.name
-                            "extensions"
-                            "force"
-                          ]
-                        )
-                      }' to acknowledge this.
+                      Using '${lib.showOption profilePath}.extensions.settings' will override all
+                      previous extensions settings. Enable
+                      '${lib.showOption profilePath}.extensions.force' to acknowledge this.
                     '';
                   }
                 ]
@@ -820,8 +807,7 @@ in
                       assertion = value.permissions == null || length packages == 1;
                       message = ''
                         Must have exactly one extension with addonId '${name}'
-                        in '${moduleName}.extensions.packages' but found
-                        ${toString (length packages)}.
+                        in '${lib.showOption profilePath}.extensions.packages' but found ${toString (length packages)}.
                       '';
                     }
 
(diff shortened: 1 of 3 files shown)


bricked-dev:
As always, since small changes to the mkFirefoxModule can result in big changes in formatting, I recommend to view the diff without changes to whitespace [here](https://github.com/nix-community/home-manager/pull/7402/files?diff=unified&w=1) or using `git diff --ignore-all-space`.

bricked-dev:
How can I add a test that checks whether an assertion failed due to unauthorized extension permissions?

bricked-dev:
Currently, this searches `extensions.packages` to find the extension whose permissions need to be checked. This can be prone to errors as it relies on the `addonId` which can be of two forms. In the future we might consider moving `extensions.packages` to `extensions.settings.<name>.package`. This way the attribute key becomes irrelevant, since it can be inferred from the package's `addonId`.

bricked-dev:
@rycee are you OK with contributing your Permissions Allow List expression?

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -721,7 +721,15 @@ in
...
-                                    description = "Json formatted options for the specified extensionID";
+                                    default = { };
+                                    description = "Json formatted options for this extension.";
+                                  };
+                                  permissions = mkOption {
+                                    type = types.nullOr (types.listOf types.str);
+                                    default = null;
+                                    example = [ "activeTab" ];
+                                    defaultText = "Any permissions";
+                                    description = "Allowed permissions for this extension.";
```
0xda157: linking to docs or providing list of options or something would be nice

bricked-dev: Good idea! It seems like this might help: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/permissions

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -721,7 +721,19 @@ in
...
+                                    description = "Json formatted options for this extension.";
+                                  };
+                                  permissions = mkOption {
+                                    type = types.nullOr (types.listOf types.str);
+                                    default = null;
+                                    example = [ "activeTab" ];
+                                    defaultText = "Any permissions";
+                                    description = ''
+                                      Allowed permissions for this extension. See
+                                      https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/permissions
```
0xda157:
```suggestion
                                      <https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/permissions>
```

awwpotato (approved):
didn't read much of the config impl but the option LGTM

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -888,82 +941,80 @@ in
...
-                    in
-                    "${extensionsEnvPkg}/share/mozilla/${extensionPath}";
-                  recursive = true;
-                  force = true;
-                };
-              }
-            ]
-            ++
-              # Add extension settings as separate attributes
-              optional (profile.extensions.settings != { }) (
```
bricked-dev, line 956: I removed this check. Can someone double check whether it was needed?

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -88,7 +88,7 @@ let
           "browser.bookmarks.file" = toString bookmarksFile;
           "browser.places.importBookmarksHTML" = true;
         }
-        // lib.optionalAttrs (extensions != { }) {
+        // lib.optionalAttrs (builtins.any (ext: ext.settings != { }) (attrValues extensions)) {
```
bricked-dev: It's important that this and the nex line I highlight remain equivalent, or else we might accidentally override settings.

modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -763,36 +775,77 @@ in
...
-                    }' to acknowledge this.
-                  '';
-                }
-              ] ++ config.bookmarks.assertions;
+              assertions =
+                [
+                  (mkNoDuplicateAssertion config.containers "container")
+                  {
+                    assertion =
+                      builtins.all (ext: ext.settings == { }) (attrValues config.extensions.settings)
```
bricked-dev: This has to be equivalent to the previous line I highlighted.

khaneliman: I think it would be worth adding inline comments so it has the living document about the requirement for safety

modules/programs/firefox/mkFirefoxModule.nix
khaneliman: same as [previous](https://github.com/nix-community/home-manager/pull/7402#discussion_r2192761223)

modules/programs/firefox/mkFirefoxModule.nix
bricked-dev: I moved the checks to a method to ensure they remain equivalent.

khaneliman: Thats even better, thanks

modules/programs/firefox/mkFirefoxModule.nix
khaneliman: I can test later, but if mapAttrsToList has nothing to iterate over it i think it handles it fine and should be okay without it.

khaneliman: Seems to build fine for me and didn't see any changes in my diff when activating config.

khaneliman:
Needs a `nix fmt` still btw

PR #7407, opened 2025-07-07
home-environment: add `home.checks`

### LINKED ISSUE #7390: Provide analog for `system.checks` (of NixOS) in Home Manager
### Description

I'd like to validate some files generated by my Home Manager configuration.  While NixOS provides the option `system.checks` which can be populated with usually `pkgs.runCommand` doing the config check. It seems, however, that Home Manager lacks this capability (or I'm not finding it).

The documentation of `system.checks` is:

> Packages that are added as dependencies of the system’s build, usually for the purpose of validating some part of the configuration.
>
> Unlike system.extraDependencies, these store paths do not become part of the built system configuration.

It is implemented in https://github.com/NixOS/nixpkgs/blob/7a732ed41ca0dd64b4b71b563ab9805a80a7d693/nixos/modules/system/activation/top-level.nix#L388

fabiancholewinski1234:
### Note

I do not fully understand how dependency tracking (build time vs "runtime") works but this implementation copied the nixpkgs mechanism so should be fine *if* `home.activationPackage` is the analogue of `baseSystem` in nixos/modules/system/activation/top-level.nix. (https://github.com/NixOS/nixpkgs/blob/f859c5d293108245d6a364ce679f6f3cf15c1f4c/nixos/modules/system/activation/top-level.nix#L57-L77)

Note for nixpkgs implementation: `system.checks` becomes part of `system.systemBuilderArgs` which is passed as an argument to `baseSystem = pkgs.stdenvNoCC.mkDerivation ...`.

### Description

This should have the same effect `system.checks` has in nixpkgs: adds paths to the build closure without being becoming part of the generated configuration.

This is useful for built-time checks as these should not leave a trace in the built home configuration.

The implementation mirrors nixpkgs: add an unused argument to the home-manager-generation derivation.

Fixes #7390

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@rycee

### DIFF
--- a/modules/home-environment.nix
+++ b/modules/home-environment.nix
@@ -389,6 +389,22 @@ in
       description = ''
         A list of paths that should be included in the home
         closure but generally not visible.
+
+        For built-time checks the `home.checks` option is more appropriate
+        for that purpose as checks should not leave a trace in the built
+        home configuration.
+      '';
+    };
+
+    home.checks = mkOption {
+      type = types.listOf types.package;
+      default = [ ];
+      description = ''
+        Packages that are added as dependencies of the home's build, usually
+        for the purpose of validating some part of the configuration.
+
+        Unlike `home.extraDependencies`, these store paths do not
+        become part of the built home configuration.
       '';
     };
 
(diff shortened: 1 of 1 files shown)


fabiancholewinski1234:
Addendum: This PR might be worth cherry-picking to release-25.05.

khaneliman:
> so should be fine if home.activationPackage is the analogue of baseSystem

Have you tested this functionality to see if it works?

fabiancholewinski1234:
> > so should be fine if home.activationPackage is the analogue of baseSystem
> 
> Have you tested this functionality to see if it works?

Yes I had and it worked.  Tests were done with

```
home.checks = [ (pkgs.runCommandLocal "mytest" { } "true; touch $out") ];
```
and
```
home.checks = [ (pkgs.runCommandLocal "mytest" { } "false; touch $out") ];
```

PR #7408, opened 2025-07-07
firefoxpwa: add module

brckd:
### Description

Adds the `firefoxpwa` module for the native Progressive Web Apps for Firefox application.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/firefoxpwa.nix
```diff
@@ -0,0 +1,193 @@
...
+                description = "Name of the profile.";
+              };
+              sites = lib.mkOption {
+                default = { };
+                description = ''
+                  Attribute set of site options for this profile. The keys of that attribute set
+                  consist of ULIDs. A ULID is made of 16 characters, each of which is one of
+                  '0123456789ABCDEFGHJKMNPQRSTVWXYZ' (Excluding I, L, O and U). See
+                  https://github.com/ulid/spec?tab=readme-ov-file#canonical-string-representation.
+                  ULIDs must be unique across profiles and sites.
```
bricked-dev: Might consider to create a custom string type for ULIDs and add assertions for ULID uniqueness.

bricked-dev:
This is what the example configuration would look like:
image

modules/programs/firefoxpwa.nix
```diff
@@ -0,0 +1,254 @@
...
+
+  sites = lib.concatMapAttrs (_: profile: profile.sites) cfg.profiles;
+
+  mkUlidAssertions =
+    path:
+    lib.concatMap (
+      { name, value }:
+      let
+        length = 26;
+        allowed = "0123456789ABCDEFGHJKMNPQRSTVWXYZ";
```
khaneliman, line 22: Might be nice to create some type helpers for these kinds of things. Just making a note for myself.

modules/programs/firefoxpwa.nix
```diff
@@ -0,0 +1,254 @@
...
+    xdg.desktopEntries = lib.mkMerge (
+      lib.mapAttrsToList (name: site: {
+        "FFPWA-${name}" = lib.mkIf site.desktopEntry.enable {
+          inherit (site.desktopEntry) icon categories;
+          name = site.settings.manifest.name;
+          exec = "firefoxpwa site launch ${name} --protocol %u";
+          terminal = false;
+        };
+      }) sites
+    );
```
khaneliman, line 247: nit: like grouping these more 

```nix
xdg = {
  dataFile...

  desktopEntries...
};
```

modules/programs/firefoxpwa.nix
bricked-dev: There is `lib.types.strMatching` for RegExps, but that can't check after keys.

andresilva:
How do you find the manifest URL? Also looking at the official docs of firefoxpwa it seems that it's also possible to wrap websites that do not provide a manifest.

> If the website is not a valid PWA, the installation action will not appear (unless enabled in the extension settings). However, you can still install such websites using the extension's browser action. This method does not require a web app manifest and can be used for any website.

andresilva:
Also regarding the profile id and site id, should we just generate both randomly?

bricked-dev:
> How do you find the manifest URL? Also looking at the official docs of firefoxpwa it seems that it's also possible to wrap websites that do not provide a manifest.
> 
> > If the website is not a valid PWA, the installation action will not appear (unless enabled in the extension settings). However, you can still install such websites using the extension's browser action. This method does not require a web app manifest and can be used for any website.

I was already wondering about that! When I tried to omit the mainfest URL, it threw an error, so we should find out where that manifest is. The manifest url is useless anyways, because it cannot be used to update the read-only manifest data declared by home manager.

You can find the manifest URL by going to the website, opening the inspector in your browser and searching for "manifest". The first result should be a link whose href is the relative URL of the manifest.

bricked-dev:
> Also regarding the profile id and site id, should we just generate both randomly?

These are ULIDs which are supposed to be randomly generated in order to not clash. Firefoxpwa itself uses ULIDs like "00000000000000000000000000" though, so I think it's fine to just use custom ones. If desired, we could generate these ULIDs in home-manager.

drupol:
Really cool ! I wanted to do that for quite a while with Firefox without much luck! https://github.com/TLATER/nix-webapps/

modules/programs/firefoxpwa.nix
```diff
@@ -0,0 +1,254 @@
...
+      default = { };
+      description = ''
+        Attribute set of profile options. The keys of that attribute set consist of
+        ULIDs. A ULID is made of 16 characters, each of which is one of
+        '0123456789ABCDEFGHJKMNPQRSTVWXYZ' (Excluding I, L, O and U). See
+        <https://github.com/ulid/spec?tab=readme-ov-file#canonical-string-representation>.
+      '';
+      example = lib.literalExpression ''
+        {
+          "0123456789ABCDEFGHJKMNPQRSTVWXYZ".sites."ZYXWVTSRQPNMKJHGFEDCBA9876543210" = {
```
drupol, line 71: This expression's ULID has 32 characters, while it should have only 26.

modules/programs/firefoxpwa.nix
```diff
@@ -0,0 +1,254 @@
...
+            name = "MDN Web Docs";
+            url = "https://developer.mozilla.org/";
+            manifestUrl = "https://developer.mozilla.org/manifest.f42880861b394dd4dc9b.json";
+            desktopEntry.icon = pkgs.fetchurl {
+              url = "https://developer.mozilla.org/favicon-192x192.png";
+              sha256 = "0p8zgf2ba48l2pq1gjcffwzmd9kfmj9qc0v7zpwf2qd54fndifxr";
```
drupol, line 77:
```suggestion
              hash = "...";
```

modules/programs/firefoxpwa.nix
```diff
@@ -0,0 +1,254 @@
...
+    home.packages = lib.mkIf (cfg.package != null) [ cfg.package ];
+
+    xdg.dataFile."firefoxpwa/config.json" = lib.mkIf (cfg.settings != { }) {
+      text = lib.generators.toJSON { } cfg.settings;
+    };
+
+    xdg.desktopEntries = lib.mkMerge (
+      lib.mapAttrsToList (name: site: {
+        "FFPWA-${name}" = lib.mkIf site.desktopEntry.enable {
+          inherit (site.desktopEntry) icon categories;
```
drupol, line 241: We should use `site.desktopEntry` as is so we can pass any parameter we want to it. For example: `class` so we can assign the proper WM class to the window and have a proper icon associated to it.

PR #7409, opened 2025-07-07
hyprshell: add module

### LINKED ISSUE #7290: Options for configuring hyprshell (former hyprswitch)
Description:
hyprswitch originally is rust based window management program, that enables Windows/MacOS/KDE like window switching for hyprland (Super+Tab). it also implements a rofi like application launcher. 

https://github.com/h3rmt/hyprshell

adding a module could streamline the setup and increase the accessability of hyprland. 

Proposal:
Currently, i need to autogenerate the hyprshell config using `hyprshell generate config`, then create a module myself to map all options to nix and then retype the config. A Different approach whould be to link it using some form of static file like `xdg.configFile. after that i add hyprshell run to the exec-once clause of hyprland. this is unintuitive and very messy.
A simple Module that creates a systemd service, generates the config based on system specification and already has defaults pointing to the users gtk config whould be a lot easier.

a module could then be used like this

```nix
{config,...):{
  home-manager.users.<name>.services.hyprshell = {
   enable = true
   gtk = {
     theme = config.gtk.theme.name;
     iconTheme = config.gtk.iconTheme.name;
  }
  windows = {
  ...
  }
  launcher = {
  ...
  }
  extraCss = ''
  '';
)
``` 

a List of Config options may be found here:
https://github.com/H3rmt/hyprshell/blob/hyprshell-release/CONFIGURE.md

Example Implementation

%

aguirre-matteo:
### Description

This pull request adds a module for configuring Hyprshell. Closes #7290.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/hyprshell.nix
+++ b/modules/services/hyprshell.nix
@@ -1,0 +1,95 @@
+{
+  lib,
+  pkgs,
+  config,
+  ...
+}:
+let
+  inherit (lib)
+    types
+    mkIf
+    mkEnableOption
+    mkPackageOption
+    mkOption
+    ;
+
+  cfg = config.services.hyprshell;
+
+  jsonFormat = pkgs.formats.json { };
+in
+{
+  meta.maintainers = with lib.hm.maintainers; [ aguirre-matteo ];
+
+  options.services.hyprshell = {
+    enable = mkEnableOption "hyprshell";
+    package = mkPackageOption pkgs "hyprshell" { nullable = true; };
+
+    settings = mkOption {
+      type = jsonFormat.type;
+      default = { };
+      description = ''
+        Configuration settings for hyprshell. All the avaiblable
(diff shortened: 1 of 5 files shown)


modules/services/hyprshell.nix
```diff
@@ -0,0 +1,95 @@
...
+        type = types.str;
+        default = "";
+        example = "-vv";
+        description = "Arguments to pass to the hyprshell service";
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+    assertions = [
```
khaneliman, line 66: can you add the platform assertion too

aguirre-matteo: Done.

H3rmt:
@jennydaman just noticed that the name of the css file is invalid. it should be `styles.css` instead of `style.css`.

https://github.com/H3rmt/hyprshell/issues/343

Issue #7413, opened 2025-07-08
bug: tests fail on aarch64-linux due to unavailable package


me-and:
### Are you following the right branch?

- [x] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Issue description

Attempting to run the test suite with `nix build --reference-lock-file flake.lock ./tests#test-all`, as described at https://github.com/nix-community/home-manager/blob/master/.github/PULL_REQUEST_TEMPLATE.md, fails on aarch64-linux:

```
error:
       … while calling the 'derivationStrict' builtin
         at <nix/derivation-internal.nix>:37:12:
           36|
           37|   strict = derivationStrict drvAttrs;
             |            ^
           38|

       … while evaluating derivation 'nmt-all-tests'
         whose name attribute is located at /nix/store/cd7fkv2r1nflz758r8bsnqfvkq7f72wj-source/pkgs/stdenv/generic/make-derivation.nix:468:13

       … while evaluating attribute 'buildCommand' of derivation 'nmt-all-tests'
         at /nix/store/cd7fkv2r1nflz758r8bsnqfvkq7f72wj-source/pkgs/build-support/trivial-builders/default.nix:80:17:
           79|         enableParallelBuilding = true;
           80|         inherit buildCommand name;
             |                 ^
           81|         passAsFile = [ "buildCommand" ] ++ (derivationArgs.passAsFile or [ ]);

       … while evaluating the option `home.activation.installPackages.data':

       … while evaluating definitions from `/nix/store/vcwmcnvw5dlrnb8phvdcfwmfmbf6rnsj-source/modules/home-environment.nix':

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: Package ‘libstrangle-unstable-202202022’ in /nix/store/cd7fkv2r1nflz758r8bsnqfvkq7f72wj-source/pkgs/tools/X11/libstrangle/default.nix:55 is not available on the requested hostPlatform:
         hostPlatform.config = "aarch64-unknown-linux-gnu"
         package.meta.platforms = [
           "x86_64-linux"
         ]
         package.meta.badPlatforms = [ ]
       , refusing to evaluate.
```

I'm seeing this building at commit fd9e55f5.

### System information

```markdown
- system: `"aarch64-linux"`
 - host os: `Linux 6.6.51, NixOS, 25.05 (Warbler), 25.05.20250629.b43c397`
 - multi-user?: `yes`
[...]

Issue #7412, opened 2025-07-08
bug: non-flake tests fail with `attribute 'withAddons' missing`


me-and:
### Are you following the right branch?

- [x] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Issue description

Attempting to run the test suite with `nix-shell --pure tests -A run.all`, as described at https://github.com/nix-community/home-manager/blob/master/.github/PULL_REQUEST_TEMPLATE.md, fails:

```
error:
       … while calling the 'derivationStrict' builtin
         at <nix/derivation-internal.nix>:37:12:
           36|
           37|   strict = derivationStrict drvAttrs;
             |            ^
           38|

       … while evaluating derivation 'nmt-run-all-tests'
         whose name attribute is located at /nix/store/j95fcik6rzsydwips4m89dmlvfj9hg9y-source/pkgs/stdenv/generic/make-derivation.nix:480:13

       … while evaluating attribute 'shellHook' of derivation 'nmt-run-all-tests'
         at /nix/store/fm1dwqb08lxr1gk02niyvb1j7v4181c0-source/default.nix:38:41:
           37|   runShellOnlyCommand = name: shellHook:
           38|     pkgs.runCommandLocal name { inherit shellHook; } ''
             |                                         ^
           39|       echo This derivation is only useful when run through nix-shell.

       … while evaluating the option `nmt.result.success':

       … while evaluating definitions from `/nix/store/fm1dwqb08lxr1gk02niyvb1j7v4181c0-source/nmt.nix':

       … while evaluating the option `home.activation.installPackages.data':

       … while evaluating definitions from `/home/adam/vcs/home-manager/modules/home-environment.nix':

       … while evaluating the option `home.packages':

       … while evaluating definitions from `/home/adam/vcs/home-manager/modules/programs/anki':

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: attribute 'withAddons' missing
       at /home/adam/vcs/home-manager/modules/programs/anki/default.nix:313:8:
          312|     home.packages = [
          313|       (cfg.package.withAddons (
             |        ^
          314|         [
```

I'm seeing this building at commit fd9e55f5 on the master branch.

### Maintainer CC

@JuneStepp 

### System information

```markdown
- system: `"aarch64-linux"`
 - host os: `Linux 6.6.51, NixOS, 25.05 (Warbler), 25.05.20250629.b43c397`
[...]

PR #7414, opened 2025-07-08
Improve error when failing to build Neomutt config

me-and:
### Description

If there's an account under `accounts.email.accounts` with `neomutt.enable` set to true but neither `passwordCommand` nor `neomutt.sendMailCommand` set, then building the Neomutt rc file will fail.  Ensure that failure has a useful error message, rather than a confusing type error.

While we're here, make the module code slightly less repetitious by just building the set of email accounts that have Neomutt enabled once, rather than multiple times in multiple contexts.

### Checklist

    or `nix build --reference-lock-file flake.lock ./tests#test-all` using Flakes. (See #7412 for failures with the non-flake test, and #7413 for failures on aarch64-linux, but flake test passes on x86_64-linux.)

- ~[ ] Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).~ No changes required

    ```
    {component}: {description}

    {long description}
    ```

  - ~[ ] Added myself as module maintainer. See [example](https://github.com/nix-community/home-manager/blob/a51598236f23c89e59ee77eb8e0614358b0e896c/modules/programs/lesspipe.nix#L11).~

#### Maintainer CC

No obvious maintainers.

### DIFF
--- a/modules/programs/neomutt/default.nix
+++ b/modules/programs/neomutt/default.nix
@@ -23,18 +23,15 @@ let
 
   accountCommandNeeded = lib.any (
     a:
-    a.neomutt.enable
-    && (
-      a.neomutt.mailboxType == "imap"
-      || (lib.any (m: !isString m && m.type == "imap") a.neomutt.extraMailboxes)
-    )
-  ) (attrValues config.accounts.email.accounts);
+    a.neomutt.mailboxType == "imap"
+    || (lib.any (m: !isString m && m.type == "imap") a.neomutt.extraMailboxes)
+  ) neomuttAccounts;
 
   accountCommand =
     let
       imapAccounts = filter (
-        a: a.neomutt.enable && a.imap.host != null && a.userName != null && a.passwordCommand != null
-      ) (attrValues config.accounts.email.accounts);
+        a: a.imap.host != null && a.userName != null && a.passwordCommand != null
+      ) neomuttAccounts;
       accountCase =
         account:
         let
--- a/modules/programs/neomutt/default.nix
+++ b/modules/programs/neomutt/default.nix
@@ -10,6 +10,7 @@ let
     concatStringsSep
     concatMapStringsSep
     filter
+    filterAttrs
     isString
     mkIf
     mkOption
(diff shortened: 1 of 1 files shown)


modules/programs/neomutt/default.nix
```diff
@@ -538,12 +537,17 @@ in
...
+    assertions =
+      [
+        {
+          assertion = ((filter (b: (lib.length (lib.toList b.map)) == 0) (cfg.binds ++ cfg.macros)) == [ ]);
+          message = "The 'programs.neomutt.(binds|macros).map' list must contain at least one element.";
+        }
+      ]
+      ++ lib.mapAttrsToList (n: a: {
+        assertion = a.neomutt.sendMailCommand != null || a.passwordCommand != null;
+        message = "'accounts.email.accounts.${n}' needs either 'neomutt.sendMailCommand' or 'passwordCommand' set.";
```
teto, line 549: would be nice if message could mention which one precisely (or both).

modules/programs/neomutt/default.nix
```diff
@@ -538,12 +537,17 @@ in
...
-      {
-        assertion = ((filter (b: (lib.length (lib.toList b.map)) == 0) (cfg.binds ++ cfg.macros)) == [ ]);
-        message = "The 'programs.neomutt.(binds|macros).map' list must contain at least one element.";
-      }
-    ];
+    assertions =
+      [
+        {
+          assertion = ((filter (b: (lib.length (lib.toList b.map)) == 0) (cfg.binds ++ cfg.macros)) == [ ]);
+          message = "The 'programs.neomutt.(binds|macros).map' list must contain at least one element.";
```
teto, line 544: unrelated to your PR but I wonder why we need this assert ^^' ?

modules/programs/neomutt/default.nix
me-and: You need to configure at least one of `neomutt.sendMailCommand` or `passwordCommand`. Whether someone wants to configure one, the other, or both, will depend on how they want to configure their system to send emails, so it's a user decision rather than something this message can mandate.

If you have any suggestions about how to make that clearer, I'd be glad to hear them!

modules/programs/neomutt/default.nix
me-and: I have no idea. Looks like it was added by @sumnerevans in 25a6a6d2984e70c9a07c8f8a69ebe24e6c700abf at #1885

PR #7422, opened 2025-07-09
jujutsu: add includes option to declare additional config files

lajp:
### Description

Add an option to declare additional configuration files. This is useful for example for applying conditional configuration with the `--when.condition` option.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@shikanime

### DIFF
--- a/modules/programs/jujutsu.nix
+++ b/modules/programs/jujutsu.nix
@@ -66,6 +66,22 @@ in
         for options.
       '';
     };
+
+    includes = mkOption {
+      type = types.listOf types.lines;
+      default = [ ];
+      example = [
+        ''
+          --when.repositories = ["~/work"]
+
+          [user]
+          email = "NAME@WORK"
+        ''
+      ];
+      description = ''
+        List of additional configuration files to include. Similar to `programs.git.includes`.
+      '';
+    };
   };
 
   config = mkIf cfg.enable {
@@ -92,8 +108,23 @@ in
       })
     ];
 
-    home.file."${configDir}/jj/config.toml" = mkIf (cfg.settings != { }) {
-      source = tomlFormat.generate "jujutsu-config" cfg.settings;
-    };
+    home.file = lib.mkMerge [
+      {
+        "${configDir}/jj/config.toml" = mkIf (cfg.settings != { }) {
+          source = tomlFormat.generate "jujutsu-config" cfg.settings;
+        };
+      }
+
+      (mkIf (cfg.includes != [ ]) (
+        let
+          files = lib.imap (i: conf: {
+            "${configDir}/jj/conf.d/include-${toString i}.toml" = {
+              text = conf;
+            };
+          }) cfg.includes;
+        in
+        lib.mergeAttrsList files
+      ))
+    ];
   };
 }


lajp:
Ideally it would be nice to write the configuration in `includes` as attribute sets but there's a problem in that the condition (`--when`...) has to be on top of the file. I don't really like the idea of splitting it into `{ condidition = {}; config = {}; }` but that certainly is one option.

Any ideas for this?

teto:
hyave you thought of creating the files yourselves with `xdg.configFile."jj/conf.d.` ? jj loads them automatically.

lajp:
That's what I do [currently](https://github.com/lajp/nix/blob/e66e0f9970381a28b37a6ff1455aeac705312edb/modules/user/programs/jujutsu.nix#L31-L48), yes, but I think it would be beneficial to have it as a part of the hm module

ambroisie:
> I don't really like the idea of splitting it into { condidition = {}; config = {}; } but that certainly is one option.

Why not? This mirrors the [`programs.git.includes`](https://github.com/nix-community/home-manager/blob/206ed3c71418b52e176f16f58805c96e84555320/modules/programs/git.nix#L32) options.

EDIT: I don't understand your point about "the condition has to be at the top of the file".

Translating [their example](https://jj-vcs.github.io/jj/latest/config/#using-multiple-files) directly to a Nix attribute set:

```nix
{
  "--when" = {
    repositories = [ "~/the/work/repo" ];
  };

  user = {
    email = "YOUR_WORK_EMAIL@workplace.com";
  };
}
```

You _could_ write a module option which would translate the following:

```nix
  programs.jujutsu.includes = {
    work = {
      condition = {
        repositories = [ "~/the/work/repo" ];
      };

      settings = {
        user = {
          email = "YOUR_WORK_EMAIL@workplace.com";
        };
      };
    };
  };
```

To (`name = "work"`, `value = { condition = <...>; settings = <...>}`)

```
  home.file."${config.xdg.configHome}/jj/conf.d/50-hm-include-${name}.toml" = {
    source = tomlFormat.generate "jujutsu-include-${name}.toml" ({ "--when" = value.condition; } // value.settings);
  };
```

Though to be honest I think letting the user write `"--when"` themselves is fine (i.e: no `condition`/`settings` split, just straight to TOML translation of the value), not worth the complexity in the module to make it _less_ transparent.

lajp:
> EDIT: I don't understand your point about "the condition has to be at the top of the file".

It might've been just a fluke but I think I ran into a problem where due to the attribute set being unordered, the lines ended up out of order in the configuration file, which caused the condition to not work.

But this might've just been a mistake on my part, I'll check later. And if it doesn't hold, then there's no reason to split it into condition and settings and we can just use what you suggested (which is what I initially wrote as well).

ambroisie:
> split it into condition and settings

As I said, I see little reason to even _have_ that split, it saves the user writing `--where` by calling it `condition` instead, and adding nesting to the rest of the settings attrset. It's strictly more work and less transparent.

shikanime:
I have always had mixed feelings about these abstractions, including the Git module. They lure you away from the official tool configuration unnecessarily and don't seem to offer any advantages. Unlike Git, Jujutsu doesn't require the use of external files with the '--scope' option, and I doubt many people examine the configuration files created using Nix.

ambroisie:
> I have always had mixed feelings about these abstractions, including the Git module. They lure you away from the official tool configuration unnecessarily and don't seem to offer any advantages. Unlike Git, Jujutsu doesn't require the use of external files with the '--scope' option, and I doubt many people examine the configuration files created using Nix.

For Git specifically, the include option takes care to include files _at the end_ of the configuration file. For Jujutsu, one could just take care to name the files correctly to be included in the correct order (always _after_ the main configuration). I tend to agree on it being less useful here.

lajp:
Closing this because it didn't seem to be a popular change and I realized that my use case is covered entirely by the `--scope` option already provided by Jujutsu

PR #7424, opened 2025-07-09
hyprland: add direct support for animation configuration

C0Florent:
### Description

This PR adds a specific support for Hyprland animation configuration.

Note it is already possible to configure animations through `programs.services.window-managers.hyprland.settings`, but I recently found a use case which I described in my added option's example in which it would be useful to have a dedicated option.

I'm not sure we want this change or not: it has the advantage of covering my niche use case and provides a more readable, less repetitive configuration option; but has the disadvantage of proposing several ways of achieving the same thing, potentially making it harder to understand for the module users. Whether or not this is considered valuable for HM, I still need something for my use case, so I would just use it as a custom module in my config if it does not land here.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@fufexan

### DIFF
--- a/modules/services/window-managers/hyprland.nix
+++ b/modules/services/window-managers/hyprland.nix
@@ -290,6 +290,43 @@ in
       '';
     };
 
+    animations = lib.mkOption {
+      type = with lib.types; attrsOf str;
+      default = { };
+      description = ''
+        Hyprland animations configuration. This option is an alternative to
+        {option}`programs.hyprland.settings.animation` which can help avoiding
+        repetition of animations, in case you want to pass an animation to
+        a bind to hyprctl for example.
+
+        ::: {.note}
+        This option has an `apply` function which prepends the attribute
+        values with their corresponding names. This is so that to write
+        a call to `hyprctl animation`, you only have to write
+        ```nix
+        "hyprctl animation ''${config.programs.hyprland.animations.''${animName}}"
+        ```
+        instead of having to repeat the animation name (which would look like this:)
+        ```nix
+        "hyprctl animation ''${animName}, ''${config.programs.hyprland.animations.''${animName}}"
+        ```
+        :::
+      '';
+      example = lib.literalExpression ''
+        {
+          ## This produces the example from the Hyprland documentation example:
+          # animation = workspaces, 1, 8, default
+          # animation = windows, 1, 10, myepiccurve, slide
+          # animation = fade, 0
+
+          workspaces = "1, 8, default";
+          windows = "1, 10, myepiccurve, slide";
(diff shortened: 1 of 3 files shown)


fufexan (approved):
Looks fine to me, and the changes are simple enough.

modules/services/window-managers/hyprland.nix
```diff
@@ -216,6 +216,43 @@ in
       '';
     };
 
+    animations = lib.mkOption {
```
khaneliman, line 293: My first gut reaction is not liking introducing a random option when we have a freeform settings option and an extraConfig explicitly for things that don't get generated how you want.

fufexan: Same, I had to think it over for a few hours before approving, but I'm not fully sure we want to implement this in HM. There's https://github.com/hyprland-community/hyprnix/ for all the bells and whistles.

C0Florent: > My first gut reaction is not liking introducing a random option when we have a freeform settings option and an extraConfig explicitly for things that don't get generated how you want.

I do agree this may not be suited for integration into HM, for the reasons you stated above.

Just to make my point clear, let me state why this option in particular has a purpose that cannot be reached by using the current `settings` or `extraConfig`

---

Yesterday when working on my personal config (https://github.com/C0Florent/config/commit/c33b8fd043e339cdd074967d9268c98adcc03722), I found the need to "repeat" an animation which was written in `settings.animation`, which is a **list** along these lines:
```nix
    animation = [
      "windows,    1, 3.5, easeOutBack, slide"
      "windowsOut, 1, 5,   easeOutBack, slide"
    ];
```
In this case, the names `windows` and `windowsOut` are special identifiers recognised by Hyprland to configure particular animations.
Later in configuration (in a completely separate file), I need to reference a specific animation configuration (which in this case is a list element). Lists are not suited to find elements in this way; so in this particular case I wish it was an attribute set rather than a list, so that I could lookup by name instead of by string prefix.

My proposed change adds another option for defining animations, which does allow defining this as an attrset, in the following fashion:
```nix
    animations = {
      windows     = "1, 3.5, easeOutBack, slide";
      windowsOut  = "1, 5,   easeOutBack, slide";
    };
```
This way, I can reference animations in other parts of my config by using `config.wayland.windowManager.hyprland.animations.<name>`, which gives me back the string I wrote in config.

---

Sorry for the repetition of my point, I simply wanted to give you a clear overview of the pros of this change, given you've already understood the cons of it. Now that this point is clear I'm curious of your definitive opinion on the subject

PR #7429, opened 2025-07-11
lib: Add ordered children and no children to `lib.hm.generators.toKDL`

quentinmit:
### Description

The current `toKDL` generator doesn't allow the generation of Zellij keybinding configuration because it requires ordered children. For example, the following KDL:

```kdl
keybinds clear-defaults=true {
    pane {
        bind "e" { TogglePaneEmbedOrFloating; SwitchToMode "locked"; }
        bind "left" { MoveFocus "left"; }
    }
}
```

cannot be generated with the current `toKDL` generator.

There are a few previous attempts to fix this (#5272, #4614, #5706), but they all involve non-backwards-compatible changes to the existing mapping.

This PR extends the existing syntax to add an `_children` attr (like the already-existing `_args` and `_props` attrs) that can be used to pass a list of ordered children for a node, which means that the above KDL can be generated with:

```nix
keybinds._props.clear-defaults = true;
keybinds.pane._children = [
  {
    bind = {
      _args = ["e"];
      _children = [
        { TogglePaneEmbedOrFloating = {}; }
        { SwitchToMode._args = ["locked"]; }
      ];
    };
  }
  {
    bind = {
      _args = ["left"];
      MoveFocus = ["left"];
    };
  }
];
```

It's probably worth adding a higher-level interface to assign keybindings to the zellij module, but I think that should be a separate PR.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@khaneliman @mainrs for the zellij module

### DIFF
--- a/modules/lib/generators.nix
+++ b/modules/lib/generators.nix
@@ -116,34 +116,34 @@
       convertAttrsToKDL =
         name: attrs:
         let
-          optArgsString = lib.optionalString (attrs ? "_args") (
-            lib.pipe attrs._args [
-              (map literalValueToString)
-              (lib.concatStringsSep " ")
-              (s: s + " ")
-            ]
+          optArgs = map literalValueToString (attrs._args or [ ]);
+          optProps = lib.mapAttrsToList (name: value: "${name}=${literalValueToString value}") (
+            attrs._props or { }
           );
 
-          optPropsString = lib.optionalString (attrs ? "_props") (
-            lib.pipe attrs._props [
-              (lib.mapAttrsToList (name: value: "${name}=${literalValueToString value}"))
-              (lib.concatStringsSep " ")
-              (s: s + " ")
-            ]
-          );
+          orderedChildren = lib.pipe (attrs._children or [ ]) [
+            (map (child: mapAttrsToList convertAttributeToKDL child))
+            lib.flatten
+          ];
+          unorderedChildren = lib.pipe attrs [
+            (lib.filterAttrs (
+              name: _:
+              !(elem name [
+                "_args"
+                "_props"
+                "_children"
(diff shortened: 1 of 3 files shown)


khaneliman:
Looking pretty good to me and testing it works well. Would be nice to add some documentation in the module related to this nix -> kdl syntax for people to discover it easier and close out https://github.com/nix-community/home-manager/issues/4659. But, I won't block on that.

terlar:
Nice, this will also aid in the niri configuration.

VAWVAW:
Have I missed something or does this just not work for toplevel children?

```nix
toKDL {} {
  _children = [{
    {
      output = {
        _args = ["a"];
        b = "c";
      };
    }
  }];
};
```
produces
```kdl
_children {
  - {
    output "a" {
      b "c"
    }
  }
}
```
for me instead of
```kdl
output "a" {
  b "c"
}
```

terlar:
It doesn't work but @quentinmit is aware of it and said it would be an easy fix:
https://github.com/NixOS/nixpkgs/pull/426828#issuecomment-3111754732

evysgarden:
> Have I missed something or does this just not work for toplevel children?

not sure whether there has been any work on this. but I stumbled upon this issue while adding a home-manager module for niri as well and fixed it downstream.

prbly not the most elegant solution but it works:

```nix
convertAttributeToKDL =
    name: value:
   ...
    else if vType == "list" then
      if name == "_children" then
        concatStringsSep "\n" (
          map (lib.flip lib.pipe [
            (mapAttrsToList convertAttributeToKDL)
            (concatStringsSep "\n")
          ]) value
        )
    ...
```

PR #7435, opened 2025-07-11
starship: set `STARSHIP_CONFIG` var and add option to set config path

rhoriguchi:
### Description

The reason for doing this is so that when you use `sudo su`, you can pass that env variable to the root shell to have the same terminal prompt as the current user.

Docs: <https://starship.rs/config/#config-file-location>

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/starship.nix
+++ b/modules/programs/starship.nix
@@ -82,13 +82,26 @@ in
         https://starship.rs/advanced-config/#transientprompt-and-transientrightprompt-in-cmd
       '';
     };
+
+    configPath = mkOption {
+      type = lib.types.str;
+      default = "${config.xdg.configHome}/starship.toml";
+      defaultText = lib.literalExpression "\${config.xdg.configHome}/starship.toml";
+      description = ''
+        Relative path to the user's home directory where the Starship config should be stored.
+      '';
+    };
   };
 
   config = mkIf cfg.enable {
-    home.packages = [ cfg.package ];
+    home = {
+      packages = [ cfg.package ];
 
-    xdg.configFile."starship.toml" = mkIf (cfg.settings != { }) {
-      source = tomlFormat.generate "starship-config" cfg.settings;
+      sessionVariables.STARSHIP_CONFIG = cfg.configPath;
+
+      file.${cfg.configPath} = mkIf (cfg.settings != { }) {
+        source = tomlFormat.generate "starship-config" cfg.settings;
+      };
     };
 
     programs.bash.initExtra = mkIf cfg.enableBashIntegration ''
--- a/tests/modules/programs/starship/settings.nix
+++ b/tests/modules/programs/starship/settings.nix
@@ -46,6 +46,11 @@
   };
 
   nmt.script = ''
+    sessionVarsFile=home-path/etc/profile.d/hm-session-vars.sh
+    assertFileExists $sessionVarsFile
+    assertFileContains $sessionVarsFile \
+        'export STARSHIP_CONFIG="/home/hm-user/.config/starship.toml"'
+
     assertFileContent \
       home-files/.config/starship.toml \
       ${./settings-expected.toml}


modules/programs/starship.nix
```diff
@@ -85,7 +85,11 @@ in
   };
 
   config = mkIf cfg.enable {
-    home.packages = [ cfg.package ];
+    home = {
+      packages = [ cfg.package ];
+
+      sessionVariables.STARSHIP_CONFIG = "${config.xdg.configHome}/starship.toml";
```
khaneliman: Makes me wonder if it would be worth creating an option, similar to we did in [docker-cli](https://github.com/nix-community/home-manager/pull/7411/files#diff-4af3e473b841c3256ff98005885ee5658038a1fd6c77948370de1f102fdbcb7fR24-R30) to support customizing the location, if someone wants it somewhere else.

rhoriguchi: Good idea, I've updated it.

modules/programs/starship.nix
```diff
@@ -82,13 +82,25 @@ in
         https://starship.rs/advanced-config/#transientprompt-and-transientrightprompt-in-cmd
       '';
     };
+
+    configPath = mkOption {
+      type = lib.types.str;
+      default = ".config/starship.toml";
```
khaneliman:
```suggestion
      default = "${config.xdg.configHome}/starship.toml";
```

rhoriguchi: Isn't this an absolute path to the user's home?

khaneliman: https://github.com/nix-community/home-manager/blob/ae62fd8ad8347e6bb5b615057f39f33d595a1c47/modules/misc/xdg.nix#L70-L79 it's what we use to hold the actual xdg config home path in case someone changes its location.

khaneliman: It is, but it gets handled properly from the activation script. We do it elsewhere in the codebase, too. 
https://github.com/nix-community/home-manager/blob/ae62fd8ad8347e6bb5b615057f39f33d595a1c47/modules/misc/xdg.nix#L70-L79

Can test though to double check. But, we have several uses of `home.file.${configDir}` that defaults to a `config.xdg.configHome`

rhoriguchi: Seems not to work

```
> cat /etc/profiles/per-user/rhoriguchi/etc/profile.d/hm-session-vars.sh
...
export STARSHIP_CONFIG="/home/rhoriguchi//home/rhoriguchi/.config/starship.toml"
...
```

rhoriguchi: It works for the home file location, but it will screw up the env variable.

rhoriguchi: Works for the `home.file` option. Had to use `lib.replaceStrings` for the env variable.

rhoriguchi: This seems to break tests, since you need to have a state version. `if then` case for a certain `stateVersion`?

khaneliman:
Not sure what issues you were running into with the session variable, but this seems to work fine. To avoid docs issues we can just move the default into the config block.

```diff
diff --git a/modules/programs/starship.nix b/modules/programs/starship.nix
index 945332d4b..23c78f085 100644
--- a/modules/programs/starship.nix
+++ b/modules/programs/starship.nix
@@ -85,7 +85,7 @@ in
 
     configPath = mkOption {
       type = lib.types.str;
-      default = "${config.xdg.configHome}/starship.toml";
+      defaultText = lib.literalExpression "\${config.xdg.configHome}/starship.toml";
       description = ''
         Relative path to the user's home directory where the Starship config should be stored.
       '';
@@ -96,15 +96,15 @@ in
     home = {
       packages = [ cfg.package ];
 
-      sessionVariables.STARSHIP_CONFIG = "${config.home.homeDirectory}/${
-        lib.replaceStrings [ config.home.homeDirectory ] [ "" ] cfg.configPath
-      }";
+      sessionVariables.STARSHIP_CONFIG = cfg.configPath;
 
       file.${cfg.configPath} = mkIf (cfg.settings != { }) {
         source = tomlFormat.generate "starship-config" cfg.settings;
       };
     };
 
+    programs.starship.configPath = lib.mkOptionDefault "${config.xdg.configHome}/starship.toml";
+
     programs.bash.initExtra = mkIf cfg.enableBashIntegration ''
       if [[ $TERM != "dumb" ]]; then
         eval "$(${lib.getExe cfg.package} init bash --print-full-init)"
diff --git a/tests/modules/programs/starship/settings.nix b/tests/modules/programs/starship/settings.nix
index 29688b099..3c7529946 100644
--- a/tests/modules/programs/starship/settings.nix
+++ b/tests/modules/programs/starship/settings.nix
@@ -46,6 +46,11 @@
   };
 
   nmt.script = ''
+    sessionVarsFile=home-path/etc/profile.d/hm-session-vars.sh
+    assertFileExists $sessionVarsFile
+    assertFileContains $sessionVarsFile \
+        'export STARSHIP_CONFIG="/home/hm-user/.config/starship.toml"'
+
     assertFileContent \
       home-files/.config/starship.toml \
       ${./settings-expected.toml}

```

0xda157:
> To avoid docs issues we can just move the default into the config block.

You don't have to move to config, adding defaultText is enough (default doesn't get evaled by docs generation if defaultText exists).

rhoriguchi:
Thank you for the inputs @khaneliman

PR #7444, opened 2025-07-12
swayidle: improve confusing "events" configuration

pks-t:
### Description

With swayidle one can configure two different kinds of hooks:

  - Idle timeouts are executed after the session has been idle for a specific amount of time.

  - Events are executed when systemd notifies us that for example the user session is locked or that the device is about to suspend.

While not obvious, there is a significant difference between how these two kinds are configured: there can be several timeouts with separate commands to be executed, but each event can only be specified once. If an event is specified multiple times, then the last command wins.

This can be very easy to miss in swayidle's documentation. Furthermore, because the config is a list of `{ event = "..."; command = "..."; }` attrset, we double down on this confusion and make it seem like having multiple handlers for an event was actually supported.

Fix this by converting from a list of "event" submodules to an attrset where the key is the event name and the value is the command to be executed. This makes it impossible to specify multiple commands for a single event by accident.

If a user _does_ want to have multiple commands executed on any event they can for example use `pkgs.writeShellScript` and manually chain the commands in that script.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@c0deaddict

### DIFF
--- a/modules/services/swayidle.nix
+++ b/modules/services/swayidle.nix
@@ -42,21 +42,30 @@ in
         };
       };
 
-      eventModule = {
+      eventsModule = {
         options = {
-          event = mkOption {
-            type = types.enum [
-              "before-sleep"
-              "after-resume"
-              "lock"
-              "unlock"
-            ];
-            description = "Event name.";
+          before-sleep = mkOption {
+            type = types.nullOr types.str;
+            default = null;
+            description = "Command to run before suspending.";
           };
 
-          command = mkOption {
-            type = types.str;
-            description = "Command to run when event occurs.";
+          after-resume = mkOption {
+            type = types.nullOr types.str;
+            default = null;
+            description = "Command to run after resuming.";
+          };
+
+          lock = mkOption {
+            type = types.nullOr types.str;
+            default = null;
+            description = "Command to run when the logind session is locked.";
+          };
+
+          unlock = mkOption {
+            type = types.nullOr types.str;
+            default = null;
+            description = "Command to run when the logind session is unlocked.";
           };
         };
       };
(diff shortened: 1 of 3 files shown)


c0deaddict (approved):

teto:
This is breaking bacwards compat. 
It seems like the user is likely to have to edit the file eventually so we dont need to go hard with backwards compatibility code but maybe throw an error explaining what to do / linking this PR when you detect a list (aka the old format). We dont want to silently disable swaylock by ignoring the old config either.
can you rebase please ? I will give it a spin 

cc @khaneliman for additionnal insight.

pks-t:
> It seems like the user is likely to have to edit the file eventually so we dont need to go hard with backwards compatibility code but maybe throw an error explaining what to do / linking this PR when you detect a list (aka the old format).

@teto I can backfill such logic if you want me to. Only question is whether to do it via an assert or via a `check` function for the specific module? Or is there a better way to do it?

> can you rebase please?

Done now, thanks!

khaneliman:
Should be simple enough to user a coercion function that throws a warning about deprecated usage of `list` for your configuration and automatically transforms it to fulfill the new attribute set based type.

pks-t:
> Should be simple enough to user a coercion function that throws a warning about deprecated usage of list for your configuration and automatically transforms it to fulfill the new attribute set based type.

@khaneliman Ah, good point. I've done that now, thanks!

khaneliman:
Appreciate the backwards compatible change, now.

alex-robbins:
Looks like you [forgot to change](https://github.com/nix-community/home-manager/pull/7444/files#diff-3e71ed5cbab7fee8058ddd0a480b6e022ba0baa554ac92ff8e766c91532c5853R111) the default for `services.swayidle.events` and now a warning is emitted for the default configuration:
```
evaluation warning: The syntax of services.swayidle.events has changed. While it
                    previously accepted a list of events, it now accepts an attrset
                    keyed by the event name.
```

The warning goes away if you manually set `services.swayidle.events = { }`.

pks-t:
@alex-robbins Ineed. This got fixed via https://github.com/nix-community/home-manager/pull/8315 now. Thanks!

PR #7466, opened 2025-07-13
gurk-rs: add module

awwpotato:
### Description
signal tui client

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/gurk-rs.nix
+++ b/modules/programs/gurk-rs.nix
@@ -1,0 +1,55 @@
+{
+  lib,
+  pkgs,
+  config,
+  ...
+}:
+let
+  tomlFormat = pkgs.formats.toml { };
+  cfg = config.programs.gurk-rs;
+in
+{
+  meta.maintainers = [ lib.maintainers.awwpotato ];
+
+  options.programs.gurk-rs = {
+    enable = lib.mkEnableOption "gurk-rs";
+    package = lib.mkPackageOption pkgs "gurk-rs" { nullable = true; };
+
+    settings = lib.mkOption {
+      inherit (tomlFormat) type;
+      default = { };
+      description = ''
+        Configuration written to {file}`$XDG_CONFIG_HOME/.config/gurk/gurk.toml`
+        or {file}`Library/Application Support/gurk/gurk.toml`. Options are
+        declared at <https://github.com/boxdot/gurk-rs/blob/main/src/config.rs>.
+        Note that `data_path` and `signal_db_path` should be set.
+      '';
+      example = lib.literalExpression ''
+        {
+          data_path = "/home/USERNAME/.local/share/gurk/gurk.data.json";
+          signal_db_path = "/home/USERNAME/.local/share/gurk/signal-db";
+          first_name_only = false;
+          show_receipts = true;
+          notifications = true;
+          bell = true;
+          colored_messages = false;
+          default_keybindings = true;
+          user = {
+            name = "MYNAME";
+            phone_number = "MYNUMBER";
+          };
+          keybindings =  { };
+        }
+      '';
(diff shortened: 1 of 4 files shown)


modules/programs/gurk-rs.nix
```diff
@@ -0,0 +1,58 @@
...
+        }
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    home.package = lib.mkIf (cfg.package != null) [ cfg.package ];
+
+    home.file."${
+      if pkgs.stdenv.hostPlatform.isDarwin && !config.xdg.enable then
```
khaneliman: If this a rust program, are we sure it respects `xdg` on darwin? Seems like the most common crate rust programs utilize doesn't respect it.

khaneliman: Yeah, looks like they use `dirs` which doesn't respect `xdg` on macos
https://docs.rs/dirs/3.0.2/dirs/fn.config_dir.html 

- https://github.com/boxdot/gurk-rs/blob/50636c70258a6b4411a22aa0fab07416aa260de6/src/config.rs#L284
- https://github.com/boxdot/gurk-rs/blob/50636c70258a6b4411a22aa0fab07416aa260de6/Cargo.toml#L46

0xda157: gurk uses [this function](https://docs.rs/dirs/3.0.2/dirs/fn.config_dir.html) to get the config dir, which does use `Library/Application Support` on darwin. EDIT: oops was thought you asked reverse of what you asked, I'll fix it

khaneliman: Yes, but you have a condition for `!config.xdg.enable`

clay53:
getting this error on current master branch (I don't have programs.gurk-rs enabled):
```
error:
         at /nix/store/bgl6ldj5ihbwcq8p42z3a0qzgqafgk2b-source/lib/attrsets.nix:1534:13:
         1533|           if length values == 1 || pred here (elemAt values 1) (head values) then
         1534|             head values
             |             ^
         1535|           else

         at /nix/store/bgl6ldj5ihbwcq8p42z3a0qzgqafgk2b-source/lib/modules.nix:1086:7:
         1085|     // {
         1086|       value = addErrorContext "while evaluating the option `${showOption loc}':" value;
             |       ^
         1087|       inherit (res.defsFinal') highestPrio;

       … while evaluating the option `system.build.toplevel':

       … while evaluating definitions from `/nix/store/bgl6ldj5ihbwcq8p42z3a0qzgqafgk2b-source/nixos/modules/system/activation/top-level.nix':

       … while evaluating the option `assertions':

       … while evaluating definitions from `/nix/store/ihz4q52pg5izk1cn8h6cw4b1k174sfqh-source/nixos/common.nix':

       error: The option `home-manager.users.clhickey.home.package' does not exist. Definition values:
       - In `/nix/store/ihz4q52pg5izk1cn8h6cw4b1k174sfqh-source/modules/programs/gurk-rs.nix':
           {
             _type = "if";
             condition = false;
             content = {
               _type = "if";
```

modules/programs/gurk-rs.nix
```diff
@@ -0,0 +1,55 @@
...
+            phone_number = "MYNUMBER";
+          };
+          keybindings =  { };
+        }
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    home.package = lib.mkIf (cfg.package != null) [ cfg.package ];
```
khaneliman, line 48:
```suggestion
    home.packages = lib.mkIf (cfg.package != null) [ cfg.package ];
```
oops how did this get through CI?

PR #7472, opened 2025-07-14
dunst: UWSM compatibility

fcoelhomrc:
Added Install field for systemd service, which is needed for autostart to work properly under compositors launched under UWSM

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/services/dunst.nix
+++ b/modules/services/dunst.nix
@@ -218,6 +218,9 @@ in
                 ]
             );
             Environment = lib.optionalString (cfg.waylandDisplay != "") "WAYLAND_DISPLAY=${cfg.waylandDisplay}";
+            Install = {
+              WantedBy = config.wayland.systemd.target;
+            };
           };
         };
       }


fcoelhomrc:
Did not follow the testing steps in checkpoint, I can figure out how to do them if really needed... but it is such a simple PR that I thought it wasn't worth it (given local tests + Github Actions tests)

SebastianStork:
I believe this wasn't tested enough... as it's now causing build failures. 

`Install.WantedBy` shouldn't be under the `Service` attribute, it is actually a top level attribute just like `Service`. Also, `WantedBy` expects a list and not a string.

denibertovic:
@SebastianStork @fcoelhomrc I'm experiencing the same issue:

```
A definition for option `systemd.user.services.dunst.Service.Install' is not of type `boolean or signed integer or string or absolute path or list of (boolean or signed integer or string or absolute path)'.
```

khaneliman:
Ugh.. missed that this went to 25.05 too.

PR #7476, opened 2025-07-15
neomutt: allow default email sending behaviour

me-and:
### Description

If neither neomutt.sendMailCommand nor passCmd are set, leave configuration for sending emails unset.  This will leave neomutt to manage sending emails itself, which matches the existing description of the `accounts.email.accounts.<name>.neomutt.sendMailCommand` configuration option.

This also removes the assertion added in #7414 / bd680a8c (neomutt: allow default email sending behaviour, 2025-07-12), since it is now valid to have neither passCmd nor neomutt.sendMailCommand on an account.

### Checklist

- ~[ ] Test cases updated/added. See [example](https://github.com/nix-community/home-manager/commit/f3fbb50b68df20da47f9b0def5607857fcc0d021#diff-b61a6d542f9036550ba9c401c80f00ef).~ There are no existing tests for `neomutt.sendMailCommand`; there probably ought to be some, but I don't want to delay this fix for writing those.

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

Paging @teto as the person who looked at #7414

### DIFF
--- a/modules/programs/neomutt/default.nix
+++ b/modules/programs/neomutt/default.nix
@@ -223,7 +223,7 @@ let
       {
         sendmail = "'${neomutt.sendMailCommand}'";
       }
-    else
+    else if passwordCommand != null then
       let
         smtpProto = if smtp.tls.enable && !smtp.tls.useStartTls then "smtps" else "smtp";
         smtpPort = if smtp.port != null then ":${toString smtp.port}" else "";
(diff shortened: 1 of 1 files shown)


me-and:
Also tagging @khaneliman and @trueNAHO as folk involved in the previous PR.

me-and:
I think my force-push to pick up the fix at #7478 accidentally removed the commit in this PR and auto-closed it, so I've now reopened it with the new commit.

PR #7479, opened 2025-07-15
direnv: fix silent clobbering global config; tests/direnv: refactor tests

khaneliman:
### Description

Fix a bug when using the `silent` option. 

Consolidate tests so we dont have to eval multiple times for things that can be asserted in the same configuration. 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/direnv.nix
```diff
@@ -132,7 +132,7 @@ in
             source = tomlFormat.generate "direnv-config" (
               cfg.config
               // lib.optionalAttrs (cfg.silent && isVersion236orHigher) {
-                global = {
+                global = (cfg.config.global or { }) // {
                   log_format = "-";
                   log_filter = "^$";
                 };
```
ambroisie: Alternative solution: use `mkIf` (and optionally `lib.mkDefault`, though IMO the user explicitly asked for the config so should use `mkForce` to override it). Should look something approximately like:

```nix
programs.direnv.config = {
  global = mkIf (cfg.silent && isVersion236orHigher) {
    log_format = "-";
    log_filter = "^$";
  };
};
```

khaneliman: Thanks, that feels nicer.

modules/programs/direnv.nix
```diff
@@ -126,18 +126,16 @@ in
...
       home.packages = [ cfg.package ];
 
+      programs.direnv.config = {
+        global = mkIf (cfg.silent && isVersion236orHigher) {
+          log_format = "-";
+          log_filter = "^$";
+        };
+      };
       xdg.configFile."direnv/direnv.toml" =
         mkIf (cfg.config != { } || (cfg.silent && isVersion236orHigher))
```
ambroisie: This became redundant.

```suggestion
        mkIf (cfg.config != { })
```

khaneliman: good catch, thanks again

PR #7483, opened 2025-07-15
iamb: using macos configuration directory when on darwin

heywoodlh:
### Description

Use the correct configuration directory when on MacOS for the iamb module

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@wikiker


modules/programs/iamb.nix
```diff
@@ -43,11 +45,23 @@ in
 
   };
 
-  config = lib.mkIf cfg.enable {
-    home.packages = lib.mkIf (cfg.package != null) [ cfg.package ];
+  config =
+    let
+      configFile = tomlFormat.generate "iamb-config" cfg.settings;
```
khaneliman, line 50:
```suggestion
      configFile = tomlFormat.generate "iamb-config" cfg.settings;
      configDir =
        if pkgs.stdenv.isDarwin then
          "Library/Application Support"
        else
          config.xdg.configHome;
```

Could do this, like we do in other modules, and then simplify the config block with one file generation condition using `home.file`

PR #7501, opened 2025-07-19
mkFirefoxModule: policies option work for darwin

auscyber:
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC
@rycee

### DIFF
--- a/modules/programs/firefox/default.nix
+++ b/modules/programs/firefox/default.nix
@@ -33,6 +33,7 @@ in
       };
       platforms.darwin = {
         configPath = "Library/Application Support/Firefox";
+        defaultsId = "org.mozilla.firefox.plist";
       };
     })
 
--- a/modules/programs/firefox/mkFirefoxModule.nix
+++ b/modules/programs/firefox/mkFirefoxModule.nix
@@ -274,6 +274,12 @@ in
       default = wrappedPackageName;
       description = "Name of the wrapped browser package.";
     };
+    darwinDefaultsId = mkOption rec {
+      type = types.nullOr types.str;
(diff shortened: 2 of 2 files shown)


modules/programs/firefox/mkFirefoxModule.nix
```diff
@@ -928,6 +935,7 @@ in
           _: profile:
           # Merge the regular profile settings with extension settings
           mkMerge [
+
```
khaneliman:
```suggestion
```

khaneliman (commented):
Could you rebase on latest master for CI.

modules/programs/firefox/default.nix
```diff
@@ -33,6 +33,7 @@ in
       };
       platforms.darwin = {
         configPath = "Library/Application Support/Firefox";
+        defaultsId = "org.mozilla.firefox.plist";
```
khaneliman, line 36: Doesn't seem to handle different versions of firefox. 

I had to change it to `org.nixos.firefoxdeveloperedition` to work in my setup.

auscyber: > Doesn't seem to handle different versions of firefox.
> 
> I had to change it to `org.nixos.firefoxdeveloperedition` to work in my setup.

well that would unfortunately hve to change per firefox version

auscyber:
sorry i kept having bizarre bizarre merge errors

modules/programs/firefox/default.nix
auscyber: i've now resolved it to be something defined in the mkFirefoxModule but also a user option that can be changed to allow for scenarios like yours

PR #7513, opened 2025-07-21
firefox: Add profile enable option

adisbladis:
### Description

Adds a `profiles.<name>.enable` option that allows disabling a declared profile.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@brckd @rycee @HPsaucii

### DIFF
--- a/modules/programs/firefox/mkFirefoxModule.nix
+++ b/modules/programs/firefox/mkFirefoxModule.nix
@@ -52,8 +52,10 @@ let
   # by future browser versions.
   extensionPath = "extensions/{ec8030f7-c20a-464f-9b0e-13a3a9e97384}";
 
-  profiles =
-    lib.flip lib.mapAttrs' cfg.profiles (
+  enabledProfiles = lib.filterAttrs (_: profile: profile.enable) cfg.profiles;
+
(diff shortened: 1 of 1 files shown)


bricked-dev:
Thank you for the contribution! What could be use cases for this?

adisbladis:
> Thank you for the contribution! What could be use cases for this?

I have a bunch of different browser profiles for different purposes that I activate on some (but not all) systems.
It's pretty nifty to be able to declare a profile which is default disabled:
``` nix
someProfile = {
  id = 1;
  enable = lib.mkDefault false;
}
```

And then activate it on a per system basis:
``` nix
programs.firefox.profiles.someProfile.enable = true;
```

I'm using this approach to compartmentalise extension installs & for getting my partners Firefox profile onto my user account on some systems.

bricked-dev:
You're right, that could be pretty useful! If I was you, Though you could also create an external option like `programs.firefox.profilePresets.someProfile,enable` for your usecase. I'm not sure if an `enable` option would be appropriate here, because I usually don't see it being used in submodules. It would have to be set to `true` by default for backwards compatibility.

PR #7527, opened 2025-07-23
xdg-terminal-exec: init

nukdokplex:
port xdg.terminal-exec from nixos to hm

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/misc/xdg-terminal-exec.nix
+++ b/modules/misc/xdg-terminal-exec.nix
@@ -1,0 +1,49 @@
+{
+  lib,
+  config,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.xdg.terminal-exec;
+in
+{
+  options = {
+    xdg.terminal-exec = {
+      enable = lib.mkEnableOption "xdg-terminal-exec, the [proposed](https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/46) Default Terminal Execution Specification";
+      package = lib.mkPackageOption pkgs "xdg-terminal-exec" { };
+      settings = lib.mkOption {
+        type = with lib.types; attrsOf (listOf str);
+        default = { };
+        description = ''
+          Configuration options for the Default Terminal Execution Specification.
+
+          The keys are the desktop environments that are matched (case-insensitively) against `$XDG_CURRENT_DESKTOP`,
+          or `default` which is used when the current desktop environment is not found in the configuration.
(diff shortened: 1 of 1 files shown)


modules/misc/xdg-terminal-exec.nix
```diff
@@ -0,0 +1,49 @@
...
+  ...
+}:
+let
+  cfg = config.xdg.terminal-exec;
+in
+{
+  options = {
+    xdg.terminal-exec = {
+      enable = lib.mkEnableOption "xdg-terminal-exec, the [proposed](https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/46) Default Terminal Execution Specification";
+      package = lib.mkPackageOption pkgs "xdg-terminal-exec" { };
```
khaneliman:
```suggestion
      package = lib.mkPackageOption pkgs "xdg-terminal-exec" { nullable = true; };
```

modules/misc/xdg-terminal-exec.nix
```diff
@@ -0,0 +1,49 @@
...
+            "com.raggesilver.BlackBox.desktop"
+            "org.gnome.Terminal.desktop"
+          ];
+        };
+      };
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    home.packages = [ cfg.package ];
```
khaneliman:
```suggestion
    home.packages = lib.mkIf (cfg.package != null) [ cfg.package ];
```

modules/misc/xdg-terminal-exec.nix
```diff
@@ -0,0 +1,49 @@
...
+}:
+let
+  cfg = config.xdg.terminal-exec;
+in
+{
+  options = {
+    xdg.terminal-exec = {
+      enable = lib.mkEnableOption "xdg-terminal-exec, the [proposed](https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/46) Default Terminal Execution Specification";
+      package = lib.mkPackageOption pkgs "xdg-terminal-exec" { };
+      settings = lib.mkOption {
```
khaneliman, line 15: Not sure of a better name atm, but it feels weird using `settings` just because its not a standard generator option.

nukdokplex: I think it would be better to leave this name to save connection between nixos and hm module

Issue #7250, opened 2025-06-10
New module: Niri compositor


asumbek:
### Description

A scrollable-tiling Wayland compositor.

- Homepage: https://github.com/YaLTeR/niri
- Documentation: https://github.com/YaLTeR/niri/wiki

Atemo-C:
- Relevant closed issue: https://github.com/nix-community/home-manager/issues/5532
_(Closed as "completed" because the user now uses a flake for bit, but it should probably have been closed with another reason, since the issue is still very much here. Oh, well.)_

I am also interested in a module to enable and configure Niri through Home Manager. Currently, I have a `home.file` option to write the configuration file where necessary, but this is not ideal.

For example, in my current Hyprland configuration, I split it and integrate it deeply into my NixOS configuration, such as having Hyprland's input option be configured in the relevant input modules, display configuration being configured in the same place where the boot and Waybar display configuration are, etc.

Being able to do so with Niri would open this possibility up in a much easier and neater way than to rely on `home.file`.

ambroisie:
Relevant work in Nixpkgs before a module can feasibly be added (or rather, its `settings` option): https://github.com/NixOS/nixpkgs/pull/412029.

terlar:
We already have somewhat working KDL support as part of Home Manager. I built a module some time ago:
https://github.com/terlar/nix-config/blob/main/modules/home/services/window-managers/niri.nix

But I wanted to take it for a spin for some time before upstreaming it.

asumbek:
Can we take an example with how Zellij module turns Nix expression into KDL format?

terlar:
Yes, this is what I did. I used this one:
https://github.com/nix-community/home-manager/blob/master/modules/lib/generators.nix#L63

The main issue is the non-support for multiple definitions of the same attribute, e.g. "spawn-at-startup".

This would be defined as:
```nix
{
  spawn-at-startup = "waybar";
}
```

But there seems to be no support for defining multiple ones. So for this case either extraConfig can be used or what I opted for was to define a spawnAtStartup option.

You can see it in action here:

https://github.com/terlar/nix-config/blob/main/modules/home/profiles/user/terje/desktop/niri/default.nix#L36

asumbek:
Oh, alright. I'll for your module to get up-streamed.

rachitvrma:
> Yes, this is what I did. I used this one: https://github.com/nix-community/home-manager/blob/master/modules/lib/generators.nix#L63
> 
[...]

PR #7534, opened 2025-07-24
niri: init module (playground for nixpkgs KDL format)

sodiboo:
### Description

- This is a playground for testing https://github.com/NixOS/nixpkgs/pull/426828.
- It's an example of what i'd consider to be a minimal and idiomatic KDL-configured module.

It is subject to change as the nixpkgs KDL format evolves. As i open this PR; i have *not* considered the implications of compatibility with niri-flake, because `wayland.windowManagers.niri` is a totally different namespace from the `programs.niri` i use. Realistically, i guess i can just stub out `wayland.windowManagers.niri` with aliases to my own modules.

Note that `xwayland-satellite` is one of the primary things we'd want to set a default for (letting niri locate it, *without* including that in `$PATH`), but this option does not exist in stable niri. It requires the latest git. The config is expected to fail validation, unless you use a newer niri package.

---

This should probably not be merged. I want to get the "KDL style" down first; *then* i think there should be a separate PR where i focus on just how to apply that to niri. In particular, a future PR can discuss which parts are worth including in such a niri module (and maybe naming of some options like `xcursor-size`, which i deviate from in niri-flake but not in this PR). For now, just accept the general set of options i expose and their naming.

Please focus on the *general structure of the KDL-related parts of the module*:
- The naming of `extraConfig`, `rendered`, `configFile`.
- The way i'm doing the rendering (it's much inspired by niri-flake at this moment).
- Does this kind of file hierarchy make sense? I like the idea of separating it into sections like this, with each one outputting a `rendered` option.

### Related

This doesn't close any of the following issues, but it's related enough that i want to create backlinks:

- https://github.com/NixOS/nixpkgs/issues/198655
- #7250
- https://github.com/sodiboo/niri-flake/issues/480
- https://github.com/nix-community/stylix/issues/1746

Also, a reminder for myself to cc ironbar on the "real" PR later on. They want a `niri-session.target`, i believe:

- https://github.com/JakeStanger/ironbar/pull/953#discussion_r2074117855

The semantics of installation and service management are *entirely out of scope* for this PR.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,16 +2,16 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1752950548,
-        "narHash": "sha256-NS6BLD0lxOrnCiEOcvQCDVPXafX1/ek1dfJHX1nUIzc=",
-        "owner": "NixOS",
+        "lastModified": 1753316869,
+        "narHash": "sha256-46I6xjQkZqk/Hvj5z6F/R69ppfs6KoRp3BL6mAwhkFY=",
+        "owner": "sodiboo",
         "repo": "nixpkgs",
-        "rev": "c87b95e25065c028d31a94f06a62927d18763fdf",
+        "rev": "88c9f09361a3591af14b1dd0d05d2f877c370cc3",
         "type": "github"
       },
       "original": {
-        "owner": "NixOS",
-        "ref": "nixos-unstable",
+        "owner": "sodiboo",
+        "ref": "ergonomic-kdl",
         "repo": "nixpkgs",
         "type": "github"
       }
--- a/flake.nix
+++ b/flake.nix
@@ -1,7 +1,7 @@
 {
   description = "Home Manager for Nix";
 
-  inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+  inputs.nixpkgs.url = "github:sodiboo/nixpkgs/ergonomic-kdl";
 
   outputs =
     {
--- a/modules/config/home-cursor.nix
+++ b/modules/config/home-cursor.nix
@@ -93,6 +93,8 @@ let
       sway = {
         enable = mkEnableOption "sway config generation for {option}`home.pointerCursor`";
       };
+
+      niri.enable = mkEnableOption "niri config generation for {option}`home.pointerCursor`";
     };
   };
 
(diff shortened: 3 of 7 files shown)


sodiboo:
> My main thought is that we already have a "KDL style" in home-manager, for better or worse, and adopting a second one or replacing the existing one both seem like they would cause unnecessary pain for users.

At least in the case of niri, *this style* is the one users are used to. There are a couple users who use it already, see [the end of this comment](https://github.com/NixOS/nixpkgs/pull/412029/#issuecomment-3090113652). It shouldn't "replace" the existing one in home-manager, and I don't expect the zellij module to adopt this style of configuration.

> I think it's painful for the `extraConfig` options to take lists - that means there's no way for a user to (say)
> 
> ```nix
> wayland.windowManager.niri.extraConfig.input.keyboard.xkb.layout = "no";
> ```
> 
> in one place, and
> 
> ```nix
> wayland.windowManager.niri.extraConfig.input.keyboard.xkb.numlock = {};
> ```
> 
> in another, in a way that's compatible with the options system's priority semantics.

Nonissue for niri. A comprehensive mapping exists, and [it allows doing](https://github.com/sodiboo/system/blob/25859a4e39cf5162181891b12de9d40d60e5ba6b/personal/niri.mod.nix#L384-L469) exactly [things like that](https://github.com/sodiboo/system/blob/25859a4e39cf5162181891b12de9d40d60e5ba6b/personal/transient-session/niri/run.mod.nix#L63-L65).

---

Note that this PR is primarily intended as an "example module" to see how a KDL-configured app can look, using the KDL format that is being discussed in https://github.com/NixOS/nixpkgs/pull/426828. I consider both of these comments to be "basically irrelevant" for this PR, since they're arguing against the basic list-of-nodes structure of that PR.

evysgarden:
@sodiboo heyo so turns out I've been doing the same (more or less) thing as you and have been mapping out the niri config to nix.

I also opened a draft pull request ( #8575  ) and sadly noticed only now that your pull-request and the niri-flake exist.

I'm gonna continue working on this since I'm almost done, just need to finish the descriptions and write some tests.

What's your status?
Should I stop, or do you wanna join forces?

I don't want to end up with a pull request that gets denied because yours is more suitable.

Brisingr05:
Since https://github.com/nix-community/home-manager/pull/9685 has been merged, this can probably be closed I assume?

PR #7535, opened 2025-07-24
docker-cli and yarn: fixups

FriedrichAltheide:
### Description

- docker-cli and yaar: Fix example text
- docker-cli: Set DOCKER_CONFIG env to the path of the config files instead of a single file
- docker-cli: Separate path to folder and config name
- docker-cli: Update tests

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/docker-cli.nix
+++ b/modules/programs/docker-cli.nix
@@ -21,18 +21,18 @@ in
   options.programs.docker-cli = {
     enable = mkEnableOption "management of docker client config";
 
-    configPath = mkOption {
+    configDir = mkOption {
       type = lib.types.str;
-      default = ".docker/config.json";
+      default = ".docker";
       description = ''
-        Relative path to the user's home directory where the Docker CLI settings should be stored.
+        Folder relative to the user's home directory where the Docker CLI settings should be stored.
       '';
     };
 
     settings = mkOption {
       type = jsonFormat.type;
       default = { };
-      example = ''
+      example = lib.literalExpression ''
         {
           "proxies" = {
             "default" = {
@@ -52,11 +52,11 @@ in
   config = mkIf cfg.enable {
     home = {
       sessionVariables = {
-        DOCKER_CONFIG = "${config.home.homeDirectory}/${cfg.configPath}";
+        DOCKER_CONFIG = "${config.home.homeDirectory}/${cfg.configDir}";
       };
 
       file = {
-        "${cfg.configPath}" = {
+        "${cfg.configDir}/config.json" = {
           source = jsonFormat.generate "config.json" cfg.settings;
         };
       };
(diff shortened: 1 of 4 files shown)


modules/programs/docker-cli.nix
```diff
@@ -23,16 +23,24 @@ in
...
+        Folder relative to the user's home directory where the Docker CLI settings should be stored.
+      '';
+    };
+
+    configFilename = mkOption {
+      type = lib.types.str;
+      default = "config.json";
+      description = ''
+        Filename of the docker CLI settings relative to `programs.docker-cli.configPath`.
       '';
```
khaneliman:
```suggestion
```

modules/programs/docker-cli.nix
```diff
@@ -23,16 +23,24 @@ in
 
     configPath = mkOption {
       type = lib.types.str;
-      default = ".docker/config.json";
+      default = ".docker";
       description = ''
-        Relative path to the user's home directory where the Docker CLI settings should be stored.
+        Folder relative to the user's home directory where the Docker CLI settings should be stored.
+      '';
+    };
```
khaneliman:
```suggestion
    configDir = mkOption {
      type = lib.types.str;
      default = ".docker";
      description = ''
        Folder relative to the user's home directory where the Docker CLI settings should be stored.
      '';
    };
```

modules/programs/docker-cli.nix
```diff
@@ -56,8 +64,8 @@ in
       };
 
       file = {
-        "${cfg.configPath}" = {
-          source = jsonFormat.generate "config.json" cfg.settings;
+        "${cfg.configPath}/${cfg.configFilename}" = {
+          source = jsonFormat.generate "${cfg.configFilename}" cfg.settings;
```
khaneliman:
```suggestion
        "${cfg.configDir}/config.json" = {
          source = jsonFormat.generate "docker-config.json" cfg.settings;
```

khaneliman (commented):
alternative, could just change from `configPath` to `configDir` maybe, for now, to fix the broken env variable and change the default behavior. new enough module. 

longterm, i think i need to expose a way for all modules to properly control the generated location of their files since this comes up a lot.

modules/programs/docker-cli.nix
```diff
@@ -21,18 +21,26 @@ in
...
+      '';
+    };
+
+    configFilename = mkOption {
+      type = lib.types.str;
+      default = "config.json";
+      description = ''
+        Filename of the docker CLI settings relative to `programs.docker-cli.configDir`.
       '';
     };
```
khaneliman: don't think there's really a benefit to this right? docker is still going to just search for `config.json` i thought. or does it just pick up any json files in the directory?

PR #7536, opened 2025-07-24
trippy: add forceUserConfig option

### LINKED ISSUE #7471: bug: trippy does not read user config file when run as root

### Issue description

when trippy is ran as root (as required to make it run) it will not use the users config file setup by this module.  it will use roots `$XDG_CONFIG_HOME/trippy/trippy.toml`

the [nixos upstream module](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/programs/trippy.nix) sets some permissions to not require to run it with sudo so it would use the users config file correctly.

my suggestion to fix is to create a wrapper to call `trip -c $XDG_CONFIG_HOME/trippy/trippy.conf $@` which would use the config file created by the hm module when ran with sudo

### Maintainer CC

@aguirre-matteo 

### System information

```markdown
fetching path input 'path:/nix/store/9hmjwrrn8k0cqyn3nwg3d8sn29ia4hhg-source'
this path will be fetched (0.02 MiB download, 0.06 MiB unpacked):
  /nix/store/nlqc1vm5psq2jjr5df6z25gkxkfqbvxx-stdenv-linux
copying path '/nix/store/nlqc1vm5psq2jjr5df6z25gkxkfqbvxx-stdenv-linux' from 'https://cache.nixos.org'...
 - system: `"x86_64-linux"`
 - host os: `Linux 5.15.167.4-microsoft-standard-WSL2, NixOS, 25.11 (Xantusia), 25.11.20250709.4c3a394`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Lix, like Nix) 2.91.3
System type: x86_64-linux
Features: gc, signed-caches
User configuration files: /home/sedlund/.config/nix/nix.conf:/etc/xdg/nix/nix.conf:/home/sedlund/.nix-profile/etc/xdg/nix/nix.conf:/nix/profile/etc/xdg/nix/nix.conf:/home/sedlund/.local/state/nix/profile/etc/xdg/nix/nix.conf:/etc/profiles/per-user/sedlund/etc/xdg/nix/nix.conf:/nix/var/nix/profiles/default/etc/xdg/nix/nix.conf:/run/current-system/sw/etc/xdg/nix/nix.conf
Store directory: /nix/store
State directory: /nix/var/nix
Data directory: /nix/store/3j073h93daccd3sk5nvm241sdpnd1aqz-lix-2.91.3/share`
 - nixpkgs: `/nix/store/9hmjwrrn8k0cqyn3nwg3d8sn29ia4hhg-source`
```

aguirre-matteo:
### Description

This pull request adds the option `programs.trippy.forceUserConfig`. Closes #7471.
### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/tests/modules/programs/trippy/example-config.nix
+++ b/tests/modules/programs/trippy/example-config.nix
@@ -1,5 +1,4 @@
 {
-
   programs.trippy = {
     enable = true;
     settings = {
--- a/tests/modules/programs/trippy/example-config.nix
+++ b/tests/modules/programs/trippy/example-config.nix
@@ -1,4 +1,5 @@
 {
+
   programs.trippy = {
     enable = true;
     settings = {
--- a/modules/programs/trippy.nix
+++ b/modules/programs/trippy.nix
@@ -6,6 +6,7 @@
 }:
 let
   inherit (lib)
+    types
     mkIf
     mkEnableOption
     mkPackageOption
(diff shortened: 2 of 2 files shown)


tests/modules/programs/trippy/example-config.nix
```diff
@@ -1,4 +1,5 @@
 {
+
```
khaneliman:
```suggestion
```

Intended or did other changes get lost?

aguirre-matteo: Sorry, it's just a mistake, I've added that newline by accident.

PR #7549, opened 2025-07-26
tmux: shortcut and prefix settings

comfysage:
### Description

adjusting the tmux lines for setting the prefix.
previously the prefix option would be set in the prefix key table, causing it to not register correctly.

the settings for the shortcut option would set an unrelated keybinding for the prefix.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/tmux.nix
+++ b/modules/programs/tmux.nix
@@ -84,23 +84,17 @@ let
     ''}
 
     ${
-      if cfg.prefix != null then
-        ''
-          # rebind main key: ${cfg.prefix}
-          unbind C-${defaultShortcut}
-          set -g prefix ${cfg.prefix}
-          bind -N "Send the prefix key through to the application" \
-            ${cfg.prefix} send-prefix
-        ''
-      else
-        optionalString (cfg.shortcut != defaultShortcut) ''
-          # rebind main key: C-${cfg.shortcut}
-          unbind C-${defaultShortcut}
-          set -g prefix C-${cfg.shortcut}
-          bind -N "Send the prefix key through to the application" \
-            ${cfg.shortcut} send-prefix
-          bind C-${cfg.shortcut} last-window
-        ''
+      let
+        defaultPrefix = "C-${defaultShortcut}";
+        prefix = if cfg.prefix != null then cfg.prefix else "C-${cfg.shortcut}";
+      in
+      optionalString (prefix != defaultPrefix) ''
+        # rebind main key: ${prefix}
+        unbind ${defaultPrefix}
+        set -g prefix ${prefix}
+        bind -n -N "Send the prefix key through to the application" \
+          ${prefix} send-prefix
+      ''
     }
 
     ${optionalString cfg.disableConfirmationPrompt ''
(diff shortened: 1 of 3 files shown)


modules/programs/tmux.nix
```diff
@@ -89,17 +89,16 @@ let
...
         optionalString (cfg.shortcut != defaultShortcut) ''
           # rebind main key: C-${cfg.shortcut}
           unbind C-${defaultShortcut}
           set -g prefix C-${cfg.shortcut}
-          bind -N "Send the prefix key through to the application" \
-            ${cfg.shortcut} send-prefix
-          bind C-${cfg.shortcut} last-window
+          bind -n -N "Send the prefix key through to the application" \
+            C-${cfg.shortcut} send-prefix
         ''
```
ambroisie: This is a lot of duplication at this point, how about something like:

```nix
  let
    prefix =
      if cfg.prefix != null then cfg.prefix
      else if cfg.shortcut != defaultShortcut then "C-${cfg.shortcut}"
      else null;
  in
  lib.optionalString (prefix != null) ''
    # rebind main key: ${prefix}
    unbind C-${defaultShortcut}
    set -g prefix ${prefix}
    bind -N "Send the prefix key through to the application" \
    bind -n -N "Send the prefix key through to the application" \
      ${prefix} send-prefix
  ''
```

ambroisie: Even better IMO, would be to make `shortcut` nullable, and set `prefix = "C-${cfg.shortcut}"` when it is non-null.

Makes it obvious to the user if they set both `shortcut` and `prefix` to conflicting values, rather than silently dropping one.

comfysage: ive simplified the code a bit to smt that makes sense to me :)

khaneliman:
Could you rebase to test latest ci

spasche:
This is a breaking change, previously with a configuration such as:
```
shortcut = "a";
```
Mapping was:
Ctrl+a a => goes to beginning of line
Ctrl+a Ctrl+a => goes to last window.

But now:
Ctrl+a a => ignored
Ctrl+a Ctrl+a => ignored

As a workaround, I added this back to my configuration, but that's no ideal IMHO:
```
      bind C-a last-window
      bind -N "Send the prefix key through to the application" \
        a send-prefix
```

comfysage:
@spasche 

yes this is breaking since it fixes a vital bug: the `last-window` command was only bound if the prefix was set using the `shortcut` option but not when the `prefix` option was used. nevertheless, there was no reason for this behavior so if you want the `last-window` keybind as you describe you'll have to add that to your config manually.

spasche:
Thanks @comfysage for the explanation.
I understand the reasoning for the `last-window` change and I'll add it to preserve the previous behavior.

I'm not sure to understand why the `-n` flag is added to the `bind` command:
```
set -g prefix C-a
bind -n -N "Send the prefix key through to the application" \
  C-a send-prefix
```
That `bind` command doesn't seem to have any effect in this case (pressing Ctrl+a does nothing in tmux or the shell).
I see the same behavior with or without it.

PR #7561, opened 2025-07-28
zsh: fix regressions on env var handling 

khaneliman:
### Description
Follow up to https://github.com/nix-community/home-manager/pull/6089#issuecomment-3124490554

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/zsh/lib.nix
```diff
@@ -31,7 +31,17 @@ rec {
   # given a relative (or unknown) path, returns absolute by prepending home dir
   # if path doesn't begin with "/"
   mkAbsPathStr =
-    pathStr: cleanPathStr ((lib.optionalString (!lib.hasPrefix "/" pathStr) "${homeDir}/") + pathStr);
+    pathStr:
+    let
+      cleanPath = lib.removeSuffix "/" pathStr;
+      hasShellVars = lib.hasPrefix "$" cleanPath || lib.hasInfix "$" cleanPath;
```
gepbird: Was this not simplified for performance or other reasons?
```suggestion
      hasShellVars = lib.hasInfix "$" cleanPath;
```

modules/programs/zsh/lib.nix
```diff
@@ -31,7 +31,17 @@ rec {
...
+    in
+    if hasShellVars then
+      # Does not escape shell variables, allowing them to be expanded at runtime
+      cleanPath
```
ambroisie: IMO this should issue a warning and be removed (maybe after a deprecation period). I don't see a good reason to support shell variables when Nix is used to manage the shell configuration.

andrevmatos:
Just tested here, I can confirm this works as expected for `zsh.history.path`;
`zsh.dotDir` didn't like receiving an env var though, I get an empty `~/.config/zsh` when I try, but I don't think this was working before either.

Regarding allowing or not shell variables, IMHO this should be upon user's discretion, and home-manager shouldn't prohibit that. 
AFAIU, it's mostly a tool to help build and manage config files; if this would be valid config (in this case, snippets interpolated into one's zshrc), then it could be supported in a best-effort basis.

gepbird:
With `programs.zsh.dotDir = ".config/zsh";` I get this warning, which is :+1: :
```
trace: evaluation warning: gep profile: Using relative paths in programs.zsh.dotDir is deprecated and will be removed in a future release.      
Consider using absolute paths or shell variables instead.                                                                                       
Current dotDir: .config/zsh    
```

But changing it to `programs.zsh.dotDir = "$HOME/.config/zsh";` won't work as expected, shell variables are not resolved:
```
dotfiles ❯ find $(nix-build /nix/store/rq89ignazz658747icxfimi8sfd5ymgk-home-manager-files.drv) | grep zsh
/nix/store/mmmwwsdh21glrsa2az5bnd2nsz31q16d-home-manager-files/'$HOME/.config/zsh'
/nix/store/mmmwwsdh21glrsa2az5bnd2nsz31q16d-home-manager-files/'$HOME/.config/zsh'/.zshenv
/nix/store/mmmwwsdh21glrsa2az5bnd2nsz31q16d-home-manager-files/'$HOME/.config/zsh'/.zshrc
/nix/store/mmmwwsdh21glrsa2az5bnd2nsz31q16d-home-manager-files/.zshenv
```

Using `programs.zsh.dotDir. = "${config.home-manager.users.gep.home.homeDirectory}/.config/zsh";` work well

khaneliman:
> But changing it to programs.zsh.dotDir = "$HOME/.config/zsh"; won't work as expected, shell variables are not resolved:

Yeah, i had it specific to a history path change, at first. Thought it might work with dotDir so just applied it to everything but didn't test it, yet. It does look like it won't support that, though.

modules/programs/zsh/default.nix
```diff
@@ -343,6 +345,30 @@ in
     in
     mkIf cfg.enable (
       lib.mkMerge [
+        {
+          assertions = [
+            {
+              assertion = !lib.hasInfix "$" cfg.dotDir;
```
khaneliman, line 351: Technically you can have a directory with `$` in the name... not sure why you would though.

modules/programs/zsh/lib.nix
khaneliman: Just flexibility to support generating valid `zsh` configurations.

modules/programs/zsh/default.nix
gepbird: That sounds like a weird use case, let's stick to protecting people from trying to use env vars as it is now

modules/programs/zsh/default.nix
```diff
@@ -343,6 +345,30 @@ in
...
+                programs.zsh.dotDir cannot contain shell variables as it is used for file creation at build time.
+                Current dotDir: ${cfg.dotDir}
+                Consider using an absolute path instead.
+              '';
+            }
+          ];
+
+          warnings =
+            lib.optionals
+              (cfg.dotDir != homeDir && !lib.hasPrefix "/" cfg.dotDir && !lib.hasInfix "$" cfg.dotDir)
```
gepbird, line 367: Nit: due to the above assertion it never has `$` infix
```suggestion
              (cfg.dotDir != homeDir && !lib.hasPrefix "/" cfg.dotDir)
```

modules/programs/zsh/default.nix
khaneliman: Don't want to deal with the assumption, just going to remove.

khaneliman: Hmm.. maybe you're right. I could add it back if you think it provides more value than risk...

modules/programs/zsh/default.nix
khaneliman: I was getting both the warning and assertion triggering, only wanted the assertion to fire. Maybe it was a test infra thing, though.

modules/programs/zsh/default.nix
gepbird: I meant I like having this assertion (which you didn't remove), because it prevents users from using `$` in `dotDir`, where most likely they wanted to use an env var instead of a `$` in the folder name. I think the test case you removed in https://github.com/nix-community/home-manager/compare/9ba29a1bc7f64b3803420372928c0d6f5989522c..6fba8803792b709e3a3c166f5b7aea480515d0b7 is not needed anymore, as `dotDir` can't contain a `$`

khaneliman: > I meant I like having this assertion (which you didn't remove), because it prevents users from using `$` in `dotDir`, where most likely they wanted to use an env var instead of a `$` in the folder name. I think the test case you removed in https://github.com/nix-community/home-manager/compare/9ba29a1bc7f64b3803420372928c0d6f5989522c..6fba8803792b709e3a3c166f5b7aea480515d0b7 is not needed anymore, as `dotDir` can't contain a `$`

The assertion was removed. That's just the latest force push cleaning up tests after yanking out the assertion.

khaneliman:
> Thanks, apart from a nit changes look good and did some testing with my previous zsh config :)
> 
> The `dotDir` change will affect quite a few users: https://grep.app/search?f.lang=Nix&regexp=true&q=dotDir+%3D+%22%5B%5E%2F%5D (not all results are using `home-manager`) I think for most people it won't take much effort to figure out that they can change it to `/home/yourusername/.config/zsh`. For those who not like hardcoding `yourusername` as a string (including myself), how can we get to `${config.home.homeDirectory}/.config/zsh` (or `${config.home-manager.users.yourusername.home.homeDirectory}/.config/zsh` for using home manager as a NixOS module)? Sorry if it's a bit off topic/noobish question, but I remember struggling with this for hours when starting with Nix. Should we mention this exact use case in the warning or link some documentation?

Oh, do you just mean some migration instructions for how to use the HM config to access use home path etc

gepbird:
> Oh, do you just mean some migration instructions for how to use the HM config to access use home path etc

Yes, one of the most common previous `dotDir` was `.config/zsh`

gepbird (approved):
I tested the old `dotDir` values, migration instructions are pretty good, thanks for the quick work!

romen:
@khaneliman The documentation for the dotDir option still states that it is supposed to be relative to the home directory. Shouldn't that be updated as well?

khaneliman:
> @khaneliman The documentation for the dotDir option still states that it is supposed to be relative to the home directory. Shouldn't that be updated as well?

Yes, we will need to update that.

PR #7564, opened 2025-07-28
hyprsunset: Migrate to v0.3 configuration files

kerfuzzle:
### Description
Updates the hyprsunset module so that it creates the hyprsunset.conf file introduced in v0.3.0.
This allows hyprsunset to manage the transition timings instead of defining extra systemd services/timers.
See https://wiki.hypr.land/Hypr-Ecosystem/hyprsunset/#configuration and https://github.com/hyprwm/hyprsunset/commit/f663619349f608e47af9de2d94f4eb264ff5294c

This does cause backwards compatibility issues with the previous transitions system used in the module, however, it feels more suitable to use the transition system provided by the program instead of creating one externally.

This is my first contribution so I am unsure how best to handle the backwards compatibility issues and I apologize if there are any glaring issues.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@khaneliman


kerfuzzle:
Decided to retain the existing services instead of migrating them as the new configuration file uses `HH:MM` for the transition timings so not all systemd calendar expressions would be able to be faithfully migrated.

Let me know if anything else can be improved. Merging this may also need to wait for v0.3 to make its way into nixpkgs stable.

khaneliman:
Looks a lot better, just rebased to clean up commit history.

khaneliman:
> Merging this may also need to wait for v0.3 to make its way into nixpkgs stable.

We are on v0.3.1 upstream right now.

PR #7566, opened 2025-07-28
nh: Add options for specific flakes

girlpunk:
### Description

nh provides an environment variable setting for defining the default flake, which is available as the programs.nh.flake setting. However, nh also provides three environment-specific settings for OS, home-manager, and Darwin. These are useful when, for one reason or another, different flakes are used for configuring the OS and the user's environment (e.g. on a shared machine).

This PR makes the other three settings available in a similar manner.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

  - [N/A] Added myself as module maintainer. See [example](https://github.com/nix-community/home-manager/blob/a51598236f23c89e59ee77eb8e0614358b0e896c/modules/programs/lesspipe.nix#L11).

#### Maintainer CC

@johnrtitor @a-jay98

### DIFF
--- a/modules/programs/nh.nix
+++ b/modules/programs/nh.nix
@@ -29,6 +29,42 @@ in
       '';
     };
 
+    osFlake = lib.mkOption {
+      type = lib.types.nullOr lib.types.str;
+      default = null;
+      description = ''
+        The string that will be used for the {env}`NH_OS_FLAKE` environment variable.
+
+        {env}`NH_OS_FLAKE` is used by nh as the default flake for performing {command}`nh os`
+        actions, such as {command}`nh os switch`.
+        Setting this will take priority over the `flake` option.
+      '';
+    };
+
+    homeFlake = lib.mkOption {
+      type = lib.types.nullOr lib.types.str;
+      default = null;
+      description = ''
+        The string that will be used for the {env}`NH_HOME_FLAKE` environment variable.
+
+        {env}`NH_HOME_FLAKE` is used by nh as the default flake for performing {command}`nh home`
+        actions, such as {command}`nh home switch`.
+        Setting this will take priority over the `flake` option.
+      '';
+    };
+
+    darwinFlake = lib.mkOption {
+      type = lib.types.nullOr lib.types.str;
+      default = null;
+      description = ''
+        The string that will be used for the {env}`NH_DARWIN_FLAKE` environment variable.
+
+        {env}`NH_DARWIN_FLAKE` is used by nh as the default flake for performing
+        {command}`nh darwin` actions, such as {command}`nh darwin switch`.
+        Setting this will take priority over the `flake` option.
+      '';
+    };
+
     clean = {
       enable = lib.mkEnableOption ''
         periodic garbage collection for user profile and nix store with nh clean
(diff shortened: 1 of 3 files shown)


modules/programs/nh.nix
```diff
@@ -64,19 +100,45 @@ in
...
+          message = "nh.osFlake must be a directory, not a nix file";
+        }
+        {
+          assertion = (cfg.homeFlake != null) -> !(lib.hasSuffix ".nix" cfg.homeFlake);
+          message = "nh.homeFlake must be a directory, not a nix file";
+        }
+        {
+          assertion = (cfg.darwinFlake != null) -> !(lib.hasSuffix ".nix" cfg.darwinFlake);
+          message = "nh.darwinFlake must be a directory, not a nix file";
+        }
```
0xda157: instead maybe just add examples to the options, something like
```nix
osFlake = lib.mkOption {
  # ...
  example = "/etc/nixos/";
};
```

modules/programs/nh.nix
```diff
@@ -64,19 +100,45 @@ in
...
+        ))
+        (lib.mkIf (cfg.osFlake != null) {
+          NH_OS_FLAKE = cfg.osFlake;
+        })
+        (lib.mkIf (cfg.homeFlake != null) {
+          NH_HOME_FLAKE = cfg.homeFlake;
+        })
+        (lib.mkIf (cfg.darwinFlake != null) {
+          NH_DARWIN_FLAKE = cfg.darwinFlake;
+        })
```
0xda157: personally I would do this, fine if you think it's not worth the extra complexity tho.
```nix
map (
  name:
  lib.mkIf (cfg."${name}Flake" != null) {
    "NH_${lib.toUpper name}_FLAKE" = cfg."${name}Flake";
  }
) [ "os" "home" "darwin" ]
```

modules/programs/nh.nix
A-jay98: I too think adding the example is sufficient. But I'm not against keeping the assertions either. If you choose to keep them perhaps you can rewrite them with a map and avoid the repetition. Something like:
 ```nix
map ( 
    name: { 
        assertion = cfg.${name} != null -> !(lib.hasSuffix ".nix" ${name});
        message = "nh.${name} must be a directory not a nix file";
    } 
) ["homeFlake" "darwinFlake" "osFlake"]
 ```

girlpunk: The assertions are included in NixOS's equivalent of the `FLAKE` setting, so I thought adding them here as well may be useful for some users?

awwpotato (approved):

modules/programs/nh.nix
```diff
@@ -64,19 +100,48 @@ in
...
+    assertions =
+      (lib.optionals pkgs.stdenv.isDarwin [
+        (lib.hm.darwin.assertInterval "programs.nh.clean.dates" cfg.clean.dates pkgs)
+      ])
+      ++
+        map
+          (name: {
+            assertion = (cfg.${name} != null) -> !(lib.hasSuffix ".nix" cfg.${name});
+            message = "nh.${name} must be a directory, not a nix file";
+          })
```
pedorich-n, line 112: Why is this assertion needed? It was removed before in https://github.com/nix-community/home-manager/pull/6470.

The reason I'm bringing it is because I use a folder called `config.nix`. So now my build fails. Since we don't have access to those paths from Nix, I don't think we should make assumptions.

PR #7568, opened 2025-07-29
television: put channel configs in correct location and cleanup

poperigby:
### Description

Channel configurations are now supposed to live in `~/.config/television/cable/<NAME>.toml`. I also cleaned up the derivation a bit to make it more readable.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

#### Maintainer CC

@awwpotato

### DIFF
--- a/modules/programs/television.nix
+++ b/modules/programs/television.nix
@@ -1,29 +1,49 @@
 {
+  config,
   lib,
   pkgs,
-  config,
   ...
 }:
 let
-  tomlFormat = pkgs.formats.toml { };
+  inherit (lib)
+    getExe
+    hm
+    literalExpression
+    maintainers
+    mapAttrs'
+    mkEnableOption
+    mkIf
+    mkMerge
+    mkOption
+    mkPackageOption
+    nameValuePair
+    ;
+
+  inherit (lib.types)
+    attrsOf
+    ;
+
   cfg = config.programs.television;
+  settingsFormat = pkgs.formats.toml { };
 in
 {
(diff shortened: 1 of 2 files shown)


khaneliman:
Please split refactoring from functional change

modules/programs/television.nix
```diff
@@ -1,29 +1,49 @@
...
+    mkPackageOption
+    nameValuePair
+    ;
+
+  inherit (lib.types)
+    attrsOf
+    ;
+
   cfg = config.programs.television;
+  settingsFormat = pkgs.formats.toml { };
```
0xda157, line 27: please don't rename to settingsFormat, tomlFormat is a more accurate name

modules/programs/television.nix
```diff
@@ -1,29 +1,49 @@
...
+    hm
+    literalExpression
+    maintainers
+    mapAttrs'
+    mkEnableOption
+    mkIf
+    mkMerge
+    mkOption
+    mkPackageOption
+    nameValuePair
```
0xda157, line 19: inheriting literally everything is very unnecessary, a good rule of thumb is only inherit it if it gets used 3 or 4 times.

modules/programs/television.nix
```diff
@@ -1,29 +1,49 @@
 {
+  config,
   lib,
   pkgs,
-  config,
```
0xda157, line 4: why? I prefer having lib and pkgs first.

awwpotato (changes requested):

PR #7575, opened 2025-07-29
treewide: add missing package options

khaneliman:
Add options to support more flexible module configurations.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/info.nix
```diff
@@ -41,11 +35,18 @@ in
...
   ];
 
-  options.programs.info.enable = lib.mkEnableOption "GNU Info";
+  options.programs.info = {
+    enable = lib.mkEnableOption "GNU Info";
+
+    # Installs this package -- the interactive just means that it
+    # includes the curses `info` program. We also use `install-info`
+    # from this package in the activation script.
+    package = lib.mkPackageOption pkgs "texinfoInteractive" { };
```
ambroisie: The documentation would read slightly better with:

```suggestion
    package = lib.mkPackageOption pkgs "texinfo" { default = "texinfoInteractive"; };
```

modules/services/barrier.nix
```diff
@@ -81,7 +83,7 @@ in
       Service.ExecStart =
         with cfg.client;
         toString (
-          [ "${pkgs.barrier}/bin/barrierc" ]
+          [ "${cfg.package}/bin/barrierc" ]
```
ambroisie: `lib.getExe'`?

modules/programs/powerline-go.nix
```diff
@@ -192,7 +194,7 @@ in
     # https://github.com/justjanne/powerline-go#fish
     programs.fish.interactiveShellInit = mkIf (cfg.enable && config.programs.fish.enable) ''
       function fish_prompt
-          eval ${pkgs.powerline-go}/bin/powerline-go -error $status -jobs (count (jobs -p))${commandLineArguments}
+          eval ${cfg.package}/bin/powerline-go -error $status -jobs (count (jobs -p))${commandLineArguments}
```
ambroisie: `lib.getExe`?

ambroisie: Same on the other ones.

modules/services/blueman-applet.nix
```diff
@@ -43,7 +48,7 @@
       };
 
       Service = {
-        ExecStart = "${pkgs.blueman}/bin/blueman-applet";
+        ExecStart = "${cfg.package}/bin/blueman-applet";
```
ambroisie: `lib.getExe'`?

modules/services/caffeine.nix
```diff
@@ -36,7 +38,7 @@ in
         ProtectSystem = "full";
         Type = "exec";
         Slice = "session.slice";
-        ExecStart = "${pkgs.caffeine-ng}/bin/caffeine";
+        ExecStart = "${cfg.package}/bin/caffeine";
```
ambroisie: Same suggestion.

modules/services/getmail.nix
```diff
@@ -19,6 +19,8 @@ in
     services.getmail = {
       enable = lib.mkEnableOption "the getmail systemd service to automatically retrieve mail";
 
+      package = lib.mkPackageOption pkgs "getmail6" { };
```
ambroisie:
```suggestion
      package = lib.mkPackageOption pkgs "getmail" { default = "getmail6"; };
```

khaneliman:
> Got fed up adding comments for all the `getExe`/`getExe'` spots, but they probably belong in most places.

This was mostly a catching of more modules missed during the last `package` treewide. I have another branch for standardizing `lib.getExe` so I didn't try and do it here. But, I probably should save myself some future effort and just do it in these since I haven't touched that in a long time.

khaneliman:
@adisbladis sorry, need to update workflow to not re-add people who removed themselves. 

Will manually add you and then remove you so the workflow doesn't tag you in subsequent force pushes.

khaneliman:
@ambroisie think i got em all

khaneliman:
oh ffs my script locally did that while testing lol

PR #7579, opened 2025-07-29
tests: chunk size 250 -> 20

khaneliman:
Just seeing how this affects CI rebuilds / speeds.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


khaneliman:
@zowoq from what you've seen monitoring buildbot. do you think more small chunks helps load balancing or fewer larger chunks? i was also hoping to benefit from fewer chunk rebuilds when one test area affects others. but maybe that doesn't matter for these symlink join chunks if the individual unit is cached, still. Also easier to find a test failure in a small log with smaller chunks.

zowoq:
Not easy to answer. If there aren't other factors like memory usage or eval time (like there was with nixvim) then really depends on what works best on the darwin CI runner. It has 10 cores, nix config is max-jobs = 10, cores = 2.

PR #7584, opened 2025-07-30
anki: set valid language default

trueNAHO:
### Description

Fixes: e8da7372fd1f ("anki: add module (#7274)")

This problem was discovered in a Stylix testbed: https://github.com/nix-community/stylix/pull/1801#discussion_r2243126248.

@JuneStepp, is there any particular reason `default = "en_US";` was not initially set?

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


JuneStepp:
> @JuneStepp, is there any particular reason default = "en_US"; was not initially set?

Anki has code to use the system language or en_US by default, buuutt it turns out that doesn't apply past the initial stages of startup, thus this error.

I hate that people not using English deal with so much friction but what can you do. Probably should have a default here since the possible values aren't well defined and intuitive.

modules/programs/anki/default.nix
```diff
@@ -23,9 +23,8 @@ in
     package = lib.mkPackageOption pkgs "anki" { };
 
     language = lib.mkOption {
-      type = with lib.types; nullOr str;
-      default = null;
-      example = "en_US";
```
JuneStepp, line 28: Can you add back in an example of your choice? "es_ES", "zh_CN", whatever.

modules/programs/anki/default.nix
```diff
@@ -23,9 +23,9 @@ in
...
+      type = lib.types.str;
```
JuneStepp:
```suggestion
      type = lib.types.nonEmptyStr;
```

modules/programs/anki/helper.nix
```diff
@@ -36,7 +36,7 @@ let
 
     # Shared options
 
-    lang_tag: str = "${toString cfg.language}"
+    lang_tag: str = "${cfg.language}"
     if lang_tag:
       profile_manager.setLang(lang_tag)
```
JuneStepp:
```suggestion
    profile_manager.setLang("${cfg.language}")
```

PR #7587, opened 2025-07-30
sesh: Add preview & use `fzf --tmux` instead of `fzf-tmux -p`

niksingh710:
### Checklist

	(After doing this the count of file changes were a lot)

    ```
    {component}: {description}

    {long description}
    ```


phucisstupid:
nice, merge please

khaneliman (commented):
Seems to be mixing in formatting, changing `fzf --tmux` change, `removing `args` from each keybind, etc... can you break up changes into separate commits so its easier to follow what's actually changing and why.

semi710:
@khaneliman Thanks for pointing out the `args` part — I had missed that earlier. I've now kept the diff minimal as well.

I'd like to get some community input: currently, `Ctrl-D` is used to delete the tmux session. Would it be okay if I change that to `Ctrl-Q` instead? I've generally seen `Ctrl-D` mapped to scroll the preview window in general.

phucisstupid:
> @khaneliman Thanks for pointing out the `args` part — I had missed that earlier. I've now kept the diff minimal as well.
> 
> I'd like to get some community input: currently, `Ctrl-D` is used to delete the tmux session. Would it be okay if I change that to `Ctrl-Q` instead? I've generally seen `Ctrl-D` mapped to scroll the preview window in general.

suggest this in sesh repo

semi710:
> Suggest this in the `sesh` repo

Yeah, great idea — I’ll do that. And if they accept it, we can update it properly in a future PR.
However, I don’t think they’ll agree to it, since *it’s just a snippet example* and is meant to be edited by the user anyway.

That’s why I asked here if we could make the change in Home Manager instead. Otherwise, I’ll just patch the string locally for myself — though it won’t be as clean.

Btw, the PR is ready to merge. The binding change can be addressed in another PR if even required.

Issue #7594, opened 2025-08-01
bug: `programs.television.channels` creates invalid files


Posiedon-Rises:
### Are you following the right branch?

- [x] My Nixpkgs and Home Manager versions are in sync

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Issue description

Currently `programs.television.channels` creates the file `~/.config/television/<name>-channels.toml`, but a television update makes it the standard to put files `~/.config/television/cable/<name>.toml`, and only allows one channel per file. Docs can be found here: https://alexpasmantier.github.io/television/docs/Users/channels.

The module should be updated to place files in the correct spot, as well as updating the docs to demonstrate how the module should be used, like this:
```nix
programs.television.channels.nix = {
  metadata.name = "nix";
  source.command = "nix-search-tv print";
  preview.command = "nix-search-tv preview {}";
};
```

### Maintainer CC

@awwpotato 

### System information

```markdown
- system: `"x86_64-linux"`
 - host os: `Linux 6.15.7-zen1, NixOS, 25.11 (Xantusia), 25.11.20250728.dc96378`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.28.4`
 - nixpkgs: `/nix/store/fy1zkcaf993w05ylfqx93b27am49i27b-source`
```

0xda157:
this will be fixed by https://github.com/nix-community/home-manager/pull/7568, we're just waiting for the pr author to respond to some feedback.

poseidon-rises:
Ah sorry, I only searched issues. I should have checked PRs as well, my bad.

0xda157:
this issue is not fixed yet so we should keep it open

PR #7601, opened 2025-08-02
television: update channel location

awwpotato:
### Description
taking the functional changes from https://github.com/nix-community/home-manager/pull/7568 (author hasn't responded to feedback after 5 days)
Closes: https://github.com/nix-community/home-manager/issues/7594
cc: @elanora96 @Posiedon-Rises 

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/television.nix
+++ b/modules/programs/television.nix
@@ -9,7 +9,10 @@ let
   cfg = config.programs.television;
 in
 {
-  meta.maintainers = [ lib.maintainers.awwpotato ];
+  meta.maintainers = with lib.maintainers; [
+    awwpotato
+    PopeRigby
+  ];
 
   options.programs.television = {
     enable = lib.mkEnableOption "television";
--- a/modules/programs/television.nix
+++ b/modules/programs/television.nix
@@ -43,29 +43,41 @@ in
       default = { };
       description = ''
         Each set of channels are written to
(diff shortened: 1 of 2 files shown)


tests/modules/programs/television/basic-config.nix
```diff
@@ -10,19 +10,18 @@
...
-          preview_command = "git show -p --stat --pretty=fuller --color=always {0}";
-        }
-        {
-          name = "my-dotfiles";
-          source_command = "fd -t f . $HOME/.config";
-          preview_command = "bat -n --color=always {0}";
-        }
-      ];
+    channels.git-log = {
+      metadata = {
```
poseidon-rises, line 14:
```suggestion
      metadata = {
        name = "git-log";
```

The tests expect the name to be provided; We should change this here to match the previous tests.

modules/programs/television.nix
```diff
@@ -9,7 +9,10 @@ let
   cfg = config.programs.television;
 in
 {
-  meta.maintainers = [ lib.maintainers.awwpotato ];
+  meta.maintainers = with lib.maintainers; [
+    awwpotato
+    PopeRigby
+  ];
```
khaneliman, line 15: separate commit, please

PR #7605, opened 2025-08-02
htop: add more fields

n0099:
### Description

sync with https://github.com/htop-dev/htop/blame/348c0a6bf4f33571835a0b6a1a0f5deb15132128/linux/ProcessField.h#L44-L53

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


rhoriguchi:
121 to 133 are off by one, all of them should be +1 https://github.com/htop-dev/htop/blob/cfb561fd342c9d5d814ac4965a6047442994ba94/linux/ProcessField.h#L48-L53

n0099:
It's by intentional as their stored value in `~/.config/htop/htoprc` is minus by one.
image

PR #7607, opened 2025-08-03
programs/nix-search-tv: init

poseidon-rises:
### Description

Created a module for [nix-search-tv](https://github.com/3timeslazy/nix-search-tv)  with options to enable and change settings. Opening as a draft, as I'm not sure if I'm creating `enableTelevisionIntegration` in the most optimal way.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/nix-search-tv.nix
+++ b/modules/programs/nix-search-tv.nix
@@ -1,0 +1,86 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+let
+  jsonFormat = pkgs.formats.json { };
+  cfg = config.programs.nix-search-tv;
+in
+{
+  meta.maintainers = with lib.hm.maintainers; [
+    poseidon-rises
+  ];
+
+  options.programs.nix-search-tv = {
+    enable = lib.mkEnableOption "nix-search-tv";
+    package = lib.mkPackageOption pkgs "nix-search-tv" { nullable = true; };
+
+    settings = lib.mkOption {
+      type = jsonFormat.type;
+      default = { };
+      description = ''
+        Configuration written to {file}`$XDG_CONFIG_HOME/nix-search-tv/config.json`.
+        See <https://github.com/3timeslazy/nix-search-tv?tab=readme-ov-file#configuration>
+        for the full list of options.
+      '';
+      example = lib.literalExpression ''
+        {
+          indexes = [ "nixpkgs" "home-manager" "nixos" ];
+
+          experimental = {
+            render_docs_indexes = {
+              nvf = "https://notashelf.github.io/nvf/options.html";
+            };
+          };
+        }
+      '';
+    };
+
+    enableTelevisionIntegration = lib.mkOption {
+      type = lib.types.bool;
+      default = config.programs.television.enable;
+      defaultText = lib.literalExpression "config.programs.television.enable";
+      description = ''
+        Enables integration with television. Creates a channel through
+        `programs.television.channels.nix-search-tv`, which are set as defaults
+        and can be overridden.
+        See [programs.television.channels](#opt-programs.television.channels)
+        for more information
+      '';
+      example = true;
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    home.packages = lib.mkIf (cfg.package != null) [ cfg.package ];
+
+    xdg.configFile."nix-search-tv/config.json" = lib.mkIf (cfg.settings != { }) {
+      source = jsonFormat.generate "config.json" cfg.settings;
+    };
+
+    programs.television.channels.nix-search-tv = lib.mkIf cfg.enableTelevisionIntegration (
+      let
+        path = lib.getExe cfg.package;
+      in
+      {
+        metadata = {
+          name = "nix-search-tv";
+          description = "Search nix options and packages";
+        };
+
+        source.command = "${path} print";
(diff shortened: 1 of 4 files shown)


poseidon-rises:
@awwpotato Thought you should get a look at this, seeing as you're the maintainer of the television module.

modules/programs/nix-search-tv.nix
```diff
@@ -0,0 +1,77 @@
...
+            };
+          };
+        }
+      '';
+    };
+
+    enableTelevisionIntegration = lib.mkOption {
+      type = lib.types.bool;
+      default = config.programs.television.enable;
+      defaultText = lib.literalExpression "\${config.programs.television.enable}";
```
0xda157:
```suggestion
      defaultText = lib.literalExpression "config.programs.television.enable";
```

modules/programs/nix-search-tv.nix
```diff
@@ -0,0 +1,77 @@
...
+  config = lib.mkIf cfg.enable {
+    home.packages = lib.mkIf (cfg.package != null) [ cfg.package ];
+
+    xdg.configFile."nix-search-tv/config.json" = lib.mkIf (cfg.settings != { }) {
+      source = jsonFormat.generate "config.json" cfg.settings;
+    };
+
+    programs.television.channels.nix-search-tv = lib.mkIf cfg.enableTelevisionIntegration (
+      let
+        path = if (cfg.package != null) then lib.getExe cfg.package else "nix-search-tv";
```
0xda157: instead just add an assertion that `cfg.enableTelevisionIntegration` requires `cfg.package` to not be null

modules/programs/nix-search-tv.nix
```diff
@@ -0,0 +1,77 @@
...
+      in
+      {
+        metadata.name = lib.mkDefault "nix-search-tv";
+        metadata.description = "Search nix options and packages";
```
0xda157:
```suggestion
        metadata = {
          name = "nix-search-tv";
          description = "Search nix options and packages";
        };
```

poseidon-rises:
Ok, if you think it looks good I'll open it up for review!

modules/programs/nix-search-tv.nix
```diff
@@ -0,0 +1,86 @@
...
+        path = lib.getExe cfg.package;
+      in
+      {
+        metadata = {
+          name = "nix-search-tv";
+          description = "Search nix options and packages";
+        };
+
+        source.command = lib.mkDefault "${path} print";
+        preview.command = lib.mkDefault "${path} preview {}";
```
trueNAHO: Why use `lib.mkDefault` instead of requiring end-users to explicitly override with `lib.mkForce`:

```suggestion
        source.command = "${path} print";
        preview.command = "${path} preview {}";
```

tests/modules/programs/nix-search-tv/television.nix
```diff
@@ -0,0 +1,22 @@
...
+      ${pkgs.writeText "settings-expected" ''
+        [metadata]
+        description = "Search nix options and packages"
+        name = "nix-search-tv"
+
+        [preview]
+        command = "${lib.getExe pkgs.nix-search-tv} preview {}"
+
+        [source]
+        command = "${lib.getExe pkgs.nix-search-tv} print"
```
trueNAHO, line 19: Can this string not be generated with `lib.generators.toINI`?

tests/modules/programs/nix-search-tv/basic-config.nix
```diff
@@ -0,0 +1,39 @@
...
+            "render_docs_indexed": {
+              "nvf": "https://notashelf.github.io/nvf/options.html"
+            }
+          },
+          "indexes": [
+            "home-manager",
+            "nixos",
+            "nixpkgs"
+          ]
+        }
```
trueNAHO, line 36: What about generating this string with `builtins.toJSON`?

trueNAHO (changes requested):

tests/modules/programs/nix-search-tv/television.nix
0xda157: > Can this string not be generated with `lib.generators.toINI`?

this is toml, also the standard in home-manager tests is using the final string instead of a generator for the expect text.

modules/programs/nix-search-tv.nix
poseidon-rises: I had a chat with someone else and that's what was recommended to me, if you think requiring `mkForce` is better behavior I'll gladly change it.

tests/modules/programs/nix-search-tv/basic-config.nix
poseidon-rises: As @awwpotato said, it's standard to use the final string for tests.

khaneliman:
Can you also clean up the commits, a bit. I'd expect a separate commit for adding yourself as a maintainer and a commit for the addition of the module.

poseidon-rises:
Should be done @khaneliman.

khaneliman:
> Should be done @khaneliman.

I see 10 commits, not 2. :P

poseidon-rises:
I assumes you just wanted me to split the first commit, that's my bad.

poseidon-rises:
Should be done (for real this time :P) @khaneliman.

PR #7608, opened 2025-08-03
fontconfig: add options for font rendering

bmrips:
### Description

Add options to the fontconfig module that allows to configure the font rendering, i.e. antialiasing, hinting, and subpixel rendering.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/misc/fontconfig.nix
```diff
@@ -17,7 +17,10 @@ let
 
 in
 {
-  meta.maintainers = [ lib.maintainers.rycee ];
+  meta.maintainers = with lib.maintainers; [
+    bmrips
+    rycee
+  ];
```
khaneliman, line 23: separate commit, please

modules/misc/fontconfig.nix
```diff
@@ -87,6 +90,33 @@ in
...
+          "none"
+          "rgb"
+          "bgr"
+          "vertical-rgb"
+          "vertical-bgr"
+        ];
+        default = "rgb";
+        description = "The sub-pixel rendering mode.";
+      };
+
```
khaneliman:
```suggestion
      antialiasing = lib.mkEnableOption "font antialiasing.";
      
      hinting = lib.mkOption {
        type = lib.types.enum [
          "none"
          "slight"
          "medium"
          "full"
        ];
        default = "none";
        description = "The font hinting mode.";
        example = "full";
      };
      
      subpixelRendering = lib.mkOption {
        type = lib.types.enum [
          "none"
          "rgb"
          "bgr"
          "vertical-rgb"
          "vertical-bgr"
        ];
        default = "none";
        description = "The sub-pixel rendering mode.";
      };
```

Not sure we should just be enabling some new functionality unexpectedly for users of this module.

modules/misc/fontconfig.nix
```diff
@@ -154,6 +184,36 @@ in
...
         '';
 
+        "fontconfig/conf.d/10-hm-rendering.conf".text =
+          let
+            set =
+              name: value:
+              let
+                xmlValue =
+                  if builtins.isBool value then
+                    "<bool>${if value then "true" else "false"}</bool>"
```
khaneliman:
```suggestion
                    "<bool>${lib.boolToString value}</bool>"
```

modules/misc/fontconfig.nix
```diff
@@ -154,6 +184,36 @@ in
...
+                    ${xmlValue}
+                  </edit>
+                </match>
+              '';
+          in
+          mkFontconfigConf ''
+            <description>Set the rendering mode</description>
+
+            ${set "antialias" cfg.antialiasing}
+            ${set "hinting" true}
```
khaneliman: Should probably be gated behind whether or not `cfg.hinting` is not `none` ?

modules/misc/fontconfig.nix
```diff
@@ -154,6 +184,36 @@ in
           <cachedir>${config.home.path}/lib/fontconfig/cache</cachedir>
         '';
 
+        "fontconfig/conf.d/10-hm-rendering.conf".text =
```
khaneliman: Would prefer to not generate anything if a user hasn't configured it. But, if we change defaults so that they don't cause an unexpected change in a system.. it should be fine.

modules/misc/fontconfig.nix
bmrips: Done.

modules/misc/fontconfig.nix
bmrips: That's right. With the new push, I set all options to null by default.

modules/misc/fontconfig.nix
bmrips: Done

modules/misc/fontconfig.nix
bmrips: I initially thought the same but found out that the `hintstyle = hintnone` is a valid assignment since it disables hinting altogether.

modules/misc/fontconfig.nix
bmrips: Right, I now only generate it if any rendering option is non-null.

PR #7612, opened 2025-08-04
less: configure options

bmrips:
### Description

- Generate `$XDG_CONFIG_HOME/lesskey` only when it is non-empty.
- Rename `programs.less.keys` to `programs.less.config` since its scope is
  broader than just keybindings.
- Add a module option to configure less options through an attrs.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/programs/less.nix
```diff
@@ -4,36 +4,65 @@
   pkgs,
   ...
 }:
+
 let
   cfg = config.programs.less;
+
```
khaneliman:
```suggestion
```

modules/programs/less.nix
```diff
@@ -4,36 +4,65 @@
   pkgs,
   ...
 }:
+
```
khaneliman:
```suggestion
```

modules/programs/less.nix
```diff
@@ -4,36 +4,65 @@
...
         type = lib.types.lines;
         default = "";
         example = ''
           s        back-line
           t        forw-line
         '';
         description = ''
-          Extra configuration for {command}`less` written to
-          {file}`$XDG_CONFIG_HOME/lesskey`.
+          Configuration for {command}`less`, written to {file}`$XDG_CONFIG_HOME/lesskey`.
```
khaneliman: Would prefer keeping it multine with `''` or just a single line `"`.

tests/modules/programs/less/custom-config.nix
```diff
@@ -1,18 +1,18 @@
+let
+  config = ''
+    s        back-line
+    t        forw-line
+  '';
+
```
khaneliman:
```suggestion
```

Pamplemousse:
LGTM :+1: 

Could you maybe add a test that adds both configuration and options, to make sure they play nice together in the resulting file?

bmrips:
> Could you maybe add a test that adds both configuration and options, to make sure they play nice together in the resulting file?

Sure!

PR #7613, opened 2025-08-04
grep: init module

bmrips:
### Description

Add a module for grep, mainly to configure its colours through `$$GREP_COLORS`>

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/grep.nix
+++ b/modules/programs/grep.nix
@@ -1,0 +1,37 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+let
+  cfg = config.programs.grep;
+
+in
+{
+
+  meta.maintainers = [ lib.maintainers.bmrips ];
+
+  options.programs.grep = {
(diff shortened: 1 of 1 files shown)


modules/programs/grep.nix
```diff
@@ -0,0 +1,34 @@
...
+  cfg = config.programs.grep;
+
+in
+{
+
+  meta.maintainers = [ lib.maintainers.bmrips ];
+
+  options.programs.grep = {
+    enable = lib.mkEnableOption "{command}`grep`.";
+    package = lib.mkPackageOption pkgs "gnugrep" { nullable = true; };
```
ambroisie: Presumably, one can use a different `grep` implementation than GNU's here. Or rather, the important part is that it is a `grep` package, not specifically a `gnugrep` one.

```suggestion
    package = lib.mkPackageOption pkgs "grep" { default = "gnugrep"; nullable = true; };
```

bmrips: Good one, thanks!

PR #7615, opened 2025-08-04
accounts.email: add 'davmail' flavor

bmrips:
### Description

Add `davmail` as an email account flavor. In order to do so, I needed to add an option to configure IMAP's and SMTP's authentication mechanism.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/accounts/email.nix
```diff
@@ -114,6 +114,26 @@ let
...
+        "clear"
+        "cram_md5"
+        "digest_md5"
+        "gssapi"
+        "login"
+        "ntlm"
+        "plain"
+        "xoauth2"
+      ]
+    );
```
khaneliman:
```suggestion
    type = types.nullOr (
      types.either types.str (
        types.enum [
          "anonymous"
          "apop"
          "clear"
          "cram_md5"
          "digest_md5"
          "gssapi"
          "login"
          "ntlm"
          "plain"
          "xoauth2"
        ]
      )
    );
```
Could make it freeform while using enum to guide a user, in case new auth types become available.

bmrips: Good idea!

PR #7620, opened 2025-08-05
🐛 fix VSCode profile dir handling on Darwin

dudeofawesome:
### Description

The VSCode module's `vscodeProfiles` activation script had 2 issues on Darwin.

1. The script executes in `/usr/local/bin`, so specifying the full path to the `Library` dir is necessary (rather than simply `Library/Application Support/…`.
2. The `mkdir` call had an unquoted variable, leading to errors when using paths with spaces (ie: `/Users/me/Library/Application Support/…`)

I've also got a branch with my changes applied to the `release-25.05` branch (`fix/vscode-profile-dir-creation-25.05`).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


dudeofawesome:
Will this get backported the the 25.05 release automatically? If not, what's the process to make that happen?

rycee:
@dudeofawesome Backports do not happen automatically. I have now cherry-picked these two commits into the release-25.05 branch so they should be available. Please give it a try.

PR #7629, opened 2025-08-06
keepassxc: add autostart option

bmrips:
### Description

Add an option to automatically start KeePassXC on login.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/keepassxc.nix
+++ b/modules/programs/keepassxc.nix
@@ -46,16 +46,36 @@ in
         for the full list of options.
       '';
     };
+
+    autostart = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      example = true;
+      description = ''
+        Whether to start Keepassxc automatically on login through the XDG autostart mechanism.
+      '';
+    };
   };
 
   config = lib.mkIf cfg.enable (
     lib.mkMerge [
 
       {
-        xdg.configFile = {
-          "keepassxc/keepassxc.ini" = lib.mkIf (cfg.settings != { }) {
-            source = iniFormat.generate "keepassxc-settings" cfg.settings;
-          };
+        assertions = [
+          {
+            assertion = cfg.autostart -> config.xdg.autostart.enable;
+            message = ''
+              {option}`xdg.autostart.enable` has to be enabled in order for
+              {option}`programs.keepassxc.autostart` to be effective.
+            '';
+          }
+        ];
+
+        xdg.autostart.entries = lib.mkIf cfg.autostart [
+          "${cfg.package}/share/applications/org.keepassxc.KeePassXC.desktop"
+        ];
+        xdg.configFile."keepassxc/keepassxc.ini" = lib.mkIf (cfg.settings != { }) {
+          source = iniFormat.generate "keepassxc-settings" cfg.settings;
         };
       }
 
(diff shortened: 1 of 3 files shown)


modules/programs/keepassxc.nix
```diff
@@ -43,16 +46,24 @@ in
         for the full list of options.
       '';
     };
+
+    autostart = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      example = true;
+      description = "Whether Keepassxc starts automatically on login.";
```
khaneliman: Should probably add some detail about how this is setup, in case a system isn't respecting `xdg.autostart` entries.

bmrips: Good idea! I just added a slightly more verbose description.

modules/programs/keepassxc.nix
```diff
@@ -43,16 +46,27 @@ in
...
       '';
     };
+
+    autostart = lib.mkOption {
+      type = lib.types.bool;
+      default = false;
+      example = true;
+      description = ''
+        Whether to start Keepassxc automatically on login through the XDG autostart mechanism.
+        Make sure to also enable {option}`xdg.autostart.enable`.
```
khaneliman: Actually, reminds me that we should have an assertion that enabling this requires that setting to be enabled.

bmrips: But what's the purpose of the `xdg.autostart.enable` flag then? I thought it is used to disable autostarting altogether.

Btw, why isn't `xdg.autostart.enable = true` the default? Because `xdg.autostart.entries` is empty by default, so enabling autostarting does not have any side-effect for users that don't care about it.

khaneliman: Well, if someone activates an option called `keepassxc.autostart = true;` i would think that means that it will autostart this program. But, if it doesn't have any effect unless `xdg.autostart.enable = true;` we add an assertion that says `You've tried to enable auto start for this, but it won't have an effect because you haven't enabled the functionality it's dependent on`

bmrips: Isn't that also an argument in favor of enabling Xdg autostart by default? Then, the functionality would work out of the box.

Additionally, I would rather use the `xdg.autostart.enable` flag to disable autostarting altogether when options like `<program>.autostart` are enabled.

khaneliman: We don't want unexpected side effects from a module, though. So, a warning/assertion is preferred so a user can explicitly opt in to a behavior from elsewhere. Just enabling this module's `autostart` doesn't mean I expect any other `xdg.autostart` mechanisms to be enabled that might be configured in a configuration. Similar to how we have configured most other cross module integrations. 

Even if it was a default enabled behavior, someone could opt out of the `xdg.autostart.enable` and we just let them know they have a config that doesn't make sense.

bmrips: As long as every `<program>.autostart` option is disabled by default, there is no side effect with Xdg autostart enabled by default; the user has to explicitly enable autostarting per program.

To conclude this PR, I have added your requested assertion. Is there any possibility to add a test for it?

khaneliman: >  Is there any possibility to add a test for it?

There are a handful of tests that exist to test assertions, you could just create another test case matching their strategy of expecting an assertion thrown with a config.

bmrips: Ok, should be done now.

PR #7630, opened 2025-08-06
git-credential-keepassxc: init module

bmrips:
### Description

Add a module for [git-credential-keepassxc](https://github.com/Frederick888/git-credential-keepassxc).

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/git-credential-keepassxc.nix
+++ b/modules/programs/git-credential-keepassxc.nix
@@ -1,0 +1,58 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+
+let
+  cfg = config.programs.git-credential-keepassxc;
+in
+{
+
+  meta.maintainer = [ lib.maintainers.bmrips ];
+
+  options.programs.git-credential-keepassxc = {
+    enable = lib.mkEnableOption "{command}`git-credential-keepassxc`.";
+    package = lib.mkPackageOption pkgs "git-credential-keepassxc" { };
+    groups = lib.mkOption {
+      type = with lib.types; nullOr (listOf str);
+      default = null;
+      example = "Git";
+      description = ''
+        The KeePassXC groups used for storing and fetching of credentials. By
+        default, the groups created by
+        {command}`git-credential-keepassxc configure [--group <GROUP>]` are used.
+      '';
+    };
+    hosts = lib.mkOption {
+      type = with lib.types; listOf str;
+      default = [ ];
+      example = [ "https://github.com" ];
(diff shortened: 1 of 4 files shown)


modules/programs/git-credential-keepassxc.nix
```diff
@@ -0,0 +1,58 @@
...
+        if cfg.groups == null then
+          "--git-groups"
+        else
+          lib.concatStringsSep " " (map (g: "--group ${g}") cfg.groups);
+    in
+    {
+      helper = "${cfg.package}/bin/git-credential-keepassxc ${groups}";
+    };
+
+in
```
khaneliman:
```suggestion
let
  cfg = config.programs.git-credential-keepassxc;
in
```

modules/programs/git-credential-keepassxc.nix
```diff
@@ -0,0 +1,58 @@
...
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    home.packages = [ cfg.package ];
+    programs.git.extraConfig.credential =
+      if cfg.hosts == [ ] then
+        helperConfig
+      else
+        lib.listToAttrs (map (host: lib.nameValuePair host helperConfig)) cfg.hosts;
```
khaneliman:
```suggestion
    programs.git.extraConfig.credential =
    let
      helperConfig =
        let
          groups =
            if cfg.groups == null then
              "--git-groups"
            else
              lib.concatStringsSep " " (map (g: "--group ${g}") cfg.groups);
        in
        {
          helper = "${cfg.package}/bin/git-credential-keepassxc ${groups}";
        };
    in
      if cfg.hosts == [ ] then
        helperConfig
      else
        lib.listToAttrs (map (host: lib.nameValuePair host helperConfig)) cfg.hosts;
```

khaneliman (commented):
Prefer keeping `let` variables scoped to usage, as much as possible.

modules/programs/git-credential-keepassxc.nix
```diff
@@ -0,0 +1,58 @@
...
+  pkgs,
+  ...
+}:
+
+let
+  cfg = config.programs.git-credential-keepassxc;
+in
+{
+
+  meta.maintainer = [ lib.maintainers.bmrips ];
```
NotAShelf, line 13: This should be `meta.maintainers`, currently breaks eval.

Could make a PR but I'm pretty positive someone can push a fix by the time I clone and prepare a PR.

khaneliman: Thanks raf, I'll push a quick fix.

PR #7639, opened 2025-08-07
codex: support XDG Base Directory specification

delafthi:
- Configuration file is now placed in XDG_CONFIG_HOME/codex/config.toml by default for versions >=0.2.0 when preferXdgDirectories is enabled.
- Falls back to ~/.codex/config.yaml for versions <0.2.0 and to ~/.codex/config.toml when preferXdgDirectories is disabled
- Sets CODEX_HOME environment variable to $XDG_CONFIG_HOME/codex when using XDG directories.
- Updated tests to verify XDG directory behavior and environment variable presence.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```

### DIFF
--- a/modules/programs/codex.nix
+++ b/modules/programs/codex.nix
@@ -15,7 +15,6 @@ let
   packageVersion = if cfg.package != null then lib.getVersion cfg.package else "0.2.0";
   isTomlConfig = lib.versionAtLeast packageVersion "0.2.0";
   settingsFormat = if isTomlConfig then tomlFormat else yamlFormat;
-  configFileName = if isTomlConfig then ".codex/config.toml" else ".codex/config.yaml";
 in
 {
   meta.maintainers = [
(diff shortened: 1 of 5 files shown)


khaneliman (commented):
Could be simplified a lot to just add a `configDir` variable that does the `config.homePreferXdgDirectories && isTomlConfig` check and sets to either `codex` or `${config.xdg.configHome}/codex` instead of introducing a mkMerge with duplicating code

modules/programs/codex.nix
```diff
@@ -63,16 +64,26 @@ in
...
-    home.file = {
-      "${configFileName}" = lib.mkIf (cfg.settings != { }) {
-        source = settingsFormat.generate "codex-config" cfg.settings;
-      };
-      ".codex/AGENTS.md" = lib.mkIf (cfg.custom-instructions != "") {
-        text = cfg.custom-instructions;
+  config =
+    let
+      useXdgDirectories = (config.home.preferXdgDirectories && isTomlConfig);
+      configDir = if useXdgDirectories then ".config/codex" else ".codex";
```
khaneliman:
```suggestion
      xdgConfigHome = lib.removePrefix config.home.homeDirectory config.xdg.configHome
      configDir = if useXdgDirectories then "${xdgConfigHome}/codex" else ".codex";
```

Think this will work to respect a configured xdg config home.

delafthi: Thanks, learned something new.

PR #7649, opened 2025-08-09
walker: add module

### LINKED ISSUE #7590: [New Module Request] Walker
### Program/Service Name

walker

### Upstream Repository

https://github.com/abenz1267/walker

### Configuration Documentation

https://github.com/abenz1267/walker/wiki/Basic-Configuration

### Reason for Addition

It's quite a widely used application launcher for Wayland. Preferably, we would also offer it as a service to be started in the background.

aguirre-matteo:
### Description

This pull request adds a module for configuring Walker. I modified a bit the [upstream module](https://github.com/abenz1267/walker/blob/master/nix/modules/home-manager.nix) to add an assertion and change the name of some options. Closes #7590.

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/services/walker.nix
```diff
@@ -0,0 +1,120 @@
...
+      description = "Whatever to enable Walker's Systemd Unit.";
+    };
+  };
+
+  config = mkIf cfg.enable (mkMerge [
+    {
+      assertions = [
+        (lib.hm.assertions.assertPlatform "services.walker" pkgs lib.platforms.linux)
+        {
+          assertion = if cfg.systemd.enable then (if cfg.package == null then false else true) else true;
```
khaneliman:
```suggestion
          assertion = cfg.systemd.enable -> (cfg.package == null);
```

Think this works. If systemd is enabled, package needs to be configured.

aguirre-matteo: I tried it and it works if we replace `==` with `!=`.

JonathanLorimer:
@aguirre-matteo The walker docs [suggest](https://benz.gitbook.io/walker/installation-and-launching#launching) that you need to have `elephant` enabled via systemd for walker to run correctly. Does this module work correctly without setting up an elephant systemd service?

aguirre-matteo:
When I was developing the module I tested walker on my PC and it worked. But now it seems they changed the upstream module to force walker.service to require elephant.service. It would be nice to double check if there's an issue in the latest version, but I'll be busy until next 2 weeks, so feel free to test it by yourself and tell me the result.

JonathanLorimer:
@aguirre-matteo It does seem to be working for me, but its very slow and hangs sometimes, so maybe requires elephant. I could potentially try and do a patch

aguirre-matteo:
I already opened an issue in walker's repo to add more maintainers to the module. Being honest, I just add initial support in Home-Manager for programs, but I'm not constantly looking up for changes since I don't use most of those tools.

JonathanLorimer:
FWIW It looks like the walker project provides its own home manager module https://github.com/abenz1267/walker/blob/master/nix/modules/home-manager.nix . So I think people can just use that!

aguirre-matteo:
Yep, I know that. But the idea was to add the module to the Home-Manager repo so it's part of it.

PR #7650, opened 2025-08-09
pizauth: init module

Swarsel:
### Description

Adds a module for [pizauth](https://github.com/ltratt/pizauth)

### Checklist

    ```
    {component}: {description}

    {long description}
    ```


modules/services/pizauth.nix
```diff
@@ -0,0 +1,194 @@
...
+  renderedAccounts = lib.concatStringsSep "\n" (
+    lib.mapAttrsToList (name: acc: renderAccount name acc) cfg.accounts
+  );
+
+  configFile = pkgs.writeTextFile {
+    name = "pizauth.conf";
+    text =
+      lib.optionalString (cfg.extraConfig != null && cfg.extraConfig != "") "${cfg.extraConfig}\n"
+      + renderedAccounts;
+  };
```
khaneliman: Able to leverage a generator or is the config a special format that doesn't have one?

modules/services/pizauth.nix
```diff
@@ -0,0 +1,194 @@
...
+  lib,
+  pkgs,
+  ...
+}:
+let
+  inherit (lib) mkOption types;
+
+  cfg = config.services.pizauth;
+
+  renderScopes =
```
khaneliman: Stylistically would prefer more of the let block be localized to usage instead of at top of file

modules/services/pizauth.nix
Swarsel: A rendered config could for example look like:

```
refresh_at_least = 15s;

account "test1" {
  auth_uri = "authUri1";
  token_uri = "tokenUri1";
  client_id = "clientId1";
  client_secret = "clientSecret1";
  login_hint = "testLogin1";
  redirectUri = "redirectUri1";
  refresh_retry = 30s;
}

account "test2" {
  auth_uri = "authUri2";
  token_uri = "tokenUri2";
  client_id = "clientId2";
  client_secret = "clientSecret2";
  scopes = [
    "scope1",
    "scope2"
  ];
}
```

I think the closest thing would be `toKeyValue` for parts of the account config, but I believe we would still need custom code for the `account "<name>"` part and the scopes (multiline lists) - but I am also not too familiar with the generators, so I might be missing something.

modules/services/pizauth.nix
```diff
@@ -0,0 +1,195 @@
...
+        );
+
+        configFile = pkgs.writeTextFile {
+          name = "pizauth.conf";
+          text =
+            lib.optionalString (cfg.extraConfig != null && cfg.extraConfig != "") "${cfg.extraConfig}\n"
+            + renderedAccounts;
+        };
+      in
+      configFile;
```
khaneliman:
```suggestion
      in
      pkgs.writeTextFile {
        name = "pizauth.conf";
        text =
          lib.optionalString (cfg.extraConfig != null && cfg.extraConfig != "") "${cfg.extraConfig}\n"
          + renderedAccounts;
      };
```

Can just reduce noise by not storing the final step in a variable just to be referenced by name immediately.

modules/services/pizauth.nix
```diff
@@ -0,0 +1,195 @@
...
+
+  };
+
+  config = lib.mkIf cfg.enable {
+    assertions = [ (lib.hm.assertions.assertPlatform "services.pizauth" pkgs lib.platforms.linux) ];
+
+    home.packages = [ cfg.package ];
+
+    xdg.configFile."pizauth.conf".source =
+      let
```
khaneliman, line 113:
```suggestion
      let
        indent = "  ";
```

modules/services/pizauth.nix
```diff
@@ -0,0 +1,195 @@
...
+        renderScopes =
+          scopes:
+          let
+            indent = "  ";
```
khaneliman:
```suggestion
```

modules/services/pizauth.nix
```diff
@@ -0,0 +1,195 @@
...
+            quoted = map (s: "\"${s}\"") scopes;
+            joined = lib.concatStringsSep ",\n${indent}${indent}" quoted;
+          in
+          "[\n${indent}${indent}${joined}\n${indent}]";
+
+        renderAccount =
+          let
+            indent = "  ";
+          in
```
khaneliman:
```suggestion
```

modules/services/pizauth.nix
```diff
@@ -0,0 +1,195 @@
...
+        Unit = {
+          Description = "Pizauth OAuth2 token manager";
+          After = [ "network.target" ];
+        };
+
+        Service = {
+          Type = "simple";
+          ExecStart = "${pkgs.pizauth}/bin/pizauth server -vvvv -d";
+          ExecReload = "${pkgs.pizauth}/bin/pizauth reload";
+          ExecStop = "${pkgs.pizauth}/bin/pizauth shutdown";
```
khaneliman:
```suggestion
          ExecStart = "${lib.getExe pkgs.pizauth} server -vvvv -d";
          ExecReload = "${lib.getExe pkgs.pizauth} reload";
          ExecStop = "${lib.getExe pkgs.pizauth} shutdown";
```

modules/services/pizauth.nix
```diff
@@ -0,0 +1,195 @@
...
+          name = "pizauth.conf";
+          text =
+            lib.optionalString (cfg.extraConfig != null && cfg.extraConfig != "") "${cfg.extraConfig}\n"
+            + renderedAccounts;
+        };
+      in
+      configFile;
+
+    systemd.user.services = {
+      pizauth = {
```
khaneliman:
```suggestion
    systemd.user.services.pizauth = {
```

Could flatten this f we're only dealing with a single service.

modules/services/pizauth.nix
```diff
@@ -0,0 +1,195 @@
...
+    accounts = mkOption {
+      type = types.attrsOf (
+        types.submodule {
+          options = {
+            name = mkOption {
+              type = types.str;
+              readOnly = true;
+              description = ''
+                    Unique identifier of the account. This is set to the
+                attribute name of the account configuration.
```
khaneliman:
```suggestion
                Unique identifier of the account. This is set to the
                attribute name of the account configuration.
```

modules/services/pizauth.nix
```diff
@@ -0,0 +1,195 @@
...
+              description = ''
+                The OAuth2 client secret.
+              '';
+            };
+
+            scopes = mkOption {
+              type = types.nullOr (types.listOf types.str);
+              description = ''
+                      The scopes which the OAuth2 token will give access to. Optional.
+                Note that Office365 requires the non-standard "offline_access" scope to be specified in order for pizauth to be able to operate successfully.
```
khaneliman:
```suggestion
                The scopes which the OAuth2 token will give access to. Optional.
                Note that Office365 requires the non-standard "offline_access" scope to be specified in order for pizauth to be able to operate successfully.
```

modules/services/pizauth.nix
```diff
@@ -0,0 +1,195 @@
...
+  inherit (lib) mkOption types;
+
+  cfg = config.services.pizauth;
+in
+{
+  meta.maintainers = [ lib.hm.maintainers.swarsel ];
+
+  options.services.pizauth = {
+    enable = lib.mkEnableOption ''
+      Pizauth, a commandline OAuth2 authentication daemon.
```
khaneliman:
```suggestion
      Pizauth, a commandline OAuth2 authentication daemon
```
`mkEnableOption` will already include an ending punctuation.

modules/services/pizauth.nix
```diff
@@ -0,0 +1,195 @@
...
+                "https://outlook.office365.com/SMTP.Send"
+                "offline_access"
+              ];
+            };
+
+            loginHint = mkOption {
+              type = types.nullOr types.str;
+              default = null;
+              description = ''
+                The login hint for the account provider. Optional.
```
khaneliman:
```suggestion
                An optional login hint for the account provider.
```

